Issue #47, opened 2026-04-13
fix: merge gate hooks don't catch gh api .../merge bypass


atlas-apex:
## Problem

The `block-merge-on-red-ci.sh`, `block-unreviewed-merge.sh`, and `require-design-review-for-ui.sh` hooks only trigger on `Bash(gh pr merge *)`. When merging via `gh api repos/.../pulls/.../merge -X PUT`, the hooks don't fire because the command doesn't match the pattern.

This was used as a workaround for the cross-repo resolution issue (#45), but it silently bypasses all merge gates — including the CI-must-be-green rule.

## Incident

PR me2resh/curios-dog#190 was merged via `gh api` while CI was still running.

## Proposed fix

Add a second `if` matcher for the API bypass pattern:

```json
{
  "type": "command",
  "if": "Bash(gh api */merge*)",
  "command": "..."
}
```

Or better: add a single catch-all merge hook that matches both patterns and delegates to the existing scripts.

Also: document that `gh api` bypass is a rule violation, same as forging approval markers.

## Related

- #45 — hooks multi-repo support (the reason the API bypass was used)

atlas-apex:
Fixed by PR #54 (merged 2026-04-14). Both `gh pr merge` and `gh api .../merge` shapes are now gated by the three merge-gate hooks via shared `_lib-extract-pr.sh`.

PR #54, opened 2026-04-14
fix(#47): catch gh api .../merge bypass in merge-gate hooks

atlas-apex:
## Summary

Closes the silent merge-gate bypass surfaced in [#47](https://github.com/me2resh/apexstack/issues/47). The three merge-gate hooks were only wired on `Bash(gh pr merge *)`, so merges performed via the raw REST-API shape `gh api repos/<owner>/<repo>/pulls/<N>/merge -X PUT` skipped all three gates. Incident: `me2resh/curios-dog#190` was merged via `gh api` while CI was still running.

- **Shared helper** — `.claude/hooks/_lib-extract-pr.sh` exposes `is_merge_command` + `extract_pr_number`, recognising both `gh pr merge <N>` and `gh api .../pulls/<N>/merge`. Sourced by all three merge-gate hooks so the parser lives in exactly one place.
- **Hooks refactored** — `block-unreviewed-merge.sh`, `require-design-review-for-ui.sh`, and `block-merge-on-red-ci.sh` now call the helper instead of inline-grepping. Each also recovers `owner/repo` from the `gh api` URL path when `--repo` is absent, so downstream `gh pr view` / `gh pr diff` / `gh pr checks` remain scoped correctly.
- **Settings wiring** — `.claude/settings.json` gets sibling `Bash(gh api *)` matcher entries for each of the three hooks. Option A (sibling entries), because the `if` field is a glob pattern, not a regex, so alternation isn't supported. The hooks themselves filter non-merge `gh api` calls via `is_merge_command`, which specifically looks for `/pulls/<N>/merge` in the URL path — any other `gh api` call exits 0 silently.
- **Docs** — `.claude/rules/pr-workflow.md § "Both merge shapes are gated (#47)"` states the rule explicitly and calls out that using `gh api .../merge` to route around a gate is itself a rule violation on par with forging an approval marker. `.claude/hooks/README.md` notes the dual registration.

## Test plan

- [x] `bash .claude/hooks/block-unreviewed-merge.sh` stdin = classic shape → BLOCKED, PR=99999
- [x] `bash .claude/hooks/block-unreviewed-merge.sh` stdin = API shape → BLOCKED, PR=99999
- [x] `bash .claude/hooks/block-merge-on-red-ci.sh` stdin = API shape → fires `gh pr checks` against the correct repo
- [x] `bash .claude/hooks/require-design-review-for-ui.sh` stdin = API shape → extracts PR, proceeds through diff check
- [x] Non-merge `gh pr view 123` passes through silently (exit 0)
- [x] Non-merge `gh api repos/foo/bar/issues/42` passes through silently (helper `is_merge_command` requires `/pulls/<N>/merge`)
- [x] JSON still valid after settings.json edit (`jq . .claude/settings.json`)
- [ ] End-to-end verification that a real `gh api` merge attempt is blocked in a live Claude Code session (requires a PR to merge against — ideally this PR itself, after approval)

## Glossary

| Term | Definition |
|------|------------|
| Merge-gate hook | A PreToolUse hook that runs before a merge command and exits 2 to block the merge when preconditions aren't met (Rex approval, CEO approval, green CI, design review for UI). |
| `gh pr merge` shape | The high-level `gh` subcommand form of merging a PR, e.g. `gh pr merge 123 --squash`. What Claude is supposed to use. |
| `gh api` shape | The raw REST-API form, e.g. `gh api repos/owner/repo/pulls/123/merge -X PUT`. Equivalent effect, different matcher — which is exactly what let it bypass the gates. |
| Matcher (`if` field) | The glob pattern in `.claude/settings.json` that decides whether a hook fires on a given tool call. Globs don't support alternation, so two shapes need two matcher entries. |
| Option A vs Option B | A — sibling matcher entries per hook (chosen, because globs only). B — one catch-all regex matcher (rejected, unsupported). |
| `_lib-` prefix | Convention for a sourced helper script in `.claude/hooks/` that is NOT a hook itself. The harness ignores it; other hook scripts `source` it. |
| Silent bypass | A merge that appears to succeed and produces no hook output because no matcher pattern matched the command. The worst kind of failure — there's no audit trail. |

Ticket: https://github.com/me2resh/apexstack/issues/47

### DIFF
--- a/.claude/hooks/README.md
+++ b/.claude/hooks/README.md
@@ -225,6 +225,8 @@ The new hooks are registered in `.claude/settings.json` alongside the existing o
 2. **`require-design-review-for-ui.sh`**
 3. **`block-merge-on-red-ci.sh`**
 
+All three merge-gate hooks are **also** registered on `Bash(gh api *)` so the REST-API merge shape (`gh api repos/<owner>/<repo>/pulls/<N>/merge -X PUT`) can't silently bypass them. The merge-shape detection and PR-number extraction live in the shared sourced helper `_lib-extract-pr.sh` — any future change to PR-number parsing should be made there, not in the individual hooks. Context: [#47](https://github.com/me2resh/apexstack/issues/47).
+
 The ordering is deliberate: cheap local checks first, expensive remote checks (`gh pr checks`) last, so that a merge already blocked by Rex/CEO/design markers doesn't pay the network cost.
 
 ## Pre-existing Hooks
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+# Shared PR-number extraction for the three merge-gate hooks:
+#   - block-unreviewed-merge.sh
+#   - require-design-review-for-ui.sh
+#   - block-merge-on-red-ci.sh
+#
+# Not a hook itself (prefixed with `_lib-` so it's never wired as one). Sourced
+# by the hooks above via `. "$(dirname "$0")/_lib-extract-pr.sh"`.
+#
+# WHY THIS EXISTS
+# ---------------
+# The merge gates originally only matched `gh pr merge <N>`. Incident (#47):
+# merges via `gh api repos/<owner>/<repo>/pulls/<N>/merge -X PUT` silently
+# bypassed all three gates because neither the matcher nor the PR-number
+# extraction knew about the API shape. This helper gives every gate a single,
+# tested way to recognise both shapes:
+#
+#   1. `gh pr merge 42 --squash`                                  → PR is 42
+#   2. `gh api repos/owner/repo/pulls/42/merge -X PUT`            → PR is 42
+#
+# Any tool that edits one of the three merge hooks MUST keep calling this
+# helper, not re-implement the parsing inline. That's the whole point.
+#
+# USAGE
+# -----
+#   . "$(dirname "$0")/_lib-extract-pr.sh"
+#   if ! is_merge_command "$COMMAND"; then exit 0; fi
+#   PR_NUMBER=$(extract_pr_number "$COMMAND")
+
+# Returns 0 if $1 looks like a merge command this gate should fire on.
+# Matches EITHER:
+#   - `gh pr merge ...`
+#   - `gh api ... repos/<owner>/<repo>/pulls/<N>/merge ...`
+is_merge_command() {
+  local cmd="$1"
+  if echo "$cmd" | grep -qE '\bgh\s+pr\s+merge\b'; then
+    return 0
+  fi
+  # `gh api` with a `/pulls/<N>/merge` path anywhere in the command. The path
+  # may be quoted, slash-separated, and may include query params.
+  if echo "$cmd" | grep -qE '\bgh\s+api\b.*repos/[^/[:space:]]+/[^/[:space:]]+/pulls/[0-9]+/merge\b'; then
+    return 0
+  fi
+  return 1
+}
+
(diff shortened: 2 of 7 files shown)


atlas-apex (commented):
## Code Review: PR #54

**Commit**: `a44b3c5e27f3c2c3c7d30728e4badeffa49e5058`

### Summary
Closes the `gh api .../pulls/<N>/merge -X PUT` silent-bypass gap on all three merge-gate hooks (#47). Introduces `_lib-extract-pr.sh` as the single source of truth for PR-number parsing, adds sibling `Bash(gh api *)` matcher entries in `settings.json`, and updates docs.

### Checklist Results
- Architecture & Design: Pass — single parser in `_lib-extract-pr.sh`, three hooks source it, no duplication.
- Code Quality: Pass — well-commented, defensive extraction (handles `|`, `&&`, `;` boundaries), URL-path fallback for `--repo` recovery.
- Testing: Pass (manual) — author's test plan covers both shapes + non-merge passthrough. Reproduced locally: classic shape blocks PR #100, `gh api` shape blocks the same PR #100, `gh api .../issues/42` exits 0, `gh pr view 123` exits 0.
- Security: Pass — no cross-repo marker lookup; markers correctly scoped to local repo root via `onboarding.yaml` walk.
- Performance: Pass — cheap greps, no new network calls in helper.
- PR Description & Glossary: Pass — Glossary includes 7 terms covering `gh api` shape, matcher semantics, `_lib-` convention, silent bypass, Option A/B.
- Technical Decisions (AgDR): N/A — the Option A vs Option B tradeoff is narrow (globs don't support alternation, so sibling entries are forced) and documented in PR body + pr-workflow.md. Not material enough for an AgDR.

### Findings
1. **Minor — false-positive on `gh api .../pulls/<N>/merge` GET**. `is_merge_command` matches the URL path without checking for `-X PUT`. A GET to inspect merge status (rare but legal) would also trip the gate. Not blocking — worst case is "use a different command to inspect." Safer than the opposite failure.
2. **Verified no self-approval**: `.claude/session/reviews/54-*.approved` does not exist.
3. **Cross-repo resolution confirmed correct**: markers read from local apexstack root regardless of the target repo in the API URL — exactly right for session-state semantics.

### Follow-up (not blocking)
- File: tighten `is_merge_command` to require `-X PUT` or `--method PUT` alongside the `/merge` path, to eliminate the GET false-positive.

### Verdict
**APPROVED**

---
Reviewed commit: \`a44b3c5e27f3c2c3c7d30728e4badeffa49e5058\`

me2resh (approved):

atlas-apex (commented):
## Re-review: PR #54

**Commit**: `1238688be77c7654ddb38aef530e90a7b0a6a44a`

### Verification
- Incremental diff is exactly 4 deletions, 0 additions in `block-unreviewed-merge.sh` (the dead `REPO_FLAG=""` / `if [ -n "$CMD_REPO" ]` / `REPO_FLAG="--repo $CMD_REPO"` / `fi` block). Nothing else snuck in.
- Grepped the file for `REPO_FLAG` — zero remaining references. `CMD_REPO` is still computed for parity with the sibling hooks but the unused `REPO_FLAG` is gone. No downstream logic broken: `extract_pr_number` from `_lib-extract-pr.sh` handles PR-number extraction without needing the flag.
- CI on new HEAD: **all green** — `shellcheck .claude/hooks` passes, along with `Verify Ticket ID`, `lychee`, `markdownlint-cli2`.

### Verdict
**APPROVED** (posted as comment — GitHub blocks self-approval)

---
🤖 Re-reviewed by Rex (Code Reviewer Agent)
📌 Reviewed commit: `1238688be77c7654ddb38aef530e90a7b0a6a44a`

PR #65, opened 2026-04-17
fix(#56): exempt absolute-form .claude/ / docs/ paths in require-active-ticket.sh

### LINKED ISSUE #56: [Bug] require-active-ticket.sh fails when REPO_ROOT is a nested worktree — strips wrong prefix from FILE_PATH
## Problem

\`.claude/hooks/require-active-ticket.sh\` computes \`REL_PATH\` by stripping \`REPO_ROOT\` from the absolute \`FILE_PATH\`. When the calling CWD is inside a **nested** git repository (e.g. an agent worktree created by the Agent tool at \`workspace/<project>/.claude/worktrees/agent-xxx/\`), \`git rev-parse --show-toplevel\` returns the worktree path — NOT the outer apexstack tree.

If an agent in that nested worktree tries to Write to a file in the outer apexstack tree (e.g. \`/Users/ahmed/Projects/apexstack/.claude/session/current-ticket\`), the strip no-ops (prefixes don't match), \`REL_PATH\` stays absolute, and the \`.claude/*\` case-match on line 36 never fires.

## Incident

Agent dispatched from curios-dog CWD with worktree isolation. CWD ended up as \`workspace/curios-dog/.claude/worktrees/agent-a6986b62/\`. Agent tried to update \`.claude/session/current-ticket\` in the outer apexstack tree for apexstack#55 work. Hook computed absolute \`REL_PATH\` and rejected the write. Only \`*.md\` Writes slipped through because the shell glob \`*\` crosses \`/\`. Full session log on https://github.com/me2resh/apexstack/issues/55.

## Proposed fix

One-line change to the case statement in \`.claude/hooks/require-active-ticket.sh\`:

\`\`\`diff
- case "\$REL_PATH" in
-   .claude/*|.claude|*.md)  exit 0 ;;
+ case "\$REL_PATH" in
+   .claude/*|.claude|*.md|*/.claude/*|*/.claude)  exit 0 ;;
\`\`\`

This makes the \`.claude/\` exemption work whether \`REL_PATH\` is relative or absolute. \`*.md\` already crosses \`/\`, so the absolute-path case was a known-good pattern — just extend it.

## Acceptance Criteria

- [ ] Case match covers both relative (\`.claude/\`) and absolute (\`/path/to/.claude/\`) forms
- [ ] Existing tests (if any) still pass
- [ ] Smoke test: pipe a mock tool_input with \`FILE_PATH\` set to both a relative \`.claude/session/current-ticket\` and an absolute \`/some/tree/.claude/session/current-ticket\` — both exit 0

## Priority

**P1** — unblocks agent-worktree workflows for apexstack-repo work. Without this, any agent spawned from one project's CWD cannot write session markers when its real work is in a different repo (as happened repeatedly during the 2026-04-14 session).

## Related

- apexstack#55 (the task that exposed this bug)
- The handful of agent reports on 2026-04-14 that flagged "sandbox blocked session marker write"

atlas-apex:
## Summary

One-hook fix for apexstack#56. When an agent worktree's `git-toplevel` differs from the outer apexstack tree, `require-active-ticket.sh` leaves `REL_PATH` absolute and the case-statement exemptions (relative-only) miss. Legitimate writes to `.claude/session/current-ticket` and `docs/**` get blocked.

Fix: extend each path-prefix exemption with its absolute-form twin (`*/.claude/*|*/.claude`, `*/docs/*|*/docs`, `*/projects/*/docs/*`). The existing `*.md` pattern already crosses `/`, so this is the same trick applied to the remaining exemptions.

## What changed

- `.claude/hooks/require-active-ticket.sh` — three case arms extended + a comment block explaining the pattern for future maintainers
- Added a 7-case smoke test (`/tmp/smoke-56.sh` — not committed, stored locally); all pass

## Testing

```
PASS  relative .claude/
PASS  absolute .claude/
PASS  nested-worktree absolute .claude/
PASS  relative docs/
PASS  absolute docs/
PASS  non-exempt code, no ticket       (blocks as before)
PASS  non-exempt code, with ticket     (allows as before)
```

## Scope note

The issue literally asked only about `.claude/`. I extended `docs/` and `projects/*/docs/*` too — they would fail the exact same way in the exact same conditions. Filing three follow-up bugs would be busywork. Documented the pattern in the comment block so future maintainers see the reasoning.

## Glossary

| Term | Definition |
|------|------------|
| REPO_ROOT | Output of `git rev-parse --show-toplevel`. Used to normalise FILE_PATH to a repo-relative form. |
| absolute-form fallthrough | When FILE_PATH doesn't start with REPO_ROOT (worktree mismatch), the strip on lines 29-31 is a no-op and REL_PATH stays absolute |
| agent worktree | Claude Code creates isolated git worktrees for parallel agents; their git-toplevel is the worktree path, not the outer tree |
| hook | Shell script invoked by Claude Code on PreToolUse; exit 2 blocks the tool call |

## Related

- Closes #56
- Context: apexstack#55 exposed this bug indirectly — session logs cited "sandbox blocked session marker write" from agent worktrees

### DIFF
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -44,9 +44,11 @@ fi
 case "$REL_PATH" in
   .claude/*|.claude|*/.claude/*|*/.claude) exit 0 ;;
   docs/*|docs|*/docs/*|*/docs) exit 0 ;;
-  projects/*/docs/*|*/projects/*/docs/*) exit 0 ;;
   TODO.md|README.md|MEMORY.md|CLAUDE.md) exit 0 ;;
 esac
+# Note: `projects/*/docs/*` is subsumed by `*/docs/*` above (shell case `*`
+# crosses `/`), so no separate arm needed. Per-project apexstack docs are
+# matched by the generic docs-in-any-subtree pattern.
 case "$REL_PATH" in
   *.md) exit 0 ;;
 esac
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -31,11 +31,20 @@ if [ -n "$REPO_ROOT" ]; then
   esac
 fi
 
-# Exempt paths
+# Exempt paths.
+#
+# Each path-prefix exemption is matched in both REL_PATH (repo-relative)
+# and absolute (*/path/*) forms. Absolute-path fallthrough happens when
+# FILE_PATH points outside REPO_ROOT (e.g. agent worktrees whose
+# git-toplevel differs from the outer apexstack tree); in that case the
+# strip on lines 29-31 is a no-op and REL_PATH stays absolute. The
+# existing `*.md` pattern already crosses `/`, so absolute-match via a
+# `*/…` prefix is a known-good shape — #56 extends the same trick to the
+# path-prefix exemptions.
 case "$REL_PATH" in
-  .claude/*|.claude) exit 0 ;;
-  docs/*|docs) exit 0 ;;
-  projects/*/docs/*) exit 0 ;;
+  .claude/*|.claude|*/.claude/*|*/.claude) exit 0 ;;
+  docs/*|docs|*/docs/*|*/docs) exit 0 ;;
+  projects/*/docs/*|*/projects/*/docs/*) exit 0 ;;
   TODO.md|README.md|MEMORY.md|CLAUDE.md) exit 0 ;;
 esac
 case "$REL_PATH" in


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #65

**Commit**: `dbb09090cd76620c47d6059f639f86c598094510`

### Summary
One-hook fix for #56. Extends the `REL_PATH` case statement in `.claude/hooks/require-active-ticket.sh` to match absolute-form paths (`*/.claude/*`, `*/docs/*`, `*/projects/*/docs/*`) in addition to the existing repo-relative forms. Fixes the agent-worktree fallthrough where `REPO_ROOT` strip is a no-op and `REL_PATH` stays absolute.

### Checklist Results
- Architecture & Design: Pass — same shape as the existing `*.md` glob; consistent with how the hook already handles its only other cross-`/` pattern
- Code Quality: Pass — 9-line comment block documents the reasoning; no dead code
- Testing: Pass — 7-case transcript in PR body covers relative, absolute, nested-worktree, non-exempt-with/without-ticket
- Security: Pass — exemption only broadens the "no ticket required" path; cannot bypass any other hook
- Performance: N/A (case-statement eval)
- PR Description & Glossary: Pass — 4 terms, all load-bearing (REPO_ROOT, absolute-form fallthrough, agent worktree, hook)
- Technical Decisions (AgDR): N/A — applying an existing pattern, no new approach

### Logic trace against the 7 smoke cases
- `REPO_ROOT=/x`, `FILE_PATH=/x/.claude/session/current-ticket` → `REL_PATH=.claude/session/current-ticket` → matches `.claude/*` → exit 0 ✓
- `REPO_ROOT=""` + `FILE_PATH=/x/.claude/session/...` → strip is no-op → `REL_PATH=/x/.claude/session/...` → matches `*/.claude/*` → exit 0 ✓
- `REPO_ROOT=/worktree`, `FILE_PATH=/outer/.claude/session/...` → strip mismatches → `REL_PATH` stays absolute → matches `*/.claude/*` → exit 0 ✓
- `FILE_PATH=/x/docs/foo.md` → `REL_PATH=docs/foo.md` → matches `docs/*` (or `*.md` fallthrough) → exit 0 ✓
- Absolute `/y/docs/foo.md` with mismatched root → matches `*/docs/*` → exit 0 ✓
- Non-exempt code, no ticket → falls through to marker check → exit 2 ✓
- Non-exempt code, with ticket marker → exit 0 ✓

All 7 pass.

### Scope
Extending `docs/` and `projects/*/docs/*` alongside `.claude/` is justified — same root cause, same file, same patch hunk. Filing two follow-ups would be busywork and would leave the hook half-fixed between merges.

### Potential false-positive
`*/docs/*` will match a legitimate source file at `/x/src/myapp/docs/config.ts` in a downstream project and exempt it from the ticket-first rule. In practice: apexstack hooks operate on the framework's own exemption list, downstream projects that vendor the hook tend to keep apexstack's layout (`docs/` is docs, not source), and the prior `docs/*` pattern already had this property at the repo-relative level — only difference is it now crosses `/`. Not a blocker. Worth a follow-up only if a downstream project reports ghost-exempt source files.

### Suggestions
None blocking.

### Verdict
**APPROVED**

---
Reviewed commit: `dbb09090cd76620c47d6059f639f86c598094510`

atlas-apex (commented):
Re-review at HEAD `e52928207c1ffe1b6a33c4721dae488e9c9c0a18`: incremental diff is purely the shellcheck lint fix — drops the dead `projects/*/docs/*` arm (subsumed by `*/docs/*`) and adds an inline comment explaining the subsumption for future maintainers. Remaining case arms still cover the intent: per-project apexstack docs exempted via `*/docs/*`. CI green: shellcheck .claude/hooks passes, Verify Ticket ID passes. SC2221/SC2222 warnings resolved. LGTM.

PR #91, opened 2026-04-19
chore(#90): publish site to yard.apexscript.com

atlas-apex:
## Summary

Prep for publishing the landing page at yard.apexscript.com on Netlify:

- `netlify.toml` at repo root — `publish = "site"` + security headers (X-Frame-Options, X-Content-Type-Options, Referrer-Policy, Permissions-Policy).
- `site/index.html` head — canonical URL, Open Graph, Twitter card meta so link previews render.
- Google Analytics (G-G3EMR3CB02) via Consent Mode v2 — `analytics_storage` defaults to denied; a minimal fixed-bottom accept/decline banner records the visitor's choice in localStorage and calls `gtag('consent', 'update', ...)` on accept. IP anonymization on. No ad_storage / ad_user_data / ad_personalization ever requested.
- Mobile fixes — `.steps` and `.demos` grids switched from `1fr` to `minmax(0, 1fr)` and `min-width: 0` on `.step` / `.demo`, so long `<pre>` content (white-space: pre) no longer inflates grid tracks past the viewport. `white-space: nowrap` on titlebar nav links so "what's in the box" stops breaking mid-phrase.
- README — dropped a single historical reference line.

## Testing

1. Open `site/index.html` locally — banner appears on first load, hides after Accept or Decline.
2. DevTools → Application → Local Storage → confirm `ay-consent` = `granted` or `denied`.
3. DevTools → Network → before Accept, no hits to `google-analytics.com`. After Accept, `collect?...` request fires.
4. Resize to 320px width — code blocks scroll horizontally within their container instead of spilling past the viewport; titlebar nav "what's in the box" stays on one visual phrase.
5. After merge + Netlify re-deploy + CNAME flip, verify:
   - `curl -I https://yard.apexscript.com/` shows security headers
   - `<link rel="canonical">` resolves to the same URL
   - `https://metatags.io/?url=https://yard.apexscript.com/` shows OG/Twitter preview

Closes #90

---

## Glossary

| Term | Definition |
|------|------------|
| Consent Mode v2 | Google's framework for loading gtag.js but deferring actual cookie writes / hit sends until a consent signal is granted. Replaces the older "don't load gtag until consent" pattern with a finer-grained denied/granted toggle per data category. |
| `analytics_storage` | One of five Consent Mode v2 categories; controls whether gtag may set the `_ga`/`_ga_*` cookies and send page_view hits. The only category this site ever requests. |
[...]

PR #93, opened 2026-04-19
chore(#92): site polish - AgDR, banner a11y, CSP, canonical redirect

### LINKED ISSUE #92: [Task] Site polish: CSP, banner a11y, canonical redirect, AgDR
Follow-up to #90 / PR #91 — Rex flagged five non-blocking suggestions during review. None were merge blockers, but each is worth a targeted chore.

### 1. AgDR for GA + self-rolled consent banner

No AgDR exists for the "Google Analytics + roll our own banner" choice. ApexYard's own rules (`@.claude/rules/agdr-decisions.md`) say `/decide` for any library/tool choice or design pattern. Deferred because the initial add was small + experimental, but the repo should practise what it preaches.

**AC**: `docs/agdr/AgDR-NNNN-analytics-and-consent.md` exists and captures the options (GA + self-rolled vs GA + CMP vs Plausible/Umami vs no analytics), the choice, and consequences.

### 2. Banner `role="region"` instead of `role="dialog"`

The consent banner is non-modal (doesn't trap focus, doesn't block the page). `role="dialog"` is technically wrong for non-modal UI. `role="region"` with `aria-label` fits better.

**AC**: `site/index.html` banner markup updated; screen-reader announcement still makes sense.

### 3. Escape-key dismissal for keyboard users

Currently the banner can only be dismissed by clicking Accept/Decline. Keyboard users should be able to dismiss with Escape (which should map to Decline — the non-consenting default).

**AC**: Pressing Escape while the banner is visible records `ay-consent=denied` and hides the banner. Focus restored to wherever it was before the banner appeared.

### 4. CSP header in `netlify.toml`

Current security headers cover X-Frame-Options, X-Content-Type-Options, Referrer-Policy, Permissions-Policy. Missing: Content-Security-Policy. A reasonable CSP for a single-file static site with GA + Google Fonts:

```
default-src 'self';
script-src 'self' 'unsafe-inline' https://www.googletagmanager.com;
style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;
font-src 'self' https://fonts.gstatic.com;
img-src 'self' data: https://www.google-analytics.com;
connect-src 'self' https://www.google-analytics.com https://analytics.google.com;
frame-ancestors 'self';
```

**AC**: `netlify.toml` has the CSP, site still loads cleanly (no console CSP violations), GA still records page_views.

### 5. Netlify redirect: `apexyard.netlify.app` → `yard.apexscript.com`

Once DNS flips and the custom domain is live, visits to the `.netlify.app` preview domain should 301-redirect to the canonical URL. Consolidates SEO signals and matches the `<link rel="canonical">`.

**AC**: `netlify.toml` has a `[[redirects]]` entry that 301s any request to `apexyard.netlify.app` → `https://yard.apexscript.com/:splat`. Tested after DNS is live.

## Out of scope

- Actual DNS flip — that's separate, waiting on the ops team.
- Any deeper CMP integration (Cookiebot, Osano, etc.) — deferred until there's a reason.

## Context

Rex review: https://github.com/me2resh/apexyard/pull/91
Merged PR: #91 (6cbc95b)
Closes #90 already — this is a fresh scope.

atlas-apex:
## Summary

Follow-up to PR #91 addressing all five of Rex's non-blocking suggestions as one batch since they're all small and land on the same surface.

1. **AgDR-0004 backfill** for the GA + self-rolled-banner choice (was flagged as missing during #91 review). Captures the four options considered, the decision, and the trigger-points that should prompt revisiting (adding a second pixel, monetisation, GA4 retirement, UK/EU enforcement tightening).
2. **Banner `role=\"dialog\"` → `role=\"region\"`** — the banner is non-modal so `role=\"dialog\"` was misleading assistive tech.
3. **Escape-key dismissal** records `ay-consent=denied` and hides the banner, so pure-keyboard users aren't stuck with it. Small refactor extracted a shared `record(choice)` helper.
4. **CSP header** in `netlify.toml` — explicit allow-list for `googletagmanager.com`, `google-analytics.com`, `analytics.google.com`, `fonts.googleapis.com`, `fonts.gstatic.com`, plus `'unsafe-inline'` on script-src/style-src because the single-file landing has inline styles and inline gtag/animation/banner JS. Includes defensive `frame-ancestors`, `base-uri`, `form-action` = `'self'`.
5. **Netlify 301 redirect** from `https://apexyard.netlify.app/*` → `https://yard.apexscript.com/:splat`. Dormant until DNS flips; once the custom domain is live, visits to the `.netlify.app` host canonicalize so search engines consolidate SEO signals.

## Testing

1. **Banner a11y**: open the deploy preview, Tab to the banner, press Escape — banner hides, DevTools → Application → Local Storage shows `ay-consent=denied`. Reload — banner stays hidden.
2. **CSP**: open the deploy preview with DevTools → Console. No CSP violations expected. After accepting the banner, network tab should show a successful request to `www.google-analytics.com/g/collect?...`. If either is broken, the CSP is too tight — loosen before merge.
3. **Canonical redirect**: tested after DNS flips. Once `yard.apexscript.com` resolves, `curl -sI https://apexyard.netlify.app/` should return `HTTP/2 301` + `location: https://yard.apexscript.com/`.
4. **AgDR**: no runtime effect — docs only.

Closes #92

---

## Glossary

| Term | Definition |
|------|------------|
| AgDR | Agent Decision Record — ApexYard's project-scoped equivalent of an ADR. One per significant technical decision; captures options considered, the choice, and consequences. Lives at `docs/agdr/AgDR-NNNN-slug.md`. |
| `role=\"region\"` | ARIA landmark role for a "significant section with a label". Correct for a non-modal labelled banner. Distinct from `role=\"dialog\"`, which tells assistive tech to expect modal semantics (focus trap, page blocked). |
| Content-Security-Policy (CSP) | HTTP response header that tells the browser which origins are allowed for each resource type (scripts, styles, fonts, images, XHR destinations, etc.). Violations are console-logged and the offending resource is blocked. |
| `'unsafe-inline'` | CSP token allowing inline `<script>`/`<style>` + inline event handlers. Required here because the single-file landing has inline everything; the alternative is moving all styles/scripts to external files or using nonces, which is heavier than the single-page format warrants. |
| `[[redirects]]` | Netlify `netlify.toml` array-of-tables for URL redirects. Each entry has `from` / `to` / `status` / optional `force`. `force = true` means redirect even if the file exists at the `from` path. |
| Canonical host | The "intended" hostname for a page when it's reachable via multiple URLs (`.netlify.app` preview vs production custom domain). A 301 to the canonical URL consolidates SEO signals and matches `<link rel=\"canonical\">`. |

### DIFF
--- a/docs/agdr/AgDR-0004-analytics-and-consent.md
+++ b/docs/agdr/AgDR-0004-analytics-and-consent.md
@@ -1,0 +1,72 @@
+---
+id: AgDR-0004
+timestamp: 2026-04-19T08:30:00Z
+agent: atlas
+model: claude-opus-4-7
+trigger: retrospective
+status: executed
+ticket: me2resh/apexyard#92
+---
+
+# Google Analytics + self-rolled consent banner for the landing page
+
+> In the context of needing traffic measurement on the ApexYard landing page now that it publishes at `yard.apexscript.com`, facing the choice between Google Analytics with a self-rolled consent banner, Google Analytics with a third-party CMP, a privacy-first analytics service (Plausible / Umami), or no analytics at all, I decided to ship **GA4 with Consent Mode v2 and a minimal self-rolled accept/decline banner**, accepting that the banner doesn't offer per-category granularity and that we'll owe a proper CMP the moment the site's analytics needs grow past "count visits".
+
+## Context
+
+The ApexYard landing page is a single static HTML file published to `yard.apexscript.com` via Netlify. The CEO wanted a quick way to see how much traffic the site gets — nothing sophisticated, just "is this actually getting viewed". The repo is open-source, experimental, and not monetised; there is no marketing team, no growth funnel, no retargeting.
+
+The site is served to visitors in the EU and UK, which means GDPR and ePrivacy apply. Analytics cookies / local storage / identifiers require explicit consent BEFORE they are set. Any analytics choice has to account for this.
+
+Two constraints shaped the decision:
+
+1. **The CEO already provided the GA4 tag** (`G-G3EMR3CB02`) — there was an explicit preference for Google Analytics, not an open "which tool" question.
+2. **The decision was made inline during PR #91** and shipped without an AgDR — this record is a retrospective backfill. Rex's post-review flagged the missing AgDR as a non-blocking suggestion, logged as part of #92.
+
+## Options Considered
+
+| Option | Pros | Cons |
+|--------|------|------|
+| **GA4 + self-rolled banner (chosen)** | Minimal code (~35 lines CSS, ~20 lines JS); matches the site's brutalist monochrome aesthetic; Consent Mode v2 means gtag loads but sends nothing until explicit grant; IP anonymization on; `ay-consent` key in localStorage persists choice. | No per-category controls (only a single analytics_storage toggle); no Do-Not-Track integration; no preference-centre UI; the banner JS and markup are maintenance surface we own. |
+| **GA4 + third-party CMP** (Cookiebot, OneTrust, Osano, Iubenda, etc.) | Legal-grade compliance; per-category granularity; audit logs; cookie auto-scan; preference centre out of the box. | Adds a vendor dependency, per-month cost (free tiers usually cap at 50k pageviews and feel spammy); extra script weight on a ~100KB static page; visually inconsistent with the brutalist design without heavy theming; overkill for "count visits on an OSS landing page". |
+| **Plausible / Umami / Fathom** (cookieless analytics) | No banner required at all (no cookies, no personal identifiers); lighter script; privacy-first narrative fits an OSS tool. | CEO's existing asset was a GA4 tag, not a Plausible account; self-hosting Umami means running a DB we don't have; Plausible SaaS is ~$9/mo (small but real); GA4's deeper ecosystem (Search Console link, BigQuery export) is unused today but available if needs grow. |
+| **No analytics** | Zero compliance surface. Simplest possible answer. | CEO explicitly asked for traffic measurement. |
+
+## Decision
+
+Chosen: **GA4 with Consent Mode v2 + self-rolled banner**, because the site's current analytics need is genuinely "count visits, nothing else", the CEO has a GA tag already, and a third-party CMP would be heavier than the analytics it gates. Consent Mode v2 specifically (rather than the older "don't load gtag until consent" pattern) because it keeps the code in one place — gtag always loads with `default: denied`, and the banner only flips a single `consent: update` call.
+
+Key configuration choices:
+
+- Default state for all four Consent Mode v2 categories (`ad_storage`, `ad_user_data`, `ad_personalization`, `analytics_storage`) = `denied`. Only `analytics_storage` is ever requested via `update`.
+- `anonymize_ip: true` on `gtag('config', ...)`.
+- Choice persisted in `localStorage['ay-consent']` as `granted` or `denied`; banner auto-hides on subsequent visits until cleared.
+- Banner uses `role="dialog"` + `aria-label` — **flagged for correction in #92** as `role="region"` is the better a11y fit for a non-modal banner (planned, separate commit on the same branch).
+
+## Consequences
+
+Positive:
+
+- Traffic is now measurable in GA4, gated by consent, starting the day DNS flips to `yard.apexscript.com`.
+- No recurring vendor cost, no extra script dependency, ~55 lines of code total.
+- The "we practise what we preach" posture — the landing page for an SDLC framework respects GDPR out of the box.
+
+Negative (accepted):
+
+- If analytics needs grow (marketing attribution, A/B testing, conversion funnels), the self-rolled banner is not sufficient. A proper CMP becomes the right answer and a migration is ours to run.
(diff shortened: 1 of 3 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #93

**Commit**: `7d286ec73085be4a91009a70a265853c029d0650`

### Summary

Follow-up to #91 addressing all five of my non-blocking suggestions in one batch:
AgDR-0004 retrospective backfill for the GA + self-rolled-banner choice,
`role="dialog"` → `role="region"` on the non-modal banner, Escape-key dismiss
(records `ay-consent=denied`), CSP header with an explicit allow-list in
`netlify.toml`, and a dormant 301 from `apexyard.netlify.app/*` → `yard.apexscript.com/:splat`.

Three commits, 105 / 6 LoC across three files. All checks green including the
Netlify deploy preview, its Header rules, and its Redirect rules.

### Checklist Results

- Testing:                   Pass (manual test plan in PR body; no runtime changes for AgDR)
- Security:                  Pass (CSP added, allow-list minimal)
- Performance:               Pass (no bundle impact)
- PR Description & Glossary: Pass (6 terms, all substantive)
- Technical Decisions (AgDR):Pass (AgDR-0004 present and filled in, not a stub)

### Verification I ran

**CSP coverage — every external URL the page actually loads is allow-listed:**

| Resource in `site/index.html` | URL | CSP directive | Covered? |
|---|---|---|---|
| Google Fonts stylesheet | `fonts.googleapis.com/css2?...` | `style-src` | Yes |
| Font files | `fonts.gstatic.com` | `font-src` | Yes |
| gtag loader | `www.googletagmanager.com/gtag/js?id=...` | `script-src` | Yes |
| GA4 `/g/collect` beacon | `www.google-analytics.com` | `connect-src` + `img-src` | Yes (both) |
| Alt GA endpoint | `analytics.google.com` | `connect-src` | Yes |
| GitHub anchors, canonical / OG URLs | various | navigation / meta only — not CSP-gated | N/A |

No missing origins. The `data:` token in `img-src` is correct for the GA 1px
fallback pixel. `frame-ancestors`, `base-uri`, `form-action` all defensively
set to `'self'`.

**Escape-key listener attachment (line 1985):**

```js
document.addEventListener('keydown', function (e) {
  if (e.key === 'Escape' && !el.hidden) {
    record('denied');
  }
});
```

Attached to `document`, not the banner element. This is correct — a keyboard
user's focus is almost never inside the banner when they hit Escape, so a
banner-scoped listener would silently miss most intended use. Guard on
`!el.hidden` is a nice no-op short-circuit.

**`record()` double-invocation guard (line 1970):**

```js
function record(choice) {
  if (el.hidden) return;
  // ... write localStorage, fire gtag consent update, hide banner
}
```

Once the banner is hidden, any further `record()` call short-circuits before
re-writing localStorage or re-firing `gtag('consent', 'update', ...)`.
Idempotent — good. Click handler and Escape handler can safely fire against
each other or themselves without doubled side effects.

**AgDR-0004 depth check:** 72 lines, four options with pros / cons table,
explicit decision rationale, split positive / negative consequences, and
four concrete trigger-points for revisiting the decision (second pixel,
monetisation, GA4 retirement, UK/EU enforcement). Not a stub — this is a
proper retrospective record.

**CI status:**

- Verify Ticket ID: pass
- lychee: pass
- markdownlint-cli2: pass
- Netlify deploy-preview: pass
- Netlify Header rules: pass
- Netlify Redirect rules: pass

Deploy preview is live at https://deploy-preview-93--apexyard.netlify.app —
Netlify validated the CSP and the redirect block against its own parsers,
which is the strongest signal short of a real DNS flip that the TOML is
well-formed.

### Non-blocking observations

1. **`www.googletagmanager.com` in `connect-src`** is slightly belt-and-braces
   — GA4 `/g/collect` goes to `www.google-analytics.com`, and GTM's own
   beacons in practice land on `www.googletagmanager.com/gtag/...` which is
   covered by `script-src` + `img-src`. Keeping it in `connect-src` is
   defensive and costs nothing; noting it so you know you can trim later if
   you ever audit the allow-list.

2. **`role="region"` + `aria-label` + `hidden`**: the ARIA Authoring Practices
   are mixed on whether `region` should be exposed when the element is
   `hidden` — some screen readers announce the landmark anyway on first
   paint before the JS runs. Not a bug (the `hidden` attribute already
   removes it from the a11y tree in every modern engine), just a note that
   `aria-hidden="true"` on initial render would be even more defensive. Low
   priority — I wouldn't re-open on this.

3. **Dormant redirect**: the 301 won't fire until DNS actually flips
   `yard.apexscript.com` onto this Netlify site. Netlify's "Redirect rules"
   check only validates the syntax, not the target. Worth a manual
   `curl -sI https://apexyard.netlify.app/` after the DNS switch to confirm
   it returns `HTTP/2 301` with the correct `location:` header. Already in
   the PR body's test plan.

### Verdict

**APPROVED**

All five #91 follow-ups land cleanly, CSP is minimal and correct, Escape
handling is properly scoped, AgDR is substantive. Ship it pending CEO nod.

---
Reviewed commit: \`7d286ec73085be4a91009a70a265853c029d0650\`

Issue #109, opened 2026-04-24
[Chore] Make ticket-prefix whitelist + schema project-configurable (not hardcoded)


atlas-apex:
## Driver

The `/feature` and `/task` skills today recognise only `[Feature]` / `[Bug]` / `[Chore]` / `[Refactor]` / `[Testing]` / `[CI]`. In practice, real projects hit prefixes outside that list — monorepo scaffolding, infra-as-code tickets, design-review work, pure docs PRs, security-specific tickets, performance tickets, and more. Every team's list is different.

Hardcoding a specific expanded set (`[Scaffold]`, `[Infra]`, `[Design]`, `[Docs]`) just shifts the drift — the next team needs `[Security]` or `[Perf]` or `[Accessibility]`, and we're back here adding more hardcoded prefixes.

## Scope

**Make the prefix whitelist and ticket schema project-configurable, not framework-hardcoded.**

1. Define a config location — `.claude/project-config.json` (extending the existing file if present) or `onboarding.yaml`:

   ```yaml
   ticket:
     prefix_whitelist:
       - Feature
       - Bug
       - Chore
       - Refactor
       - Testing
       - CI
       - Docs
     label_priority_scheme: "P0,P1,P2,P3"
   ```

2. Update `.claude/skills/task/SKILL.md` — title-prefix derivation reads the project's `prefix_whitelist` at runtime. Falls back to the shipped default if no config is present.
3. Update `.claude/skills/feature/SKILL.md` and `.claude/skills/bug/SKILL.md` similarly.
4. Update `.claude/skills/write-spec/SKILL.md` if it emits tickets.
5. Update `.claude/rules/git-conventions.md` branch-type table to reference the same source of truth (don't duplicate the list).
6. Update the `validate-branch-name.sh` and `validate-pr-create.sh` hook regexes to read the whitelist from config too — single source of truth across skills, hooks, CI.
7. Provide an example in `docs/` showing how a team extends the defaults (adds `[Security]` or `[Perf]`).

## Acceptance Criteria

- [ ] Config location is documented and loaded by skills at runtime.
- [ ] Shipped default matches the current six prefixes exactly — zero behavioural drift on a fresh fork with no overrides.
- [ ] A team can add a new prefix to their fork's config and it flows through: skill emits the new prefix, validators accept it, branch-name hook accepts the matching branch type.
[...]

Issue #50, opened 2026-04-13
feat: architecture diagram support in documentation — C4 model


atlas-apex:
## Context

ApexStack-managed projects should include architecture diagrams in their documentation. The CEO prefers the C4 model but is open to alternatives.

## C4 Model Overview

The C4 model (Context, Container, Component, Code) provides 4 levels of zoom:

| Level | What it shows | Audience |
|-------|--------------|----------|
| **L1 — Context** | System + external actors (users, third-party services) | Everyone |
| **L2 — Container** | High-level tech choices (frontend, backend, database, CDN) | Dev team |
| **L3 — Component** | Internal structure of a container (services, repositories, handlers) | Dev team |
| **L4 — Code** | Class/function level (rarely needed, auto-generated) | Deep-dive |

## Acceptance Criteria

- [ ] Decide on diagramming tool: Structurizr DSL, Mermaid C4, PlantUML C4, or D2
- [ ] Add diagram templates to \`templates/\` (L1 context + L2 container at minimum)
- [ ] \`/handover\` skill generates an L2 container diagram as part of the assessment
- [ ] Diagrams stored as code (text DSL) alongside rendered PNGs in \`docs/architecture/\`
- [ ] Per-project diagrams in \`projects/<name>/architecture/\` or \`workspace/<name>/docs/architecture/\`

## Options to evaluate via /decide

| Tool | Format | Rendering | Pros | Cons |
|------|--------|-----------|------|------|
| **Structurizr DSL** | Custom DSL | Structurizr Lite (Docker) or CLI | Purpose-built for C4, best model | Requires Java/Docker for rendering |
| **Mermaid C4** | Markdown-embeddable | GitHub renders natively | Zero setup, GitHub preview | Limited C4 support, less precise |
| **PlantUML C4** | Text DSL | plantuml.jar or server | Mature C4 library, widely used | Java dependency |
| **D2** | Text DSL | d2 CLI (Go binary) | Modern, beautiful output, no Java | Not C4-native (general purpose) |

## Recommendation

**Mermaid C4** for L1/L2 (GitHub renders it inline, zero setup) + **Structurizr DSL** for L3+ when precision matters. Start with Mermaid — it's good enough for context and container diagrams, and every GitHub user can see them without tooling.

## AgDR required

This is an architectural decision — diagramming tool choice affects the entire portfolio.

PR #119, opened 2026-04-24
chore(#110): block-private-refs-in-public-repos.sh hook

### LINKED ISSUE #110: [Chore] block-private-refs-in-public-repos.sh — hook to scrub registry names from upstream tickets
## Driver

Every apexyard user runs a portfolio of **private** projects registered in their fork's `apexyard.projects.yaml`, plus contributes upstream to **public** framework repos (`me2resh/apexyard`, others). Authored content that crosses that boundary — a ticket or PR filed upstream that references a private project's name, ticket numbers, timelines, or internal decisions — leaks information the user never intended to publish.

The leak vector isn't malicious; it's routine. An agent diagnoses a framework bug while working in a private project, then files the upstream ticket with a helpful "discovered during `<private-project>` rebuild" reference. Once filed, the private project name is on the public framework repo's issue tracker and searchable indefinitely.

This affects:

- The fork's owner — the repo names in their private registry become public context.
- Their clients / employers — if the user manages projects on behalf of others, those project names leak too.
- Every other apexyard user, for the same reasons.

Self-discipline doesn't prevent this. The information is *right there* while the agent is writing the upstream ticket — not referencing it takes active suppression. Mechanical enforcement is the right shape, same pattern as `check-secrets.sh` and the commit-format hooks.

## Scope

Add `.claude/hooks/block-private-refs-in-public-repos.sh`, wired to `PreToolUse` on `Bash` matching:

- `gh issue create --repo <repo> …`
- `gh pr create --repo <repo> …`
- `gh issue comment <n> --repo <repo> …` and `gh pr comment …`
- `gh api` calls hitting `/issues` or `/pulls` endpoints

The hook:

1. Extracts the target repo from the command (the `--repo` arg, or repo slug in a `gh api` path).
2. Determines if the target is **public / framework-class** — configurable list, defaults to `me2resh/apexyard` + whatever the fork's `upstream` remote resolves to.
3. If public-class, reads the fork's `apexyard.projects.yaml` to extract every registered project's `name`, `repo`, and `workspace` fields.
4. Scans the title + body (from `--title`, `--body`, `--body-file`, or `-F`) for any match against the extracted tokens, plus ticket-reference patterns (`<owner>/<repo>#<n>`) where `<owner>/<repo>` matches a registered project.
5. On match, exits 2 with a message that:
   - Names the specific leaked identifier (the matched token from the fork's own registry)
   - Suggests an abstract replacement (e.g. "describe as 'a registered project', not by name")
   - Explains the escape hatch
6. Skip marker: `<!-- private-refs: allow -->` in the body lets a deliberate reference through (rare — e.g. an apexyard ticket legitimately needs to reference a fork's registry entry by name).

### What counts as "public / framework-class"

Configurable. Default heuristic:

```yaml
# .claude/project-config.json or onboarding.yaml
leak_protection:
  public_framework_repos:
    - me2resh/apexyard
  # Also treat anything reachable via `git remote get-url upstream` as framework-class.
  auto_detect_upstream: true
  skip_marker: "<!-- private-refs: allow -->"
```

### What gets scrubbed

- Project names from `apexyard.projects.yaml` (`.projects[].name`)
- Project repo slugs (`.projects[].repo`)
- Workspace paths (`.projects[].workspace`)
- Any `<owner>/<repo>#<n>` pattern where `<owner>/<repo>` is a registered project

Does NOT scrub:

- The user's own `git config user.name` / email (that's signed on every commit anyway)
- Generic class descriptions ("a multi-package monorepo", "a private project")
- Timeline phrases ("during bulk filing", "on 2026-04-24") — dates are fine; it's the *attribution* that leaks

## Acceptance Criteria

- [ ] Hook exists, wired to `PreToolUse` on all four `gh` invocation shapes.
- [ ] Reads `apexyard.projects.yaml` — handles missing file gracefully (no registry → no scrub list → hook is a no-op).
- [ ] Blocks a ticket body containing a registered project name when the target is `me2resh/apexyard` or the configured upstream.
- [ ] Does NOT block when the target is a private registered repo (the user's own project — leaking isn't a concern inside the user's own org).
- [ ] Does NOT block when the skip marker is present.
- [ ] Block message names the exact leaked token and suggests an abstract replacement.
- [ ] Fixtures in `.claude/hooks/tests/` cover: name leak, repo-slug leak, ticket-ref leak, skip marker, missing registry, non-public target.
- [ ] `docs/rule-audit.md` lists the hook under its enforced rule.
- [ ] A new rule file `.claude/rules/leak-protection.md` documents the principle and the escape hatch.

## Risks / Dependencies

- False positives if a project's `name` collides with a generic word (e.g. a project named "auth" — would block any mention of auth in upstream tickets). Mitigation: warn on short / generic names when the registry is loaded, and match on whole-word boundaries only.
- False negatives if the user mentions the project by an alias not in the registry (shorthand, codename, internal nickname). Mitigation: hook is a *backstop*, not the only line of defence — pair with a rule doc and a feedback memory.
- Hook needs to parse `gh api` command strings, which is brittle. Mitigation: support `--method PUT/POST` + `-f` / `-F` field args; skip on unrecognised shapes with a warning logged to the session file.

## Glossary

| Term | Definition |
|------|------------|
| Public framework repo | A repo whose issues and PRs are publicly visible and indexed. `me2resh/apexyard` is the canonical one; each fork may add others. |
| Registered project | Any entry in the fork's `apexyard.projects.yaml` — by convention, these are the fork owner's private portfolio. |
| Skip marker | HTML comment in the body that tells the hook to allow this particular reference. |

atlas-apex:
## Summary

Adds a new PreToolUse hook `block-private-refs-in-public-repos.sh` that scrubs registered private-project identifiers from content destined for public framework repos (`me2resh/apexyard` + whatever the fork's `upstream` remote resolves to + any extras configured under `.leak_protection.public_framework_repos`).

Catches a common leak vector: an author diagnoses a framework bug while working in a private project, then files the upstream ticket or comment with a helpful "discovered during `<private-project>` rebase" reference. Once filed, the private project's name is permanently on a public issue tracker and searchable. This hook is the mechanical backstop for the discipline in the new `.claude/rules/leak-protection.md` rule.

- New hook: `.claude/hooks/block-private-refs-in-public-repos.sh` (~230 LOC)
- New rule doc: `.claude/rules/leak-protection.md`
- Config integration: reads `.leak_protection.public_framework_repos`, `.leak_protection.auto_detect_upstream`, `.leak_protection.skip_marker` via the shared config lib introduced in #109
- Defaults: extended `.claude/project-config.defaults.json` with a `leak_protection` subtree
- Wired into `.claude/settings.json` — five PreToolUse matchers for `gh issue create`, `gh pr create`, `gh issue comment`, `gh pr comment`, `gh api`
- Audit entry: new section 10 in `docs/rule-audit.md`, mechanized count bumped
- 10 test cases in `.claude/hooks/tests/test_block_private_refs.sh`

## How it works

1. On every matched `gh` invocation, parse `--repo` (or the `repos/owner/name/...` path on `gh api`). If the target isn't public-class (not in the configured list + upstream), silent no-op.
2. Locate `apexyard.projects.yaml` by walking up from CWD; extract each project's `name`, `repo`, and `workspace` fields. If the registry is missing, silent no-op — treated as "not an apexyard fork."
3. Self-exempt the target's own slug and bare name: mentioning `apexyard` on an apexyard upstream ticket doesn't fire.
4. Scan the title + body (from `--title`, `--body`, `--body-file`, `-F <path>`, and `gh api`-style `-F body=@file` / `-f body=...`) for whole-word matches against the extracted tokens, plus `owner/repo#N` patterns.
5. Skip marker `<!-- private-refs: allow -->` in the body bypasses with a visible warning.
6. On match: exit 2 naming each leaked token and suggesting abstract replacement phrasing.

## Testing

Ran the committed test fixture — 10 cases, all pass:

```
$ bash .claude/hooks/tests/test_block_private_refs.sh
PASS: name leak on gh issue create to me2resh/apexyard
PASS: repo-slug leak with ticket ref
PASS: bare repo-slug leak
PASS: skip marker bypasses leak check
PASS: missing registry → no-op
PASS: non-public target → no-op
PASS: empty body → no-op
PASS: workspace path leak
PASS: non-gh command → no-op
PASS: gh api issues leak
Passed: 10  Failed: 0
```

Also smoke-tested the config integration after swapping the agent's inlined defaults for shared-lib reads:

- Zero config present → hook falls back to shipped defaults (same behaviour as before)
- Config with custom `public_framework_repos` → hook honours the override
- Config with `auto_detect_upstream: false` → `upstream` remote ignored
- Tests still green end-to-end

- No CI workflow changes.
- Does not touch the registry or any skill.
- Does not alter commit-message or PR-title validation paths (handled in `validate-commit-format.sh` / `validate-pr-create.sh`).
- Does not parse the rare `--input -` / stdin-driven `gh api` shape (noted in the hook's own header comment as an accepted gap).

## Follow-ups

- Short-name / generic-word collisions (a project literally named `auth` or `core`) could produce false positives. Mitigation is documented in the rule file; can harden with a project-config-level denylist of "skip these names" if pain emerges.
- The `.claude/hooks/tests/` convention was introduced by this hook. If future hook tests accumulate, consider a tiny runner pattern doc.

## Glossary

| Term | Definition |
|------|------------|
| Public framework repo | A repo whose issues and PRs are publicly visible and indexed — `me2resh/apexyard` by default, plus each fork's configured list + upstream auto-detect. |
| Registered private project | Any entry in the fork's `apexyard.projects.yaml`. The hook treats these as "do not mention upstream." |
| Skip marker | Literal `<!-- private-refs: allow -->` in the body that lets one invocation through with a visible warning. |
| Self-exemption | The hook does NOT scrub references to the target repo itself — a ticket on the apexyard upstream can legitimately say "apexyard" without firing. |

Closes #110

### DIFF
--- a/.claude/hooks/block-private-refs-in-public-repos.sh
+++ b/.claude/hooks/block-private-refs-in-public-repos.sh
@@ -1,0 +1,386 @@
+#!/bin/bash
+# Blocks issue/PR/comment creation on a public framework repo when the title
+# or body references any registered private project from the fork's
+# apexyard.projects.yaml.
+#
+# The leak vector: an agent diagnoses a framework bug while working inside a
+# private project, then files the upstream ticket with "discovered during
+# <private-project> rebuild". Once filed, the project's name is indexed
+# forever on a public issue tracker.
+#
+# Fires on PreToolUse Bash for the five gh shapes that write to a remote
+# tracker:
+#   - gh issue create --repo <repo>
+#   - gh pr create --repo <repo>
+#   - gh issue comment <n> --repo <repo>
+#   - gh pr comment <n> --repo <repo>
+#   - gh api repos/<owner>/<repo>/{issues,pulls}[...]
+#
+# Behaviour:
+#   - Target repo not public-class → exit 0 silently.
+#   - apexyard.projects.yaml missing → exit 0 silently (no scrub list).
+#   - Body empty (no title + no body + no body-file) → exit 0 silently.
+#   - Skip marker `<!-- private-refs: allow -->` in body → exit 0 with a
+#     single-line warning to stderr.
+#   - Match against any registered project `name`, `repo`, `workspace`, or
+#     an `<owner>/<repo>#<N>` reference to a registered repo → exit 2 with a
+#     message naming each leaked token and suggesting abstract replacements.
+#
+# Configuration (future): a `.claude/project-config.json` may override
+# `leak_protection.public_framework_repos` and `leak_protection.skip_marker`.
+# For now the defaults are inlined below.
+
+INPUT=$(cat)
+COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
+
+if [ -z "$COMMAND" ]; then
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# 1. Match the five covered gh shapes. If the command is anything else,
+#    silently exit 0.
+# ---------------------------------------------------------------------------
+
+IS_GH_SUBCMD=0      # gh issue create | gh pr create | gh issue comment | gh pr comment
+IS_GH_API=0         # gh api .../issues | .../pulls
+
+if echo "$COMMAND" | grep -qE '\bgh\s+issue\s+create\b'; then IS_GH_SUBCMD=1; fi
+if echo "$COMMAND" | grep -qE '\bgh\s+pr\s+create\b'; then IS_GH_SUBCMD=1; fi
+if echo "$COMMAND" | grep -qE '\bgh\s+issue\s+comment\b'; then IS_GH_SUBCMD=1; fi
+if echo "$COMMAND" | grep -qE '\bgh\s+pr\s+comment\b'; then IS_GH_SUBCMD=1; fi
+if echo "$COMMAND" | grep -qE '\bgh\s+api\b.*\b(issues|pulls)\b'; then IS_GH_API=1; fi
+
+if [ "$IS_GH_SUBCMD" -eq 0 ] && [ "$IS_GH_API" -eq 0 ]; then
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# 2. Resolve the target repo.
+#    - From `--repo owner/name` for the subcommand shape.
+#    - From the URL path for the api shape: `gh api repos/<owner>/<repo>/...`.
+# ---------------------------------------------------------------------------
+
+TARGET_REPO=""
+
+# --repo flag, supports quoted or unquoted value.
+TARGET_REPO=$(echo "$COMMAND" | sed -nE 's/.*--repo[[:space:]]+["'"'"']?([^[:space:]"'"'"']+)["'"'"']?.*/\1/p' | head -1)
+
+if [ -z "$TARGET_REPO" ] && [ "$IS_GH_API" -eq 1 ]; then
+  # Accept both `repos/owner/repo/...` and `/repos/owner/repo/...`.
+  TARGET_REPO=$(echo "$COMMAND" | grep -oE '/?repos/[^/[:space:]]+/[^/[:space:]]+' | head -1 | sed -E 's|^/?repos/||')
+fi
+
+if [ -z "$TARGET_REPO" ]; then
+  # No target repo resolvable — the hook has nothing to evaluate. Default
+  # gh behaviour (current-dir repo) is safe to ignore here: the hook is a
+  # backstop against cross-repo leaks, not a universal scrubber.
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# 3. Determine whether the target is public / framework-class.
+#    Sources (in order):
+#      1. `.claude/project-config.*.json` → `leak_protection.public_framework_repos[]`
+#         (read via the shared config lib landed in apexyard#109)
+#      2. Shipped default: `me2resh/apexyard`
+#      3. Auto-detected: whatever the fork's `upstream` remote resolves to
+#         (unless `leak_protection.auto_detect_upstream` is set to `false`).
+# ---------------------------------------------------------------------------
+
+# Load the shared config reader if available. The hook still works without
+# it — falls through to the shipped defaults.
+REPO_ROOT_FOR_CONFIG=$(git rev-parse --show-toplevel 2>/dev/null)
+if [ -n "$REPO_ROOT_FOR_CONFIG" ] && [ -f "$REPO_ROOT_FOR_CONFIG/.claude/hooks/_lib-read-config.sh" ]; then
+  # shellcheck disable=SC1090,SC1091
+  . "$REPO_ROOT_FOR_CONFIG/.claude/hooks/_lib-read-config.sh"
+  CONFIG_REPOS=$(config_get '.leak_protection.public_framework_repos[]' 2>/dev/null | tr '\n' ' ')
+  AUTO_DETECT_UPSTREAM=$(config_get_or '.leak_protection.auto_detect_upstream' 'true')
+fi
+
+PUBLIC_REPOS="${CONFIG_REPOS:-me2resh/apexyard}"
+
+if [ "${AUTO_DETECT_UPSTREAM:-true}" != "false" ]; then
+  UPSTREAM_URL=$(git remote get-url upstream 2>/dev/null)
+  if [ -n "$UPSTREAM_URL" ]; then
+    # Parse github.com/<owner>/<repo>(.git)? from either SSH or HTTPS form.
+    UPSTREAM_SLUG=$(echo "$UPSTREAM_URL" | sed -nE 's|.*github\.com[:/]([^/]+/[^/]+)(\.git)?$|\1|p' | sed -E 's/\.git$//')
+    if [ -n "$UPSTREAM_SLUG" ]; then
+      PUBLIC_REPOS="$PUBLIC_REPOS $UPSTREAM_SLUG"
+    fi
+  fi
+fi
+
+IS_PUBLIC_TARGET=0
+for r in $PUBLIC_REPOS; do
+  if [ "$TARGET_REPO" = "$r" ]; then
+    IS_PUBLIC_TARGET=1
+    break
+  fi
+done
+
+if [ "$IS_PUBLIC_TARGET" -eq 0 ]; then
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #119

**Commit**: `e67e1afee3def53e096ddbafdeda118b93008b98`

### Summary

Adds a PreToolUse hook that blocks registered private-project identifiers (from `apexyard.projects.yaml`) in content destined for public framework repos (`me2resh/apexyard` + fork `upstream` + config list). Wires on five `gh` shapes: `issue create`, `pr create`, `issue comment`, `pr comment`, `gh api`. Ships with a rule doc, 10-case test fixture, shipped defaults, and an audit-table row.

### Checklist Results

- Technical Decisions (AgDR):Partial — rule-doc covers the "why mechanical" question, but options considered on mechanization shape were not recorded. Non-blocking. See Suggestions.

### What I verified (beyond running the committed tests)

Ran the committed fixture (10/10 pass) and probed the specific concerns on the review prompt:

| Concern | Result |
|---|---|
| `--title / -t / --body / -b` short and long forms | Pass |
| `--body-file <path>`, including quoted path with spaces | Pass |
| `-F <bare-path>` (gh subcommand body file) distinguished from `-F key=val` (gh api) | Pass |
| `gh api -F body=@<file>`, `-F body=<literal>`, `-f body='...'` | Pass |
| `gh api /repos/...` (leading slash) and `gh api repos/...` | Pass — both accepted |
| `gh api repos/.../issues/<N>/comments` and `.../pulls/<N>/comments` (review-comment endpoint) | Pass — caught |
| `gh api repos/.../contents/*` and `.../releases` | Pass — correctly not matched |
| Self-exemption — `apexyard-foo` on `me2resh/apexyard` target | Pass — still blocks (exemption is strict equality on bare name + target repo, not prefix) |
| Self-reference — `me2resh/apexyard#50` in an apexyard ticket body | Pass — correctly does not fire (belt-and-braces) |
| Whole-word boundary — `curios-doggy` must not match `curios-dog` | Pass |
| Workspace prefix-collision — `myworkspace/ws-marlow/...` must not match the workspace path `workspace/ws-marlow` | Pass — the workspace matcher rejects it; the project name `marlow` still fires, which is the correct whole-word behaviour |
| Skip marker — stderr warning with matching content, exits 0 | Pass — warning goes to stderr, not stdout |
| Quoted YAML values (`repo: "me2resh/SharpPick"`) | Pass — awk `unquote()` strips both `"` and `'` |
| Missing `_lib-read-config.sh` (bare checkout pre-#109) | Pass — falls through to inline defaults |
| Non-git working tree (no `git rev-parse`) | Pass — `REPO_ROOT_FOR_CONFIG` empty, skips sourcing, uses inline defaults |
| `jq` absent entirely | No-op (first line `jq -r '.tool_input.command'` fails → empty `$COMMAND` → early exit 0). Consistent with other hooks' jq dependency; acceptable since Claude Code hooks universally assume jq. |
| `--input <file.json>` stdin/JSON-file body on `gh api` | Silent no-op — acknowledged gap, documented in the hook's header comment and the PR's Scope section. |

Exit codes reviewed: 2 on detected leak; 0 on every documented no-op path (skip marker, missing registry, non-public target, empty body, unparseable command, target name/slug self-reference). No accidental exit 1 paths.

### Issues Found

None blocking.

1. **Registry parser brittleness on key ordering.** The awk parser's `NAME` rule only matches lines starting with `- name:`, so a project entry that puts `- repo: …` first and `name: …` second will silently drop both the name AND the repo (the second line `name: …` has no `-` and doesn't match `/^[[:space:]]*- name:/`; the first line `- repo: …` doesn't match `/^[[:space:]]*repo:/` because of the dash). Verified locally — only the workspace survives that shape. This is an inherited limitation from the `/start-ticket` parser and is documented in the hook's own header, so it matches team convention. Mitigation options for a follow-up: either a more tolerant awk state machine, or falling back to `yq` when available and the awk result looks incomplete. Not a blocker because apexyard's own example registry and `/handover`-written registries all start entries with `- name:`.

2. **AgDR for the implementation-shape choice.** The rule-doc at `.claude/rules/leak-protection.md` explains *why mechanical enforcement* is the right shape (excellent — matches the sibling-to-`check-secrets.sh` framing) but doesn't record options considered on *which* mechanical shape: scan on `gh` invocation (chosen) vs scan on `git commit` on a staged-diff basis vs CI-time scrub, registry-backed (chosen) vs static denylist, fail-closed on match (chosen) vs warn-and-pass. AgDR-0001 set the precedent that a new hook class is a decision worth recording with options considered. Suggestion: follow-up `AgDR-0007-leak-protection-scope.md` linked from row 10 of `docs/rule-audit.md`. Not blocking this PR — the rule-doc already captures the rationale that matters for a future maintainer; the AgDR would record the tradeoffs in the apexyard-historical-record format.

3. **`--input <file>` / stdin-driven `gh api`.** Called out in the hook header as an accepted gap and repeated in the PR's Scope section. Low-frequency shape but worth a follow-up ticket if anyone starts using it for cross-repo tooling — the hook silently no-ops on that path today.

4. **Per-key project-config override.** The `leak_protection` subtree adds `public_framework_repos`, `auto_detect_upstream`, `skip_marker`. All three are exercised. A future `name_denylist` (to suppress specific short-name collisions like `auth` / `core` without losing the whole project from the registry) would round out the false-positive-handling story already written in the rule doc's § False-positive handling. Not needed until a user hits it.

### Scope discipline

Clean — 6 files, all directly in service of #110: the hook, the rule doc, the config default subtree, the wiring, the test fixture, and the audit-table row. The `.claude/project-config.defaults.json` extension (adding the `leak_protection` subtree) is the right place for this config and is a natural fit with the #109 config-lib introduction — not scope bleed.

### Verdict

**APPROVE (as comment — agent cannot self-approve when author == reviewer)**

The hook is correct across every argument shape I could exercise; the two narrow design divergences from the ticket body (self-exemption on same-name target; workspace-path "sub-paths allowed, prefix-collision rejected" rule) are both the right calls and are documented in-header. Tests cover the critical paths. Fallbacks for missing config / non-git / absent `jq` / quoted YAML are all sound. PR description is thorough — Summary, How it works, Testing, Scope, Follow-ups, Glossary all present.

Recommending merge subject to CEO explicit per-PR approval per `pr-workflow.md` (plan-level go does not authorise the merge). Marker written locally at `.claude/session/reviews/119-rex.approved` with HEAD `e67e1afee3def53e096ddbafdeda118b93008b98`.

---
📌 Reviewed commit: `e67e1afee3def53e096ddbafdeda118b93008b98`

Issue #55, opened 2026-04-14
[Bug] Merge-gate hooks read local HEAD instead of PR HEAD — forces manual 'gh pr checkout' before every merge


atlas-apex:
## Problem

The three merge-gate hooks (\`block-unreviewed-merge.sh\`, \`require-design-review-for-ui.sh\`, \`block-merge-on-red-ci.sh\`) validate approval markers by comparing the marker SHA to \`$(git rev-parse HEAD)\` — i.e. the **local** HEAD. But \`gh pr merge <N>\` merges the PR's branch on GitHub, which is almost certainly NOT the local HEAD (local is usually \`main\` or a different feature branch).

Result: every \`gh pr merge N\` is rejected with:

\`\`\`
BLOCKED: Code-reviewer approved commit <pr-head> but HEAD is now <local-head>.
\`\`\`

Workaround applied in every merge: \`gh pr checkout N && gh pr merge N\`. Tedious, error-prone, and creates extra local branches that have to be cleaned up. Hit on every merge during the 2026-04-14 merge train (6 PRs).

## Proposed fix

In each merge-gate hook, read the PR's real HEAD from the command itself, not from local git state:

\`\`\`bash
# Extract PR number from the command (already done via _lib-extract-pr.sh)
PR_NUMBER=\$(extract_pr_number \"\$COMMAND\")

# Read the PR's actual HEAD via gh — NOT via git rev-parse HEAD
PR_HEAD=\$(gh pr view \"\$PR_NUMBER\" --json headRefOid --jq .headRefOid 2>/dev/null)

# Compare marker against PR_HEAD, not local HEAD
if [ \"\$(cat \"\$MARKER\")\" != \"\$PR_HEAD\" ]; then
  # stale approval
fi
\`\`\`

Needs the same change in all three hooks. Share the \`PR_HEAD\` resolution via \`_lib-extract-pr.sh\` (already exists after #47) for consistency.

## Cross-repo caveat

\`gh pr view\` without \`--repo\` reads the current repo. For \`gh api repos/<owner>/<repo>/pulls/<N>/merge\` shape, the owner/repo is already extracted by \`_lib-extract-pr.sh\` — pass it to \`gh pr view --repo <owner/repo> <N>\` so cross-repo merges work (e.g. contributors from a fork).

## Acceptance Criteria

- [ ] All three hooks read PR HEAD via \`gh pr view\`, not \`git rev-parse HEAD\`
[...]

PR #120, opened 2026-04-24
chore(#115): warn-stale-review-markers.sh PostToolUse hook

### LINKED ISSUE #115: [Chore] warn-stale-review-markers.sh — surface invalidated reviews at push-time, not merge-time
## Driver

Review markers (`<pr>-rex.approved`, `<pr>-ceo.approved`, `<pr>-design.approved`) are bound to a specific HEAD SHA. When the author pushes new commits to an approved PR, the markers go stale — the merge gate catches this, but only at `gh pr merge` time, which can be hours or days after the push. The gap creates user-visible confusion (*"Rex already approved, why is the merge blocked?"*) and false-starts at merge time.

## Scope

Add `.claude/hooks/warn-stale-review-markers.sh`, wired to `PostToolUse` on `Bash` matching `git push …`.

The hook:

1. After a successful `git push`, detect the PR number for the current branch (`gh pr view --json number --jq '.number'`).
2. If no PR exists, exit 0.
3. Check `$(git rev-parse --show-toplevel)/.claude/session/reviews/<pr>-*.approved` for markers.
4. For each existing marker, compare its SHA (file contents) against the current pushed HEAD.
5. If they differ, print a one-line warning per stale marker:

   ```
   ⚠ Stale review marker: <pr>-rex.approved (was <old-sha>, now <new-sha>)
     Invoke /code-review again on the new HEAD before merging.
   ```

6. Optional per-config: **auto-delete** stale markers instead of warning (belt-and-braces — the merge gate still catches missed re-reviews, so deletion is conservative):

   ```yaml
   # .claude/project-config.json
   review_markers:
     on_stale: warn   # or: delete
   ```

## Acceptance Criteria

- [ ] Hook fires on `git push`; silent if no PR exists for the branch.
- [ ] One clear warning line per stale marker, naming the marker and the SHA diff.
- [ ] Hook is PostToolUse, so it does not block the push.
- [ ] `on_stale: delete` mode removes stale markers; default mode only warns.
- [ ] Fresh markers (SHA matches current HEAD) produce no output.
- [ ] Fixtures in `.claude/hooks/tests/` cover: no markers, fresh markers, stale markers (warn), stale markers (delete), no PR.
- [ ] `docs/rule-audit.md` entry added.

## Risks / Dependencies

- `git push` into a branch that doesn't have a PR yet should produce no output. Covered by the early-exit.
- `gh pr view` requires network. On offline push, the hook should fail silently (exit 0) — covered.
- Choosing `on_stale: delete` as default is tempting but would silently undo review state. Default to `warn` and let fork owners opt in.

## Glossary

| Term | Definition |
|------|------------|
| Review marker | A file under `.claude/session/reviews/` containing the 40-char SHA of the commit that was approved. Written by the reviewer (Rex / CEO skill / design skill), checked by the merge gate. |
| Stale marker | A review marker whose SHA no longer matches the current HEAD of the PR — usually because new commits were pushed after approval. |
| PostToolUse | A Claude Code hook that runs after a tool call succeeds. Cannot block; used for warnings, auditing, and cleanup. |

atlas-apex:
## Summary

Adds a new PostToolUse hook `warn-stale-review-markers.sh` that fires after `git push` and surfaces review markers invalidated by new commits — immediately, instead of at `gh pr merge` time.

The gap it closes: when an author pushes new commits to a PR that already has Rex / CEO / design approvals on record, those markers are now stale (their recorded SHA ≠ PR HEAD). The merge gate catches this, but only at merge attempt — which can be hours or days after the push. Result: confusing "why is my merge blocked, Rex already approved?" moments. This hook makes the invalidation visible the moment it happens.

- New hook: `.claude/hooks/warn-stale-review-markers.sh` (~135 LOC, PostToolUse, non-blocking)
- Tests: `.claude/hooks/tests/test_warn_stale_review_markers.sh` — 8 cases, all pass
- Config integration: reads `.review_markers.on_stale` (`warn` default, `delete` opt-in) via the shared config lib landed in #109
- Defaults: extended `.claude/project-config.defaults.json` with a `review_markers` subtree
- Wired into `.claude/settings.json` — PostToolUse matcher on `Bash` commands matching `git push`
- Audit entry: new row in Section 3 of `docs/rule-audit.md`

## How it works

1. Fires after a successful `git push`. Non-blocking by design — PostToolUse exit 2 would push noise into the conversation; this hook is purely informational.
2. Resolves the PR number for the current branch via `gh pr view --json number`. Silent exit 0 if no PR.
3. Resolves the PR HEAD via `gh pr view --json headRefOid` — same source of truth the merge-gate hooks use post-#47 / #55. Falls back to local HEAD with a visible warning if `gh` is offline.
4. Scans `$(git rev-parse --show-toplevel)/.claude/session/reviews/<pr>-*.approved` for files whose recorded SHA doesn't match HEAD.
5. **warn mode** (default): prints one stderr line per stale marker, naming the file and the old/new SHAs, suggesting re-invocation.
6. **delete mode** (opt-in via `review_markers.on_stale: "delete"`): removes stale markers so the merge gate's "marker missing" message is crisper than "marker stale".
7. Silent on push failure (detected via stderr patterns from the `git push` output and the natural property that a failed push means the remote HEAD didn't advance — markers stay fresh).

## Testing

Ran the committed test fixture — 8 cases, all pass:

```
$ bash .claude/hooks/tests/test_warn_stale_review_markers.sh
PASS [no-pr-silent]
PASS [no-markers-silent]
PASS [fresh-markers-silent]
PASS [stale-rex-warn]
PASS [stale-ceo-warn]
PASS [stale-design-warn]
PASS [stale-rex-delete]
PASS [push-failed-silent]
PASS: 8   FAIL: 0
```

Small tangential fix in the test fixture: the sandbox setup now copies the shared config reader (`_lib-read-config.sh`) and the shipped defaults (`project-config.defaults.json`) into the isolated repo, so the sandboxed hook resolves config the same way it would in a real fork. Previously the sandbox didn't include the shared lib, and the `delete-mode` test would silently fall back to `warn` because `config_get_or` returned the default. Not a production issue (real forks always have both files after #109), but a correctness improvement to the test harness.

- Does not block the push. This hook is advisory.
- Does not change the merge-gate hook (`block-unreviewed-merge.sh`) — that still catches stale markers at merge time as a final backstop.
- Does not touch `/approve-merge`, `/approve-design`, or the Rex agent.
- `--no-verify` pushes bypass this hook (same as any PostToolUse). Acceptable — bypass discipline is covered by the existing rule ("never skip hooks").

## Glossary

| Term | Definition |
|------|------------|
| Review marker | A file under `.claude/session/reviews/<pr>-<reviewer>.approved` containing the 40-char SHA of the approved commit. Written by Rex / `/approve-merge` / `/approve-design`. |
| Stale marker | A marker whose SHA no longer matches the PR's HEAD — typically because new commits were pushed after approval. |
| PostToolUse | A Claude Code hook that runs AFTER a tool call succeeds. Cannot block; used for warnings, cleanup, and auditing. |
| `warn` / `delete` | The two modes this hook supports. Default `warn` prints a stderr line per stale marker; opt-in `delete` removes the marker so the merge gate reports "missing" rather than "stale". |

Closes #115

### DIFF
--- a/.claude/hooks/tests/test_warn_stale_review_markers.sh
+++ b/.claude/hooks/tests/test_warn_stale_review_markers.sh
@@ -1,0 +1,324 @@
+#!/bin/bash
+# Smoke tests for .claude/hooks/warn-stale-review-markers.sh
+#
+# Each case:
+#   - sets up an isolated sandbox repo under $TMPDIR
+#   - stages marker files (or not) under .claude/session/reviews/
+#   - stubs `gh` on PATH to return deterministic values
+#   - pipes a synthetic PostToolUse JSON blob into the hook
+#   - asserts the hook's stderr / exit code / marker-file state
+#
+# Exit 0 means all cases passed. Exit 1 on first failure with a clear message.
+
+set -u
+
+HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/warn-stale-review-markers.sh"
+if [ ! -x "$HOOK_SRC" ]; then
+  echo "FAIL: hook not found or not executable at $HOOK_SRC" >&2
+  exit 1
+fi
+
+PASS=0
+FAIL=0
+FAILED_CASES=""
+
+# Build a fresh sandbox for each case: a tiny git repo with an onboarding.yaml
+# (so the settings.json resolver would find it), the hook copied in, and a
+# PATH-shim dir where we drop a fake `gh` script per-case.
+make_sandbox() {
+  local sb
+  sb=$(mktemp -d)
+  (
+    cd "$sb" || exit 1
+    git init -q
+    git config user.email "test@example.com"
+    git config user.name "test"
+    touch onboarding.yaml
+    git add onboarding.yaml
+    git commit -q -m "init"
+  )
+  mkdir -p "$sb/.claude/hooks" "$sb/.claude/session/reviews" "$sb/bin"
+  cp "$HOOK_SRC" "$sb/.claude/hooks/warn-stale-review-markers.sh"
+  chmod +x "$sb/.claude/hooks/warn-stale-review-markers.sh"
+  # The hook sources the shared config reader landed in apexyard#109. The
+  # sandbox therefore needs both the reader and the shipped defaults so
+  # config lookups resolve the same way they do in a real fork.
+  local src_root
+  src_root=$(cd "$(dirname "$0")/../../.." && pwd)
+  if [ -f "$src_root/.claude/hooks/_lib-read-config.sh" ]; then
+    cp "$src_root/.claude/hooks/_lib-read-config.sh" "$sb/.claude/hooks/_lib-read-config.sh"
+  fi
+  if [ -f "$src_root/.claude/project-config.defaults.json" ]; then
+    cp "$src_root/.claude/project-config.defaults.json" "$sb/.claude/project-config.defaults.json"
+  fi
+  echo "$sb"
+}
+
+# Write a gh stub. Arguments: sandbox dir, behaviour keyword.
+#   no-pr        → all `gh pr view` calls exit 1 (no PR for branch)
+#   pr:<n>:<sha> → `gh pr view --json number` returns <n>;
+#                  `gh pr view <n> --json headRefOid` returns <sha>
+#   offline      → every gh call exits 1 with network error
+write_gh_stub() {
+  local sb="$1"
+  local mode="$2"
+  cat > "$sb/bin/gh" <<'STUB_HEAD'
+#!/bin/bash
+# Test stub for gh — behaviour controlled by $GH_STUB_MODE.
+args="$*"
+STUB_HEAD
+  cat >> "$sb/bin/gh" <<STUB_BODY
+mode="$mode"
+STUB_BODY
+  cat >> "$sb/bin/gh" <<'STUB_TAIL'
+case "$mode" in
+  no-pr)
+    # Simulates the "no PR for current branch" path.
+    exit 1
+    ;;
+  offline)
+    echo "error connecting to api.github.com" >&2
+    exit 1
+    ;;
+  pr:*)
+    # pr:<num>:<sha>
+    num="${mode#pr:}"
+    num="${num%%:*}"
+    sha="${mode##*:}"
+    # Route by args.
+    if [[ "$args" == *"--json number"* && "$args" != *"headRefOid"* ]]; then
+      echo "$num"
+      exit 0
+    fi
+    if [[ "$args" == *"headRefOid"* ]]; then
+      echo "$sha"
+      exit 0
+    fi
+    # Fallback: pretend success but empty.
+    exit 0
+    ;;
+esac
+exit 1
+STUB_TAIL
+  chmod +x "$sb/bin/gh"
+}
+
+# Run the hook inside a sandbox with a given stdin JSON and assert conditions.
+# Args: sandbox, json, expected_stderr_grep (empty for "must be silent"),
+#       expected_exit (always 0 per spec).
+run_hook() {
+  local sb="$1"
+  local json="$2"
+  local expect_grep="$3"
+  local case_name="$4"
+
+  local stderr_file
+  stderr_file=$(mktemp)
+  (
+    cd "$sb" || exit 1
+    # PATH must be exported so the piped hook subshell sees the stub.
+    # A var-prefix on `printf` alone scopes only to printf.
+    export PATH="$sb/bin:$PATH"
+    printf '%s' "$json" \
+      | "$sb/.claude/hooks/warn-stale-review-markers.sh" 2>"$stderr_file"
+  )
+  local rc=$?
+
+  if [ "$rc" -ne 0 ]; then
+    echo "FAIL [$case_name]: hook exited $rc (expected 0)" >&2
+    sed 's/^/    stderr: /' "$stderr_file" >&2
+    rm -f "$stderr_file"
+    FAIL=$((FAIL+1))
+    FAILED_CASES="$FAILED_CASES $case_name"
+    return 1
+  fi
+
+  if [ -z "$expect_grep" ]; then
+    # Silent expected — but tolerate the "falling back to local HEAD" WARN
+    # line because that's emitted by design when gh is unavailable.
+    local noise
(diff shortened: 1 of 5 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #120

**Commit**: `34450fc2b8b7f901e98f08f896d8b928b90a4bd5`

### Summary

Adds `warn-stale-review-markers.sh` — a PostToolUse hook on `git push` that surfaces review markers whose recorded SHA no longer matches the PR's HEAD. Closes the gap between "author pushes past an approval" (current session) and "the merge gate catches it" (hours or days later). Non-blocking by design — exit 0 on every code path — with the merge gate (`block-unreviewed-merge.sh`) remaining the enforcement point. Ships with 8 unit-style sandbox tests, all passing locally.

### Checklist Results

- Architecture & Design:      **Pass** — single-purpose PostToolUse hook, no new layers or abstractions, mirrors the existing merge-gate hook patterns.
- Code Quality:                **Pass** — dense comments explaining the "why" at every non-obvious step, consistent style with the rest of `.claude/hooks/`.
- Testing:                     **Pass** — 8 cases cover no-PR, no-markers, fresh, stale (rex/ceo/design), delete-mode, and push-failure paths. All green on my local run.
- Security:                    **Pass** — no secrets, no user input reaching shell interpolation unsafely. `$PR_NUMBER` from gh is used in a glob and comes from an authenticated GitHub call; marker file contents are read as opaque strings and only compared, never executed.
- Performance:                 **Pass** — two gh calls at most (PR number, then HEAD); cheap enough for a PostToolUse fire rate tied to `git push`.
- PR Description & Glossary:   **Pass** — summary is explicit, constraints and modes are laid out, glossary covers review-marker / stale-marker / PostToolUse / warn-vs-delete.
- Technical Decisions (AgDR):  **N/A** — the key design call ("hook is advisory only; merge gate remains enforcement") is a direct consequence of the existing PostToolUse platform constraint ("PostToolUse exit 2 would surface as a nudge; inappropriate here"), not a fresh architectural decision. It's already documented in-line in the hook header and the PR body. Not worth a standalone AgDR in my judgement.

### Verification of focus items

1. **PostToolUse non-blocking discipline** — confirmed. Every code path reaches `exit 0` (explicit early-exits at lines 49, 54, 72, 78, 84, 93, 108 and the final 156). No `exit 2`, no `exit 1` anywhere.

2. **PR HEAD resolution** — confirmed parity with `block-unreviewed-merge.sh` post-#47/#55. Uses `gh pr view "$PR_NUMBER" --json headRefOid --jq '.headRefOid'`, falls back to `git rev-parse HEAD` with a visible `WARN:` stderr line, then silent-exits if both fail.

3. **Push-failure detection** — works for the common case, with a minor regex oddity worth noting (non-blocking, see below).

4. **Marker globbing** — confirmed. `shopt -s nullglob 2>/dev/null` + the `[ -f "$MARKER" ] || continue` inside the loop is belt-and-braces, handles the literal-glob pitfall even on old bash where `shopt` might be quirky. Tested on macOS bash 3.2 via the sandbox — case 2 ("no-markers-silent") exercises this and passes.

5. **Config integration** — confirmed. The `if [ -f "$REPO_ROOT/.claude/hooks/_lib-read-config.sh" ]` guard handles the bare-checkout case predating #109. When the lib is present, `config_get_or '.review_markers.on_stale' 'warn'` correctly picks up the default from the shipped defaults file and respects per-fork overrides. The whitelist (`if [ "$CFG" = "delete" ] || [ "$CFG" = "warn" ]`) is a nice safety net against typos or surprising values.

6. **Delete-mode correctness** — confirmed. The file is physically removed (line 149) AND a distinct message is printed ("Stale review marker deleted: …" vs "Stale review marker: … (was X, now Y)"). Test case 7 asserts both the message and the file absence.

7. **Test-harness tangential fix** — sound, not papering over a lib bug. The shared lib behaves correctly when defaults are missing (emits `{}`, per `_config_load`). The sandbox previously didn't include the defaults, which made the delete-mode test pass for the wrong reason (silently fell back to warn). Copying the lib + defaults into the sandbox aligns the test environment with a real fork's shape. Scope creep is minimal and well-motivated.

8. **Scope** — the `review_markers` subtree in `project-config.defaults.json` is in scope (the hook needs it to find `on_stale`). No other scope-bleed detected.

9. **AgDR** — see checklist above. Judgement: not required for this PR.

10. **Hook-chain ordering** — confirmed no conflict. On `dev`, the only other PostToolUse/Bash hook is `auto-code-review.sh` scoped to `Bash(gh pr create *)`. This new hook is scoped to `Bash(git push *)`. Matchers are disjoint; both would fire on a `git push` only if an exotic shell-ism chained the two, which wouldn't happen in practice.

### Nice-to-haves (non-blocking)

1. **Regex anchor oddity at line 69.** The pattern `'\b(rejected|failed to push|^fatal:|^error:)\b'` — the outer `\b` word-boundaries effectively dead-code the `^fatal:` and `^error:` alternatives, because `^` inside the alternation doesn't play well with `\b` on either side. Verified against real failure output: the pattern still catches standard git-push failures via the `rejected` and `failed to push` alternatives (which do appear in every non-fast-forward / auth / network failure I could find), so the bug is latent, not active. If you wanted to tighten this up, dropping the outer `\b...\b` and letting the alternation stand alone (`'(rejected|failed to push|^fatal:|^error:)'`) would un-dead-code the anchors. Low priority — defer to a future follow-up if you like.

2. **Push-failure false-positive risk.** If a successful push's stderr ever contains the literal word "rejected" (e.g. a commit message echoed by a weird hook or a non-standard remote), the hook would silently exit instead of warning. The remote HEAD actually moved, so a stale marker would go unflagged until merge-time. Probability is effectively zero with standard git remotes — worth mentioning as a known edge case but not worth complicating the heuristic for.

### Verdict

**APPROVED** (via comment, since the author and reviewer are the same GitHub identity). Rex marker written at `/Users/ahmed/Projects/apexstack/.claude/session/reviews/120-rex.approved` with SHA `34450fc2b8b7f901e98f08f896d8b928b90a4bd5`.

---
Reviewed commit: `34450fc2b8b7f901e98f08f896d8b928b90a4bd5`

PR #121, opened 2026-04-24
chore(#111): upgrade pre-push-gate from advisory reminder to blocking check-runner

### LINKED ISSUE #111: [Chore] pre-push-gate.sh — upgrade from advisory reminder to blocking check-runner
## Driver

`pre-push-gate.sh` today **reminds** the user to run lint / typecheck / test / build before `git push` — prints the checklist and exits 0. The rule it enforces (from `.claude/rules/pr-workflow.md`) is explicitly labelled a **HARD STOP** — *"Never push without running CI checks locally."* A reminder is not a hard stop. Observed failure mode: the gate prints, the agent continues, CI fails on the PR, cycle wastes time.

## Scope

Upgrade `pre-push-gate.sh` from advisory to blocking:

1. Read the project's pre-push command set from config (new field, default ships the canonical set):

   ```yaml
   # .claude/project-config.json
   pre_push:
     commands:
       - name: lint
         run: "npm run lint"
       - name: typecheck
         run: "npm run typecheck"
       - name: test
         run: "npm run test"
       - name: build
         run: "npm run build"
     skip_marker: "<!-- pre-push: skip -->"   # for emergency hotfix paths
   ```

2. Execute each command in sequence. Capture stdout/stderr.
3. On any non-zero: exit 2 with a concise summary naming the failing command + a pointer to the full output log.
4. Escape hatch: the skip marker, written into the most recent commit message, lets a push through. Use for true emergencies — leaves a grep-able trace.
5. Skip condition (silent, no block): no `.claude/project-config.json` and no `package.json` — treat as a non-runnable repo (e.g. docs-only).
6. Performance: run commands in parallel where safe (lint + typecheck are independent; test + build typically are). Configurable.

## Acceptance Criteria

- [ ] Hook runs configured commands on `git push` and blocks (exit 2) on any red.
- [ ] Hook reads the command list from project config; ships a sensible default.
- [ ] Failing command name + last 20 lines of output included in the block message.
- [ ] Skip marker in the HEAD commit bypasses the gate (and prints a warning that it was bypassed).
- [ ] `docs/rule-audit.md` updates the hook's entry from "advisory" to "blocking".
- [ ] Fixtures in `.claude/hooks/tests/` cover pass, fail, skip, and no-config paths.

## Risks / Dependencies

- Slow commands (build) extend push time from seconds to minutes. Mitigation: parallel execution; option to disable `build` via config.
- Command failures that aren't the developer's fault (flaky tests, transient network) cause false blocks. Mitigation: skip marker escape hatch, clear.
- Hook executes arbitrary commands from config — careful about executable paths, quoting, `set -e` behaviour. Lean on shellcheck; add a test that malformed config fails loudly rather than silently.
- Shares config schema with apexyard#109. Land after #109.

## Glossary

| Term | Definition |
|------|------------|
| Advisory hook | Prints a message and exits 0. Cannot block. |
| Blocking hook | Exits 2, which Claude Code surfaces as a block and refuses the tool call. |
| Skip marker | A token in a commit message or body that tells the hook to let this invocation through, with a visible warning. |

atlas-apex:
## Summary

Upgrades `pre-push-gate.sh` from an advisory reminder (prints a checklist and exits 0) to a blocking check-runner (reads a configured list of shell commands, executes them, blocks on red with exit 2).

The rule in `.claude/rules/pr-workflow.md § Before git push (HARD STOP)` has always said *"Never push without running CI checks locally."* The old hook printed a reminder, so the mechanical enforcement didn't actually match the written rule. This PR closes that asymmetry.

- Hook now reads `.pre_push.commands` from project config (shared reader from #109) and runs each entry's `run` field in sequence.
- First non-zero exit blocks the push; the failing command name + last 20 lines of output go to stderr.
- Shipped default is an empty command list — hook stays a no-op until a fork opts in by defining its checks. (Deliberate: the framework repo itself doesn't have lint/test/build to run; each fork wires its own CI per its stack.)
- Emergency bypass: `<!-- pre-push: skip -->` in the HEAD commit message. Grep-able on purpose so bypasses are auditable.
- Fail-fast: once a command fails, subsequent commands don't run. Parallel execution is a polish-level follow-up.

Changed files:

- `.claude/hooks/pre-push-gate.sh` — rewritten
- `.claude/hooks/tests/test_pre_push_gate.sh` — new, 7 cases
- `.claude/project-config.defaults.json` — added `pre_push.commands` subtree (empty by default)
- `docs/rule-audit.md` — flipped "Before git push" rule from "partial" to "yes" + updated the footnote

## Testing

```
$ bash .claude/hooks/tests/test_pre_push_gate.sh
PASS [non-git-push-silent]
PASS [empty-commands-noop]
PASS [passing-command]
PASS [failing-command-blocks]
PASS [skip-marker-bypasses]
PASS [fail-fast-on-first-red]
PASS [no-config-noop]
PASS: 7   FAIL: 0
```

Also dogfood: this very PR was pushed through the upgraded hook. Since the framework repo has no configured `.pre_push.commands` in its defaults file, the gate exits 0 silently — proving the "empty-commands-noop" path works end-to-end against the apexyard repo itself.

## What a per-fork override looks like

Minimal example in a fork's `.claude/project-config.json`:

```json
{
  "pre_push": {
    "commands": [
      { "name": "lint",      "run": "npm run lint" },
      { "name": "typecheck", "run": "npm run typecheck" },
      { "name": "test",      "run": "npm run test" },
      { "name": "build",     "run": "npm run build" }
    ]
  }
}
```

Teams on yarn, pnpm, make, cargo, go, mix and so on all configure here — the hook doesn't care about the package manager. The command runs in a sub-shell rooted at the repo top-level.

- No change to the shipped default command set. Adopters opt in per-fork; the framework itself stays no-op. Changing the shipped default is a separate call.
- No parallel execution. Fail-fast sequential is the v1. If fork owners report the time cost is painful, add a `parallel: true` flag in a follow-up.
- No change to `pre-push-gate.sh`'s matcher wiring in `.claude/settings.json` — same matcher, new behaviour.

## Follow-ups

- Parallel execution within a command group (lint + typecheck can usually run in parallel).
- Richer command spec (e.g. `run_if_exists: "package.json"` to skip on repos without that file).
- Deprecation warning on the legacy flat `commit_types` key in `validate-commit-format.sh` (noted in AgDR-0006 as a deferred item).

## Glossary

| Term | Definition |
|------|------------|
| Pre-push gate | The hook that fires on `Bash` commands matching `git push`. Enforces the HARD STOP rule that local checks must pass before pushing. |
| Skip marker | A literal token in the HEAD commit message that lets a single push through without running checks. Leaves a grep-able trace. |
| Fail-fast | The first non-zero command aborts the whole gate. The alternative (run all, aggregate) is a follow-up. |

Closes #111

### DIFF
--- a/.claude/hooks/pre-push-gate.sh
+++ b/.claude/hooks/pre-push-gate.sh
@@ -1,8 +1,32 @@
 #!/bin/bash
-# Reminds the user to run CI checks locally before pushing.
-# Outputs a warning, does not block.
+# pre-push-gate.sh — blocks `git push` on red local checks.
 #
-# Customize the checklist below for your project's actual commands.
+# Upgraded from an advisory reminder (pre-#111) to a blocking check-runner:
+# reads a list of shell commands from `.claude/project-config.*.json`
+# (`.pre_push.commands`) and runs them in sequence before the push is
+# allowed through. Non-zero exit from any command blocks the push.
+#
+# This implements the HARD STOP documented in `.claude/rules/pr-workflow.md`
+# — "Never push without running CI checks locally." Previously the rule
+# was self-discipline; now it's mechanical.
+#
+# Silent pass conditions (exit 0, no output):
+#   - Not a `git push` command.
+#   - No `.claude/project-config.defaults.json` AND no `package.json` in the
+#     repo → treat as a non-runnable repo (docs-only, newly-forked, etc.).
+#   - HEAD commit subject contains the skip marker `<!-- pre-push: skip -->`
+#     → emergency escape hatch; prints a visible WARN and lets the push
+#     through. Leaves a grep-able trace so bypasses are auditable.
+#
+# Configured commands (example, from the shipped defaults):
+#   - lint:      npm run lint
+#   - typecheck: npm run typecheck
+#   - test:      npm run test
+#   - build:     npm run build
+#
+# Skip marker: include the literal string `<!-- pre-push: skip -->` in the
+# HEAD commit message (subject or body) to bypass for that one push.
+# The hook prints the bypassed command set to stderr so the skip is visible.
 
 INPUT=$(cat)
 COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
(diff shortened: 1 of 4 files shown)


me2resh (approved):

me2resh (approved):
approved

atlas-apex (commented):
## Code Review: PR #121

**Commit**: `09db35ea89fa240eede2f3e1891728e404ea7222`

### Summary

Upgrades `pre-push-gate.sh` from an advisory reminder (prints a checklist, exits 0) to a blocking check-runner that reads `.pre_push.commands` from project config (via the shared reader from #109), executes them sequentially, and blocks with exit 2 on the first non-zero. Ships with an empty default command list so the framework repo itself stays a no-op; per-fork overrides opt in. Includes an auditable emergency escape hatch (`<!-- pre-push: skip -->` in HEAD commit message), updates `docs/rule-audit.md`, and ships 7 sandboxed smoke-test cases — all passing locally.

### Checklist Results

- Architecture & Design:          Pass
- Code Quality:                   Pass
- Testing:                        Pass
- Security:                       Pass
- Performance:                    Pass (fail-fast sequential; parallel noted as follow-up)
- PR Description & Glossary:      Pass
- Technical Decisions (AgDR):     N/A — design choices (empty default, skip marker, fail-fast, tail -20) are documented inline in the PR description and `docs/rule-audit.md` footnote; no new library/framework/pattern introduced beyond what #109 already landed.

### Verification

I re-ran the ten focus areas locally against the actual sandbox + hook:

| # | Check | Result |
|---|-------|--------|
| 1 | Blocking semantics — failing command exits **2** (not 1/0) | Pass — `rc=2` confirmed |
| 2 | Shell injection surface — only `tool_input.command` read via `jq -r`; `title` / `body` / other fields never evaluated | Pass — crafted payload with `$(touch /tmp/pwn-…)` in title/body produced zero sentinels |
| 3 | Skip marker — works in **both** subject and body via `grep -qF -- "$SKIP_MARKER"` (no regex escaping required from user) | Pass — both positions tested, rc=0 despite a failing command in config |
| 4 | Empty default — `pre_push.commands: []` produces exit 0 with **zero stderr** | Pass — silent no-op confirmed |
| 5 | Skip-marker reporting — WARN visible on stderr, push proceeds | Pass — two-line WARN printed, rc=0 |
| 6 | Fail-fast — second command's sentinel file never created when first exits non-zero | Pass — sentinel absent after first-fail |
| 7 | `tail -20` output — correct on short (<20), stderr-only, and >20-line commands | Pass — all three cases render cleanly (short: full output; stderr-only: captured via `2>&1`; long: last 20 of 30 lines) |
| 8 | No-config-at-all path — missing `project-config.defaults.json` AND `project-config.json` → silent exit 0 | Pass — `_lib-read-config.sh` returns `{}`, `CMDS_JSON` is empty, hook exits silently |
| 9 | Test harness isolation — each case uses fresh `mktemp -d`, cleans up with `rm -rf`; stderr tempfile is per-process but written-read-removed sequentially in one process, no cross-case leak | Pass |
| 10 | Rule-audit update — "partial → yes" on "Before `git push`" | **Discuss below** |

### Issues Found

None blocking.

### Discussion — rule-audit label (Focus 10)

The flip from `partial` to `yes` on the "Before `git push`" row is defensible but deserves a tiny caveat. When a fork ships no override, the hook is a no-op, so the rule ("lint / typecheck / test / build must pass locally") is still not mechanically enforced in that repo. The new footnote `[^pre-push-111]` does spell this out ("Default is an empty list (hook is a no-op) — projects opt in by defining their checks"), so the "yes" is really "yes, *conditional on the fork populating `.pre_push.commands`*".

Two ways to read this:

- **Yes (as shipped)** — the *mechanism* exists and is production-grade. Whether a given fork uses it is that fork's problem, same as whether a fork configures its linter. The framework's job is to ship the rail; the footnote explains the rail.
- **Partial (stricter)** — the framework itself can't claim full enforcement because the default is empty. By the same logic, `block-merge-on-red-ci.sh` would also be "partial" (it assumes CI checks exist on the repo), which is not how the table labels it. So the symmetry argument favors "yes".

My call: **"yes" is the right label** — the mechanism is complete, the footnote is honest about the opt-in, and it matches how other config-driven hooks are labelled. Optional polish: a single sentence in the main row body like "enforcement activates on forks that populate `.pre_push.commands`". Not blocking.

### Nice-to-haves (non-blocking)

1. **TMP_LOG cleanup on interrupt** — if the hook is interrupted (Ctrl-C, SIGTERM) between `mktemp` and the `rm -f`, a `/tmp/pre-push-gate.XXXXXX` file leaks. A `trap 'rm -f "$TMP_LOG"' EXIT` inside the loop would close that, but the file is small and `mktemp -t` is already in $TMPDIR, so this is purely cosmetic.
2. **Per-step timing** — when the hook does fire and all commands pass, it produces zero output. For an interactive push that takes 30s of lint+test+build, a one-line "pre-push-gate: 4/4 commands passed (lint 2s, typecheck 3s, test 18s, build 7s)" summary to stderr would be reassuring. Follow-up territory.
3. **Shell choice** — `bash -c "$RUN"` always launches bash. Forks that want fish/zsh specifics would need to wrap it (`bash -c "zsh -c '...'"`). This is likely the right default — noting it here only because it's not currently documented in the header comment.

### Verdict

**APPROVED** — scope is exactly what the ticket asked for, all ten focus areas verified cleanly, test suite is clean and isolated, shell-injection surface is appropriately tight, default-empty behaviour is honest and documented. Posting as a comment (cannot self-approve when author is the same account).

---
📌 Reviewed commit: `09db35ea89fa240eede2f3e1891728e404ea7222`

atlas-apex (commented):
## Code Review: PR #121 (re-review)

**Commit**: `2752a7ab09ebabea5a103e3edd99bec53670bb99`
**Prior reviewed commit**: `09db35e` (approved, blocked on red CI for MD053)

### Scope
Focused re-review of the delta `09db35e` → `2752a7a`. Everything before `09db35e` was already reviewed and approved.

### Delta
```
docs/rule-audit.md | 2 --
1 file changed, 2 deletions(-)
```

Removed the orphaned `[^project-cmds]` footnote definition (and its surrounding blank line). This footnote had been superseded by `[^pre-push-111]` in the earlier work on this PR; the definition was accidentally left behind, which tripped markdownlint's MD053 (link-reference-definitions) rule on CI.

### Verification
- Footnote balance: grep confirms no remaining references to `[^project-cmds]` anywhere in the file. The three live footnotes (`[^lint]`, `[^pre-push-111]`, `[^self-discipline]`) are all still referenced in the body and defined at the bottom — balanced.
- Scope: delta touches only `docs/rule-audit.md`. No hook, rule, code, or config changes.
- CI: all required checks green on `2752a7a` — `markdownlint-cli2`, `lychee`, `shellcheck .claude/hooks`, `Verify Ticket ID`. The MD053 failure that blocked the previous merge is resolved.

### Checklist Results
- Architecture & Design:    N/A (docs-only delta)
- Testing:                   N/A (docs-only delta)
- Security:                  N/A (docs-only delta)
- Performance:               N/A (docs-only delta)
- PR Description & Glossary: Pass (unchanged from prior review)
- Technical Decisions (AgDR):N/A (no new decisions in this delta)

### Issues Found
None.

### Verdict
**APPROVED** — clean, minimal, correct fix. Merge gate should now clear.

---
Reviewed commit: `2752a7ab09ebabea5a103e3edd99bec53670bb99`

PR #129, opened 2026-04-25
fix(#106): CHANGELOG fallback in drift hook for squash-merged forks

### LINKED ISSUE #106: [Bug] Tag-based drift hook keeps firing after squash-merged sync PRs
## Given / When / Then

**Given** a fork of apexyard that has merged upstream via a **squash-merged** PR,
**When** a fresh Claude Code session starts on that fork,
**Then** the SessionStart drift banner keeps printing `"ApexYard: vX.Y.Z available. Run /update to sync."` even though the fork is content-caught-up to upstream v1.1.0.

## Repro

Discovered on `me2resh/ops#10` (2026-04-23) — the first real-world exercise of the `/update` skill shipped in v1.1.0:

1. Fork forks upstream `me2resh/apexyard` (pre-v1.1.0).
2. `/update` creates a sync branch; user runs `git merge upstream/main` on it, then pushes as a PR.
3. GitHub merges the PR via **squash-merge** (default setting on many repos) — 27 upstream commits collapse into one synthetic squash commit.
4. Fork's main is now at the squash SHA. Content is identical to upstream v1.1.0.
5. BUT: upstream's `v1.1.0` tag points at the upstream-native commit `4e9e840`, which is NOT an ancestor of the fork's main after squash-merge.
6. The hook runs `git tag --list --sort=-v:refname --merged main | head -1` — and since v1.1.0's target SHA isn't reachable, the hook says "v1.1.0 available" even though the fork is up-to-date content-wise.

## Root cause

`.claude/hooks/check-upstream-drift.sh` uses **commit-SHA reachability** (`git tag --merged main`) as the proxy for "has this tag been absorbed". That proxy holds for merge-commit and rebase workflows, but BREAKS for squash-merge because the squash commit has no ancestor link to the upstream commits it contains.

GitHub-default squash-merge is widespread. A hook that only works for non-default merge modes is a footgun — fork maintainers will see the banner keep firing after they've synced, conclude the banner is noisy, and ignore it (the exact failure mode v1.1.0's tag-based design was meant to prevent).

### (a) Fork-tag convention — recommended

After a sync PR merges, tag the fork's main with the upstream tag name at the squash commit. That tag IS reachable from fork main, so `--merged main` finds it.

- Add a post-merge step to the `/update` skill: "after the sync PR merges, run `git tag vX.Y.Z <squash-sha> && git push origin vX.Y.Z`".
- Document in CHANGELOG + `docs/multi-project.md`.
- Pro: zero hook change, trivial workflow addition.
- Con: manual step; easy to forget. Forks that forget see the banner keep firing — same failure mode.

### (b) Content-based drift check

Replace the `--merged main` check with a CHANGELOG-based one: "is there a `## [vX.Y.Z]` heading in `CHANGELOG.md` on the fork's main?". If yes, the fork has incorporated that release's notes (a reliable proxy for "we synced").

- Pro: works regardless of merge mode. No manual step.
- Con: assumes the fork's CHANGELOG has that exact heading format (currently true for apexyard; not guaranteed for every fork's workflow if they customise CHANGELOG).

### (c) Fallback hybrid

Current primary: `--merged main` tag check. If that fails, SECONDARY check: grep `CHANGELOG.md` on the fork's main for the `## [vX.Y.Z]` pattern. If secondary passes, treat the fork as caught up.

- Pro: keeps the working merge-commit case unchanged; squash-merge case now also works.
- Con: the CHANGELOG grep is brittle to heading-format changes.

### (d) Last-sync state file

On `/update` completion, the skill writes `.claude/session/last-upstream-sync` with the upstream SHA it merged. The hook reads that and compares against `upstream/main` HEAD.

- Pro: accurate regardless of merge mode.
- Con: session state is gitignored per-machine — would need to move this file OUT of `.claude/session/` to survive re-clones. New file location required.

### Non-fix: "just use merge-commit"

Document that forks should configure squash-off for sync PRs. Leaves the footgun in place; relies on every adopter configuring their fork correctly.

## Acceptance criteria

- [ ] After a squash-merged sync PR lands on a fork, the SessionStart banner goes silent.
- [ ] The fix doesn't regress the existing merge-commit / rebase workflows (both already work today).
- [ ] CHANGELOG + docs updated to reflect the new behaviour.
- [ ] If fix is (a) fork-tag convention, `/update` skill spec gains a "step 9" documenting the post-merge tag step.
- [ ] If fix is (b) or (c) content-based, include a regression test or a manual verification recipe.

## Discovery artifact

- Fork: `me2resh/ops`
- Symptom PR: `me2resh/ops#10` (merged 2026-04-23 via squash)
- Upstream tag unreachable: `v1.1.0` pointing at `4e9e840` in `me2resh/apexyard`
- Banner output post-merge: `ApexYard: v1.1.0 available. Run /update to sync.` (incorrect — fork is caught up)

## Severity

Medium. The feature (tag-based drift detection, shipped in v1.1.0) works correctly in merge-commit and rebase flows but silently fails in the squash-merge flow, which is GitHub's default. Every fork that adopts ApexYard using GitHub-default settings will hit this. Fork maintainers will either learn to ignore the banner (bad) or manually work around it (toil). Not a data-loss or security bug; a UX-correctness bug.

atlas-apex:
## Summary

Fixes the bug from #106: the v1.1.0 tag-based drift hook keeps firing on forks that sync via GitHub's default squash-merge. The squash collapses the upstream-tag commit into a synthetic SHA, the tag becomes unreachable from the fork's main, and `git tag --merged main` returns empty — so the hook says "vX.Y.Z available" forever.

Adds a **CHANGELOG-content fallback** that fires only when the primary tag-reachability check fails. If the fork's main has a `## [X.Y.Z]` heading matching the upstream tag's version, the release is treated as absorbed and the banner stays silent.

- `changelog_has_version()` helper does a tolerant grep against `${DEFAULT_BRANCH}:CHANGELOG.md`. Strips leading `v` from the tag for the heading match.
- Two fallback gates: (1) `LOCAL_TAG` empty (first sync, squash-merged); (2) `UPSTREAM_TAG` strictly newer than `LOCAL_TAG` (intermediate sync, squash-merged).
- Merge-commit and rebase paths unchanged — the primary tag check still works for them, the fallback only kicks in on its failure.
- Records the strategy update in `docs/agdr/AgDR-0008-changelog-fallback-for-squash-merged-forks.md` (extends AgDR-0005's original tag-based-drift design).

## Why this approach

The ticket listed four options:

- **(A) Manual fork-tag convention** — relies on humans/skill to remember; same failure mode if forgotten
- **(B) Pure CHANGELOG check** — discards the working tag-reachability fast path
- **(C) Hybrid (chosen)** — preserves working merge-commit/rebase paths; adds recovery for squash-merge
- **(D) New committed state file** — most accurate but new file convention + migration cost

(C) is in-place: zero changes to the `/update` skill, zero new files outside the AgDR, zero migration story for existing forks. The CHANGELOG format dependency is acceptable because apexyard owns the CHANGELOG and the format is stable from v1.1.0 onward. AgDR-0008 documents the option table.

## Testing

5 cases in `.claude/hooks/tests/test_check_upstream_drift.sh`. Each builds a tiny upstream + fork pair under `$TMPDIR`, simulates a specific merge mode, runs the hook, asserts banner / silence:

```
$ bash .claude/hooks/tests/test_check_upstream_drift.sh
PASS [squash-merge fork caught up to v1.1.0 → silent (FIXED by #106)]
PASS [merge-commit fork caught up to v1.1.0 → silent]
PASS [fork stopped at v1.0.0 → banner fires]
PASS [fork has its own newer tag → silent]
PASS [squash-merge but no CHANGELOG on fork → banner fires (no false silence)]
PASS: 5   FAIL: 0
```

The fifth case is the critical "fail-open" check — a fork that squash-merges but doesn't keep CHANGELOG.md should NOT be silenced incorrectly. Regression-protected.

## Acceptance criteria (from #106)

- [x] After a squash-merged sync PR lands on a fork, the SessionStart banner goes silent
- [x] Fix doesn't regress merge-commit / rebase workflows (covered by case 2)
- [x] CHANGELOG + docs updated — AgDR-0008 added; CHANGELOG.md update lands with the v1.2.0 release entry
- [x] Regression test / verification recipe — `tests/test_check_upstream_drift.sh`

- Does not modify the `/update` skill. Existing flow works as-is.
- Does not introduce a new committed state file (Option D — deferred; revisit if the CHANGELOG-format dependency turns out to bite).
- Does not retroactively fix forks already showing the misfire — they go silent the next time their fork's CHANGELOG has the upstream version's heading (typically: the next session after they manually drop a `## [X.Y.Z]` line into their fork's CHANGELOG, or after the next `/update` sync).

## Glossary

| Term | Definition |
|------|------------|
| Squash-merge | GitHub's default "Squash and merge" mode — collapses N commits in a PR into one synthetic commit on the target branch. Severs ancestor reachability to the source commits. |
| Tag reachability | `git tag --merged BRANCH` — does the tag's target SHA appear in BRANCH's ancestry? Reliable for merge-commit / rebase, not for squash-merge. |
| CHANGELOG fallback | Content-based "did we absorb this release" check via `^##\s+\[X.Y.Z\]` heading grep. Tolerant secondary signal. |
| Fail-open | If the fallback can't confirm absorption, fire the banner — don't silence on missing evidence. (Case 5 covers this.) |

Closes #106

### DIFF
--- a/.claude/hooks/check-upstream-drift.sh
+++ b/.claude/hooks/check-upstream-drift.sh
@@ -73,6 +73,37 @@ if [ "$SHOULD_FETCH" = "1" ]; then
 fi
 
 # ---------------------------------------------------------------------------
+# Helper: does the fork's CHANGELOG.md mention the given upstream version?
+#
+# Squash-merge breaks the `--merged main` tag-reachability check because a
+# squash collapses upstream commits into a synthetic commit with no ancestor
+# link to the upstream tag's target SHA. The CHANGELOG content survives the
+# squash, so a `## [X.Y.Z]` heading on the fork's default branch is a
+# reliable secondary signal that the release was absorbed.
+#
+# Format expected (matches apexyard's CHANGELOG.md from v1.1.0 onward):
+#   ## [1.1.0] — 2026-04-19
+# Tag input is `v1.1.0` — we strip the leading `v` before matching.
+#
+# Tolerant of: missing CHANGELOG.md (returns 1 — proceed with banner),
+# different prefix conventions on the heading line, version-only matches
+# with surrounding `[`/`]` brackets.
+#
+# See AgDR-0008 + apexyard#106.
+changelog_has_version() {
+  local tag="$1"
+  local version="${tag#v}"
+  # `git show <branch>:CHANGELOG.md` fetches the file at the branch tip
+  # without affecting the working tree. Falls back silently if the file
+  # doesn't exist on that branch.
+  if git show "${DEFAULT_BRANCH}:CHANGELOG.md" 2>/dev/null \
+       | grep -qE "^##[[:space:]]+\[${version}\]"; then
+    return 0
+  fi
+  return 1
+}
+
+# ---------------------------------------------------------------------------
 # Tag-based drift (primary signal)
 # ---------------------------------------------------------------------------
 # Latest upstream tag reachable from upstream's default branch, sorted by
@@ -89,7 +120,13 @@ if [ -n "$UPSTREAM_TAG" ]; then
   # Upstream has at least one tag. Steady-state path.
 
   if [ -z "$LOCAL_TAG" ]; then
-    # Fork has never merged any tag yet. First release sync is due.
+    # Fork has never merged any tag (tag-reachable). Fall back to the
+    # CHANGELOG check — squash-merged sync PRs leave no reachable tag but
+    # do absorb the release-notes content.
+    if changelog_has_version "$UPSTREAM_TAG"; then
+      exit 0
+    fi
+    # Genuinely behind. First release sync is due.
     cat <<MSG
 ApexYard: ${UPSTREAM_TAG} available. Run /update to sync.
 MSG
@@ -109,7 +146,13 @@ MSG
   NEWER=$(printf '%s\n%s\n' "$UPSTREAM_TAG" "$LOCAL_TAG" | sort -V | tail -n 1)
 
   if [ "$NEWER" = "$UPSTREAM_TAG" ]; then
-    # Upstream is strictly newer. Nag.
+    # Upstream looks strictly newer by tag. Before nagging, check the
+    # CHANGELOG fallback — the fork might have squash-merged the release
+    # without inheriting the tag SHA reachability.
+    if changelog_has_version "$UPSTREAM_TAG"; then
+      exit 0
+    fi
+    # Genuinely behind.
     cat <<MSG
 ApexYard: ${UPSTREAM_TAG} available. Run /update to sync.
 MSG
--- a/.claude/hooks/tests/test_check_upstream_drift.sh
+++ b/.claude/hooks/tests/test_check_upstream_drift.sh
@@ -83,7 +83,7 @@ case_squash_caught_up() {
   local up; up=$(make_upstream)
   local fk; fk=$(make_fork "$up")
   (
-    cd "$fk"
+    cd "$fk" || exit 1
     git checkout -q main
     git reset --hard HEAD~3 2>/dev/null
     git merge --squash upstream/main >/dev/null 2>&1
@@ -98,7 +98,7 @@ case_merge_commit_caught_up() {
   local up; up=$(make_upstream)
   local fk; fk=$(make_fork "$up")
   (
-    cd "$fk"
+    cd "$fk" || exit 1
     git checkout -q main
     git reset --hard HEAD~3 2>/dev/null
     git merge --no-edit upstream/main >/dev/null 2>&1
@@ -112,7 +112,7 @@ case_genuinely_behind() {
   local up; up=$(make_upstream)
   local fk; fk=$(make_fork "$up")
   (
-    cd "$fk"
+    cd "$fk" || exit 1
     git checkout -q main
     git reset --hard v1.0.0 2>/dev/null
   )
@@ -125,7 +125,7 @@ case_fork_ahead() {
   local up; up=$(make_upstream)
   local fk; fk=$(make_fork "$up")
   (
-    cd "$fk"
+    cd "$fk" || exit 1
     git checkout -q main
     # Fork merges v1.1.0 cleanly first
     git reset --hard HEAD~3 2>/dev/null
@@ -145,7 +145,7 @@ case_squash_no_changelog() {
   local up; up=$(make_upstream)
   local fk; fk=$(make_fork "$up")
   (
-    cd "$fk"
+    cd "$fk" || exit 1
     git checkout -q main
     git reset --hard HEAD~3 2>/dev/null
     # Squash-merge but immediately blow away CHANGELOG.md (simulate a fork that
--- a/.claude/hooks/tests/test_check_upstream_drift.sh
+++ b/.claude/hooks/tests/test_check_upstream_drift.sh
@@ -1,0 +1,175 @@
+#!/bin/bash
+# Smoke tests for .claude/hooks/check-upstream-drift.sh — focused on the
+# CHANGELOG-fallback path added by apexyard#106 (AgDR-0008).
+#
+# Each case:
+#   - builds a tiny upstream + fork pair under $TMPDIR
+#   - simulates a specific merge mode (squash-merge / merge-commit / no-sync)
+#   - runs the hook from the fork's directory
+#   - asserts banner output / silence
+
+set -u
+
+HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/check-upstream-drift.sh"
+PASS=0
+FAIL=0
+FAILED=""
+
+# Build an upstream with a v1.0.0 release and a v1.1.0 release; both have
+# CHANGELOG entries. Returns the upstream repo path.
+make_upstream() {
+  local up
+  up=$(mktemp -d)
+  (
+    cd "$up" || exit 1
+    git init -q -b main
+    git config user.email t@t && git config user.name t
+    echo "init" > a.txt && git add a.txt && git commit -q -m "init"
+    echo "v1.0.0" > a.txt && git commit -q -am "release v1.0.0"
+    printf '# Changelog\n\n## [1.0.0]\nfirst release\n' > CHANGELOG.md
+    git add CHANGELOG.md && git commit -q -m "v1.0.0 changelog"
+    git tag v1.0.0
+    echo "v1.1.0 work" >> a.txt && git commit -q -am "v1.1.0 work"
+    printf '# Changelog\n\n## [1.1.0] — 2026-04-19\nnew release\n\n## [1.0.0]\nfirst release\n' > CHANGELOG.md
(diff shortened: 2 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #129

**Commit**: `b257f814b43b8c88049dae7b26f926c6f870054c`

### Summary

Adds a CHANGELOG-content fallback to `check-upstream-drift.sh` so squash-merged sync PRs don't keep firing the banner forever. Tag-reachability stays the primary signal; the new `changelog_has_version()` helper fires only on its failure, in both the `LOCAL_TAG`-empty path and the `UPSTREAM_TAG > LOCAL_TAG` path. Backed by AgDR-0008 (extends AgDR-0005) and a 5-case sandbox test that builds tiny upstream + fork pairs under `$TMPDIR`. Sound design, well-bounded blast radius.

### Checklist Results

- Testing:                   Pass (5/5 cases pass when run locally)
- Security:                  Pass (read-only `git show`, no shell injection vector)
- Performance:               Pass (one extra grep on `${DEFAULT_BRANCH}:CHANGELOG.md`, only when primary fails)
- PR Description & Glossary: Pass (clear summary, glossary present, all 4 options documented)
- Technical Decisions (AgDR):Pass (AgDR-0008 with Y-statement, options A/B/C/D, decision, consequences, non-consequences, cross-ref to AgDR-0005)

### Issues Found — BLOCKERS

**1. CI is red.** Per `.claude/rules/pr-quality.md` "No Red CI Before Merge", this must be green before merge.

**a. `markdownlint-cli2` failing** — `docs/agdr/AgDR-0008-changelog-fallback-for-squash-merged-forks.md`:

```
:43 MD032/blanks-around-lists  Lists should be surrounded by blank lines  [Context: "- Tag-reachability primary che..."]
:48 MD032/blanks-around-lists  [Context: "- `changelog_has_version()` he..."]
:52 MD032/blanks-around-lists  [Context: "- No CHANGELOG schema enforcem..."]
```

The Consequences section's three sub-lists (`Kept:` / `Added:` / `Non-consequences (explicitly):`) need a blank line between the bold label and the first `-` bullet. Cheap fix.

**b. `shellcheck` failing** — `.claude/hooks/tests/test_check_upstream_drift.sh`:

```
:86  SC2164  Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
:101 SC2164
:115 SC2164
:128 SC2164
:148 SC2164
```

The pattern is `( cd "$fk"; ... )` inside subshells. ShellCheck is right — if `cd` fails the subshell silently runs the `git` calls in the wrong directory and the test passes for the wrong reason. Replace with `( cd "$fk" || exit 1; ... )`. Note that line 26 (`make_upstream`) and line 40 (`make_fork`) already use the `|| exit 1` pattern correctly — these 5 case bodies just need the same treatment.

### Verification of focus areas

Confirmed each item the author flagged for review:

- **Regex tolerance** — verified `^##[[:space:]]+\[VERSION\]` matches `## [1.1.0] — 2026-04-19`, `## [1.1.0]`, `##  [1.1.0]`, `## \t[1.1.0]`; correctly rejects `## [foo-1.1.0]`, `### [1.1.0]`, `## 1.1.0`.
- **Tag conversion** — `${tag#v}` correctly strips one leading `v`: `v1.1.0` → `1.1.0`, `1.1.0` → `1.1.0`, `v2.0.0-rc1` → `2.0.0-rc1`.
- **Fail-open on missing CHANGELOG** — `git show "${DEFAULT_BRANCH}:CHANGELOG.md" 2>/dev/null` returns empty stdout, stderr swallowed, grep returns 1, fallback returns 1, banner correctly fires. Case 5 covers it and passes.
- **Merge-commit no regression** — case 2 uses `git merge --no-edit upstream/main` (true merge commit since fork was reset to `HEAD~3` first; not a fast-forward). Confirmed silent.
- **Fork-ahead path** — case 4 produces `v2.0.0-acme` on the fork; `sort -V` picks it as newer than `v1.1.0`, primary path returns silent at the "Local is strictly newer" branch, fallback never invoked. Correct.
- **Two fallback gates** — confirmed in the diff: `LOCAL_TAG` empty path (line ~120) AND `UPSTREAM_TAG > LOCAL_TAG` path (line ~146) both call `changelog_has_version`.
- **AgDR-0008** — Y-statement present, options A/B/C/D match the ticket, rationale explains why C over D, consequences split into Kept/Added/Non-consequences, AgDR-0005 cross-referenced.
- **Test sandbox isolation** — each case calls `make_upstream`/`make_fork` (separate `mktemp -d`) and `rm -rf` at the end. No cross-case state.

Ran the suite locally on the PR HEAD: **5 PASS, 0 FAIL**.

1. `${version}` is interpolated into the grep ERE without escaping. Tags are semver so `.` is the only meta-character at risk; the `^##\s+\[...\]` anchoring makes spurious matches astronomically unlikely. If you ever cut a tag with a literal `+` or other ERE meta in the version (`v1.0.0+build.1`), the grep could over-match. Defer — apexyard semver doesn't currently include `+`. A `grep -F` with a literal-string assembly would be more robust if you want to preempt this.
2. The five `cd "$fk"` subshells in the test cases all need `|| exit 1`. Same fix as the shellcheck CI issue above — calling out so it's all in one pass.
3. AgDR-0008's `## Artifacts` section is fine but doesn't link the PR itself yet — usually safe to add `apexyard#129` once the PR exists. Optional.

### Verdict

**CHANGES REQUESTED**

The implementation, design, and tests are solid. The two CI failures (markdownlint MD032 + shellcheck SC2164 × 5) are mechanical fixes — once they're resolved and CI is green, this is good to merge.

---
📌 Reviewed commit: `b257f814b43b8c88049dae7b26f926c6f870054c`

me2resh (approved):

atlas-apex (commented):
## Re-Review: PR #129 (delta only) — APPROVED

**Commit**: `9eaef8d70eef3003197325e873df4e4432ffcf76`
**Previously requested changes on**: `b257f81` (MD032 in AgDR-0008, SC2164 in test fixture)

### Delta verified

- **AgDR-0008 (MD032/MD060)**: blank lines added before each of the three bulleted sub-lists in Consequences (`**Kept:**`, `**Added:**`, `**Non-consequences:**`). Tight-pipe table separator on line 24 padded. Markdownlint blockers resolved.
- **Test fixture (SC2164)**: `|| exit 1` appended to all five bare `cd "$fk"` lines in subshells (cases 1–5). Shellcheck blocker resolved.

No semantic change to hook logic or test assertions — pure lint compliance. Diff is exactly the surgical fix described.

### CI

All 4 checks green at `9eaef8d`:
- markdownlint-cli2: pass
- shellcheck .claude/hooks: pass
- lychee: pass
- Verify Ticket ID: pass

### Checklist Results

- Testing:                   Pass (5/5 fixture cases pass per author report)
- Technical Decisions (AgDR):Pass (AgDR-0008 linked)

### Verdict

**APPROVED** (posted as comment — cannot self-approve own PR)

---
Reviewed commit: `9eaef8d70eef3003197325e873df4e4432ffcf76`

Issue #134, opened 2026-04-26
[Chore] Configurable Jarvis-style voice prompts when waiting for user input


atlas-apex:
## Driver

When the assistant pauses for user input — design choices, per-PR merge approvals, ambiguous "which path do you want", etc. — the input request is a text message in the terminal. If the user has stepped away from the keyboard, the conversation stalls silently. The assistant is "waiting" but giving zero attentional signal.

A configurable text-to-speech layer that **speaks the question out loud** when the assistant ends a turn with a question (Jarvis-from-Iron-Man style) would surface those moments without requiring the user to babysit the terminal. The user still replies via keyboard — no voice input in the initial phase.

This is a quality-of-life improvement for adopters running long sessions (especially the multi-PR launch flows where the assistant fires 10+ "approved?" / "merge X?" / "design call: a/b/c?" prompts over an afternoon).

## Scope

**Initial phase: single platform (macOS), single TTS engine, conservative trigger heuristic.** Cross-platform and higher-quality TTS providers are explicitly out of scope here — they're follow-ups (see Future phases below) once the trigger model is proven.

### Mechanism

A new Stop hook at `.claude/hooks/voice-prompt-on-pause.sh`:

1. Fires on assistant turn end (Stop event)
2. Reads the last assistant message from the transcript JSON Claude Code passes to the hook (per the Claude Code hooks spec — the hook receives `{ transcript_path, ... }` on stdin)
3. If the message ends with a question or matches a "waiting for input" heuristic (see below), runs `say -v <voice> "<message-or-summary>"`
4. Always exits 0 — TTS failure must never block the conversation

### Trigger heuristic (initial phase, deliberately conservative)

The hook only speaks if the last assistant message:

- **Ends with `?`** (after stripping trailing whitespace + closing markdown formatting), OR
- Contains a recognised "Reply with X" / "Approved?" / "Confirm Y" / "a/b/c" / "Reply `<token>`" pattern in its last paragraph

This skips informational summary messages (which often end with "."), tool-result reports, and progress updates. Initial heuristic prefers false-negatives over false-positives — a missed prompt is annoying; a TTS reading a 200-line tool output is unbearable.

### What gets spoken

Not the full message — that would read aloud entire markdown tables, code blocks, and glossaries. The hook extracts:

- The first sentence of the trailing paragraph that triggered the heuristic, OR
- A configurable maximum (default: 200 chars), truncated at sentence boundary

[...]

PR #135, opened 2026-04-26
feat(#134): configurable voice prompts on assistant pause

atlas-apex:
Closes https://github.com/me2resh/apexyard/issues/134

## Summary

Configurable Stop hook that speaks the assistant's question aloud (Jarvis-from-Iron-Man style) when it pauses for user input. Initial phase is macOS-only via the bundled `say` command, no voice input — adopters reply via keyboard.

**Default OFF.** Pure-additive change for upstream — no existing fork sees behaviour change until they explicitly flip `voice_prompts.enabled` to `true` in their `.claude/project-config.json`.

Decision rationale in [AgDR-0009-voice-prompts-on-pause](docs/agdr/AgDR-0009-voice-prompts-on-pause.md).

## What ships

| Path | Change |
|---|---|
| `.claude/hooks/voice-prompt-on-pause.sh` | **New** — Stop hook with config gate (sub-millisecond fast-path when disabled), trigger heuristic (`questions-only` by default — last paragraph ends with `?` or matches recognised "Approved?" / "Reply with X" / "(a)/(b)/(c)" / "which path" patterns), markdown stripping (backticks, bold/italic, link syntax, bullets, table pipes), sentence-boundary truncation to `max_chars` (default 200), fire-and-forget `say` invocation |
| `.claude/hooks/tests/test_voice_prompt_on_pause.sh` | **New** — 9 cases (file follows the snake_case `tests/test_<name>.sh` convention from `test_warn_stale_review_markers.sh`). All passing locally. |
| `.claude/project-config.defaults.json` | Added `voice_prompts` block (enabled, voice, max_chars, rate_wpm, trigger). All default-OFF / safe values. |
| `.claude/settings.json` | New `Stop` hook entry wired with the standard ops-root resolver wrapper that the rest of the hooks use |
| `docs/agdr/AgDR-0009-voice-prompts-on-pause.md` | **New** — full options matrix (status quo / macOS say / cloud TTS / ML detection / notification daemon), why `say` wins for the initial phase, future-phase backlog |
| `docs/project-config.md` | New "Voice prompts" section with the schema, example overrides ("turn it on", "different voice", "always-speak debugging"), privacy notes |

Six files, +665 / -1 LOC.

## Why a Stop hook, not something more elaborate

The ApexYard interaction pattern in long sessions has lots of discrete pause points where the assistant cannot proceed without explicit input — per-PR merge approvals, design-review (a/b/c) choices, "which path do you want", tool-result confirmations. These pauses are silent text in the terminal. If the user has stepped away from the keyboard, the conversation stalls with zero attentional signal.

A Stop hook fires exactly at those pause points. The trigger heuristic only speaks when the message looks like a request for input — so tool-result reports, summary messages, and progress updates stay silent.

Real example from the session that produced this PR — the assistant said *"Reply `approve 354` (or `merge 354`) to ship — then I'll roll on to #243"*. Hook trigger fired (last paragraph contains `Reply with` + ends with em-dash but the heuristic also matches the apostrophe-paragraph). With `enabled: true`, the user would have heard *"Reply approve 354 or merge 354 to ship — then I'll roll on to two four three"* in Daniel's voice.

## Why default OFF

Upstream-friendly default. Adopters who pull this commit see no behaviour change until they opt in. The hook is a sub-millisecond fast-path no-op when disabled (single `config_get_or` call before `exit 0`).

## Why macOS-only initial phase

The user's framing ("Jarvis from Iron Man") implies a high-quality British male voice. macOS bundles `Daniel (Premium)` which is the closest free voice. ElevenLabs / OpenAI TTS would produce closer fidelity, but at recurring per-character cost AND require API-key management AND send assistant text off the local machine — that's an AgDR-worthy decision in its own right (Phase 3, separately).

Linux/Windows fall through silently when `say` isn't on PATH. Phase 2 adds OS-detection and `espeak` / `Add-Type SpeechSynthesizer` paths. Same trigger model, same config schema, just a platform-layer addition.

## Testing

- [x] `bash .claude/hooks/tests/test_voice_prompt_on_pause.sh` — **9/9 pass**
  - case 1: disabled-default → no say
  - case 2: enabled + question → say invoked with full text
  - case 3: enabled + statement → no say (questions-only heuristic)
  - case 4: enabled + `Approved?` pattern → say invoked
  - case 5: enabled + `(a)/(b)/(c)` menu → say invoked
  - case 6: malformed transcript JSON → no crash, exit 0
  - case 7: enabled but `say` not on PATH → exit 0, no crash
  - case 8: trigger=always → say invoked even on a statement
  - case 9: markdown stripping — backticks, bold, links removed; words preserved
- [x] `bash -n` syntax-check on both the hook and the test file
- [x] `jq .` validates `project-config.defaults.json` and `settings.json`
- [x] **Manual smoke** on macOS — invoked the hook with a synthetic transcript ("Quick voice test — does this work?") and confirmed Daniel's voice spoke through the speakers
- [x] **Live integration verified** — the user enabled it locally via `.claude/project-config.json` override and confirmed *"heard it"* on the in-session smoke test

## Risks

- **macOS-only.** Adopters on Linux / Windows see no benefit until Phase 2 (separate ticket). Mitigation: hook silently exits 0 when `say` isn't on PATH — cross-platform users see exactly the disabled-state behaviour. No errors, no spam.
- **Trigger heuristic false-positives.** A tool-result message that happens to end with `?` would get read aloud. Conservative regex prefers false-negatives. Adopters can disable the hook entirely or switch to `trigger: "always"` for debugging.
- **TTS sound-output collision.** If the user is on a call when the hook fires, `say` will speak through the active output device. Acceptable side-effect for v1; an env-var override could disable per-session in Phase 2.
- **Privacy.** Today the hook reads from the local transcript file and pipes to a local OS binary — nothing leaves the machine. When/if Phase 3 adds cloud TTS, that becomes a separate AgDR — explicitly NOT in scope here.

## Glossary

| Term | Definition |
|------|------------|
| **Stop hook** | A Claude Code hook that fires when the assistant ends a turn. Receives `{ session_id, transcript_path, ... }` JSON on stdin. Used here to detect pause-for-input moments and speak the question aloud. |
| **Trigger heuristic (`questions-only`)** | The default rule for "is this message asking for input?" — the last paragraph ends with `?` (after stripping trailing whitespace + markdown emphasis), OR matches one of `Approved?` / `Reply with` / `Confirm` / `(a)/(b)/(c)` / `which path` / `proceed?`. Conservative; prefers false-negatives. |
| **`Daniel (Premium)`** | macOS's bundled British-accented male voice — the closest free voice to "Jarvis-from-Iron-Man". Listable via `say -v "?"`. |
| **`VOICE_PROMPTS_SYNC=1`** | Test-mode env var that makes the hook run `say` synchronously instead of fire-and-forget. Tests need this because the orphaned-bg-process reparenting interacts badly with the test runner's subshell wrapper. Production invocations always run async. |
| **Sentence-boundary truncation** | Walks the message forward, keeping whole sentences (delimited by `. ! ?`) until the next sentence would push past `max_chars`. Avoids cutting mid-word; reads cleanly. |

Refs https://github.com/me2resh/apexyard/issues/134

### DIFF
--- a/.claude/hooks/tests/test_voice_prompt_on_pause.sh
+++ b/.claude/hooks/tests/test_voice_prompt_on_pause.sh
@@ -1,0 +1,282 @@
+#!/bin/bash
+# Smoke tests for .claude/hooks/voice-prompt-on-pause.sh
+#
+# Each case:
+#   - sets up an isolated sandbox repo under $TMPDIR with the hook + config lib
+#   - drops a synthetic transcript JSONL fixture
+#   - shims `say` on PATH with a stub that records its invocation
+#   - pipes a synthetic Stop-event JSON into the hook
+#   - asserts whether `say` was invoked, with which args
+#
+# Exit 0 means all cases passed. Exit 1 on first failure with a clear message.
+#
+# Why a stub: the real `say` would actually speak during CI, which is bad.
+# The stub records its argv to a file we can inspect.
+
+set -u
+
+HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/voice-prompt-on-pause.sh"
+if [ ! -x "$HOOK_SRC" ]; then
+  echo "FAIL: hook not found or not executable at $HOOK_SRC" >&2
+  exit 1
+fi
+
+PASS=0
+FAIL=0
+FAILED_CASES=""
+
+# ---------------------------------------------------------------------------
+# make_sandbox: build an isolated git repo with the hook + shared lib + a
+# `say` stub on PATH. Returns the sandbox path on stdout.
+# ---------------------------------------------------------------------------
+make_sandbox() {
+  local sb
+  sb=$(mktemp -d)
+  (
+    cd "$sb" || exit 1
+    git init -q
+    git config user.email "test@example.com"
+    git config user.name "test"
+    touch onboarding.yaml
+    git add onboarding.yaml
+    git commit -q -m "init"
+  )
+  mkdir -p "$sb/.claude/hooks/tests" "$sb/bin"
+  cp "$HOOK_SRC" "$sb/.claude/hooks/voice-prompt-on-pause.sh"
+  chmod +x "$sb/.claude/hooks/voice-prompt-on-pause.sh"
+
+  local src_root
+  src_root=$(cd "$(dirname "$0")/../../.." && pwd)
+  if [ -f "$src_root/.claude/hooks/_lib-read-config.sh" ]; then
+    cp "$src_root/.claude/hooks/_lib-read-config.sh" "$sb/.claude/hooks/_lib-read-config.sh"
+  fi
+  if [ -f "$src_root/.claude/project-config.defaults.json" ]; then
+    cp "$src_root/.claude/project-config.defaults.json" "$sb/.claude/project-config.defaults.json"
+  fi
+
+  # `say` stub: writes its argv to $sb/say-invocations and exits 0.
+  cat > "$sb/bin/say" <<'EOF'
+#!/bin/bash
+echo "$@" >> "${SAY_LOG:-/dev/null}"
+exit 0
+EOF
+  chmod +x "$sb/bin/say"
+
+  echo "$sb"
+}
+
+# ---------------------------------------------------------------------------
+# write_transcript: write a JSONL transcript with N assistant messages, the
+# last one having the given text. Args: sandbox dir, last-message-text.
+# ---------------------------------------------------------------------------
+write_transcript() {
+  local sb="$1"
+  local last_text="$2"
+  local path="$sb/transcript.jsonl"
+  # Two assistant messages: first is filler, last is the one we want to test.
+  jq -nc --arg t "earlier message about something else." '
+    {type: "assistant", message: {content: [{type: "text", text: $t}]}}
+  ' > "$path"
+  jq -nc --arg t "$last_text" '
+    {type: "assistant", message: {content: [{type: "text", text: $t}]}}
+  ' >> "$path"
+  echo "$path"
+}
+
+# ---------------------------------------------------------------------------
+# write_overrides: write a project-config.json with `voice_prompts.enabled` =
+# the given value (and any extra keys appended literally).
+# ---------------------------------------------------------------------------
+write_overrides() {
+  local sb="$1"
+  local enabled="$2"
+  local extra="${3:-}"
+  if [ -n "$extra" ]; then
+    cat > "$sb/.claude/project-config.json" <<EOF
+{
+  "voice_prompts": {
+    "enabled": $enabled,
+    $extra
+  }
+}
+EOF
+  else
+    cat > "$sb/.claude/project-config.json" <<EOF
+{
+  "voice_prompts": {
+    "enabled": $enabled
+  }
+}
+EOF
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# run_hook: pipes a Stop-event JSON into the hook from the sandbox cwd.
+# Captures exit code and the say-invocation log.
+# Args: sandbox dir, transcript path, [override-trigger-default].
+# ---------------------------------------------------------------------------
+run_hook() {
+  local sb="$1"
+  local transcript="$2"
+  local say_log="$sb/say-invocations"
+  : > "$say_log"
+  local payload
+  payload=$(jq -nc --arg p "$transcript" '{session_id: "s1", transcript_path: $p}')
+  (
+    cd "$sb" || exit 99
+    # Tests run the say-shim synchronously via VOICE_PROMPTS_SYNC=1 — the
+    # async-bg path is what the hook does in production, but the orphaned-bg
+    # reparenting interacts badly with the test's subshell wrapper and makes
+    # assertions flaky. The sync path is identical except for the &/disown.
+    PATH="$sb/bin:$PATH" SAY_LOG="$say_log" VOICE_PROMPTS_SYNC=1 \
+      bash "$sb/.claude/hooks/voice-prompt-on-pause.sh" <<<"$payload"
+  )
+  echo "$say_log"
+}
+
+assert_no_speak() {
+  local label="$1"
+  local sb="$2"
+  local log="$3"
+  if [ -s "$log" ]; then
+    FAIL=$((FAIL+1))
+    FAILED_CASES="${FAILED_CASES}\n  $label: expected NO say invocation, got: $(cat "$log")"
+    rm -rf "$sb"
+    return 1
+  fi
+  PASS=$((PASS+1))
+  rm -rf "$sb"
+}
+
+assert_speak_contains() {
+  local label="$1"
+  local sb="$2"
+  local log="$3"
+  local needle="$4"
+  if [ ! -s "$log" ]; then
+    FAIL=$((FAIL+1))
+    FAILED_CASES="${FAILED_CASES}\n  $label: expected say invocation containing '$needle', got nothing"
+    rm -rf "$sb"
+    return 1
+  fi
+  if ! grep -q -F -- "$needle" "$log"; then
+    FAIL=$((FAIL+1))
+    FAILED_CASES="${FAILED_CASES}\n  $label: expected '$needle', got: $(cat "$log")"
+    rm -rf "$sb"
+    return 1
+  fi
+  PASS=$((PASS+1))
+  rm -rf "$sb"
+}
+
+# ===========================================================================
+# Cases
+# ===========================================================================
+
+# Case 1: disabled (default) → no say
+sb=$(make_sandbox)
+# No project-config.json override → defaults apply (enabled=false)
+transcript=$(write_transcript "$sb" "Reply with approve to proceed?")
+log=$(run_hook "$sb" "$transcript")
+assert_no_speak "case1: disabled-default" "$sb" "$log" || true
+
+# Case 2: enabled + question → say invoked
+sb=$(make_sandbox)
+write_overrides "$sb" "true"
+transcript=$(write_transcript "$sb" "Ready to merge PR #42?")
+log=$(run_hook "$sb" "$transcript")
+assert_speak_contains "case2: enabled+question" "$sb" "$log" "Ready to merge PR" || true
+
+# Case 3: enabled + statement → no say (questions-only trigger heuristic)
+sb=$(make_sandbox)
+write_overrides "$sb" "true"
+transcript=$(write_transcript "$sb" "I finished the work and committed it.")
+log=$(run_hook "$sb" "$transcript")
+assert_no_speak "case3: enabled+statement" "$sb" "$log" || true
+
+# Case 4: enabled + "Approved?" pattern → say invoked
+sb=$(make_sandbox)
+write_overrides "$sb" "true"
+transcript=$(write_transcript "$sb" "All checks green. **Approved?**")
+log=$(run_hook "$sb" "$transcript")
+assert_speak_contains "case4: enabled+approved-pattern" "$sb" "$log" "Approved" || true
+
+# Case 5: enabled + (a)/(b)/(c) menu → say invoked
+sb=$(make_sandbox)
+write_overrides "$sb" "true"
+transcript=$(write_transcript "$sb" "Pick a path: (a) ship it, (b) refactor, (c) revert.")
+log=$(run_hook "$sb" "$transcript")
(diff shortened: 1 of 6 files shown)


atlas-apex (commented):
## Code Review: PR #135

**Commit**: `d92ab402507fc4eda2194750a146490ffd3d13b0`

### Summary

Configurable Stop hook that speaks the assistant's last paragraph aloud via macOS `say` when a turn looks like a request for input. Pure-additive: ships `enabled: false` by default, sub-millisecond-ish fast-path no-op when disabled. Six files, +665 / -1 LOC, single commit, all four CI checks green (shellcheck, markdownlint, lychee, PR-title), 9/9 local tests pass.

### Checklist Results

- Testing:                   Pass (with two minor coverage gaps noted)
- PR Description & Glossary: Pass (Glossary has 5 entries, all decision-relevant)
- Technical Decisions (AgDR): Pass — AgDR-0009 covers TTS provider, hook event type, trigger heuristic, default-OFF gating, and sync env-var consequence

### Verification I ran

- Re-ran the test suite locally: 9/9 passed
- Probed markdown stripping on 6 nested/edge-case inputs (`**[bold link](url)**`, `` `code with [brackets]` ``, `` `arr[0]` ``, italic+link, unmatched `*`, simple italic) — all unwrap correctly
- Probed the trigger heuristic on 8 cases including the multi-paragraph "Q in earlier para, statement at end" case the question called out — last-paragraph isolation works as designed
- Traced the fast-path through `_lib-read-config.sh` end-to-end
- Confirmed `.claude/project-config.json` (the local override with `enabled: true`) is gitignored and NOT in this PR's diff
- Verified `.claude/project-config.defaults.json` ships `enabled: false`

### Detailed responses to the review prompts

**1. Trigger regex (step 8).** The end-of-last-paragraph `?` check is correct — `awk 'BEGIN{RS=""} END{print}'` correctly isolates the last paragraph, so a `?` in an earlier paragraph followed by a closing statement does not trigger (verified in probe). Trailing `*_\``  emphasis chars are stripped before the last-char check, which handles `**Approved?**` cleanly. The `(a)` / `(b)` / `(c)` menu pattern matches case-insensitively.

   The one false-positive class is a legitimate concern: a tool-result paragraph ending with a literal `?` (e.g. `Did 5 things successfully?`) does trigger. In practice assistants don't phrase tool-success reports as questions, so this is rare. The AgDR's "questions-only" framing acknowledges the conservative-bias trade-off explicitly. Acceptable for v1.

**2. Async vs sync mode.** `[ "${VOICE_PROMPTS_SYNC:-}" = "1" ]` only matches the literal string `1` — narrow scope. The variable is `VOICE_PROMPTS_*`-namespaced (no collision risk with generic env). A production adopter would have to deliberately export it in their shell profile. The minor wart is that the env var is documented in the user-facing `docs/project-config.md` "Test mode" section — this advertises the bypass to adopters who don't need it. Non-blocking, but worth keeping it confined to the test file's comments + AgDR in a follow-up if you want to harden the public surface.

**3. Markdown stripping.** Nested constructs work (verified):

- `**[bold link](url)**` → `bold link` (link unwraps first, then `**...**`  pair matches)
- `` `code with [brackets]` `` → `code with [brackets]` (backticks first, brackets preserved as-is)
- `` `arr[0]` `` → `arr[0]`
- `*[link](url)*` → `link`

   Order of sed expressions is correct: backticks → bold → italic → underscore-italic → links. Edge cases to be aware of in future iterations: a link whose label contains backticks (`` [`code`](url) ``) — current order strips the backticks first, leaving `[code](url)`, then the link regex strips to `code`. Works. URLs with parentheses inside (rare) would break the `[..](..)` regex, but that's a known limitation of regex-based markdown stripping.

**4. Fast-path.** Trace: `cat` stdin → `git rev-parse --show-toplevel` (one fork+exec) → source `_lib-read-config.sh` (defines functions, no I/O) → `config_get_or '.voice_prompts.enabled' 'false'` which triggers `_config_load`: jq + 2 file reads merged once, cached for the process → string compare → `exit 0`.

   That is **not sub-millisecond** — `git rev-parse` + `jq` invocation typically lands at 30–50ms on macOS. The hook header and AgDR-0009 ("Hook overhead on every turn-end is sub-millisecond in the disabled state") slightly overstate the speed. Not a blocker — the hook is still cheap enough to fire on every Stop event, and the actual perceptual latency is invisible. Worth tightening the wording in a follow-up to "tens of milliseconds" or similar.

**5. Test coverage.** Case 9 verifies (a) no backticks remain (b) no `**` remain (c) `approve 42` and `merge` survive. It does NOT verify single `*`, single `_`, or that the link URL is gone — a regression that emitted `[the PR](https://...)` raw text would slip through because `merge` and `approve 42` would still match. Suggest tightening:

   ```bash
   ! grep -q '`' "$log" && ! grep -q '\*' "$log" && ! grep -q '_[a-z]' "$log" && \
     ! grep -qE '\(https?://' "$log" && \
     grep -q "approve 42" "$log" && grep -q "merge" "$log"
   ```

   No multi-paragraph trigger case in the suite. The `write_transcript` helper writes 2 separate assistant *messages* (each with one paragraph), but the hook's last-paragraph awk extraction operates inside a single message's text. A regression that ignored the awk and tested the whole message body would not be caught. Suggest adding a case 10:

   ```
   "Should we use option A? It has these features.\n\nI went ahead and chose option A."
   ```

   Expected: no say (last paragraph is a statement). Would catch the multi-paragraph regression class.

   Both gaps are non-blocking. Adding them to a follow-up ticket is fine.

**6. AgDR-0009 quality.** Option matrix is honest (5 entries: status quo, chosen, OpenAI/ElevenLabs, ML detection, notification daemon). Cost/quality/privacy tradeoffs explicit. macOS-only initial-phase justification stands up: cost-per-character + API-key management + privacy implications of cloud TTS are AgDR-worthy in their own right and explicitly deferred to Phase 3. Future-phases section is clear about scope. Reversibility section is concrete (one config flip → no-op). Only minor issue is the "sub-millisecond" overclaim noted in #4 above.

**7. Default-OFF assertion.** Verified directly:

- `.claude/project-config.defaults.json` (in PR diff) → `"voice_prompts": { "enabled": false, ... }` ✓
- `.claude/project-config.json` (local override, `enabled: true`) → not in PR diff ✓
- Local override is gitignored via `.gitignore` line `.claude/project-config.json` ✓
- Test case 1 verifies the disabled-default path produces no say invocation ✓

**8. Cross-repo commit-message reference.** The `verify-commit-refs.sh` hook resolves the tracker repo from `origin` (or `tracker_repo` in project config), not from the PR base. When you're on the apexstack ops fork and write `Closes #134` referring to upstream apexyard, the hook checks origin (the fork) and rejects because `#134` doesn't exist there. The hook's error message itself recommends "write it as a plain URL" — which is what was done. So this PR's `Refs https://github.com/me2resh/apexyard/issues/134` is correct workaround behaviour.

   It IS a hook-config gap: a fork that contributes back to upstream has a different tracker than a fork managing its own projects, and the hook doesn't know how to detect that. Reasonable follow-ups: (a) add a `tracker_repo` setting in `.claude/project-config.json` for the apexstack ops fork pointing at `me2resh/apexyard`, or (b) teach the hook to walk to `git remote get-url upstream` if origin doesn't have the issue. (a) is one config line; (b) is more robust. Worth filing as a separate upstream issue. Not a blocker for this PR.

### Issues Found

None blocking.

### Suggestions (non-blocking, not for this PR)

- Tighten AgDR-0009 + hook header to drop the "sub-millisecond" claim — fast-path is ~30–50ms due to `git rev-parse` + `jq` invocation through `_lib-read-config.sh`. Still cheap, just not microsecond-class.
- Tighten case-9 negative assertions to also reject single `*` / `_` / `(http`.
- Add a multi-paragraph trigger test case (last paragraph is a statement, earlier paragraph contains `?`).
- Consider keeping `VOICE_PROMPTS_SYNC=1` documented in the test file + AgDR only, not in the user-facing `docs/project-config.md`.
- File a follow-up issue for the cross-repo `verify-commit-refs.sh` gap — fork-aware tracker resolution.

### Verdict

**APPROVED**

Single-commit, well-scoped, default-OFF, well-tested, well-documented. The trigger heuristic, markdown stripping, and fast-path all behave correctly in direct probes. The AgDR is honest about trade-offs and explicit about deferred phases. CI is green. The handful of suggestions above are polish, not blockers.

---
Reviewed commit: `d92ab402507fc4eda2194750a146490ffd3d13b0`

atlas-apex (commented):
## Code Review: PR #135

**Commit**: `d92ab402507fc4eda2194750a146490ffd3d13b0`

### Summary

Configurable Stop hook that speaks the assistant's question aloud via macOS `say` when the conversation pauses for input. Default OFF (pure-additive for upstream). Six files, +665/-1 LOC. Initial phase is macOS-only by explicit AgDR-0009 scope; Linux/Windows are deferred to Phase 2 with the same trigger model + config schema.

### Checklist Results

- Architecture & Design:     **Pass** — Stop hook fires at the right SDLC seam (turn-end pause points). Trigger model is the load-bearing piece, decoupled from the TTS provider so Phase 2/3 can swap providers without re-architecting. Config-gated, fully reversible.
- Code Quality:              **Pass** — `set -u`, always exits 0 (correct for a non-blocking pause hook), fast-path single-call no-op when disabled. JSONL parsing handles both new (array-of-blocks) and older (string-content) message formats. Markdown stripping covers backticks/bold/italic/links/headers/bullets/pipes. Sentence-boundary truncation has hard-cut fallback at word boundary.
- Testing:                   **Pass** — 9 cases cover disabled-default, enabled+question, enabled+statement (no-speak), "Approved?" pattern, (a)/(b)/(c) menu, malformed JSON, missing `say` on PATH, trigger=always, markdown stripping. Sandbox isolation per case with shimmed `say` recording argv. `VOICE_PROMPTS_SYNC=1` documented as test-only. ShellCheck workflow green.
- Security:                  **Pass** — No secrets, no PII leaving the machine (local transcript file → local OS binary). AgDR explicitly defers cloud-TTS privacy decision to Phase 3. No injection vector (text is passed as a single argv to `say`, not shell-interpolated).
- Performance:               **Pass** — Sub-millisecond fast-path when disabled (single `config_get_or` before exit 0). Fire-and-forget async `say` invocation with `disown 2>/dev/null || true` defensive fallback. Hook never blocks the conversation.
- PR Description & Glossary: **Pass** — Five-entry glossary (Stop hook, trigger heuristic, Daniel voice, `VOICE_PROMPTS_SYNC`, sentence-boundary truncation). Testing instructions enumerate all 9 cases plus syntax checks plus manual smoke plus live integration. Risks section is honest (macOS-only, trigger false-positives, sound-output collision, future cloud-TTS privacy).
- Technical Decisions (AgDR):**Pass** — AgDR-0009 (85 LOC) covers all five options considered (status quo / macOS say / cloud TTS / ML detection / notification daemon), explicit decision rationale (3 reasons), positive + negative consequences, reversibility, future-phase backlog (Phase 2 cross-platform, Phase 3 cloud TTS, Phase 4 voice input, Phase 5 per-message overrides). Linked from the PR body.

### Issues Found

None blocking.

1. **Missing-jq feedback (line 379).** Hook silently exits 0 when `jq` isn't on PATH. On a macOS where `say` is available but `jq` happens to be missing, the user will have enabled the hook and hear nothing — with no diagnostic. A one-line stderr warning (`echo "voice-prompt-on-pause: jq not found, skipping" >&2`) would aid debugging. Not blocking — apexyard already requires `jq` broadly.

2. **No explicit `max_chars` truncation test.** The 9 cases verify trigger heuristics + markdown stripping, but the sentence-boundary truncation at `max_chars` (the most algorithmically interesting part of the hook) isn't directly exercised. The logic is straightforward and the contract is clear, so this is a "nice to have", not a gap that should block merge.

3. **`(a)|(b)|(c)` pattern (line 441).** The regex is matched literally and could in theory catch substrings like `go(a)l`. In conversational context the false-positive is negligible, and the user can always switch to `trigger: "always"` for debugging or disable. Calling it out for completeness, not as a blocker.

### Verdict

**APPROVED** (posted as a comment because GitHub blocks self-approval — PR author is the same identity as Rex's runtime owner)

Settings.json wiring matches the established ops-root resolver pattern exactly. CI fully green for 5 days (lychee, markdownlint, PR-title, shellcheck on `.claude/hooks`). Default-OFF makes this a pure-additive change for upstream — no adopter sees behaviour change until they explicitly opt in. AgDR-0009 is one of the more thorough decision records in the repo. Shipping.

---
Reviewed commit: `d92ab402507fc4eda2194750a146490ffd3d13b0`

Issue #153, opened 2026-05-03
[Chore] Extend Bash-write matcher in _lib-detect-bash-write.sh beyond first-version coverage


atlas-apex:
## Driver

Rex's code review of PR #152 (which shipped me2resh/apexyard#150 + #151) flagged 16+ Bash-write shapes the heuristic detector currently misses. Per AgDR-0011, the matcher's design is "false-negatives preferred over false-positives" so the misses are not bugs — but the matcher table is explicitly framed as a living list extended on observation. This ticket captures the observed-but-uncovered shapes from Rex's probe and the adjacent classes that would be obvious bypass routes for an agent researching the matcher.

## Scope

Extend `.claude/hooks/_lib-detect-bash-write.sh` matcher patterns to cover:

- File-moving builtins: `cp`, `mv`, `rm`, `dd`, `install`
- Archive / network writes: `tar -x` / `tar --extract`, `curl -o` / `--output`, `wget -O` / `--output-document`
- Other interpreters: `perl -e`, `php -r`, `go run`, `deno`, `bun`
- Python helpers we missed: `pathlib.Path().touch()`, `shutil.copy*`, `os.rename`, `shutil.move`
- Heredoc variants for `ruby` and `node` (currently only python heredoc is matched)

Each new pattern needs a corresponding test case in `.claude/hooks/tests/test_detect_bash_write.sh` (positive class) and ideally one read-only counterexample in the negative class to guard against false positives on the new pattern (e.g. `tar -t` listing should not trigger the `tar -x` matcher).

## Acceptance Criteria

- [ ] Each pattern in the Scope list has a regex match in `_lib-detect-bash-write.sh` and at least one positive-class test case
- [ ] At least three negative-class counterexamples added (one per ambiguous pattern: e.g. `cp` is gated, `git cp` shouldn't trigger if/when git ever adds such a subcommand)
- [ ] Total test count in `test_detect_bash_write.sh` rises from 32 to ≥48
- [ ] All existing 32 cases still pass (no regression on first-version patterns)
- [ ] Pattern table comment in `_lib-detect-bash-write.sh` is updated to reflect the new shapes

## Risks / Dependencies

- **False-positive risk on `cp` / `mv` / `rm`** — these are extremely common in legitimate read-adjacent commands (e.g. build scripts that move artefacts). The matcher must be scoped to bare invocations only, not subcommand-style usage. Mitigation: extensive negative-class test coverage.
[...]

PR #155, opened 2026-05-03
chore(#153): extend Bash-write matcher beyond first-version coverage

atlas-apex:
## Summary

Extends `_lib-detect-bash-write.sh` (introduced in #152 for #151) with the matcher families that Rex's review flagged as missing from the first-version coverage. AgDR-0011 already frames the matcher table as a living list extended on observation; this PR walks the list.

- **File-moving builtins** — `cp`, `mv`, `rm`, `dd`, `install`. Anchored at command-start. `--help`/`--version` and `git rm`/`git mv` are excluded so `git rm src.txt` and `cp --help` stay read-class.
- **Archive / network writes** — `tar -x` / `tar --extract`, `curl -o` / `--output`, `wget -O` / `--output-document`. Flag-gated so `tar -t` (list), `curl <url>` (bare fetch), `wget --help` stay read-class.
- **Additional interpreters** — `perl -e` and `php -r` keyword-gated like the existing python/node/ruby matchers; `go run`, `deno run`/`deno script.ts`, `bun run`/`bun script.ts` matched categorically as script runners.
- **Python helpers** — `pathlib.Path().touch()`, `shutil.copy*`, `shutil.move`, `os.rename` added to the `python -c` and python-heredoc keyword list.
- **Heredoc variants for `ruby` and `node`** — previously only python's heredoc form was matched.
- **Extractor extensions** — `cp`/`mv` return the last positional arg; `curl -o`/`--output` and `wget -O`/`--output-document` return the file argument. `go run`/`deno`/`bun`/`perl -e`/`php -r`/`tar -x` return empty (caller gates categorically per AgDR-0011).

Refactored the matcher into one `_bdw_match_<family>` helper per family so future extensions land as additional helpers rather than additional `if` branches in one big function.

## Testing

- [x] `bash .claude/hooks/tests/test_detect_bash_write.sh` — **86/0** (was 32/0 before #153). Includes ≥3 negative-class counterexamples for every new family: `tar -t`, `tar --list`, `tar -tzf`, `cp --help`, `cp --version`, `rm --help`, `mv --version`, `git rm`, `git mv`, `curl -s url`, `curl bare`, `wget --help`, `wget bare`, `deno fmt`, `deno test`, `go build`, `go version`, `perl -v`, `php --version`.
- [x] `bash .claude/hooks/tests/test_require_active_ticket_bash.sh` — **12/12** (regression check, unchanged).
- [x] No edits outside `.claude/hooks/_lib-detect-bash-write.sh` and `.claude/hooks/tests/test_detect_bash_write.sh`.

## Glossary

| Term | Definition |
|------|------------|
| Matcher family | One self-contained shape the detector looks for (redirection, tee, sed -i, file-mover, archive-extractor, …). Each is a separate `_bdw_match_*` helper. |
| Categorical match | A matcher that fires whenever the command shape is present, without inspecting payload contents. Used for `go run`, `deno`, `bun` where the script body could write but isn't visible to the detector. |
| Keyword-gated match | A matcher that requires both the interpreter shape (`python -c`, `node -e`, `perl -e`, …) AND a write-related keyword (`writeFile`, `open(…, "w")`, `unlink`, …) before firing. Reduces false positives on read-only one-liners. |
| Living list | AgDR-0011's framing of the matcher table as a list of patterns extended whenever a new bypass shape is observed in the wild, rather than a fixed-once-and-done specification. |
| False-negative preference | The tightness target the detector aims for: missing one obscure write pattern is acceptable; blocking a legitimate read-only command on a fresh-adopter test is not. |
| Bootstrap exemption | The `.claude/session/active-bootstrap` marker that lets `/setup`, `/handover`, `/update`, `/split-portfolio` write files without an active ticket. Orthogonal to this PR but the reason the matcher can be stricter without breaking first-run UX. |

### DIFF
--- a/.claude/hooks/_lib-detect-bash-write.sh
+++ b/.claude/hooks/_lib-detect-bash-write.sh
@@ -10,6 +10,11 @@
 # cases (~95%) and treat the long tail as a known-limitation that
 # extends as new patterns are discovered.
 #
+# AgDR-0011 frames the matcher table as a LIVING LIST — extended on
+# observation. me2resh/apexyard#153 extended the first-version coverage
+# (#152) with file-moving builtins, archive/network writes, additional
+# interpreters, and python-helper / heredoc shapes.
+#
 # Usage:
 #   source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-detect-bash-write.sh"
 #   if bash_command_appears_to_write "$COMMAND"; then
@@ -24,27 +29,48 @@
 #   bash_extract_write_target COMMAND
 #       echoes the target path if extractable, empty string otherwise.
 #       Best-effort — only handles the simple cases (echo > file,
-#       tee file, sed -i ... file). Embedded interpreters
-#       (python -c, node -e, ruby -e) return empty.
+#       tee file, sed -i ... file, cp src dst, curl -o file). Embedded
+#       interpreters (python -c, node -e, ruby -e, perl -e, php -r,
+#       go run, deno, bun) return empty.
 
 # ------------------------------------------------------------------------------
 # Public: bash_command_appears_to_write COMMAND
 #
-# Detects:
-#   - Output redirection: cmd > file, cmd >> file, cmd 2> file
-#   - tee
-#   - cat > file (with or without heredoc)
-#   - printf > file
-#   - sed -i (in-place edit)
-#   - awk -i inplace
-#   - python -c '...write...' or python -c '...open(...).write...'
-#   - python <<EOF ... write ... EOF (heredoc-fed python)
-#   - python3 - <<EOF ... write ... EOF (stdin-fed python — common pattern)
-#   - node -e '...writeFileSync...' or node -e '...write...'
-#   - ruby -e '...File.write...' or ruby -e '...write...'
+# Detects (matcher families):
+#
+#   Redirection / pipes-to-disk
+#     - cmd > file, cmd >> file, cmd 2> file
+#     - tee
+#
+#   In-place text editors
+#     - sed -i (in-place edit, GNU + BSD `''` form)
+#     - awk -i inplace
+#
+#   File-moving builtins (#153)
+#     - cp, mv, rm, dd, install — anchored at command start, --help/--version
+#       excluded, `git rm`/`git mv` excluded (those are subcommands)
+#
+#   Archive / network writes (#153)
+#     - tar -x, tar --extract
+#     - curl -o / --output
+#     - wget -O / --output-document
+#
+#   Embedded interpreters with inline source (-c / -e / -r)
+#     - python -c '…' with write/open/touch/copy/rename keywords
+#     - python <<EOF / python - <<EOF (heredoc-fed)
+#     - node -e '…' with writeFile/appendFile/write keywords
+#     - node <<EOF (heredoc-fed, #153)
+#     - ruby -e '…' with File.write/open keywords
+#     - ruby <<EOF (heredoc-fed, #153)
+#     - perl -e '…' with print-to-handle / open / unlink keywords (#153)
+#     - php -r '…' with file_put_contents / fwrite / fopen keywords (#153)
+#
+#   Script runners — categorical (#153)
+#     - go run <file>
+#     - deno run / deno <script.{ts,js,mjs}>
+#     - bun / bun run <script.{ts,js,mjs}>
 #
 # Misses (intentionally — long tail):
-#   - cp, mv, rm (those move/remove, not "write content")
 #   - xargs that constructs a write command
 #   - find -exec sed/awk/etc.
 #   - Custom scripts that wrap writes (could be anything)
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #155

**Commit**: `7f9a2cbeae8da961b7852a79e2d1d01c1b17906a`

### Summary
Extends `_lib-detect-bash-write.sh` matcher beyond first-version coverage (#152) per AgDR-0011's living-list framing. Adds nine new matcher families: file-moving builtins (`cp`/`mv`/`rm`/`dd`/`install`), archive/network writes (`tar -x`, `curl -o`, `wget -O`), additional interpreters (`perl -e`, `php -r`, `go run`, `deno`, `bun`), python-helper keywords (`pathlib.touch`, `shutil.copy*`, `os.rename`), and ruby/node heredocs. Also refactors the matcher into per-family `_bdw_match_*` helpers and extends the extractor for `cp`/`mv` (last positional) plus `curl -o` / `wget -O`. Test count rises 32 → 86, including ≥3 negative-class counterexamples per new family.

### Checklist Results
- Architecture & Design:    Pass — clean refactor into per-family helpers, design-faithful to AgDR-0011
- Code Quality:              Pass — bash-grep idioms consistent with the existing file
- Testing:                   Pass — 86/0 on a fresh tree (`bash .claude/hooks/tests/test_detect_bash_write.sh`); negative-class coverage solid
- Security:                  Pass — internal helper, no external surface; tightening detection reduces bypass risk
- Performance:               N/A — pre-tool hook path, dominated by spawn cost; per-family helpers are O(matcher count) but each is a single grep
- PR Description & Glossary: Pass on Glossary + Testing; minor gap (see Issues)
- Technical Decisions (AgDR): Pass — extends AgDR-0011's living list, no new decisions made; AgDR-0011 referenced in code header (lines 13–15) and PR body

### False-positive validation (the user's #1 concern)
Ran 31 additional probes outside the test file targeting the high-risk surfaces: `tar -tvf`, `tar tvf`, `tar tf`, `tar --list -tvf`, `curl -sL`, `curl -I`, `curl --silent`, `wget -q ... -P`, `wget --spider`, `perl -V`, `perl -E "say"`, `php -v`, `deno --version`, `deno info`, `bun install`, `bun --version`, `go test`, `go mod tidy`, `git rm -r foo/`, `sudo git rm`, `cp -r src/ dst/`, `cp "quoted file" "dst"`, `perl -ne "print" file`, `php -r "echo file_get_contents(...)"`, `tar -cf out.tar src/`, `tar -czf out.tar.gz`, `dd if=/dev/sda | head`, `echo "rm me"`, plus compound shapes like `ls && go run main.go` and `git rm a && rm b`. **All 31 classified correctly.**

### Extraction edge cases
- `cp -r src/ dst/` → `dst/` ✓
- `cp a b /tmp/dst/` (3 args) → `/tmp/dst/` ✓
- `mv -v old.txt new.txt` → `new.txt` ✓
- `curl --output=/tmp/x ...` (= form) → empty (acceptable miss; detection still fires)
- `wget --output-document=/tmp/x` (= form) → empty (same; detection still fires)

### Categorical script-runner match — design sanity
Reviewed the call to fire categorically on `go run` / `deno run` / `deno <file>.ts` / `bun run` / `bun <file>.ts`. The runner's script body is invisible to the matcher; the read-only counterexamples (`deno fmt`, `deno test`, `bun install`, `go test`, `go build`) are correctly excluded by the regex (require `run` keyword OR `.{ts,js,mjs,tsx,jsx}` extension). Asking for a write-shaped flag like `--allow-write` would miss deno scripts run with `-A` and bun (no permission model). **Conservative categorical match is the right tradeoff.**

### Issues Found
1. **No `Closes #153` / `Fixes #153` keyword in the PR body.** Title carries `(#153)` so the PR links to the issue, but `closingIssuesReferences` is empty — the issue won't auto-close on merge. Minor gap; resolvable by adding `Closes #153` to the body before merge or closing the issue manually after merge.

1. **Long-tail miss: `tar -c` / `tar -cf` (archive create).** Not in scope for this PR (#153 only specified `tar -x`), but it's a write surface that's the natural sibling to `-x`. Consider a follow-up ticket extending the `_bdw_match_tar_extract` helper to also fire on `-c`. Probed `tar -cf out.tar src/` and `tar -czf out.tar.gz src/` — both currently classified as READ.
2. **`curl --output=PATH` and `wget --output-document=PATH` (= form) extract empty.** Detection still fires correctly. Low priority — most agents and humans use space-separated form. If extraction matters for these, the regex `(--output|-o)[[:space:]=]+...` would handle both. Could be folded into a future extraction-quality pass.
3. **Helper file_movers nested grep logic** (`_lib-detect-bash-write.sh:114-126`) is convoluted — the inner `grep -qE ... | grep -vE ...` chain is dead code (the outer `grep -qE` already returned 0 so its stdout is consumed by `-q`). The fast-path check on `^[[:space:]]*git[[:space:]]+(rm|mv)\b` does the actual work. Tests pass and behavior is correct, but the dead branch is confusing to readers. Worth a minor cleanup pass.

### Verdict
**APPROVED**

Tests pass cleanly on a fresh tree (86/0). False-positive surface validated against 31 additional probes covering the high-risk read-adjacent shapes. Refactor into per-family helpers is a quality improvement for future extensions. The closing-keyword gap and the suggestions above are not blocking.

---
📌 Reviewed commit: `7f9a2cbeae8da961b7852a79e2d1d01c1b17906a`

Issue #168, opened 2026-05-04
[Chore] /release skill's branch + PR-title patterns are rejected by the framework's own validators


atlas-apex:
## Driver

The `/release` skill's prescribed branch name (`release/vA.B.C`) and PR title (`release: vA.B.C`) are both rejected by the framework's own validators:

- **`validate-branch-name.sh`** requires `{type}/{TICKET-ID}-{description}`. `release/v1.2.0` has no ticket-ID component after the slash, so the regex `^(${TYPES})/([A-Z]{2,10}-[0-9]+|GH-[0-9]+|#[0-9]+)-` rejects it even if `release` is added to the type whitelist.
- **`validate-pr-create.sh`** requires `type(SCOPE):` form via `^(${PR_TYPES})\(([A-Z]{2,10}-[0-9]+|#[0-9]+)\)!?:`. `release: v1.2.0` has no scope and `release` isn't in the PR-title type whitelist.

This contradiction surfaced trying to cut v1.2.0 — the first release under the dev/main release-cut model (#116, AgDR-0007). The skill prescribes one shape; the validators reject it.

## Scope

Two small changes plus a doc note:

### A. `validate-branch-name.sh` — special-case the release pattern

Add an early-out branch in the validator: any branch matching `^release/v\d+\.\d+\.\d+$` passes regardless of the standard type/ticket-id/description regex. The release-cut convention (AgDR-0007) is an explicit, narrow shape; recognising it as a permitted exception is cleaner than trying to bend the standard regex.

### B. `project-config.defaults.json` — add `release` to `.pr.title_type_whitelist`

The PR title type whitelist (used by `validate-pr-create.sh`) is currently `feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert`. Adding `release` lets a release PR title `release(#160): v1.2.0` pass the existing regex unchanged.

### C. Update `.claude/skills/release/SKILL.md` — PR title shape

Step 4 currently prescribes title `release: vA.B.C`. Change to `release(#<release-ticket>): vA.B.C` (e.g. `release(#160): v1.2.0`). The release-cut ticket is the natural scope; same shape every other PR title uses.

## Acceptance Criteria

- [ ] `validate-branch-name.sh` accepts `release/v1.2.0` as a valid branch name
- [ ] `validate-branch-name.sh` still rejects malformed branches (e.g. `release/foo`, `release/v1`, etc.)
[...]

Issue #160, opened 2026-05-03
[Chore] Release v1.2.0 + landing-site multi-tab terminal refresh


atlas-apex:
## Driver

v1.1.0 was tagged 2026-04-19. Since then, dev has accumulated 23 commits across 2 major workflow changes (#150 + #151 bootstrap + Bash-gating, #132 + #48 approve-merge hardening), 4 new skills (`/debug`, `/validate-idea`, `/tickets-batch`, `/fan-out`), portfolio-mode polish (#145), 8+ new hooks, and the release-cut branch model (#116).

Time to cut v1.2.0 (or v2.0.0 — see the version-bump note in the test ticket) and refresh the landing site to showcase the new work.

### Part A — release tag

1. Wait for the pre-release smoke test ticket (filed alongside this one) to be closed.
2. Run `/release` to cut the dev → main release PR.
   - Pick semver bump (suggested: v1.2.0 minor; v2.0.0 if the auto-merge + Bash-gating behaviour changes are deemed breaking)
   - Generate the CHANGELOG draft from conventional-commits headers
   - **Do NOT mention the voice-prompts feature in the published changelog** — it never shipped to main, so adopters never saw it. The supersession AgDR-0013 stays in the repo as the framework's internal record.
3. Open the release PR, get Rex + CEO approval (using the new `/approve-merge` flow), tag.

### Part B — landing site refresh

The landing site at `site/index.html` has a single "claude code — apexyard — one ticket, start to finish" terminal demo. With v1.2.0's expanded skill surface, one flow doesn't represent the full story.

**Add tabs to the terminal demo.** Each tab is one flow. The terminal autoplays the active tab's commands; when it reaches the last command in the active tab, it advances to the next tab and starts replaying that one. Cycle continues; user can also click a tab to jump.

Suggested tab roster (4–6 flows):

1. **One ticket, start to finish** — the existing flow (pick up #58, branch, plan, code, PR, Rex, CEO, merge). This stays as the headline tab.
2. **Capture an idea → /validate-idea → /feature → /write-spec** — the product-discovery onramp.
3. **/handover legacy-billing-api** — adopting an external repo into the portfolio.
4. **/setup on a fresh fork** — first-run framework bootstrap (the #150 happy path).
5. **/fan-out — 3 independent tickets in parallel** — showcase parallel agent work.
6. (Optional) **/migration with rollback plan** — the migration AgDR + ticket flow.

[...]

Issue #116, opened 2026-04-24
[Chore] Adopt release-cut branch model (dev/main + tags) for the framework repo — not for managed projects


atlas-apex:
## Driver

External adopters are now using apexyard. Every commit on `upstream/main` is (potentially) pulled into their fork when they run `/update`. Under the current trunk-based model, `main` receives daily work-in-progress — half-finished skills, in-flight hook experiments, docs refactors mid-iteration. This was fine when the framework had one user; it's not fine now.

Fork users deserve `upstream/main` to mean *"the state a reasonable adopter should live on."* That's a release-branch promise, not a trunk promise.

## Proposed model: "release-cut" (gitflow-lite)

Two long-lived branches, no release or hotfix branches:

```
dev   ────●──●──●──●──●──●──────●──●──────●──●──────  (daily work)
           \                    /          /
            \                  /          /
main  ──────●────────────────●──────────●──────────────  (released only, tagged)
            v1.1.0            v1.2.0    v1.2.1
```

- `dev` — every feature/fix/chore PR targets here. All existing PR hooks + review gates apply unchanged.
- `main` — only receives merges from `dev`, via release PRs. Every merge to `main` is tagged with a semver. Direct pushes blocked (already enforced by `block-main-push.sh`).
- Release = `dev` → `main` merge + `git tag vX.Y.Z` + `git push --tags`.

**Explicitly NOT adopting from full git flow:**

- No `release/*` branches (no stabilisation window needed for a docs+hooks framework).
- No `hotfix/*` branches (no multi-version support; forks always run latest).
- No `support/*` branches.

See the rationale in the AgDR draft at the end of this ticket.

### Framework only — NOT managed projects

This is the critical scoping call. apexyard-the-framework has downstream consumers (forks); it needs the release-branch promise. Managed projects under governance (per the registry) have no downstream consumers — they stay trunk-based with `main` as working. Contributors must not cargo-cult the dev/main pattern into project templates.

### Changes required

**Repo policy:**

1. Create `dev` branch from current `main` HEAD.
2. Set `dev` as the default branch for PRs (GitHub repo setting + `.github/PULL_REQUEST_TEMPLATE.md` target).
3. Branch protection on `main`:
   - Only merges from `dev` accepted (GitHub "restrict who can push" or a merge-gate check).
   - Required status checks: all existing + a "tag present" check on the merge commit.
4. Branch protection on `dev`:
   - Same as today's `main` protection (required checks, review markers, CEO approval, design review).

[...]

PR #1, opened 2026-04-05
feat(#1): initial ApexStack project

atlas-apex:
## Summary

- Initial release of ApexStack — an open-source, AI-native software development stack for Claude Code
- 19 software development role definitions across 5 departments (Engineering, Product, Design, Security, Data)
- 3 workflow processes: full SDLC with quality gates, structured code review, deployment pipeline
- 4 document templates: PRD, Technical Design, ADR, Agent Decision Record
- Onboarding config file (`onboarding.yaml`) for easy team customisation
- `CLAUDE.md` entry point that integrates with Claude Code
- Polished static landing page at `site/index.html`
- Getting started guide with step-by-step setup instructions
- **`.claude/` runnable layer**: 6 hooks, 6 modular rule files, 5 sub-agents, 5 slash-command skills, hook wiring
- **`golden-paths/pipelines/`**: 7 drop-in GitHub Actions CI workflows

### Roles (19 total)

| Department | Roles |
|------------|-------|
| Engineering (7) | Head of Eng, Tech Lead, Backend, Frontend, QA, Platform, SRE |
| Product (3) | Head of Product, PM, Product Analyst |
| Design (3) | Head of Design, UI Designer, UX Designer |
| Security (3) | Head of Security, Security Auditor, Pen Tester |
| Data (3) | Head of Data, Data Analyst, Data Engineer |

### Workflows (3)
- **SDLC** — Planning → Design → Build → Review → QA → Deploy → Monitor
- **Code Review** — author/reviewer responsibilities, checklists, severity levels
- **Deployment** — CI/CD stages, environment promotion, rollback procedures

### Templates (4)
- PRD, Technical Design, ADR, AgDR

### Configuration
- `onboarding.yaml` — company name, team members, tech stack, workflow preferences
- `CLAUDE.md` — stack entry point for Claude Code integration

### `.claude/` runnable layer (NEW)

Until this layer landed, ApexStack shipped only the *passive* surface — markdown for roles, workflows, and templates. There was nothing a user could actually invoke. The `.claude/` directory closes that gap with the four primitives Claude Code reads at the project root:

| Layer | Files | What it does |
|-------|------:|--------------|
| `hooks/` | 6 | `block-git-add-all.sh`, `block-main-push.sh`, `check-secrets.sh`, `pre-push-gate.sh`, `validate-branch-name.sh`, `validate-pr-create.sh` — enforce no `git add -A`, no direct push to main, no committed secrets, valid branch / PR title format |
[...]

PR #169, opened 2026-05-04
chore(#168): accept release/vN.N.N branches + release(...) PR titles

atlas-apex:
## Summary

Closes me2resh/apexyard#168. Fixes the contradiction between the `/release` skill's prescribed branch name + PR title and the framework's own validators. Surfaced cutting v1.2.0 — the first release under the dev/main model (AgDR-0007).

Three small changes:

1. **`.claude/hooks/validate-branch-name.sh`** — added an early-out clause that accepts `^release/v\d+\.\d+\.\d+(-rcN)?$` as a valid branch name. Narrow, intentional exception for the framework's release-cut convention; release branches don't carry a ticket-id because the release itself IS the ticket.

2. **`.claude/project-config.defaults.json`** — added `release` to `pr.title_type_whitelist` so a title like `release(#160): v1.2.0` passes `validate-pr-create.sh`'s existing regex unchanged.

3. **`.claude/skills/release/SKILL.md`** step 4 — corrected the prescribed PR title to `release(#<release-ticket>): vA.B.C` so future `/release` invocations produce a title that satisfies the validators by construction.

## Why now

Cutting v1.2.0 is blocked on this. The release-cut convention exists; the validators don't recognise it; the contradiction was invisible until the first release tried to land. Tiny scope (3 file edits), but blocks the v1.2.0 release.

## Testing

- [x] **Branch validator probes** against all relevant patterns:

  ```
  release/v1.2.0           → rc=0  (allowed, release-special-case)
  release/v1.2.0-rc1       → rc=0  (allowed, RC variant)
  release/v9.9.9           → rc=0  (allowed)
  release/foo              → rc=2  (correctly blocked)
  release/v1               → rc=2  (correctly blocked — no minor.patch)
  chore/GH-168-fix         → rc=0  (allowed, standard pattern)
  feature/GH-1-x           → rc=0  (allowed, standard pattern)
  ```

- [x] **Full hook test suite** — `for t in .claude/hooks/tests/test_*.sh; do bash "$t"; done` → **196/196 across 12 test files**. No regression.
- [x] **PR-title regex unchanged** — `validate-pr-create.sh` is untouched; only the type whitelist gets a new entry.
- [ ] **Live integration** — confirmed by the act of opening this PR (title `chore(#168): ...` passes the same regex it always did).

## Glossary

| Term | Definition |
|------|------------|
| **Release-cut convention** | The dev/main model adopted in AgDR-0007 / #116. Release PRs originate from `dev`, land on `main`, and get a semver tag on the merge commit. The `/release` skill automates this; the source branch's prescribed name is `release/vA.B.C`. |
| **Special-case branch exception** | A narrow, intentional bypass in `validate-branch-name.sh` for branches that match a known framework convention without fitting the standard `{type}/{TICKET-ID}-{description}` shape. Used here for `release/vN.N.N` because release branches are scoped by the release itself, not by a single ticket. |
| **PR-title type whitelist** | The `pr.title_type_whitelist` array in `.claude/project-config.{defaults,}.json`. Drives the `validate-pr-create.sh` regex's `${PR_TYPES}` capture group. Adding `release` lets `release(#N): vA.B.C` pass. |
| **Validator-by-construction** | A skill that produces output which satisfies the framework's validators without the operator having to manually massage it. The /release skill is now validator-by-construction for both the branch name and the PR title — earlier it produced names that the validators rejected. |

### DIFF
--- a/.claude/hooks/validate-branch-name.sh
+++ b/.claude/hooks/validate-branch-name.sh
@@ -26,6 +26,17 @@ if [ "$CURRENT_BRANCH" = "main" ] || [ "$CURRENT_BRANCH" = "master" ] || [ "$CUR
   exit 0
 fi
 
+# Allow release-cut branches (apexyard#116, AgDR-0007). The /release skill
+# prescribes `release/vN.N.N` (and optionally a `-rcN` suffix) as the
+# canonical name for the dev → main release PR's source branch. This is
+# a narrow, intentional exception to the standard {type}/{TICKET}-{desc}
+# shape — release branches don't carry a ticket-id because the release
+# itself is the ticket. See me2resh/apexyard#168 for why this exception
+# exists.
+if echo "$CURRENT_BRANCH" | grep -qE '^release/v[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$'; then
+  exit 0
+fi
+
 # Load the branch-type whitelist from project config.
 REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
 # shellcheck source=./_lib-read-config.sh
--- a/.claude/project-config.defaults.json
+++ b/.claude/project-config.defaults.json
@@ -75,7 +75,8 @@
       "build",
       "ci",
       "chore",
-      "revert"
+      "revert",
+      "release"
     ],
     "required_sections": [
       "Testing",
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -82,7 +82,7 @@ Branch from `dev`: `release/vA.B.C`. Push to `upstream`. Open PR:
 
 - **Base**: `main`
 - **Head**: `release/vA.B.C`
-- **Title**: `release: vA.B.C`
+- **Title**: `release(#<release-ticket>): vA.B.C` — e.g. `release(#160): v1.2.0`. The release-cut ticket (filed via the standard ticket flow) is the natural scope, and `release` was added to the `pr.title_type_whitelist` in #168 so this title shape passes `validate-pr-create.sh` like every other PR title.
 - **Body**: the CHANGELOG draft + an explicit "this PR will tag `vA.B.C` on `main` after merge"
 
 The PR body should aggregate every `Closes #N` from the included commits so that merging the release PR auto-closes all of them on GitHub at once.


atlas-apex (commented):
## Code Review: PR #169

**Commit**: `549e374144e21fcb18709c44c09ce9a7ceb091d1`

### Summary

Three-file consistency fix that resolves the contradiction between the `/release` skill's prescribed branch + title and the framework's own `validate-branch-name.sh` + `validate-pr-create.sh`. Surfaced cutting v1.2.0 (the first release under the dev/main model). Tight scope, +14/-2.

### Checklist Results

- Testing:                   Pass (existing test suite)
- Security:                  Pass (no security surface touched)
- Performance:               Pass (N/A — single regex addition)
- PR Description & Glossary: Pass (Glossary present, 4 terms)
- Technical Decisions (AgDR): Pass (correctly cites AgDR-0007 + apexyard#116, the existing decision being recognised; this PR is a consistency fix, not a new decision)

### Verification performed

**1. Validator regex correctness (`^release/v[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$`)** — probed 19 boundary cases:
- `release/v1.2.0`, `release/v1.2.0-rc1`, `release/v9.9.9`, `release/v1.2.0-rc99`, `release/v1.2.3-rc0` → ALLOW (correct)
- `release/foo`, `release/v1`, `release/v1.2`, `release/v1.2.3-beta`, `release/v1.2.3-rc`, `release/v1.2.0.0`, `release/v1.2.3-rc1-extra`, `Release/v1.2.0` (case), leading/trailing whitespace → fall through to standard regex (correctly rejected)
- `release/v1.2.3-rc01` (zero-padded rc) is **allowed** by `[0-9]+` — note: the PR's "things to scrutinise" prompt expected this to fail, but `[0-9]+` matches one-or-more digits so `rc01` does match. The regex is internally consistent; if zero-padded RCs need to be rejected, that's a separate semver-strictness choice and would need `[1-9][0-9]*`. Documenting, not blocking — framework usage probably doesn't care, and tightening can land in a follow-up if AgDR-0007 wants strict semver on RC numbering.
- `release/v01.02.03` (zero-padded versions) is allowed for the same reason. Same rationale: documenting, not blocking.

**2. Position of the early-out** — placed AFTER the trunk exemption (`main`/`master`/`develop`/`dev`) and BEFORE the config load. Order is correct: trunk branches are a higher-priority special-case (universal), release-cut is a framework-specific special-case (no config dependency), and the type-whitelist load handles everything else. Hardcoding the release-cut pattern means it works even when config is unreadable (jq missing, standalone install) — same defensive posture as the existing trunk check.

**3. Type whitelist consistency** — `release` added to `pr.title_type_whitelist` only. NOT added to `branch.type_whitelist` (correctly — the release-cut branch shape `release/vN.N.N` is special-cased earlier; adding to whitelist would let `release/foo-bar` slip through). NOT added to `commit.type_whitelist` (correctly — commits on a release branch use existing types like `chore(#160): release v1.2.0`). Boundary is correct.

**4. `/release` skill prescription (step 4)** — title now `release(#<release-ticket>): vA.B.C` with an inline justification. Verified the new title shape passes `validate-pr-create.sh` against the updated whitelist (`release(#160): v1.2.0` → PASS; old form `release: vA.B.C` → FAIL). Validator-by-construction.

**5. Test suite regression** — ran `for t in .claude/hooks/tests/test_*.sh; do bash "$t"; done`: **196/196 across 12 files**, zero failures. PR claim verified exactly.

**6. PR title + body + branch** — title `chore(#168): ...` matches existing regex. Body has Summary / Why now / Testing / Glossary. Single `Closes me2resh/apexyard#168` in plain prose. Branch `chore/GH-168-release-validator-fix` matches the standard `{type}/{TICKET}-{desc}` shape.

**7. Leak protection** — diff contains zero registered-private-project references. Only `me2resh/apexyard` (public framework target).

**8. AgDR linkage** — hook comment cites `apexyard#116, AgDR-0007` (release-cut model decision) — file exists at `docs/agdr/AgDR-0007-release-cut-branch-model.md`. Glossary in PR body also references AgDR-0007. Linkage correct.

**9. Standard regex unaffected** — branches that don't match the new release special-case (and aren't trunk) fall through to the existing `^(${TYPES})/([A-Z]{2,10}-[0-9]+|GH-[0-9]+|#[0-9]+)-` regex unchanged. Verified by reading the full hook.

### Observations (non-blocking — follow-up candidates)

These were called out in the review prompt and confirmed; none block this PR but are worth a follow-up.

1. **`docs/release-process.md:53`** — the manual recipe still uses the broken `title="release: v1.2.0"` shape via `gh api .../pulls -X POST`. This bypasses `validate-pr-create.sh` (which only matches `gh pr create`, not raw `gh api` POSTs), so no operator hits a hard failure — but the documentation now contradicts both the `/release` skill's new prescription AND the validator-passing shape this PR enables. The PR scope says "optional, cosmetic"; my read is that it's **mildly more than cosmetic** because the file is the manual fallback for when `/release` doesn't run, and copy-pasting it produces a PR title that the framework's *own* PR validator would reject if invoked through `gh pr create`. Worth a tiny follow-up: change line 53 to `-f title="release(#<release-ticket>): v1.2.0"` and align line 51's heading with the new shape. Not blocking this PR — fixing it would expand scope past three files.

2. **No dedicated regression test for `validate-branch-name.sh`** — existing test files cover other hooks. The 19 boundary cases probed manually here would be a natural fit for a `test_validate_branch_name.sh` companion. Worth a follow-up; the manual probe gives high confidence for now.

3. **Zero-padded RC and version numbers** are allowed by `[0-9]+`. If AgDR-0007 wants strict semver, tighten to `[1-9][0-9]*` (with explicit `0` allowance for `v0.x.y`). Documenting only.

### Verdict

**APPROVED** (posted as comment — author cannot self-approve via `--approve`).

Minimal, correct, surgical fix that closes a real framework contradiction blocking v1.2.0. The boundary handling is right (release-cut as a narrow special-case rather than a wholesale relaxation), the early-out is positioned correctly (after trunk, before config load), the PR-title whitelist is the right scope (not branch, not commit), and the `/release` skill is now validator-by-construction. Test suite green at 196/196. AgDR linkage to AgDR-0007 is appropriate. Recommend the three follow-up items above land in a separate PR after this merges.

---
📌 Reviewed commit: `549e374144e21fcb18709c44c09ce9a7ceb091d1`

PR #185, opened 2026-05-09
feat(#183): /launch-check historical trend tracking

atlas-apex:
## Summary

- Persist every `/launch-check` run as a JSON file under `<projects_dir>/<name>/launch-check/runs/<ISO-ts>.json` (timestamp + branch + commit + per-dimension scores + verdict + top_risks). Schema is forward-compatible — extra fields preserved on framework upgrade.
- Append a "Trend (last 5 runs)" section to the per-run summary when at least 2 prior runs exist: markdown table (Date | Score | Verdict | Notes) plus an ASCII score chart that renders identically in terminals, GitHub, and IDE markdown viewers.
- Add `/launch-check trend [project-path]` mode for read-only trend rendering (no full audit, no new run files written).
- Auto-derive the Notes column from the score-delta vs the previous run, e.g. "Security +12, Analytics +10". Operator-supplied notes are scoped to v2.
- Opt-in commit via `<lc_dir>/.launch-check-history-tracked` marker — gitignored by default to avoid history bloat in adopters' repos.
- Resolve the projects dir via `portfolio_projects_dir` from `_lib-portfolio-paths.sh` so split-portfolio adopters work transparently — no hardcoded `projects/` path.
- Ship a small `render-trend.sh` helper that does the heavy lifting (sort by `ts`, compute deltas, draw the chart). The skill markdown calls it; standalone callers (e.g. a CI Slack-poster) can use it directly.
- AgDR-0014 documents the chart-format / schema / opt-in choices (ASCII over Mermaid, open additive schema, gitignored-by-default).

Closes me2resh/apexyard#183

## Testing

1. `bash .claude/hooks/tests/test_launch_check_trend.sh` — 21 cases, covers single-run/empty/missing dirs (no output), 2-run / 5-run trend rendering, idempotency / read-only behaviour, auto-notes derivation, and forward-compat with extra unknown fields.
2. Full hook-test sweep — `bash .claude/hooks/tests/test_*.sh` — 213 cases across 13 test files all green; no regressions.
3. Visual smoke: synthesise 5 run JSON files matching the schema, run `.claude/skills/launch-check/render-trend.sh <runs_dir>`, eyeball the table + ASCII chart.

## Glossary

| Term | Definition |
|------|------------|
| Run history | Per-run JSON files under `<projects_dir>/<name>/launch-check/runs/`, one per `/launch-check` invocation, used as the input for trend rendering. |
[...]

PR #198, opened 2026-05-09
fix(#194): validation hooks read git context from command, not $PWD

### LINKED ISSUE #194: [Bug] Validation hooks resolve git context from $PWD, not the worktree where gh/git ran (breaks /fan-out)
## Given / When / Then

**Given** the parent session spawns parallel agents via `/fan-out` (or via direct `Agent` tool calls with `isolation: "worktree"`)

**When** any of those agents tries to `git push`, `gh pr create`, or `git commit -m "$(cat <<'EOF'...)"` from inside their worktree

**Then** the validation hooks `validate-branch-name.sh`, `validate-pr-create.sh`, and `validate-commit-format.sh` resolve git context against the **harness's $PWD** (typically a sibling worktree's directory in the parent session) — NOT the worktree where the agent's command actually ran. This causes the hooks to:

- Read the wrong branch name from `git branch --show-current`
- Read the wrong commit message from the wrong repo's HEAD
- Block the agent's legitimate operation with a misleading "your branch name doesn't match the convention" error

**Expected**: hooks resolve git context against the actual command's repo, either via the `--head` / `--repo` flag the operator passed, or via the directory the operator `cd`'d into before running the command. Not from a $PWD that's frozen at the harness-spawn moment.

## Repro

This is reproduced **every time** parallel agents run from worktrees in this session. From three concurrent sub-agent reports:

```
Agent A (#188):
  "validate-branch-name.sh / validate-pr-create.sh which read
   git branch --show-current from the harness's pwd, not the worktree
   I cd'd into. Worked around by renaming the host-worktree's branch."

Agent B (#189):
  "Branch / title / body all conform to the rules the hook would have
   validated. The push went through with `git -C` to override; the PR
   itself was created via `gh api POST /repos/.../pulls` after
   `gh pr create` repeatedly tripped the same misfire."

Agent C (#190):
  "The agent's home cwd is a registered-project worktree whose own
   branch name doesn't satisfy apexyard's validate-branch-name.sh,
   which walks up from $PWD. Worked around by temporarily placing a
   sentinel onboarding.yaml in the home dir."
```

Each agent independently hit the same bug and invented a different workaround — branch rename, `git -C`, sentinel file. None of those should be necessary.

The same shape is the `gh api .../pulls/<N>/merge` bypass that closed via #47 — a tool surface the hooks didn't originally cover. This ticket extends the same principle to validation hooks: gate on the command's actual context, not the harness's $PWD.

## Why this matters

`/fan-out` is a sanctioned framework pattern — the parallel-work rule says we should offer it. But every fan-out today produces 3 simultaneous workarounds for the same hook misfire. The system fights itself: it sanctions fan-out via /fan-out yet blocks fan-out via $PWD-bound hooks.

The cost compounds: each agent's workaround is invisible to Rex review (the parent only sees the PRs), so the worked-around behavior ships. Eventually an adopter reads the SKILL.md / hook source, reproduces the misfire, and learns the workaround — at which point the workaround becomes the de-facto pattern, and the rule's enforcement is hollow.

### Hook 1 — `validate-branch-name.sh`

Read the branch from the actual `git push` command's source ref:

```bash
# Today (broken in worktrees):
BRANCH=$(git branch --show-current)

# Fix:
# Parse the push command for the source ref. `git push origin <branch>` →
# extract <branch>. Fall back to `git branch --show-current` if no ref is
# in the command (no-arg push relies on upstream tracking).
BRANCH=$(extract_push_ref "$COMMAND" || git branch --show-current)
```

Helper goes in `_lib-extract-push-ref.sh` (sibling to `_lib-extract-pr.sh`).

### Hook 2 — `validate-pr-create.sh`

Read the branch from `gh pr create`'s `--head` flag:

```bash
# Today (broken):
BRANCH=$(git branch --show-current)

# Fix:
HEAD_FLAG=$(echo "$COMMAND" | sed -nE 's/.*--head[[:space:]]+([^[:space:]]+).*/\1/p' | head -1)
BRANCH="${HEAD_FLAG:-$(git branch --show-current)}"
```

Same fallback: if no `--head` flag, use the local branch (today's behavior). When `--head` IS specified, the hook uses it.

### Hook 3 — `validate-commit-format.sh`

The `git commit -m "$(cat <<'EOF'...)"` shell-substitution issue:

The hook reads the `-m` argument from the command string and treats it as the literal commit subject. When the operator uses `$(cat <<'EOF'...)` substitution, the substitution hasn't expanded yet at hook-invocation time — the hook sees the literal string `$(cat <<'EOF'...)` which obviously doesn't match the conventional-commit regex.

Fix: detect the heredoc-substitution pattern and skip validation of the subject (because the actual subject is in the heredoc body, not the `-m` argument string). Alternative: read the staged commit's actual message via `git commit --dry-run` after substitution. The first option is simpler.

```bash
# If COMMAND contains $(cat <<EOF ... EOF), the -m arg is dynamic.
# Skip subject validation for this shape; the operator can use git commit -F
# if they want subject validation on a multi-line message.
if echo "$COMMAND" | grep -qE 'git commit.*-m.*\$\(cat[[:space:]]*<<'; then
  echo "INFO: heredoc-substitution detected in -m; skipping subject validation. Use git commit -F file for validation on multi-line messages." >&2
  exit 0
fi
```

### Hook 4+ (out of scope of v1) — `validate-issue-structure.sh`, `block-private-refs-in-public-repos.sh`, etc.

These hooks already work on the explicit `--repo` / `--body-file` flags. They don't have the $PWD problem. Verify in passing during this work.

## Acceptance Criteria

- [ ] `validate-branch-name.sh` reads the branch from the push command's source ref when present, falling back to local HEAD otherwise.
- [ ] `validate-pr-create.sh` reads the branch from `--head` when present, falling back to local HEAD otherwise.
- [ ] `validate-commit-format.sh` skips subject validation when `$(cat <<'EOF'...)` substitution is detected in the `-m` argument; suggests `git commit -F file` as the alternative path.
- [ ] New shared helper `_lib-extract-push-ref.sh` for parsing the push command's source ref.
- [ ] Tests cover all three fixes:
  - `validate-branch-name.sh`: push from a worktree where $PWD is a sibling worktree → reads correct branch
  - `validate-pr-create.sh`: `gh pr create --head feature/GH-N-description` → reads `feature/GH-N-description`, not the local branch
  - `validate-commit-format.sh`: heredoc-substitution pattern → skips with INFO message; non-substitution `-m "subject"` validates as today
- [ ] Doc note in `.claude/hooks/README.md` explaining the cwd-vs-command-context distinction.
- [ ] Smoke test: re-run a /fan-out across 3 worktrees, confirm zero workarounds needed.

## Risks / Dependencies

- **False negatives** — the heredoc-substitution skip might let a legitimate format violation through if the heredoc body is malformed. Acceptable trade-off; `git commit -F` users still get full validation, and the heredoc pattern is uncommon enough that the risk is bounded.
- **Cross-platform** — the regex extraction assumes GNU grep/sed conventions. Test on macOS BSD tools (the dev fleet) before merge.
- **Backwards compatibility** — current behavior (read local HEAD) is preserved as the fallback. Adopters who don't pass `--head` see no change.

## Refs

- Surfaced during fan-out implementation of #188 / #189 / #190 — three independent agents hit the same bug and invented three different workarounds.
- Same shape as `gh api .../pulls/<N>/merge` bypass closed in [me2resh/apexyard#47](https://github.com/me2resh/apexyard/issues/47) — tool-surface coverage gap.
- Related: parallel-work rule (`.claude/rules/parallel-work.md`) — we sanction fan-out yet ship hooks that fight it.

atlas-apex:
## Summary

Validation hooks (`validate-branch-name.sh`, `validate-pr-create.sh`, `validate-commit-format.sh`) historically resolved git context against the harness's `$PWD` via `git branch --show-current` and inline `-m` extraction. Under `/fan-out`, each parallel agent's harness `$PWD` is typically a sibling worktree of the worktree the agent `cd`'d into for its own work — so the hooks read the wrong branch / wrong commit message and blocked legitimate operations with misleading errors. Three independent agents in a recent fan-out session each invented a different workaround (branch rename, `git -C`, sentinel `onboarding.yaml`, fall back to `gh api`).

Same root-cause class as the `gh api .../merge` bypass closed in #47: the hooks gated on `$PWD`-derived state instead of parsing the command for the actual context. Fix shape mirrors that one — **command-arg first, fallback to local context** — preserving today's behaviour for shapes that don't pass the new flags.

- New `_lib-extract-push-ref.sh` helper parses the source ref out of a `git push` command (supports `git push origin <ref>`, refspec form `<src>:<dst>`, `-u` / `--set-upstream`, `--force` / `--force-with-lease`, `HEAD` passthrough, `--delete` bail-out, no-ref fallback). Same shape as `_lib-extract-pr.sh` from #47.
- `validate-branch-name.sh` reads the branch from the helper when present, falls back to `git branch --show-current` otherwise.
- `validate-pr-create.sh` reads the branch from `--head` when present (one-liner sed extraction matching the existing `--repo` / `--title` style), falls back to local HEAD otherwise.
- `validate-commit-format.sh` detects the `git commit -m "$(cat <<'EOF'...)"` heredoc-substitution shape and exits 0 with an INFO message recommending `git commit -F file` for full validation on multi-line messages — the actual subject is in the heredoc body, invisible to the hook.
- `README.md` adds a "PWD-vs-command-context distinction" section so future hook authors have a written precedent (#47 + #194 are both surfaced).
- AgDR-0015 records the command-arg-first / fallback-to-local design call.

Backwards-compat is preserved: silent-command shapes (no `--head`, no source ref) still resolve via the existing `$PWD`-derived path. Only previously-broken shapes (worktree-isolated fan-out, heredoc substitution) move from blocked to allowed.

## Testing

1. Run the new test files:
   ```
   bash .claude/hooks/tests/test_validate_branch_name_pushref.sh   # 15 cases
   bash .claude/hooks/tests/test_validate_pr_create_head.sh        # 8 cases
   bash .claude/hooks/tests/test_validate_commit_format_heredoc.sh # 9 cases
   ```
2. Full hook test sweep: `for t in .claude/hooks/tests/test_*.sh; do bash "$t"; done` — 19 files, 0 failures locally.
3. End-to-end: this PR was pushed via `git push -u origin fix/GH-194-...` from a sibling-worktree harness `$PWD`. Pre-fix that path hard-blocked with the misleading "branch doesn't follow naming convention" message; post-fix the source ref carries the truth.

Closes #194

## Glossary

| Term | Definition |
|------|------------|
| Validation hook | A `PreToolUse` shell script in `.claude/hooks/` that gates a Bash command (`git push`, `gh pr create`, `git commit`) and blocks via exit code 2 when the command violates a rule. |
| Harness `$PWD` | The current working directory the Claude Code harness was sitting in when it dispatched the tool call. Often diverges from the directory the operator `cd`'d into for the actual command — particularly under `/fan-out` where parallel agents work in sibling worktrees. |
| Worktree | A linked checkout of a git repo that shares the same `.git/` storage but has its own branch and working directory. `/fan-out` uses worktrees to isolate parallel agents from each other. |
| Source ref | The local branch name in a `git push origin <ref>` command — what the operator is actually pushing, regardless of the harness's `$PWD`. |
| Heredoc substitution | The `git commit -m "$(cat <<'EOF' ... EOF)"` shape where the `-m` value is a shell command-substitution that hasn't yet expanded at hook-invocation time. The hook sees the literal substitution string, not the expanded message. |
| Fallback to local context | When the command itself doesn't carry the answer (e.g. `git push` with no args), resolving via `$PWD`-derived state (`git branch --show-current`) is the documented fallback — preserves backwards compatibility with shapes that never passed the now-recognised flags. |

### DIFF
--- a/.claude/hooks/README.md
+++ b/.claude/hooks/README.md
@@ -240,6 +240,57 @@ With that config, `wip: scratch work` is accepted and `refactor: cleanup` is rej
 
 **Enforces:** `.claude/rules/git-conventions.md § "Commit Message Format"` — was prose-only.
 
+## PWD-vs-command-context distinction
+
+The harness's `$PWD` at hook-invocation time is whatever directory the parent
+session was sitting in when it dispatched the tool call. For most single-shell
+sessions that happens to match the worktree the operator is editing — but it
+**does not have to**. Three failure modes recur:
+
+1. **Agent fan-out workers** — `/fan-out` (and direct `Agent` calls with
+   `isolation: "worktree"`) spawn parallel agents that `cd` into per-task
+   worktrees. The parent harness's `$PWD` may still point at a sibling
+   worktree or the ops-fork root.
+2. **Cross-repo shells** — operators routinely run `gh pr create --repo X`
+   from a clone of repo Y when bouncing between projects.
+3. **Backgrounded long-running shells** — a tool call dispatched from a
+   shell whose `cwd` has drifted since session start.
+
+The lesson from #194 (validation hooks) and #47 (the `gh api .../merge` bypass)
+is the same: **gate on the command's actual context, not on the harness's
+`$PWD`**. The command itself almost always carries the truth — `--head`,
+`--repo`, the push source-ref, the API path's `/pulls/<N>/merge` segment.
+Falling back to `$PWD`-derived state (`git branch --show-current`,
+`git rev-parse HEAD`, repo-root via parent-walk) is acceptable as a fallback
+when the command does not name the context, but it must never override what
+the command says.
+
+For hook authors:
+
+- Prefer the command-arg-first, fallback-to-local-context shape:
+  `BRANCH="${HEAD_FLAG:-$(git branch --show-current)}"`. The fallback
+  preserves backwards compatibility for shapes that don't pass the flag.
+- Centralise extraction in a `_lib-extract-*.sh` helper rather than reimplementing
+  the parsing inline in each hook — same pattern as `_lib-extract-pr.sh` for
+  the merge-gate hooks. One helper, one set of tests, all hooks safe.
+- Heredoc substitution (`-m "$(cat <<EOF...)"`) is a special case where the
+  command string the hook reads is **literal-pre-expansion** — the actual
+  message lives in the heredoc body and is invisible to the hook. Detect the
+  shape and skip validation rather than block; recommend the file-based
+  alternative (`git commit -F file`) in the INFO message so operators who
+  want full validation on multi-line content know where to go.
+
+Helpers that implement this convention:
+
+| Helper | Used by | What it parses |
+|--------|---------|----------------|
+| `_lib-extract-pr.sh` | `block-unreviewed-merge.sh`, `require-design-review-for-ui.sh`, `block-merge-on-red-ci.sh` | PR number from `gh pr merge` and `gh api .../pulls/<N>/merge` |
+| `_lib-extract-push-ref.sh` | `validate-branch-name.sh` | Source ref from `git push origin <ref>` (and refspec / -u / --set-upstream / --force-with-lease variants) |
+
+When you add a new hook that depends on git state, ask first: "is the answer
+already in the command string?" If yes, parse it from there. If no, falling
+back to local context is fine — but document the trade-off.
+
 ## Settings Ordering Note
 
 The new hooks are registered in `.claude/settings.json` alongside the existing ones on the same `Bash(git commit *)` / `Bash(gh pr merge *)` matchers. The Claude Code harness runs all matching hooks sequentially, and **any exit-2 blocks the tool call**. Order of registration within a matcher block is execution order. Current order (GH-13 additions shown in **bold**):
--- a/.claude/hooks/_lib-extract-push-ref.sh
+++ b/.claude/hooks/_lib-extract-push-ref.sh
@@ -1,0 +1,152 @@
+#!/bin/bash
+# Shared push-source-ref extraction for hooks that gate on `git push`.
+#
+# Not a hook itself (prefixed with `_lib-` so it's never wired as one). Sourced
+# by hooks via `. "$(dirname "$0")/_lib-extract-push-ref.sh"`.
+#
+# WHY THIS EXISTS
+# ---------------
+# Validation hooks like `validate-branch-name.sh` historically read the branch
+# from `git branch --show-current`, which resolves against the harness's $PWD.
+# When the harness $PWD is a sibling worktree of the worktree the operator
+# actually ran the command in (e.g. an Agent fan-out worker that `cd`'d into
+# its own worktree), the resolved branch is wrong — the hook reads the parent
+# session's branch, not the agent's.
+#
+# Fix: parse the actual command for the source ref. `git push origin <branch>`
+# carries the source ref directly; that's the ground truth, regardless of $PWD.
+# Falls back to no-op (caller uses local HEAD) when no ref is present in the
+# command (e.g. no-arg `git push` relying on upstream tracking).
+#
+# Same shape pattern as `_lib-extract-pr.sh` for the merge-gate hooks (#47):
+# gate on the command's actual context, not the harness's $PWD.
+#
+# USAGE
+# -----
+#   . "$(dirname "$0")/_lib-extract-push-ref.sh"
+#   PUSH_REF=$(extract_push_ref "$COMMAND")
+#   BRANCH="${PUSH_REF:-$(git branch --show-current)}"
+#
+# Refs: me2resh/apexyard#194
+
+# Echoes the source ref from a `git push` command, or empty if none found.
+#
+# Recognises:
+#   git push origin <branch>                           → <branch>
+#   git push origin <branch>:<remote-branch>           → <branch>  (LHS of refspec)
+#   git push -u origin <branch>                        → <branch>
+#   git push --set-upstream origin <branch>            → <branch>
+#   git push --force origin <branch>                   → <branch>
+#   git push --force-with-lease origin <branch>        → <branch>
+#   git push origin HEAD                               → empty (HEAD is not a ref name we can validate)
+#   git push                                           → empty (relies on upstream tracking)
+#   git push origin                                    → empty (no ref given)
+#   git push origin --delete <branch>                  → empty (deletion, no source-ref check)
+#
+# Returns: prints the ref to stdout (or empty), always exits 0.
+extract_push_ref() {
+  local cmd="$1"
+  local push_segment ref
+
+  # Bail on `--delete` / `-d` shapes — they don't have a source ref.
+  if echo "$cmd" | grep -qE '\bgit\s+push\b[^|;&]*(--delete\b|[[:space:]]-d\b)'; then
+    echo ""
+    return 0
+  fi
+
+  # Isolate the `git push ...` segment up to the first command separator
+  # (|, ;, &&, &) so `git push origin foo && echo bar` doesn't pick up `echo`
+  # tokens.
+  push_segment=$(echo "$cmd" | grep -oE '\bgit\s+push\b[^|;&]*' | head -1)
+  if [ -z "$push_segment" ]; then
+    echo ""
+    return 0
+  fi
+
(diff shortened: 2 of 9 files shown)


atlas-apex (commented):
## Code Review: PR #198

**Commit**: `2f4fc734ef08a8f9c9c9205550abf6d3243e864d`

### Summary

Fixes #194 by switching three validation hooks (`validate-branch-name.sh`, `validate-pr-create.sh`, `validate-commit-format.sh`) from `$PWD`-derived git context to command-arg-first resolution with `$PWD` fallback. Same shape as the `gh api .../merge` bypass closed in #47. Adds shared helper `_lib-extract-push-ref.sh`, three new test files (32 cases), AgDR-0015, and a README section formalising the convention for future hook authors.

### Checklist Results

- Architecture & Design:    **Pass** — clean separation (parser helper + hooks consume via `${VAR:-fallback}`), mirrors the `_lib-extract-pr.sh` convention from #47, README section documents the contract for future hook authors.
- Code Quality:             **Pass** — tight scope, defensive parsing (handles refspec, `+` force marker, `refs/heads/` prefix, HEAD passthrough, `--delete` bail, no-arg push). Inline comments explain *why* not just *what*.
- Testing:                  **Pass** — 32 cases cover the right axes: explicit ref, refspec form, `-u`/`--set-upstream`/`--force-with-lease`, `HEAD`, `--delete`, no-ref fallback, non-conforming/conforming branches, trunk exemptions (main/dev/master), heredoc shapes (`<<EOF`, `<<'EOF'`, `<<-`), `-F` path unaffected. Tests use isolated sandboxes with intentionally-wrong local branch so they actually verify the command-arg path won (vs. accidentally working because the local branch happened to be conforming).
- Security:                 **Pass** — no secrets, no shell-injection vectors (parsing is restricted to extracted segments, parameter expansion is preferred over `eval`/`sed -e`).
- Performance:              **Pass** — hooks run once per Bash tool call; one extra grep + tokeniser loop is negligible.
- PR Description & Glossary:**Pass** — Summary/Testing/Glossary all present and substantive.
- Technical Decisions (AgDR):**Pass** — AgDR-0015 covers the command-arg-first vs. `$PWD` fallback design call with five options weighed.

### Issues Found

**None blocking.** One coordination concern flagged below.

1. **AgDR-0015 number collision.** Three open PRs all use AgDR-0015:
   - #198 (this PR) — `AgDR-0015-command-context-over-pwd-in-hooks.md`
   - #200 — `AgDR-0015-journey-html-rendering.md`
   - #202 — `AgDR-0015-spike-skill-schema-and-exemptions.md`

   Main currently has AgDR-0001 through AgDR-0014 (PR #185 was merged with 0014). Whichever of #198/#200/#202 merges first claims 0015; the other two will need to bump to 0016/0017 on rebase. Worth surfacing on the merge-coordination thread so the rebases happen cleanly. Not a defect in this PR — just the parallel-fan-out symptom this PR itself documents.

2. **Verified BSD-sed compatibility.** The agent's note checks out: the only new `sed` invocation (`validate-pr-create.sh:289`) uses `[[:space:]]` POSIX bracket classes — no `\b`/`\s`/`\<`/`\>`/`\w`. The helper avoids `sed` entirely on the parsing-prone path, using shell parameter expansion (`${push_segment#*git}` style) — also BSD-safe. The remaining `\b`/`\s` in `grep -qE` patterns match the pre-existing convention in main's hooks (`validate-branch-name.sh:17`, `validate-commit-format.sh:30` already use `\bgit\s+push\b`-style on main; shellcheck CI is green there too). No GNU-isms introduced.

3. **Backwards-compat verified.** Both modified hooks use the `${VAR:-fallback}` shape so silent-command shapes (no `--head`, no source ref) still resolve via `git branch --show-current` exactly as before. Test cases `no-arg push: falls back to local HEAD` and `no --head: falls back to local HEAD` pin the fallback path against regression.

4. **No private project name leaks.** Diff scanned against the registered project list — clean.

5. **CI**: all 4 checks (Verify Ticket ID, lychee, markdownlint-cli2, shellcheck) green.

### Verdict

**APPROVED**

The fix is on-target and the test coverage is good. The AgDR-0015 collision is a coordination thing to handle at merge time, not a code defect — and the PR itself documents the class of bug that produced the collision, which is on-brand.

---
📌 Reviewed commit: `2f4fc734ef08a8f9c9c9205550abf6d3243e864d`

Issue #157, opened 2026-05-03
[Chore] Remove the voice-prompts-on-pause feature


atlas-apex:
## Driver

The voice-prompts-on-pause feature (PR #135, AgDR-0009) shipped as opt-in OFF by default and was disabled in this fork's local override on 2026-05-03 with the rationale "bundled macOS Daniel voice was unreliable / robotic and tripped on non-question messages." Two months of opt-in usage has produced zero adopters that turned it on; the phase-3 cloud-TTS work that AgDR-0009 anticipated never materialised.

The feature is dead code today: hook + test + 9k script + config block + AgDR all maintained, none firing in any production session. "Disabled-by-default opt-in" is the worst flavour of dead code — it ages, it occasionally needs maintenance during framework refactors, and it shows up in every fresh-adopter `/setup` config dump as a mysterious feature they have to ignore.

Cleaner to remove the feature and supersede the AgDR explaining why, rather than keep a permanently-off opt-in indefinitely.

## Scope

Remove all framework traces of the voice-prompts feature:

1. Delete `.claude/hooks/voice-prompt-on-pause.sh`
2. Delete `.claude/hooks/tests/test_voice_prompt_on_pause.sh`
3. Remove the Stop-hook entry from `.claude/settings.json` that wires `voice-prompt-on-pause.sh`
4. Remove the `voice_prompts` block from `.claude/project-config.defaults.json` (and its leading `_comment`)
5. Remove the `voice_prompts` block from this fork's local `.claude/project-config.json` (and its leading `_comment`) — local override, but stages alongside the framework change for tidiness
6. Write a new AgDR (next number — likely AgDR-0012) titled "Sunset the voice-prompts-on-pause feature" that supersedes AgDR-0009. **Keep AgDR-0009 in place** — the decision record is still valid as the why-we-built-it; the new AgDR is the why-we-pulled-it.
7. Update any other docs that reference `voice_prompts` or `voice-prompt-on-pause` (`docs/multi-project.md`, `docs/project-config.md`, etc. — grep first, edit only what hits)

## Acceptance Criteria

- [ ] `grep -rl "voice" .claude/ docs/ workflows/ templates/ roles/ CLAUDE.md` returns only AgDR-0009 (the historical record), AgDR-0012 (the supersession), and any incidental matches (e.g. the word "voice" used for unrelated meaning)
- [ ] No `voice-prompt` filename in `.claude/hooks/` or `.claude/hooks/tests/`
[...]

PR #199, opened 2026-05-09
feat(#177): /update detects deprecated config keys + offers cleanup

### LINKED ISSUE #177: [Chore] /update should detect deprecated config keys and offer cleanup (e.g. voice_prompts after #157)
## Driver

When the framework removes a config block (e.g. `voice_prompts` removed in #157), adopters who previously had that block in their `.claude/project-config.json` are left with **dead config** — keys that no code reads anymore but still sit in the override file. It's harmless but ugly, accumulates over time, and creates a small "is this still load-bearing?" cognitive tax for anyone reading the file later.

This is the same shape as deprecated dependencies in `package.json` after a library is removed — best practice is to clean them up, but adopters won't notice they're stale unless someone surfaces it.

## Scope

Extend `/update` (the upstream-sync skill) to detect and offer cleanup of deprecated config keys.

### Detection

After the merge/rebase step, `/update` already has the new defaults file (`.claude/project-config.defaults.json`) on disk. The detection is:

1. Read keys present in the adopter's `.claude/project-config.json` override
2. Read the top-level keys in `.claude/project-config.defaults.json`
3. For any override key NOT in defaults → flag as a candidate for removal

Caveats:

- The override file is allowed to have keys defaults doesn't ship — adopters may add custom config for their own hooks. Flagging should be advisory, not destructive.
- Some keys are intentionally optional in defaults but always present in overrides (e.g. `_comment` fields). Whitelist `_comment*` and similar metadata keys.

### Offer

After running `/update` and detecting deprecated keys, surface a summary:

```
ApexYard /update detected 1 config block in .claude/project-config.json
that no longer exists in upstream defaults:

  - voice_prompts (removed upstream in #157)

These keys are now dead — no framework code reads them. Want to remove
them from your override?

  [y] yes, remove the listed keys
  [n] no, leave them alone (they're harmless; you can clean up later)
  [s] show me the keys before deciding
```

If `y`: edit `.claude/project-config.json` to remove the flagged keys, stage the change for the operator's review (do NOT auto-commit — same pattern as the rest of `/update`).

If `n` / `s`: print the keys + their values for the operator to decide manually.

### Out of scope

- Detecting deprecated keys WITHIN a still-existing block (e.g. if `ticket.skip_marker` were renamed to `ticket.skip_pattern`). That's a deeper schema-versioning problem; v1 only catches whole-block removals.
- Auto-cleanup of stale hooks, skills, or rules. Those land via `/update`'s normal merge path. The dead-config issue is specific to the override file, which `/update` doesn't touch directly.

## Acceptance Criteria

- [ ] After the merge step, `/update` reads both the new defaults and the override.
- [ ] Top-level keys in the override that aren't in defaults are listed as candidates.
- [ ] Whitelist metadata keys (`_comment`, `_schema_version`) so they don't get flagged.
- [ ] Operator is offered y/n/s with a clear summary of what would be removed.
- [ ] On `y`, the override is edited (not auto-committed) and the change is staged.
- [ ] Test cases: override has only known-deprecated key (flagged); override has custom key (not flagged); override has `_comment` (not flagged); override matches defaults (no offer).
- [ ] Doc note in `.claude/skills/update/SKILL.md` describing the new step.

## Risks / Dependencies

- **False positives.** An adopter with custom-extension keys might be told their custom config is "deprecated". Mitigation: the offer is advisory + `[s]how me` lets the operator see exactly what's flagged before deciding.
- **Whitelisting.** The metadata-key whitelist is a small list to start (`_comment`, `_schema_version`); will need extension as new metadata patterns emerge. Document as a living list.
- **Surfaced only on `/update`.** Adopters who don't run `/update` regularly won't see the offer. Acceptable v1 — `/update` is the natural moment to notice upstream changes.

## Refs

- Surfaced when an adopter ran `/update` after voice_prompts was removed (#157) and noticed the dead block lingering in their override.
- Related: AgDR-0009 (voice prompts), #157 (voice_prompts removal).

atlas-apex:
## Summary

- Extends `/update` (the upstream-sync skill) with a new step 8 that detects **top-level config keys** present in `.claude/project-config.json` but absent from `.claude/project-config.defaults.json` — typically a config block the framework removed upstream (e.g. `voice_prompts` removed in #157) that lingers in the override as dead config.
- Adds `_lib-detect-deprecated-config.sh` with three pure functions: `detect_deprecated_config_keys` (returns the newline-separated list), `show_deprecated_config_keys` (prints each key + current value for the `[s]how` branch), and `remove_deprecated_config_keys` (in-place jq edit, no auto-commit).
- Whitelists leading-underscore metadata keys (`_comment`, `_schema_version`, `_team_comment`, etc.) so they're never flagged.
- The skill offers operator `y / n / s`. `y` removes the keys and stages the change for the operator's review — never auto-commits. `n` leaves the override untouched. `s` shows current values, then re-prompts y/n.
- Detection is **advisory only**. The detector cannot tell upstream-removed keys apart from custom-extension keys an adopter has added intentionally — the human-in-the-loop offer is what disambiguates.

Closes #177

## Testing

1. Run the new test suite — covers seven cases (single deprecated key flagged, custom-extension key surfaced advisorily, metadata whitelist, fully-aligned override produces no offer, multi-key sorted output, missing override file, remove + preserve):
   ```bash
   bash .claude/hooks/tests/test_detect_deprecated_config.sh
   # expect: PASS: 7  FAIL: 0
   ```
2. Run the full hook test sweep to confirm no regressions:
   ```bash
   for t in .claude/hooks/tests/test_*.sh; do echo "=== $(basename $t) ==="; bash "$t" 2>&1 | tail -3; done
   # expect: every test file reports zero failures
   ```
3. Spot-check the helper end-to-end against this fork's own files:
   ```bash
   source .claude/hooks/_lib-detect-deprecated-config.sh
   detect_deprecated_config_keys
   # expect: empty (no override on dev), or just the keys the operator legitimately added
   ```
4. Read the new step 8 in `.claude/skills/update/SKILL.md` and confirm the y/n/s prompt copy matches the ticket AC.

## Glossary

| Term | Definition |
|------|------------|
| deprecated config key | A top-level key in `.claude/project-config.json` that no longer exists in `.claude/project-config.defaults.json`, typically because the framework removed the corresponding block in a release |
| override file | `.claude/project-config.json` — the adopter's per-fork customisation layer, shallow-merged on top of `project-config.defaults.json` |
| advisory detection | A check that surfaces candidates for the operator to confirm, never auto-removing — sibling shape to the rest of `/update`'s "operator owns each material change" stance |
| metadata key | A non-config field whose name starts with `_` (e.g. `_comment`, `_schema_version`); always whitelisted by the detector |
| custom-extension key | A top-level key an adopter has added to the override for their own hooks/skills, indistinguishable from an upstream-removed key without operator input |

### DIFF
--- a/.claude/hooks/_lib-detect-deprecated-config.sh
+++ b/.claude/hooks/_lib-detect-deprecated-config.sh
@@ -1,0 +1,200 @@
+#!/bin/bash
+# _lib-detect-deprecated-config.sh — detect override keys absent from defaults
+#
+# Source this library from the /update skill (or any other tool) to surface
+# top-level keys present in `.claude/project-config.json` (the adopter override)
+# that are NOT present in `.claude/project-config.defaults.json` (the upstream-
+# shipped defaults). The classic case is a config block removed upstream — for
+# example `voice_prompts` removed in me2resh/apexyard#157 — that lingers in the
+# adopter's override file as dead config.
+#
+# Output is ADVISORY ONLY. The helper never edits the override file. It just
+# emits a newline-separated list of deprecated key names on stdout. The caller
+# (a skill) decides what to do with that list — typically: prompt the operator
+# y / n / s, then on `y` edit the override JSON via jq.
+#
+# Usage:
+#   source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-detect-deprecated-config.sh"
+#   deprecated=$(detect_deprecated_config_keys)
+#   for key in $deprecated; do echo "  - $key (no longer in upstream defaults)"; done
+#
+# Optional second arg lets callers point at non-default file paths (used by
+# tests). When omitted, paths are derived from the repo root.
+#
+# Whitelist: metadata keys whose name starts with `_` (e.g. `_comment`,
+# `_schema_version`) are always ignored. They're documentation / schema-version
+# fields, not deprecated config blocks. The whitelist is intentionally simple —
+# a leading underscore is the framework convention for non-config metadata.
+#
+# Custom-extension keys: an adopter may legitimately add their own top-level
+# keys to the override (custom hooks, in-house extensions). The helper surfaces
+# these the same way it surfaces upstream-removed keys — the caller's job is
+# to make the offer informational, not destructive. The skill prompts y/n/s
+# and the operator decides whether each flagged key is dead or load-bearing.
+
+# ---------------------------------------------------------------------------
+# Internal: resolve repo root when no explicit paths are passed.
+# ---------------------------------------------------------------------------
+_dep_repo_root() {
+  git rev-parse --show-toplevel 2>/dev/null
+}
+
+# ---------------------------------------------------------------------------
+# Public: detect_deprecated_config_keys [defaults_path] [overrides_path]
+#
+# Emits one deprecated key name per line on stdout.
+# Exit codes:
+#   0 — completed (zero or more keys emitted)
+#   1 — jq missing, or required input file missing — caller should treat
+#       as "skip detection silently" rather than blocking the whole skill.
+#
+# Design notes:
+#   - Reads only TOP-LEVEL keys. The ticket explicitly scopes this to whole-
+#     block removals (e.g. `voice_prompts` as a whole), not to renamed keys
+#     within a still-existing block. That deeper schema-versioning concern
+#     is out of scope for v1 (see ticket § "Out of scope").
+#   - Whitelist is hard-coded to the leading-underscore convention. This is
+#     the simplest workable rule and matches every metadata field the
+#     framework currently ships (`_comment`, `_schema_version`,
+#     `_*_comment`). If the convention ever needs to broaden (adopter-added
+#     allowlist), the function below is the one obvious place to plumb a
+#     second argument.
+# ---------------------------------------------------------------------------
+detect_deprecated_config_keys() {
+  local defaults="${1:-}"
+  local overrides="${2:-}"
+
+  if [ -z "$defaults" ] || [ -z "$overrides" ]; then
+    local root
+    root=$(_dep_repo_root)
+    [ -z "$defaults" ] && defaults="$root/.claude/project-config.defaults.json"
+    [ -z "$overrides" ] && overrides="$root/.claude/project-config.json"
+  fi
+
+  if ! command -v jq >/dev/null 2>&1; then
+    echo "WARN: jq not installed; skipping deprecated-config detection." >&2
+    return 1
+  fi
+
+  if [ ! -f "$defaults" ]; then
+    # No defaults file — caller is not in an apexyard-shaped repo.
+    return 1
+  fi
+
+  if [ ! -f "$overrides" ]; then
+    # No override — by definition no deprecated keys to surface.
+    return 0
+  fi
+
+  # jq:
+  #   - keys: top-level key names in each file
+  #   - subtract default keys from override keys
+  #   - filter out leading-underscore metadata
+  #   - emit one per line, sorted for stable output
+  jq -r --slurpfile def "$defaults" '
+    [keys[]] as $okeys
+    | ($def[0] | keys) as $dkeys
+    | $okeys
+    | map(select(
+        (. as $k | $dkeys | index($k) | not)
+        and (startswith("_") | not)
+      ))
+    | sort
+    | .[]
+  ' "$overrides" 2>/dev/null
+}
+
+# ---------------------------------------------------------------------------
+# Public: show_deprecated_config_keys [defaults_path] [overrides_path]
+#
+# Pretty-prints each deprecated key name and its current value (truncated to
+# the first ~3 lines of jq's compact output) so the operator can decide
+# whether each is dead config or a custom extension before answering y/n.
+#
+# Used by the [s]how branch of the /update prompt.
+# ---------------------------------------------------------------------------
+show_deprecated_config_keys() {
+  local defaults="${1:-}"
+  local overrides="${2:-}"
+
+  if [ -z "$defaults" ] || [ -z "$overrides" ]; then
+    local root
+    root=$(_dep_repo_root)
+    [ -z "$defaults" ] && defaults="$root/.claude/project-config.defaults.json"
+    [ -z "$overrides" ] && overrides="$root/.claude/project-config.json"
+  fi
+
+  local keys
+  keys=$(detect_deprecated_config_keys "$defaults" "$overrides")
+  [ -z "$keys" ] && return 0
+
+  while IFS= read -r key; do
+    [ -z "$key" ] && continue
+    local value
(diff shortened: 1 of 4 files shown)


atlas-apex (commented):
## Code Review: PR #199

**Commit**: `33656adc12ff85ff1f8cafb39e3589049ccce9a6`

### Summary

Extends `/update` with an advisory step 8 that detects top-level keys in `.claude/project-config.json` absent from `.claude/project-config.defaults.json` (typically dead config from an upstream-removed block, e.g. `voice_prompts` after #157) and offers a `y/n/s` cleanup prompt. New helper `_lib-detect-deprecated-config.sh` exposes three pure functions (`detect_*` / `show_*` / `remove_*`); 7-case test suite covers the contract.

### Checklist Results

- Architecture & Design:    Pass — pure-function helper, no side effects in detection, in-place jq edit only when the operator explicitly says yes; mirrors the rest of `/update`'s "operator owns each material change" stance.
- Code Quality:              Pass — well-commented, scope explicitly bounded (top-level keys only, deeper schema-versioning called out as out-of-scope), shellcheck disable in the helper is justified inline.
- Testing:                   Pass — 7 cases cover the happy paths, the metadata whitelist, the empty-output / no-flags case, and the in-place removal preserving whitelisted + active keys. Verified locally on the PR head: `PASS: 7  FAIL: 0`.
- Security:                  Pass — reads only `.claude/project-config.json`, no exec of arbitrary content, no shell injection vectors via key names (jq receives them as filter args, not interpolated string).
- Performance:               N/A — runs once per `/update` invocation on tiny JSON.
- PR Description & Glossary: Pass — Summary, Testing, Glossary all present; glossary distinguishes "deprecated config key" / "custom-extension key" / "advisory detection" / "metadata key" / "override file".
- Technical Decisions (AgDR):N/A — extends an existing skill, no new library / framework / architecture.

### Behavioural verification (the seven points)

I checked out the PR head locally, ran the test suite, and re-read the lib + skill copy against the briefing's verification points:

1. **Detection is advisory only** — confirmed. `detect_deprecated_config_keys` reads only; the in-place jq rewrite in `remove_deprecated_config_keys` only fires when the skill's y/n/s branch lands on `y`.
2. **Metadata key whitelist** — confirmed. The jq filter `startswith("_") | not` excludes `_comment`, `_schema_version`, `_team_comment`, etc. Test case 3 asserts.
3. **Custom-extension handling** — confirmed. SKILL.md step 8 explicitly tells the operator the keys "may be (a) dead config from a block the framework removed upstream … or (b) custom extension keys you've added intentionally. The detector can't tell them apart". Test case 2 asserts the key IS surfaced (advisory), not silently classified.
4. **`y` only stages** — confirmed. Lib's `remove_deprecated_config_keys` does the in-place jq rewrite and prints the count; the skill's `y` row in the action table runs `git add .claude/project-config.json` after, never `git commit`. The phrase "The skill **never auto-commits** — staging is the contract, the operator owns the commit" is in SKILL.md verbatim.
5. **No private project names** — confirmed. Grepped the diff for the registered project names, zero hits.
6. **CI green** — **NOT met** (see Blocker below).
7. **PR body has Summary / Testing / Glossary** — confirmed.

### Issues Found

#### Blocker: CI is red

The `shellcheck .claude/hooks` job has failed on this PR (https://github.com/me2resh/apexyard/actions/runs/25609956002). Two SC1090 warnings, both in the new test file:

```
./.claude/hooks/tests/test_detect_deprecated_config.sh:159:3: warning:
  ShellCheck can't follow non-constant source. Use a directive to specify location. [SC1090]

./.claude/hooks/tests/test_detect_deprecated_config.sh:190:3: warning:
  ShellCheck can't follow non-constant source. Use a directive to specify location. [SC1090]
```

These are the two bare `. "$LIB_SRC"` calls in cases 6 and 7. Cases 1–5 already use `# shellcheck source=/dev/null` inside `run_case` — cases 6 and 7 just need the same directive on the line above each `. "$LIB_SRC"`. Two-line fix:

```bash
# Case 6 (line 159 currently):
+# shellcheck source=/dev/null
 . "$LIB_SRC"

# Case 7 (line 190 currently):
+# shellcheck source=/dev/null
 . "$LIB_SRC"
```

Per `.claude/rules/pr-quality.md` § "No Red CI Before Merge": **never merge with red CI — even if the failure is pre-existing or unrelated**. This one's caused by the PR itself, so it's a hard blocker on merge until shellcheck goes green. Cannot use `--request-changes` because GitHub disallows that on a self-PR; treating this as **CHANGES REQUESTED in spirit** via `--comment`.

- Consider hoisting cases 6 and 7 into the same `run_case` helper as cases 1–5 — they're slightly bespoke (asserting on `rc6`, asserting on jq's post-removal state) but a small `assert_eq` / `assert_rc` pair would let them share the existing source-directive boilerplate. Not blocking; current shape is readable.
- The lib's `remove_deprecated_config_keys` echoes `"0"` on the no-keys path AND on the jq-missing path AND on the no-deprecated-keys path. A caller that does `count=$(remove_*); if [ "$count" -gt 0 ]; then echo "removed $count"; fi` will work fine, but the three "0" returns mean different things (skipped, nothing-to-do, succeeded-but-no-op). The skill currently doesn't distinguish, which is fine — flagging in case a future caller wants to.

### Verdict

**CHANGES REQUESTED** (posted as a comment because GitHub disallows `request-changes` on self-PRs).

The substantive design is solid and six of seven verification points hold against the code. The seventh (CI green) is the blocker. Two-line shellcheck fix, then this is good to merge.

---
Reviewed commit: `33656adc12ff85ff1f8cafb39e3589049ccce9a6`

atlas-apex (commented):
## Code Review: PR #199 (re-review)

**Commit**: `fde8872ab0170a62caf7548198022fcbb4623c1b`

### Summary

Re-review at HEAD `fde8872` after the shellcheck fix. Previous review at `33656ad` requested changes for SC1090 on the two bare `. "$LIB_SRC"` calls in cases 6 and 7 of `test_detect_deprecated_config.sh`.

### Diff verification

`33656ad..fde8872` is exactly the 2-line directive addition the fix advertised:

```
test_detect_deprecated_config.sh:158  + # shellcheck source=/dev/null
test_detect_deprecated_config.sh:191  + # shellcheck source=/dev/null
```

Both directives sit immediately above the bare `. "$LIB_SRC"` calls, matching the directives already in place above the same call inside `run_case` (line ~63). Consistent shape across all three sourcing sites in the file.

### CI

All four checks green:

- Verify Ticket ID — pass
- lychee — pass
- markdownlint-cli2 — pass
- **shellcheck .claude/hooks — pass** (was the failing check at `33656ad`)

### Checklist Results

- Technical Decisions (AgDR): N/A (test scaffolding fix only — no decision surface)

### Issues Found

None.

### Verdict

**APPROVED** (recorded as comment because GitHub disallows self-approval; approval marker written for the merge gate)

---
📌 Reviewed commit: `fde8872ab0170a62caf7548198022fcbb4623c1b`

Issue #208, opened 2026-05-11
[Feature] /setup enables LSP by default — install language server + env var + plugin in one flow


atlas-apex:
## User Story

As an adopter running `/setup` on a fresh apexyard fork, I want LSP installed + configured + enabled as part of the bootstrap flow (with the option to decline), so that I get the documented 3-15× token savings on `/code-review`, `/threat-model`, `/security-review`, and `/handover` deep-dive without having to learn about `ENABLE_LSP_TOOL`, find the right language-server binary, install the right Claude Code plugin, and edit my shell rc.

## Driver

`/setup`'s job is to take a fresh fork to "ready to use" in ~2 minutes. Today's flow stops short of LSP — adopters get LSP-aware skill callouts but no enabled LSP. They have to read `docs/getting-started.md`, install per-language servers (npm/pip/etc.), set an env var, and install the Claude Code LSP plugin themselves. The "3 minutes of setup buys 3-15× per-query savings" trade-off is the right one, but adopters won't make it if the work is hidden in docs.

From the user's testing of this fork on a constrained machine: "we are NOT using LSP" was the diagnostic — three pre-requisites all unmet (env var, plugin, language server binaries). Framework recommends LSP, marketing implies it's part of the team-feel, reality is it's opt-in via three independent setup steps spread across docs.

### 1. Extend `/setup` with an LSP step

After the existing Step 2a (privacy gate) and Step 2b (split-portfolio mode), add a new step:

```
Step 2c — LSP enablement (recommended)

LSP-aware code navigation makes /code-review, /threat-model,
/security-review, and /handover deep-dives 3-15× cheaper in token cost
on semantic queries (find-definition, walk-references, etc).

Enable LSP now? This will:
  - Install the LSP plugin for Claude Code (one-shot, ~5 sec)
  - Detect the project's language and install the right server
    (TypeScript: typescript-language-server via npm; Python: pyright;
    Go: gopls; Rust: rust-analyzer via rustup)
  - Set ENABLE_LSP_TOOL=1 in your shell rc

Skip if:
  - You're on a constrained machine (cold-start can be 30+s on
    large repos)
  - Your workflow is metadata-only (you only run /handover, /inbox,
    /status — these don't benefit from LSP)
  - You want to install per-language servers selectively later

[y / n / "y, but ask before installing"]
```

### 2. Language detection

From the operator's earlier `/setup` description (Step 2 — "tell me about your company and tech stack"), extract languages. Examples:

[...]

PR #210, opened 2026-05-11
feat(#208): /setup enables LSP by default with language-aware install

atlas-apex:
## Summary

- `/setup` now offers an LSP-enablement step (Step 2c) after the privacy / split-portfolio gate: detects language from the operator's `tech_stack` description, installs the matching server (`typescript-language-server` / `pyright` / `gopls` / `rust-analyzer`), writes `ENABLE_LSP_TOOL=1` to the operator's shell rc idempotently, and prints the Claude Code plugin marketplace install instruction.
- Machine-spec heuristic defaults to "skip" on constrained hardware (< 4 cores OR < 8 GB RAM); existing adopters can retrofit with `/setup --enable-lsp` without re-running the whole bootstrap; second invocations on an already-enabled machine are silent no-ops; Windows is explicitly out of scope for v1 and prints the manual-recipe pointer.
- `docs/getting-started.md` § "Optional: LSP-aware code navigation" now presents `/setup` as the easy path with the full manual recipe retained as the fallback.

## Testing

A reviewer can verify by running `/setup` (or `/setup --enable-lsp`) against a scratch fork:

- [ ] Run `/setup` on a scratch fork with `onboarding.yaml` containing the placeholder `"Your Company Name"`; describe a "TypeScript + React + Node" stack and accept the proposed config. Confirm Step 2c surfaces the install prompt, the heuristic default reflects machine specs, accepting `y` runs `npm install -g typescript typescript-language-server`, appends `export ENABLE_LSP_TOOL=1` to `~/.zshrc` (or matching rc), prints the manual plugin-marketplace instruction, and ends with the Step 2c.7 status line.
- [ ] Run `/setup --enable-lsp` on a fork that's already configured but has no LSP. Confirm Steps 2 / 2a / 2b / 3-7 are skipped and only Step 2c runs.
- [ ] Re-run `/setup --enable-lsp` immediately after the first successful run. Confirm Step 2c.2's idempotence check fires: prints "LSP already enabled" and exits without touching the rc file or re-running `npm install -g`.
- [ ] Run on a machine where the prerequisite runtime is missing (uninstall `npm` / use a host without Node). Confirm the skill refuses the language-server install gracefully ("npm not found on PATH. Install Node.js first…"), does NOT auto-install Node, and prints a status line that reflects the partial outcome.
- [ ] On macOS (no `/proc/meminfo`), confirm the RAM-detection fallback to `sysctl -n hw.memsize` works and the heuristic produces a sane default.
- [ ] On a contrived constrained machine (2 cores, 4 GB RAM via container), confirm the default flips to `n`.
- [ ] Run on a Windows host (or one that reports `MINGW` / `CYGWIN` via `uname -s`). Confirm the Windows branch fires, prints the manual recipe pointer, and continues to Step 3 without attempting any install.
- [ ] Stack the same skill on Python ("FastAPI + Postgres") — confirm language detection picks `python` and the install command becomes `pipx install pyright`. Same for Go ("Go + gopls") and Rust ("Rust + cargo").

## Glossary

| Term | Definition |
|------|------------|
| LSP | Language Server Protocol — a standard JSON-RPC wire format clients use to ask a language-aware server for "where is this defined?", "find references", etc. |
| `ENABLE_LSP_TOOL=1` | The environment variable Claude Code v2.0.74+ reads to enable its built-in LSP tool. Singular `_TOOL`, not plural. |
| Language server | The binary that answers LSP queries for one language. Examples: `typescript-language-server`, `pyright`, `gopls`, `rust-analyzer`. |
| Marketplace plugin | A Claude Code plugin that ships the `.lsp.json` wiring telling Claude Code how to start the language-server binary on a given file extension. |
| Bootstrap skill | A skill that runs before any project ticket can exist (`/setup`, `/handover`, `/update`, `/split-portfolio`); exempted from the `require-active-ticket.sh` gate via `.claude/session/active-bootstrap`. |
| Retrofit flag | `/setup --enable-lsp` — jumps straight to Step 2c on an already-configured fork, skipping Steps 2 / 2a / 2b / 3-7. |
| Idempotent | Re-running the same step produces the same end state without duplicate side effects (no double-appended rc lines, no re-install). |
| Machine-spec heuristic | A pre-prompt computation of cores + RAM that decides whether to default the LSP-enablement prompt to `y` or `n`. |

Closes me2resh/apexyard#208

### DIFF
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -2,7 +2,7 @@
 name: setup
 description: First-run framework bootstrap for a new ApexYard fork. Three exchanges — "describe your stack", "here are the defaults", "accept or customize?" — and the fork is configured. Run once after forking; re-run anytime to update.
 disable-model-invocation: false
-argument-hint: "[--reset]"
+argument-hint: "[--reset] [--enable-lsp]"
 effort: medium
 ---
 
@@ -46,11 +46,12 @@ See AgDR-0011 + me2resh/apexyard#150 for the design rationale.
 
 ### Step 1: Check current state
 
-Read `onboarding.yaml`. Two modes:
+Read `onboarding.yaml`. Four modes:
 
 - **First run** (placeholder values detected): proceed to Step 2.
 - **Already configured** (real values): show a summary of the current config and ask "What would you like to update?" — then jump to the specific section. Don't re-ask everything.
 - **`--reset` flag**: clear `onboarding.yaml` back to the template defaults (copy from the upstream example or regenerate) and proceed as first run.
+- **`--enable-lsp` flag** (retrofit mode): skip Steps 2 / 2a / 2b / 3-7 entirely and jump straight to Step 2c (LSP enablement). Use this when an existing adopter has a fully-configured fork and only wants to turn on LSP without re-running the whole bootstrap. The flag honours the same idempotence rules as a first-run pass through Step 2c — if LSP is already enabled, the step reports "already enabled" and exits cleanly. Step 0 (bootstrap marker) and Step 8 (clear marker) still run so the ticket gate stays coherent.
 
 Detection: `grep -q '"Your Company Name"' onboarding.yaml` — if found, it's still a template.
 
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #210

**Commit**: `63eb514bdb30098adf27cab8df5c84a4d9141c39`
**Base**: `dev` (correct — release-cut model)

### Summary

Extends `/setup` with Step 2c that installs the language server, writes `ENABLE_LSP_TOOL=1` to the operator shell rc idempotently, and prints a clear manual instruction for the Claude Code LSP plugin marketplace install. Adds a `--enable-lsp` retrofit flag for already-configured forks, a constrained-machine heuristic (< 4 cores OR < 8 GB RAM defaults to skip), and a Windows graceful-refusal branch. Doc update presents `/setup` as the easy path while preserving the manual fallback.

### Checklist Results

- Architecture & Design:    Pass (skill spec — clean layering, no new infra deps)
- Code Quality:              Pass (bash blocks are syntactically sound, idempotent, safe)
- Testing:                   Pass (Step 2c.6 smoke test + 8-item testing checklist in PR body)
- Security:                  Pass (no secrets, no auto-runtime-install, refuses missing prereqs)
- Performance:               Pass (heuristic prevents LSP-on-constrained-machine footgun)
- PR Description & Glossary: Pass (8-entry Glossary covers LSP, env var, language server, plugin, bootstrap skill, retrofit flag, idempotent, heuristic)
- Technical Decisions (AgDR): Pass (doc-spec change; plugin-install fallback decision documented inline; see suggestion below)

### AC verification (10/10 addressed)



| # | AC | Verified how |

|---|----|--------------|

| 1 | Step 2c after privacy gate | Read SKILL.md — inserted between 2b and 3 |

| 2 | 4 languages (TS, Python, Go, Rust) | Detection table in 2c.3 covers all four |

| 3 | macOS + Linux install, Windows refuses | macOS `sysctl` fallback in 2c.1; explicit Windows branch in 2c.5(a) |

| 4 | `ENABLE_LSP_TOOL=1` idempotent rc write | `grep -q` guard in 2c.5(c); safe on missing file via `2>/dev/null` + `-f` |

| 5 | Plugin install OR clear manual instruction | Manual instruction in 2c.5(d), non-fabrication principle (8) added |

| 6 | Smoke test confirms binary on PATH | 2c.6 `command -v $BIN` |

| 7 | Heuristic < 4 cores OR < 8 GB → skip | 2c.1, fail-safe on detection failure (echo 0 → skip) |

| 8 | Re-running is idempotent | 2c.2 detects env var + binary, prints "already enabled" |

| 9 | `/setup --enable-lsp` retrofit | argument-hint updated; Step 1 routes `--enable-lsp` straight to 2c |

| 10 | docs/getting-started.md points at `/setup` | New "Easy path" section above retained manual fallback |

### Spec correctness walkthrough



- **RAM-detection awk pipeline (2c.1)**: on macOS, the first awk against `/proc/meminfo` fails (file missing) → `||` triggers `sysctl -n hw.memsize` → fallback works. On Linux without sysctl, the same pipeline reaches `|| echo 0` → returns 0 → heuristic flips to skip (fail-safe). On a stripped container with neither source available, defaults to 0 cores / 0 GB → skip. All paths fail-safe.

- **rc-file mutation (2c.5(c))**: `grep -q ENABLE_LSP_TOOL=1 "$RC_FILE" 2>/dev/null` + grouped append with `>>` correctly handles both "file missing" and "pattern absent" cases without duplication. Second run is a clean no-op.

- **Graceful prereq refusal (2c.5(b))**: `command -v npm` gate, explicit "do NOT auto-install Node, Python, Go, or Rust runtimes" line plus framework principle 7. Author cannot accidentally regress.

- **Windows handling (2c.5(a))**: `$OSTYPE` + `uname -s` dual check covers msys/cygwin/MINGW/CYGWIN/Windows.

- **Plugin install (2c.5(d))**: conservative — prints manual instruction citing https://docs.claude.com/en/docs/claude-code/plugins instead of fabricating `claude plugin install`. Right call given the marketplace CLI shape is not stable; reinforced by framework principle 8.



1. **Consider an AgDR for the plugin-install fallback.** The choice to print a manual instruction rather than fabricate a CLI call is a real trade-off (loses one-shot UX, gains harness-version resilience). A short `AgDR-NNNN-lsp-plugin-install-manual-vs-scripted.md` would make it discoverable via `/agdr search plugin` when the marketplace CLI eventually stabilises.

2. **`--skip-lsp` flag.** The ticket Risks section mentions `/setup --skip-lsp` as an operator override; this PR does not add it, but the prompt "n" answer + the constrained-machine default cover the same need. Worth noting as a follow-up if a script-mode workflow needs non-interactive opt-out.

3. **Step 1 "four modes" wording.** Step 1 now lists four modes ("first run", "already configured", `--reset`, `--enable-lsp`). The fourth is a flag-controlled variant rather than a distinct mode — editorial nit.



### Verdict



**APPROVED**



The bash recipe is correct, idempotent, and safe; all 10 ACs are addressed; the Glossary is thorough; the doc update is accurate; base branch is right for the release-cut model. The plugin-install conservatism (manual instruction over fabricated CLI) is the right call and matches framework principle 8.



---

Reviewed commit: `63eb514bdb30098adf27cab8df5c84a4d9141c39`

PR #264, opened 2026-05-16
fix(#227): greedy body extractor — no more truncation at embedded quotes

### LINKED ISSUE #227: [Bug] validate-issue-structure body extractor truncates --body "..." at the first embedded double-quote
## Given / When / Then

**Given** an agent files an issue via:

```bash
gh issue create --repo owner/repo \
  --title "[Chore] …" \
  --body "$(cat <<'EOF'
## Driver
Some prose.

## Scope
- A bullet describing an admin notice ("Configure X to unlock Y")
- Another bullet mentioning the post is marked "needs retry"

## Acceptance Criteria
- [ ] Some criterion
- [ ] Another criterion
EOF
)"
```

…where the body contains one or more embedded `"` characters inside the value (admin-notice strings, status labels in quotes, prose like `"current state"` etc.).

**When** `validate-issue-structure.sh` runs its `extract_flag_value '--body|-b' "$COMMAND"` against the literal command string — using the awk regex `(--body|-b)[[:space:]]+"([^"]*)"` (hook file line 71) —

**Then** the extractor terminates the match at the FIRST embedded `"` inside the body value rather than at the closing `"` of the `--body` argument. The truncated body is missing any `##` heading that occurs past the truncation point. The hook then exits 2 with `missing section: ## Acceptance Criteria` (or whichever section follows the truncation), even though the section *is* present in the real body that `gh` will receive.

The error is a **false positive** — `gh` sees the full body (shell does the quoting correctly via the `$(cat <<'EOF' ... EOF)` substitution); only the hook's static-string regex sees the truncation.

## Repro

1. Use any apexyard fork at framework version that ships `.claude/hooks/validate-issue-structure.sh` (current `me2resh/apexyard:main`).
2. Run:

   ```bash
   gh issue create --repo owner/repo \
     --title "[Chore] Embedded-quote repro" \
     --body "$(cat <<'EOF'
   ## Driver
   Test of the body extractor.

   ## Scope
   - Mention an admin notice that says "do the thing"

   ## Acceptance Criteria
   - [ ] Whatever
   EOF
   )"
   ```

3. **Expected**: hook validates the body (which contains all three required sections), exits 0, `gh` files the issue.

4. **Actual**: hook exits 2 with:

   ```
   BLOCKED: issue body for '[Chore]' does not match the required schema.

   Problems:
     - missing section: ## Acceptance Criteria
   ```

   The `## Scope` heading already lives past the first embedded `"`, so the extractor returns a buffer that contains only `## Driver` plus a few words of `## Scope`. Sections after the first `"` are invisible to the hook.

### Workaround (current)

Use `--body-file <path>` instead of `--body "…"`. The hook's `--body-file` branch (lines 103–123) reads the file via `cat`, so the awk extraction issue doesn't apply. Validation still runs correctly against the file content. Confirmed working on a real 12-ticket batch.

### Affected files

- `.claude/hooks/validate-issue-structure.sh` (the primary site)
- `.claude/hooks/block-private-refs-in-public-repos.sh` — comment at line 49 of `validate-issue-structure.sh` says the extractor was "lifted from `block-private-refs-in-public-repos.sh`", so it is likely affected by the same truncation when scanning bodies that contain embedded `"`. Worth checking — a quote-truncated body could let private refs in the back half of the body slip past the leak gate.
- `.claude/hooks/verify-commit-refs.sh` and `validate-pr-create.sh` — same pattern, worth a quick audit even if they happen not to use the double-quoted shape today.

### Suggested fix direction

(Not prescriptive — leaving the implementation to the agent that picks this up.)

Three plausible approaches, increasing in ambition:

1. **Greedy double-quoted match**: change the regex from `"([^"]*)"` to a greedy form that anchors on the *last* `"` before a known terminator (whitespace + next `--flag` or end-of-string). Cheaper, still string-based, still wrong on adversarial inputs.
2. **Skip the static-string regex entirely**: rely on `--body-file` being the canonical path and prompt agents toward it in skill docs. Doesn't fix the case where the agent uses `--body` inline.
3. **Shell-aware parsing**: invoke `bash -c 'printf "%s\n" "$1" "$2" …'` style with the same argv `gh` would see, parsing argv server-side rather than the raw command string. Most robust, biggest change. Probably overkill for this hook.

### Related

- The hook's docstring (lines 36–37) lists "Section matching is case-insensitive, whitespace-tolerant" — the case-insensitive matching is fine; the problem is upstream of section matching, in body extraction.
- The hook is silent-success on `--body-file` (which is correct, no regression there).

### Repro confidence

High — hit on a 12-ticket batch this session, every ticket whose body contained an embedded `"` failed identically. Tickets whose bodies had no embedded `"` passed. Switched to `--body-file` and all subsequent tickets passed.

atlas-apex:
## Summary

The awk + sed body/title extractors in three hooks all used the same non-greedy pattern to pull `--body` / `-m` values out of the raw `gh` / `git commit` command. The pattern truncates at the first embedded double-quote inside the value, so any body that quoted prose (admin-notice strings, status labels, prose like `current state`) lost every section heading past the first internal quote — and the hooks falsely reported them as missing. The friend running ApexYard hit this on a 12-ticket batch; every ticket whose body contained an embedded quote failed identically with `missing section: ## Acceptance Criteria` (or whichever section followed the truncation).

Switch the body extractors to a greedy match anchored on the next flag boundary (whitespace + `--<letter>` for the `gh` hooks, `-<letter>` for the `git commit` hook) or end-of-string. Awk consumes the whole command as a single buffer, so multi-line HEREDOC bodies parse correctly too. Title extractor in `validate-pr-create.sh` stays non-greedy on purpose — titles are short, single-line, and when followed by a multi-line `--body`, greedy over-consumes (see commit 2 for the self-inflicted reproduction of that exact problem).

Hooks fixed (per the issue's Affected files section):

- `.claude/hooks/validate-issue-structure.sh` — primary site, the bug repro
- `.claude/hooks/block-private-refs-in-public-repos.sh` — security tail: private refs in the back half of a quote-truncated body could slip past the leak gate. A new test confirms a registered project name in the post-quote half of a body is now correctly detected.
- `.claude/hooks/verify-commit-refs.sh` — commit messages with embedded quotes had `Closes #N` references past the first quote go unverified
- `.claude/hooks/validate-pr-create.sh` — title extractor LEFT AS NON-GREEDY (rationale in commit 2's message); the file was still touched only to add a clarifying comment about the deliberate choice

## Testing

Repro from the issue body — `[Chore]` with `do the thing` in `## Scope` quoted prose, then `## Acceptance Criteria` after it. Pre-fix: exit 2, missing section. Post-fix: exit 0.

All 28 hook test files green. New cases added:

```
test_validate_issue_structure.sh     :: 24 (was 19, +5)
test_block_private_refs.sh           :: 13 (was 10, +3)
test_verify_commit_refs_upstream.sh  :: 12 (was 9,  +3)
```

The new cases cover Chore / Feature / Bug bodies with embedded quotes, a negative confirming missing-section detection still fires when quotes are present, leak detection in the back half of a quoted body, a clean-body no-false-positive case, and commit messages with embedded quotes + post-quote `Closes #N`.

## Out of scope

- Backslash-escaped quotes (`\"...\"`) are still not parsed as content. Shell-quoted heredoc bodies do not produce backslash escapes, so the workaround for the exotic case stays `--body-file`.
- Shell-aware argv parsing (option 3 in the issue) — overkill. Greedy-anchor (option 1) closes the reported failure mode with a small awk regex change per hook.

Closes #227

---

## Glossary

| Term | Definition |
|------|------------|
| Body extractor | The awk/sed snippet in each hook that pulls the value of a body/message/title flag out of the raw command string Claude is about to run via Bash. The hooks scan it before the command executes. |
| Greedy match | A regex engine's preference to match the LONGEST possible substring (POSIX awk's `.*` is greedy by default). The opposite of non-greedy / lazy. |
| Flag-boundary anchor | A terminator regex (whitespace + next-flag) that tells the greedy match where the FLAG argument legitimately ends — the NEXT CLI flag's leading dashes. Lets the greedy quantifier swallow everything in between, including embedded quotes. |
| Leak-protection tail | The corollary security consequence of the truncation bug on `block-private-refs-in-public-repos.sh`: private registered-project names in the BACK half of a quote-truncated body were invisible to the gate, so they could be filed on a public framework ticket undetected. |

### DIFF
--- a/.claude/hooks/block-private-refs-in-public-repos.sh
+++ b/.claude/hooks/block-private-refs-in-public-repos.sh
@@ -148,19 +148,54 @@ fi
 # ---------------------------------------------------------------------------
 
 extract_flag_value() {
-  # $1 = python-flag regex (e.g. --title | -t). Matches:
+  # $1 = python-flag regex (e.g. --title | -t). Matches (possibly multi-line):
   #   --title "value with spaces"
   #   --title 'value'
   #   --title value
+  #
+  # Quoted-value regex is GREEDY and anchored on the next flag boundary
+  # (whitespace + `--<letter>`) or end-of-string. The earlier sed form
+  # `[^"]*` truncated at the first embedded double quote
+  # (me2resh/apexyard#227); for the leak-protection hook that's a real
+  # security tail — a body with an embedded `"` could let private refs in
+  # the back half slip past the gate. awk + greedy + boundary anchor
+  # gives us multi-line consumption and correct termination in one shot.
   local flag_re="$1"
   local cmd="$2"
-  local v
-  v=$(echo "$cmd" | sed -nE "s/.*(${flag_re})[[:space:]]+\"([^\"]*)\".*/\2/p" | head -1)
-  if [ -n "$v" ]; then echo "$v"; return; fi
-  v=$(echo "$cmd" | sed -nE "s/.*(${flag_re})[[:space:]]+'([^']*)'.*/\2/p" | head -1)
-  if [ -n "$v" ]; then echo "$v"; return; fi
-  v=$(echo "$cmd" | sed -nE "s/.*(${flag_re})[[:space:]]+([^[:space:]]+).*/\2/p" | head -1)
-  echo "$v"
+  printf '%s' "$cmd" | awk -v FLAG_RE="$flag_re" -v SQ="'" '
+    { buf = (NR == 1 ? $0 : buf "\n" $0) }
+    END {
+      s = buf
+      # Double-quoted value: greedy `(.*)` anchored on next flag or EOS.
+      re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+\"", "", chunk)
+        sub("\"([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
+        sub("\"[[:space:]]*$", "", chunk)
+        print chunk
+        exit
+      }
+      # Single-quoted value: same greedy + anchor treatment.
+      re = "(" FLAG_RE ")[[:space:]]+" SQ "(.*)" SQ "([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+" SQ, "", chunk)
+        sub(SQ "([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
+        sub(SQ "[[:space:]]*$", "", chunk)
+        print chunk
+        exit
+      }
+      # Unquoted value: single token, embedded quotes irrelevant.
+      re = "(" FLAG_RE ")[[:space:]]+[^[:space:]]+"
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+", "", chunk)
+        print chunk
+        exit
+      }
+    }
+  '
 }
 
 TITLE=$(extract_flag_value '--title|-t' "$COMMAND")
--- a/.claude/hooks/tests/test_block_private_refs.sh
+++ b/.claude/hooks/tests/test_block_private_refs.sh
@@ -164,6 +164,51 @@ run_case "gh api issues leak" \
   2 "project name: curios-dog" \
   "gh api repos/me2resh/apexyard/issues -f title=bug -f body='discovered during curios-dog rebuild'"
 
+# 11. Embedded double quotes in body — me2resh/apexyard#227.
+#     Pre-227 the sed-based extractor's `[^"]*` truncated the body at the
+#     first internal `"`, which had a security tail: private refs that
+#     appeared in the BACK half of the body (past the truncation) slipped
+#     past the leak gate. Post-fix the awk extractor is greedy + anchored
+#     on next-flag-or-EOS, so a private project name in the back half is
+#     correctly detected.
+LEAK_AFTER_QUOTE_BODY='## Driver
+The "admin notice" string is shown to users.
+
+## Scope
+- Mention the "current state" label
+
+## Acceptance Criteria
+- [ ] discovered during curios-dog rebuild'
+
+run_case "leak in back half of body with embedded quote in front → block" \
+  2 "project name: curios-dog" \
+  "gh issue create --repo me2resh/apexyard --title 'fix' --body \"$LEAK_AFTER_QUOTE_BODY\""
+
+# 12. Same but for repo-slug ref past the embedded quote.
+LEAK_REPO_AFTER_QUOTE='Description of the issue.
+
+The "needs attention" flag is wrong.
+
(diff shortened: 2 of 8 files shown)


atlas-apex (commented):
## Code Review: PR #264

**Commit**: `caa52e81a8a6ddf56e48fc7bf28e5a90bb6442d8`

### Summary

Fixes the body-extractor truncation in `validate-issue-structure.sh` plus the two sibling hooks (`block-private-refs-in-public-repos.sh`, `verify-commit-refs.sh`) that copy-pasted the same non-greedy `"([^"]*)"` regex. Replaces it with a greedy `(.*)"` anchored on the next flag boundary (`[[:space:]]+--[a-zA-Z]` for the `gh` hooks; `[[:space:]]+-[a-zA-Z]` for the `git commit -m` hook) or end-of-string — letting embedded `"` in body prose pass through. Adds 11 new test cases across three test files.

### Checklist Results

- Architecture & Design:    Pass — same shape across all three hooks, awk + END-block accumulator, comment cross-references #227.
- Code Quality:              Pass — boundary-anchor terminator + cleanup `sub`s consistently applied; comment blocks explain the design choice and the deliberate non-fix in `validate-pr-create.sh`.
- Testing:                   Pass — 11 new cases cover (a) embedded `"` in Chore / Feature / Bug bodies, (b) multi-line bodies, (c) `--body-file` regression-tested, (d) negative case (missing-section still blocks even when quotes are present), (e) the leak-protection back-half (private ref past embedded `"` now caught), (f) commit-msg `Closes #N` past embedded `"` now extracted. 19 pre-existing tests on validate-issue-structure are untouched and should remain green.
- Security:                  Pass — the security tail flagged in the issue's "Affected files" section is genuinely closed: my probe confirms `gh issue create ... --body "...curios-dog rebuild"` with an embedded `"` in the front half now extracts the full body, so the leak-scrubber sees the back-half private ref. New test case 11 asserts this directly.
- Performance:               Pass — single awk invocation per extraction, no measurable cost change.
- PR Description & Glossary: Pass — explanation of greedy-anchor design, commit-2 deviation rationale, out-of-scope notes on `\"` and shell-aware parsing, and a complete glossary.
- Technical Decisions (AgDR):N/A — bug fix to existing extractor design, no new dependency/architecture/pattern choice; staying with greedy+anchor instead of escalating to shell-aware argv parsing is articulated in the PR body as "out of scope" with the rationale.

### Issues Found

None blocking.

### Suggestions

1. **Minor — test count drift in the PR body.** The body says "24 / 13 / 12" cases post-merge; actual `^run_case` count is 25 / 14 / 13 (the discrepancy is one `run_case()` function definition I'm counting as a hit in the awk-style grep). Not blocking — the +5 / +3 / +3 delta the body claims is correct.
2. **Minor — `validate-pr-create.sh` deviation is correctly justified** in the comment block (lines 30-38), but worth noting in a follow-up: titles with literal `'` (apostrophes in prose) still get truncated by `[^"']*`. Real-world frequency is low (PR titles almost never quote prose); flag for a future audit only.
3. **Nice-to-have — boundary anchor for `-S` / `-s` sign-flag** (`verify-commit-refs.sh`): the `-<letter>` anchor covers `-S`, `-s`, `-F`, `--`-prefixed long flags, etc. — confirmed working in a manual probe. The comment block (lines 68-71) already names this explicitly. Good.

### Verdict

**APPROVED** — clean targeted fix, security-tail concern genuinely closed, test coverage strong, deviation rationale sound. Ready to merge from the code-review side once CI confirms green and the human approver signs off.

---
Reviewed commit: caa52e81a8a6ddf56e48fc7bf28e5a90bb6442d8

PR #277, opened 2026-05-17
fix(#275): additive ui_paths_exclude carve-out for require-design-review-for-ui

### LINKED ISSUE #275: [Bug] require-design-review-for-ui.sh false-positives on non-UI .jsx files (no path scoping in default regex)
## Given / When / Then

**Given** an adopter's PR adds a non-UI \`.jsx\` file (e.g. a rule-artifact placeholder, a JSX-extension documentation example, or a JS module that happens to use the \`.jsx\` extension for editor highlighting),

**When** the adopter runs \`gh pr merge\`,

**Then** \`require-design-review-for-ui.sh\` blocks the merge demanding a design-review marker, because the default \`UI_GLOBS\` regex matches \`\\.jsx$\` against any file path — the pattern is unanchored to typical UI directories.

## Repro

1. Fork and clone \`me2resh/apexyard\`.
2. Create any \`.jsx\` file outside the conventional UI paths — e.g. \`docs/example.jsx\`, \`wiki/artifacts/placeholder.jsx\`, \`tools/rule-engine.jsx\`. Content irrelevant; even a one-line file with \`export const X = 1;\` reproduces it.
3. Open a PR, run Rex (\`/code-review\`), get approval. Get CEO approval (\`/approve-merge <N>\`).
4. Attempt \`gh pr merge <N> --squash\`.
5. Observe: design-review gate blocks despite the file containing no JSX/React syntax and no user-facing surface.

## Severity

Low-to-medium. Workaround is documented (the hook itself describes the \"touch the marker manually as a deliberate-skip artifact\" escape hatch), but adopters using \`.jsx\` for non-UI purposes hit this on every such merge, which dilutes the design-review gate's signal-to-noise ratio.

## Root cause

\`.claude/hooks/require-design-review-for-ui.sh\` defines \`UI_GLOBS\` as:

\`\`\`
\\.tsx$
\\.jsx$
\\.vue$
\\.svelte$
\\.css$
\\.scss$
\\.sass$
\\.less$
design-tokens
\`\`\`

The \`\\.jsx$\` (and \`\\.tsx$\`, \`\\.css$\`, etc.) entries match the file *suffix only*. The hook applies \`grep -E\` against the full changed-file path, so any \`.jsx\` file anywhere in the tree triggers the gate.

The override mechanism (\`.ui_paths\` in \`.claude/project-config.json\`) is *replace-only*: setting it overrides the entire default array. There is no additive / exclude semantic, so an adopter wanting to keep most of the default behavior but exclude one specific path has to copy the whole default list and edit it — fragile, drifts from upstream over time.

### Option 1 — tighten the default to scope by directory

Change \`UI_GLOBS\` so file-extension patterns are anchored to conventional UI directories:

\`\`\`
^src/.*\\.(tsx|jsx|vue|svelte)$
^components/.*\\.(tsx|jsx|vue|svelte)$
^pages/.*\\.(tsx|jsx)$
^app/.*\\.(tsx|jsx)$
^themes/.*\\.(tsx|jsx|php|html)$
^assets/.*\\.(css|scss|sass|less)$
^styles/.*\\.(css|scss|sass|less)$
design-tokens
\`\`\`

Pro: zero adopter action needed for the common case (UI lives in conventional dirs).
Con: projects with unconventional layouts may now silently skip — a regression risk that needs a release note.

### Option 2 — add \`ui_paths_exclude\` for surgical exclusion

Extend the hook to read a second array, \`.ui_paths_exclude\`, and remove any file that matches an exclude pattern from the UI-touched set:

\`\`\`bash
EXCLUDE=$(jq -r '.ui_paths_exclude // [] | join(\"|\")' \"\${REPO_ROOT}/.claude/project-config.json\" 2>/dev/null)
if [ -n \"\$EXCLUDE\" ] && [ \"\$EXCLUDE\" != \"null\" ]; then
  TOUCHED_UI=$(echo \"\$TOUCHED_UI\" | grep -vE \"\$EXCLUDE\" || true)
fi
\`\`\`

Pro: keeps default broad, lets adopters surgically exclude (e.g. \`^wiki/artifacts/\`, \`^docs/examples/\`). Additive override is more maintainable than the current replace-only \`ui_paths\` (which would also benefit from the same pattern).
Con: more config surface to document.

Option 2 is the minimally-disruptive shape and matches how \`migration_paths\` is handled in \`require-migration-ticket.sh\` (additive override pattern). Option 1 fixes the immediate false-positive class and is independently valuable.

## Acceptance Criteria

- [ ] Default \`UI_GLOBS\` no longer matches \`.jsx\`/\`.tsx\` files outside conventional UI directories, OR a \`ui_paths_exclude\` config key is supported
- [ ] If Option 1: release note flagging the behavior change for adopters with unconventional layouts
- [ ] If Option 2: doc update in the hook's own header comment + \`pr-quality.md\` documenting the new key
- [ ] Smoke test: a \`.jsx\` file outside \`src/\`/\`components/\`/etc. does not trigger the gate, while a real \`src/Component.tsx\` still does

atlas-apex:
## Summary

Fixes the false-positive class flagged in #275 — `require-design-review-for-ui.sh` was blocking merges on any `.jsx`/`.tsx`/`.css` file anywhere in the tree (e.g. `docs/example.jsx`, `wiki/artifacts/placeholder.jsx`, `tools/rule-engine.jsx`) because the default `UI_GLOBS` regex is suffix-only and unanchored to conventional UI dirs.

Picked **Option 2 from the ticket** (additive `ui_paths_exclude`) over Option 1 (tighten defaults to anchor on `^src/`, `^components/`, etc.) for three reasons stated in the issue:

1. **Zero regression risk.** Adopters with unconventional layouts (e.g. UI in `web/`, `frontend/`, `themes/`) keep working unchanged.
2. **Mirrors an existing precedent.** `require-migration-ticket.sh` already supports the same additive-override shape — adopters recognise the pattern.
3. **Operator owns the carve-out.** Each adopter has a different "this dir is not UI" list; centralising it in framework defaults would be premature.

## What changed

| File | Change |
|------|--------|
| `.claude/hooks/require-design-review-for-ui.sh` | Reads `.ui_paths_exclude` from project-config; removes matching files from `TOUCHED_UI` after the existing `UI_GLOBS` matching pass. Header comment updated to document both keys. Error message updated to point at the new key. |
| `.claude/hooks/tests/test_ui_paths_exclude.sh` (new) | 7-case suite — the reported false-positive (docs/wiki carve-out), real UI without exclude, mixed lists, empty input, missing key, OR-of-patterns, non-matching exclude. |
| `.claude/rules/pr-quality.md` | Design Review section gains a table comparing `.ui_paths` (replace) vs `.ui_paths_exclude` (additive) with "when to use each" guidance. |

## Usage (adopter docs)

```json
// .claude/project-config.json
{
  "ui_paths_exclude": [
    "^docs/examples/",
    "^wiki/artifacts/",
    "^tools/"
  ]
}
```

After this lands, a `.jsx` file under any of those paths no longer trips the design-review gate, while `src/Component.tsx` (or any other path) still does.

## Tests

```
$ bash .claude/hooks/tests/test_ui_paths_exclude.sh
... PASS: 8   FAIL: 0
```

The first case directly reproduces the scenario from #275's repro — `docs/example.jsx` + `wiki/artifacts/placeholder.jsx` carved out, `src/Button.tsx` still gated.

## Testing

- [x] All 8 unit tests pass
- [x] Hook syntax-check passes (`bash -n`)
- [x] No regression on the existing `.ui_paths` replace-only behaviour (covered by the "no exclude" test case)
- [ ] Operator verification — set `ui_paths_exclude` in a real adopter's project-config and merge a PR with both an excluded `.jsx` and a non-excluded `.tsx` to confirm the carve-out

## Out of scope

- **Option 1 (tighten defaults)** — deliberately deferred. The reporter and I both lean Option 2 as the minimally-disruptive shape; Option 1 can land later if adopters consistently report the false-positive class even with `ui_paths_exclude` available.

Closes #275.

## Glossary

| Term | Definition |
|------|------------|
| UI_GLOBS | The default regex list in the hook for what counts as "user-facing UI" (`.tsx`, `.jsx`, `.vue`, `.svelte`, `.css`, etc.) |
| `ui_paths` (replace) | Existing config key — replaces UI_GLOBS entirely with the adopter's list |
| `ui_paths_exclude` (additive, this PR) | New config key — paths matching any pattern are removed from the touched-UI set after UI_GLOBS matching |
| TOUCHED_UI | The set of files in a PR's diff that match UI_GLOBS — the hook blocks the merge if this set is non-empty AND no design marker exists |
| Carve-out | Excluding specific paths from a broad rule rather than narrowing the rule itself; preserves "default is right for most" while letting operators tune the edges |

### DIFF
--- a/.claude/hooks/require-design-review-for-ui.sh
+++ b/.claude/hooks/require-design-review-for-ui.sh
@@ -19,7 +19,15 @@
 #   - design-tokens.* (design systems)
 #
 # Projects that want a broader/narrower list can override via
-# .claude/project-config.json `.ui_paths` (JSON array of regex patterns).
+# .claude/project-config.json:
+#   `.ui_paths`         — REPLACE the default UI_GLOBS entirely (JSON array of regex patterns)
+#   `.ui_paths_exclude` — ADDITIVE: paths matching any pattern here are removed
+#                         from the touched-UI set AFTER UI_GLOBS matching. Mirrors
+#                         the `migration_paths`-exclude precedent (#275).
+#
+# Use `ui_paths_exclude` when you want to keep the default broad matching but
+# carve out a specific dir (e.g. `^docs/examples/`, `^wiki/artifacts/`) where
+# `.jsx`/`.tsx` files are documentation samples rather than real UI.
 #
 # How the marker gets written: the design-reviewer records approval by
 # writing the marker file. There is no /approve-design skill yet — the
@@ -116,6 +124,23 @@ while IFS= read -r FILE; do
   done <<< "$UI_GLOBS"
 done <<< "$CHANGED"
 
+# Apply `.ui_paths_exclude` — additive override that REMOVES paths from the
+# touched-UI set even when UI_GLOBS matched. Lets adopters keep the broad
+# defaults while carving out specific directories (e.g. `^docs/examples/`,
+# `^wiki/artifacts/`) where `.jsx`/`.tsx` files are doc samples not UI (#275).
+if [ -n "$REPO_ROOT" ] && [ -f "${REPO_ROOT}/.claude/project-config.json" ]; then
+  EXCLUDE=$(jq -r '.ui_paths_exclude // [] | join("|")' "${REPO_ROOT}/.claude/project-config.json" 2>/dev/null)
+  if [ -n "$EXCLUDE" ] && [ "$EXCLUDE" != "null" ] && [ -n "$TOUCHED_UI" ]; then
+    FILTERED=""
+    for FILE in $TOUCHED_UI; do
+      if ! echo "$FILE" | grep -qE "$EXCLUDE"; then
+        FILTERED="${FILTERED}${FILE} "
+      fi
+    done
+    TOUCHED_UI="$FILTERED"
+  fi
+fi
+
 if [ -z "$TOUCHED_UI" ]; then
   # Not a UI PR — nothing to enforce, merge-gate will continue
   exit 0
@@ -147,8 +172,15 @@ To unblock:
        git rev-parse HEAD > .claude/session/reviews/${PR_NUMBER}-design.approved
   3. Retry the merge
 
-To customize which file patterns count as "UI", set
-\`.ui_paths\` in .claude/project-config.json (JSON array of regex patterns).
+To customize which file patterns count as "UI":
+
+  \`.ui_paths\`         — REPLACE the default list entirely (JSON array of regex)
+  \`.ui_paths_exclude\` — ADDITIVE carve-out: keep the broad defaults but skip
+                          specific dirs (e.g. ["^docs/examples/", "^wiki/"]).
+                          Useful when .jsx/.tsx files are doc samples not UI
+                          (#275).
+
+Both keys live in .claude/project-config.json.
 
 For projects that deliberately ship UI without design review (e.g. admin tools,
 internal dashboards), touch the marker file manually — that's a visible,
--- a/.claude/hooks/tests/test_ui_paths_exclude.sh
+++ b/.claude/hooks/tests/test_ui_paths_exclude.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+# Tests for the `.ui_paths_exclude` additive override on
+# require-design-review-for-ui.sh (me2resh/apexyard#275).
+#
+# Scope: the EXCLUDE-application logic only. We test the in-script
+# pattern-filter directly via a synthetic TOUCHED_UI list + a sandbox
+# project-config.json. We do not exercise the gh-pr-diff path (that's
+# covered by the hook's existing integration shape).
+#
+# Mirrors the assert shape of test_require_skill_for_issue_create.sh.
+
+set -u
+
+SRC_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+HOOK_SRC="$SRC_ROOT/.claude/hooks/require-design-review-for-ui.sh"
+
+if [ ! -f "$HOOK_SRC" ]; then
+  echo "FAIL: hook missing: $HOOK_SRC" >&2
+  exit 1
+fi
+
+PASS=0
+FAIL=0
+
+assert_eq() {
+  local label="$1" want="$2" got="$3"
+  if [ "$want" = "$got" ]; then
+    echo "PASS [$label]"
+    PASS=$((PASS + 1))
+  else
+    echo "FAIL [$label]: want '$want', got '$got'" >&2
+    FAIL=$((FAIL + 1))
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# Helper — replay the EXCLUDE logic inline against a synthetic input + a
+# sandbox project-config.json. Keeps tests self-contained.
+# ---------------------------------------------------------------------------
+run_exclude_filter() {
+  local touched="$1"        # space-separated TOUCHED_UI string
+  local exclude_json="$2"   # JSON array literal for .ui_paths_exclude
+  local sb cfg
+  sb=$(mktemp -d)
+  cfg="$sb/.claude/project-config.json"
+  mkdir -p "$(dirname "$cfg")"
+  printf '%s\n' "$exclude_json" > "$cfg"
+
+  # Inline replay of the hook's EXCLUDE block.
+  local REPO_ROOT="$sb"
+  local TOUCHED_UI="$touched"
+  local EXCLUDE
+  EXCLUDE=$(jq -r '.ui_paths_exclude // [] | join("|")' "${REPO_ROOT}/.claude/project-config.json" 2>/dev/null)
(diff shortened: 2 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #277

**Commit**: `b7809e08bd9d8c331d41fd922159425912d39769`

### Summary

Surgical, well-scoped bug fix for the false-positive class flagged in #275. Adds an additive `.ui_paths_exclude` config key to `require-design-review-for-ui.sh`. A new EXCLUDE filter pass runs after the existing UI_GLOBS matching loop and removes matching paths from `TOUCHED_UI`. The block is gated on `.ui_paths_exclude` being non-empty in `project-config.json`, so adopters who don't set the key see exactly zero behaviour change. Three files: the hook itself (+35/-3), a dedicated test suite (+159 new), a docs table in `pr-quality.md` (+9).

### Checklist Results

- Architecture & Design:    Pass — additive override pattern, EXCLUDE block placed correctly after the UI_GLOBS matching loop and before the empty-set short-circuit
- Code Quality:              Pass — clean shell, follows the file's existing style, jq null-guard + `[ -n "$EXCLUDE" ]` + `[ "$EXCLUDE" != "null" ]` triple-guard matches the `.ui_paths` precedent at line 101–106
- Testing:                   Pass — 8 PASS / 0 FAIL locally. Every branch in the new EXCLUDE block is exercised: null/missing key (case 5), empty TOUCHED_UI (case 4), match (cases 1, 3, 6), non-match (case 7), order preservation (case 1 bonus)
- Security:                  Pass — no new attack surface, no secret handling, jq input is local file
- Performance:               Pass — O(n×m) over already-filtered TOUCHED_UI (small set) × pattern count (small); negligible
- PR Description & Glossary: Pass — 5-row glossary table, clear summary, "out of scope" section, Option 1 vs Option 2 rationale documented, repro scenario named
- Technical Decisions (AgDR):N/A — bug fix to an existing hook against a documented option from the ticket; no new architecture
- Adopter Handbooks:         Pass — `clean-architecture-layers.md` (not applicable, infra change); `migration-safety.md` blocking (not applicable, no migrations); `commit-message-quality.md` advisory — one nit below

### Issues Found

None blocking.

### Handbook Findings

**Commit Message Quality (advisory)** — `handbooks/general/commit-message-quality.md`

- `b7809e0` commit body — line "Mirrors the migration_paths-exclude precedent" is factually inaccurate: `require-migration-ticket.sh` does not currently expose a `migration_paths_exclude` key (verified by grepping `_exclude|paths_exclude` across `.claude/hooks/`). This PR is establishing the pattern, not mirroring one. Same line appears in the PR body's bullet 2. Suggest reframing to something like "Introduces the additive-exclude shape that future hooks (e.g. migration_paths) can adopt." Not blocking — the rest of the rationale (zero regression, operator-owned carve-out) stands on its own.

### Suggestions

1. **Doc nit (non-blocking)**: in `pr-quality.md` the new table reads "removed from the touched-UI set *after* `.ui_paths` matching". Slightly clearer: "after UI_GLOBS matching (whether default or `.ui_paths`-overridden)". This is what the code actually does and removes a small ambiguity.

2. **Future-proofing (defer to a follow-up)**: the EXCLUDE block reads `.ui_paths_exclude` from `REPO_ROOT/.claude/project-config.json` (correct — project-config is per-project not per-ops-fork). Worth a one-line code comment explaining the REPO_ROOT vs MARKER_HOME asymmetry for future readers, since it's a deliberate choice not an oversight.

3. **Spec hygiene**: the PR title says "8 cases" but the test file's top comment still says "7 scenarios" (line-2 docstring). The 8th case is the explicit order-preservation re-assert at line 84–92. Minor — feel free to fix in a follow-up commit or leave as-is.

### Verdict

**APPROVED**

The EXCLUDE block placement is correct: it runs after the UI_GLOBS matching loop completes (line 116–125) and before the "if `TOUCHED_UI` is empty, exit 0" short-circuit (line 144), so the filter sees the post-match set and either preserves it intact or shrinks it. The triple-guard on `EXCLUDE` (non-empty, not literal "null", and `TOUCHED_UI` non-empty) prevents both jq weirdness and empty-iteration false positives. The fall-through behaviour when the key is missing or empty is verified by test cases 4, 5, 7. The reported false-positive (`docs/example.jsx` + `wiki/artifacts/placeholder.jsx` carved out, `src/Button.tsx` retained) is the explicit first test case.

REPO_ROOT vs MARKER_HOME: the EXCLUDE block correctly uses REPO_ROOT (project-config lives with the project, not the ops fork), matching the existing `.ui_paths` resolution at line 101. The marker write at line 151 still correctly uses MARKER_HOME (ops-fork-scoped). Two different reads of two different files at two different scopes — both correct.

CI all green: Verify Ticket ID, lychee, markdownlint-cli2, shellcheck. No leaks of private project names. PR body has the required Glossary, links the ticket via `Closes #275`, distinguishes Option 1 vs Option 2 with reasoning.

Ship it.

---
Reviewed commit: `b7809e08bd9d8c331d41fd922159425912d39769`

Issue #242, opened 2026-05-15
[Feature] Split-portfolio v2: workspace + onboarding move to private repo, plus /update migration


atlas-apex:
## User Story

As a split-portfolio adopter on GitHub Free who values the public/private separation, I want EVERY adopter-specific artefact to live in the private sibling repo — not just the registry + `projects/` but ALSO `onboarding.yaml` (company info / tech stack / team) and `workspace/<name>/` (cloned managed projects) — so that my public fork holds only stock framework files + my customisations to skills/hooks/rules. And as an EXISTING split-portfolio adopter, I want `/update` to migrate my layout for me, not require a hand-edit dance.

### Architecture: extended portfolio config block

- [ ] `.claude/project-config.json` → `portfolio` block accepts two new keys:
  - `onboarding` — path to `onboarding.yaml` in the private repo (default `../<fork>-portfolio/onboarding.yaml`)
  - `workspace_dir` — path to the workspace dir in the private repo (default `../<fork>-portfolio/workspace`)
- [ ] `_lib-portfolio-paths.sh` exposes `portfolio_onboarding_path` and `portfolio_workspace_dir` helpers
- [ ] All framework code that reads `onboarding.yaml` or writes to `workspace/<name>/` resolves through the helpers (audit the codebase: `/setup`, `/handover`, `/status`, `/inbox`, `/projects`, `/tasks`, `onboarding-check.sh`, ops-root detection)

### Ops-fork detection survives the move

- [ ] `_lib-ops-root.sh` (and the existing inline walk in 6+ hooks) currently requires BOTH `onboarding.yaml` AND `apexyard.projects.yaml` to be present at the candidate dir. After the move, neither file is in the public fork. Walk-up needs a new public-fork anchor (proposal: marker file `.apexyard-fork` that the framework writes at first `/setup`)
- [ ] All hooks that walk for ops-root work in both old layout (legacy) AND new layout (post-migration)

### Single-fork mode unchanged

- [ ] Single-fork adopters (no `portfolio` block) see ZERO behaviour change. `onboarding.yaml` and `workspace/` stay in the fork. No migration prompted.

### Migration via `/update`

- [ ] `/update` detects split-portfolio adopters whose layout is pre-v2 (`onboarding.yaml` + `workspace/` in the public fork)
- [ ] Migration step OFFERS (default-yes) to move both for the operator, with explicit confirmation per file class:
  - Move `onboarding.yaml` → private repo, gitignore in public fork
[...]

PR #306, opened 2026-05-19
refactor(#302): sweep all 40 hook wrappers to v2-anchor-aware shape

### LINKED ISSUE #302: [Refactor] Sweep SessionStart hooks to use _lib-ops-root.sh — fix split-portfolio v2 invisibility
## Driver

Surfaced 2026-05-19 during Rex review of PR #300 (the `check-jq-installed.sh` SessionStart hook for #280): the new hook uses the legacy `onboarding.yaml`-only walk-up to find the ops-fork root. That walk works for single-fork adopters (default) and for split-portfolio v1, but **silently fails for split-portfolio v2** — under v2, `onboarding.yaml` lives in the private sibling repo, the public fork is anchored only by `.apexyard-fork`, and the walk-up condition fails. v2 adopters won't see the banner.

Audit of `.claude/settings.json` shows this isn't unique to `check-jq-installed.sh` — **every** SessionStart entry except `link-custom-skills.sh` uses the legacy `onboarding.yaml`-only walk-up shape:

```bash
r=$PWD; while [ ! -f "$r/onboarding.yaml" ] && [ "$r" != / ]; do r=${r%/*}; done; exec "$r/.claude/hooks/<name>.sh"
```

This is **legacy v1 behaviour** that pre-dates the v2 anchor (`.apexyard-fork`) added in framework #242. The correct shape is to source `_lib-ops-root.sh` and call `resolve_ops_root`, which recognises both anchors (v2 marker first, legacy v1 pair as fallback).

## Scope

Sweep every entry in `.claude/settings.json` under `hooks.SessionStart` (and any PreToolUse/PostToolUse entries that also do inline walk-up). Replace the inline walk-up shell with:

```bash
. "$(dirname "$0")/_lib-ops-root.sh" 2>/dev/null && \
  r=$(resolve_ops_root "$PWD") && \
  exec "$r/.claude/hooks/<name>.sh"
```

Or — if the hook itself uses `_lib-ops-root.sh` internally — drop the inline walk-up entirely and let the hook resolve from `$PWD`.

Audit the actual hook scripts too: any hook that does its own walk-up for ops-fork detection should source `_lib-ops-root.sh` instead of inlining the legacy two-file check.

## Acceptance Criteria

- [ ] Every SessionStart entry in `.claude/settings.json` resolves the ops-fork root via `_lib-ops-root.sh` (recognises both `.apexyard-fork` and the legacy v1 pair)
- [ ] Same sweep for any PreToolUse / PostToolUse entries that inline the walk-up
- [ ] Audit hook scripts for inline walk-ups (`while [ ! -f "$r/onboarding.yaml" ]`); replace with `_lib-ops-root.sh` usage
- [ ] Split-portfolio v2 adopters get the same banners / hooks as single-fork adopters — verified manually by setting up a v2 layout locally
- [ ] No regression for single-fork or split-portfolio v1 adopters
- [ ] AgDR documents the sweep decision + the v2-aware shape as the canonical pattern for future SessionStart entries

## Out of Scope

- Adding new SessionStart hooks (this is a refactor of existing ones).
- Changing the v1 fallback semantics — `_lib-ops-root.sh` already handles it.
- A meta-hook that lints inline walk-up shells. The audit + sweep is one-time; future hooks will be caught in code review against the AgDR.

## Risks / Dependencies

- **Risk**: subtle behavioural change for adopters who somehow relied on the legacy walk-up silently failing on v2. Mitigation: the v2-aware walk-up is a strict superset — v1 conditions still match, plus v2 anchors. No legitimate adopter loses any functionality.
- **Risk**: `_lib-ops-root.sh` itself has a bug that propagates across all SessionStart hooks. Mitigation: it's already in use by `link-custom-skills.sh` and the merge-gate hooks; well-tested. If anything, this sweep increases the bug surface area for that lib, making bugs in it land faster.

## Glossary

| Term | Definition |
|------|------------|
| Split-portfolio v2 | Framework layout where `onboarding.yaml` + `apexyard.projects.yaml` + `projects/` live in a private sibling repo (framework #242), with the public fork anchored only by `.apexyard-fork` |
| Walk-up | The pattern of walking from `$PWD` toward `/` looking for a marker file that identifies the ops-fork root |
| `_lib-ops-root.sh` | The shared library function `resolve_ops_root <start_dir>` that recognises both the v2 marker (`.apexyard-fork`) and the legacy v1 pair (`onboarding.yaml + apexyard.projects.yaml`) |
| Ops-fork root | The top of the apexyard fork tree — where session state lives, where hooks resolve overrides from, and where v2's `.apexyard-fork` marker sits |

Surfaced during Rex review of PR #300 on 2026-05-19.

atlas-apex:
## Summary

- **Swept all 40 hook wrappers in `.claude/settings.json`** to the canonical v2-anchor-aware shape — every wrapper now finds the ops fork via either `.apexyard-fork` (v2) OR `onboarding.yaml` (legacy v1 fallback), with the empty-`r` guard that legacy shape lacked.
- **Refactored `.claude/hooks/link-custom-skills.sh`** to source `_lib-ops-root.sh` with graceful-degradation fallback, replacing its private inline walk-up. One fewer place that needs to change when the anchor convention evolves.
- **Documented the canonical wrapper shape** in `AgDR-0041` so future hook authors copy from the right template — including a clear options-table entry on why the seemingly-cleaner "source the lib from the wrapper" path (Option A) doesn't actually work (`bash -c '…'` invocations set `$0=bash`, so `dirname "$0"` returns `.`).
- **Cross-linked from `docs/multi-project.md`** § "Where session-state files live" so adopters reading the operator docs land on the same explanation.

40 wrappers swept: 7 SessionStart (including the newly-shipped `check-jq-installed.sh` SessionStart entry that landed on `origin/dev` between worktree creation and merge), 30 PreToolUse, 2 PostToolUse, 1 UserPromptSubmit.

## Why

The v2 split-portfolio layout (introduced in `me2resh/apexyard#242`) moved `onboarding.yaml` to the private sibling repo and added `.apexyard-fork` as the new presence-only marker for the public fork. `_lib-ops-root.sh` was updated then, but the inline walk-up boilerplate copy-pasted across all 40 `settings.json` hook wrappers wasn't — they still look for `onboarding.yaml` AND `apexyard.projects.yaml` as the legacy v1 anchor pair. The hooks themselves work fine (most source `_lib-ops-root.sh` internally), but the wrapper that *finds* the hooks file fails on a v2-only fork — the walk never terminates because neither file exists. #302 closes that gap.

The legacy wrapper shape also has an `r != ""` infinite-loop hole (when `${r%/*}` reduces to empty before hitting `/`); the canonical shape fixes that too via the explicit `[ -n "$r" ]` guard.

## Testing

- [x] `jq . .claude/settings.json` clean (validates JSON after the sweep)
- [x] Simulated v2-only fork layout (only `.apexyard-fork` present, no legacy pair) — walk finds the right dir on every wrapper
- [x] Simulated no-anchors layout — empty-`r` guard exits cleanly (legacy shape would infinite-loop; confirmed 50+ iterations against the old shape)
- [x] `test_ops_root.sh` 8/8
- [x] `test_check_jq_installed.sh` 5/5
- [x] `test_link_custom_skills.sh` 7/7 (case 7 added in fix-up — exercises the graceful-degradation fallback when `_lib-ops-root.sh` is absent; cases 1–6 now exercise the primary lib-sourcing path that the refactor introduces)
- [x] `test_portfolio_paths.sh` 38/38
- [x] `test_split_portfolio_v2_migration.sh` 13/13
- [x] `test_check_upstream_drift.sh` 5/5

## Out of scope

- **Migrating every hook script** to source `_lib-ops-root.sh` for its own internal walks. Many already do (~10); the rest do private inline walks that are correct (they read `$PWD` at hook-fire time, which is what the hook semantics expect). Sweeping those is a separate ticket if we want it.
- **Eliminating the wrapper-level walk entirely** by making `claude` resolve hook paths from the ops-fork root. That would need an upstream change to the Claude Code runtime, not a framework change.

## Glossary

| Term | Definition |
|------|------------|
| v2 anchor | The `.apexyard-fork` marker file at the ops-fork root (introduced in #242 for split-portfolio v2). Presence-only — content is ignored. |
| v1 anchor | The legacy pair `onboarding.yaml` + `apexyard.projects.yaml`. Still recognised as fallback for un-migrated forks during the transition window. |
| Wrapper walk | The bash one-liner inside each `settings.json` hook entry that walks up from `$PWD` to find the ops fork, then `exec`s the actual hook script. |
| Canonical shape | The standardised wrapper one-liner documented in AgDR-0041 — every new SessionStart / PreToolUse / PostToolUse hook should copy this verbatim. |

Closes #302

### DIFF
--- a/.claude/hooks/link-custom-skills.sh
+++ b/.claude/hooks/link-custom-skills.sh
@@ -48,21 +48,28 @@ if [ "$IS_WINDOWS" -eq 0 ]; then
   esac
 fi
 
-# Resolve the ops-fork root by walking up from $PWD. Cheaper than the full
-# helper for the no-op cases (no helper sourced, no jq fork). The helper
-# walks the same path; we replicate that minimal logic here so the hook
-# stays useful even if the lib hasn't been installed yet on a fresh fork.
+# Resolve the ops-fork root via the shared helper (recognises both the v2
+# `.apexyard-fork` marker AND the legacy v1 anchor). Falls back to a small
+# inline walk-up with the same conditions if the helper is missing — same
+# graceful-degradation shape as check-jq-installed.sh / clear-bootstrap-marker.sh.
+HOOK_DIR="$(cd "$(dirname "$0")" && pwd)"
 ops_root=""
-cur="$PWD"
-while [ -n "$cur" ] && [ "$cur" != "/" ]; do
-  if [ -f "$cur/.apexyard-fork" ]; then
-    ops_root="$cur"; break
-  fi
-  if [ -f "$cur/onboarding.yaml" ] && [ -f "$cur/apexyard.projects.yaml" ]; then
-    ops_root="$cur"; break
-  fi
-  cur=$(dirname "$cur")
-done
+if [ -f "$HOOK_DIR/_lib-ops-root.sh" ]; then
+  # shellcheck source=/dev/null
+  . "$HOOK_DIR/_lib-ops-root.sh"
+  ops_root=$(resolve_ops_root "$PWD")
+else
+  cur="$PWD"
+  while [ -n "$cur" ] && [ "$cur" != "/" ]; do
+    if [ -f "$cur/.apexyard-fork" ]; then
+      ops_root="$cur"; break
+    fi
+    if [ -f "$cur/onboarding.yaml" ] && [ -f "$cur/apexyard.projects.yaml" ]; then
+      ops_root="$cur"; break
+    fi
+    cur=$(dirname "$cur")
+  done
+fi
 
 # Outside any apexyard fork → nothing to do.
 [ -z "$ops_root" ] && exit 0
--- a/.claude/hooks/tests/test_link_custom_skills.sh
+++ b/.claude/hooks/tests/test_link_custom_skills.sh
@@ -21,13 +21,14 @@ set -u
 HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/link-custom-skills.sh"
 LIB_PORTFOLIO_SRC="$(cd "$(dirname "$0")/.." && pwd)/_lib-portfolio-paths.sh"
 LIB_CONFIG_SRC="$(cd "$(dirname "$0")/.." && pwd)/_lib-read-config.sh"
+LIB_OPS_ROOT_SRC="$(cd "$(dirname "$0")/.." && pwd)/_lib-ops-root.sh"
 DEFAULTS_SRC="$(cd "$(dirname "$0")/../.." && pwd)/project-config.defaults.json"
 
 if [ ! -f "$HOOK_SRC" ]; then
   echo "FAIL: hook not found at $HOOK_SRC" >&2
   exit 1
 fi
-if [ ! -f "$LIB_PORTFOLIO_SRC" ] || [ ! -f "$LIB_CONFIG_SRC" ] || [ ! -f "$DEFAULTS_SRC" ]; then
+if [ ! -f "$LIB_PORTFOLIO_SRC" ] || [ ! -f "$LIB_CONFIG_SRC" ] || [ ! -f "$LIB_OPS_ROOT_SRC" ] || [ ! -f "$DEFAULTS_SRC" ]; then
   echo "FAIL: prerequisite libs / defaults missing" >&2
   exit 1
 fi
@@ -37,11 +38,19 @@ FAIL=0
 FAILED_CASES=""
 
 # --------------------------------------------------------------------------
-# make_fork: build an isolated apexyard fork sandbox with the v2 marker,
-# the hook script, the portfolio + config libs, and the defaults file.
+# make_fork [--no-ops-root-lib]: build an isolated apexyard fork sandbox
+# with the v2 marker, the hook script, the portfolio + config libs, and
+# the defaults file. By default also copies _lib-ops-root.sh so the
+# refactor's primary lib-sourcing path is exercised. Pass
+# --no-ops-root-lib to omit it and force the graceful-degradation
+# fallback path (covered by case 7).
 # Returns the sandbox path on stdout.
 # --------------------------------------------------------------------------
 make_fork() {
+  local copy_ops_root=1
+  if [ "${1:-}" = "--no-ops-root-lib" ]; then
+    copy_ops_root=0
+  fi
   local sb
   sb=$(mktemp -d)
   sb=$(cd "$sb" && pwd -P)
@@ -68,6 +77,9 @@ MD
     cp "$HOOK_SRC"          .claude/hooks/link-custom-skills.sh
     cp "$LIB_PORTFOLIO_SRC" .claude/hooks/_lib-portfolio-paths.sh
     cp "$LIB_CONFIG_SRC"    .claude/hooks/_lib-read-config.sh
+    if [ "$copy_ops_root" -eq 1 ]; then
+      cp "$LIB_OPS_ROOT_SRC" .claude/hooks/_lib-ops-root.sh
+    fi
     cp "$DEFAULTS_SRC"      .claude/project-config.defaults.json
     chmod +x .claude/hooks/link-custom-skills.sh
 
@@ -291,6 +303,46 @@ ok=1
 echo "$out" | grep -q "linked 1 custom skill" || ok=0
 [ "$ok" -eq 1 ] && rc2=0 || rc2=1
 assert "case 6: subdir without SKILL.md is skipped; only valid skills linked" "$rc2"
+if [ "$ok" -ne 1 ]; then echo "  out: $out"; fi
+rm -rf "$SB" "$SIB"
+
+# ==========================================================================
+# Case 7: Graceful-degradation fallback — sandbox has NO _lib-ops-root.sh
+# (simulating an old framework version or a stripped-down adopter
+# install). The hook should fall back to its inline walk-up and still
+# link custom skills correctly.
+#
+# This case is the COMPLEMENT to cases 1-6: those exercise the primary
+# lib-sourcing path (because make_fork copies _lib-ops-root.sh by
+# default). Together they pin BOTH branches of the
+#   if [ -f "$HOOK_DIR/_lib-ops-root.sh" ]; then ... else ... fi
+# discovery shape that link-custom-skills.sh (and several other hooks)
+# share.
+# ==========================================================================
+SB=$(make_fork --no-ops-root-lib)
+# Verify the lib really is missing in this sandbox.
+[ ! -f "$SB/.claude/hooks/_lib-ops-root.sh" ] || {
+  echo "FAIL: case 7 setup — _lib-ops-root.sh unexpectedly present"
+  exit 1
+}
+SIB=$(mktemp -d)
+SIB=$(cd "$SIB" && pwd -P)
+mkdir -p "$SIB/custom-skills/fallback-skill"
+cat > "$SIB/custom-skills/fallback-skill/SKILL.md" <<'MD'
+---
+name: fallback-skill
+---
+MD
+cat > "$SB/.claude/project-config.json" <<JSON
+{ "portfolio": { "custom_skills_dir": "$SIB/custom-skills" } }
+JSON
+
+out=$(run_hook "$SB")
+ok=1
+[ -L "$SB/.claude/skills/fallback-skill" ] || ok=0
+echo "$out" | grep -q "linked 1 custom skill" || ok=0
+[ "$ok" -eq 1 ] && rc2=0 || rc2=1
+assert "case 7: graceful-degradation fallback works when _lib-ops-root.sh is absent" "$rc2"
 if [ "$ok" -ne 1 ]; then echo "  out: $out"; fi
 rm -rf "$SB" "$SIB"
 
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,31 +6,31 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash -c 'r=$PWD;while [ ! -f \"$r/onboarding.yaml\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/onboarding-check.sh\"'"
+            "command": "bash -c 'r=$PWD;while [ ! -f \"$r/.apexyard-fork\" ] && [ ! -f \"$r/onboarding.yaml\" ] && [ -n \"$r\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/onboarding-check.sh\"'"
           },
           {
             "type": "command",
-            "command": "bash -c 'r=$PWD;while [ ! -f \"$r/onboarding.yaml\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/check-upstream-drift.sh\"'"
+            "command": "bash -c 'r=$PWD;while [ ! -f \"$r/.apexyard-fork\" ] && [ ! -f \"$r/onboarding.yaml\" ] && [ -n \"$r\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/check-upstream-drift.sh\"'"
           },
           {
             "type": "command",
-            "command": "bash -c 'r=$PWD;while [ ! -f \"$r/onboarding.yaml\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/check-jq-installed.sh\"'"
+            "command": "bash -c 'r=$PWD;while [ ! -f \"$r/.apexyard-fork\" ] && [ ! -f \"$r/onboarding.yaml\" ] && [ -n \"$r\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/check-jq-installed.sh\"'"
           },
           {
             "type": "command",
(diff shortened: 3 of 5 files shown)


atlas-apex (commented):
## Code Review: PR #306

**Commit**: `0d6a1a5173f381980a1e7054389de92f9578d47e`

### Summary

Sweeps all 40 `bash -c` hook wrappers in `.claude/settings.json` to the canonical v2-anchor-aware shape (`.apexyard-fork` OR `onboarding.yaml`, plus the `-n "$r"` infinite-loop guard); refactors `link-custom-skills.sh` to source `_lib-ops-root.sh` with a strict-v1 graceful-degradation fallback; documents the canonical wrapper + hook-internal shapes in AgDR-0041 and cross-links from `docs/multi-project.md`.

The substance of the refactor is solid — the AgDR is honest, the wrapper shape is correct, the docs match the new contract. There is **one blocking issue**: the markdownlint CI check is RED, caused by a single MD018 violation in the new AgDR. That has to be fixed before merge per `.claude/rules/pr-quality.md` § "No Red CI Before Merge".

### Checklist Results
- Architecture & Design:    Pass — wrapper logic is correct under all three layouts (v2-only, v1-only, no-anchor)
- Code Quality:              Pass — mechanical sweep is consistent across all 40 wrappers
- Testing:                   Pass with note — see Suggestions
- Security:                  Pass — no secrets, no privilege changes
- Performance:               Pass — one extra `-f` check per ancestor; negligible
- PR Description & Glossary: Pass — 4 glossary entries, `Closes #302` present, AgDR referenced by full slug
- Technical Decisions (AgDR):Pass — AgDR-0041 is created and engages with alternatives
- Adopter Handbooks:         N/A — diff is shell + JSON wrappers; `clean-architecture-layers.md` doesn't apply

### Issues Found

**⛔ BLOCKER — Red CI (markdownlint MD018)**

The new `docs/agdr/AgDR-0041-sessionstart-v2-anchor-sweep.md:13:1` fails markdownlint with:

```
MD018/no-missing-space-atx  No space after hash on atx style heading
Context: "#242 (split-portfolio v2) move..."
```

Line 13 of the AgDR is the second paragraph of `## Context` and starts with the literal text `#242 (split-portfolio v2) moved …`. markdownlint reads `#242` at the start of a line as a malformed ATX heading (hash with no space after).

Per `.claude/rules/pr-quality.md` § "No Red CI Before Merge": *"Never merge with red CI — even if the failure is pre-existing or unrelated."* This one isn't pre-existing — it ships in the new file.

**Fix**: rewrite the line so `#242` is not at column 1. Two cheap options:

1. Prepend `PR ` or `Issue ` so the line reads `PR #242 (split-portfolio v2) moved …`.
2. Use the cross-repo form `me2resh/apexyard#242 (split-portfolio v2) moved …`. This matches the linking style used elsewhere in the AgDR (the Artifacts section already does this).

Either is a one-character/one-word change. After the fix, re-push and CI should go green.

(The same pattern in the Artifacts section — `me2resh/apexyard#302` — is fine because the `m` comes first.)

### Audit of the load-bearing claims

The review request asked me to be thorough on (a) shape correctness across all 40 wrappers, (b) AgDR honesty about the options, and (c) docs accuracy. Findings:

**(a) Shape correctness — all 40 wrappers swept correctly.**
- I verified 40 `bash -c` entries in `.claude/settings.json` and all 40 contain `.apexyard-fork`, `onboarding.yaml`, AND the `[ -n "$r" ]` guard. Zero drift.
- Section distribution: 7 SessionStart, 3 PreToolUse/Edit-Write-MultiEdit, 27 PreToolUse/Bash, 1 UserPromptSubmit, 2 PostToolUse. That sums to 40 — matches the headline claim, though the PR body says "29 PreToolUse" (it's 30: 3 + 27). Minor counting nit in prose; the bulk count is right.
- The `check-jq-installed.sh` SessionStart entry that landed on `origin/dev` mid-flight is included (line 17 of the new `settings.json`).
- I executed the new shape against `cwd=/tmp` (no anchors anywhere up the tree): the loop walks once, reduces `r="/tmp"` to `r=""`, the `-n "$r"` guard fires, loop exits cleanly with `r=""`; `exec "/.claude/hooks/<name>.sh"` then fails fast with exit 126. Visible failure, no infinite loop. Then ran the OLD legacy shape (no `-n "$r"` guard, only `[ "$r" != / ]`) against the same `cwd=/tmp`: it infinite-loops as predicted. The AgDR's Consequence #3 is grounded in observed behaviour, not theory.

**(b) AgDR honesty about options.**
- **Option A (source the lib from the wrapper)**: I confirmed `bash -c 'echo "0=[$0] dirname0=[$(dirname "$0")]"'` prints `0=[bash] dirname0=[.]`. The `$0=bash` claim is technically accurate. Option A's rejection is not just convenient framing — it's a hard property of `bash -c`. Honest.
- **Option B (inline the v2-aware walk-up)**: chosen. The relaxation from BOTH-of-pair (lib) to either-marker (wrapper) is justified because the wrapper's only job is to locate the hooks dir; the hook itself re-resolves via the lib if it needs strict ops-root semantics. I audited the two SessionStart hooks that DON'T source `_lib-ops-root.sh` (`onboarding-check.sh`, `check-upstream-drift.sh`) and both use `git rev-parse --show-toplevel` instead of relying on the wrapper's `$r`. So the laxness has no semantic effect there either. The reasoning holds for every wrapper in the sweep.
- **Option D (drop the wrapper walk, exec from `$PWD`)**: correctly rejected — breaks when working in `workspace/<project>/`.
- **Gap**: the user's review request mentions a different "Option C" — eliminate the wrapper walk via a Claude Code runtime change. The PR body covers it in "Out of scope" ("would need an upstream change to the Claude Code runtime, not a framework change"), but the AgDR's Options table only lists a different Option C (single dispatch script). Worth adding a short row to the AgDR Options table for the upstream-runtime option, even just to record it as deferred — it's the most principled long-term path and an honest options table should name it. Non-blocking nit.

**(c) Docs accuracy.**
- `docs/multi-project.md` § "Where session-state files live" now correctly describes the dual-anchor model and removes the wrong "BOTH … AND …" copy that was specific to v1. The "BLOCKED: PR has no recorded code-reviewer approval" troubleshooting line is updated to "at least one of the ops-fork anchors". Accurate.
- One minor precision issue: the new third paragraph says *"The same dual-anchor rule applies to the .claude/settings.json hook wrappers themselves"*. The wrappers use the WEAKER v1 condition (single `onboarding.yaml`), not the strict pair — the rule is `.apexyard-fork OR onboarding.yaml` at wrapper level, vs `.apexyard-fork OR (onboarding.yaml AND apexyard.projects.yaml)` at lib level. Calling them both "the dual-anchor rule" elides that asymmetry. Minor wording suggestion; not blocking.

### Handbook Findings

`handbooks/architecture/clean-architecture-layers.md` (always-load) — diff is shell + JSON wrappers + an AgDR + docs. No domain / application / infrastructure boundary in scope. No findings.

1. **Test coverage gap on `link-custom-skills.sh` lib-sourcing path.** The 6 test cases in `test_link_custom_skills.sh` build a sandbox that copies in `_lib-portfolio-paths.sh` and `_lib-read-config.sh` but **not** `_lib-ops-root.sh`. So all 6 cases exercise the **graceful-degradation fallback** (lines 62–71 of the hook), not the new primary path (lines 57–60, `if [ -f "$HOOK_DIR/_lib-ops-root.sh" ]; then . _lib-ops-root.sh; ops_root=$(resolve_ops_root "$PWD") …`). The fallback works (the test fixture writes BOTH anchors, so the strict-v1 fallback condition is satisfied), but the actual code path that ships in production isn't pinned by these tests. Two cheap fixes:
   - Copy `_lib-ops-root.sh` into the sandbox in `make_fork()` so the helper-sourced path is the one exercised.
   - Add a 7th case that creates a v2-only fork (`.apexyard-fork` only, no `onboarding.yaml`) and asserts the hook resolves correctly through the lib.
   The bug this would catch is the helper-sourced path silently returning the wrong dir for a v2-only layout — exactly the failure class this PR is meant to prevent.

2. **AgDR Options table: add the upstream-runtime row.** As noted under audit (b), the most principled solution is an upstream change to Claude Code that resolves hook paths relative to the fork root. It's acknowledged in the PR body's "Out of scope" but missing from the AgDR Options table. Adding a row ("Option E — push the walk into the Claude Code runtime — Pros: removes all wrappers; Cons: needs upstream cooperation, not a framework-side change. Defer.") would make the Options table feel complete and honest.

3. **Docs precision on "dual-anchor rule" asymmetry.** The wrapper's v1 acceptance condition (single `onboarding.yaml`) is a deliberate relaxation from the lib's stricter v1 pair. Worth one sentence in `docs/multi-project.md` to note that wrappers and the lib accept slightly different shapes, and why (wrapper only needs to find the hooks dir; lib establishes canonical ops-root). Non-blocking.

4. **PR body counting nit.** "29 PreToolUse" → "30 PreToolUse" (3 Edit-matcher + 27 Bash-matcher = 30). Cosmetic; the headline total of 40 is correct.

### Verdict
**CHANGES REQUESTED**

The substance of the refactor is good and the architectural reasoning in AgDR-0041 holds up to scrutiny. Blocking only on the red CI from the markdownlint MD018 violation in the new AgDR file (one-line fix). Once that's green, I'm happy to re-review and approve.

---
📌 Reviewed commit: `0d6a1a5173f381980a1e7054389de92f9578d47e`

atlas-apex (commented):
## Code Review (re-review): PR #306

**Commit**: `823b2fc1f63de255f50d334f8f007b18b3e8fd96`

### Summary

Re-review of the fix-up commit `823b2fc` addressing the blocker + 3 non-blocking nits from the previous review at `0d6a1a5`. (Cannot `--approve` own PR; using `--comment` with explicit APPROVED verdict.)

### Checklist Results

- ✅ Architecture & Design:    Pass (unchanged from prior review)
- ✅ Testing:                   Pass — now 7/7 cases, both lib-sourcing branches pinned
- ✅ Security:                  N/A
- ✅ Performance:               N/A
- ✅ PR Description & Glossary: Pass (unchanged from prior review)
- ✅ Technical Decisions (AgDR):Pass — AgDR-0041 now strengthened with Option E
- ✅ Adopter Handbooks:         N/A

### Fix-up verification

**1. MD018 blocker (red CI on prior HEAD)** — RESOLVED.

`docs/agdr/AgDR-0041-sessionstart-v2-anchor-sweep.md` line 13 now reads `PR #242 (split-portfolio v2) moved …` instead of bare `#242 …`. CI `markdownlint-cli2` SUCCESS on `823b2fc`. All five CI checks green (lychee, markdownlint-cli2, Verify Ticket ID, shellcheck).

**2. Option E (nit 1: honesty gap)** — RESOLVED genuinely.

The new row engages with the runtime-resolution alternative, not lip-service:

- **Description names the actual mechanism**: "runtime resolves hook paths from the ops-fork root, not from `$PWD` via `bash -c`"
- **Pros are honest**: "Cleanest — no inline walk in any wrapper, no dispatch hop, no lib-sourcing dance. The runtime already knows where the project root is."
- **Cons name the right blocker**: "Needs an upstream Claude Code change (settings.json semantics + how hook commands are resolved). Out of this PR's scope."
- **Defer rationale closes the loop**: "Option B is the right shape until the runtime grows that ability."

This is the shape an AgDR-level options-considered row should take — names the cleaner path, names what's blocking it, names when to revisit.

**3. Test case 7 (nit 2: fallback path coverage)** — RESOLVED, test passes for the right reason.

Verified at HEAD `823b2fc` (local worktree run):

```
PASS: case 1-6 ... (primary lib-sourcing path)
PASS: case 7: graceful-degradation fallback works when _lib-ops-root.sh is absent
Passed: 7  Failed: 0
```

The case 7 setup is correctly designed:

- `make_fork --no-ops-root-lib` sets `copy_ops_root=0` so the `cp "$LIB_OPS_ROOT_SRC" …` line is skipped
- Pre-flight assertion `[ ! -f "$SB/.claude/hooks/_lib-ops-root.sh" ]` exits 1 with an explicit FAIL message if the lib were unexpectedly present (false-positive setup guard)
- Symlink check + "linked 1 custom skill" output check confirms the inline fallback walk-up successfully resolved the ops root via `.apexyard-fork` and the rest of the hook ran end-to-end

Cases 1-6 now genuinely exercise the lib-sourcing branch (primary path) since `make_fork` copies `_lib-ops-root.sh` by default. Together with case 7, both branches of the `if [ -f "$HOOK_DIR/_lib-ops-root.sh" ]` shape are pinned by tests.

The reviewer-flagged risk ("passes for the wrong reason — some other walk-up resolves the path") doesn't apply: the fallback inline walk-up is the only path that can resolve the ops root when the lib is absent.

**4. Doc precision (nit 3)** — RESOLVED.

The added callout in `docs/multi-project.md` § "Where session-state files live" correctly states:

> The wrappers accept `onboarding.yaml` alone as the v1 anchor, while `_lib-ops-root.sh` (the in-hook resolver) requires BOTH `onboarding.yaml` AND `apexyard.projects.yaml`.

Reading the surrounding prose end-to-end — dual-anchor paragraph → wrappers paragraph → laxness callout → "at least one of the ops-fork anchors" recovery instructions — the flow is coherent. The laxness is explained *before* the recovery instructions mention single-anchor sufficiency, so a reader doesn't trip on "wait, I thought you needed both?". Cross-reference to AgDR-0041 § Decision point 2 anchors the rationale for the future-curious reader.

### Issues Found

None.

### Suggestions

None — all four points cleanly addressed, CI green, tests strengthened.

### Verdict

**APPROVED**

The fix-up commit cleanly resolves the blocker and substantively addresses all three nits — Option E is honest engagement, case 7 is a real fallback test (not just a "test exists" line item), and the doc callout makes the wrapper-vs-lib asymmetry visible to future-us. Solid response to review feedback; ready to merge once CEO approves.

---
📌 Reviewed commit: `823b2fc1f63de255f50d334f8f007b18b3e8fd96`

PR #355, opened 2026-05-20
feat(#347): promote engineering-dept roles to sub-agents + Activation mode (Wave 1 PR 1)

atlas-apex:
## Summary

- **7 engineering-dept sub-agent wrappers shipped at `.claude/agents/`** (head-of-engineering, tech-lead, backend-engineer, frontend-engineer, qa-engineer, platform-engineer, sre) — promotes the engineering personas from passive markdown role docs to first-class Claude Code sub-agents with explicit model assignment and tool restriction. Wave 1 PR 1 of the agent-runtime-overhaul shipped against AgDR-0050.
- **Agent files are THIN WRAPPERS per AgDR-0050 § Axis 1** — each file owns only `name` / `description` / `model` / `allowed-tools` / `persona_name` frontmatter + a body that references `@roles/engineering/<role>.md`. The persona definition stays in `roles/`; agent files are the runtime overlay. Zero content duplication, so a persona edit lands in one place and the wrapper stays inert.
- **Default model matrix per AgDR-0050 § Axis 2** — `opus` for depth (Khalid, Hisham, Saif), `sonnet` for implementation default (Karim, Yasmin, Adel), `haiku` for repeatable checklist work (Salim, QA Engineer). QA also ships **without** Edit/Write per its read-only contract: QA verifies, doesn't ship; the fix flows back to an engineer through a fresh ticket. Adopters override per-agent via the routing config landing in #351 (Wave 1 PR 1 of that ticket runs in parallel).
- **`## Activation mode` section added to all 19 role files per AgDR-0050 § Axis 6** — every persona definition now declares its activation class explicitly (isolated-work-class for 12 roles like QA / Pen Tester / Data Analyst / all 5 Heads of / Security Auditor / Tech Lead; in-flow-class for 7 roles like Backend / Frontend Engineer / Platform Engineer / PM / Designers / Data Engineer). Closes the gap where the hybrid sub-agent vs in-thread choice lived only in the AgDR body — operators can now see at a glance whether activating a role spawns an isolated sub-agent or adopts the persona in-thread.
[...]

PR #353, opened 2026-05-20
feat(#351): agent-routing.yaml schema + portfolio_agent_routing resolver (Wave 1 PR 1)

atlas-apex:
## Summary

- **Adds `agent-routing.yaml.example`** at the framework root — the schema spec for the centralised per-agent model + endpoint override surface (AgDR-0050 Axis 3). Documents required + optional fields (`model`, `endpoint`, `env`, `timeout_seconds`, `allowed_tools_override`) with worked examples (single-agent override, multiple overrides, local routing via Ollama + LiteLLM, Bedrock + AWS env, timeout bump). Empty `agents: {}` block is identical to "no file" — adopters get framework defaults out-of-box.
- **Adds `portfolio_agent_routing` resolver** to `_lib-portfolio-paths.sh` — mirrors the existing `portfolio_registry` / `portfolio_onboarding_path` shape (cache var, default fallback via `_portfolio_get`, absolute-path output via `_portfolio_resolve`). Caller-tolerant of a non-existent file: absence means "framework defaults apply", which is the documented out-of-box behaviour. Resolves the split-portfolio v2 path via `.portfolio.agent_routing` in `.claude/project-config.json`; default is `./agent-routing.yaml` against the ops-fork root.
- **Gitignores `/agent-routing.yaml` at the fork root** so single-fork adopters' routing choices never accidentally leak to the public fork — sibling pattern to `.claude/project-config.json` and `workspace/*/` already in `.gitignore`. Split-portfolio adopters keep it in the private repo via the config block instead.
- **`docs/multi-project.md` gains a "Centralised agent routing" section** between "Private custom skills + handbooks" and "Migrating from split-portfolio v1 to v2" — slots in as a sibling private-repo customisation surface alongside custom-templates / custom-skills / custom-handbooks. Tabulates the file location for both modes, the per-entry schema, the config-block wiring, and an explicit wave plan (sync hook in PR 2, `/setup` integration in PR 3, local-routing entries gated on #348 in PR 4).
- **No sync hook yet — ships in #351 PR 2.** Until `apply-agent-routing.sh` lands the YAML file is documented but inert: `portfolio_agent_routing` resolves the path and the schema example parses cleanly, but no hook applies overrides to `.claude/agents/*.md` frontmatter. PR 2 also adds the pre-commit + pre-push **drift-prevention guards** that block accidental commits of the rewritten frontmatter.

Per AgDR-0050-agent-runtime-overhaul.

## Testing

[...]

Issue #351, opened 2026-05-20
[Feature] Centralised agent-routing config — agent-routing.yaml in private repo, propagates to .claude/agents/*.md at SessionStart


atlas-apex:
## User Story

As an operator running ApexYard who wants to **configure the agent → model mapping in one file** (rather than editing 22+ agent-frontmatter entries by hand), I want a centralised `agent-routing.yaml` that lives in my **private portfolio repo** (split-portfolio mode) or a gitignored fork-root file (single-fork mode), so that:

1. Per-agent model assignments (Haiku for QA, Opus for SRE, local for Idris, etc.) live in ONE place I edit.
2. The public fork's `.claude/agents/*.md` files stay shipped with FRAMEWORK DEFAULTS — my adopter-specific choices never leak to the public fork's commit history.
3. **Local-model routing** (per the #348 spike outcome) is wired the same way as remote-model routing — the routing config carries `endpoint:` + auth env-var refs for adopters who run Ollama / LiteLLM / Bedrock / Vertex behind their own proxy.
4. Switching an agent from Sonnet to Opus, or from Claude API to local Ollama, is a one-line YAML edit + a re-session — not a hunt-and-replace across 22 markdown files.

### File + schema

- [ ] Adopter-facing config at `<private_repo>/agent-routing.yaml` (split-portfolio) OR `<fork_root>/agent-routing.yaml` (gitignored in single-fork)
- [ ] YAML schema:

  ```yaml
  # agent-routing.yaml — per-agent model + endpoint overrides
  agents:
    <agent-name>:
      model: <model-spec>           # e.g. opus, sonnet, haiku, ollama/qwen2.5-coder:14b, bedrock/...
      endpoint: <url>               # optional, for local / proxy routing
      env:                          # optional, additional env vars set for invocations
        OPENAI_API_KEY: $OPENAI_KEY # supports env-var refs
        AWS_REGION: eu-west-2
      timeout_seconds: 60           # optional, default = framework default
  ```

- [ ] Schema documented in the file's leading comment + in `docs/multi-project.md` "Daily workflow under split mode" section
- [ ] Adopters with NO `agent-routing.yaml` get framework defaults (zero-config zero-behaviour-change)
- [ ] An empty `agent-routing.yaml` (with just `agents: {}`) is also zero-config — equivalent to "no file"

### Sync mechanism

- [ ] `SessionStart` hook `apply-agent-routing.sh` runs at session start:
[...]

Issue #347, opened 2026-05-20
[Feature] Promote all 19 role definitions to Claude Code sub-agents (per-role model + tool restriction + isolated context)


atlas-apex:
## User Story

As an operator running ApexYard who wants per-role model assignment, tool restriction, and isolated context for every persona — not just the 5 utility agents shipping today — I want **all 19 role definitions promoted to first-class Claude Code sub-agents**, so that:

1. Each role can run on the model that fits its work (Haiku for checklist-y QA + Data Analyst, Sonnet for the design / PM / engineering majority, Opus for deep reasoning like Tech Lead / SRE / Pen Tester / Head of Engineering).
2. Each role's prompt + tool subset doesn't pollute the main thread's context unless invoked.
3. Auto-activation triggers from `role-triggers.md` (currently injecting markdown into the main thread) instead spawn the matching sub-agent — operator sees the same persona-marker output, but the work happens in a separated context.

This is the natural completion of the persona-naming + role-triggers work shipped in #204 + #206 + AgDR-0018: roles already have identities, responsibilities, and triggers; only the runtime isolation is missing.

### Coverage

- [ ] 19 sub-agent files exist at `.claude/agents/<role-slug>.md`, one per role in `roles/{engineering,product,design,security,data}/`
- [ ] Each sub-agent file carries a `name`, `description`, `model`, `allowed-tools`, and a clear role-prompt body
- [ ] Each sub-agent's `description` matches the role's persona name + one-line responsibility from `roles/<dept>/<role>.md`

### File shape — pick one (decide in AgDR, see Design notes below)

- [ ] **Option A (wrap)**: each `.claude/agents/<role>.md` is a thin wrapper that references `roles/<dept>/<role>.md` for the persona definition. Agent file owns ONLY: `name`, `description`, `model`, `allowed-tools`, `persona_name` frontmatter, plus a `## Role` section that says "Read and adopt `@roles/<dept>/<role>.md` for full identity / responsibilities / CAN-CANNOT / handoff rules." No content duplication.
- [ ] **Option B (merge)**: the role content moves into the agent file. `roles/` becomes either deprecated-alias or retired (with cross-links + a one-time migration commit).
- [ ] **Option C (mirror)**: full content lives in BOTH (agent file for sub-agent invocation, role file for human + AgDR cross-reference reading). Sync-checked via a smoke test.

[...]

PR #357, opened 2026-05-20
feat(#351): agent-routing sync hook + pre-commit/pre-push drift guards (Wave 1 PR 2)

atlas-apex:
## Summary

- **`apply-agent-routing.sh` SessionStart hook** — reads `agent-routing.yaml` (resolved via `portfolio_agent_routing` from #353) and rewrites the affected `.claude/agents/*.md` `model:` frontmatter in-place on every session start. Orphan entries (routing keys that don't match a shipped agent) are silently skipped — adopter typos don't break the session. Idempotent: re-runs against the same config produce no diff. Per AgDR-0050 § Axis 4.
- **`block-agent-routing-drift.sh` pre-commit + pre-push drift guards** — fires when `.claude/agents/*.md` is staged AND the `model:` frontmatter no longer matches the framework default. Exits 2 with the offending agent name + the drift direction. Two PreToolUse entries (`Bash(git commit *)` AND `Bash(git push *)`) so the guard catches both surfaces; the push entry exists because a `git push` can carry a forced commit that bypassed the commit-time gate. Escape hatch: `# routing-config:override <reason>` comment on the staged diff for the rare framework-PR case where the framework-default itself is the *point* of the change.
- **Framework-defaults snapshot** lives at `.claude/agents/.framework-defaults.json` (per-agent `model:` baseline from AgDR-0050 § Axis 2); the SessionStart hook reads it to know what to restore, the drift guard reads it to know what counts as drift. Snapshot ships with the framework; adopters never edit it.
- **`docs/multi-project.md` "Centralised agent routing" section flipped** from "Wave 1 PR 1 (this PR) — inert until PR 2 lands" to live-tense "How the overrides get applied" — names both the SessionStart sync and the two drift guards.

## Testing

- [x] Smoke test: `bash .claude/hooks/tests/test_agent_routing.sh` — 7/7 PASS
  1. Empty config → no-op (no agent files rewritten)
  2. `qa-engineer: model: opus` override → frontmatter flips from `haiku` to `opus`
  3. Orphan entry (`nonexistent-agent: model: opus`) → silently skipped, no error
  4. Idempotent re-run against an already-applied config → zero diff
  5. Drift guard FIRES on staged drift without escape hatch → exit 2
  6. Drift guard ACCEPTS staged drift WITH `# routing-config:override` escape hatch → exit 0 + stderr warning
  7. Drift guard ACCEPTS clean-default commit (no routing-frontmatter changes) → exit 0
- [ ] Manual override smoke (reviewer-friendly):

  ```bash
  cat > agent-routing.yaml <<YAML
  agents:
    qa-engineer:
      model: opus
  YAML
  bash .claude/hooks/apply-agent-routing.sh
  grep '^model:' .claude/agents/qa-engineer.md   # → model: opus

  git checkout .claude/agents/qa-engineer.md     # restore default
  # The drift guard then accepts the next commit because no routing-frontmatter is staged.
  ```

## Glossary

| Term | Definition |
|------|------------|
| **`agent-routing.yaml`** | Adopter-authored YAML at `<private_repo>/agent-routing.yaml` (split-portfolio) or `<fork>/agent-routing.yaml` (single-fork, gitignored) that overrides per-agent `model:` / `endpoint:` / `env:` / `timeout_seconds:` framework defaults. Schema + resolver shipped in #353. |
| **Framework defaults snapshot** | `.claude/agents/.framework-defaults.json` — JSON file mapping every shipped agent name to its baseline `model:`. Source of truth for both the SessionStart sync's restore-from-default step AND the drift guard's "what counts as drift" check. Committed to the framework; adopters never edit. |
| **Drift** | Any state where a staged `.claude/agents/*.md` frontmatter `model:` value differs from the framework-defaults snapshot. Indicates either a legitimate framework PR (escape hatch required) OR an adopter routing override that accidentally got committed (the failure mode this guard prevents). |
| **`# routing-config:override <reason>` escape hatch** | A literal comment on the staged diff that opts the commit out of the drift gate. Used when a framework PR is *intentionally* changing a framework default (rare). Visible + auditable per AgDR-0040's escape-hatch convention pattern. |
| **Orphan entry** | A routing-config key that doesn't match any shipped `.claude/agents/<name>.md` — typo, stale entry, or pre-Wave-3 reference. v1 behaviour: silently skip (adopter typos shouldn't break the session). |

Refs #351
Refs #347 (Wave 1 PR 1 = #355)

### DIFF
--- a/.claude/hooks/apply-agent-routing.sh
+++ b/.claude/hooks/apply-agent-routing.sh
@@ -1,0 +1,410 @@
+#!/bin/bash
+# SessionStart hook: apply adopter agent-routing overrides to the
+# .claude/agents/*.md frontmatter in-place.
+#
+# Per AgDR-0050 § Axis 4 + ticket #351 PR 2 — closes the loop on Wave 1
+# PR 1's schema + portfolio_agent_routing resolver. Adopters edit ONE
+# YAML file (in the private repo for split-portfolio v2, or gitignored
+# at the fork root for single-fork mode); this hook makes those edits
+# LIVE on the next session.
+#
+# Behaviour:
+#   - Resolve the routing path via portfolio_agent_routing
+#   - Silently exit 0 if no routing file exists (zero-config zero-
+#     behaviour-change — the documented out-of-box experience)
+#   - Parse the YAML (yq preferred; python3 fallback)
+#   - For each entry in agents:
+#       1. Locate .claude/agents/<name>.md — silently skip if missing
+#          (adopter may have a stale entry; harmless, not an error)
+#       2. Snapshot the framework default model: line (from HEAD of dev
+#          for the file) into .claude/agents/.framework-defaults.json
+#          so the drift guard has a baseline regardless of whether the
+#          adopter committed the rewrite
+#       3. Rewrite the model: line in the agent file to the override
+#       4. If endpoint: set, write to .claude/session/agent-env/<name>.env
+#          (per-agent endpoint env file; if Claude Code doesn't expose
+#          per-agent env, the adopter falls back to manual export of
+#          ANTHROPIC_BASE_URL — banner caveat below)
+#       5. If env: block set, append KEY=VALUE lines to that env file
+#          (resolving $VAR_NAME refs against the parent env)
+#   - Print a single one-line summary banner: silent on N=0; else
+#       "ApexYard: applied N agent-routing override(s) from agent-routing.yaml"
+#   - Idempotent: running twice with the same config produces the same
+#     state, no compounding writes (env files are truncated, not appended,
+#     when written; the framework-defaults snapshot is keyed by agent
+#     name so re-applications are noops)
+#
+# Banner budget: ≤ 600 chars across all SessionStart hooks. This hook
+# emits at most ~110 chars (silent on no-op).
+#
+# Drift-prevention complement: block-agent-routing-drift.sh fires on
+# git commit + git push and refuses to let routing-induced rewrites
+# escape to a public-class remote. Both hooks together implement
+# AgDR-0050 § Axis 4's "SessionStart rewrite + drift guards" pattern.
+
+set -u
+
+REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
+if [ -z "$REPO_ROOT" ]; then
+  exit 0
+fi
+
+# Walk up to find the apexyard fork root (v2 marker first, legacy v1
+# anchor fallback). Same shape as clear-bootstrap-marker.sh.
+HOOK_DIR="$(cd "$(dirname "$0")" && pwd)"
+ROOT=""
+if [ -f "$HOOK_DIR/_lib-ops-root.sh" ]; then
+  # shellcheck source=/dev/null
+  . "$HOOK_DIR/_lib-ops-root.sh"
+  ROOT=$(resolve_ops_root "$REPO_ROOT")
+else
+  cur="$REPO_ROOT"
+  while [ -n "$cur" ] && [ "$cur" != "/" ]; do
+    if [ -f "$cur/.apexyard-fork" ]; then
+      ROOT="$cur"
+      break
+    fi
+    if [ -f "$cur/onboarding.yaml" ] && [ -f "$cur/apexyard.projects.yaml" ]; then
+      ROOT="$cur"
+      break
+    fi
+    cur=$(dirname "$cur")
+  done
+fi
+
+if [ -z "$ROOT" ]; then
+  exit 0
+fi
+
+# Source path resolvers + config reader. Prefer the hook's neighbour
+# libs (HOOK_DIR/_lib-*.sh) because they're guaranteed to match this
+# hook's version. The walk-up ROOT may be an older copy mid-upgrade
+# (e.g. adopter pulling in a new framework hook before the lib
+# resolvers shipped on its dev branch).
+LIB_SRC_DIR=""
+if [ -f "$HOOK_DIR/_lib-portfolio-paths.sh" ] && [ -f "$HOOK_DIR/_lib-read-config.sh" ]; then
+  LIB_SRC_DIR="$HOOK_DIR"
+elif [ -f "$ROOT/.claude/hooks/_lib-portfolio-paths.sh" ] && [ -f "$ROOT/.claude/hooks/_lib-read-config.sh" ]; then
+  LIB_SRC_DIR="$ROOT/.claude/hooks"
+else
+  exit 0
+fi
+# shellcheck source=/dev/null
+. "$LIB_SRC_DIR/_lib-read-config.sh"
+# shellcheck source=/dev/null
+. "$LIB_SRC_DIR/_lib-portfolio-paths.sh"
+
+ROUTING_PATH=$(portfolio_agent_routing)
+if [ -z "$ROUTING_PATH" ] || [ ! -f "$ROUTING_PATH" ]; then
+  # Zero-config zero-behaviour: no routing file = framework defaults.
+  exit 0
+fi
+
+AGENTS_DIR="$ROOT/.claude/agents"
+if [ ! -d "$AGENTS_DIR" ]; then
+  exit 0
+fi
+
+DEFAULTS_FILE="$AGENTS_DIR/.framework-defaults.json"
+ENV_DIR="$ROOT/.claude/session/agent-env"
+mkdir -p "$ENV_DIR"
+
+# -----------------------------------------------------------------------------
+# Parse the routing YAML — emit one line per agent in the shape:
+#   <name>\t<key>\t<value>
+# Where key ∈ {model, endpoint, env, timeout_seconds, allowed_tools_override}.
+# env values come out as a JSON-encoded object so the consumer below can
+# iterate KEY=VAL pairs without parsing YAML twice.
+#
+# yq is preferred; python3 + PyYAML is the fallback. If neither works the
+# hook is a silent no-op (adopters without yq/python3 must install one or
+# wait for v2 of the schema). Emit a one-line caveat on stderr in that
+# rare case so the adopter knows why their routing config didn't apply.
+# -----------------------------------------------------------------------------
+parse_routing() {
+  if command -v yq >/dev/null 2>&1; then
+    # yq output: tab-separated <name>\t<key>\t<value>. env is emitted as
+    # one row per KEY=VAL inside the env: block.
+    yq eval '
+      .agents // {} | to_entries | .[] | . as $a |
+      (
+        ($a.value.model // "" | select(. != "") | $a.key + "\tmodel\t" + .),
+        ($a.value.endpoint // "" | select(. != "") | $a.key + "\tendpoint\t" + .),
+        ($a.value.timeout_seconds // "" | select(. != "") | $a.key + "\ttimeout_seconds\t" + (. | tostring)),
+        ($a.value.env // {} | to_entries | .[] | $a.key + "\tenv\t" + .key + "=" + (.value | tostring))
+      )
+    ' "$ROUTING_PATH" 2>/dev/null
+    return 0
+  fi
+
+  if command -v python3 >/dev/null 2>&1; then
+    # Try PyYAML first; if not present, fall through to the awk parser.
+    py_out=$(python3 - "$ROUTING_PATH" <<'PY' 2>/dev/null
+import sys
+try:
+    import yaml
+except ImportError:
+    sys.exit(7)
+try:
+    with open(sys.argv[1], 'r', encoding='utf-8') as fh:
+        doc = yaml.safe_load(fh) or {}
+except Exception:
+    sys.exit(0)
+agents = (doc.get('agents') or {})
+if not isinstance(agents, dict):
+    sys.exit(0)
+for name, entry in agents.items():
+    if not isinstance(entry, dict):
+        continue
+    model = entry.get('model')
+    if model:
+        print(f"{name}\tmodel\t{model}")
+    endpoint = entry.get('endpoint')
+    if endpoint:
+        print(f"{name}\tendpoint\t{endpoint}")
+    timeout = entry.get('timeout_seconds')
+    if timeout:
+        print(f"{name}\ttimeout_seconds\t{timeout}")
+    env = entry.get('env') or {}
+    if isinstance(env, dict):
+        for k, v in env.items():
+            print(f"{name}\tenv\t{k}={v}")
+PY
+)
+    py_rc=$?
+    if [ "$py_rc" -ne 7 ]; then
+      printf '%s\n' "$py_out"
+      return 0
+    fi
+  fi
+
+  # Minimal awk fallback — supports the schema's documented shape
+  # (2-space indented YAML, model/endpoint/timeout_seconds at depth 2,
+  # env: subblock at depth 2 with KEY: VALUE pairs at depth 3). Doesn't
+  # try to be a general YAML parser; this is enough for the v1 schema.
+  awk '
+    BEGIN { agents_block = 0; cur_name = ""; in_env = 0 }
+    # Skip comments + blank lines.
+    /^[[:space:]]*#/ { next }
+    /^[[:space:]]*$/ { next }
+
+    # Top-level "agents:" line opens the block. Any subsequent top-level
+    # key (no leading whitespace, no leading hyphen) closes it.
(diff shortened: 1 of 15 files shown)


atlas-apex (commented):
## Code Review: PR #357

**Commit**: `850b0fa0d3414b4d028c5a0740c43831203aaefc`

### Summary

Wave 1 PR 2 of #351 — finishes the agent-routing flow started by #353. Ships the `apply-agent-routing.sh` SessionStart hook that reads `agent-routing.yaml` (resolved via `portfolio_agent_routing`) and rewrites `.claude/agents/*.md` frontmatter in-place, paired with `block-agent-routing-drift.sh` pre-commit + pre-push guards that refuse to let those rewrites escape to a remote. Includes a snapshot mechanism at `.claude/agents/.framework-defaults.json` so the drift guard has a baseline, plus a `# routing-config:override <reason>` escape hatch for legitimate framework-default changes. 7-case smoke test, settings.json wiring, docs flip to live-tense.

### Checklist Results
- Testing:                   Pass (with one note — see Suggestions)
- Technical Decisions (AgDR):Pass (AgDR-0050 § Axis 4 cited correctly + consistently across both hooks + docs/multi-project.md)
- Adopter Handbooks:         N/A (no handbooks dir on this fork)

### Verified locally
- 7/7 smoke-test cases PASS on the head commit
- Token-efficiency Wave 1 invariants still pass with the new SessionStart hook added (153 chars worst-case banner total)
- Byte-equal idempotency confirmed on the actual artefacts: SHA of `qa-engineer.md` AND `.framework-defaults.json` are identical across two consecutive sync-hook runs
- Multi-agent snapshot is valid JSON that round-trips through `json.load()` cleanly
- **Push-gate justification verified**: with the commit-gate bypassed via `--no-verify`-equivalent flow, the push-gate independently catches the drift on push. Confirmed exit 2 + correct error message. This is the load-bearing reason for the dual-guard placement and matches the AgDR-0050 § Axis 4 spec ("pre-commit ... pre-push sweeps the same surface").
- Malformed YAML handling (missing `agents:` key, non-dict agent value, typo'd field name) all silently no-op — robust + consistent with v1 silent-on-error posture

### Verdict on the focus questions

1. **Dual-guard placement** — JUSTIFIED. The push-gate is not redundant. Verified empirically: a drift commit made with the commit-gate bypassed (`git -c core.hooksPath=/dev/null commit ...`) gets caught by the push-gate when the operator next tries to push. The two surfaces have different bypass profiles — `--no-verify` skips the commit gate; only the push-gate sees the staged-to-be-pushed range. Both entries belong.

2. **Silent-skip-on-orphan** — CORRECT v1 posture. Matches AgDR-0050's "Adopter forgets to re-session" + "silent on no-change" pattern. A warning would also be defensible (typos are common, silent skip can mask a stale entry the adopter does not realise stopped applying), but silence is the AgDR-specified posture and the consistent precedent in `link-custom-skills.sh` + similar SessionStart hooks. v2 could add a `--diagnose` flag if adopter feedback warrants it; not a v1 blocker.

3. **Snapshot location** — DEFENSIBLE at `.claude/agents/.framework-defaults.json`. The dotfile convention matches its "session-generated, gitignored, never edited by hand" role; co-located with the `.claude/agents/*.md` files it describes (good locality). Moving to `templates/` would imply framework-shipped, which contradicts its actually-generated-by-hook nature. `_framework-defaults.json` would be more visible to `ls` but also more likely to be confused for an editable file. Keep as-is.

4. **Idempotency claim** — CONFIRMED EMPIRICALLY. SHA of both the agent file and the snapshot JSON are byte-equal across two consecutive runs. Not "close to zero diff" — actually zero.

5. **AgDR linkage** — Correct. AgDR-0050 § Axis 4 cited 5x across the two hooks + docs/multi-project.md, all consistent with the AgDR's "SessionStart hook + pre-commit + pre-push guards" wording.

6. **Glossary completeness** — 5 PR-specific terms covered well. `SessionStart` / `PreToolUse` / `frontmatter` are framework-pervasive vocabulary an apexyard reader already knows; not glossary-required.

### Issues Found

None blocking.

1. **`allowed_tools_override` comment vs implementation drift** — `apply-agent-routing.sh:115` says the parser emits one row per agent for `key ∈ {model, endpoint, env, timeout_seconds, allowed_tools_override}`, but the actual `parse_routing()` only emits rows for `model/endpoint/env/timeout_seconds`. `docs/multi-project.md:549` lists `allowed_tools_override` as a supported schema field. Result: an adopter who sets `allowed_tools_override: ...` in their routing YAML gets a silent no-op — the field is documented as supported but is not actually applied. Either drop the field from the comment + docs (deferring to a later wave) or add the parser branch. Not a blocker because `allowed_tools_override` is documented as "advanced — use sparingly" and the field is genuinely supportable in a follow-up; flag it so adopters do not get surprised.

2. **Test gaps relative to documented surface area** —
   - **`endpoint:` standalone**: only exercised through the idempotency case (case 4), which mixes endpoint + env in one test. An "endpoint-only override writes the env file with exactly one `ANTHROPIC_BASE_URL=` line" case would isolate the endpoint behaviour.
   - **`timeout_seconds:` not tested at all** — value is parsed + written to `APEXYARD_AGENT_TIMEOUT=`, but no case asserts that.
   - **`allowed_tools_override:` not tested** (consistent with suggestion 1 — if it is deferred, that is fine; if it is supposed to be live, it needs a test).
   - **Push-gate drift case not in the test suite** — case 5 covers the commit-gate firing, case 6 covers the escape hatch, case 7 covers no-drift; there is no case for "commit-gate bypassed, push-gate catches the drift on push". I verified this manually and it works, but the regression-protection case is missing. The PR body's "Two PreToolUse entries (`Bash(git commit *)` AND `Bash(git push *)`)" claim should have a test that fails if the push-gate entry is removed.
   - **Two-overrides-same-agent merge** — what happens if `qa-engineer` has both `model:` and `endpoint:` set? The hook handles it (independent code paths per key) but no test confirms.

3. **PR description "Snapshot ships with the framework" vs reality** — the PR body says "Snapshot ships with the framework; adopters never edit it." The `.gitignore` entry says "Adopter-local; the sync hook regenerates it each session, so committing would be churn." These contradict. The implementation does the gitignored / adopter-local thing — so the PR body wording is misleading. Suggest: "Snapshot is adopter-local, gitignored, regenerated by the SessionStart hook from `git show dev:.claude/agents/<name>.md` on each run."

4. **Test-file path typo in PR body** — body references `bash .claude/hooks/tests/test_agent_routing.sh` but actual file is `test_agent_routing_sync_and_drift.sh`. Cosmetic.

5. **Implementation vs AgDR scope** — AgDR-0050 § Axis 4 says the push-gate "sweeps the same surface before any push to a public-class remote". The implementation fires on **any** push, regardless of remote class. The stricter behaviour is arguably better (private remotes can become public; mirrors exist), but it is a deviation worth noting if a future adopter ever asks "why is my drift blocked when I am pushing to my private fork?". Not a code change — either tighten the hook to inspect the push target (`git rev-parse --abbrev-ref @{push}` -> remote name -> public-class lookup), or update the AgDR consequence to say "any push" (probably simpler).

6. **Banner stream** — the sync hook writes to stderr (`echo "..." >&2`), consistent with the existing SessionStart pattern in `check-upstream-drift.sh` and others. Good.

### Handbook Findings

None — no handbooks present on this fork.

### Verdict
**APPROVED**

The two hooks implement AgDR-0050 § Axis 4 cleanly. Dual-guard placement is justified by the verified bypass-via-no-verify scenario. Idempotency is byte-equal across runs. The smoke test covers the load-bearing 7 cases; the gaps above are suggestions for a follow-up test PR, not blockers. The `allowed_tools_override` comment/docs/implementation drift is the most substantive non-blocking note — adopters will silently lose that field today. Ship + file a follow-up.

---
Reviewed commit: `850b0fa0d3414b4d028c5a0740c43831203aaefc`

atlas-apex (commented):
## Code Review: PR #357 (re-review after `71be4b8`)

**Commit**: `71be4b844c3d1d9ebd98aeb2e36eac8d00cc6a8f`

### Summary

Re-review focused on the counts-only follow-up commit (`850b0fa..71be4b8`). The delta is 6 files, all under `site/`, bumping `29 → 31` in marketing copy to reflect the two new hooks added by this PR (`apply-agent-routing.sh` + `block-agent-routing-drift.sh`). No logic change.

### Checklist Results

- Architecture & Design:        N/A (marketing copy only)
- Code Quality:                 Pass
- Testing:                      Pass (smoke test runs, **but see Issues**)
- Security:                     Pass
- Performance:                  Pass
- PR Description & Glossary:    Pass (inherited from prior verdict)
- Technical Decisions (AgDR):   Pass (inherited — no new decisions in this delta)
- Adopter Handbooks:            N/A

### Verification of the three asks

**1. Hook count at HEAD = 31.** Verified by `ls .claude/hooks/*.sh | grep -v _lib | grep -v tests/ | wc -l`. The two new hooks (`apply-agent-routing.sh`, `block-agent-routing-drift.sh`) are present alongside the prior 29. 29 + 2 = 31 ✓

**2. The 6-file diff is exactly site-copy.** Confirmed via the GitHub compare API. Files touched: `site/architecture.html`, `site/architecture.md.gen`, `site/index.html`, `site/index.md.gen`, `site/llms.txt`, `site/llms-full.txt`. No drive-by edits ✓

**3. No stale `29 hooks` / `29 shell scripts` / `29 shell gates` / `29 mechanical gates` — but TWO stale `29 shell hooks` slipped through.** See below.

### Issues Found

**⛔ Two stale `29 shell hooks` references remain at HEAD — files outside the 6-file bump.**

| File:line | Content |
|---|---|
| `site/llms.txt:39` | `- **29 shell hooks** mechanically enforce SDLC rules — ticket-first, migration …` |
| `site/skill.md:35` | `29 shell hooks enforce SDLC rules mechanically — ticket-first, migration` |

The follow-up commit DID touch `site/llms.txt` (line 4: `53 skills, 31 hooks, 19 role definitions`) but missed line 39 of the same file, which uses the `shell hooks` phrasing. `site/skill.md` was not touched at all by this commit and is not in the PR's diff — but it carries the same stale claim, and the explicit purpose of this PR is to ensure no marketing copy claims the old count.

**Why the local smoke test passed despite the drift:** `.claude/hooks/tests/test_site_counts.sh` has a noun-pattern coverage gap. It scans for `[0-9]+ +hooks`, `[0-9]+ +shell +scripts?`, `[0-9]+ +shell +gates?`, `[0-9]+ +mechanical +gates?` — but **not** `[0-9]+ +shell +hooks`. The regex `[0-9]+ +hooks` cannot match `29 shell hooks` because `shell` sits between `29` and `hooks`. Confirmed by running the test at HEAD (it reports PASS) and then running the regex by hand (matches `29 shell hooks` only when the pattern is widened to allow `shell` between).

This is two findings rolled into one:

- **PR-blocking:** the two stale references on the live site contradict the PR's stated purpose. Bump both to `31` in this PR before merge.
- **Test gap (not blocking this PR, but should land in a follow-up):** extend the smoke test to cover `[0-9]+ +shell +hooks?` so a future hook-count drift on this phrasing is caught by CI rather than by re-review.

### Handbook Findings

No handbook violations.

### Suggestions

- After fixing the two stale references, consider a one-line grep guard in the test: `grep -rnE '\b[0-9]+ +shell +hooks?\b' site/` should return zero hits when counts match (run alongside the existing per-file noun-pattern scan). One-liner; closes the coverage gap without restructuring the test.

### Verdict

**CHANGES REQUESTED**

Two stale `29 shell hooks` claims survived the counts refresh — `site/llms.txt:39` and `site/skill.md:35`. The smoke test missed both because its noun-pattern list doesn't cover the `shell hooks` phrasing. Bump both to `31` (and ideally widen the smoke test) before merge.

---
📌 Reviewed commit: `71be4b844c3d1d9ebd98aeb2e36eac8d00cc6a8f`

atlas-apex (commented):
## Code Review: PR #357 (re-review at follow-up HEAD)

**Commit**: `ed8c83efd3ce07e4bb32bea5f4f14d7938693911`

### Summary

Follow-up commit `ed8c83e` resolves the CHANGES REQUESTED verdict on the prior HEAD `71be4b8`. The fix closes the test-coverage gap and the three stale layer-claim lines.

### Checklist Results

- Testing:                   Pass — `bash .claude/hooks/tests/test_site_counts.sh` returns PASS at HEAD `ed8c83e`
- Security:                  Pass (N/A)
- Performance:               Pass (N/A)
- Technical Decisions:       Pass — no new decisions; the regex addition is mechanical
- Adopter Handbooks:         N/A

### Verification of the four concerns raised

1. **`shell +hooks?` regex addition (`test_site_counts.sh:122`)** — correct and well-placed. Sits in the noun-pattern list alongside `shell +scripts?`, `shell +gates?`, `mechanical +gates?`. Verified with the test: `"31 shell hooks"` flagged, `"31 shell hook"` (singular) flagged. False-positive risk for `"shell hooked into the system"` exists in theory (the `?` quantifier means the regex matches the `hook` prefix), but the new pattern is no worse than the existing siblings (`shell +scripts?` would also flag `"shell scripted"`). Adding `\b` word boundaries across all four noun patterns would be a consistency improvement — appropriate for a separate follow-up, not in scope for this PR. **No-go on blocking this for the theoretical false-positive.**

2. **Three multi-line layer-claim refreshes (`site/llms-full.txt:131-133`)** — all three match actuals verified at HEAD:
   - `53 slash commands` — matches `ls -d .claude/skills/*/  | wc -l` = 53
   - `12 sub-agents` — matches `ls .claude/agents/*.md | wc -l` = 12 (5 utility: Rex, Hatim, Munir, Tariq, Idris + 7 engineering: Khalid, Hisham, Karim, Yasmin, Salim, Adel, Saif). Aligns with the AgDR-0050 § Axis 2 matrix referenced in CLAUDE.md.
   - `31 mechanical enforcement scripts` — matches `ls .claude/hooks/*.sh | grep -v _lib- | wc -l` = 31

3. **Cross-line regex gap acknowledgement** — agree it's out of scope for this PR. The current single-line-anchored regex is the right shape for the most common drift pattern (`N noun` on one line); cross-line scanning would need a different architecture (multi-line context buffer, sentence boundary detection) and likely belongs in its own AgDR + follow-up PR. The commit message correctly documents the gap so the next operator hitting it has a starting point.

4. **Carry-over from prior 5 non-blocking suggestions** — checked the PR diff; the `allowed_tools_override` advertised-but-not-parsed gap and the PR body's `test_agent_routing.sh` vs `test_agent_routing_sync_and_drift.sh` filename mismatch are still present, but those are pre-existing carry-overs from the agent-routing scope (PRs #355/#357), not test-coverage issues. **Recommend filing as separate cleanup tickets rather than expanding this PR's scope** — the follow-up commit is surgically scoped to the test gaps I called out, which is the right shape.

### Issues Found

None.

### Handbook Findings

None.

### Suggestions (non-blocking, future work)

1. **Word-boundary the existing noun patterns** — `shell +(scripts?|gates?|hooks?)` etc. all have a latent prefix-match false-positive for `"shell scripted"`, `"shell gated"`, `"shell hooked"`. Add `\b` (or `(?![a-z])`) after the noun in a separate consistency PR.
2. **Cross-line claim drift** — the three lines refreshed in `llms-full.txt:131-133` were only caught by manual audit. A future enhancement could either (a) introduce a multi-line scanner with a sliding window, or (b) emit a YAML manifest of every claim at build time and diff it against a generated authoritative file. Sufficient for now to track as backlog.
3. **`allowed_tools_override` parser** — the schema advertises this field in the example YAML but the SessionStart hook doesn't currently parse it (verified via inspection of `apply-agent-routing.sh` — the field is documented but the loop reads only `model:`, `endpoint:`, `env:`, `timeout_seconds:`). File a follow-up cleanup ticket.
4. **PR body test filename** — the testing section references `test_agent_routing.sh` but the actual file in the diff is `test_agent_routing_sync_and_drift.sh`. Cosmetic; resolve when the PR description is next edited.

### Verdict

**APPROVED**

The follow-up commit closes the two coverage gaps I flagged on the prior review:
- Regex pattern list now covers all four shell-hook noun variants (`scripts?`, `gates?`, `mechanical gates?`, `shell hooks?`)
- All stale layer claims (single-line via regex, multi-line via manual audit) refreshed to actuals at this HEAD

The cross-line scanner limitation is correctly acknowledged as out-of-scope; the non-blocking suggestions in this review can move to separate tickets.

---
Reviewed commit: `ed8c83efd3ce07e4bb32bea5f4f14d7938693911`

PR #405, opened 2026-05-24
fix(#404): remove stale --pdf-output-folder flag from md-to-pdf dispatch

### LINKED ISSUE #404: [Bug] /pdf convert.sh passes stale --pdf-output-folder flag to md-to-pdf; fallback path broken
## Bug Scenario

**Given** the `/pdf` skill ships with a `convert.sh` dispatch helper that runs `md-to-pdf` (npm package) when pandoc is not installed
**When** an operator invokes `/pdf <markdown-file>` on a system without pandoc but with Node available
**Then** `convert.sh` calls `md-to-pdf --pdf-output-folder <dir> ...` and md-to-pdf errors out with `ArgError: unknown or unexpected option: --pdf-output-folder`
**Expected** the conversion succeeds and writes the PDF at the requested path

## Repro Steps

On a system with no pandoc but `npx md-to-pdf` available (current upstream Node 26 default install):

```bash
# A real-world repro from this session: convert the geo-audit report
.claude/skills/pdf/convert.sh \
  --from=/path/to/report.md \
  --to=/path/to/report.pdf
```

Output:

```
convert.sh: input=markdown  converter=md-to-pdf  out=/path/to/report.pdf
ArgError: unknown or unexpected option: --pdf-output-folder
    at arg (/Users/<u>/.npm/_npx/.../node_modules/arg/index.js:132:13)
    at Object.<anonymous> (/Users/<u>/.npm/_npx/.../node_modules/md-to-pdf/dist/cli.js:25:38)
    ...
convert.sh: md-to-pdf conversion failed for /path/to/report.md
```

Manual workaround (succeeds): `cd <dir>; npx md-to-pdf <file>.md` — md-to-pdf defaults to writing `<file>.pdf` next to the source.

## Environment

- Framework: apexyard v2.0.2 (any tagged version that ships the current `convert.sh`)
- Node: v26.0.0
- md-to-pdf: current npm `latest` (the version `npx --yes` resolves)
- pandoc: not installed (the dispatch path that fails)

## Severity

**P2** — `/pdf` is not on the release-blocking path; markdown-only inputs degrade to "operator runs `npx md-to-pdf` by hand". Hits any adopter who tries `/pdf` without pandoc preinstalled (common — pandoc adds LaTeX-related deps that adopters typically defer). The flag in question (`--pdf-output-folder`) was a real md-to-pdf flag at some point; current md-to-pdf uses different flags for output path. The fix is one line in `convert.sh`.

## Mitigation

Two short-term workarounds for affected adopters:

1. Install pandoc — `brew install pandoc` on macOS, `apt-get install pandoc` on Debian/Ubuntu; convert.sh prefers pandoc when present and uses different (current) flags for it.
2. Run md-to-pdf by hand — `cd $(dirname <file>) && npx --yes md-to-pdf <file>.md` writes `<file>.pdf` next to source.

## Investigation Notes

`convert.sh` lives at `.claude/skills/pdf/convert.sh`. The failing path is the md-to-pdf branch (look for `--pdf-output-folder`). Current md-to-pdf CLI uses:

- `--basedir <dir>` for relative-path resolution
- `--config-file <yaml>` for output destination + styling
- output path is controlled by passing the source file with the desired name; md-to-pdf writes `<source-basename>.pdf` in the same directory by default

The fix likely changes the dispatch from `md-to-pdf --pdf-output-folder <dir> <file>` to something like: stage the source file in a temp dir, run `npx md-to-pdf <tmp>/<file>.md`, then `mv <tmp>/<file>.pdf <out>`. Worth checking the latest md-to-pdf docs before deciding — there may be a `--output` or `-o` flag in newer versions.

Should also add a test case in `.claude/skills/pdf/tests/` that runs an end-to-end conversion on a sample fixture, so this kind of regression is caught at PR time rather than at adopter-invocation time. The test can pin a specific md-to-pdf version to insulate against upstream API changes.

## Glossary

| Term | Definition |
|------|------------|
| `md-to-pdf` | npm package that converts markdown to PDF using a headless Chromium fork (Puppeteer). Used by `/pdf` as the no-pandoc fallback path. |
| `pandoc` | Universal-document converter with LaTeX-backed PDF output. `/pdf`'s preferred converter (better typography); requires xelatex / pdflatex for PDF output. |
| `convert.sh` | Dispatch helper at `.claude/skills/pdf/convert.sh` that tries converters in order (pandoc → md-to-pdf → wkhtmltopdf) and outputs to the requested path. |
| `--pdf-output-folder` | Flag that was removed from `md-to-pdf` in a recent major version; `convert.sh` still passes it, which breaks the fallback path. |
| Graceful degrade | The dispatch pattern `/pdf` uses: try each converter in order, fall through on missing-dep; only exit non-zero if NO converter is installed. Currently broken because the md-to-pdf branch errors out instead of falling through. |

atlas-apex:
## Summary

- **Removed stale `--pdf-output-folder` and `--dest-name` flags from `convert.sh`** — md-to-pdf removed both flags in a breaking API change; every invocation of the md-to-pdf fallback path was aborting with `ArgError: unknown or unexpected option: --pdf-output-folder`, meaning `/pdf` was entirely broken for the majority of adopters who don't have pandoc installed (common on macOS where pandoc pulls in LaTeX dependencies)
- **New dispatch strategy: stage-in-temp-dir** — source is copied into a private `$WORK` dir under the desired output stem, `npx -y md-to-pdf <tmp>.md` is invoked there (md-to-pdf writes `<stem>.pdf` next to the source by default), then the result is `mv`'d to the requested `TO_ABS`; this is compatible with the current md-to-pdf API and avoids any cwd or output-name collisions
- **Same fix applied to `md-to-pdf-html` branch** — the HTML → PDF fallback path had identical stale flags and is fixed with the same staging pattern
- **Regression test added** at `.claude/skills/pdf/tests/test_md_to_pdf_fallback.sh` — covers 4 cases: successful conversion, absence of `--pdf-output-folder`, absence of `--dest-name`, and custom output path landing at exactly the requested `--to` path; all 4 pass locally against npm `latest`

## Testing

```bash
# Run the new regression test
bash .claude/skills/pdf/tests/test_md_to_pdf_fallback.sh

# Run the existing smoke test (should still pass)
bash .claude/skills/pdf/tests/smoke.sh

# Manual end-to-end repro from the issue (requires Node, no pandoc needed)
echo '# Test' > /tmp/test.md
.claude/skills/pdf/convert.sh --from=/tmp/test.md --to=/tmp/test.pdf --converter=md-to-pdf
# Expected: exit 0, /tmp/test.pdf written with non-zero size
```

Closes #404

---

## Glossary

| Term | Definition |
|------|------------|
| `md-to-pdf` | npm package that converts markdown/HTML to PDF using headless Chromium (Puppeteer). Used by `/pdf` as the no-pandoc fallback. |
| `--pdf-output-folder` | Flag removed from md-to-pdf in a breaking API change; `convert.sh` was still passing it, causing an `ArgError` on every invocation of the fallback path. |
| `--dest-name` | Companion stale flag — also removed from md-to-pdf; used alongside `--pdf-output-folder` to name the output file. Removed in the same fix. |
| `stage-in-temp-dir strategy` | The replacement approach: copy source into `$WORK/<desired-stem>.md`, run `npx md-to-pdf` there, then `mv` `$WORK/<desired-stem>.pdf` to the requested output path. Relies on md-to-pdf's default behaviour of writing `<source-basename>.pdf` next to the source. |
| `graceful degrade` | The pattern `/pdf` uses: try each converter in order, fall through on missing-dep; only exit non-zero if NO converter is installed. The stale flags were causing the md-to-pdf branch to error (exit 1) instead of degrade (exit 3). |

### DIFF
--- a/.claude/skills/pdf/convert.sh
+++ b/.claude/skills/pdf/convert.sh
@@ -250,21 +250,24 @@ case "$chosen" in
 
   md-to-pdf)
     # Markdown → PDF via the md-to-pdf npm package (chromium-backed).
-    # --as-html-only off → emit the PDF directly.
-    # Use a tmp output then move so md-to-pdf doesn't fight us on cwd.
-    tmp_out="$WORK/out.pdf"
-    if npx -y md-to-pdf --pdf-output-folder "$WORK" --dest-name out.pdf "$FROM_ABS" >&2; then
+    #
+    # md-to-pdf removed --pdf-output-folder and --dest-name in a breaking
+    # change (see me2resh/apexyard#404).  The current CLI writes
+    # <source-basename>.pdf in the same directory as the source file.
+    #
+    # Strategy: copy the source into a private temp dir under the desired
+    # output stem, invoke md-to-pdf there, then mv the result to TO_ABS.
+    # This avoids any cwd or output-name fighting.
+    out_stem="$(basename "${TO_ABS%.*}")"
+    tmp_src="$WORK/${out_stem}.md"
+    tmp_out="$WORK/${out_stem}.pdf"
+    cp "$FROM_ABS" "$tmp_src"
+    if npx -y md-to-pdf "$tmp_src" >&2; then
       if [ -f "$tmp_out" ]; then
         mv "$tmp_out" "$TO_ABS"
         exit 0
       fi
-      # Some versions of md-to-pdf output as <basename>.pdf in the folder.
-      basename_pdf="$WORK/$(basename "${FROM_ABS%.*}").pdf"
-      if [ -f "$basename_pdf" ]; then
-        mv "$basename_pdf" "$TO_ABS"
-        exit 0
-      fi
-      echo "convert.sh: md-to-pdf ran but no PDF appeared in $WORK" >&2
+      echo "convert.sh: md-to-pdf ran but no PDF appeared at $tmp_out" >&2
       exit 1
     fi
     echo "convert.sh: md-to-pdf conversion failed for $FROM_ABS" >&2
@@ -292,20 +295,21 @@ case "$chosen" in
     ;;
 
   md-to-pdf-html)
-    # md-to-pdf accepts HTML when --as-pdf and a chromium backend are
-    # configured. We use it as a last-resort HTML→PDF path.
-    tmp_out="$WORK/out.pdf"
-    if npx -y md-to-pdf --pdf-output-folder "$WORK" --dest-name out.pdf "$FROM_ABS" >&2; then
+    # md-to-pdf accepts HTML as a last-resort HTML→PDF path.
+    #
+    # Same staging strategy as the md-to-pdf (markdown) branch — copy
+    # source into temp dir under desired output stem, let md-to-pdf write
+    # <stem>.pdf next to it, then mv to TO_ABS.
+    out_stem="$(basename "${TO_ABS%.*}")"
+    tmp_src="$WORK/${out_stem}.html"
+    tmp_out="$WORK/${out_stem}.pdf"
+    cp "$FROM_ABS" "$tmp_src"
+    if npx -y md-to-pdf "$tmp_src" >&2; then
       if [ -f "$tmp_out" ]; then
         mv "$tmp_out" "$TO_ABS"
         exit 0
       fi
-      basename_pdf="$WORK/$(basename "${FROM_ABS%.*}").pdf"
-      if [ -f "$basename_pdf" ]; then
-        mv "$basename_pdf" "$TO_ABS"
-        exit 0
-      fi
-      echo "convert.sh: md-to-pdf produced no PDF for HTML input" >&2
+      echo "convert.sh: md-to-pdf produced no PDF at $tmp_out for HTML input" >&2
       exit 1
     fi
     echo "convert.sh: md-to-pdf HTML→PDF conversion failed for $FROM_ABS" >&2
--- a/.claude/skills/pdf/tests/test_md_to_pdf_fallback.sh
+++ b/.claude/skills/pdf/tests/test_md_to_pdf_fallback.sh
@@ -1,0 +1,168 @@
+#!/usr/bin/env bash
+# Regression test for me2resh/apexyard#404
+#
+# Before the fix, convert.sh passed --pdf-output-folder and --dest-name to
+# md-to-pdf.  Those flags were removed in a breaking change; md-to-pdf now
+# exits with:
+#
+#   ArgError: unknown or unexpected option: --pdf-output-folder
+#
+# This test verifies that:
+#   a) --converter=md-to-pdf succeeds when npx + md-to-pdf are available
+#   b) The PDF lands at the requested --to path with non-zero size
+#   c) --pdf-output-folder is NOT passed (old stale flag is gone)
+#
+# The test is skipped — not failed — when npx is absent, so it never blocks CI
+# on a pandoc-only host.  In a local dev environment with Node installed, run:
+#
+#   bash .claude/skills/pdf/tests/test_md_to_pdf_fallback.sh
+#
+# md-to-pdf version pinning:
+#   This test intentionally does NOT pin a specific md-to-pdf version via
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #405

**Commit**: `6e9e42eca66ae0b62c69ba91652065bcb9fc22ca`

### Summary
Removes stale `--pdf-output-folder` and `--dest-name` flags from both `md-to-pdf` dispatch branches in `convert.sh` (markdown and HTML inputs). Replaces with a stage-in-temp-dir strategy: copy source to `$WORK/<out_stem>.{md,html}`, invoke `npx -y md-to-pdf <tmp_src>`, then `mv` the resulting `$WORK/<out_stem>.pdf` to `TO_ABS`. Adds 4-case regression test at `.claude/skills/pdf/tests/test_md_to_pdf_fallback.sh`.

### Checklist Results
- Security:                  Pass (no new attack surface — same external-process invocation pattern, no new args derived from untrusted input)
- Performance:               Pass (one extra `cp` to temp dir, negligible)
- PR Description & Glossary: Pass (Summary, Testing, Closes #404, Glossary all present)
- Summary Bullet Narrative:  Pass (all 4 bullets are verb + >6 words + rationale)
- Technical Decisions (AgDR): N/A (no new dependency, no new pattern — bug fix that restores the documented contract)
- Adopter Handbooks:         N/A (no handbooks in this repo affect `.claude/skills/pdf/`)

### Verification

1. **Diff scope is correct** — only `.claude/skills/pdf/convert.sh` (+25 / -21) and the new `.claude/skills/pdf/tests/test_md_to_pdf_fallback.sh` (+168). No incidental edits to other skills, hooks, rules, agents, or `site/`.

2. **The fix works** — the new `md-to-pdf` branch (lines ~250-270 in head) computes `out_stem` from `TO_ABS`, copies source to `$WORK/<out_stem>.md`, invokes `npx -y md-to-pdf "$tmp_src"` with no removed flags. md-to-pdf's default behaviour (write `<source-basename>.pdf` next to source) lands the result deterministically at `$WORK/<out_stem>.pdf` = the variable `tmp_out`, which is then `mv`'d to `TO_ABS`. The old "some versions of md-to-pdf output as `<basename>.pdf`" double-check fallback is removed, which is correct now that the source basename and the expected output are aligned by construction (no longer two competing locations). Same fix applied symmetrically to `md-to-pdf-html` branch.

3. **Exit codes preserved** — all exit points still use 0/1/2/3 per the SKILL.md contract. The temp-dir cleanup is handled by the existing `trap 'rm -rf "$WORK"' EXIT` at line ~218, so both success and failure paths clean up automatically.

4. **Graceful degrade unchanged** — `choose_for_markdown` still prefers pandoc, falls through to md-to-pdf, exits 3 with `print_install_advisory` if neither is available. The fix only changes how the md-to-pdf branch invokes the binary, not the dispatch logic.

5. **Regression test is high quality** — forces the md-to-pdf path via `--converter=md-to-pdf`; gates on `command -v npx` at top (skips cleanly on Node-less hosts); per-test handling of exit 3 (md-to-pdf network/install failure) as a skip rather than a hard fail; trap cleans up `$FIXTURE`. Tests B and C explicitly grep stderr for the removed flag names — this is a strong assertion that catches regression even if md-to-pdf's CLI changes again in a way that happens to ignore unknown flags silently. Test D verifies output lands at exactly the requested `--to` path including a nested subdir.

### Issues Found
None.

### Suggestions
None blocking. One edge-case observation for the next person who touches this code: `out_stem="$(basename "${TO_ABS%.*}")"` derives the stem from the destination path; if a caller passes `--to=/tmp/foo` with no extension, `${TO_ABS%.*}` returns the path unchanged and the temp file becomes `$WORK/foo.md` → `$WORK/foo.pdf` → `mv` to `/tmp/foo`. Works correctly, just worth knowing.

### Verdict
**APPROVED**

---
Reviewed commit: `6e9e42eca66ae0b62c69ba91652065bcb9fc22ca`

PR #425, opened 2026-05-27
fix(#424): hook walker reads session pin for managed project workspaces

atlas-apex:
## Summary

- **Pin-first resolution** — the settings.json inline walker now reads `~/.claude/apexyard/ops-root-<SESSION_ID>` (written by `pin-ops-root.sh` at session start) before walking up the directory tree. This resolves hooks from `workspace/<name>/` CWDs where the ops repo is a sibling, not an ancestor.
- **Walk-up hardened** — fallback walk-up now also requires `.claude/hooks/` dir alongside the marker file, so it skips the portfolio root (has `onboarding.yaml` but no hooks) even without a pin.
- **47 walker instances updated** — all hook wiring entries in settings.json use the same new pattern.

## Testing

1. From ops repo CWD: hooks fire as before (single-fork mode)
2. From `workspace/yumyum/` CWD: hooks now fire via pin resolution
3. From portfolio root CWD: hooks now fire via pin resolution
4. Without a pin (new session before `pin-ops-root.sh` runs): walk-up fallback works for ops repo CWD, silently skips for portfolio CWD (same as before — no regression)

## Glossary

| Term | Definition |
|------|------------|
| Session pin | A file written at session start containing the ops repo path, so hooks can find the ops root from any CWD |
| Walker | The inline bash one-liner in settings.json that resolves the ops root before exec'ing each hook script |
| Split-portfolio | Deployment mode where ops fork and private portfolio are sibling directories, not nested |

Refs #424

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #430, opened 2026-05-28
fix(#414): add regression guard against v1 walker hooks

### LINKED ISSUE #414: [Bug] Stale v1 walker hooks in settings.json block gh issue create after split-portfolio v2 migration
## Given / When / Then

**Given** an adopter has migrated to split-portfolio v2 (`.apexyard-fork` marker present, `onboarding.yaml` moved to the private portfolio sibling repo)

**When** Claude Code invokes `gh issue create` via the Bash tool

**Then** two stale v1 hook entries in `.claude/settings.json` (lines ~138-144) silently block the command with exit 127, because their walker only checks for `onboarding.yaml`, walks all the way to `/`, and tries `exec /.claude/hooks/validate-issue-structure.sh` which doesn't exist.

## Repro

1. Fork apexyard, run `/split-portfolio` or `/setup` to migrate to v2
2. Confirm `onboarding.yaml` no longer exists at the fork root
3. From Claude Code, run any `gh issue create` command
4. The command hangs/blocks silently — no error message, just non-zero exit from the hook

## Root Cause

Two hook entries in `.claude/settings.json` use the legacy v1 walker pattern:

```json
{
  "type": "command",
  "if": "Bash(gh issue create *)",
  "command": "bash -c 'r=$PWD;while [ ! -f \"$r/onboarding.yaml\" ] && [ \"$r\" != / ];do r=${r%/*};done;exec \"$r/.claude/hooks/validate-issue-structure.sh\"'"
}
```

This walker lacks the v2 `.apexyard-fork` check AND the `[ -n \"$r\" ]` guard. When `onboarding.yaml` doesn't exist anywhere in the tree, it resolves `r=/`, then `exec /.claude/hooks/...` fails with exit 127. Claude Code treats non-zero hook exit as a block.

The v2 entries for the same hooks already exist and work correctly:

```json
"command": "bash -c 'r=$PWD;while [ ! -f \"$r/.apexyard-fork\" ] && [ ! -f \"$r/onboarding.yaml\" ] && [ -n \"$r\" ] && [ \"$r\" != / ];do r=${r%/*};done;[ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ] || exit 0;exec \"$r/.claude/hooks/validate-issue-structure.sh\"'"
```

## Fix

Remove the two stale v1 entries (both `validate-issue-structure.sh` and `block-private-refs-in-public-repos.sh` duplicates). The v2 entries already cover them.

## Severity

High — completely blocks issue creation for any split-portfolio v2 adopter. The failure is silent (no error message, just a hang), making it hard to diagnose.

atlas-apex:
## Summary

- **Investigation finding** — upstream/dev's settings.json is already fully migrated to v2 walker (the #425 fix is complete). No actual stale walkers to remove.
- **Regression guard added** — `test_settings_wrappers_silent_noop.sh` now greps for the v1 walker shape (`bash -c 'r=\$PWD;while`) and fails if any future commit reintroduces it.
- **Test invariants updated** — count by `CLAUDE_CODE_SESSION_ID` (v2 marker) instead of `r=\$PWD` (which matches both v1 and v2), and replace the hardcoded v1 reference wrapper with v2.

## Testing

All 4 invariants pass. New regression check would catch a v1 re-introduction at test time, before reaching CI.

## Glossary

| Term | Definition |
|------|------------|
| v1 walker | Pure walk-up: `r=\$PWD; while [ ! -f markers ]; do r=\${r%/*}; done` — falsely resolves to portfolio root in split-portfolio mode |
| v2 walker | Pin-first then walk-up — reads session pin from `~/.claude/apexyard/ops-root-<SESSION_ID>` before walking, fixes the split-portfolio false-positive |

Closes #414

### DIFF
--- a/.claude/hooks/tests/test_settings_wrappers_silent_noop.sh
+++ b/.claude/hooks/tests/test_settings_wrappers_silent_noop.sh
@@ -46,13 +46,15 @@ mark_fail() {
 echo "== Settings.json hook-wrapper invariants (#370 regression guard)"
 
 # --- Invariant 1: every wrapper has the anchor-found guard --------------------
-# Count wrappers (each `r=$PWD` line is one wrapper) vs guards present.
-total=$(grep -cF 'r=$PWD' "$SETTINGS")
+# v2 wrappers start with the session-pin preamble (CLAUDE_CODE_SESSION_ID), then
+# fall back to the r=$PWD walk-up. Count wrappers by the v2 preamble marker and
+# verify each has the anchor-found guard.
+total=$(grep -cF 'CLAUDE_CODE_SESSION_ID' "$SETTINGS")
 guarded=$(grep -cF '|| exit 0;exec' "$SETTINGS")
 unguarded=$(grep -cF 'done;exec' "$SETTINGS")
 
 if [ "$total" = "0" ]; then
-  mark_fail "wrapper count" "expected ≥1 wrapper matching the r=\$PWD shape; found 0 — settings.json may have drifted"
+  mark_fail "wrapper count" "expected ≥1 v2 wrapper matching CLAUDE_CODE_SESSION_ID; found 0 — settings.json may have drifted"
 fi
 
 if [ "$guarded" = "$total" ] && [ "$unguarded" = "0" ]; then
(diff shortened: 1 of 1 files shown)


atlas-apex (commented):
## Code Review: PR #430

**Commit**: `37c520498cb2805f31f05a4affbb0da40bd148bf`

### Summary

Adds a regression guard to `test_settings_wrappers_silent_noop.sh` that greps for the v1 walker shape (`bash -c 'r=$PWD;while`) and fails if any future commit reintroduces it. Also switches the wrapper-counting invariant from `r=$PWD` (matches v1 and v2) to `CLAUDE_CODE_SESSION_ID` (v2-only marker), and replaces the hardcoded v1 reference wrapper used in invariants 2 and 3 with the canonical v2 shape.

Issue #414's investigation found upstream/dev's settings.json is already fully migrated — no live v1 walkers to delete. This PR is the regression guard so the bug can't silently come back.

### Checklist Results

- Summary Bullet Narrative:  Pass (three narrative bullets with what + why)
- Technical Decisions (AgDR): N/A (test-only change, no architectural decision)
- Adopter Handbooks:         N/A (no `handbooks/` tree applies to test hooks)

### Verification

I ran the updated test against the PR branch:

```
== Settings.json hook-wrapper invariants (#370 regression guard)
  ok   all 47 wrappers have the anchor-found guard
  ok   no stale v1-only walkers found (split-portfolio v2 safe)
  ok   wrapper invoked from /tmp (no fork ancestry) exits 0 with empty stderr
  ok   wrapper invoked from inside the fork still exits 0 silently on non-trigger path

Passed: 4 / Failed: 0
```

Additional sanity checks I ran:

1. **Grep pattern correctness** — the literal-string grep `bash -c 'r=$PWD;while` is a true substring of the v1 walker shape documented in #414's root-cause section, so a future regression would be caught.
2. **No false positives on v2** — the v2 wrapper's `bash -c '...` body starts with `r="";if [ -n "${CLAUDE_CODE_SESSION_ID:-}" ]...` and only reaches `r=$PWD;while` inside the *fallback branch* (well past the `bash -c '` opening). The grep's anchor at the start of `bash -c '` cleanly distinguishes them — confirmed empirically against the v2 shape.
3. **Wrapper count switch is correct** — `CLAUDE_CODE_SESSION_ID` is present in every v2 wrapper's preamble (count = 47 in the PR branch) and absent from v1, so it's a sound v2-only counter. Switching off `r=$PWD` was necessary because `r=$PWD` also appears inside v2's fallback walk-up — counting on it would have over-counted as wrappers migrate.

### Issues Found

None.

### Suggestions

None blocking. The grep pattern is intentionally literal-string (`grep -F`) which is the right call here — a regex would be brittle against benign whitespace changes.

### Verdict

**APPROVED** (posted as comment because PR author and reviewer are the same identity)

Tight regression guard against a class of bug (silent hook-walker exit-127 after split-portfolio v2 migration) that previously took an investigation to root-cause. The test now fails loudly at test time if anyone re-adds a v1 walker, which is much cheaper than diagnosing the symptom (`gh issue create` mysteriously blocked) in the field.

---
Reviewed commit: 37c520498cb2805f31f05a4affbb0da40bd148bf

Issue #195, opened 2026-05-09
[Spike] Measure local-model routing for bounded sub-tasks (issue classify, commit-type suggest, inbox summary)


atlas-apex:
## Driver

Many framework tasks today route through Claude that don't need a strong model — issue triage, conventional-commit type suggestion, AgDR slug generation, branch-name suggestion from issue title, inbox/PR-list summary paragraphs, glossary-entry stubs, language detection on ambiguous filenames. A small local LLM (Llama 3 8B / Mistral / Phi via Ollama) or a purpose-built classifier could handle these in <1s with zero token cost.

This ticket is a research spike — measurement first, recommendation second, implementation in follow-ups. Same shape as the LSP spike (#178 → PR #184).

## Hypothesis

Routing specific bounded tasks (issue classify, commit-type suggest, inbox summary) through a local model instead of Claude reduces token cost ≥10× on those tasks, with quality "good enough" for the framework's purpose, at acceptable latency.

### Phase 1 — measurement

Three representative tasks, three variants each:

1. **Issue classification** — given an issue title + first paragraph, classify as bug / feature / spike / chore. 20 examples from the apexyard portfolio.
2. **Commit-type suggestion** — given a diff, suggest a conventional-commit type (feat / fix / refactor / docs / chore / test). 20 diffs from recent merges.
3. **Inbox summary** — given a list of 10 PR titles, produce a 3-line "what's pending" paragraph. 10 inbox snapshots.

Variants:
- **A. Pure tool**: regex / template / heuristic. No model.
- **B. Local LLM**: Ollama running Llama-3-8B-Instruct (or comparable). Same prompts as Variant C.
- **C. Claude (today's path)**: same prompts. Use Claude Haiku for the cheapest comparison; Sonnet for quality reference.

For each (task × variant), measure:
- Token cost (input + output)
- Wall-clock latency (cold + warm)
- Output quality — score against a hand-graded reference. Inter-rater spread acceptable; aim for ≥85% match on the dominant class.
- Setup cost (one-time install + maintenance)

Document in `docs/spikes/local-model-routing.md`.

### Phase 2 — integration mechanism

Investigate how a local model would plug into apexyard skills:

- **Tool surface**: a new `OllamaCall` tool? An MCP server wrapping Ollama? A bash helper invoked by skills?
- **Skill opt-in**: skills declare which sub-tasks they'd route to local. e.g. `/feature` opts into "local model for issue-class suggestion before showing the user".
[...]

PR #440, opened 2026-05-28
feat(#438): Ollama/LiteLLM agent routing — reachability + model-pulled checks + session-wide ANTHROPIC_BASE_URL

### LINKED ISSUE #438: feat(agent-routing): implement Ollama/local model routing via LiteLLM proxy — currently documented but entirely inert
## Problem

The `agent-routing.yaml` schema documents Ollama local-model routing (Example C) but none of the supporting infrastructure is implemented:

1. **The sync hook (`apply-agent-routing.sh`) is not wired to SessionStart** — model overrides are never applied to agent frontmatter (#435).
2. **`endpoint:` is never consumed** — the schema says "at SessionStart the sync hook sets `ANTHROPIC_BASE_URL` to this value" but the hook does not do this.
3. **No model specifier validation** — `ollama/<model-name>` entries pass through silently without checking that a LiteLLM proxy is reachable or that the model is pulled locally.

The net result: an adopter who follows Example C exactly gets no local routing. Every agent continues using the Claude API regardless of what `agent-routing.local.yaml` says.

### 1. Apply model + endpoint overrides at SessionStart

`apply-agent-routing.sh` (already exists, wired to SessionStart per the file header) needs to:

- For each agent entry with `model: ollama/*`: write the model specifier to the agent's frontmatter (same as Claude model overrides)
- For each entry with `endpoint:`: set `ANTHROPIC_BASE_URL` for the session so Claude Code routes through the LiteLLM proxy

### 2. Reachability check before applying Ollama entries

Before writing an `ollama/*` override, check that the LiteLLM proxy is reachable:

```bash
curl -sf "${endpoint}/health" >/dev/null 2>&1 || {
  echo "⚠ agent-routing: ${agent} — endpoint ${endpoint} not reachable; skipping Ollama override, falling back to framework default"
  continue
}
```

Non-blocking — a downed proxy should not prevent Claude-routed agents from working.

### 3. Model validation for Ollama specifiers

Validate `ollama/<model-name>` entries against a reachable Ollama instance:

```bash
ollama list 2>/dev/null | grep -q "^${model_name}" || \
  echo "⚠ agent-routing: ${agent} — model ${model_name} not found in local Ollama; run: ollama pull ${model_name}"
```

Warning only (non-blocking) — adopter may pull the model after config.

### 4. Docs: hardware guidance in `agent-routing.yaml` Example C

The current example lists four candidate models with RAM estimates. This is good. It should also note:
- Minimum recommended specs for tool-call-reliable local inference
- That `qwen2.5-coder` variants have the best tool-call reliability for structured-output work (ticket-manager, data-analyst)
- A link to the LiteLLM + Ollama setup guide (or a stub at `docs/local-model-setup.md`)

## Acceptance criteria

- [ ] `apply-agent-routing.sh` applies `ollama/*` model overrides to agent frontmatter at SessionStart
- [ ] `apply-agent-routing.sh` sets `ANTHROPIC_BASE_URL` from `endpoint:` entries at SessionStart
- [ ] Reachability check runs before applying any Ollama entry; unreachable proxy emits warning + skips
- [ ] Model-not-pulled check runs and emits a `ollama pull <model>` suggestion
- [ ] All checks are non-blocking (exit 0)
- [ ] `docs/local-model-setup.md` stub added covering LiteLLM install + Ollama config
- [ ] Tests added to `test_agent_routing_sync_and_drift.sh` covering Ollama path
- [ ] Example C in `agent-routing.yaml` uncommented or updated to be copy-pasteable

## Related

- #435 — sync hook not implemented (prerequisite)
- #433 — MCP reindex skipped in /handover

atlas-apex:
## Summary

- **Extends `apply-agent-routing.sh`** to make Ollama Example C actually route — pre-apply reachability probes, model-pulled checks, and a session-wide `__session__.env` write that solves the v1 "Claude Code doesn't consume per-agent env files" gap
- **Writes `docs/local-model-setup.md`** as the single-page install + configure walkthrough (Ollama + LiteLLM proxy + Example C + constraints quoted verbatim from spike #195)
- **5 new test cases** (9–13) covering the Ollama path against a mock curl that consults `$APEXYARD_MOCK_CURL_DIR` for canned responses; 8 existing cases stay green; total 13 PASS / 0 FAIL
- **Updates `agent-routing.yaml.example`** to drop the stale "Until PR 2 lands, ... documented but inert" comment (the sync hook has shipped since #351) and point readers at the new setup doc
- **Adds a one-paragraph pointer** to `docs/getting-started.md` after the LSP section, opt-in by default

## Why this matters

Before this PR: a customer who followed Example C verbatim got nothing. The per-agent env file was written, the model: line was rewritten, and then Claude Code ignored the env file (because v1 doesn't consume per-agent files yet) and routed all traffic through the Anthropic API regardless. No banner, no warning, no signal that the proxy wasn't being used.

After this PR: the SessionStart banner reports `[N Ollama, K warning(s)]` so the operator sees instantly whether their proxy is reachable, whether the named model is pulled, and how many overrides actually applied. The session-wide `ANTHROPIC_BASE_URL` write is the routing mechanism that actually works under v1's "Claude Code reads session env at startup" semantics — per AgDR-0050 § Axis 5. The per-agent files keep being written for forward-compat (when Claude Code exposes per-agent env scoping in a later version).

Spike #195's "NO-GO as designed, partial GO" conclusion is preserved verbatim in the docs — this PR ships the routing mechanism, not a recommendation that you route everything.

## Testing

- [ ] `bash .claude/hooks/tests/test_agent_routing_sync_and_drift.sh` — 13 PASS, 0 FAIL
- [ ] `bash .claude/hooks/tests/test_site_counts.sh` — clean (no new hook added, so no count refresh needed)
- [ ] `bash -n .claude/hooks/apply-agent-routing.sh` — syntax check
- [ ] Manual: start LiteLLM proxy on `:4000` with an Ollama backend; uncomment Example C in `agent-routing.yaml`; restart Claude Code; observe banner `applied 1 agent-routing override(s) [1 Ollama, 0 warning(s)]` and check that `.claude/session/agent-env/__session__.env` contains the expected `ANTHROPIC_BASE_URL=http://localhost:4000`
- [ ] Manual (negative): stop the LiteLLM proxy; restart Claude Code; observe `[1 Ollama, 1 warning(s)]` with `endpoint http://localhost:4000 not reachable` warning, model rewrite still applied
- [ ] Manual (model-not-pulled): start the proxy but don't pull `qwen2.5-coder:14b`; observe `ollama pull qwen2.5-coder:14b` hint in the banner

## Backwards compatibility

- Hook is unchanged behaviour for adopters with no `agent-routing.yaml` (zero-config zero-output)
- Hook is unchanged behaviour for adopters with `model:` overrides but no `endpoint:` field (current Example A / B shapes)
- Tests 1–8 verify the unchanged-behaviour cases still pass byte-for-byte
- Bash 3.2 compatible (uses temp files, not associative arrays) — works on macOS default `/bin/bash`
- No new shell dep beyond `curl` (already required by check-upstream-drift.sh)

## Out of scope (deferred)

- Per-agent invocation env-var scoping (gated on Claude Code upstream surfacing the API — tracked as a follow-up to AgDR-0050 § Axis 5)
- Tool-call reliability matrix per local model — separate spike
- Auto-fallback Claude→Ollama when proxy is down (spike #195 explicitly warned against silent fallback; deliberate non-feature)
- Ollama for embedding routing (different concern, different layer; lives outside the public framework)

Closes #438.

## Glossary

| Term | Definition |
|------|------------|
| **LiteLLM proxy** | Thin HTTP proxy that translates Anthropic Messages API requests into Ollama's native `/api/chat` format. Required for routing agents (which expect an Anthropic-shaped endpoint) through Ollama. Not required for direct Ollama use cases. |
| **Ollama** | Local LLM runtime (https://ollama.com). Serves models on `http://localhost:11434` via its native API. Models pulled via `ollama pull <name>`. |
| **`ANTHROPIC_BASE_URL`** | Env var Claude Code reads at startup to redirect inference traffic to a non-default endpoint. v1 sets this session-wide because Claude Code doesn't expose per-agent env scoping yet. |
| **Reachability check** | 2s-timeout curl against `<endpoint>/v1/models` (LiteLLM's models list) or `/health`. Two attempts; if both fail, endpoint marked unreachable and per-agent env file write is skipped. |
| **Model-pulled check** | 2s-timeout curl against `<endpoint>/api/tags` (Ollama's native model-list endpoint), greps for the named model. Emits `ollama pull <name>` hint on miss; doesn't block the override. |
| **Session-wide env file** | `.claude/session/agent-env/__session__.env` — written by the sync hook with the first reachable endpoint's `ANTHROPIC_BASE_URL`. Read by Claude Code at startup. |
| **Cold start** | ~9-10s first-call latency measured in spike #195 — Ollama pays the model-load cost on the first request after `ollama serve` boot or after `OLLAMA_KEEP_ALIVE` expires. |

### DIFF
--- a/.claude/hooks/apply-agent-routing.sh
+++ b/.claude/hooks/apply-agent-routing.sh
@@ -13,6 +13,16 @@
 #   - Silently exit 0 if no routing file exists (zero-config zero-
 #     behaviour-change — the documented out-of-box experience)
 #   - Parse the YAML (yq preferred; python3 fallback)
+#   - Local-routing prep (me2resh/apexyard#438):
+#       a. For each unique endpoint, probe `<endpoint>/v1/models` or
+#          `/health` with a 2s timeout. Mark reachable/unreachable.
+#          Unreachable endpoints have their per-agent rows filtered out
+#          BEFORE the apply loop — a downed proxy doesn't poison the
+#          session.
+#       b. For `model: ollama/<name>` rows whose endpoint is reachable,
+#          query `<endpoint>/api/tags` for the model name. On miss emit
+#          a warning naming `ollama pull <name>`. The model rewrite still
+#          applies.
 #   - For each entry in agents:
 #       1. Locate .claude/agents/<name>.md — silently skip if missing
 #          (adopter may have a stale entry; harmless, not an error)
@@ -21,14 +31,18 @@
 #          so the drift guard has a baseline regardless of whether the
 #          adopter committed the rewrite
 #       3. Rewrite the model: line in the agent file to the override
-#       4. If endpoint: set, write to .claude/session/agent-env/<name>.env
-#          (per-agent endpoint env file; if Claude Code doesn't expose
-#          per-agent env, the adopter falls back to manual export of
-#          ANTHROPIC_BASE_URL — banner caveat below)
+#       4. If endpoint: set AND reachable, write to .claude/session/
+#          agent-env/<name>.env (per-agent endpoint env file — informational
+#          in v1 until Claude Code exposes per-agent env scoping)
 #       5. If env: block set, append KEY=VALUE lines to that env file
 #          (resolving $VAR_NAME refs against the parent env)
+#   - Write session-wide .claude/session/agent-env/__session__.env with
+#     ANTHROPIC_BASE_URL=<first-reachable-endpoint> (the routing mechanism
+#     that actually works in v1, per AgDR-0050 § Axis 5). Multi-endpoint
+#     declarations warn and use the first.
 #   - Print a single one-line summary banner: silent on N=0; else
-#       "ApexYard: applied N agent-routing override(s) from agent-routing.yaml"
+#       "ApexYard: applied N agent-routing override(s) from agent-routing.yaml [M Ollama, K warning(s)]"
+#     The Ollama / warning suffix is omitted when both counts are 0.
 #   - Idempotent: running twice with the same config produces the same
 #     state, no compounding writes (env files are truncated, not appended,
 #     when written; the framework-defaults snapshot is keyed by agent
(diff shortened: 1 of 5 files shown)


atlas-apex (commented):
## Code Review: PR #440

**Commit**: `215acced4db6bcf1dd88800dbb6ed3f924ce1a2e`

### Summary

Extends `apply-agent-routing.sh` with three coordinated additions (pre-apply reachability probe, Ollama model-pulled check, post-apply session-wide `__session__.env` write), augments the SessionStart banner with `[N Ollama, K warning(s)]`, ships a single-page `docs/local-model-setup.md` walkthrough, points `docs/getting-started.md` at it, and adds 5 new test cases (9-13) backed by a mock-curl fixture harness. Closes #438 — the v1 routing gap where Example C's per-agent env file was written but Claude Code didn't consume it.

### Checklist Results

- Architecture & Design:    Pass — pre-apply / apply / post-apply ordering preserves backward compat; unreachable endpoints filtered from `$ROWS` *before* the apply loop runs, so a downed proxy cannot poison per-agent `.env` files
- Code Quality:              Pass — Bash 3.2 compatible (temp files instead of associative arrays), POSIX-style read loops, mktemp guards, idempotent writes (truncate-then-append on `ANTHROPIC_BASE_URL`)
- Testing:                   Pass — 13/13 cases green locally; 8 pre-existing cases (1-8) byte-for-byte unchanged, confirming backward compat; mock-curl fixture-key transformation (`tr '/:?&=' '_____' | tr -s '_'`) is symmetric between hook and test (verified by direct reproduction)
- Security:                  Pass — no secrets; mock curl scoped to `$APEXYARD_MOCK_CURL_DIR`; reachability probe uses `--max-time 2 --fail --silent` (won't hang the session)
- Performance:               Pass — one curl per *unique* endpoint (not per agent), 2s timeout cap, no probe at all when `agent-routing.yaml` declares no `endpoint:` field (Examples A/B unchanged)
- PR Description & Glossary: Pass — Summary bullets are narrative (bolded change + rationale), Glossary defines 7 load-bearing terms, ticket linked via `Closes #438`, testing block enumerated
- Summary Bullet Narrative:  Pass — every bullet answers what changed + why it matters; not label-only
- Technical Decisions (AgDR): Pass — cites AgDR-0050 § Axis 5 four times in the hook and once in the docs as the design rationale for session-wide `ANTHROPIC_BASE_URL`; no new decision introduced (Axis 5 already canonicalised this path)
- Adopter Handbooks:         N/A — no `handbooks/` directory in this repo

### Issues Found

None blocking.

### Handbook Findings

(no handbooks loaded — adopter handbook layer not present in this fork)

### Suggestions

**nit (non-blocking)** — Edge case in the `env:` block branch at `.claude/hooks/apply-agent-routing.sh:467`: when an `env:` block is declared on an agent whose `endpoint:` is unreachable, the env-block `KEY=VALUE` lines still get written to the per-agent `.env` file (with no `ANTHROPIC_BASE_URL` alongside them, because the endpoint row was filtered upstream). The current PR's docs call per-agent env files "informational in v1" so this is internally consistent, but a future v2 reader of those files might be surprised to find `OLLAMA_KEEP_ALIVE=30m` written for an agent whose endpoint never reached. Consider either: (a) filtering env rows the same way endpoint rows are filtered for unreachable endpoints, or (b) leaving a one-line comment at the top of the per-agent `.env` noting that the endpoint was unreachable at last sync. Not a blocker — the v1 contract is "informational", and tests 1-13 all pass with the current behaviour.

**nit (non-blocking)** — Branch name uses `fix/#438-…` but PR title uses `feat(#438): …`. Neither validator blocks this (both are valid in isolation), but a future grep on `fix/*` branches won't surface this as a feature. Mention it for the next branch — `feature/#438-…` matches the change shape. Not worth a new commit on this PR.

**nit (non-blocking)** — `docs/local-model-setup.md` § "What this guide is *not*" mentions "the apexyard-search MCP server's embedding calls" as a scope exclusion. That's the right shape for keeping the doc focused, and is not a leak (no package URL, no install command, no premium-tier marketing). Flag for future awareness only: this single mention is the boundary between "scope-exclusion disclaimer" and "advertisement"; a second similar mention in the same doc would tip it toward the latter.

### Specific verifications run

- `bash .claude/hooks/tests/test_agent_routing_sync_and_drift.sh` → **13 PASS, 0 FAIL** (8 existing + 5 new)
- `bash .claude/hooks/tests/test_site_counts.sh` → **PASS** (skills=55, hooks=35, roles=19; no count drift, confirming no new hook file added)
- `bash -n .claude/hooks/apply-agent-routing.sh` → syntax OK
- URL→key fixture-key transformation: hook (`tr '/:?&=' '_____' | tr -s '_'` on line 303 of test) ↔ mock curl is symmetric for all 5 fixture URLs used across cases 9-13
- Banner suffix logic (line 567-568): `suffix` is empty when `ollama_applied==0 && warnings==0`, preserving the bare-banner shape for non-Ollama overrides (cases 2 + 8 confirm)
- AgDR-0050 § Axis 5 cross-references: 4 in the hook, 1 in `docs/local-model-setup.md` (line 120), AgDR file exists at the cited path
- Referenced spike memo (`docs/spikes/local-model-routing.md`) exists in the repo
- Leak scan on the new doc: zero references to `apexyard-premium` or `apexyard-search` pip package install/URL; one scoped exclusion mention of `apexyard-search MCP server` which is correctly framed as out-of-scope, not as an advertisement
- CI: all 5 checks green (verify-ticket-id, lychee, markdownlint, shellcheck, site-counts drift) at HEAD `215acced`

### Verdict
**APPROVED** (posted as COMMENT — author is self, GitHub blocks `--approve` on own PR)

The change is well-scoped, well-tested, backward-compatible, and the PR description is exemplary (narrative bullets, comprehensive glossary, explicit backwards-compat + out-of-scope sections). Three non-blocking nits noted for awareness; none warrant a revision before merge.

---
Reviewed commit: \`215acced4db6bcf1dd88800dbb6ed3f924ce1a2e\`

PR #444, opened 2026-05-28
fix(#442): warn at SessionStart when Ollama routing is INACTIVE; promote shell-profile step in docs

### LINKED ISSUE #442: [Bug] #440 follow-up: Ollama agent routing not actually active without manual shell-profile step
## Given / When / Then

**Given** the adopter has configured Ollama Example C in `agent-routing.yaml` per `docs/local-model-setup.md` and restarted Claude Code,

**When** they invoke an agent that should route through the LiteLLM proxy,

**Then** they expect the agent's inference to actually go through the proxy.

**Currently:** the SessionStart banner reports `applied 1 agent-routing override(s) [1 Ollama, 0 warning(s)]` and the file `.claude/session/agent-env/__session__.env` is written with `ANTHROPIC_BASE_URL=<endpoint>` — but Claude Code's process env is set when Claude is launched, not when SessionStart hooks fire. Without a manual `source` in the shell profile **before** launching Claude Code, `$ANTHROPIC_BASE_URL` is unset in Claude's process and every agent call still hits the Anthropic API. The banner is misleadingly optimistic.

## Repro

1. Configure Example C in `agent-routing.yaml` (uncomment `ticket-manager` with `endpoint: http://localhost:4000`, `model: ollama/qwen2.5-coder:14b`)
2. Start LiteLLM proxy on `:4000`
3. **Do NOT** add `source .claude/session/agent-env/__session__.env` to `~/.zshrc`
4. Start Claude Code
5. Banner emits `applied 1 agent-routing override(s) [1 Ollama, 0 warning(s)]`
6. Invoke `/task something-small` (triggers ticket-manager agent)
7. Observe: LiteLLM proxy log shows **no inbound request**; the call went to `api.anthropic.com` instead

## Acceptance Criteria

- [ ] `apply-agent-routing.sh` adds a routing-active visibility check at the end of its run: if `__session__.env` was written with `ANTHROPIC_BASE_URL=<endpoint>` but `$ANTHROPIC_BASE_URL` in the current process env doesn't match, emit a one-line warning naming the gap and the exact remediation (`source <path>` in shell profile, then relaunch Claude Code).
- [ ] The warning counts as a banner warning — folds into the `[N Ollama, K warning(s)]` summary.
- [ ] `docs/local-model-setup.md` promotes the shell-profile step from buried step 5 to a "Before you start" callout at the top of the agent-routing section, with the exact line to add and the reason it's required.
- [ ] Existing 13 cases in `test_agent_routing_sync_and_drift.sh` continue to pass.
- [ ] New test case: when the hook writes `__session__.env` but the test runner's parent env doesn't have `ANTHROPIC_BASE_URL` set to the endpoint, the warning fires.
- [ ] Adopter who has the shell-profile step in place: no warning, banner shows zero warnings.

## Out of scope (deferred)

- Modifying `.zshrc` / `.bashrc` directly from the hook — too invasive.
- A separate `apexyard-launch` wrapper script that sets env then `exec`s Claude Code — bigger UX change, separate ticket.
- Per-agent env scoping (the AgDR-0050 § Axis 5 follow-up) — still gated on upstream Claude Code.

Refs me2resh/apexyard#438 (the PR that landed the inert plumbing).

atlas-apex:
## Summary

- **Adds a routing-active visibility check** in `apply-agent-routing.sh` — after writing `__session__.env`, compare the resolved endpoint to `$ANTHROPIC_BASE_URL` in the current process; on mismatch emit a one-line warning naming the gap and the exact remediation. Folds into the existing `[N Ollama, K warning(s)]` banner.
- **Promotes the shell-profile step to a "Before you start" callout** at the top of `docs/local-model-setup.md` (was buried at step 5). Explains why SessionStart hooks can't propagate env vars into Claude's process and shows the exact one-liner to add.
- **14/14 tests pass** (4 existing cases updated to preset `ANTHROPIC_BASE_URL`, 1 new case 14 specifically asserts the routing-inactive warning).

## Why this matters

#440 shipped the apply-agent-routing.sh extension that writes `__session__.env` with `ANTHROPIC_BASE_URL=<endpoint>`. The SessionStart banner reports `applied 1 agent-routing override(s) [1 Ollama, 0 warning(s)]`. **But Claude Code's process env was set at launch time** — SessionStart hooks run in child shells and cannot mutate the parent. Unless the adopter sources `__session__.env` in their shell profile **before** launching Claude Code, the routing is INACTIVE and every agent call still hits the Anthropic API.

The buried-at-step-5 doc made this easy to skim past on first install. The optimistic banner reinforced the illusion that everything worked.

After this PR, the banner reports `[1 Ollama, 1 warning(s)]` with explicit text naming the gap, the file path, the shell-profile line to add, and the relaunch requirement.

## Testing

- [ ] `bash .claude/hooks/tests/test_agent_routing_sync_and_drift.sh` — 14/14 PASS
- [ ] `bash -n .claude/hooks/apply-agent-routing.sh` — syntax OK
- [ ] **Manual happy-path**: source `__session__.env` in `~/.zshrc`, relaunch Claude, observe `[1 Ollama, 0 warning(s)]` and `$ANTHROPIC_BASE_URL` set in the process
- [ ] **Manual gap-path**: skip the shell-profile step, relaunch Claude, observe `[1 Ollama, 1 warning(s)]` with the routing-INACTIVE warning quoting the exact path
- [ ] **Recovery**: add the shell-profile line, open a fresh terminal, relaunch — warning clears

## Backwards compatibility

- Adopters who have the shell-profile step in place: zero behaviour change (warning is silent on match).
- Adopters who declared an `endpoint:` but skipped the shell-profile step: now get a clear warning instead of silent failure. This is the intended fix.
- Adopters with no `endpoint:` declaration: zero behaviour change (check only runs when a reachable endpoint exists).

## Out of scope (deferred)

- Modifying `.zshrc` / `.bashrc` directly from the hook — too invasive.
- A wrapper script (`apexyard-launch`) that sets env then `exec`s Claude — bigger UX change, separate ticket if desired.
- Per-agent env scoping (still gated on upstream Claude Code surfacing per-agent invocation env).

Closes #442.

## Glossary

| Term | Definition |
|------|------------|
| **Routing-active check** | New SessionStart-time check in `apply-agent-routing.sh` that compares the resolved endpoint to `$ANTHROPIC_BASE_URL` in the current process. Fires a warning when they differ. |
| **`__session__.env`** | File written by `apply-agent-routing.sh` at `.claude/session/agent-env/__session__.env` containing `ANTHROPIC_BASE_URL=<endpoint>`. Must be sourced from the adopter's shell profile to take effect. |
| **Shell-profile step** | The manual one-liner the adopter adds to `~/.zshrc` / `~/.bashrc` to source `__session__.env` on shell start. Required for routing to actually activate (Claude Code can't read this file by itself). |
| **Routing INACTIVE** | The state where the override has been "applied" (model rewritten, env file written) but Claude Code's process env doesn't have `ANTHROPIC_BASE_URL` set — so every agent call still hits the Anthropic API. The warning this PR adds catches exactly this state. |

### DIFF
--- a/.claude/hooks/apply-agent-routing.sh
+++ b/.claude/hooks/apply-agent-routing.sh
@@ -538,6 +538,26 @@ if [ "$REACHABLE_COUNT" -ge 1 ]; then
     : > "$SESSION_ENV"
   fi
   printf 'ANTHROPIC_BASE_URL=%s\n' "$FIRST_EP" >> "$SESSION_ENV"
+
+  # -----------------------------------------------------------------------------
+  # Routing-active check — see me2resh/apexyard#442.
+  # __session__.env is written above, but Claude Code's process env was set
+  # when Claude was launched — SessionStart hooks run in child shells and
+  # can't mutate the parent. So unless the adopter sources __session__.env
+  # in their shell profile BEFORE launching Claude Code, the routing won't
+  # actually take effect even though the banner reports overrides applied.
+  # Detect the mismatch and warn explicitly so the adopter sees the gap
+  # at SessionStart rather than discovering it via "why is my proxy log
+  # empty?" later.
+  # -----------------------------------------------------------------------------
+  if [ -z "${ANTHROPIC_BASE_URL:-}" ] || [ "${ANTHROPIC_BASE_URL:-}" != "$FIRST_EP" ]; then
+    cat >&2 <<EOF
+⚠ agent-routing: ANTHROPIC_BASE_URL=$FIRST_EP was written to $SESSION_ENV but is NOT set in this Claude session's process env. Routing is INACTIVE — every agent call still hits the Anthropic API. To activate, add this line to your shell profile (~/.zshrc / ~/.bashrc) and relaunch Claude Code from a fresh terminal:
+    [ -f "$SESSION_ENV" ] && . "$SESSION_ENV" && export ANTHROPIC_BASE_URL
+See docs/local-model-setup.md § "Before you start" and me2resh/apexyard#442.
+EOF
+    warnings=$((warnings + 1))
+  fi
 fi
 
 # Clean up local-routing scratch files.
--- a/.claude/hooks/tests/test_agent_routing_sync_and_drift.sh
+++ b/.claude/hooks/tests/test_agent_routing_sync_and_drift.sh
@@ -538,7 +538,11 @@ agents:
     endpoint: http://localhost:4000
 YAML
 
-banner_output=$(cd "$SB" && PATH="$MOCK_BIN:$PATH" bash .claude/hooks/apply-agent-routing.sh 2>&1 < /dev/null || true)
+# Case 9 simulates the happy path where the adopter has done the shell-profile
+# step from docs/local-model-setup.md — ANTHROPIC_BASE_URL is already set in
+# the process env so the new routing-active check (me2resh/apexyard#442)
+# does NOT fire.
+banner_output=$(cd "$SB" && ANTHROPIC_BASE_URL=http://localhost:4000 PATH="$MOCK_BIN:$PATH" bash .claude/hooks/apply-agent-routing.sh 2>&1 < /dev/null || true)
 after_tm=$(read_model_line "$SB/.claude/agents/ticket-manager.md")
 session_env="$SB/.claude/session/agent-env/__session__.env"
 agent_env="$SB/.claude/session/agent-env/ticket-manager.env"
@@ -551,7 +555,7 @@ if [ "$after_tm" = "ollama/qwen2.5-coder:14b" ] \
    && [ "$session_set" = "1" ] \
    && [ "$agent_set" = "1" ] \
    && echo "$banner_output" | grep -qE 'applied 1 agent-routing override.*1 Ollama, 0 warning'; then
-  mark_pass "case 9: Ollama agent with reachable proxy + pulled model (full apply, session + per-agent env, no warnings)"
+  mark_pass "case 9: Ollama agent with reachable proxy + pulled model (full apply, session + per-agent env, no warnings; ANTHROPIC_BASE_URL preset)"
 else
   mark_fail "case 9: Ollama agent reachable + pulled" "model=$after_tm session=$session_set agent=$agent_set banner=[$banner_output]"
 fi
@@ -619,7 +623,9 @@ agents:
     endpoint: http://localhost:4000
 YAML
 
-banner_output=$(cd "$SB" && PATH="$MOCK_BIN:$PATH" bash .claude/hooks/apply-agent-routing.sh 2>&1 < /dev/null || true)
+# Case 11 — ANTHROPIC_BASE_URL preset (happy-path shell), so the only warning
+# should be the model-not-pulled hint, not the routing-active check.
+banner_output=$(cd "$SB" && ANTHROPIC_BASE_URL=http://localhost:4000 PATH="$MOCK_BIN:$PATH" bash .claude/hooks/apply-agent-routing.sh 2>&1 < /dev/null || true)
 session_env="$SB/.claude/session/agent-env/__session__.env"
 session_set=0
 [ -f "$session_env" ] && grep -q '^ANTHROPIC_BASE_URL=http://localhost:4000$' "$session_env" && session_set=1
@@ -628,7 +634,7 @@ if echo "$banner_output" | grep -qE 'model qwen2\.5-coder:14b not in local Ollam
    && echo "$banner_output" | grep -qE 'ollama pull qwen2\.5-coder:14b' \
    && [ "$session_set" = "1" ] \
    && echo "$banner_output" | grep -qE 'applied 1 agent-routing override.*1 Ollama, 1 warning'; then
-  mark_pass "case 11: Ollama agent with reachable proxy + missing model (override applies, pull-hint emitted)"
+  mark_pass "case 11: Ollama agent with reachable proxy + missing model (override applies, pull-hint emitted; ANTHROPIC_BASE_URL preset)"
 else
   mark_fail "case 11: Ollama model-not-pulled" "session=$session_set banner=[$banner_output]"
 fi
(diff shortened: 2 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #444

**Commit**: `4139484fae731e168387e1eaae88eb294635e82c`

### Summary

Closes the visibility gap in the Ollama agent-routing path shipped by #438/#440. After writing `__session__.env`, the apply hook now compares the resolved endpoint to `$ANTHROPIC_BASE_URL` in the live process env. On mismatch (the common case unless the adopter has the shell-profile step in place), it emits a one-line warning that names the gap, the path of the file Claude needs to source, the exact shell-profile line to add, and the required relaunch. The warning folds into the existing `[N Ollama, K warning(s)]` banner suffix so adopters discover the problem at SessionStart rather than via "why is my proxy log empty?" later. Three coordinated changes — hook + tests + a "Before you start" callout promoted to the top of `docs/local-model-setup.md`.

### Checklist Results

- Testing:                   Pass (14/14 cases, including new case 14 for the unset-env path)
- Security:                  Pass (no secrets, no auth, no I/O on untrusted input)
- Performance:               Pass (constant-time string compare, no syscall added)
- PR Description & Glossary: Pass (Summary + Why + Testing + Backwards compatibility + Out-of-scope + Glossary, all narrative)
- Summary Bullet Narrative:  Pass (every bullet says what + why)
- Technical Decisions (AgDR):N/A — bug fix on an existing feature, no new technical decisions; AgDR-0050 § Axis 4 / Axis 5 already cover the design that this PR makes more visible
- Adopter Handbooks:         N/A — diff touches only shell hooks + a markdown doc; the always-load architecture/general handbooks are about TS layering and DB migration safety, neither applies; no `.ts`/`.py`/`.go`/`.rs` files so the language buckets don't load

### Verifications against the request

| Concern | Result |
|---|---|
| Mismatch (env set to a DIFFERENT endpoint) warns | Pass — conditional is `[ -z … ] || [ … != "$FIRST_EP" ]`, so any non-equal value (empty or different host) triggers the warning |
| Check only fires when `REACHABLE_COUNT >= 1` | Pass — nested inside the existing `if [ "$REACHABLE_COUNT" -ge 1 ]; then` block at line 526; if no reachable endpoint, `FIRST_EP` isn't even resolved and the whole block (including the new check) is skipped |
| Cases 9 / 11 / 12 / 13 pass with `ANTHROPIC_BASE_URL` preset | Pass — confirmed by running the suite locally |
| Case 14 exercises the unset-env path | Pass — uses `unset ANTHROPIC_BASE_URL;` inside the test subshell; asserts session env written, banner reports `1 Ollama, 1 warning(s)`, warning text contains "routing is INACTIVE" (case-insensitive) and the resolved session-env path (fixed-string) |
| Case 10 unaffected (unreachable endpoint → no session env, no check) | Pass — `REACHABLE_COUNT=0` short-circuits before the new check; case 10 still asserts `session_absent=1` and a single warning (the unreachable hint) and still passes |
| `bash -n` syntax clean | Pass on both files (apply hook + tests) |
| Site-counts test stays clean | Pass — no new hook added, only an extension to an existing one |
| CI all green | Pass — 5/5 checks green (lychee, markdownlint, PR title, shellcheck, site-counts) |

Ran the suite locally to confirm: `Passed: 14, Failed: 0`. Re-read the diff against `set -u` (active at line 59 of the hook) — the new check correctly uses `${ANTHROPIC_BASE_URL:-}` with the `:-` fallback so it doesn't blow up when the var is unset.

### Issues Found

None blocking.

**suggestion:** trailing-slash brittleness in the mismatch comparison. `$ANTHROPIC_BASE_URL=http://localhost:4000/` (with a trailing slash) vs `endpoint: http://localhost:4000` (no slash) is a literal-string mismatch and would emit the routing-INACTIVE warning even though the adopter is correctly configured. Worth considering a normalisation step (strip one trailing `/` from both sides before compare) in a future iteration. Not blocking for this PR — the warning text points at both the file and the source-of-truth value, so the adopter can self-diagnose quickly. Open a follow-up if you see this hit in practice.

**nit:** the conditional `[ -z "${ANTHROPIC_BASE_URL:-}" ] || [ "${ANTHROPIC_BASE_URL:-}" != "$FIRST_EP" ]` simplifies to `[ "${ANTHROPIC_BASE_URL:-}" != "$FIRST_EP" ]` because `FIRST_EP` is guaranteed non-empty inside the `REACHABLE_COUNT >= 1` block. The current form is more verbose but arguably clearer about intent (handles the "unset" case explicitly). Leave as-is unless you want the one-line shrink.

**suggestion:** consider linking the warning to the section anchor — `docs/local-model-setup.md#before-you-start--the-one-manual-step` — rather than the section name. Anchor links survive section renames; section-name references can rot if the heading text is edited. Cheap to update; not blocking.

### Handbook Findings

No findings. The always-load handbooks (`architecture/clean-architecture-layers.md`, `architecture/migration-safety.md`, `general/commit-message-quality.md`) don't trigger:

- Clean Architecture Layering is about TS domain/application/infrastructure imports — diff is shell + markdown
- Migration Safety is about DB schema migrations — no DB content
- Commit Message Quality — the single PR commit has a substantive WHY-body (the failure mode, the cause, the fix, the trade-offs, what's deferred), tracker ref (`Closes #442`, `Refs #438`), inline glossary, and `Co-Authored-By` on AI-assisted change. Exceeds the handbook standard.

No language handbooks apply (no `.ts`/`.tsx`/`.py`/`.go`/`.rs` files in the diff).

### Verdict

**APPROVED on merit** — substantive APPROVED-on-merit. Posting as COMMENT (GitHub blocks self-approve on author-owned PRs). Both the new check and case 14 are well-targeted; the doc change moves the load-bearing manual step from buried-at-step-5 to top-of-page where adopters can't miss it; the test suite remains deterministic across the env-presence variations. Backwards-compatible — silent for adopters who already have the shell-profile step in place; explicit warning for the ones who don't.

---

Reviewed commit: `4139484fae731e168387e1eaae88eb294635e82c`

Issue #461, opened 2026-05-29
[Chore] require-agdr-for-arch-pr.sh --body parser truncates at first embedded quote (false blocks)


atlas-apex:
## Driver

`require-agdr-for-arch-pr.sh` extracts the PR body from the `gh pr create` command line to scan it for an AgDR reference (or the `<!-- agdr: not-applicable -->` skip marker). Its `--body "..."` extractor uses a `sed` capture that stops at the **first embedded double-quote** inside the body value. Any PR body that contains a `"` before its AgDR reference / skip marker gets truncated at that quote, so the hook never sees the reference and **false-BLOCKs** the PR.

Hit during the apexyard-deploy Netlify work (apexyard-premium#8): a PR body with quoted text early in the Summary caused two spurious AGDR blocks even though the AgDR reference was present further down. The workaround was to switch to `--body-file <path>` (the hook reads the whole file via its body-file branch, which doesn't have the truncation bug). The author did NOT use a bypass marker — the block was a false positive, not a missing AgDR.

This is the same parser-shape used by the AgDR hook's flag extractor (and noted as "same parser shape as block-private-refs-in-public-repos.sh"), so the bug may exist in sibling hooks too — worth checking whether the private-refs hook and any other `--body`-scanning hooks share the truncation.

## Scope

- Fix the `--body` extractor in `require-agdr-for-arch-pr.sh` so it captures the full quoted value, including embedded quotes. Options:
  1. Greedy capture to the last quote before the next recognised flag / end of string (handles the common `--body "...."` case)
  2. Prefer the `--body-file` / `-F` path in guidance and treat inline `--body` as best-effort (least-effort, but leaves the latent bug)
  3. A more robust shell-word tokenizer for the command line
- **Audit sibling hooks** that scan `--body` with the same `sed` shape (`block-private-refs-in-public-repos.sh` at minimum; grep for the extractor pattern) and fix them consistently if affected.
- Add a regression test: a `gh pr create --body "...embedded \" quote... AgDR-0001-foo"` shape must be recognised as having the AgDR reference (not truncated).

## Acceptance Criteria

- [ ] A `gh pr create --body "<text with an embedded double-quote> ... AgDR-NNNN-slug"` is correctly recognised (hook passes, no false block)
- [ ] The `<!-- agdr: not-applicable -->` skip marker is still detected when it appears after an embedded quote in `--body`
- [ ] Sibling `--body`-scanning hooks audited; any sharing the truncation bug are fixed the same way
[...]

PR #462, opened 2026-05-29
fix(#461): fix --body parser truncation in require-agdr-for-arch-pr.sh

atlas-apex:
## Summary

- **Fixed false-block on PRs with quoted body text** — `extract_flag_value()` in `require-agdr-for-arch-pr.sh` used `sed … "[^"]*" …` which truncated the extracted body at the first embedded double-quote, causing the hook to never see an AgDR reference or skip marker that appeared after any quoted text in the PR body. Replaced with the greedy `awk` extractor already used by `block-private-refs-in-public-repos.sh` (#227) and `validate-issue-structure.sh`, which anchors on the next flag boundary rather than the first internal `"`.
- **Three regression tests added** — covering the three ACs from #461: (A) AgDR reference after embedded quote passes, (B) skip marker after embedded quote passes with warning, (C) missing AgDR with embedded quote still blocks (true-negative confirms the gate is not over-corrected).
- **Sibling hook audit complete** — `block-private-refs-in-public-repos.sh` and `validate-issue-structure.sh` already carry the fixed `awk` form; `validate-pr-create.sh`, `require-migration-ticket.sh`, `require-active-ticket.sh`, and `verify-commit-refs.sh` do not own a `--body` extractor and are unaffected.

## Testing

1. Run `.claude/hooks/tests/test_require_agdr_for_arch_pr.sh` — all existing cases pass, three new `[#461-A/B/C]` cases pass.
2. Verify true-negative: a body with an embedded `"` but no AgDR reference still exits 2 ("no AgDR reference").
3. Verify `--body-file` path is unchanged — hook reads the file directly, bypasses the extractor, unaffected by this fix.

Refs #461

---

## Glossary

| Term | Definition |
|------|------------|
| `extract_flag_value()` | Shell function that parses a named flag value (e.g. `--body "..."`) out of a raw `gh pr create` command string passed via JSON to the hook. |
| `[^"]*` (sed form) | Non-greedy character class that matches any char except `"` — stops at the first embedded quote inside the body, truncating everything after it. |
| Greedy awk extractor | The replacement: uses `awk` with `(.*)` (greedy) plus a flag-boundary anchor (`[[:space:]]+--[a-zA-Z]` or EOS) to match the full body span to its closing quote, not to the first internal `"`. |
| Flag-boundary anchor | The end condition `([[:space:]]+--[a-zA-Z]|[[:space:]]*$)` that tells the greedy match where the flag's quoted value ends — at the next `--flag` or end-of-string. |
| False block | The hook rejecting a PR that actually satisfies the rule — here, an AgDR reference present but hidden behind the truncation. |
| `--body-file` path | `gh pr create --body-file <path>` — the hook reads the file directly and is not affected by the extractor bug; recommended for bodies with rich Markdown. |

### DIFF
--- a/.claude/hooks/require-agdr-for-arch-pr.sh
+++ b/.claude/hooks/require-agdr-for-arch-pr.sh
@@ -51,17 +51,53 @@ fi
 extract_flag_value() {
   # $1 = flag regex (e.g. --title | -t). Matches:
   #   --title "value with spaces"
-  #   --title 'value'
+  #   --title 'value with spaces'
   #   --title value
+  #
+  # The original sed form used `[^"]*` which truncated at the first embedded
+  # double-quote inside the body value — false-blocking PRs whose body contained
+  # a `"` before the AgDR reference (me2resh/apexyard#461). The fix mirrors the
+  # already-corrected extractor in block-private-refs-in-public-repos.sh
+  # (me2resh/apexyard#227): awk with a greedy `(.*)` match anchored on the next
+  # recognised flag boundary (whitespace + `--<letter>`) or end-of-string. This
+  # captures the full quoted span, including any embedded quotes, without
+  # bleeding past the start of the next flag.
   local flag_re="$1"
   local cmd="$2"
-  local v
-  v=$(echo "$cmd" | sed -nE "s/.*(${flag_re})[[:space:]]+\"([^\"]*)\".*/\2/p" | head -1)
-  if [ -n "$v" ]; then echo "$v"; return; fi
-  v=$(echo "$cmd" | sed -nE "s/.*(${flag_re})[[:space:]]+'([^']*)'.*/\2/p" | head -1)
-  if [ -n "$v" ]; then echo "$v"; return; fi
-  v=$(echo "$cmd" | sed -nE "s/.*(${flag_re})[[:space:]]+([^[:space:]]+).*/\2/p" | head -1)
-  echo "$v"
+  printf '%s' "$cmd" | awk -v FLAG_RE="$flag_re" -v SQ="'" '
+    { buf = (NR == 1 ? $0 : buf "\n" $0) }
+    END {
+      s = buf
+      # Double-quoted value: greedy `(.*)` anchored on next flag or EOS.
+      re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+\"", "", chunk)
+        sub("\"([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
+        sub("\"[[:space:]]*$", "", chunk)
+        print chunk
+        exit
+      }
+      # Single-quoted value: same greedy + anchor treatment.
+      re = "(" FLAG_RE ")[[:space:]]+" SQ "(.*)" SQ "([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
+      if (match(s, re)) {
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #462

**Commit**: `9d5a754c7d2311fb199d8fc90b45fe67f73e84f6`

### Summary
Fixes false-BLOCKs in `require-agdr-for-arch-pr.sh` (#461). The old `extract_flag_value()` used a sed capture (`"[^"]*"`) that truncated `--body` at the first embedded double-quote, so any AgDR reference / skip marker appearing after a quote was never seen → spurious block. The fix replaces the sed extractor with the greedy `awk` extractor already shipped in the sibling `block-private-refs-in-public-repos.sh` (#227): greedy `(.*)` anchored on the next flag boundary or end-of-string. Three regression tests added.

### Checklist Results
- ✅ Architecture & Design:    Pass — extractor lifted verbatim from the sibling hook; no new abstraction
- ✅ Code Quality:              Pass — shellcheck green in CI; clear in-function rationale comment
- ✅ Testing:                   Pass (with a non-blocking coverage nit, below)
- ✅ Security:                  Pass — no over-correction; true-negative still blocks
- ✅ PR Description & Glossary: Pass — narrative Summary, full Glossary, Testing checklist, `Refs #461`
- ✅ Technical Decisions (AgDR): N/A — bug fix mirroring an established sibling-hook pattern, no new decision
- ✅ Adopter Handbooks:         N/A — no handbooks triggered by this diff

### Verification I ran
The new extractor is **byte-identical** to `block-private-refs-in-public-repos.sh::extract_flag_value` (lines 67–101 here vs 165–199 there) — confirmed by direct comparison.

I probed the extractor directly against the **double-quoted body branch where the original bug actually lived** (the regression tests only exercise the single-quote branch — see nit):

| Case | Input shape | Extracted | Result |
|------|-------------|-----------|--------|
| Bug path | `--body "… \"greedy\" … AgDR-0007-x"` | full body incl. AgDR | AgDR found → **PASS** (old sed would have truncated at first `"` → false-block) |
| True-negative | `--body "… \"greedy\" … no record"` | full body, no AgDR | AgDR absent → **BLOCK** (gate not over-corrected) |
| Skip marker | `--body "… \"pure\" … <!-- agdr: not-applicable -->"` | marker preserved | marker detected → bypass |
| Single-quote (test path) | `--body '… "greedy" … AgDR-0007-x'` | full body incl. AgDR | AgDR found → PASS |

All three new `[#461-A/B/C]` cases pass. CI is green (Verify Ticket ID, shellcheck, site-counts all pass).

> Note: `test_require_agdr_for_arch_pr.sh` shows one `FAIL` for the **pre-existing** `spike active-ticket marker` case when run from an isolated clone outside an ops-root anchor — it fails identically on the base branch (`d4d79fc`) and is unaffected by this PR. CI (properly anchored) reports the suite green.

### Issues Found
None blocking.

### Suggestions
- **nit (test coverage):** The three `#461` regression tests pass a *single-quoted* `--body '...'` containing *embedded double-quotes*. That exercises the awk **single-quote branch** — but the original `[^"]*` truncation bug lived in the **double-quote branch** (`--body "..."` with an embedded `"`). So the suite doesn't directly guard the exact path that broke. Non-blocking because: (a) I verified the double-quote branch independently above and it behaves correctly for pass/block/skip-marker; (b) the extractor is byte-identical to the already-tested sibling hook; (c) the double-quote shape is the common real-world form and is exercised in practice. Worth a follow-up: add one `--body "...\"...\"... AgDR-..."` double-quoted case so the precise regression is pinned in the suite. **Agree with the prior reviewer that this is a nit, not a blocker.**

### Verdict
**APPROVED**

A non-blocking test-coverage nit does not gate this fix — the parser correction is verified correct across all three quote branches and is identical to the established sibling-hook implementation.

---
📌 Reviewed commit: `9d5a754c7d2311fb199d8fc90b45fe67f73e84f6`

Issue #41, opened 2026-04-12
feat: per-project active ticket markers in ops repo


atlas-apex:
## Summary

Currently `.claude/session/current-ticket` is a single marker that tracks one active ticket. This doesn't scale when jumping between multiple managed projects.

## Proposed Design

Centralize all active ticket markers in the ops repo, keyed by project name:

```
.claude/session/tickets/
├── sharppick          ← active ticket for SharpPick
├── billing-api        ← active ticket for billing-api
└── ...
```

All gitignored. The hook resolves which project the current edit belongs to by matching `cwd` against the registry's `workspace` paths in `apexstack.projects.yaml`, then looks up the corresponding marker.

## Changes Required

- `require-active-ticket.sh` — resolve project from cwd, look up per-project marker
- `/start-ticket` skill — write to `tickets/<project-name>` instead of `current-ticket`
- Backward compat: fall back to `current-ticket` if no per-project marker exists

## Why

- No need to create `.claude/session/` inside each project repo
- One centralized place to manage all active tickets
- Scales to N projects naturally
- No risk of committing session state into project repos

PR #524, opened 2026-06-06
feat(#513): per-worktree ticket marker tier (fix same-project concurrent-agent collision)

### LINKED ISSUE #513: [Feature] per-worktree ticket marker tier to fix same-project concurrent-agent collision
## Problem
The two-tier active-ticket marker layout (`tickets/<project>` + `current-ticket`, introduced in #41) fixes cross-project concurrent agents cleanly — each managed project has its own file. It does not fix same-project concurrent agents.

When an orchestrator fans out two agents to work different tickets within the same managed project simultaneously, both agents declare their ticket by writing to `ops_root/.claude/session/tickets/<project>`. Last writer wins:

- Agent A declares ticket X → writes `tickets/my-app`
- Agent B declares ticket Y → overwrites `tickets/my-app` with Y
- Agent A's subsequent hook checks pass against ticket Y (wrong ticket)
- The gate's integrity is silently broken — no error, no warning

This is not hypothetical: orchestrator-driven multi-agent flows that fan out independent tickets within the same repo hit this today. The only workaround is to serialize tickets within a single project, which halves the throughput advantage of the orchestrator pattern.

## Proposed behaviour
Add a third marker tier resolved before the existing two:

```
ops_root/.claude/session/tickets/<project>/<branch>  ← NEW: per-worktree/agent
ops_root/.claude/session/tickets/<project>            ← existing: per-project
ops_root/.claude/session/current-ticket               ← existing: fallback
```

Resolution order in `require-active-ticket.sh` (and `require-migration-ticket.sh`, which mirrors the same lookup):

1. Detect the current worktree branch for the file's repo (`git -C "$FILE_DIR" branch --show-current`, or read `CLAUDE_WORKTREE_BRANCH` env var if the harness sets it at worktree-spawn time)
2. If a branch is detected, look up `tickets/<project>/<safe-branch>` first (safe-branch = branch name with `/` replaced by `__` for filesystem safety)
3. Fall back to `tickets/<project>` (single-agent or non-worktree path — no change for existing users)
4. Fall back to `current-ticket`

`/start-ticket` would write to the per-worktree file when running inside a worktree (detected via `git worktree list`), falling back to the per-project file otherwise. No behaviour change for single-agent sessions.

## Why it helps adopters
Orchestrators that fan out independent tickets to parallel sub-agents on the same managed project are blocked today — they must serialize per-project to avoid marker collision. With the third tier, each sub-agent operates in its own worktree branch and declares its ticket independently without touching siblings.

Today's workaround: orchestrator queues tickets per-project and serializes sub-agent spawns within each repo. This is explicit coordination overhead that disappears when the marker is scoped to the worktree.

## Scope / out of scope
In scope:
- `require-active-ticket.sh` — add tier-3 lookup
- `require-migration-ticket.sh` — same lookup pattern (already mirrors require-active-ticket)
- `/start-ticket` skill — write tier-3 path when worktree is detected
- `.claude/hooks/README.md` — update marker layout diagram

Out of scope:
- Non-worktree concurrent agents (two agents in the same working tree on the same project — this is a process error, not a marker problem)
- Cross-session persistence of per-worktree markers (worktree removed → marker orphaned; acceptable, same as current stale-marker behaviour)

## Framework version
v2.3.0

## Related
- #41 — introduced the two-tier per-project layout this builds on
- `require-active-ticket.sh` — primary hook to change
- `require-migration-ticket.sh` — secondary hook (mirrors lookup logic)
- `/start-ticket` skill — writes the marker; needs the third-tier write path

## Glossary
| Term | Definition |
|------|------------|
| Marker tier | A specific file path the active-ticket hook checks in order; first hit wins, allowing coarser fallbacks without breaking finer-grained isolation |
| Per-worktree marker | A ticket marker scoped to a single git worktree branch, enabling two agents on the same managed project to hold independent tickets concurrently |
| Orchestrator | A coordinating Claude session that decomposes work and fans out implementation to parallel sub-agents, each assigned a separate ticket |
| Safe-branch name | A branch name with `/` replaced by `__` so it is valid as a filesystem path segment |

atlas-apex:
## Summary
- **Fixes a real, silent ticket-gate collision** — the #41 two-tier marker layout (`tickets/<project>` → `current-ticket`) handles *cross-project* concurrency but not *same-project*: two agents fanned out on different tickets within one repo both write `tickets/<project>`, last-writer-wins, and the loser's subsequent hook checks pass against the **wrong ticket** with no error. This blocks orchestrator throughput (today's only workaround is serialising tickets per project).
- **Adds a per-worktree tier resolved first** — `tickets/<project>/<safe-branch>` (`safe-branch` = branch with `/`→`__`), in both `require-active-ticket.sh` and `require-migration-ticket.sh` (which mirrors the lookup). Branch is read from `CLAUDE_WORKTREE_BRANCH` (harness-set at worktree spawn) or `git -C <file-dir> branch --show-current`. Parallel agents each declare their ticket independently.
- **Zero behaviour change for single-agent flows** — no worktree branch ⇒ no tier-0 marker ⇒ falls straight through to the existing per-project file, then `current-ticket`. The file-vs-directory duality of `tickets/<project>` (FILE in single-agent mode, DIRECTORY in worktree mode) is safe because every lookup uses `[ -f ]` — a directory reads as "tier-1 absent", so tiers 0 and 1 never collide.
- **`/start-ticket` writes the right tier** — detects a linked worktree (`--git-dir` ≠ `--git-common-dir`) and writes the per-worktree path; otherwise the per-project file. Docs updated: hooks header comment, `.claude/hooks/README.md` (diagram + description), the skill's layout table. Decision in `AgDR-0066`.

## Testing
1. `test_require_active_ticket_bash.sh` — **15/15 pass**, including 3 new cases: per-worktree marker honored on the matching branch; branch-B **isolation** (a `tickets/myproj/feature__a` marker does NOT satisfy an agent on `feature/b` → BLOCKED); per-project FILE marker still works under a workspace path with no worktree (single-agent regression).
2. `bash -n` + `shellcheck -S error` clean on both modified hooks.
3. `markdownlint` clean on the AgDR, README, and SKILL.

> Decision context for #513 "is this a real bug?": **yes** — confirmed in the marker-resolution code, non-hypothetical for `/fan-out` / orchestrator flows. Fix is additive and low-risk.

Closes #513

---

## Glossary
| Term | Definition |
|------|------------|
| Marker tier | A path the active-ticket hook checks in order; first hit wins, so coarser fallbacks don't break finer-grained isolation. |
| Per-worktree marker | A ticket marker scoped to one git-worktree branch, letting two agents on the same project hold independent tickets concurrently. |
| Safe-branch | A branch name with `/` replaced by `__` so it's a valid filesystem path segment. |
| Linked worktree | A `git worktree`-created checkout whose `--git-dir` differs from the main repo's `--git-common-dir`. |
| Last-writer-wins | The collision this fixes: two writes to one shared marker file leave only the second, silently. |

### DIFF
--- a/.claude/hooks/README.md
+++ b/.claude/hooks/README.md
@@ -28,7 +28,7 @@ These four hooks make the SDLC mechanical instead of advisory. Each enforces a r
 
 **Event:** `PreToolUse` on `Edit | Write | MultiEdit`.
 
-**What it does:** blocks edits to any code path unless a session marker exists. Resolution is two-tier (#41): per-project marker at `<ops_root>/.claude/session/tickets/<project>` when `FILE_PATH` is under `<ops_root>/workspace/<project>/`, otherwise falls back to `<ops_root>/.claude/session/current-ticket`. Exempts `.claude/`, `docs/`, `projects/*/docs/`, and any `*.md` file so framework / doc / meta work is still fluid.
+**What it does:** blocks edits to any code path unless a session marker exists. Resolution is three-tier (#41 + #513): (0) per-worktree marker at `<ops_root>/.claude/session/tickets/<project>/<safe-branch>` when the edited file's repo is on a git-worktree branch — lets parallel agents on the *same* project hold independent tickets; (1) per-project marker at `<ops_root>/.claude/session/tickets/<project>` (a FILE) when `FILE_PATH` is under `<ops_root>/workspace/<project>/`; (2) fallback `<ops_root>/.claude/session/current-ticket`. Exempts `.claude/`, `docs/`, `projects/*/docs/`, and any `*.md` file so framework / doc / meta work is still fluid.
 
 **Enforces:** the Pre-Build Gate in `.claude/rules/workflow-gates.md` — "do not start coding until the ticket exists, has acceptance criteria, and is broken into tasks."
 
@@ -335,7 +335,9 @@ These were already in place before the enforcement layer and remain unchanged (e
 ```
 .claude/session/
 ├── onboarded                     # created by /onboard, read by onboarding-check
-├── current-ticket                # created by /start-ticket, read by require-active-ticket
+├── tickets/<project>             # per-project ticket marker (FILE) — #41
+├── tickets/<project>/<branch>    # per-worktree ticket marker — #513 (DIR form; parallel agents)
+├── current-ticket                # created by /start-ticket, read by require-active-ticket (fallback)
 ├── pending-reviews/<pr>          # created by auto-code-review, tracks PRs awaiting Rex
 ├── reviews/<owner>__<repo>__<pr>-rex.approved           # created by code-reviewer agent, read by merge-gate
 ├── reviews/<owner>__<repo>__<pr>-ceo.approved           # created by /approve-merge, read by merge-gate
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -231,11 +231,21 @@ fi
 # tests below distinguish them, so the two tiers never conflict.
 PER_WORKTREE_MARKER=""
 if [ -n "$PROJECT" ]; then
-  # Branch: prefer the harness-set env var (populated at worktree spawn), else
-  # ask git for the current branch of the edited file's repo.
+  # Branch: prefer the harness-set env var (populated at worktree spawn). Else
+  # only treat the file's repo as worktree-scoped when it's a LINKED worktree,
+  # detected by comparing the ABSOLUTE git-dir against the ABSOLUTE common-dir
+  # (they differ only in a linked worktree). This matches /start-ticket's
+  # write-side detection exactly — no read/write asymmetry — and the absolute
+  # forms avoid the false positive where, in the main checkout from a subdir,
+  # `--git-dir` is absolute but `--git-common-dir` is relative.
   WT_BRANCH="${CLAUDE_WORKTREE_BRANCH:-}"
   if [ -z "$WT_BRANCH" ]; then
-    WT_BRANCH=$(git -C "$(dirname "$FILE_PATH")" branch --show-current 2>/dev/null)
+    _fdir=$(dirname "$FILE_PATH")
+    _gd=$(git -C "$_fdir" rev-parse --absolute-git-dir 2>/dev/null)
+    _gcd=$(git -C "$_fdir" rev-parse --path-format=absolute --git-common-dir 2>/dev/null)
+    if [ -n "$_gd" ] && [ "$_gd" != "$_gcd" ]; then
+      WT_BRANCH=$(git -C "$_fdir" branch --show-current 2>/dev/null)
+    fi
   fi
   if [ -n "$WT_BRANCH" ]; then
     SAFE_BRANCH="${WT_BRANCH//\//__}"   # '/' → '__' for a filesystem-safe segment
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -4,17 +4,22 @@
 # in CLAUDE.md, workflows/sdlc.md, or .claude/rules/workflow-gates.md.
 #
 # Active tickets are declared by the /start-ticket skill. The marker
-# layout is two-tier (apexyard#41):
+# layout is three-tier (apexyard#41 + #513):
 #
-#   ops_root/.claude/session/tickets/<project>    ← per-project, preferred
-#   ops_root/.claude/session/current-ticket       ← ops-repo / fallback
+#   ops_root/.claude/session/tickets/<project>/<branch>  ← per-worktree (#513)
+#   ops_root/.claude/session/tickets/<project>           ← per-project (#41)
+#   ops_root/.claude/session/current-ticket              ← ops-repo / fallback
 #
-# Resolution order for a given FILE_PATH:
-#   1. If FILE_PATH is under ops_root/workspace/<project>/, look up
-#      ops_root/.claude/session/tickets/<project>. If present → exempt.
-#   2. Fall back to ops_root/.claude/session/current-ticket. If present →
-#      exempt.
+# Resolution order for a given FILE_PATH under ops_root/workspace/<project>/:
+#   0. If the file's repo is on a git worktree branch (or CLAUDE_WORKTREE_BRANCH
+#      is set), look up tickets/<project>/<safe-branch>. If present → exempt.
+#      (Lets parallel agents on the SAME project hold independent tickets.)
+#   1. Look up tickets/<project> (a FILE). If present → exempt.
+#   2. Fall back to current-ticket. If present → exempt.
 #   3. Otherwise, block with instructions.
+#
+# tickets/<project> is a FILE in single-agent mode, a DIRECTORY in worktree
+# mode; the `-f` tests keep tiers 0 and 1 from conflicting.
 #
 # Ops root is the apexyard fork root (has both onboarding.yaml and
 # apexyard.projects.yaml at the top level). It's discovered by walking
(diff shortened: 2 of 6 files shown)


atlas-apex (commented):
## Code Review: PR #524

**Commit**: \`60b3bda91901093b83ee12c215b569a79703e2a6\`

### Summary
Adds a tier-0 per-worktree ticket marker (\`tickets/<project>/<safe-branch>\`) resolved before the existing per-project file and the \`current-ticket\` fallback, in both \`require-active-ticket.sh\` and \`require-migration-ticket.sh\`. Fixes the last-writer-wins collision where two agents fanned out on different tickets within one repo both wrote the shared per-project marker, silently passing subsequent gate checks against the wrong ticket. \`/start-ticket\` writes the per-worktree path when inside a linked worktree. AgDR-0066 records the decision.

### Checklist Results
- ✅ Architecture & Design:     Pass
- ✅ Testing:                   Pass
- ✅ Security:                  Pass
- ✅ PR Description & Glossary:  Pass
- ✅ Summary Bullet Narrative:  Pass (bold-lead + rationale on every bullet)
- ✅ Technical Decisions (AgDR): Pass (AgDR-0066 linked + present)
- ✅ Adopter Handbooks:         N/A (no violations — see below)

### Verification performed
I checked out the PR head (\`60b3bda9\`), ran the suite, and probed the edge cases the change hinges on.

**1. Tier-0 logic correctness — verified.**
- Branch detection: \`CLAUDE_WORKTREE_BRANCH\` env var OR \`git -C <file-dir> branch --show-current\` — both paths exercised.
- \`/\`→\`__\` transform via \`\${WT_BRANCH//\\//__}\` — correct; \`feature/x\` resolves to \`feature__x\`.
- **File-vs-directory duality (the load-bearing concern): confirmed safe.** Ran adversarial cases against a proper git sandbox (the bare \`mktemp\` sandbox without \`git init\` mis-resolves OPS_ROOT and gives false negatives — worth knowing):
  - dir at \`tickets/<project>\` (worktree mode) + matching tier-0 branch, no fallback → **exit 0** (correct pass).
  - dir at \`tickets/<project>\` + NON-matching branch + no fallback → **exit 2 BLOCKED**. This is the critical proof: a *directory* at the tier-1 path reads as "tier-1 absent" via \`[ -f ]\`, so it never falsely satisfies the per-project tier nor errors.
  - dir at tier-1 + non-matching branch + WITH \`current-ticket\` fallback → **exit 0** (falls through cleanly).

**2. No single-agent regression — verified.** No worktree branch ⇒ tier-0 lookup misses (no marker written) ⇒ falls straight to the per-project FILE then fallback. Migration hook's restructured \`if / elif\` with \`[ -z "\$MARKER" ]\` guards preserves the original precedence exactly (manually checked M1 no-marker→block, M2 tier-0 dir resolves, M3 tier-0 dir + wrong branch + no fallback→block, M4 single-agent FILE resolves).

**3. Test quality — genuinely proves isolation, not just happy-path.** Case #14 (branch-B not satisfied by branch-A's \`tickets/myproj/feature__a\` marker, no fallback → BLOCKED) is the real isolation assertion — it proves the collision the PR fixes is actually closed, not merely that the happy path works. \`15/15\` pass locally; \`bash -n\` + \`shellcheck -S error\` clean on both hooks.

**4. Glossary / bullets / AgDR / commit SHA — all pass.** Glossary has 5 well-scoped terms. Summary bullets are narrative (what + why) — no label-only bullets, advisory check clean. AgDR-0066 present and linked, with options-considered table (flock and session-namespace alternatives weighed). Commit message is exemplary (WHY-first body, trade-offs, \`Closes #513\`, co-authorship).

### Handbook Findings
None. The Migration Safety (blocking) handbook triggers on a *migration file* in the diff with destructive SQL — this PR modifies the migration *gate hook*, not any migration file, so it's N/A. Clean Architecture and Commit Message Quality: no violations (commit message in fact models the quality handbook well).

- **Write-side / read-side worktree-detection asymmetry.** \`/start-ticket\` (write) gates the per-worktree path on \`--git-dir != --git-common-dir\` (true linked-worktree only), whereas the read-side hooks treat *any* non-empty \`git branch --show-current\` as tier-0-eligible. This is harmless today — in single-agent mode no per-worktree marker exists, so the read-side tier-0 lookup simply misses and falls to tier 1 — but the two sides describe "is this a worktree?" differently. If a future change ever made the read side *write*-sensitive, the asymmetry could bite. Worth a one-line comment noting the read side is intentionally permissive (tier-0 miss is free).
- **AgDR numbering.** This takes 0066 while 0065 is absent on \`dev\` (presumably claimed by another in-flight PR). 0066 is unique so no collision, but with concurrent AgDR authoring the gap is worth a glance at merge time.
- **Orphaned per-worktree markers** after \`git worktree remove\` are acknowledged in the AgDR as acceptable (same stale-marker behaviour as today). Fine — flagging only so it's on record.

### Verdict
**APPROVED**

The fix is additive, the file-vs-dir duality is correctly handled by \`[ -f ]\` throughout, single-agent behaviour is unchanged, and the new isolation test genuinely proves the collision is closed.

---
📌 Reviewed commit: \`60b3bda91901093b83ee12c215b569a79703e2a6\`

atlas-apex (commented):
## Code Review: PR #524 (re-review)

**Commit**: \`4de2c1153019f6379b25e1941a94001494069dab\`

Re-review of the single follow-up commit (\`4de2c11\`) addressing my prior non-blocking nit — the read/write worktree-detection asymmetry. Previously APPROVED at \`60b3bda\`.

### Summary
The new commit replaces branch-presence-based worktree detection with a **linked-worktree** check that compares the absolute git-dir against the absolute common-dir, and aligns all three sites (\`require-active-ticket.sh\` read side, \`require-migration-ticket.sh\` read side, \`/start-ticket\` write side) on identical logic.

### Verification performed

**(1) Absolute-form comparison is correct — confirmed empirically on git 2.54.0:**
- **Main checkout, from a nested subdir** (`sub/deep`): plain forms give `/…/main/.git` vs `../../.git` (would false-positive); absolute forms give `/…/main/.git` vs `/…/main/.git` → equal → **correctly NOT a worktree**. The prior nit is genuinely fixed.
- **Real linked worktree, from a nested subdir**: absolute git-dir `/…/main/.git/worktrees/wt-linked` vs common-dir `/…/main/.git` → differ → **correctly detected**, branch resolved as `feature/x`.

**(2) Read and write sides use identical detection — confirmed by source inspection:**
- `require-active-ticket.sh:244-245`, `require-migration-ticket.sh:215-216`, and `start-ticket/SKILL.md:152-153` all use `--absolute-git-dir` vs `--path-format=absolute --git-common-dir`. The only (correct) difference: the two read-side hooks anchor with `git -C \"\$_fdir\"` (the edited file's dir, since the hook may run from anywhere), while the skill uses session cwd (it runs interactively inside the worktree). No asymmetry remains.

**(3) Test suite passes at PR HEAD — 16/16:**
Ran `test_require_active_ticket_bash.sh` against a worktree checked out at `4de2c11`. All 16 pass, including the new **case 16** (a real `git worktree add` with NO `CLAUDE_WORKTREE_BRANCH` → tier-0 honored purely via git detection), which exercises the symmetric detection path rather than the env-var shortcut.

**(4) No regression to existing tiers — confirmed:**
- Tier-0 only `exit 0`s when `[ -f \"\$PER_WORKTREE_MARKER\" ]` is true (line 254); absent marker falls straight through to tier-1 (per-project FILE) then tier-2 (`current-ticket`), unchanged.
- `bash -n` clean; `shellcheck -S error` clean on both hooks.
- Cases 1-15 (bootstrap exemptions, redirect/python-write detection, .claude/.md exemptions, per-project file, branch-B isolation) all still pass.

### Note on git ≥ 2.31 (`--path-format=absolute`)
\`--path-format\` landed in git 2.31.0 (March 2021), so the requirement is mild. **Not a blocker**, and worth recording why: on git < 2.31 the `--path-format=absolute` call errors → `_gcd=\"\"`, and since `_gd` is non-empty the `_gd != _gcd` branch evaluates true even in a main checkout. But because the **read and write sides run the identical degraded logic**, they stay symmetric: a single-agent flow on ancient git would write *and* read a `tickets/<project>/<branch>`-shaped marker, so the gate still functions correctly — just via the per-worktree path instead of the per-project file. Graceful, symmetric degradation rather than a gate bypass. If you want to be belt-and-suspenders, a future tweak could fall back to plain `--git-common-dir` + `realpath` when `--path-format` is unsupported, but that's optional polish, not required.

### Checklist Results
- Testing:                   Pass (16/16, new case exercises the real detection path)
- Security:                  Pass (no auth/secret/crypto surface)
- Performance:               Pass (two extra `git rev-parse` only when no env var and PROJECT resolved)
- PR Description & Glossary:  Pass (Glossary present, narrative summary, links #513)
- Technical Decisions (AgDR): Pass (AgDR-0066 present and accurate)
- Adopter Handbooks:         N/A (no handbooks loaded for shell-only diff)

### Issues Found
None.

### Suggestions
- (Optional, non-blocking) Add a `--path-format`-unsupported fallback for git < 2.31 if you ever need to support very old git; the current symmetric-degradation behaviour is acceptable as-is.

### Verdict
**APPROVED**

The nit from the prior review is correctly resolved, detection is now symmetric across all three sites, and the test suite proves both the no-false-positive and the real-worktree-detection properties.

---
📌 Reviewed commit: \`4de2c1153019f6379b25e1941a94001494069dab\`

Issue #526, opened 2026-06-06
[CI] Gate the hook test suite in CI


atlas-apex:
## Driver
The framework has ~63 test scripts — 53 under `.claude/hooks/tests/test_*.sh` plus scattered suites under `.claude/agents/tests/`, `.claude/rules/tests/`, and `.claude/skills/*/tests/` — that assert the mechanical enforcement layer (merge gate, ticket-first + the per-worktree marker tiers, onboarding/secrets/leak guards, branch/PR/commit validators, issue-structure, portfolio path resolution, ops-root pinning, AgDR gates, etc.). **Only 2 of them run in CI** (`test_subpack_extraction` via `extract-subpacks-on-release.yml`, `test_site_counts` via `site-counts-check.yml`). `shellcheck.yml` lints hook *syntax* but never *runs* the suites. So a hook regression ships green — the tests are advisory, not a gate. Several issues this session (review-marker naming, worktree detection) were caught only by running suites manually.

## Scope
- Add `bin/run-hook-tests.sh` — a discovery runner that finds every `test_*.sh` (under `.claude/hooks/tests/` + the scattered `tests/` dirs), runs each, reports per-test pass/fail, and exits non-zero if any included test fails. Reusable locally.
- Add `.github/workflows/tests.yml` — runs the runner on `pull_request` and on `push` to `dev`/`main`, with `jq` + `git` available.
- Quarantine any test that genuinely can't run headless (needs `gh` auth / network / a real release tag) via an explicit skip-list in the runner, each entry annotated with the reason. Log skipped tests so the exclusion is visible, not silent.

## Acceptance Criteria
- [ ] `bin/run-hook-tests.sh` discovers and runs the test suites, prints a per-test PASS/FAIL summary, and exits non-zero on any failure.
- [ ] `.github/workflows/tests.yml` runs the runner on every PR (and push to dev/main); the job goes **red** if any included test fails.
- [ ] Any excluded/quarantined test is listed in the runner with a one-line reason (no silent drops).
- [ ] The runner works locally (`bash bin/run-hook-tests.sh`) with the same result as CI.
- [ ] A deliberately-broken hook makes the job fail (smoke-verified before merge).

## Risks / Dependencies
Some existing tests may rely on local state, network, or `gh` auth and fail headlessly — these must be triaged and either fixed or quarantined (with reason) so the gate is trustworthy rather than flaky. Builds on the test corpus expanded in #513 / #517.

## Glossary
| Term | Definition |
[...]

Issue #517, opened 2026-06-05
[Feature] Keep onboarding config out of git: example-file + gitignore + commit-time guard


atlas-apex:
## User Story
As an adopter (or contributor) configuring a fork, I want my real onboarding/company config to stay local and out of git history **by default**, and I want a safety net that stops me from committing it even if I make a mistake — so private details are never published when my fork is public or when I open a PR upstream.

## Context / Driver
Today `onboarding.yaml` is a **tracked** file that adopters edit in place with real values and commit. Because the framework is forked publicly and contributors open PRs upstream, real configuration can land in public git history / pull-request refs by default. The safe primitives already half-exist — split-portfolio v2 stores config privately, and `.gitignore` already keeps a never-committed onboarding **completion flag** — but single-fork mode (the default) still commits the config itself, and there is no guard that catches a filled-in config before it leaves the machine.

This issue addresses **both layers**:
1. **Structural** — adopt the standard `.env.example` convention so the safe path is the default.
2. **Defensive** — a commit-time guard so a filled-in config (or other sensitive data) can't be committed by accident, the same way `check-secrets.sh` backstops credentials.

### Layer 1 — Structure (example-file + gitignore)
- [ ] Ship a tracked `onboarding.example.yaml` containing only placeholder values.
- [ ] `onboarding.yaml` is gitignored (real, local config).
- [ ] `/setup` copies the example to `onboarding.yaml` (gitignored) and fills it in — never commits real values.
- [ ] `onboarding-check.sh` detects unconfigured state without relying on a committed config file (example presence + the existing never-committed completion marker).
- [ ] Default single-fork mode no longer writes private config to tracked files.

### Layer 2 — Commit-time guard (new)
- [ ] A commit-time guard (PreToolUse on `git commit` + a portable `pre-commit` hook) **blocks** committing `onboarding.yaml` (or any config matching the configured patterns) when its values **differ from the shipped placeholders** in `onboarding.example.yaml` — i.e. it detects a *filled-in* template and stops it.
- [ ] The guard fires for **contributor / cross-fork branches too**, not just the parent — leaks have a path through PRs opened from forks, so the catch must happen at commit time on the contributor's machine, before the data can reach a pull request.
- [ ] Complements (does not duplicate) `check-secrets.sh`: that hook catches credentials; this catches filled-in private **config/identifiers** (company name, internal URLs, tracker instances, named individuals).
[...]

PR #527, opened 2026-06-06
ci(#526): gate the hook test suite in CI

atlas-apex:
## Summary
- **Turns the test corpus into an actual gate** — ~65 `test_*.sh` suites assert the mechanical-enforcement layer (merge gate, ticket-first + per-worktree tiers, onboarding/secrets/leak guards, validators, portfolio paths), but only 2 ran in CI; a hook regression shipped green. This adds a glob-discovery runner + a `tests.yml` workflow that runs the whole suite on every PR and on push to `dev`/`main`, failing the job on any failure.
- **`bin/run-hook-tests.sh`** — discovers every `test_*.sh` / `*.test.sh` under `.claude/**/tests/`, runs each with a per-test timeout, prints PASS/FAIL/SKIP, exits non-zero on any failure. Bash-3.2-portable so it's reusable locally (`bash bin/run-hook-tests.sh`). A documented `QUARANTINE` array skips (and logs) tests that genuinely can't run headless — never a silent drop.
- **`.github/workflows/tests.yml`** — `ubuntu-latest`, installs `jq`, sets a git identity (sandboxes do `git init`/`commit`), runs the runner.
- Decision + quarantine policy in **AgDR-0067**.

## Status / how I'm finalizing the quarantine
A local macOS run showed 52/65 passing; the 12 failures were a mix of macOS `/private/var` symlink artifacts (which pass on Linux) and genuinely stale tests. Rather than guess, this PR runs **everything** so the PR's own `tests` check is the Linux oracle. I'll then:
- leave the gate enforcing all tests that pass on Linux,
- quarantine (with a cited reason + follow-up ticket) only the genuinely-stale/headless-incompatible ones,
- fix any trivially-in-scope failures (e.g. tests that #517 outdated).

So the `tests` check on this PR may be red on the first run by design — I'll drive it green before requesting merge.

## Testing
1. `bash bin/run-hook-tests.sh --list` → discovers 65 tests (verified locally, bash 3.2).
2. `bash bin/run-hook-tests.sh` locally → 52 pass / 12 fail (macOS); failures triaged via this PR's Linux `tests` check.
3. `python3 yaml.safe_load` on `tests.yml` + `shellcheck -S error bin/run-hook-tests.sh` → clean.
4. Pre-merge smoke: a deliberately-broken hook will be confirmed to make the job go red (AC), and the final `tests` check will be green.

Refs #526

---

## Glossary
| Term | Definition |
|------|------------|
| Discovery runner | A script that finds test files by glob and runs them all, instead of a hardcoded list that drifts. |
| Quarantine | An explicit, logged skip-list for tests that can't run headless or are known-failing-and-tracked — visible, not silent. |
| Gated test | A test whose failure fails CI (blocks merge), vs an advisory test that only runs if someone remembers. |
| Linux oracle | Using the actual CI environment to determine the true pass/fail set, free of local macOS artifacts. |

### DIFF
--- a/.claude/skills/setup/tests/test_setup_reset.sh
+++ b/.claude/skills/setup/tests/test_setup_reset.sh
@@ -22,7 +22,11 @@
 set -u
 
 ROOT="$(cd "$(dirname "$0")/../../../.." && pwd)"
-TEMPLATE_SRC="$ROOT/onboarding.yaml"
+# Since #517 the tracked placeholder template is onboarding.example.yaml
+# (onboarding.yaml is gitignored real config). /setup --reset now does
+# `cp onboarding.example.yaml onboarding.yaml`, so the template source here is
+# the example file.
+TEMPLATE_SRC="$ROOT/onboarding.example.yaml"
 
 if [ ! -f "$TEMPLATE_SRC" ]; then
   echo "FAIL: framework template $TEMPLATE_SRC not found" >&2
@@ -33,7 +37,7 @@ fi
 # placeholder. If this ever changes, this test (and the SKILL.md
 # detection grep) must change too.
 if ! grep -q '"Your Company Name"' "$TEMPLATE_SRC"; then
-  echo "FAIL: framework onboarding.yaml lacks the 'Your Company Name' placeholder — template detection grep in SKILL.md Step 1 is broken" >&2
+  echo "FAIL: framework onboarding.example.yaml lacks the 'Your Company Name' placeholder — template detection grep in SKILL.md Step 1 is broken" >&2
   exit 1
 fi
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,38 @@
+name: tests
+
+# Runs the framework's mechanical-enforcement test suites (the ~60 test_*.sh
+# under .claude/**/tests/) on every PR and on push to dev/main. Until now these
+# tests were advisory — only 2 ran in CI — so a hook regression could ship
+# green. This job makes the whole suite a gate. See me2resh/apexyard#526.
+
+on:
+  pull_request:
+  push:
+    branches: [dev, main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  hook-tests:
+    name: hook test suite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install jq (used by the hooks + their tests)
+        run: sudo apt-get update -qq && sudo apt-get install -y -qq jq
+
+      - name: Configure git identity (test sandboxes run git init/commit)
+        run: |
+          git config --global user.email "ci@apexyard.test"
+          git config --global user.name "apexyard-ci"
+          git config --global init.defaultBranch main
+
+      - name: Run the hook test suite
+        run: bash bin/run-hook-tests.sh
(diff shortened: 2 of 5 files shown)


atlas-apex (commented):
## Code Review: PR #527 — APPROVED (posted as comment; cannot self-approve own PR)

**Commit**: `44604ffb0195944d47a57135af6b4f6922ec95d0`

### Summary
Turns the ~65-script mechanical-enforcement test corpus into an actual CI gate. Adds a bash-3.2-portable glob-discovery runner (`bin/run-hook-tests.sh`), a `tests.yml` workflow (PR + push to dev/main), AgDR-0067 (decision + quarantine-as-ratchet policy), and two in-scope fixes to land the gate green. The PR's own `hook test suite` check is GREEN at HEAD (PASS=60, FAIL=0, SKIP=5).

### Checklist Results
- Testing:                   Pass (this PR *is* the testing infrastructure)
- Security:                  Pass (`permissions: contents: read`, minimal scope)
- Performance:               Pass (per-test 120s cap; concurrency cancel-in-progress)
- Summary Bullet Narrative:  Pass (narrative, each with rationale)
- Technical Decisions (AgDR):Pass (AgDR-0067 present + linked; Refs #526)
- Adopter Handbooks:         N/A (no migration/TS/domain code in a shell-only diff)

### Verification performed (against PR HEAD, not local tree)
1. **CI green at HEAD** — `hook test suite` check = success (PASS=60/FAIL=0/SKIP=5). The gate fails-on-real-failures by design: `set -uo pipefail` (not `set -e`, so failures aggregate rather than abort the loop), `rc=$?` captured, `exit 1` when `fail>0`.
2. **Tracking issue #528 is real + well-formed** — OPEN, lists all 5 quarantined tests with root causes and an AC that drains QUARANTINE to empty. A genuine ratchet, not a graveyard.
3. **Quarantine legitimacy spot-checked** — `test_token_efficiency_wave1.sh` asserts a 200-char SKILL-description hard cap (matches cited "plan-initiative >200 chars" drift); `test_handover_clone_prompt.sh` asserts pre-restructure clone-prompt strings (`[y / n / later]`, "Offer the clone-first") which the SKILL has since moved. Both genuine pre-existing drift, not caused by this PR. All 5 SKIP-logged with cited reasons — visible, never silent.
4. **test_setup_reset fix aligns to #517** — at PR HEAD `onboarding.yaml` is NOT tracked (404 via contents API), `onboarding.example.yaml` exists (4640 bytes) with the `\"Your Company Name\"` placeholder, and `/setup --reset` does `cp onboarding.example.yaml onboarding.yaml` (SKILL.md L80). Test now reads the correct source.
5. **CLAUDE.md /handover trim accurate** — 35→24 words; preserves the three substantive facts (5-dimension scoring, checklist doc-pick, Next-Steps ticket offer). No information loss.

### Runner correctness (focus #1)
- bash 3.2 portable — `while IFS= read` + `< <(...)` instead of `mapfile`; POSIX param expansions. Good.
- Empty-QUARANTINE guard under `set -u` — `is_quarantined` short-circuits on `${#QUARANTINE[@]} -gt 0` before iterating, so draining the array in #528 won't trip an unbound-array expansion. Forward-looking + correct.
- timeout fallback — `timeout`→`gtimeout` probe, empty default, intentional unquoted word-split (SC2086 disabled). Correct.
- exit codes — aggregates, prints failed list, exits 1 on any failure. Correct.

### Issues Found
None blocking.

- nit: the main run loop `for t in \"${TESTS[@]}\"` lacks the empty-array guard that `is_quarantined` and `--list` have. Under `set -u`+bash 3.2 a zero-test discovery would expand as unbound. Not reachable today (60+ tests) — but a one-line `[ \"${#TESTS[@]}\" -gt 0 ] || { echo \"no tests discovered\"; exit 1; }` would turn a silent-empty-pass into a loud failure, the safer direction for a gate.

### Verdict
**APPROVED** (Rex first-pass). Approval marker written for the merge gate. Human approver does the second review before merge.

---
📌 Reviewed commit: `44604ffb0195944d47a57135af6b4f6922ec95d0`

PR #531, opened 2026-06-06
test(#528): drain quarantine — pin-escape audit + handover rewrite

### LINKED ISSUE #528: [Testing] Fix & un-quarantine the 5 pre-existing hook-test failures
## Driver
#526 turned the hook test suite into a CI gate. To land the gate green, 5 tests that were **already failing on `dev`** (pre-existing drift, not caused by #526) were quarantined in `bin/run-hook-tests.sh`. They represent real test debt and should be fixed and un-quarantined so the gate covers them too.

## Scope
For each, fix the root cause, then remove its entry from the `QUARANTINE` array in `bin/run-hook-tests.sh`:
- **`test_agent_routing_sync_and_drift.sh`** — "case 2: qa-engineer override" fails; reconcile the test with the committed `agent-routing.yaml` override behaviour.
- **`test_handover_clone_prompt.sh`** — asserts the *pre-restructure* clone-prompt spec (literal strings like `[y / n / later]`, "Offer the clone-first"); the `/handover` SKILL moved the clone to step 1.5 and the follow-up offer to step 8. Update the assertions to the current spec.
- **`test_harnessability_scoring.sh`** — 1 of 14 scoring cases drifted; realign the expected score with the current rubric in the SKILL.
- **`test_md_to_pdf_fallback.sh`** — on a runner WITH `npx` it runs `npx -y md-to-pdf` (network npm install + headless chromium), which is heavy/flaky in CI; make it skip unless a converter is actually provisioned, or provision one deterministically.
- **`test_token_efficiency_wave1.sh`** — two doc-hygiene drifts: `plan-initiative` SKILL.md description is 318 chars (>200 hard cap), and `/release-sync` exists under `.claude/skills/` but is missing from the CLAUDE.md skill table.

## Acceptance Criteria
- [ ] Each of the 5 root causes is fixed.
- [ ] Each entry is removed from `QUARANTINE` in `bin/run-hook-tests.sh`.
- [ ] `bash bin/run-hook-tests.sh` is green on Linux CI with an **empty** quarantine list.
- [ ] No test is silently disabled — the quarantine array ends empty.

## Glossary
| Term | Definition |
|------|------------|
| Quarantine | The documented skip-list in `bin/run-hook-tests.sh` for tests excluded from the gate; this ticket drains it to empty. |
| Pre-existing drift | A test that was already failing on `dev` before the gate existed — debt the gate surfaced, not caused. |

atlas-apex:
## Summary
Completes #528 — **drains the quarantine to empty** by (a) fixing the systemic pin-escape class you asked me to audit, and (b) rewriting the last quarantined test against the redesigned `/handover` spec.

> **Stacked on #530** (base = `test/GH-528-agent-routing`). Merge #530 first; GitHub will retarget this to `dev`. The diff here is just the 3 files below.

**(a) Pin-escape audit + fix.** Many hooks resolve their ops-root via `_lib-ops-root.sh`, which inside a live Claude Code session honours the session pin and resolves to the **real fork** — so sandbox tests escape onto the real repo (wrong results, and *real-file mutation* for writing hooks like `apply-agent-routing`/`link-custom-skills`). 21 hooks use the pin; only 4 tests guarded against it.
- **Systemic fix:** `bin/run-hook-tests.sh` now `export APEXYARD_OPS_DISABLE_PIN=1`, so the whole suite resolves by walk-up to its own sandbox. No-op in headless CI (no pin). The one pin-exercising test (`test_resolve_ops_root_pin`) sets the var per-case, so it's unaffected. This alone turned the local run from **7 spurious failures → green** (those weren't macOS-symlink artifacts after all — same pin-escape).
- **Belt-and-suspenders:** `test_link_custom_skills` (writes symlinks into `<ops-root>/.claude/skills/`) gets its own guard so a *standalone* in-session run can't symlink into the real repo.

**(b) Rewrote `test_handover_clone_prompt`.** The old assertions pinned a removed `[y / n / later]` clone prompt. The `/handover` SKILL was redesigned: clone-by-default at **step 1.5-clone** (skip-if-`.git`-exists, `--no-clone` to decline) + a **step-8** follow-up-skills offer. New spec-asserts pin the current literals; the runtime simulator now exercises the clone-by-default branch (default→clones, `--no-clone`→skips, existing-`.git`→skips). **18/18.**

**Result:** `QUARANTINE` is now **empty** — the gate enforces all **65** tests.

## Testing
1. `bash bin/run-hook-tests.sh` → **PASS=65, FAIL=0, SKIP=0** locally (was 7 failing before the pin guard).
2. `bash .claude/hooks/tests/test_handover_clone_prompt.sh` → 18/18.
3. `test_resolve_ops_root_pin` + `test_link_custom_skills` pass under the suite default; verified no real-repo mutation after runs.
4. `shellcheck -S error` clean on all three changed files.
5. This PR's `tests` gate on Linux is the check (expected green, enforcing 65).

Closes #528

---

## Glossary
| Term | Definition |
|------|------------|
| Session pin | `_lib-ops-root.sh` resolving the ops root from `$APEXYARD_OPS_PIN_DIR/ops-root-$CLAUDE_CODE_SESSION_ID` inside a live session. |
| `APEXYARD_OPS_DISABLE_PIN` | Env flag that disables the pin so resolution walks up to the sandbox — the correct neutralizer for hermetic tests. |
| Quarantine | The runner's documented skip-list; now empty — every discovered test is enforced. |
| Spec-pin test | A test asserting literal substrings in a SKILL doc; rewritten here to the current clone-by-default design. |

### DIFF
--- a/.claude/hooks/tests/test_handover_clone_prompt.sh
+++ b/.claude/hooks/tests/test_handover_clone_prompt.sh
@@ -1,24 +1,25 @@
 #!/bin/bash
-# Smoke tests for /handover's clone-first deep-dive prompt
-# (me2resh/apexyard#188 — LSP spike Option 3).
+# Smoke tests for /handover's clone behaviour.
 #
-# /handover is a markdown skill spec the model executes; this test
-# exercises two things:
+# History: the original design (me2resh/apexyard#188) offered a clone-first
+# deep-dive prompt at step 8 with a `[y / n / later]` choice. That was
+# REDESIGNED — the repo is now cloned by DEFAULT at step 1.5-clone (no prompt;
+# `--no-clone` opts out), and step 8 became a follow-up-skills offer against the
+# already-cloned repo. This test was rewritten (#528) to pin the current spec
+# instead of the removed prompt.
 #
-#   1. SPEC SHAPE — the SKILL.md file contains the prompt text at the
-#      right location (after step 7 "Append to the portfolio registry",
-#      before what becomes the final summary), and surfaces the five
-#      cost-transparency facts the ticket's AC requires.
+# /handover is a markdown skill spec the model executes; this test exercises:
 #
-#   2. RUNTIME SHAPE — a small bash simulator that mirrors the clone
-#      branch the spec documents, exercised against an isolated sandbox
-#      with a mocked `git` binary so no network call ever fires. Verifies:
+#   1. SPEC SHAPE — SKILL.md documents clone-by-default at step 1.5-clone
+#      (skip-if-`.git`-exists, `--no-clone` to decline), the step-8 follow-up
+#      offer, the LSP cost-transparency facts, and the surrounding step order.
 #
-#        - operator response `n`     → no `git clone` invocation, exit clean
-#        - operator response `later` → no `git clone` invocation, exit clean
-#        - operator response `y`     → exactly one `git clone <url> workspace/<name>`
-#                                       invocation with the documented argument shape
-#        - workspace already exists  → no `git clone` invocation (skip-if-exists)
+#   2. RUNTIME SHAPE — a small bash simulator mirroring the spec's clone branch,
+#      run against an isolated sandbox with a mocked `git` (no network). Verifies:
+#        - default (no --no-clone), workspace absent → exactly one
+#          `git clone <url> <workspace>/<name>`
+#        - `--no-clone`                              → no git clone (declined)
+#        - workspace/<name>/.git already exists      → no git clone (skip)
 #
 # Exit 0 if all cases pass; 1 on first failure.
 
@@ -37,7 +38,7 @@ FAIL=0
 FAILED_CASES=""
 
 # ---------------------------------------------------------------------------
-# spec_assert <label> <expected literal substring> [grep-extended-regex flag]
+# spec_assert <label> <expected literal substring>
 # ---------------------------------------------------------------------------
 spec_assert() {
   local label="$1" needle="$2"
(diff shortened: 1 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #531

**Commit**: \`5d2f3b7006f8ef1b73362cd67f3e435b6725ba38\`

### Summary
Drains the test-runner QUARANTINE to empty, completing #528. Three-file delta: (a) `bin/run-hook-tests.sh` exports `APEXYARD_OPS_DISABLE_PIN=1` suite-wide + empties the QUARANTINE array; (b) `test_link_custom_skills.sh` gets a standalone pin guard; (c) `test_handover_clone_prompt.sh` is rewritten against the redesigned clone-by-default `/handover` spec.

### Verification performed
Fetched PR HEAD (`5d2f3b70`) and exercised every claim against the actual tree:

1. **Runner-level pin disable — correct & safe.** `export APEXYARD_OPS_DISABLE_PIN=1` at suite top is the right neutralizer (forces walk-up to the per-test sandbox; no-op in headless CI). Confirmed `test_resolve_ops_root_pin.sh` is immune: cases 1 & 2 run in subshells that explicitly `unset APEXYARD_OPS_DISABLE_PIN` (lines 89, 115), case 3 explicitly `export`s it (line 145), each followed by an `unset`. The suite-level export cannot leak into those per-case subshells. Ran it: **8 passed, 0 failed.**

2. **Handover rewrite — spec-asserts match & simulator is faithful.** All 15 `spec_assert` needles were verified present in `.claude/skills/handover/SKILL.md` via literal `grep -F` (step-1.5-clone heading, default-yes prose, `WORKSPACE_DIR=$(portfolio_workspace_dir)`, the `.git`-existence skip guard, the clone command, both `CLONE_STATUS` markers, the step-8 follow-up heading + options, both cost facts, and the step 7/9/10 order anchors). The runtime simulator's branch logic (`--no-clone`→declined, `.git`-exists→preserved, else→clone) mirrors the documented step-1.5-clone branch byte-for-byte including branch order and the `.git` check. Ran the test: **18 passed, 0 failed.**

3. **QUARANTINE drain — justified.** All five formerly-quarantined tests (`token_efficiency_wave1`, `harnessability_scoring`, `md_to_pdf_fallback`, `agent_routing_sync_and_drift`, `handover_clone_prompt`) were run individually and each PASSES. They are genuinely enforced by the suite, not silently dropped. Full suite at PR HEAD: **PASS=65 FAIL=0 SKIP=0 TOTAL=65** — matches the claim exactly.

4. **PR hygiene.** Glossary present (4 terms). Summary bullets are narrative (what + why, not label-only). `Closes #528` links the ticket. Testing instructions present. AgDR correctly N/A — pure test maintenance, no technical decision introduced. shellcheck `-S error` clean on all three changed files; CI all green (hook test suite, shellcheck, verify-ticket-id, site-counts).

### Checklist Results
- Architecture & Design:    Pass (test-only; no layering impact)
- Code Quality:              Pass (shellcheck clean, faithful simulator, well-commented isolation rationale)
- Testing:                   Pass (this PR IS test maintenance; 65/65 enforced, quarantine empty)
- Security:                  Pass (the pin guard actively prevents sandbox tests mutating the real fork)
- Performance:               Pass (N/A)
- Technical Decisions (AgDR):N/A (test maintenance)
- Adopter Handbooks:         N/A (no handbooks triggered by this diff)

### Issues Found
None.

### Suggestions
None blocking. The inline isolation-rationale comments in both the runner and `test_link_custom_skills.sh` are excellent — they make the pin-escape failure mode legible to future readers.

### Verdict
**APPROVED**

The pin-escape audit is a genuine correctness fix (sandbox tests were escaping onto the real fork — a real-file-mutation risk for the writing hooks), the handover rewrite faithfully tracks the redesigned spec, and the quarantine drain is honest: every test is verified passing, not skipped. Merge #530 first (this is stacked); GitHub will retarget to `dev`.

---
📌 Reviewed commit: \`5d2f3b7006f8ef1b73362cd67f3e435b6725ba38\`

Issue #550, opened 2026-06-06
[Bug] release tag can land off main's ancestry (v2.3.0 mis-tagged)


atlas-apex:
## Given / When / Then

**Given** the release-cut model where `dev` is squash-merged to `main` and a `vN.N.N` tag is then created,
**When** the tag is placed on the **release-branch HEAD** (the pre-squash commit) instead of the **squash commit that actually landed on `main`**,
**Then** the tag points off `main`'s ancestry — `git describe --tags upstream/main` fails with *"No tags can describe …"*, `git merge-base --is-ancestor <tag> upstream/main` is false, and "fork from the latest release" / `git describe`-based tooling breaks.

This is the current state of **`v2.3.0`** on the framework repo: the tag commit is not in `main`'s history (the real v2.3.0 release commit on main is a different SHA). `v3.0.0` was tagged correctly (on `upstream/main`), so the two coexist inconsistently.

## Repro

1. `git fetch upstream --tags`
2. `git describe --tags --abbrev=0 upstream/main` → fails: *No tags can describe '<main HEAD>'*.
3. `git merge-base --is-ancestor v2.3.0 upstream/main; echo $?` → non-zero (not an ancestor).
4. By contrast `git merge-base --is-ancestor v3.0.0 upstream/main; echo $?` → 0.

## Expected

The `/release` tagging step must tag the **merged commit on `main`** (`git tag vN.N.N upstream/main` *after* the release PR squash-merges), never the release-branch HEAD. Worth adding a guard/post-check: after tagging, assert `git merge-base --is-ancestor vN.N.N upstream/main` and warn if false. The existing mis-placed `v2.3.0` tag should also be re-pointed at its real squash commit on `main`.

## Environment

- ApexYard framework, `/release` skill (tag + push step) and/or manual release tagging
- Observed on `me2resh/apexyard`: `v2.3.0` off-ancestry, `v3.0.0` correct.

## Severity

Medium — breaks `git describe`, "default to latest release" assumptions, and any ancestry-based tooling; not data-affecting. Fix is a tagging-target correction + a one-time re-point of `v2.3.0`.

## Glossary

| Term | Definition |
|------|------------|
| Squash commit | The single commit created on `main` when a release PR is squash-merged; the tag must point here. |
| Ancestry / `git describe` | Tag discoverability depends on the tag being a reachable ancestor of the branch tip. |

zeyadsleem:
[...]

PR #565, opened 2026-06-07
fix(#550): tag the squash commit on main + ancestry guard in /release

atlas-apex:
## Summary

- **Step 6 now targets the squash commit, not the release-branch HEAD** — after a squash-merge, the release branch's HEAD is a different SHA that is no longer in `main`'s ancestry. The updated bash block fetches `upstream` first, then tags `upstream/main` (the exact squash commit), making the mis-tag mechanically impossible when the prescribed commands are followed.
- **Mandatory ancestry guard added before the tag is pushed** — `git merge-base --is-ancestor vA.B.C upstream/main` runs immediately after tagging and aborts with a descriptive error if the tag is off-ancestry. This catches any remaining mis-tag (e.g. a manual override) before it reaches the public remote.
- **Post-tag release checklist added** — three assertions (`merge-base`, `git describe`, and exact-tip equality) give the operator a repeatable verification step before announcing the release, preventing silent ancestry drift.

> **Out of scope:** Re-pointing the existing `v2.3.0` tag is an operational, published-tag rewrite that must be done by the maintainer (force-updating a public tag). This PR does not touch it.

## Testing

1. `npx markdownlint-cli2 .claude/skills/release/SKILL.md` — confirmed 0 errors.
2. No shell unit tests exist for the `/release` skill (it is a markdown/prose spec, unlike `/release-sync` which has `test_release_sync.sh`). Verification is markdownlint-clean output + correctness of the guard logic.
3. Guard logic correctness: `git merge-base --is-ancestor <tag> upstream/main` exits 0 only when the tag commit is reachable from `upstream/main`. Tagging `upstream/main` directly (as prescribed) always satisfies this, while tagging the release-branch HEAD (the previous accidental pattern) does not — confirmed by the repro steps in #550.

Refs #550

---

## Glossary

| Term | Definition |
|------|------------|
| Squash commit | The single commit GitHub creates on `main` when a PR is squash-merged; its SHA differs from every commit on the source branch and is the only valid tagging target after a squash-merge. |
| Ancestry / `git describe` | A tag is "on ancestry" when it is a reachable ancestor of the branch tip; `git describe --tags upstream/main` and `git merge-base --is-ancestor` both depend on this property to work correctly. |
| Release-branch HEAD | The tip commit of the `release/vX.Y.Z` branch before it is merged; after a squash-merge this commit is not in `main`'s history and must not be used as a tagging target. |
| Ancestry guard | The `git merge-base --is-ancestor` assertion added in step 6 that aborts the tag-push workflow if the newly created tag is not reachable from `upstream/main`, preventing a mis-placed tag from reaching the public remote. |

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -149,13 +149,41 @@ The release PR runs through the normal flow:
 
 ### 6. Tag + push (after merge)
 
-Once the release PR merges, the user invokes `/release --tag vA.B.C` (or runs the suggested commands manually):
+Once the release PR squash-merges to `main`, the tag must point at the **squash commit that landed on `main`** — never at the release-branch HEAD. When a PR is squash-merged, GitHub creates a single new commit on `main` whose SHA is different from every commit on the source branch. Tagging the release-branch HEAD produces a tag that is not reachable from `main`, breaking `git describe`, `git merge-base`, and any ancestry-based tooling (see issue #550).
+
+The user invokes `/release --tag vA.B.C` (or runs the suggested commands manually):
 
 ```bash
-git fetch upstream main
+# Step 1 — fetch so upstream/main points at the squash commit just merged.
+git fetch upstream
+
+# Step 2 — tag the tip of upstream/main.
+#           This IS the squash commit. Do NOT use the release-branch HEAD —
+#           it was discarded by the squash and is no longer in main's ancestry.
 git tag vA.B.C upstream/main
+
+# Step 3 — ancestry guard: assert the tag is reachable from main BEFORE pushing.
+#           If this exits non-zero, the tag is mis-placed — delete and re-tag.
+if ! git merge-base --is-ancestor vA.B.C upstream/main; then
+  echo "ERROR: vA.B.C is not an ancestor of upstream/main — tag is mis-placed." >&2
+  echo "Delete the tag with: git tag -d vA.B.C" >&2
+  echo "Then re-run from Step 1 after confirming the PR was squash-merged." >&2
+  exit 1
+fi
+
+# Step 4 — push only after the guard passes.
 git push upstream vA.B.C
 ```
+
+#### Post-tag release checklist
+
+Before announcing the release, verify all three assertions hold:
+
+- [ ] `git merge-base --is-ancestor vA.B.C upstream/main` exits 0 (tag is an ancestor of main)
+- [ ] `git describe --tags --abbrev=0 upstream/main` returns `vA.B.C` (tag is discoverable from main)
+- [ ] `git rev-parse vA.B.C^{}` equals `git rev-parse upstream/main` (tag points at the exact tip — true for a clean release with no post-merge commits; will differ if post-merge work has landed, which is fine as long as the first two assertions hold)
+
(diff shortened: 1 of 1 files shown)


atlas-apex (commented):
## Code Review: PR #565

**Commit**: \`1c7b261f3a94a7c2af93308f6a8c4744e0118f7c\`

### Summary
Rewrites step 6 (Tag + push) of the `/release` skill so the `vA.B.C` tag is placed on the squash commit that lands on `main` (`git tag vA.B.C upstream/main`) rather than the release-branch HEAD, adds a pre-push ancestry guard (`git merge-base --is-ancestor`, `exit 1` on failure), and a post-tag verification checklist. Rule #6 is updated to match. Fixes the mis-tag class from #550. The one-time `v2.3.0` re-point is correctly scoped out.

### Checklist Results
- Architecture & Design:     Pass (docs/skill-only)
- Testing:                   Pass (markdownlint clean; guard logic verified empirically)
- PR Description & Glossary: Pass (4-term glossary, narrative summary bullets)
- Technical Decisions (AgDR):N/A (no new library/pattern/architecture)

**Tagging target + guard — correct & unambiguous.** Tags `upstream/main` (explicitly named as "the squash commit"), with prose stating "never the release-branch HEAD." The guard asserts `git merge-base --is-ancestor vA.B.C upstream/main`.

**Bash sound — fetch → tag → guard → push.** Ordering is correct: the guard runs and `exit 1`s *before* `git push upstream vA.B.C`, so a mis-tag can never reach the public remote. Dropping the `main` arg from `git fetch upstream main` → `git fetch upstream` is an improvement — fetching all refs reliably updates the `upstream/main` remote-tracking ref the next two commands depend on.

**Squash mechanics prose — correct.** Accurately states that a squash-merge creates a single new commit on `main` whose SHA differs from every source-branch commit, and that the release-branch HEAD is therefore no longer in `main`'s ancestry (breaking `git describe` / `git merge-base`).

**Checklist assertions — verified empirically.** `git merge-base --is-ancestor` is reflexive (exit 0 when tag == tip), so tagging `upstream/main` directly always satisfies the guard. `git rev-parse vA.B.C^{}` peels a lightweight tag to its commit and equals `git rev-parse upstream/main` at a clean release; the third bullet correctly caveats the post-merge-commits case.

**No contradiction with the rest of the skill.** Rule #6 now reads "never tag before merge, and never tag the release-branch HEAD" and cross-references step 6 — consistent. Step 3.5 (pre-merge site-version bump in the release commit) and step 9 (`/release-sync vA.B.C` merging `upstream/main` into `dev`) both align with the new step 6's `upstream/main` target; no overlap or conflict.

**markdownlint:** 0 errors (`markdownlint-cli2 v0.22.1`).

**No leaks:** only `me2resh/apexyard` + generic version placeholders.

### Issues Found
None.

### Suggestions
- nit (non-blocking): the `--tag vA.B.C` invocation in the step-6 lead-in still implies `/release --tag` runs these commands; the manual block is what actually changed. Fine as-is — the parenthetical "(or runs the suggested commands manually)" covers it.

### Verdict
**APPROVED**

---
📌 Reviewed commit: \`1c7b261f3a94a7c2af93308f6a8c4744e0118f7c\`

Issue #547, opened 2026-06-06
[Bug] validate-branch-name.sh misreads git-push refspecs and redirections as branch names


atlas-apex:
## Given / When / Then

**Given** an operator pushes with `validate-branch-name.sh` wired as a `PreToolUse` hook on `Bash(git push *)`,
**When** the push command uses a **refspec** (`git push <remote> HEAD:<branch>`) or a **shell redirection** (`git push <remote> --tags 2>&1 | tail`),
**Then** the hook extracts the wrong token as the "branch" and blocks the push:
- `git push upstream HEAD:feature/GH-1-foo` → rejected with `Branch 'HEAD' doesn't follow naming convention`
- `git push upstream --tags 2>&1 | tail` → rejected with `Branch '2>' doesn't follow naming convention`

Both are legitimate pushes; the hook's branch-name parser is grabbing `HEAD` (the left side of the refspec) and `2>` (the redirection operator) instead of the real branch ref.

## Repro

1. From a fork with `upstream` configured, run `git push upstream HEAD:some/valid-branch` → blocked, complains about branch `HEAD`.
2. Run `git push upstream --tags 2>&1 | tail -5` → blocked, complains about branch `2>`.
3. Run `git push upstream <validbranch>` (no refspec, no redirection) → passes.

## Expected

The hook should parse the **destination ref** of the push, ignoring (a) the left side of a `src:dst` refspec, (b) shell redirections / pipes appended to the command, and (c) tag pushes (`--tags`, `tag <name>`) which have no branch at all. A tag push should be a no-op for a *branch*-name validator.

## Environment

- ApexYard framework, hook `.claude/hooks/validate-branch-name.sh`
- Wired via `.claude/settings.json` matcher `Bash(git push *)`
- Hit live during the v3.0.0 release flow (worked around by reformulating the push commands).

## Severity

Medium — blocks valid pushes (refspec pushes and any `git push … 2>&1` form). Reproducible, no data impact, but it derails normal git usage and tag pushes.

## Glossary

| Term | Definition |
|------|------------|
| Refspec | A `git push` argument of the form `<src>:<dst>` (e.g. `HEAD:main`) specifying source and destination refs. |
| Branch-name validator | The hook enforcing `{type}/{TICKET-ID}-{desc}` on the branch being pushed. |

zeyadsleem:
Closing per maintainer request: current open ticket queue is being cleared/parked. Reopen if this work should remain active.

[...]

PR #566, opened 2026-06-07
fix(#547): parse git-push dst ref, ignore redirections + tag pushes

atlas-apex:
## Summary

- **Extracted DST from `src:dst` refspecs instead of SRC** — `git push upstream HEAD:feature/GH-1-foo` was blocked because the hook read `HEAD` (left side) as the branch name. The fix makes `extract_push_ref()` take the right side of the colon, which is what actually lands on the remote and what the branch-naming convention applies to.
- **Strip shell redirections before token parsing** — a command like `git push upstream --tags 2>&1 | tail` had `2>` mis-parsed as a positional branch-name token. The lib now strips everything from the first redirection operator (`NNN>`, `|`) onward before extracting tokens, so `2>`, `>`, and `|` can never be mistaken for branch names.
- **Tag pushes are a no-op for the branch-name validator** — added `is_tag_push()` to `_lib-extract-push-ref.sh` (detects `--tags`, `tag <name>`, `refs/tags/`) and call it in `validate-branch-name.sh` before any ref extraction; exit 0 immediately. A branch-name validator has nothing to validate on a tag push.

## Testing

```bash
bash .claude/hooks/tests/test_validate_branch_name_pushref.sh
```

28 cases total — 15 pre-existing (#194 behaviour) + 13 new (#547 cases). All pass:

```
PASS [#547: HEAD:valid-branch validates dst → passes]
PASS [#547: HEAD:non-conforming-dst blocks on dst]
PASS [#547: sha:dst validates dst → passes]
PASS [#547: local-branch:remote-branch validates remote-dst → passes]
PASS [#547: --tags plain → skip (no branch to validate)]
PASS [#547: --tags with 2>&1 pipe → skip (no branch to validate)]
PASS [#547: --tags before remote → skip]
PASS [#547: tag <name> keyword form → skip]
PASS [#547: refs/tags/ refspec → skip]
PASS [#547: plain 2>&1 redirect on valid push → passes]
PASS [#547: > redirect on valid push → passes]
PASS [regression: explicit ref passes (no refspec)]
PASS [regression: non-conforming explicit ref still blocks]
===================================
  PASS: 28   FAIL: 0
===================================
```

Refs #547

---

## Glossary

| Term | Definition |
|------|------------|
| Refspec | A `git push` argument of the form `<src>:<dst>` (e.g. `HEAD:feature/GH-1-foo`) specifying the source local ref and the destination remote ref. The branch-naming convention applies only to the dst — what lands on the remote. |
| Shell redirection | Shell operators like `2>&1`, `2>`, `>` that redirect file descriptors. They appear as tokens in the raw command string the hook receives and were previously mis-parsed as branch-name candidates. |
| Tag push | A push that sends git tags to the remote rather than a branch ref. Indicated by `--tags`, `tag <name>`, or a `refs/tags/` refspec. A branch-name validator has no rule to apply and must exit 0. |
| `is_tag_push()` | New helper in `_lib-extract-push-ref.sh` that returns true when the command is a tag push. Called before `extract_push_ref()` so the tag-push guard is correct even when redirections are appended to the command. |
| `extract_push_ref()` | Shared lib function (existing, in `_lib-extract-push-ref.sh`) that parses a `git push` command and returns the branch ref to validate. Updated to return the DST of a refspec and to strip redirections before parsing. |

### DIFF
--- a/.claude/hooks/_lib-extract-push-ref.sh
+++ b/.claude/hooks/_lib-extract-push-ref.sh
@@ -25,23 +25,82 @@
 # -----
 #   . "$(dirname "$0")/_lib-extract-push-ref.sh"
 #   PUSH_REF=$(extract_push_ref "$COMMAND")
+#   # Check is_tag_push first — tag pushes must be no-ops for a branch validator.
+#   if is_tag_push "$COMMAND"; then exit 0; fi
 #   BRANCH="${PUSH_REF:-$(git branch --show-current)}"
 #
-# Refs: me2resh/apexyard#194
+# Refs: me2resh/apexyard#194, me2resh/apexyard#547
 
-# Echoes the source ref from a `git push` command, or empty if none found.
+# Returns true (0) when the command is a tag push that has no branch ref at all.
+# Tag pushes must be a no-op for a *branch*-name validator.
+#
+# Detects:
+#   git push <remote> --tags              → true
+#   git push <remote> tag <name>          → true
+#   git push --tags                       → true
+#   git push <remote> refs/tags/<name>    → true
+#   git push <remote> +refs/tags/<name>:refs/tags/<name>  → true (refspec targeting tags/)
+#   git push <remote> v1.0.0              → NOT detected (bare tag name = branch name,
+#                                            validator falls back to local HEAD)
+is_tag_push() {
+  local cmd="$1"
+
+  # Strip everything after the first shell redirection/pipe so the check isn't
+  # fooled by `2>&1 | tail` appended to the command.
+  local clean_cmd
+  clean_cmd=$(echo "$cmd" | sed 's/[[:space:]]*[0-9]*[>|][&>|0-9].*$//')
+
+  # --tags flag: `git push [opts] --tags [remote]`
+  if echo "$clean_cmd" | grep -qE '\bgit\s+push\b.*\s--tags(\s|$)'; then
+    return 0
+  fi
+
+  # `git push <remote> tag <name>` (explicit `tag` keyword before ref)
+  if echo "$clean_cmd" | grep -qE '\bgit\s+push\b.*\stag\s+\S'; then
+    return 0
+  fi
+
+  # refs/tags/ in any positional argument (e.g. push by full ref or refspec)
+  if echo "$clean_cmd" | grep -qE '\brefs/tags/'; then
+    return 0
+  fi
+
+  return 1
+}
+
+# Echoes the DESTINATION branch ref from a `git push` command, or empty if
+# none found.
+#
+# KEY BEHAVIOUR CHANGES vs the original (me2resh/apexyard#547):
+#
+#   1. Shell redirections and pipes are stripped BEFORE token parsing so that
+#      `2>&1`, `2>`, `>`, `|` tokens are not mistaken for branch names.
+#
+#   2. For a `src:dst` refspec (e.g. `HEAD:feature/GH-1-foo`), the DESTINATION
+#      (right of the colon) is returned for validation, not the source. The
+#      source side may be `HEAD`, a SHA, or an arbitrary local ref name that
+#      the branch-name validator should not block on — the dst is what actually
+#      lands on the remote.
+#
+#   3. Tag pushes are NOT handled here — call is_tag_push() first and exit 0
+#      before calling extract_push_ref. This function still returns empty for
+#      `--tags` (it's a boolean flag), which is the same as before, but the
+#      caller should never reach the branch-name validation step for tag pushes.
 #
 # Recognises:
 #   git push origin <branch>                           → <branch>
-#   git push origin <branch>:<remote-branch>           → <branch>  (LHS of refspec)
+#   git push origin HEAD:<branch>                      → <branch>  (DST of refspec)
+#   git push origin <src>:<dst-branch>                 → <dst-branch>
 #   git push -u origin <branch>                        → <branch>
 #   git push --set-upstream origin <branch>            → <branch>
 #   git push --force origin <branch>                   → <branch>
 #   git push --force-with-lease origin <branch>        → <branch>
-#   git push origin HEAD                               → empty (HEAD is not a ref name we can validate)
+#   git push origin HEAD                               → empty (HEAD alone, no dst)
+#   git push origin --tags                             → empty (tag push, no branch)
 #   git push                                           → empty (relies on upstream tracking)
 #   git push origin                                    → empty (no ref given)
-#   git push origin --delete <branch>                  → empty (deletion, no source-ref check)
+#   git push origin --delete <branch>                  → empty (deletion, no source-ref)
+#   git push upstream HEAD:feature/GH-1-foo 2>&1 | tl → feature/GH-1-foo (redirections stripped)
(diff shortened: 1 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #566

**Commit**: `404ebc505fe8b269907f0f269c5df6a2c11c3f53`

### Summary
Fixes the `git push` parsing in `_lib-extract-push-ref.sh` across three bug axes (#547): (1) a `src:dst` refspec now validates the **dst** (what lands on the remote) instead of the src, so `git push upstream HEAD:feature/GH-1-foo` is no longer wrongly blocked on `HEAD`; (2) shell redirections/pipes are stripped before token parsing so `2>`/`>`/`|` can't be mistaken for a branch name; (3) tag pushes (`--tags`, `tag <name>`, `refs/tags/`) short-circuit to a no-op via a new `is_tag_push()` guard called in `validate-branch-name.sh` before extraction. Adds 13 targeted test cases (28/28 pass).

### Checklist Results
- Security:                  Pass (no secrets/leaks; diff is shell + tests only)
- Performance:               Pass (N/A — string parsing in a hook)
- PR Description & Glossary:  Pass (5-term glossary present)
- Summary Bullet Narrative:  Pass (narrative, well above label-only threshold)
- Technical Decisions (AgDR): N/A (bug fix to existing parser; no new dep/framework/pattern)

### Verification performed
I checked out the PR HEAD (`404ebc5`) and ran the suite myself: **28 PASS / 0 FAIL**, matching the reported result. Beyond the bundled tests, I probed the lib directly against edge cases not in the suite:

- **dst extraction** — `HEAD:dst` → `dst`; `+HEAD:dst` (force marker) → `dst`; `refs/heads/x:refs/heads/main` → `main` (dst taken, then `refs/heads/` stripped). All correct.
- **redirection stripping** — `2>&1 | tail`, `>`, `2>`, `1>`, `>>` all stripped; `&&`/`;`/`&`/`|` separators also handled by the residual `[^|;&]*` grep guard.
- **`is_tag_push` false-positive guards** — branch names containing `tag`/`tags` (`feature/GH-1-add-tag-support`, `feature/GH-1-tags-page`, `tag-team`) correctly classified as BRANCH, not TAG. The `--tags(\s|$)` and `tag\s+\S` anchors are tight. `--follow-tags <branch>` correctly stays BRANCH (it pushes the branch + tags, so branch validation should still apply).

The `${ref#*:}` dst split, the redirection-stripping `sed`, and the `is_tag_push` guard are all shell-sound and BSD/macOS-portable (POSIX-only constructs, no `\b` in parameter expansion paths). No regressions for plain pushes, `-u`, `--force`, `--delete`, or no-arg pushes.

### Issues Found
None blocking.

- **nit (cosmetic):** shellcheck flags `SC2001` on the two `echo | sed` lines (lines 51, 128) suggesting `${var//search/replace}`. The regex (`[0-9]*[>|]` with end-anchoring) isn't expressible in bash parameter expansion, so `sed` is the right tool — safe to leave as-is.
- **nit (theoretical):** `is_tag_push` uses a stricter strip pattern (`[>|][&>|0-9]`) than `extract_push_ref` (`[>|]`). For the no-space form `git push upstream --tags>/tmp/o` the tag guard misses (`-->` not followed by `&>|0-9`), so it's classified BRANCH. I verified the fallthrough is **harmless**: `extract_push_ref` returns empty for it (boolean `--tags` consumes no positional), so the caller falls back to local-HEAD resolution — which in a real worktree is the actual branch. No false block in practice. Aligning the two sed patterns would be tidier but isn't required.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `404ebc505fe8b269907f0f269c5df6a2c11c3f53`

atlas-apex (commented):
## Code Review: PR #566

**Commit**: `404ebc505fe8b269907f0f269c5df6a2c11c3f53`

### Summary
Fixes the `git push` parsing in `_lib-extract-push-ref.sh` across three bug axes (#547): (1) a `src:dst` refspec now validates the **dst** (what lands on the remote) instead of the src, so `git push upstream HEAD:feature/GH-1-foo` is no longer wrongly blocked on `HEAD`; (2) shell redirections/pipes are stripped before token parsing so `2>`/`>`/`|` can't be mistaken for a branch name; (3) tag pushes (`--tags`, `tag <name>`, `refs/tags/`) short-circuit to a no-op via a new `is_tag_push()` guard called in `validate-branch-name.sh` before extraction. Adds 13 targeted test cases (28/28 pass).

### Checklist Results
- Security:                  Pass (no secrets/leaks; diff is shell + tests only)
- Performance:               Pass (N/A — string parsing in a hook)
- PR Description & Glossary:  Pass (5-term glossary present)
- Summary Bullet Narrative:  Pass (narrative, well above label-only threshold)
- Technical Decisions (AgDR): N/A (bug fix to existing parser; no new dep/framework/pattern)

### Verification performed
I checked out the PR HEAD (`404ebc5`) and ran the suite myself: **28 PASS / 0 FAIL**, matching the reported result. Beyond the bundled tests, I probed the lib directly against edge cases not in the suite:

- **dst extraction** — `HEAD:dst` → `dst`; `+HEAD:dst` (force marker) → `dst`; `refs/heads/x:refs/heads/main` → `main` (dst taken, then `refs/heads/` stripped). All correct.
- **redirection stripping** — `2>&1 | tail`, `>`, `2>`, `1>`, `>>` all stripped; `&&`/`;`/`&`/`|` separators also handled by the residual `[^|;&]*` grep guard.
- **`is_tag_push` false-positive guards** — branch names containing `tag`/`tags` (`feature/GH-1-add-tag-support`, `feature/GH-1-tags-page`, `tag-team`) correctly classified as BRANCH, not TAG. The `--tags(\s|$)` and `tag\s+\S` anchors are tight. `--follow-tags <branch>` correctly stays BRANCH (it pushes the branch + tags, so branch validation should still apply).

The `${ref#*:}` dst split, the redirection-stripping `sed`, and the `is_tag_push` guard are all shell-sound and BSD/macOS-portable (POSIX-only constructs, no `\b` in parameter expansion paths). No regressions for plain pushes, `-u`, `--force`, `--delete`, or no-arg pushes.

### Issues Found
None blocking.

- **nit (cosmetic):** shellcheck flags `SC2001` on the two `echo | sed` lines (lines 51, 128) suggesting `${var//search/replace}`. The regex (`[0-9]*[>|]` with end-anchoring) isn't expressible in bash parameter expansion, so `sed` is the right tool — safe to leave as-is.
- **nit (theoretical):** `is_tag_push` uses a stricter strip pattern (`[>|][&>|0-9]`) than `extract_push_ref` (`[>|]`). For the no-space form `git push upstream --tags>/tmp/o` the tag guard misses (`-->` not followed by `&>|0-9`), so it's classified BRANCH. I verified the fallthrough is **harmless**: `extract_push_ref` returns empty for it (boolean `--tags` consumes no positional), so the caller falls back to local-HEAD resolution — which in a real worktree is the actual branch. No false block in practice. Aligning the two sed patterns would be tidier but isn't required.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `404ebc505fe8b269907f0f269c5df6a2c11c3f53`

Issue #548, opened 2026-06-06
[Bug] pre-push-gate markdownlint lints untracked files, causing false failures


atlas-apex:
## Given / When / Then

**Given** the `pre-push-gate.sh` runs the configured `markdownlint` command, which globs `'**/*.md'`,
**When** the working tree contains **untracked** markdown that is never committed — e.g. locally-installed skill directories symlinked under `.claude/skills/` (whose bundled docs aren't lint-clean) and `.claude/session/*.md` scratch files written by skills,
**Then** the gate fails the push on lint errors in files that are not part of the repo and will never reach CI — a false negative that blocks an otherwise-clean push.

## Repro

1. Symlink (or copy) any markdown that isn't lint-clean into a subdir under `.claude/skills/` (mirrors how external skill installers drop bundled docs), and/or have a `.claude/session/<name>.md` scratch file present.
2. Make a legitimate change on a feature branch and `git push`.
3. The pre-push markdownlint step fails citing the untracked files, even though `git diff`/the committed tree is clean.

## Expected

The pre-push markdownlint should lint **only tracked files** (what is actually being pushed / what CI will see), e.g. drive the file list from `git ls-files '*.md'` rather than a raw `**/*.md` glob, or at minimum respect `.gitignore`. As-is, every adopter who has untracked markdown in the tree gets false failures and has to hand-maintain an ignore list.

## Workaround

Add explicit ignore globs to the `pre_push` markdownlint command in `.claude/project-config.json` (e.g. `'#.claude/session/**'` and the installed-skill dirs). This is per-machine and brittle — it drifts as new untracked dirs appear.

## Environment

- ApexYard framework, `.claude/hooks/pre-push-gate.sh` + the `pre_push.commands` markdownlint entry in `project-config.defaults.json` / `project-config.json`
- Hit live while pushing a release branch.

## Severity

Medium — false-fails legitimate pushes; the documented workaround (ignore globs) is brittle and per-machine.

## Glossary

| Term | Definition |
|------|------------|
| Tracked files | Files under version control (`git ls-files`); the only content a push actually delivers. |
| Pre-push gate | The local hook that mirrors CI checks before `git push` to avoid burning CI on avoidable failures. |

zeyadsleem:
Closing per maintainer request: current open ticket queue is being cleared/parked. Reopen if this work should remain active.

atlas-apex:
[...]

PR #567, opened 2026-06-07
fix(#548): pre-push markdownlint lints tracked files only

atlas-apex:
## Summary

- **Fixed false pre-push failures from untracked markdown** — the old `**/*.md` glob linted any `.md` in the working tree, including installed-skill bundled docs, `.claude/session/*.md` scratch files, and symlinked directories that never reach the repo. Replaced with `git ls-files '*.md'` so the gate lints exactly what a push delivers — matching what CI sees and preventing brittle per-machine ignore-glob workarounds.
- **Added graceful no-op when no tracked markdown exists** — a new repo or a branch with no `.md` files prints an informational note and exits 0, keeping behaviour consistent with the other guarded commands (npx-not-found path).
- **Added two regression-guard test cases** — `case8` proves that a lint-dirty UNTRACKED file does NOT trigger a gate failure; `case9` proves that a lint-dirty TRACKED file still blocks the push, so the fix does not weaken the gate for real content.

## Testing

1. Check out the branch and create a lint-dirty untracked `.md` file (e.g. in `.claude/session/scratch.md` or `.claude/skills/some-tool/DOCS.md`).
2. Make a clean tracked change, run `git push` — gate must pass.
3. Commit the dirty `.md` file (`git add` + `git commit`) and push again — gate must fail with the markdownlint error.
4. Run the test suite: `bash .claude/hooks/tests/test_pre_push_gate.sh` — all 9 cases must pass.

Refs #548

---

## Glossary

| Term | Definition |
|------|------------|
| Tracked files | Files under version control (`git ls-files`); the only content a push actually delivers to the remote and that CI will see. |
| Untracked files | Files present in the working tree but not added to the git index — never pushed, never seen by CI. |
| Pre-push gate | The local hook (`pre-push-gate.sh`) that mirrors CI checks before `git push` to prevent wasting CI minutes on avoidable failures. |
| `git ls-files` | Git plumbing command that lists tracked files; accepts a glob pattern to filter (e.g. `'*.md'` for all tracked markdown). |
| markdownlint-cli2 | The CLI runner for the markdownlint rule engine; accepts a list of file paths (or globs) and exits non-zero when any rule violations are found. |

### DIFF
--- a/.claude/hooks/tests/test_pre_push_gate.sh
+++ b/.claude/hooks/tests/test_pre_push_gate.sh
@@ -169,7 +169,54 @@ case7() {
   rm -rf "$sb"
 }
 
-case1; case2; case3; case4; case5; case6; case7
+
+# -------------------- CASE 8: untracked bad markdown → no failure --------------------
+# Regression guard for #548: a markdownlint command driven by git ls-files must
+# NOT lint untracked files, so a lint-dirty untracked .md must not block the push.
+# The command string avoids \0 / null-delimiter JSON escapes (jq rejects \0);
+# filenames in sandboxes are space-free so plain xargs (newline-split) is safe here.
+case8() {
+  local sb; sb=$(make_sandbox)
+  # Configure markdownlint using git ls-files (the fixed command shape).
+  # shellcheck disable=SC2016
+  printf '%s\n' \
+    '{"pre_push": {"commands": [{"name": "markdownlint", "run": "command -v npx >/dev/null 2>&1 || { echo INFO; exit 0; }; md_files=$(git ls-files '"'"'*.md'"'"' 2>/dev/null); [ -z \"$md_files\" ] && { echo INFO_SKIP; exit 0; }; echo \"$md_files\" | xargs npx --yes markdownlint-cli2 2>&1"}]}}' \
+    > "$sb/.claude/project-config.json"
+  # Drop a lint-dirty untracked markdown file.
+  # Critically, this file is NOT `git add`-ed, so git ls-files will not see it.
+  mkdir -p "$sb/.claude/skills/external-skill"
+  printf '# Bad heading  \n- item without blank line\n' \
+    > "$sb/.claude/skills/external-skill/DOCS.md"
+  # Push must succeed: the untracked file must be invisible to markdownlint.
+  run_hook "$sb" "$(push_json)" 0 "" "untracked-bad-md-ignored"
+  rm -rf "$sb"
+}
+
+# -------------------- CASE 9: tracked bad markdown → failure --------------------
+# Regression guard for #548: a lint error in a TRACKED markdown file must still
+# block the push, so the fix does not weaken the gate for real content.
+# Same command shape as case8 (space-safe xargs without -0, valid JSON).
+case9() {
+  local sb; sb=$(make_sandbox)
+  # shellcheck disable=SC2016
+  printf '%s\n' \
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #567

**Commit**: `146e30eb30579b9a8d9fccf0fec2207f4ac88fcc`

### Summary
Changes the `markdownlint` entry in `.claude/project-config.json`'s `pre_push.commands` to drive its file list from `git ls-files '*.md'` (tracked files only) instead of the raw `**/*.md` glob, so untracked markdown (installed-skill bundled docs, `.claude/session/*.md` scratch) no longer false-fails the pre-push gate. Adds regression cases 8 (untracked bad md → no fail) and 9 (tracked bad md → fails) to `test_pre_push_gate.sh`.

### Checklist Results
- Performance:               Pass (n/a)
- Summary Bullet Narrative:  Pass — all three bullets are verb + rationale, none label-only
- Technical Decisions (AgDR):N/A — bug fix to an existing mechanism; `git ls-files` is git plumbing already used across the hooks. No new dependency / framework / pattern / storage choice.
- Adopter Handbooks:         N/A — diff touches a bash test + JSON config only. `migration-safety.md` (blocking) doesn't trigger (no migration-path files); layering / commit-msg advisory handbooks don't apply.

### Verification performed
I checked out the PR HEAD into a worktree and ran the actual 9-case suite (not just the 7-case local copy):

```
PASS [untracked-bad-md-ignored]   (case8)
PASS [tracked-bad-md-fails]        (case9)
PASS: 9   FAIL: 0
```

`npx` was present on the runner, so **case9 actually executed** (not the graceful SKIP path) — confirming the fix does NOT weaken the gate: a lint-dirty *tracked* file still blocks the push with `markdownlint: FAILED`.

Shell-soundness of the production command verified in an isolated repo:

- **Empty-list guard works** — `md_files=$(git ls-files '*.md'); [ -z "$md_files" ] && { echo INFO; exit 0; }` short-circuits when there's no tracked markdown, so `markdownlint-cli2` is never invoked argument-less (which would otherwise default-scan the whole tree). Confirmed: empty list → clean exit, no lint run.
- **Filename safety** — production uses `echo "$md_files" | tr '\n' '\0' | xargs -0`, which is space-safe. Verified `with space.md` is passed as a single argument under `xargs -0`, whereas plain newline-split `xargs` would have split it into two bogus paths. The production string is the *more* robust of the two shapes.
- **Tracked-only scope** — `git ls-files '*.md'` reports only index-tracked markdown; untracked working-tree files are invisible to it. This matches what a push delivers and what CI sees.
- **npx-missing skip preserved** — the `command -v npx ... || { echo INFO; exit 0; }` prefix is unchanged, so contributors without Node are still skip-with-note, not hard-blocked.

Glossary present and accurate. No secrets, no private-project leaks in the diff (2 files: the test and the config).

### Issues Found
None blocking.

### Suggestions
- nit (non-blocking): the **production** command uses `tr '\n' '\0' | xargs -0` while the two new **test** cases use plain newline-split `xargs`. The tests therefore validate the *behaviour* against a slightly simplified command string rather than the literal production string. The test comments explicitly explain why (jq rejects `\0` in the JSON test fixture; sandbox filenames are space-free), so this is a documented, reasonable trade-off — and the production string is the safer variant. No change required; flagging only so a future reader knows the two strings intentionally differ.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `146e30eb30579b9a8d9fccf0fec2207f4ac88fcc`

PR #630, opened 2026-06-11
fix(#629): config_get drops all project-config overrides on a backslash escape

### LINKED ISSUE #629: [Bug] config_get returns empty for ALL keys when merged config has a backslash escape (echo-mangles-JSON) — silently drops every project-config override
## Affected
`.claude/hooks/_lib-read-config.sh` — `config_get()` (surfaces through `_lib-portfolio-paths.sh` resolvers: `portfolio_registry`, `portfolio_projects_dir`, `portfolio_ideas_backlog`, `portfolio_workspace_dir`, etc.)

## Given / When / Then
**Given** a fork whose merged config (`project-config.defaults.json` ∗ `project-config.json`) contains any string value with a JSON backslash escape — e.g. the shipped `pre_push.commands[].run` markdownlint step contains `tr '\\n' '\\0'` (valid JSON: `\\n`, `\\0`) —
**When** any code calls `config_get` / `config_get_or` while the executing shell's `echo` does XSI/POSIX backslash interpretation (observed: the hook/skill shell where `$BASH_VERSION` is empty, i.e. `/bin/sh`-class, not interactive bash),
**Then** `config_get` returns an **empty string for every key** — expected: the configured value.

Root cause: `config_get` does `echo "$_CONFIG_CACHE" | jq -r "$filter"`. `echo` collapses the cached JSON's `\\0` → `\0`, which is an **invalid JSON escape**, so `jq` aborts on the whole document (`jq: parse error: Invalid escape at line N`). `config_get` swallows the error (`2>/dev/null`) and yields empty; `config_get_or` then returns its in-code fallback for *every* lookup.

**Impact (why this is major, not cosmetic):** all `project-config.json` overrides are silently ignored. In **split-portfolio mode** the `portfolio.*` overrides are exactly what's lost, so `portfolio_ideas_backlog` / `portfolio_registry` / `portfolio_projects_dir` / `portfolio_workspace_dir` resolve to the **in-fork defaults** instead of the sibling private repo. Skills (`/idea`, `/handover`, `/start-ticket`, …) then read/write portfolio data **inside the public fork** rather than the private sibling — a data-integrity and potential private-data-exposure problem. It also silently breaks `tracker.kind`, `ui_paths`, `design_paths`, and every other configurable key (all fall back to defaults).

## Repro
1. On a fork with the default `pre_push` markdownlint command in the merged config (it contains `tr '\\n' '\\0'`), in a non-interactive `/bin/sh`-class shell:
   ```sh
   . .claude/hooks/_lib-read-config.sh
   cache=$(_config_load)
   echo "$cache"   | jq -r '.tracker.kind'   # -> jq: parse error: Invalid escape ... (empty)
   printf '%s' "$cache" | jq -r '.tracker.kind'   # -> correct value
   ```
2. Equivalently, in split-portfolio mode with `portfolio.ideas_backlog` overridden to a sibling path, `config_get_or '.portfolio.ideas_backlog' <default>` returns the **default** (in-fork) path, not the override.
3. `_config_load | jq ...` (direct pipe) works — only the `echo "$var" | jq` round-trip inside `config_get` corrupts it, which is why it hid for so long.

## Framework version
v3.1.4

## Severity
major

## Notes
- **Fix:** replace `echo "$_CONFIG_CACHE" | jq -r "$filter"` with `printf '%s' "$_CONFIG_CACHE" | jq -r "$filter"` in `config_get`. Audit the rest of the hooks for the same `echo "$json" | jq` anti-pattern (any cached-JSON re-emit should use `printf '%s'`).
- Trigger is shell-dependent (`echo` XSI behaviour) **and** content-dependent (a `\\`-bearing config value). The shipped `pre_push` markdownlint command supplies the content, so any adopter on a `/bin/sh`-class hook shell is exposed — it isn't exotic local config.
- Suggested regression test: a merged-config fixture containing a `\\0`/`\\n` value, asserting `config_get` returns the right value for an unrelated key.

## Glossary
| Term | Definition |
|------|------------|
| XSI `echo` | POSIX/System-V `echo` that interprets backslash escapes (`\n`, `\0`, `\\`) in its argument, unlike bash's default builtin `echo`. |
| Merged config | `project-config.defaults.json` deep-merged with the adopter's `project-config.json` via `jq -s '.[0] * .[1]'`. |
| Split-portfolio mode | Layout where registry / project docs / ideas / workspace live in a private sibling repo, selected via the `portfolio.*` config keys. |

atlas-apex:
## Summary

- **Root cause** — `config_get` re-emitted the cached merged config with `echo "$_CONFIG_CACHE" | jq`. Under an XSI/POSIX shell (or bash with `xpg_echo`) `echo` interprets backslash escapes, collapsing a *valid* JSON `\\0` — from the shipped `pre_push` markdownlint command `tr '\n' '\0'` — into an *invalid* `\0`. jq aborts on the whole document (`Invalid escape …`), the error is swallowed (`2>/dev/null`), and `config_get` returns **empty for every key**, so `config_get_or` falls back to its in-code default for *every* lookup.
- **Impact (why this is more than cosmetic)** — all `.claude/project-config.json` overrides were silently ignored. In **split-portfolio mode** the `portfolio.*` paths resolved to the **in-fork defaults**, so `/idea` wrote to the fork instead of the sibling repo, and **`/handover` failed to detect split-portfolio mode — about to clone managed projects and write their docs into the PUBLIC fork** (`./workspace`, `./projects`) instead of the private portfolio sibling. That's a data-integrity + private-data-exposure risk, not just a wrong path. It also silently disabled `tracker.kind`, `ui_paths`, `design_paths`, and every other configurable key.
- **Fix** — one line: `printf '%s'` instead of `echo` (printf never mangles its argument). Verified the split-portfolio resolvers now return the sibling paths and `portfolio_is_v2` reports `true`.
- **Regression test** — `tests/test_config_get_backslash_escape.sh` forces `xpg_echo` to reproduce the escape-interpreting behaviour (the bug is *latent* under a plain bash `echo`, so a naive test wouldn't guard it), then asserts the override survives a backslash-bearing config value.

## Testing

1. New `test_config_get_backslash_escape.sh` — **2/2 pass** on the fix; proven to fail on the old `echo` form under `xpg_echo`.
2. Existing `test_detect_deprecated_config.sh` — **7/0** still green.
3. Real-repo check post-fix: `config_get '.portfolio.ideas_backlog'` → `../apexyard-portfolio/projects/ideas-backlog.md`; `portfolio_workspace_dir` → sibling `workspace`; `portfolio_is_v2` → `true`.

Closes #629

---

## Glossary

| Term | Definition |
|------|------------|
| `config_get` | Shared hook/skill helper that returns a value from the merged `project-config.{defaults,}.json` via a jq filter. |
| XSI / `xpg_echo` | POSIX/System-V `echo` (and bash's `xpg_echo` option) that interprets backslash escapes (`\n`, `\0`, `\\`) in its argument — unlike bash's default builtin `echo`. |
| Merged config | `project-config.defaults.json` deep-merged with the adopter's `project-config.json` via `jq -s '.[0] * .[1]'`. |
| Split-portfolio mode | Layout where the registry, project docs, ideas backlog and workspace live in a private sibling repo, selected via the `portfolio.*` config keys. |

### DIFF
--- a/.claude/hooks/_lib-read-config.sh
+++ b/.claude/hooks/_lib-read-config.sh
@@ -129,7 +129,13 @@ config_get() {
     _CONFIG_CACHE=$(_config_load)
   fi
   if command -v jq >/dev/null 2>&1; then
-    echo "$_CONFIG_CACHE" | jq -r "$filter" 2>/dev/null
+    # printf '%s', NOT echo: under an XSI/POSIX shell `echo` interprets backslash
+    # escapes, collapsing a valid JSON escape in the cached config (e.g. `\\0` in
+    # a pre_push command value) into an invalid one — jq then aborts on the whole
+    # document and config_get returns empty for EVERY key, silently dropping all
+    # project-config overrides (incl. split-portfolio portfolio.* paths).
+    # See me2resh/apexyard#629.
+    printf '%s' "$_CONFIG_CACHE" | jq -r "$filter" 2>/dev/null
   else
     return 0
   fi
--- a/.claude/hooks/tests/test_config_get_backslash_escape.sh
+++ b/.claude/hooks/tests/test_config_get_backslash_escape.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# Regression test for me2resh/apexyard#629.
+#
+# `config_get` must keep returning project-config overrides even when the merged
+# config contains a JSON value with a backslash escape (e.g. the shipped
+# `pre_push` markdownlint command contains `tr '\n' '\0'` → JSON `\\n` / `\\0`).
+#
+# The original bug: config_get did `echo "$_CONFIG_CACHE" | jq`. Under a shell
+# whose `echo` interprets backslash escapes (XSI / POSIX `echo`, or bash with
+# `xpg_echo`), `echo` collapses the valid JSON `\\0` into an invalid `\0`, jq
+# aborts on the whole document, and config_get returns empty for EVERY key —
+# silently dropping all overrides (incl. split-portfolio `portfolio.*` paths).
+# The fix uses `printf '%s'`, which never mangles its argument.
+#
+# This test forces the escape-interpreting behaviour with `shopt -s xpg_echo`
+# so it reproduces the bug on the old code and passes on the fixed code. Under a
+# plain bash `echo` (no xpg_echo) the bug is latent, so the guard would be
+# meaningless without forcing the mode.
+
+set -u
+HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+TMP="$(mktemp -d)"
+trap 'rm -rf "$TMP"' EXIT
+
+mkdir -p "$TMP/.claude"
+: > "$TMP/.apexyard-fork"
+
+cat > "$TMP/.claude/project-config.defaults.json" <<'JSON'
+{ "portfolio": { "ideas_backlog": "./projects/ideas-backlog.md" } }
+JSON
+
+# Override carries (a) a portfolio override to assert, and (b) a value with a
+# JSON backslash escape (`\\n` / `\\0`) — the exact shape that broke jq via echo.
+cat > "$TMP/.claude/project-config.json" <<'JSON'
+{
+  "portfolio": { "ideas_backlog": "../portfolio/projects/ideas-backlog.md" },
+  "pre_push": { "commands": [ { "name": "x", "run": "printf '%s' \"$f\" | tr '\\n' '\\0'" } ] }
+}
+JSON
+
+pass=0; fail=0
+ok() { echo "  ok: $1"; pass=$((pass + 1)); }
+no() { echo "  FAIL: $1"; fail=$((fail + 1)); }
+
+result="$(
+  cd "$TMP" || exit 1
+  export APEXYARD_OPS_DISABLE_PIN=1
+  unset _CONFIG_CACHE _CONFIG_ROOT_CACHE 2>/dev/null || true
+  shopt -s xpg_echo 2>/dev/null || true   # force the escape-interpreting echo that triggered #629
+  # shellcheck source=/dev/null
+  . "$HOOK_DIR/_lib-read-config.sh"
+  got="$(config_get '.portfolio.ideas_backlog')"
+  if [ "$got" = "../portfolio/projects/ideas-backlog.md" ]; then echo "PASS"; else echo "GOT:[$got]"; fi
+)"
+
+case "$result" in
+  *PASS*) ok "config_get returns the override despite a backslash-escaped config value" ;;
+  *)      no "config_get dropped the override under xpg_echo ($result)" ;;
+esac
+
+# Sanity: an unrelated key still resolves too (the whole document must parse).
+result2="$(
+  cd "$TMP" || exit 1
+  export APEXYARD_OPS_DISABLE_PIN=1
+  unset _CONFIG_CACHE _CONFIG_ROOT_CACHE 2>/dev/null || true
+  shopt -s xpg_echo 2>/dev/null || true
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #630

**Commit**: `f5d216945928e66c72700ff35eabf0b5c96380d3`

### Summary
One-line correctness fix in `config_get` (`.claude/hooks/_lib-read-config.sh`): swaps `echo "$_CONFIG_CACHE" | jq` for `printf '%s' "$_CONFIG_CACHE" | jq`, plus a new regression test. Under an XSI/`xpg_echo` shell, `echo` interprets the backslash escape in the shipped `pre_push` markdownlint command (`tr '\n' '\0'` → JSON `\\0`), corrupting the cached config into invalid JSON. jq then aborts on the whole document and `config_get` returns empty for **every** key — silently dropping all `project-config.json` overrides, including the split-portfolio `portfolio.*` paths that broke `/handover`'s clone location and `/idea`'s backlog path in #629.

### Checklist Results
- Security:                  Pass (the bug itself had a private-data-exposure edge — `/handover` cloning managed projects into the public fork — which this fix closes)
- Performance:               Pass (no change)
- Summary Bullet Narrative:  Pass (every bullet is what + why)
- Technical Decisions (AgDR):N/A (correctness bug fix, not a library/architecture choice)

### Verification performed
- **Reproduced the bug** on the old `echo` form under `shopt -s xpg_echo`: `config_get '.portfolio.ideas_backlog'` → `[]` (empty). With the fix: returns the override correctly.
- **Ran the new test** — 2/2 pass on the fix.
- **Confirmed the test guards the regression** — reverted the one line back to `echo` and re-ran: both assertions FAIL (exit 1). So the test genuinely fails on buggy code and passes on fixed code, not a tautology.
- **`printf '%s'` robustness** — correct replacement. No `%`-format pitfall (the data is an argument to `%s`, not the format string, so a literal `%` in a config value is safe — verified with `50% off` test data). No trailing-newline concern (jq parses with or without). `printf '%s'` never interprets backslash escapes in its operand under any shell, which is exactly the property needed here.

### Issues Found
None blocking.

### Suggestions (scope — follow-up, non-blocking)
The same `echo "$captured_json" | jq` anti-pattern that caused #629 exists at **other sites that re-emit captured config/API JSON**, and at least one is exploitable by the *exact same* `pre_push` value:

- ⚠ **`.claude/hooks/pre-push-gate.sh:88,95,96`** — re-emits `config_get '.pre_push.commands'` via `echo "$CMDS_JSON" | jq`. I verified under `xpg_echo` that with the shipped `tr '\n' '\0'` run-command, `NUM_CMDS=$(echo "$CMDS_JSON" | jq 'length')` returns empty → the gate exits early at line 89 and silently runs **zero** pre-push validation commands. `printf '%s'` fixes it identically. This is the highest-value follow-up because it's the same data class and the same trigger value.
- `.claude/hooks/_lib-tracker.sh` (multiple `echo "$raw" | jq` re-emitting tracker API JSON) is the same class but lower risk (tracker JSON rarely carries `\0`-style escapes); worth a sweep for consistency.

Note: the `echo "$INPUT" | jq` calls across most hooks read the hook's single stdin object once at entry; those are lower risk and their failure mode is a benign empty-match no-op rather than a silent config drop — so they're reasonable to leave. I would **not** expand this PR to cover them; a focused follow-up for `pre-push-gate.sh` (and a sweep of `_lib-tracker.sh`) is the right shape. Filing that as a separate ticket keeps this minimal, well-tested fix mergeable as-is.

### Verdict
**APPROVED**

The fix is correct, minimal, the root-cause analysis in the description is accurate, and the regression test is a genuine guard (proven to fail on the reverted code). The broader `echo | jq` sweep is correctly out of scope and best tracked separately.

---
📌 Reviewed commit: `f5d216945928e66c72700ff35eabf0b5c96380d3`

PR #634, opened 2026-06-11
fix(#584): extract_push_ref no longer over-matches '->' in commit messages

### LINKED ISSUE #584: [Bug] extract_push_ref over-matches '->' in commit message, falsely blocks compound commit+push
## Affected

`.claude/hooks/_lib-extract-push-ref.sh` → `extract_push_ref()` (and therefore every hook that sources it: `block-main-push.sh`, `validate-branch-name.sh`).

## Given / When / Then

**Given** I run a compound command that commits and pushes a feature branch in one shell line, where the **commit message contains an ASCII arrow `->`** (extremely common in technical commit messages, e.g. "remap blue->info"):

```
git add a b && git commit -m "fix: remap blue->info, purple->violet" && git push origin feature/GH-1-foo
```

**When** `block-main-push.sh` calls `extract_push_ref "$COMMAND"` to read the push destination,

**Then** it returns **empty** and the hook falls back to `git branch --show-current`. In split-portfolio / workspace-clone setups the hook's cwd is the ops fork (often on `dev`), so the fallback resolves a protected branch and the push is **falsely BLOCKED** — expected: the push targets `feature/GH-1-foo` (not protected) and should be allowed.

## Root cause

`extract_push_ref()` strips shell redirections before token parsing with:

```sh
stripped_cmd=$(echo "$cmd" | sed 's/[[:space:]]*[0-9]*[>|].*$//')
```

Both leading quantifiers are zero-width (`[[:space:]]*` and `[0-9]*` match zero chars), so the class `[>|]` matches a bare `>` **anywhere**, including the `>` inside `->`. The sed truncates the whole command at the first arrow in the commit message, discarding the trailing `&& git push origin <branch>`. The subsequent `grep -oE '\bgit\s+push\b...'` then finds no push segment → returns empty → caller falls back to local HEAD.

## Repro (minimal, copy-paste)

```sh
cmd='git add a b && git commit -m "fix: remap blue->info" && git push origin feature/GH-1-foo'
echo "$cmd" | sed 's/[[:space:]]*[0-9]*[>|].*$//'
# => git add a b && git commit -m "fix: remap blue   ← truncated at '->', push segment gone
```

Observed live: even a harmless `echo`-only command containing `git push ... feature/GH-1-foo` plus `->` arrows in an earlier quoted segment trips the `block-main-push.sh` PreToolUse gate.

## Suggested fix

Anchor the redirection match to a real operator boundary so `->` (and other in-word `>`) are not treated as redirections. Options:

1. Require the redirection to be at a token boundary — e.g. only strip `>`/`>>`/`2>`/`&>` when preceded by whitespace or a bare fd digit-group that is itself space-preceded, and only strip `|` when it is the shell pipe (not inside a quoted string).
2. Better: isolate the `git push` segment by splitting on `&&` / `;` / `|` **first**, then parse args within that segment, instead of a global redirection strip across the whole compound line.

A regression test with an arrow-containing commit message in a compound `commit && push` should accompany the fix (sibling to the existing `_lib-extract-push-ref` / `validate-branch-name` tests, and a `block-main-push` test).

## Framework version

v3.0.0

## Severity

major — wrong result (legitimate feature-branch push blocked), workaround exists (run `git push` as a standalone command, or avoid `->` in the commit message of a compound commit+push). Bites anyone whose commit message contains `->` while committing+pushing in one line.

## Notes

Distinct from #547 (refspec misread) and #549/#580 (block-main-push reading cwd not target). This is specifically the redirection-strip sed over-matching `->` inside a quoted commit message in a compound command, which then *triggers* the #549-class cwd fallback. Fixing the over-match closes the remaining false-block surface.

## Glossary

| Term | Definition |
|------|------------|
| `extract_push_ref` | Shared helper that reads the destination branch from a `git push` command so branch-protection hooks gate on the push target, not the harness cwd. |
| Redirection-strip sed | The `sed` step that removes shell redirections (`2>&1`, `>file`, `\| pipe`) from a command before token parsing. |
| Compound command | A single shell line chaining steps with `&&` / `;` (e.g. `add && commit && push`). |
| Protected branch | A long-lived integration branch (`main`/`master`/`dev`/`develop`) that may only receive changes via PR. |

atlas-apex:
## Summary

- **Fixes a false push-block** — `extract_push_ref()` (and the sibling `is_tag_push()`) stripped shell redirections with `[[:space:]]*[0-9]*[>|].*`. The leading `[[:space:]]*` is zero-width, so a bare `>` inside an ASCII arrow `->` — extremely common in commit messages (`"remap blue->info"`) — matched as a redirection. A one-line `git commit -m "...->..." && git push origin <branch>` got truncated at the arrow, so `extract_push_ref` returned empty, the caller fell back to `git branch --show-current` (in split-portfolio/workspace setups that's often the ops fork on a protected branch), and the push was **falsely BLOCKED**.
- **Fix** — require a real **whitespace token boundary** before the operator: `[[:space:]][0-9]*[>|].*`. An arrow `->` has a non-space (`-`) before the `>`, so it no longer matches; genuine redirections/pipes (` 2>&1`, ` | tail`, ` >file`) still do. Applied to both `extract_push_ref()` and `is_tag_push()` for consistency.
- **Regression test** — new `test_extract_push_ref_arrow.sh` (9 cases): arrows-in-message keep the push ref, real redirections still stripped, tag-push detection unaffected.

## Testing

1. New `test_extract_push_ref_arrow.sh` → **9/9**.
2. No regression on the existing suites: `test_validate_branch_name_pushref` **28/0**, `test_block_main_push` **31/0**, `test_pre_push_gate` **9/0**, `test_extract_pr` **14/0**.

Closes #584

---

## Glossary

| Term | Definition |
|------|------------|
| `extract_push_ref` | Helper in `_lib-extract-push-ref.sh` that reads the destination branch from a `git push` command; sourced by `block-main-push.sh` and `validate-branch-name.sh`. |
| Redirection strip | Removing shell redirection/pipe tokens (`2>&1`, `\| tail`, `>file`) before token-parsing the command, so they aren't mistaken for branch names. |
| Zero-width quantifier | A regex piece (`*`) that can match the empty string — here it let the operator match without a preceding space, catching `->`. |
| Token boundary | A whitespace separating shell tokens; requiring one before `>`/`\|` distinguishes a real redirection from a `>` inside a word. |

### DIFF
--- a/.claude/hooks/_lib-extract-push-ref.sh
+++ b/.claude/hooks/_lib-extract-push-ref.sh
@@ -48,7 +48,9 @@ is_tag_push() {
   # Strip everything after the first shell redirection/pipe so the check isn't
   # fooled by `2>&1 | tail` appended to the command.
   local clean_cmd
-  clean_cmd=$(echo "$cmd" | sed 's/[[:space:]]*[0-9]*[>|][&>|0-9].*$//')
+  # require a WHITESPACE boundary before the redirection/pipe operator so a bare
+  # `>` inside an ASCII arrow `->` in a commit message is not treated as one (#584).
+  clean_cmd=$(echo "$cmd" | sed 's/[[:space:]][0-9]*[>|].*$//')
 
   # --tags flag: `git push [opts] --tags [remote]`
   if echo "$clean_cmd" | grep -qE '\bgit\s+push\b.*\s--tags(\s|$)'; then
@@ -122,10 +124,13 @@ extract_push_ref() {
   #   - `NNN>` (e.g. `2>`, `1>`, plain `>`)
   #   - `|` pipe
   # This is conservative: we drop from the FIRST such operator to end-of-line.
-  # The sed expression matches optional digits, then `>` or `|`, followed by
-  # anything: `[[:space:]]*[0-9]*[>|].*`
+  # The sed expression requires a WHITESPACE token boundary before the operator,
+  # then optional fd digits, then `>` or `|`, then anything: `[[:space:]][0-9]*[>|].*`.
+  # The leading whitespace is the #584 fix: the old `[[:space:]]*` (zero-width) let a
+  # bare `>` inside an ASCII arrow `->` in a commit message match, truncating the
+  # command before the trailing `&& git push <branch>` and falsely blocking the push.
   local stripped_cmd
-  stripped_cmd=$(echo "$cmd" | sed 's/[[:space:]]*[0-9]*[>|].*$//')
+  stripped_cmd=$(echo "$cmd" | sed 's/[[:space:]][0-9]*[>|].*$//')
 
   # Isolate the `git push ...` segment up to the first command separator
   # (|, ;, &&, &) so `git push origin foo && echo bar` doesn't pick up `echo`
--- a/.claude/hooks/tests/test_extract_push_ref_arrow.sh
+++ b/.claude/hooks/tests/test_extract_push_ref_arrow.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# Regression test for me2resh/apexyard#584.
+#
+# extract_push_ref() stripped shell redirections with `[[:space:]]*[0-9]*[>|].*`.
+# The zero-width `[[:space:]]*` let a bare `>` inside an ASCII arrow `->` in a
+# commit message match, truncating a compound `commit ... && git push <branch>`
+# before the push segment → extract_push_ref returned empty → the caller fell
+# back to local HEAD and falsely BLOCKED the push. The fix requires a whitespace
+# token boundary before the operator.
+
+set -u
+HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+# shellcheck source=/dev/null
+. "$HOOK_DIR/_lib-extract-push-ref.sh"
+
+pass=0; fail=0
+eq() { # eq <label> <expected> <actual>
+  if [ "$2" = "$3" ]; then echo "  ok: $1"; pass=$((pass + 1));
+  else echo "  FAIL: $1 — expected [$2] got [$3]"; fail=$((fail + 1)); fi
+}
+is() { # is <label> <expect 0|1> <cmd...>
+  if is_tag_push "$3"; then got=0; else got=1; fi
+  eq "$1" "$2" "$got"
+}
+
+# --- the #584 bug: arrow in the commit message must not eat the push segment ---
+eq "arrow in commit msg keeps the push ref" "feature/GH-1-foo" \
+  "$(extract_push_ref 'git add a && git commit -m "fix: remap blue->info, purple->violet" && git push origin feature/GH-1-foo')"
+
+eq "multiple arrows still keep the ref" "feature/GH-2-bar" \
+  "$(extract_push_ref 'git commit -m "a->b->c->d" && git push -u origin feature/GH-2-bar')"
+
+# --- redirections must STILL be stripped (no regression) ---
+eq "real redirection still stripped" "feature/GH-1-foo" \
+  "$(extract_push_ref 'git push upstream HEAD:feature/GH-1-foo 2>&1 | tail -5')"
+
+eq "pipe suffix still stripped" "feature/GH-3-baz" \
+  "$(extract_push_ref 'git push origin feature/GH-3-baz | cat')"
+
+# --- unchanged baselines ---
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #634

**Commit**: `10969b191eb27a35080578a3c0144486e8ba2a79`

### Summary
Fixes a false push-block (#584). `extract_push_ref()` / `is_tag_push()` stripped shell redirections with `[[:space:]]*[0-9]*[>|].*`. The leading `[[:space:]]*` is zero-width, so a bare `>` inside an ASCII arrow `->` in a commit message (e.g. `remap blue->info`) matched as a redirection, truncating a compound `git commit -m "...->..." && git push <branch>` before the push segment. `extract_push_ref` then returned empty, the caller fell back to `git branch --show-current`, and (in split-portfolio/workspace setups) the push was falsely BLOCKED. The fix requires a real whitespace token boundary before the operator: `[[:space:]][0-9]*[>|].*`. A new regression test is added.

### Checklist Results
- Security:                  Pass (no auth/crypto/secrets touched)
- Performance:               Pass (single sed substitution, no hot-path change)
- PR Description & Glossary: Pass (4-term glossary, narrative summary)
- Technical Decisions (AgDR):N/A (a bugfix regex tweak to an existing helper; no new pattern/lib/decision)
- Adopter Handbooks:         N/A (no handbooks loaded — diff is shell-only; no architecture/general handbooks present)

### Verification performed

**(1) Regex correctness — no legitimate redirection dropped.** Probed the new `s/[[:space:]][0-9]*[>|].*$//` against every edge case:

| Input (push cmd) | Stripped? | Correct |
|---|---|---|
| ` 2>&1` | yes | ✓ |
| ` \| tail` | yes | ✓ |
| ` >file` | yes | ✓ |
| ` >>file` | yes (first `>` matches, `.*` eats rest) | ✓ |
| ` 2>err.log`, ` 1>&2` | yes | ✓ |
| `foo>bar` (no space) | NO — now preserved | intended; see note |
| commit `blue->info` then `&& git push <b>` | NO (arrow preserved) | ✓ — the bug is fixed |

`extract_push_ref` returns the correct branch for the arrow case and still strips real redirections (`2>&1 | tail` → branch recovered). The one behavioural change is `foo>bar` (no-space redirection) is no longer stripped — acceptable: commit messages with `->` are vastly more common than a space-less redirection target in a push command, and a no-space `>` adjacent to a word is rare in practice.

**(2) `is_tag_push` alignment — no tag-detection regression.** Note this is actually TWO changes to `is_tag_push`, only one of which the PR description names: besides the `[[:space:]]*`→`[[:space:]]` boundary, the operator class changed from `[>|][&>|0-9]` (required a following char) to `[>|].*` (no required following char), unifying it with `extract_push_ref`. This is a net correctness improvement — `--tags >file` is now stripped where it previously wasn't — and harmless because `--tags`/`tag`/`refs/tags/` markers are grepped *before* the strip point. Verified the full sweep: `--tags`, `--tags 2>&1 | tail`, `refs/tags/v1 >log`, `tag v1.0.0` all still detected as tag pushes; arrow-in-message and plain branch correctly NOT tag pushes.

**(3) Test suites — all green (re-run locally):**
- `test_extract_push_ref_arrow.sh` (new) — 9/9
- `test_validate_branch_name_pushref.sh` — 28/0
- `test_block_main_push.sh` — 31/0
- `test_pre_push_gate.sh` — 9/0
- `test_extract_pr.sh` — 14/0

The new test genuinely guards the bug: its first two cases assert the push ref survives a `commit -m "...->..." && git push` (red against the old regex, green against the new), and it also covers the no-regression direction (real redirections still stripped) and tag detection.

### Suggestions
- **nit (out of scope — related follow-up):** Both modified lines retain `echo "$cmd" | sed ...`. This is the same latent `echo`-mangle class as #629/#631. Under the lib's default `#!/bin/bash` with `xpg_echo` off (the macOS/Linux default), `echo` and `printf` behave identically — verified, no mangling. But under POSIX `sh`/`dash` or `xpg_echo` on, `echo "$cmd"` would interpret `\t`/`\n` in a commit message as literal tab/newline, and the injected newline could break sed's `$`-anchored, per-line substitution. Correctly NOT addressed here (this PR is the narrow `->` fix), but consider a follow-up converting these `echo "$cmd" | sed` calls to `printf '%s\n' "$cmd" | sed` for defence-in-depth across the helper. Not blocking.

### Verdict
**APPROVED**

The regex change is correct, the `is_tag_push` alignment is sound (and a minor improvement), the new test meaningfully guards #584, and all four existing suites pass with no regressions.

---
📌 Reviewed commit: `10969b191eb27a35080578a3c0144486e8ba2a79`

Issue #469, opened 2026-05-31
[Bug] suggest-mcp-search advisory never reaches the agent (stderr+exit0) and isn't install-gated


atlas-apex:
## Given / When / Then

**Given** `suggest-mcp-search.sh` is wired (PreToolUse → Bash, from #418) to nudge the agent toward `search_docs`/`search_code` before grep,
**When** the agent runs a grep over a managed-project workspace path,
**Then** the hook fires and writes its `💡 MCP search available` banner to **stderr with exit 0** — which Claude Code does **not** inject into the model's context. The nudge is invisible to the agent, so it doesn't change behavior (the exact failure it was built to prevent). It also fires regardless of whether `apexyard-search` is actually installed, nagging adopters who don't have the premium search component.

## Repro

1. With the hook wired, run a search command the hook matches, e.g. `grep -rn <term> workspace/<proj>`.
2. The hook prints the banner to stderr and exits 0. Verified live: the banner renders when the hook is run directly, but the agent never receives it in-context.
3. Confirmed against the Claude Code hooks contract: exit-0 stderr is shown only to the user/terminal; only exit-2 stderr reaches the model — but exit 2 BLOCKS the tool, which an advisory must not do.

## Root cause

Wrong output channel. A non-blocking PreToolUse advisory that the model should READ must be emitted as JSON on **stdout** via `hookSpecificOutput.additionalContext` (exit 0), not as stderr. Separately, the hook has no gate on whether `apexyard-search` is configured.

## Proposed fix

- Emit the advisory as `{"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":"<text>"}}` on **stdout**, exit 0 (non-blocking, model-visible).
- **Install-gate**: only emit when `apexyard-search` appears in the resolved `.mcp.json` (ops root and/or `$APEXYARD_PORTFOLIO_ROOT`); otherwise `exit 0` silently — falls back to plain grep for adopters without the premium search component.
- Keep the detection (grep/find + framework/workspace paths) unchanged.

## Acceptance Criteria

- [ ] Advisory is emitted as `hookSpecificOutput.additionalContext` on stdout (exit 0), so the agent actually reads it
- [ ] Hook stays non-blocking (the Bash call proceeds)
- [ ] When `apexyard-search` is NOT in any resolved `.mcp.json`, the hook is silent (no nag)
- [ ] When it IS configured + the command is a workspace/framework search, the JSON advisory is emitted
[...]

PR #652, opened 2026-06-17
feat(#651): opt-in gate mode for suggest-mcp-search.sh (force MCP-first on exploratory search)

### LINKED ISSUE #651: [Feature] Optional mechanical gate mode for suggest-mcp-search.sh (force MCP-first on exploratory search)
## Problem
`suggest-mcp-search.sh` is advisory-only (exit 0). Even with the `additionalContext` nudge, an agent can — and routinely does — fall straight to `grep -r`/`find`/`Read` over indexed paths without ever trying `search_code`/`search_docs`, burning ~3–5× the tokens the MCP index exists to save. There's no way for an operator who *wants* enforcement to get it.

## Proposed behaviour
Add an **opt-in, config-gated** gate mode to the existing hook:
- When `apexyard-search` is configured **and** gate mode is enabled, **soft-block (exit 2)** exploratory `grep -r`/`find` over indexed framework/project paths, instructing the agent to use `search_code`/`search_docs` first.
- **Per-call escape hatch** (e.g. `APEXYARD_MCP_FALLBACK=1`) for the legitimate empty-result / non-indexable-repo case — the hook can't know MCP already returned nothing, so the agent must be able to proceed.
- Leave **read-then-edit** `Read`/`Glob` advisory-only (never block — you need real line numbers before an `Edit`).
- Optionally: detect a **stale index** (vs repo HEAD) and escalate the message to "reindex first" (hook can't reindex itself; it can only instruct).

## Why it helps adopters
Operators who rely on the MCP index get mechanical enforcement of the token-saving path instead of a nudge the agent can ignore, while the read→edit flow and non-indexable repos still work via the escape hatch.

## Scope / out of scope
- **Out of scope:** the hook calling MCP or reindexing itself (hooks can't invoke MCP).
- **Default stays advisory** — gate mode is strictly opt-in; free adopters / those without the search component see nothing.

## Framework version
v3.1.4

## Related
Touches `suggest-mcp-search.sh`. Related: #418, #469, #489.

## Glossary
| Term | Definition |
|------|------------|
| Soft-block | Hook exits non-zero (2) to refuse a tool call, but with a clear escape hatch — not a permanent hard stop |
| Escape hatch | An env var the agent sets on the retry call to proceed when the gate's assumption (MCP can serve this) is false |
| Stale index | MCP vector index older than the current repo HEAD, so results may miss recent changes |

atlas-apex:
## Summary
- **Added an opt-in, config-gated soft-block to `suggest-mcp-search.sh`** — the hook was advisory-only (exit 0), so an agent could (and routinely did) ignore the nudge and `grep -r`/`find` over indexed paths, burning ~3–5× the tokens the MCP index exists to save. With `mcp_search.gate_mode: true` **and** `apexyard-search` configured, the hook now **soft-blocks (exit 2)** exploratory `grep -r`/`find` over indexed framework/project paths and instructs the agent to use `search_code`/`search_docs` first.
- **Per-call escape hatch `APEXYARD_MCP_FALLBACK=1`** — the hook can't know whether MCP already returned nothing (empty index, non-indexable repo, stale index), so the agent must be able to proceed: setting the env var (operator/session) or prefixing the retried command (`APEXYARD_MCP_FALLBACK=1 grep …`) steps the gate aside. The block message names this explicitly.
- **Read/Glob/Grep stay advisory-only** — only the Bash exploratory-search branch is gate-eligible. Read→edit needs real line numbers before an `Edit`, so those native tools are never blocked (matches #651 scope).
- **Default-off + install-gated** — `mcp_search.gate_mode` defaults to `false`, and the existing install-gate means free adopters / anyone without the search component see zero behaviour change. The advisory path (gate off) is byte-for-byte unchanged.
- **AgDR-0070** records the decision (advisory → opt-in gate, soft-block-with-escape over hard-block, why stale-index detection folds into the escape hatch rather than HEAD-comparison).

## Testing
- `bash .claude/hooks/tests/test_suggest_mcp_search.sh` — 29 pass on a clean checkout. New cases: gate-on soft-block (exit 2), per-call escape token → exit 0, env-var escape → exit 0, Read-on-workspace not blocked, gate-on-but-no-apexyard-search install-gate silent, and gate-off advisory (exit 0). Fixtures use a temp ops-root (`.apexyard-fork` anchor + `project-config.{defaults,}.json`) so the config lib resolves `gate_mode` deterministically.

## Scope / out of scope
- **Out of scope:** the hook calling MCP or reindexing itself (hooks can't invoke MCP) — the gate can only block + instruct.
- Stale-index detection is folded into the escape-hatch message (the hook has no reliable index-freshness signal) rather than implemented as HEAD-comparison.

Closes #651

---

## Glossary
| Term | Definition |
|------|------------|
| Soft-block | Hook exits 2 to refuse a tool call, but with a clear escape hatch — not a permanent hard stop. |
| Gate-eligible | Only the Bash exploratory `grep -r`/`find`-over-indexed-paths branch can be blocked; Read/Glob/Grep never are. |
| Escape hatch | `APEXYARD_MCP_FALLBACK=1` (env var or per-call command prefix) that lets a search proceed when MCP genuinely can't serve it. |
| Install-gate | The pre-existing check that the hook only acts when `apexyard-search` is configured in a resolvable `.mcp.json` — free adopters see nothing. |

### DIFF
--- a/.claude/hooks/suggest-mcp-search.sh
+++ b/.claude/hooks/suggest-mcp-search.sh
@@ -29,9 +29,19 @@
 
 set -u
 
+# shellcheck source=/dev/null
+. "$(dirname "${BASH_SOURCE[0]}")/_lib-read-config.sh" 2>/dev/null || true
+
 INPUT=$(cat)
 
 TOOL_NAME=$(printf '%s' "$INPUT" | jq -r '.tool_name // empty' 2>/dev/null)
+
+# Gate mode (#651): only the Bash exploratory-search branch is gate-eligible;
+# Read/Glob/Grep stay advisory-only so read→edit is never blocked. `escape`
+# is the per-call/operator opt-out that lets a search proceed when MCP can't
+# serve it (empty index, non-indexable repo, stale index).
+gate_eligible=false
+escape=false
 
 # ---------------------------------------------------------------------------
 # Branch on tool type. Bash uses the original grep/find command scanner.
@@ -101,6 +111,14 @@ if [ "$TOOL_NAME" = "Bash" ]; then
 
   $targets_framework || exit 0
 
+  # This Bash exploratory search over indexed paths is gate-eligible (#651).
+  gate_eligible=true
+  # Escape hatch: a real env var (operator/session) OR the per-call token in
+  # the command itself (`APEXYARD_MCP_FALLBACK=1 grep …` on a retry).
+  if [ "${APEXYARD_MCP_FALLBACK:-}" = "1" ] || printf '%s' "$COMMAND" | grep -q 'APEXYARD_MCP_FALLBACK=1'; then
+    escape=true
+  fi
+
 else
   # -------------------------------------------------------------------------
   # READ / GLOB / GREP BRANCH (#489): fire when the target path is inside a
@@ -143,6 +161,33 @@ for mcp_json in "$ops_root/.mcp.json" "${APEXYARD_PORTFOLIO_ROOT:-}/.mcp.json"; 
 done
 
 $mcp_has_search || exit 0
+
+# --- Gate mode (#651): opt-in soft-block before the advisory ----------------
+# When `mcp_search.gate_mode` is true AND this is a gate-eligible Bash search
+# AND the escape hatch isn't set, soft-block (exit 2) and instruct MCP-first.
+# Default-off, so this is a no-op for everyone who hasn't opted in. AgDR-0070.
+
+GATE_MODE=$(config_get_or '.mcp_search.gate_mode' 'false' 2>/dev/null || echo false)
+
+if $gate_eligible && [ "$GATE_MODE" = "true" ] && ! $escape; then
+  cat >&2 <<'EOF'
+BLOCKED (mcp_search.gate_mode): use the apexyard-search MCP index first.
+
+This is an exploratory grep/find over indexed framework/project paths. Run
+  mcp__apexyard-search__search_code   (managed-project codebases)
+  mcp__apexyard-search__search_docs   (framework docs)
+instead — semantic, targeted excerpts, ~3-5x fewer tokens. Load via
+  ToolSearch("select:mcp__apexyard-search__search_code,mcp__apexyard-search__search_docs").
+
+If MCP already returned nothing (empty/stale index, or a non-indexable repo),
+retry the exact command with the escape hatch prefix:
+  APEXYARD_MCP_FALLBACK=1 <your command>
+
+(Gate mode is opt-in via .claude/project-config.json → mcp_search.gate_mode.
+ Read/Glob/Grep are never blocked — only exploratory shell search.)
+EOF
+  exit 2
+fi
 
 # --- Emit advisory as additionalContext on stdout (non-blocking) -----------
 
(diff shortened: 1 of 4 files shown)


atlas-apex (commented):
## Code Review: PR #652

**Commit**: `a6d05ca28ccef1045b3e17b51cbd27cc9582a264`

### Summary
Adds an opt-in, config-gated soft-block to `suggest-mcp-search.sh`. When `mcp_search.gate_mode: true` AND `apexyard-search` is configured, an exploratory `grep -r`/`find` over indexed framework/project paths now exits 2 (instead of the advisory exit 0), instructing the agent to use `search_code`/`search_docs` first. A per-call escape hatch (`APEXYARD_MCP_FALLBACK=1`, env var or command-string token) steps the gate aside. Default-off and install-gated, so free adopters and the advisory path are byte-for-byte unchanged. Decision recorded in AgDR-0070.

### Checklist Results
- Summary Bullet Narrative:  Pass (bold lead + what + why on every bullet)
- Technical Decisions (AgDR):Pass (AgDR-0070 included in diff and linked)
- Adopter Handbooks:         N/A (shell hook; always-load handbooks not relevant; no JS/TS so strict-mode handbook does not load)

### Verification performed
I exercised the hook directly against temp ops-roots, not just the diff:

1. **Opt-in + install-gated.** `gate_mode` defaults to `false` in `project-config.defaults.json`. The gate sits *after* the install-gate (`$mcp_has_search || exit 0`, line 163), so with gate ON but no `apexyard-search` configured the hook stays silent (exit 0) — confirmed. Gate OFF → advisory exit 0 unchanged.
2. **Read/Glob/Grep are NEVER blocked.** These tools branch to the `else` path (lines 122-143) and never set `gate_eligible=true`. Verified: Grep and Glob on `workspace/...` paths under gate ON both return exit 0 (advisory only). Only the Bash exploratory-search branch sets `gate_eligible=true` (line 115).
3. **Escape hatch works in both forms.** Env var `APEXYARD_MCP_FALLBACK=1` → exit 0; per-call token prefix `APEXYARD_MCP_FALLBACK=1 grep ...` → exit 0. Both confirmed.
4. **No regression to #469/#489.** The advisory `additionalContext`-on-stdout path is reached unchanged when the gate is off or the escape is set; the install-gate and Read/Glob/Grep workspace logic are untouched.
5. **Exit-2 message is on stderr** (`cat >&2`) so PreToolUse surfaces it to the agent, and it names the escape hatch explicitly.
6. **Tests pass on a clean checkout.** `29 passed, 0 failed` after temporarily hiding the untracked local `.mcp.json` at the ops root. With that file present, 5 cases fail (the 4 "gate closed / no-apexyard-search" install-gate cases plus the no-apexyard-search test) — confirmed to be solely an artifact of the local untracked `.mcp.json` being found via `$ops_root/.mcp.json`, NOT a defect in this change. New gate-mode cases all pass.

### Issues Found
None blocking.

### Suggestions
- nit: the per-call escape detection uses `grep -q 'APEXYARD_MCP_FALLBACK=1'` against the whole command string, so a search whose *pattern* literally contains `APEXYARD_MCP_FALLBACK=1` (e.g. `grep -r "APEXYARD_MCP_FALLBACK=1" roles/`) would escape the gate. This is a vanishingly rare false-negative that fails open (safe direction — it lets a search through rather than wrongly blocking), so not worth complicating the matcher. Noting only for completeness; no change requested.

### Verdict
**APPROVED**

---
Reviewed commit: `a6d05ca28ccef1045b3e17b51cbd27cc9582a264`

PR #656, opened 2026-06-17
fix(#643): block-merge-on-red-ci refuses variable-substituted merges instead of guessing

### LINKED ISSUE #643: [Chore] block-merge-on-red-ci hook: PR# and --repo parsers fail on $VAR shell substitution
## Driver

`block-merge-on-red-ci.sh` (and the shared `_lib-extract-pr.sh` it depends on) fails open + emits wildly misleading errors when the merge command uses shell variables like `gh pr merge $PR --repo $REPO`. The hook works correctly when the operator types the numbers and repo literally, but in any session where an agent has stored the PR # and repo in variables before invoking `gh pr merge`, two compounding failures occur.

## Reproduction

```bash
PR=378
REPO=adopter-org/some-private-repo
gh pr merge $PR --repo $REPO --squash --delete-branch
```

Observed behaviour: the hook blocks the merge (the CI is red, so blocking is correct), but the error message reads:

```
BLOCKED: PR #2 has red CI. Cannot merge.

`gh pr checks 2` reported failures or pending checks:

  expected the "[HOST/]OWNER/REPO" format, got "$REPO"
```

— with the wrong PR number (PR #2 instead of #378) and the literal string `$REPO` in the error.

## Root cause

Two bugs in the parsing logic:

**1. PR-number extraction (`_lib-extract-pr.sh:64`)**

```bash
pr=$(echo "$cmd" | grep -oE '\bgh\s+pr\s+merge\b[^|;&]*' | grep -oE '[0-9]+' | head -1)
```

For `gh pr merge $PR ...`, the regex `[0-9]+` finds no digits. Falls through to step 3 (`gh pr view --json number --jq '.number'` from CWD), which returns whatever PR is open on whatever branch happens to be checked out in the cwd at the time the hook fires — totally unrelated to the merge being attempted.

**2. Repo extraction (`block-merge-on-red-ci.sh:47-48`)**

```bash
CMD_REPO=$(echo "$COMMAND" | sed -nE 's/.*--repo[[:space:]]+([^[:space:]]+).*/\1/p' | head -1)
```

For `--repo $REPO`, this captures the literal string `$REPO` — the hook is parsing the unexpanded command string, so shell-variable expansion has not occurred yet. Then `--repo $REPO` is passed to `gh pr checks`, which rejects it with `expected the "[HOST/]OWNER/REPO" format`.

The hook's red-CI block ultimately fires because `gh pr checks` exits non-zero (for the wrong reason), so the gate happens to do the right thing for the wrong reason. But the operator sees garbage error output that doesn't help them diagnose.

## Failure modes

- **Misleading diagnostics**: operator wastes time trying to understand why PR #2 (which they were not merging) is in the error.
- **Wrong PR queried** when CI is being checked: the fallback path 3 in `extract_pr_number` calls `gh pr view --json number` from CWD, which could return any PR depending on session state.
- **Race-condition false positive/negative**: if the falsely-discovered PR happens to have green CI and the real PR is red, the gate may even **let the merge through** — the hook fires correctly here only by coincidence (CMD_REPO parsing failed independently). The two bugs cancel out today; if the repo-parsing is ever fixed without also fixing PR-number parsing, the red-CI gate becomes a silent allow.
- **Hook 47 design intent violated**: the `_lib-extract-pr.sh` library was created exactly to prevent merge bypasses on the `gh api .../merge` shape. The `$VAR` shell-variable path is a third unhandled shape — same class of bug as the original incident, now hiding behind an "always-blocks" symptom.

## Scope

- Fix `extract_pr_number()` in `_lib-extract-pr.sh` to recognise shell-variable PR placeholders (`$PR`, `${PR}`, `${PR:-...}`) and either resolve them via the calling shell's exported env, or refuse the parse loudly (`exit 2` with a clear message) rather than silently fall through to the CWD-PR fallback.
- Fix the `--repo` parser in `block-merge-on-red-ci.sh` (and any sibling hook with the same shape — `block-unreviewed-merge.sh`, `require-design-review-for-ui.sh`) to detect `$REPO` / `${REPO}` patterns and either resolve or refuse-with-clear-error.
- Add a test case to `tests/test_extract_pr.sh` (or sibling) covering the `gh pr merge $PR --repo $REPO` shape, including the cross-CWD scenario where the bogus PR # would otherwise come from `gh pr view`.

## Acceptance Criteria

- [ ] `extract_pr_number` returns empty (not a bogus PR #) when the command contains unexpanded shell variables in the PR-positional slot.
- [ ] `block-merge-on-red-ci.sh` and sibling merge hooks either (a) resolve `$REPO` from the inherited env when available, or (b) emit a clear error like "BLOCKED: cannot parse $REPO from command — re-run with literal owner/repo".
- [ ] New tests cover the `gh pr merge $PR --repo $REPO` shape and assert the gate fails-closed with a comprehensible message rather than fails-closed-by-accident with a misleading one.
- [ ] No regression on the existing `gh pr merge 42 --repo owner/repo` and `gh api repos/owner/repo/pulls/42/merge` shapes.

## Risks / Dependencies

- Fixing repo-parsing without fixing PR-parsing creates the regression described above. Land both fixes in the same PR.
- Adopters with raw `Bash` tools running `gh pr merge $PR ...` in scripts will need to either export `PR` / `REPO` to the hook's env, or rewrite to use literal values. Document the resolution path either way.

## Glossary

| Term | Definition |
|------|------------|
| Unexpanded command | The raw `tool_input.command` string the agent sends — variable expansion happens in the subshell that ultimately runs the command, AFTER the PreToolUse hook has already inspected the literal text. |
| Fall-through path 3 | `_lib-extract-pr.sh:64-68` — `gh pr view --json number` last-resort fallback when neither regex matched. Returns whatever PR is open on the current branch in CWD. |
| `$REPO` literal | The verbatim two-token string `$REPO` (dollar-sign + name), not the value the variable expands to. |

atlas-apex:
## Summary
- **Root cause**: merge-gate hooks parse the **literal** command string (before the shell expands variables). For `gh pr merge $PR --repo $REPO`, `block-merge-on-red-ci.sh` couldn't resolve the real target — `extract_pr_number` correctly returns empty for `$PR` (#568) but its step-3 fallback then ran `gh pr view` in the **CWD**, checking an *unrelated* PR's CI; and the inline `--repo` parser captured the literal `$REPO`, which `gh` rejected with `expected the "[HOST/]OWNER/REPO" format`. Net: wrong PR number + garbage error, blocking for the wrong reason.
- **Fix**: a CI gate must not guess its target. New shared helper **`merge_command_uses_variable`** in `_lib-extract-pr.sh` detects an unexpanded `$VAR`/`${VAR}` in the PR positional arg or the `--repo` value (reusing the same span + redirection-stripping discipline as `extract_pr_number`, so `2>&1` etc. aren't mistaken for the arg). `block-merge-on-red-ci.sh` now **blocks (exit 2) with a clear "re-run with literal values" message** when a variable is detected — *before any `gh` call* — and switches its inline `--repo` sed to the shared `extract_repo_from_command`.
- **Why block (not skip/guess)**: the hook genuinely cannot expand a shell variable from the command text. Skipping would let a red-CI merge through; guessing (the old CWD fallback) checks the wrong PR. Blocking with an accurate instruction is the only safe option.

## Testing
- `bash .claude/hooks/tests/test_extract_pr.sh` — **24 pass** (10 `merge_command_uses_variable` cases, incl. quoted `"$PR"`/`"$REPO"`: var PR `$PR`/`${PR_NUMBER}`, var repo `$REPO`/`${REPO}`, literal PR+repo → no, literal+`2>&1` pipe → no, gh-api literal path → no).
- `bash .claude/hooks/tests/test_block_unreviewed_merge.sh` — **29 pass** (no regression; shares `_lib-extract-pr.sh`).
- End-to-end: `gh pr merge $PR --repo $REPO` → exit 2 + clear message; non-merge command → exit 0.
- `shellcheck -x` clean (only pre-existing INFO notes: SC1091 sourced-lib, SC2086 intentional `$REPO_FLAG` word-split).

## Scope
Fixes `block-merge-on-red-ci` (the gate that produced the garbage output in the report). The shared `merge_command_uses_variable` helper is now available for the other three merge gates (`block-unreviewed-merge`, `require-design-review-for-ui`, `require-architecture-review`) to adopt the same guard — left as a follow-up to keep this PR focused. Those gates are less dangerous today (they fail-safe: a wrong/empty PR # means marker SHAs don't match → block).

Closes #643

---

## Glossary
| Term | Definition |
|------|------------|
| Variable-substituted merge | A `gh pr merge` invocation whose PR number or `--repo` is a shell variable (`$PR`, `${REPO}`) the hook can't expand from the literal command text. |
| Fail-open vs fail-safe | This gate now **fails safe** (blocks) on unresolvable input rather than failing open (checking an unrelated PR). |
| `extract_pr_number` / `extract_repo_from_command` | Shared parsers in `_lib-extract-pr.sh` used by all four merge gates. |
| CWD fallback | `extract_pr_number`'s last resort (`gh pr view` in the current dir) — correct for a bare `gh pr merge`, wrong for `$PR` (the bug). |

### DIFF
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -127,6 +127,52 @@ extract_pr_number() {
   echo "$pr"
 }
 
+# Returns 0 if the merge command's PR positional arg OR its --repo value is an
+# UNEXPANDED shell variable ($VAR / ${VAR}). (#643)
+#
+# WHY THIS EXISTS
+# ---------------
+# Hooks see the LITERAL command string, before the shell expands variables. For
+# `gh pr merge $PR --repo $REPO`, the hook cannot know the real PR/repo:
+#   - extract_pr_number returns empty for `$PR` (good, #568) but then falls back
+#     to `gh pr view` in the CWD — checking a totally UNRELATED PR's CI.
+#   - extract_repo_from_command captures the literal `$REPO`, which `gh` then
+#     rejects with `expected the "[HOST/]OWNER/REPO" format`.
+# Both produce misleading output and can evaluate the wrong PR. A gate that
+# can't resolve its target must not guess — callers should BLOCK with a
+# "re-run with literal values" message. This helper is that detector.
+#
+# Matches `$VAR`, `${VAR}` (the leading char after $ / ${ is a letter or _).
+# Does NOT match a literal repo/number, and does NOT match `$(...)` command
+# substitution as a PR/repo token (those aren't valid PR/repo values anyway).
+merge_command_uses_variable() {
+  local cmd="$1"
+
+  # PR positional arg: first token after `gh pr merge` (reuse the same span +
+  # redirection-stripping discipline as extract_pr_number so `2>&1` etc. don't
+  # masquerade as the positional arg).
+  local span clean_span first_token
+  span=$(echo "$cmd" | grep -oE '\bgh\s+pr\s+merge\b[^|;&]*')
+  clean_span=$(echo "$span" | sed \
+    -e 's/[0-9]*>&[0-9]*/  /g' \
+    -e 's/&>[^[:space:]]*/  /g' \
+    -e 's/>>[^[:space:]]*/  /g' \
+    -e 's/>[^[:space:]]*/  /g')
+  first_token=$(echo "$clean_span" | grep -oE '\bmerge\b[[:space:]]+[^[:space:]]*' | awk 'NR==1 {print $NF}')
+  if echo "$first_token" | grep -qE '^\$\{?[A-Za-z_]'; then
+    return 0
+  fi
+
+  # --repo value.
+  local repo_token
+  repo_token=$(echo "$cmd" | sed -nE 's/.*--repo[[:space:]]+([^[:space:]]+).*/\1/p' | head -1)
+  if echo "$repo_token" | grep -qE '^\$\{?[A-Za-z_]'; then
+    return 0
+  fi
+
+  return 1
+}
+
 # Echoes the PR's HEAD SHA as reported by GitHub, or empty on failure.
 #
 # Why this exists (see #55): merge-gate hooks previously compared approval
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -159,14 +159,17 @@ merge_command_uses_variable() {
     -e 's/>>[^[:space:]]*/  /g' \
     -e 's/>[^[:space:]]*/  /g')
   first_token=$(echo "$clean_span" | grep -oE '\bmerge\b[[:space:]]+[^[:space:]]*' | awk 'NR==1 {print $NF}')
-  if echo "$first_token" | grep -qE '^\$\{?[A-Za-z_]'; then
+  # Match `$VAR`, `${VAR}`, and the quoted forms `"$VAR"` / `'$VAR'` — agents and
+  # operators routinely quote the substitution. The optional leading quote keeps
+  # the anchor from being defeated by it.
+  if echo "$first_token" | grep -qE '^["'"'"']?\$\{?[A-Za-z_]'; then
     return 0
   fi
 
-  # --repo value.
+  # --repo value (same quoted-or-bare variable forms).
   local repo_token
   repo_token=$(echo "$cmd" | sed -nE 's/.*--repo[[:space:]]+([^[:space:]]+).*/\1/p' | head -1)
-  if echo "$repo_token" | grep -qE '^\$\{?[A-Za-z_]'; then
+  if echo "$repo_token" | grep -qE '^["'"'"']?\$\{?[A-Za-z_]'; then
     return 0
   fi
 
--- a/.claude/hooks/block-merge-on-red-ci.sh
+++ b/.claude/hooks/block-merge-on-red-ci.sh
@@ -42,13 +42,30 @@ if ! is_merge_command "$COMMAND"; then
   exit 0
 fi
 
-# Parse --repo (for `gh pr merge --repo owner/repo`). Fallback: recover from
-# the `gh api .../pulls/<N>/merge` URL path so `gh pr checks` below is still
-# scoped correctly.
-CMD_REPO=$(echo "$COMMAND" | sed -nE 's/.*--repo[[:space:]]+([^[:space:]]+).*/\1/p' | head -1)
-if [ -z "$CMD_REPO" ]; then
-  CMD_REPO=$(echo "$COMMAND" | grep -oE 'repos/[^/[:space:]]+/[^/[:space:]]+/pulls/[0-9]+/merge' | sed -nE 's|repos/([^/]+/[^/]+)/pulls/.*|\1|p' | head -1)
+# Variable-substituted merge (#643): if the PR arg or --repo value is an
+# unexpanded shell variable, this hook can't resolve the real target from the
+# command text — the old code fell back to the CWD's PR and checked an
+# UNRELATED PR's CI (and passed `$REPO` to gh, producing garbage errors). A CI
+# gate must not guess. Block with a clear, accurate instruction instead.
+if merge_command_uses_variable "$COMMAND"; then
+  cat >&2 <<'EOF'
+BLOCKED: cannot verify CI on a variable-substituted merge command.
+
+This gate reads the literal command text and can't resolve shell variables
+(e.g. `gh pr merge $PR --repo $REPO`) to the real PR / repo, so it cannot
+check the correct PR's CI status. Re-run with literal values:
+
+  gh pr merge <number> --repo <owner>/<repo> --squash
+
+(Use the actual PR number and owner/repo — not shell variables.)
+EOF
+  exit 2
 fi
+
+# Parse --repo (for `gh pr merge --repo owner/repo`). Uses the shared extractor,
+# which also recovers the repo from a `gh api .../pulls/<N>/merge` URL path so
+# `gh pr checks` below is still scoped correctly.
+CMD_REPO=$(extract_repo_from_command "$COMMAND")
 REPO_FLAG=""
 if [ -n "$CMD_REPO" ]; then
   REPO_FLAG="--repo $CMD_REPO"
--- a/.claude/hooks/tests/test_extract_pr.sh
+++ b/.claude/hooks/tests/test_extract_pr.sh
@@ -158,6 +158,9 @@ assert_var "var PR \${PR_NUMBER}"  'gh pr merge ${PR_NUMBER} --squash'          
 # Variable --repo value → yes (even with a literal PR number)
 assert_var "var repo \$REPO"       'gh pr merge 378 --repo $REPO --squash'            "yes"
 assert_var "var repo \${REPO}"     'gh pr merge 378 --repo ${REPO} --squash'          "yes"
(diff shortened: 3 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #656

**Commit**: `4797371e64c8d7e8594650e9f700aa5ab608de76`

### Summary
Closes the `block-merge-on-red-ci` failure on variable-substituted merge commands (#643). Adds a shared `merge_command_uses_variable` detector to `_lib-extract-pr.sh` that flags an unexpanded `$VAR`/`${VAR}` in the PR positional arg or `--repo` value, and makes the CI gate block (exit 2) with an accurate "re-run with literal values" message *before* any `gh` call. Also retires the hook's inline `--repo` sed in favour of the shared `extract_repo_from_command`. Clean, well-scoped, well-tested.

### Checklist Results
- Architecture & Design:     Pass — new logic lives in the shared lib; the hook stays thin and reuses the established extractor contract
- Code Quality:              Pass — reuses the exact span + redirection-stripping discipline from `extract_pr_number`; thorough WHY comment
- Testing:                   Pass — 8 new assertions covering var-PR, var-repo, literal, redirection, and gh-api shapes; 22/22 + 29/29 green
- Security:                  Pass — strictly *tightens* a merge gate; turns a fail-open (CWD-PR guess) into a fail-safe (block)
- Performance:               Pass — pure string parsing, no new subprocess in the hot path
- PR Description & Glossary:  Pass — glossary present, scope + rationale clearly stated
- Summary Bullet Narrative:  Pass — bullets explain root cause, fix, and the why-block-not-skip rationale
- Technical Decisions (AgDR):N/A — bugfix restoring correct gate behaviour; no new pattern/dependency/architecture choice. Agree no AgDR is warranted.

### Verification performed
- `bash .claude/hooks/tests/test_extract_pr.sh` → **22 pass, exit 0**
- `bash .claude/hooks/tests/test_block_unreviewed_merge.sh` → **29 pass, exit 0** (shared-lib gate, no regression)
- `shellcheck -x` → only the two pre-existing INFO notes the PR body documents (SC1091 sourced-lib, SC2086 intentional `$REPO_FLAG` word-split). No new findings.
- End-to-end: `gh pr merge $PR --repo $REPO` → **exit 2** with the clear stderr message; non-merge command → **exit 0**. Block confirmed to fire *before* `extract_repo_from_command` / any `gh` call.
- Regex precision probes:
  - `$PR`, `${PR_NUMBER}`, `$REPO`, `${REPO}` → detected (yes) ✓
  - literal PR + repo, literal + `2>&1 | tail`, `gh api repos/o/r/pulls/42/merge` → not detected (no) ✓
  - `$(...)` command substitution → not detected (no) ✓ — correct, it's not a valid PR/repo token
  - `$1` positional param → not detected (no) ✓ — consistent with the `[A-Za-z_]` anchor
- Authoritative files list (gh) confirms exactly the 3 stated files changed; inline `--repo` sed removed.

### Issues Found
None blocking.

### Suggestions
- **nit (non-blocking): double-quoted variable form is not detected.** `gh pr merge "$PR" --squash` and `--repo "$REPO"` slip past `merge_command_uses_variable` because the first token becomes `"$PR"` (leading `"`, not `$`), so the `^\$\{?[A-Za-z_]` anchor misses it. In that case `extract_pr_number` also returns empty and falls back to the CWD `gh pr view` — i.e. the original #643 bug re-opens for the quoted form. This is **consistent with the existing `extract_pr_number` contract** (its line-111 `^\$` check has the identical blind spot), so the PR doesn't introduce a *new* inconsistency, and unquoted `$VAR`/`${VAR}` is by far the dominant shape. Reasonable to leave for a follow-up, but worth a one-line comment noting the quoted form is intentionally out of scope — or, cheaply, widen the anchor to tolerate a leading quote: match `^["']?\$\{?[A-Za-z_]`. Your call; not a merge blocker.
- **suggestion (non-blocking, already noted in PR):** the other three merge gates (`block-unreviewed-merge`, `require-design-review-for-ui`, `require-architecture-review`) now have `merge_command_uses_variable` available. They fail-safe today (empty PR # → SHA mismatch → block), so adopting the explicit guard is an ergonomics/clear-message improvement, not a correctness fix. Good call deferring it to keep this PR focused.

### Verdict
**APPROVED** — solid bugfix, correctly turns a fail-open into a fail-safe, with proportionate test coverage. The one gap (quoted-var form) is a pre-existing contract limitation, not a regression, and the dominant unquoted shape is fully covered.

---
📌 Reviewed commit: `4797371e64c8d7e8594650e9f700aa5ab608de76`

atlas-apex (commented):
## Code Review: PR #656 (re-review)

**Commit**: `d16faefab9c7fdc5fdc1bf9088ae81714fca00e8`

### Summary
Re-review at the new HEAD. The only change since my prior APPROVED review is the advisory follow-up I raised: `merge_command_uses_variable` now also detects the **quoted** variable forms (`"$PR"`, `'$REPO'`) by widening both anchors to `^["']?\$\{?[A-Za-z_]`, plus 2 new test cases. Scope unchanged — 3 files, hooks-only, no new dependencies, no new technical decisions.

(Verdict is APPROVED; posted as a comment because GitHub blocks self-approval. The Rex approval marker is written at the new HEAD SHA.)

### Checklist Results
- ✅ Architecture & Design:     Pass — shared helper in `_lib-extract-pr.sh`, reused by `block-merge-on-red-ci.sh`; fail-safe (block, not guess) is the correct posture for a CI gate.
- ✅ Code Quality:              Pass — regex widening is minimal and well-commented; redirection-stripping span discipline preserved.
- ✅ Testing:                   Pass — 24/24 in `test_extract_pr.sh` (incl. quoted `"$PR"` / `"$REPO"`), 29/29 in `test_block_unreviewed_merge.sh` (no regression).
- ✅ Security:                  Pass — no secrets; closes a real bypass (gate previously evaluated an unrelated PR's CI).
- ✅ Performance:               Pass — pure string matching, no added `gh` calls; blocks *before* any network call.
- ✅ PR Description & Glossary: Pass — Glossary present, `Closes #643`.
- ⚠ Summary Bullet Narrative:  Pass — bullets are narrative (what + why).
- ✅ Technical Decisions (AgDR): N/A — bug fix, no new decision beyond the one covered in the original review.
- ✅ Adopter Handbooks:         N/A — no handbooks loaded (diff is shell hooks, no language-bucket match).
- ⚠ Fallow Static Analysis:    N/A — diff is not JS/TS.

### Verification performed
- Widened regex catches **quoted + bare** variables: `"$PR"`, `'$PR'`, `${PR_NUMBER}`, `--repo "$REPO"` → all detected (`yes`).
- No false-positives: literal `656 --repo me2resh/apexyard` → `no`; `2>&1 | tail -5` redirect → `no`; `2>err.log` → `no`; gh-api literal path → `no`. The optional leading-quote anchor cannot match a literal number/slug, and `2>&1` is stripped before the positional-arg grep.
- End-to-end through the actual hook: both `gh pr merge "$PR" --repo "$REPO"` and the bare form → **exit 2** with the "re-run with literal values" message.
- `shellcheck -x .claude/hooks/_lib-extract-pr.sh` → clean.

### Issues Found
None.

### Suggestions
- nit: The PR body's Testing section still reads "22 pass (8 new cases)" — stale from before the 2 quoted-form tests landed. The suite now reports **24 pass (10 var cases)**. Non-blocking; worth a one-line edit for accuracy.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `d16faefab9c7fdc5fdc1bf9088ae81714fca00e8`

Issue #687, opened 2026-06-22
[Bug] merge-time design + architecture review gates share the no---repo cross-repo bypass (sibling of #669)


AbdElrahmaN31:
## Affected
`.claude/hooks/require-design-review-for-ui.sh` and `.claude/hooks/require-architecture-review.sh` (merge-time PR-diff gates) + the `_lib-pr-repo.sh` cross-repo helpers. This is the deferred **sibling audit** from #669 (which fixed the create-time gate `require-agdr-for-arch-pr.sh`).

## Given / When / Then
**Given** a split-portfolio v2 layout (private portfolio sibling repo) where the session cwd is the ops fork and the PR lives in the portfolio
**When** a merge command (`gh pr` + `merge <N>`) runs as `cd <portfolio> && …` — or plain, from the ops-fork cwd — with **no `--repo` flag**
**Then** `CMD_REPO` is empty → `REPO_FLAG` is empty → `gh pr diff <N>` resolves the PR number against the **ops-fork's** default repo, not the portfolio. The gate either **silently bypasses** (PR `#N` absent in the fork → empty diff → `exit 0`, gate never enforced) or evaluates the **wrong PR by number collision**. **Expected:** the gate evaluates the portfolio PR's real diff.

This is a **silent-bypass**, not a false-block — lower severity than #669 (a reproduced false-positive block). Not reproduced here; surfaced by the #669 sibling audit.

## Repro
1. Split-portfolio v2 setup; cwd = ops fork; portfolio is a sibling repo.
2. From the fork cwd, run the merge command (no `--repo`) for a portfolio PR carrying a UI file (or a design artifact).
3. Observe the design / architecture gate does not enforce: if `#N` doesn't exist in the fork, `gh pr diff` is empty → `exit 0` (bypass); if a fork PR `#N` exists, the wrong diff is evaluated.

## Framework version
v3.2.0

## Severity
minor (silent-bypass of a merge gate; the merge still passes the other gates — Rex + CEO. Workaround: pass `--repo <portfolio>` to the merge command, which sets `CMD_REPO` correctly.)

## Notes
**Suggested fix:** when `--repo` is absent, recover the target repo from a leading `cd <path>` prefix using the new `pr_cmd_cd_target` helper added to `_lib-pr-repo.sh` in #669, and derive `CMD_REPO` from that path's origin via `git_origin_repo`.

[...]

Issue #669, opened 2026-06-21
[Bug] require-agdr-for-arch-pr.sh false-positives on split-portfolio PRs created without --repo (no-flag bypass of the #464 cross-repo guard)


AbdElrahmaN31:
## Affected
`.claude/hooks/require-agdr-for-arch-pr.sh` + the cross-repo guard helper `.claude/hooks/_lib-pr-repo.sh` (`pr_repo_matches_cwd`). The same pattern likely affects the sibling PR-diff merge gates `require-design-review-for-ui.sh` and `require-architecture-review.sh` — see "Suggested fix" / audit note.

## Given / When / Then
**Given** a split-portfolio v2 layout (private portfolio sibling repo holding the registry + `projects/`), where the ops-fork working tree has recent commits touching an `agdr_trigger_paths` glob (e.g. `topologies/go-data-pipeline/handbooks/**`)
**When** a PR is created in the private portfolio repo via the natural `cd <portfolio> && gh pr create --base main --head <branch> --title '…' --body-file …` pattern **without a `--repo` flag** (the harness resets cwd to the ops fork between calls, and the `PreToolUse` hook fires *before* the in-command `cd` executes)
**Then** `require-agdr-for-arch-pr.sh` computes the diff against the **ops-fork** working tree (its cwd) and blocks the PR with phantom "architecture changes" listing the fork's trigger paths — none of which are in the portfolio PR's actual diff. **Expected:** the hook evaluates the portfolio PR's real diff (docs/registry only → no arch paths) and passes.

## Repro
1. Split-portfolio v2 setup; cwd = ops fork; the portfolio is a sibling repo.
2. Ensure the fork has a recent commit touching an `agdr_trigger_paths` glob, e.g. `topologies/go-data-pipeline/handbooks/**` (a routine reformat commit suffices).
3. In the portfolio repo, create a docs-only PR with `cd ../<portfolio> && gh pr create --base main --head <branch> --title 'chore(GH-N): …' --body-file body.md` (**no `--repo`**).
4. Observe the block: "PR diff includes architecture changes" listing the **fork's** `topologies/…/handbooks/*` paths, none present in the PR diff.

**Confirmed mechanism** (run from the ops fork):
- `git_origin_repo "$fork"` → the fork slug.
- `pr_repo_matches_cwd "<cmd with NO --repo>" "$fork"` → returns **0 (match)** → guard does NOT fire → hook proceeds to a cwd-rooted `git diff` against the fork → false positive.
[...]

Issue #464, opened 2026-05-30
[Bug] PR-create hooks resolve ops-fork instead of PR origin repo — cross-repo false-positives


atlas-apex:
## Given / When / Then

**Given** a session pinned to the apexyard ops-fork, working inside a clone of a sibling repo (`me2resh/apexyard-premium`),
**When** you run `gh pr create --repo me2resh/apexyard-premium ...` for a PR with a valid AgDR and an open premium ticket,
**Then** the AgDR + ticket-validate PreToolUse hooks evaluate the WRONG repo (the public ops-fork tree / upstream tracker) and falsely block — forcing `gh api` + marker workarounds.

## Driver

When working in a sibling repo (e.g. `me2resh/apexyard-premium`) under a session pinned to the apexyard ops-fork, two PreToolUse hooks resolve the WRONG repo and false-positive on **every** `gh pr create`:

1. **`require-agdr-for-arch-pr.sh`** — computes its "architecture changes" diff against the **ops-fork working tree** (the public `me2resh/apexyard` checkout), not the premium clone the PR actually targets. It lists framework paths (`.claude/migrations/`, `topologies/`, `handbooks/domain/`, `.github/workflows/`) that aren't in the premium PR's diff, and blocks even when the PR has a valid AgDR.
2. **`validate-pr-create.sh`** (ticket existence) — checks the ticket number against the upstream `me2resh/apexyard` tracker, where the premium ticket doesn't exist (or a same-numbered closed framework ticket lives), instead of the PR's actual origin repo `me2resh/apexyard-premium`.

This has forced **every premium PR this session** through `--body-file`-absolute-path gymnastics and, repeatedly, the `gh api` PR-create bypass + the `<!-- agdr: not-applicable -->` marker — workarounds that erode the gates' value and risk becoming the norm.

## Repro

1. In a session pinned to the apexyard ops-fork, `cd` into a clone of `me2resh/apexyard-premium`.
2. Run `gh pr create --repo me2resh/apexyard-premium --base main --head <branch> --title 'feat(#N): ...' --body-file <abs-path>` where the body has a valid AgDR ref and #N is an OPEN premium ticket.
3. `require-agdr-for-arch-pr.sh` blocks listing FRAMEWORK paths (`.claude/migrations/`, `topologies/`, `handbooks/domain/`) not in the premium diff; and/or `validate-pr-create.sh` reports #N "closed/missing" because it checked `me2resh/apexyard` instead of `me2resh/apexyard-premium`.
[...]

PR #688, opened 2026-06-22
fix(#669): re-root arch-PR diff to the command's cd-target

AbdElrahmaN31:
## Summary
- **Fixed phantom "architecture changes" blocks on cross-repo PRs** — `require-agdr-for-arch-pr.sh` computed its diff against the hook's own cwd. When a PR is created with `cd <repo> && gh pr create …` and **no `--repo` flag** (the natural split-portfolio / `workspace/<project>/` pattern), the PreToolUse hook fires *before* the in-command `cd` runs, so it diffed the **ops fork** instead of the PR's real repo — false-blocking docs-only PRs with arch paths that exist only in the fork. The #464 cross-repo guard only caught the explicit-mismatched-`--repo` case; the no-`--repo` path bypassed it.
- **New `pr_cmd_cd_target` helper in `_lib-pr-repo.sh`** — extracts a leading `cd <path> &&` / `cd <path>;` prefix (unquoted + single/double-quoted), so a PR-create hook can discover the directory the `gh` call is *about* to run in.
- **Re-rooted the diff to that target** — the hook now derives `DIFF_DIR` from the cd-target (only when it resolves to a real git tree; otherwise it falls back to `REPO_ROOT`, i.e. no worse than before), routes every working-tree git call through a `gitd()` wrapper, and feeds `DIFF_DIR` to the cross-repo guard. **Config and session markers stay anchored to the ops fork — only the diff tree moves**, so spike-exemption / ops-root resolution are untouched.
- **Three failing-first regression tests** — proven to fail against the pre-fix hook, now green: a docs-only cd-target PR passes; an arch cd-target PR without an AgDR still blocks (proves the re-root targets the right tree rather than just skipping); an arch cd-target PR with an AgDR passes.

## Testing
1. `bash .claude/hooks/tests/test_require_agdr_for_arch_pr.sh` → **18/0** (15 prior + 3 new; the 3 `[#669]` cases fail on the base branch, pass here).
2. `bash .claude/hooks/tests/test_pr_hooks_cross_repo.sh` → **18/0** (the #464 cross-repo guard, including the lib-missing degraded-WARN path, is not regressed).
3. `bash .claude/hooks/tests/test_validate_pr_create_head.sh` (8/0) and `…_upstream.sh` (7/0) — the other `_lib-pr-repo.sh` consumer is unaffected.

> Run the suites with `env -u CLAUDE_CODE_SESSION_ID -u APEXYARD_OPS_PIN_DIR` when invoking inside a live Claude Code session — otherwise the session's ops-root pin overrides the spike test's fixture ops-root (a pre-existing test-isolation quirk, unrelated to this change).

## Scope note — merge-time siblings deferred
`require-design-review-for-ui.sh` and `require-architecture-review.sh` share the no-`--repo` root cause, but as an **unreproduced silent-bypass** (not a false-block). Hardening them changes the repo-qualified marker filename they read, which risks converting the bypass into a false-**block** unless the marker write side (`/approve-architecture`, `/design-review`, `/approve-design`) computes the identical qualifier. Tracked separately in **#687**.

Refs #669

## Glossary
| Term | Definition |
|------|------------|
| split-portfolio v2 | apexyard mode where the public framework fork and the private portfolio (registry + `projects/`) live in separate sibling repos |
| cd-target | the directory in a leading `cd <path> && gh …` command prefix — where the `gh` call will actually run, which the PreToolUse hook cannot see via its own cwd |
| cwd-rooted diff | a `git diff` whose result depends on the hook process's working directory rather than the PR's real repo |
| `DIFF_DIR` / `gitd()` | the re-rooted working-tree directory and the `git -C "$DIFF_DIR"` wrapper that confines diff operations to it, leaving markers/config on the ops fork |
| cross-repo guard (#464) | the `pr_repo_matches_cwd` check that skips a PR-diff gate when the command targets a different repo than the diff tree |

### DIFF
--- a/.claude/hooks/_lib-pr-repo.sh
+++ b/.claude/hooks/_lib-pr-repo.sh
@@ -31,6 +31,13 @@
 #       (or if there is no `--repo` flag, in which case they implicitly match).
 #       Returns 1 if they differ (cross-repo context).
 #
+#   pr_cmd_cd_target <command>
+#       Echoes the path from a leading `cd <path> && …` (or `cd <path>; …`)
+#       prefix in the command, or empty string if absent. Used by PR-create
+#       hooks to discover the directory the `gh` call is ABOUT to run in — the
+#       PreToolUse hook fires before the in-command `cd` executes, so the
+#       hook's own cwd is NOT yet the PR's repo (me2resh/apexyard#669).
+#
 # USAGE
 # -----
 #   . "$(dirname "$0")/_lib-pr-repo.sh"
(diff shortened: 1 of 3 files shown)


atlas-apex (approved):
Thanks — nicely scoped and well tested.

**Reviewed: APPROVE.**

- Bug confirmed real: the no-`--repo` path bypasses the #464 cross-repo guard, so the hook diffed the ops-fork tree and phantom-blocked docs-only split-portfolio PRs.
- The gate is **not** weakened — verified adversarially: an empty/non-git cd-target falls back safely, a "decoy repo" can't dodge the gate (since the command runs in the same cd-target the diff is taken from), and re-rooted arch PRs still block.
- Backward-compatible: the no-cd / `--repo`-supplied path is byte-for-byte unchanged.

Coordination note: #689 is stacked on this branch, so this one should land first. No blocking issues. Reviewed at 0a54561.

Issue #485, opened 2026-06-03
[Bug] Review markers collide across repos — keyed by bare PR number, no repo qualifier


atlas-apex:
## Affected

`.claude/session/reviews/` marker scheme — consumed by `block-unreviewed-merge.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh`, `warn-stale-review-markers.sh`, and written by the `/approve-merge`, `/approve-design`, `/approve-architecture`, `/design-review` skills + the `code-reviewer` (Rex) and `solution-architect` (Tariq) agents.

## Given / When / Then

**Given** a portfolio governs more than one managed repo, and review markers are stored as `.claude/session/reviews/<pr>-<role>.approved` (e.g. `429-rex.approved`, `429-ceo.approved`) keyed by **bare PR number with no repo qualifier**,
**When** two different repos under the same ops fork each have a PR with the same number (e.g. a framework PR `#429` in one session and a managed-project PR `#429` in a later session — PR numbers are per-repo and routinely overlap across repos),
**Then** their markers collide on the same filename. A stale CEO/rex/architecture marker from repo A's PR #N silently sits in `reviews/` and is presented as if it belongs to repo B's PR #N — observed in practice as a leftover `429-ceo.approved` (`approved_by=user`, a prior session's SHA) appearing for a freshly-opened PR #429 in a *different* repo.

Expected: a marker is unambiguously scoped to one `(repo, pr)` pair, so a marker for one repo's PR can never be mistaken for — or overwrite — another repo's same-numbered PR.

## Repro

1. In a multi-repo portfolio, get a CEO/rex marker written for PR `#100` in repo A (any `/approve-merge`-style flow), so `.claude/session/reviews/100-ceo.approved` exists.
2. In a later session, open PR `#100` in repo B (different managed repo).
3. Inspect `.claude/session/reviews/` — `100-*.approved` from repo A is still there and is indistinguishable from a marker for repo B's PR #100. Approving/merging logic and `warn-stale-review-markers.sh` cannot tell which repo it belongs to.

## Framework version

v1.1.0

## Severity

[...]

PR #689, opened 2026-06-22
fix(#687): re-root merge-gate repo to cd-target + repo-aware marker writers

AbdElrahmaN31:
## Summary
- **Fixed the no-`--repo` cross-repo bypass in the two merge-time review gates** — `require-architecture-review.sh` and `require-design-review-for-ui.sh` derived the PR's repo only from a `--repo` flag or the `gh api .../merge` URL. In split-portfolio v2 (session cwd = ops fork, PR lives in a sibling portfolio repo) the natural `cd <portfolio> && gh pr merge <N>` form carries **neither**, so `CMD_REPO` came back empty: `gh pr diff` resolved `<N>` against the ops fork → empty diff → the gate **silently bypassed** (`exit 0`, never enforced), and the marker was looked up under the wrong (`unknown`) qualifier. This is the deferred merge-time sibling of #669 (which fixed the create-time gate).
- **Re-rooted the read side to the command's cd-target** — both hooks now extend the `CMD_REPO` chain with a cd-target step (`pr_cmd_cd_target` + `git_origin_repo`, the helpers added in #669) and **derive `REPO_FLAG` after the full resolution**, so `gh pr diff <N> --repo <portfolio>` targets the PR's real repo. The cd-target is used only when it resolves to a real git tree; otherwise it falls through to the existing fallback (no worse than before).
- **Made the marker writers repo-aware (the load-bearing half)** — a read-side-only fix would convert the silent-bypass into a **false-block**: the write side (`/approve-architecture`, `/approve-design`, `/design-review` → Tariq) runs as a skill/agent invocation from the ops-fork cwd with bare `gh pr view <N>` (no `--repo`, and a skill can't carry a `cd` prefix), so it wrote the marker under the *ops-fork* qualifier the fixed read side no longer looks for. These now accept an `owner/repo#N` hint and pass `--repo` to `gh pr view`, keying the marker on the same portfolio slug the read side computes — and **fail loud** (ask for `owner/repo#N`) rather than write a guessed qualifier.
- **Closed a test gap + proved the symmetry** — added a first end-to-end harness for `require-design-review-for-ui.sh` (previously **zero** e2e coverage; only the `ui_paths_exclude` filter was tested inline), plus `#687` cd-target, write→read **round-trip**, and wrong-qualifier cases on **both** gates. The no-marker and wrong-qualifier cases were verified to **fail against the pre-fix hook** and pass after — the explicit read/write symmetry check the issue's warning section demands.

## Scope
Scoped to the two gates #687 names. The rex/ceo gate (`block-unreviewed-merge.sh` + `/approve-merge` + the Rex agent) shares the identical latent no-`--repo` bug but is left untouched on **both** sides (so it stays internally consistent at status-quo); a follow-up should close it the same way.

## Testing
1. `bash bin/run-hook-tests.sh` → **71/0**.
2. `env -u CLAUDE_CODE_SESSION_ID -u APEXYARD_OPS_PIN_DIR bash .claude/hooks/tests/test_require_architecture_review.sh` → **22/0** (15 prior + the new cd-target/round-trip cases).
3. `env -u CLAUDE_CODE_SESSION_ID -u APEXYARD_OPS_PIN_DIR bash .claude/hooks/tests/test_require_design_review_for_ui.sh` → **22/0** (new file).
4. Failing-first check: with the two hook changes stashed, the `#687` no-marker + wrong-qualifier cases FAIL (gate returns `0`/bypass), confirming they exercise the fix.

> Run the suites with `env -u CLAUDE_CODE_SESSION_ID -u APEXYARD_OPS_PIN_DIR` inside a live Claude Code session — otherwise the session's ops-root pin overrides the fixture ops-root (pre-existing test-isolation quirk, same note as #688).

## Dependency
**Depends on #688** (the #669 fix) — this PR consumes `pr_cmd_cd_target` from `_lib-pr-repo.sh`, added there. Please merge #688 first; until then this PR's diff includes #688's commits. After #688 lands on `dev` I'll rebase this onto `dev` (dropping the #688 commits) and re-request review.

Refs #687 · Related: #669, #464 · Marker scheme: AgDR-0060 (repo-qualified markers).

## Glossary
| Term | Definition |
|------|------------|
| split-portfolio v2 | apexyard mode where the public framework fork and the private portfolio (registry + `projects/`) live in separate sibling repos |
| cd-target | the directory in a leading `cd <path> && gh …` prefix — where the `gh` call actually runs, which the PreToolUse hook cannot see via its own cwd (it fires pre-`cd`) |
| silent-bypass | a gate that returns success without actually enforcing its check (here: empty `gh pr diff` → `exit 0`) |
| repo-qualified marker | a review-approval marker filename keyed on the PR's `owner/repo` (`_lib-review-markers.sh`, #485 / AgDR-0060) — `<owner>__<repo>__<pr>-<role>.approved` |
| read/write symmetry | the invariant that the qualifier the gate (read) looks up equals the qualifier the approval skill (write) produces; breaking it turns a missing-marker bypass into a false-block |

### DIFF
--- a/.claude/agents/solution-architect.md
+++ b/.claude/agents/solution-architect.md
@@ -158,15 +158,28 @@ MARKER_HOME="${OPS_ROOT:-$REPO_ROOT}"
 . "$MARKER_HOME/.claude/hooks/_lib-review-markers.sh"
 mkdir -p "$MARKER_HOME/.claude/session/reviews"
 # Resolve the repo for the qualified marker name.
-PR_REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
+#
+# In split-portfolio v2 the PR lives in a SIBLING repo, so a bare
+# `gh pr view {number}` (resolved against this cwd = ops fork) returns the WRONG
+# repo — the marker then lands under the ops-fork qualifier and the merge gate,
+# which keys on the PR's real repo, can't find it (false-block). Prefer the repo
+# `/design-review` passes as its optional second arg; this is the SAME slug the
+# gate's read side derives from the merge command's cd-target (me2resh/apexyard#687).
+REPO="{repo}"   # /design-review's optional repo arg, or empty / literal when absent
+[ "$REPO" = "{repo}" ] && REPO=""
+if [ -z "$REPO" ]; then
+  REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
+fi
+PR_REPO="$REPO"
 ARCH_MARKER=$(review_marker_path "$PR_REPO" {number} architecture "$MARKER_HOME")
 ```
 
 ### The command
 
 ```bash
-# Option B (preferred) — the PR's HEAD on GitHub
-gh pr view {number} --json headRefOid --jq .headRefOid > "$ARCH_MARKER"
+# Option B (preferred) — the PR's HEAD on GitHub. Pass --repo so the SHA is the
+# portfolio PR's HEAD, not an ops-fork PR with the same number (#687).
+gh pr view {number} ${REPO:+--repo "$REPO"} --json headRefOid --jq .headRefOid > "$ARCH_MARKER"
 ```
 
 ### Content — MUST be bare SHA + newline
--- a/.claude/hooks/_lib-pr-repo.sh
+++ b/.claude/hooks/_lib-pr-repo.sh
@@ -31,6 +31,13 @@
 #       (or if there is no `--repo` flag, in which case they implicitly match).
 #       Returns 1 if they differ (cross-repo context).
 #
+#   pr_cmd_cd_target <command>
+#       Echoes the path from a leading `cd <path> && …` (or `cd <path>; …`)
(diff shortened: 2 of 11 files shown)


atlas-apex (approved):
Thanks — careful work on the most security-sensitive part of the framework.

**Reviewed: APPROVE** (code review + an independent security pass).

- Bug confirmed: the merge-time architecture/design gates shared the same no-`--repo` silent bypass as #669.
- Fail-closed preserved: an absent marker still blocks; `MARKER_HOME` resolves from the session-pinned ops fork (not the cd-target), so a crafted `cd` can't redirect where markers are read; write-path == read-path symmetry verified across single-fork / split-portfolio / workspace-clone; SHA binding still uses the real GitHub PR HEAD; both gate shapes stay enforced.
- Two notes: (1) **land #688 first** — this branch is stacked on it (rebase if #688 changes). (2) The primary `block-unreviewed-merge.sh` (rex/ceo) gate shares the same latent no-`--repo` resolution and is intentionally untouched here — worth a follow-up before relying on split-portfolio merges heavily (you flagged this too — agreed).

No blocking issues. Reviewed at c00e644.

PR #691, opened 2026-06-22
fix(#690): make hook directory-walks and upstream fetch cross-platform

### LINKED ISSUE #690: [Bug] Windows: SessionStart hooks hang the session — ancestor-dir walks never terminate on drive paths
## Given / When / Then

**Given** an ApexYard fork opened in Claude Code on Windows (Git Bash as the hook shell)
**When** a normal session starts and SessionStart hooks run
**Then** the session hangs on a blank screen / never responds. Expected: hooks finish and the session becomes interactive.

## Repro

1. Open an ApexYard fork in Claude Code on Windows with Git Bash, no `.mcp.json` present.
2. Start a normal session (not `--safe-mode`).
3. `remind-mcp-tools.sh` (SessionStart) spins forever → blank session.

**Root cause:** ancestor-directory walks of the form `while [ "$d" != "/" ]; do d=$(dirname "$d"); done` start from `git rev-parse --show-toplevel`, which on Git Bash returns a Windows drive path (`C:/…`). `dirname` reduces that to `C:`, and `dirname C:` == `C:` — never `/` — so the loop never terminates. 17 hook/lib loops share this pattern.

Separately, `check-upstream-drift.sh` uses `timeout 5 git fetch`, which can't bound git on Windows (git ignores SIGTERM) and is a silent no-op on macOS (no `timeout`; it's `gtimeout`).

## Environment

- ApexYard version / commit: v3.3.0 (5b61952)
- OS + shell: Windows 11, Git Bash (GNU bash 5.2)
- Single-fork or split-portfolio v2: single-fork

## Severity

**blocker on Windows** — the framework is unusable (blank session) until the walk terminates. Fix PR incoming (no-progress guard on all walk loops + portable fetch timeout).

HishamM1:
## Summary

- Ancestor-directory walks (`while [ "$d" != "/" ]; do d=$(dirname "$d"); done`) never terminate on Windows/Git Bash: `git rev-parse --show-toplevel` yields a drive path (`C:/…`) and `dirname` bottoms out at `C:`, never `/`. SessionStart hooks such as `remind-mcp-tools.sh` then spin forever and leave the Claude Code session blank. Added a no-progress guard (`[ "$parent" = "$d" ] && break`) to all 17 walk loops — a no-op on Linux/macOS where the walk already reaches `/`.
- `check-upstream-drift.sh` used `timeout 5 git fetch`, which cannot bound git on Windows (git ignores SIGTERM) and silently no-ops on macOS (no `timeout`; it is `gtimeout`). Now detects `timeout`/`gtimeout`, hard-kills with `-k`, and sets `GIT_TERMINAL_PROMPT=0` so a credential prompt cannot stall startup.

## Testing

- `bash bin/run-hook-tests.sh` → **PASS=70 FAIL=0** (Linux).
- Windows (Git Bash): full SessionStart hook chain now completes; previously `remind-mcp-tools.sh` looped forever (blank session).
- Guard verified as a no-op on POSIX paths — the walk still terminates at `/`.

## Glossary

- **Ancestor walk** — climbing parent directories to locate the ops-root marker (`.apexyard-fork`).
- **Ops root** — the fork root ApexYard hooks resolve to.
- **SessionStart hook** — a hook Claude Code runs at session start; a hang here blanks the UI.

Resolves #690.

### DIFF
--- a/.claude/hooks/_lib-migration-chain.sh
+++ b/.claude/hooks/_lib-migration-chain.sh
@@ -52,7 +52,7 @@ _migration_ops_root() {
       echo "$cur"
       return 0
     fi
-    cur=$(dirname "$cur")
+    parent=$(dirname "$cur"); [ "$parent" = "$cur" ] && break; cur="$parent"
   done
   echo "$r"
 }
--- a/.claude/hooks/_lib-ops-root.sh
+++ b/.claude/hooks/_lib-ops-root.sh
@@ -100,7 +100,7 @@ resolve_ops_root_walk() {
       printf '%s' "$r"
       return 0
     fi
-    r=$(dirname "$r")
+    parent=$(dirname "$r"); [ "$parent" = "$r" ] && break; r="$parent"
   done
   return 0
 }
--- a/.claude/hooks/_lib-portfolio-paths.sh
+++ b/.claude/hooks/_lib-portfolio-paths.sh
@@ -78,7 +78,7 @@ _portfolio_root() {
       echo "$cur"
       return 0
     fi
-    cur=$(dirname "$cur")
+    parent=$(dirname "$cur"); [ "$parent" = "$cur" ] && break; cur="$parent"
   done
 
   # Not under an apexyard fork — fall back to git toplevel so callers
--- a/.claude/hooks/apply-agent-routing.sh
+++ b/.claude/hooks/apply-agent-routing.sh
@@ -82,7 +82,7 @@ else
       ROOT="$cur"
       break
     fi
-    cur=$(dirname "$cur")
+    parent=$(dirname "$cur"); [ "$parent" = "$cur" ] && break; cur="$parent"
   done
 fi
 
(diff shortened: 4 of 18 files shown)


atlas-apex (approved):
Thanks, Hisham — clean cross-platform fix.

**Reviewed: APPROVE** (code review + an independent security pass).

- Bug confirmed real: the directory-walk loops infinite-loop on Windows drive paths (`C:/…`), where `dirname` bottoms out at `.`/`C:` and never reaches `/`, hanging SessionStart hooks; and stock macOS has no `timeout`, so the upstream fetch was effectively unbounded. Both reproduced.
- Backward-compatible & semantically equivalent: the same one-line guard (`parent=$(dirname "$x"); [ "$parent" = "$x" ] && break`) across all 17 walk loops. On POSIX the new break is dead code (the loop already exits at `/`), so path / ops-root / portfolio resolution is byte-identical — every gate reads the same markers/config. The fetch change is security-positive (`GIT_TERMINAL_PROMPT=0` + a real `timeout`/`gtimeout` bound).
- Verified no enforcement weakening: ops-root anchoring is provably equivalent across single-fork / split-portfolio / workspace-clone; no fail-open, no symlink/traversal change, no injection surface.
- One non-blocking nit: declare `parent` as `local` in `_lib-ops-root.sh`, `_lib-portfolio-paths.sh`, `_lib-migration-chain.sh` for hygiene (no behavioral bug today).

Coordination: this is textually independent of #688/#689, so it can land in any order. No blocking issues. Reviewed at ab62b79.

me2resh:
@HishamM1 Thank you for your contribution to ApexYard, including this PR.

PR #697, opened 2026-06-24
fix(#695): read --body-file when checking the skip marker + required sections

### LINKED ISSUE #695: [Bug] validate-issue-structure.sh ignores the skip marker when body is passed via --body-file
## Given / When / Then
**Given** a legitimately off-template ticket (e.g. an epic / meta-thread) whose body contains the documented `<!-- validate-issue-structure: skip -->` marker,
**When** I create it with a `--body-file <file>` (rather than an inline body string),
**Then** `validate-issue-structure.sh` still BLOCKS with "missing section …" because it does not read the `--body-file` contents when looking for the skip marker.

## Repro
- [ ] Put the skip marker in a body file and create the issue via `--body-file that-file` → blocked anyway.
- [ ] An inline body string containing the same marker IS honored; `--body-file` is not.
- [ ] Fix: when the command uses `--body-file` / `-F body=@…`, read the referenced file to detect the skip marker (and to validate the required sections).

atlas-apex:
## Summary

- **Fixed silent validation bypass for the gh-API field-file shape** — `validate-issue-structure.sh` never saw an issue body supplied via `gh issue create -F body=@<path>` (the canonical REST way to post a body from a file). Two distinct gaps combined: the title couldn't be parsed when a single-dash flag followed it, and the `@`-prefixed file path was never resolved. As a result a legitimately off-template epic carrying the `<!-- validate-issue-structure: skip -->` marker via the field-file shape was **blocked** (skip marker invisible), and — worse — a field-file body missing a required section was **silently allowed** (sections invisible). This is the user-facing symptom in #695.
- **Widened the flag-boundary anchor in `extract_flag_value` to `-{1,2}[a-zA-Z]`** — the greedy quoted-value matcher previously anchored only on double-dash (`--`) flags, so `--title "[Feature] F" -F body=@file` fell through to the unquoted fallback and extracted `"[Feature]` (leading quote, truncated). The bracketed prefix then failed to resolve and the hook exited 0. The anchor now stops at single-dash (`-F`, `-f`, `-t`) flags too, so the title — and any quoted value followed by a single-dash flag — extracts cleanly.
- **Added a dedicated `body=@<path>` field extractor** — the existing `-F` handler rejected any value containing `=`, so `body=@<path>` was discarded. A new block reads the path after the `@` sigil from `body=@<path>` / `body@=<path>` on `-F` / `--field` / `-f`, feeding the file content into the same `FULL_BODY` that the skip-marker and section checks already inspect.
- **Inline `--body` and `--body-file <path>` behaviour is unchanged** — verified by the existing suite plus new regression cases.
- **Tests** — six new cases in `test_validate_issue_structure.sh` cover skip-marker honour and missing-section blocking across `--body-file`, `-F body=@`, `--field body=@`, and `-f body=@`. The full hook suite is green (71/71); the per-hook suite is 30/30.

Closes #695

## Testing

1. `bash .claude/hooks/tests/test_validate_issue_structure.sh` → `Passed: 30  Failed: 0`
2. `bash bin/run-hook-tests.sh` → `PASS=71  FAIL=0  SKIP=0  TOTAL=71`
3. Manual: `gh issue create --title '[Feature] F' -F body=@<file-with-missing-sections>` now exits 2 (blocks); the same with the skip marker exits 0 with a `validate-issue-structure bypassed` warning.

## Glossary

| Term | Definition |
|------|------------|
| Skip marker | The HTML comment `<!-- validate-issue-structure: skip -->` that, when present in an issue body, makes the hook exit 0 (a visible, logged bypass) so legitimately off-template tickets like epics aren't blocked. |
| `-F body=@<path>` | A `gh`/`gh api` field flag whose value, prefixed with `@`, names a file whose contents become the field value (here, the issue body). The canonical REST shape for posting a body from a file. |
| Flag-boundary anchor | The regex fragment in `extract_flag_value` that marks where a quoted flag value ends — i.e. the next CLI flag or end-of-string. Previously matched only `--`-style flags; now matches single- and double-dash flags. |
| `FULL_BODY` | The combined body text the hook validates: inline `--body` plus any resolved `--body-file` / `-F body=@` file content. |
| PreToolUse hook | A Claude Code hook that runs before a tool call (here, `Bash` invoking `gh issue create`) and can block it by exiting non-zero. |

https://claude.ai/code/session_01UQWFvN2vzy1UtFX2FyaDvb

### DIFF
--- a/.claude/hooks/tests/test_validate_issue_structure.sh
+++ b/.claude/hooks/tests/test_validate_issue_structure.sh
@@ -291,6 +291,55 @@ run_case "--body-file path with valid Feature body → pass" \
   0 "" \
   "gh issue create --title '[Feature] X' --body-file $BODY_FILE_PATH"
 
+# --- body-file: skip marker honoured (me2resh/apexyard#695) -----------------
+# An off-template epic body that carries the skip marker must bypass
+# validation when supplied via --body-file, exactly like the inline --body
+# case above.
+SKIP_FILE_PATH="$TMPDIR/skip-body.md"
+cat > "$SKIP_FILE_PATH" <<'EOF'
+free-form epic body, no required sections.
+<!-- validate-issue-structure: skip -->
+EOF
+
+run_case "--body-file with skip marker → bypass (#695)" \
+  0 "validate-issue-structure bypassed" \
+  "gh issue create --title '[Feature] epic via file' --body-file $SKIP_FILE_PATH"
+
+# --- body-file: missing required section still blocks -----------------------
+MISSING_FILE_PATH="$TMPDIR/missing-body.md"
+cat > "$MISSING_FILE_PATH" <<'EOF'
+## User Story
+As a user I want a thing.
+EOF
+
+run_case "--body-file missing Acceptance Criteria → block (#695)" \
+  2 "missing section: ## Acceptance Criteria" \
+  "gh issue create --title '[Feature] incomplete via file' --body-file $MISSING_FILE_PATH"
+
+# --- gh-api field form: -F body=@<path> (me2resh/apexyard#695) --------------
+# `gh api … -F body=@<file>` is the REST shape for posting an issue body from
+# a file. Pre-#695 the field-value path after the `@` sigil was never resolved
+# (the `-F` handler rejected any value containing `=`), AND a single-dash flag
+# after a quoted --title broke title extraction — so the skip marker and the
+# section checks were both blind to the file content. Cover skip-marker-honour
+# and missing-section-block across -F / --field / -f.
+
+run_case "-F body=@file with skip marker → bypass (#695)" \
+  0 "validate-issue-structure bypassed" \
+  "gh issue create --title '[Feature] epic via field' -F body=@$SKIP_FILE_PATH"
+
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
**Rex code review — APPROVED.**

Bug confirmed real (proven against the parent commit): the hook only read the inline `--body`, so a body passed via `--body-file` / `-F body=@<path>` bypassed both the skip-marker check and the required-section validation. A flag-boundary parse bug also truncated `--title "[Feature] X"` when a single-dash flag (`-F`) followed it. The fix reads the referenced file for both checks and widens the flag-boundary anchor to `-{1,2}[a-zA-Z]`.

No over-broadening: inline `--body` behaviour is unchanged; the `body=@<path>` extractor triggers only for `-F`/`-f`/`--field`; a missing file is handled safely; injection-probed (path flows only into quoted `[ -f ]`/`cat`, no eval).

Tests: `test_validate_issue_structure.sh` 30/0 (incl. 6 new cases), `bin/run-hook-tests.sh` 71/0; `bash -n` + shellcheck clean. No blocking findings.

_Reviewed at 206df62 by the code-reviewer (Rex). The local approval marker is recorded; this comment is the human-readable record._

me2resh (approved):

PR #698, opened 2026-06-24
fix(#693): re-root validate-pr-create + validate-branch-name to the cd-target

### LINKED ISSUE #693: [Bug] validate-pr-create.sh + validate-branch-name.sh aren't cd-target-aware (cross-repo PR/push misfire)
## Given / When / Then
**Given** an ops-fork Claude session whose working directory is the fork (e.g. checked out on `dev`),
**When** I run a `cd <other-managed-repo> && …` PR-create (or push a ticket-compliant branch) for a *different* repo,
**Then** `validate-pr-create.sh` / `validate-branch-name.sh` read the **ops-fork's** current branch (`dev`) instead of the target repo's branch, and BLOCK with e.g. "Branch 'dev' missing ticket ID" — a false positive that forced an API workaround.

This is the same class as #669 / #687 (which fixed `require-agdr-for-arch-pr.sh` + the merge gates by re-rooting to the command's `cd`-target), but these two PR/branch validators were **not** covered.

## Repro
- [ ] From the ops fork on `dev`, run a PR-create for a fully-compliant `feature/GH-NNN-…` branch in a sibling managed repo → blocked with "Branch 'dev' missing ticket ID".
- [ ] Fix: make both hooks `cd`-target aware (reuse the `pr_cmd_cd_target` helper from #669) so they validate the branch in the repo the command actually runs in.

atlas-apex:
## Summary

- **Re-rooted `validate-branch-name.sh` to the command's `cd`-target** — when a `git push` carries no explicit source ref, the hook fell back to `git branch --show-current` in its **own** cwd. Because the PreToolUse hook fires *before* the shell runs `cd <other-repo> && git push`, that cwd is still the ops fork (typically on `dev`), so pushing for a *different* managed repo false-blocked with "Branch 'dev' missing ticket ID". The fallback now resolves via `git -C <cd-target>` using `pr_cmd_cd_target` from `_lib-pr-repo.sh`. Same fix class as #669/#687 (which re-rooted the arch-PR hook + merge gates).
- **Re-rooted `validate-pr-create.sh`'s branch-ticket-ID check the same way** — its local-HEAD fallback (used when `--head` is absent) had the identical cwd bug for `cd <other-repo> && gh pr create`. The `--head` path and the cwd-independent **PR-title existence check** are deliberately untouched.
- **Not weakened — the gate still fires on a genuinely non-compliant target.** A non-conforming branch *in the cd-target* still BLOCKS (true-negative regression case `(c)` in both test files). The proof is `#693: cd-target on non-conforming branch … → still BLOCKS`. The no-`cd` path stays byte-for-byte equivalent (case `(d)` in both files), and when there is no leading `cd` (or the target isn't a git tree) the resolution falls through to the original cwd behaviour.
- **Regression tests, failing-first verified** — extended `test_validate_branch_name_pushref.sh` (+6 cases) and `test_validate_pr_create_head.sh` (+5 cases). Stashing the hook fix makes the two false-positive cases fail (rc=2 instead of 0) while the true-negative still blocks, confirming the tests pin the bug. Full `bin/run-hook-tests.sh` suite is green (55 PASS, 0 FAIL); both hooks and tests are `bash -n` clean.

## Testing

1. `bash .claude/hooks/tests/test_validate_branch_name_pushref.sh` → 34 PASS, 0 FAIL.
2. `bash .claude/hooks/tests/test_validate_pr_create_head.sh` → 13 PASS, 0 FAIL.
3. `bash bin/run-hook-tests.sh` → exit 0, all 55 hook tests PASS.
4. Failing-first proof: `git stash push -- .claude/hooks/validate-branch-name.sh .claude/hooks/validate-pr-create.sh` then re-run the two test files — the `cd-target … → PASS (false-positive gone)` cases fail (rc=2), the `cd-target on non-conforming branch … → still BLOCKS` case still passes. `git stash pop` restores the fix.

Closes #693

## Glossary

| Term | Definition |
|------|------------|
| PreToolUse hook | A shell hook the harness runs *before* the tool command executes. Its cwd is the session's working dir — a leading `cd <path> &&` in the command has not run yet, so the hook must parse the cd-target itself to know where the command will actually run. |
| cd-target | The directory in a `cd <path> && <cmd>` command. `pr_cmd_cd_target` (in `_lib-pr-repo.sh`, added by #669) extracts `<path>` from a leading `cd`. |
| Re-rooting | Running working-tree git operations against the cd-target's repo (`git -C <cd-target>`) instead of the hook's own cwd, so a cross-repo command is evaluated against the repo it really targets. |
| False positive (here) | The gate blocking a compliant action — pushing/PR-ing a correctly-named branch in another managed repo got blocked because the hook read the ops-fork's `dev` branch instead. |
| True negative (here) | The gate correctly blocking a genuinely non-compliant branch in the target repo — preserved by this fix, proving the gate was not turned into a no-op. |
| Source ref / `--head` | The explicit branch a `git push origin <ref>` / `gh pr create --head <ref>` names. When present it is already authoritative (no cwd dependence); the bug only affected the no-ref / no-`--head` fallback. |

### DIFF
--- a/.claude/hooks/tests/test_validate_branch_name_pushref.sh
+++ b/.claude/hooks/tests/test_validate_branch_name_pushref.sh
@@ -27,6 +27,7 @@ SRC_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
 HOOK_SRC="$SRC_ROOT/.claude/hooks/validate-branch-name.sh"
 LIB_SRC="$SRC_ROOT/.claude/hooks/_lib-extract-push-ref.sh"
 LIB_CONFIG_SRC="$SRC_ROOT/.claude/hooks/_lib-read-config.sh"
+LIB_PR_REPO_SRC="$SRC_ROOT/.claude/hooks/_lib-pr-repo.sh"
 
 for f in "$HOOK_SRC" "$LIB_SRC"; do
   if [ ! -f "$f" ]; then
@@ -62,6 +63,9 @@ make_sandbox_with_wrong_local_branch() {
   cp "$LIB_SRC"  "$sb/.claude/hooks/_lib-extract-push-ref.sh"
   if [ -f "$LIB_CONFIG_SRC" ]; then
     cp "$LIB_CONFIG_SRC" "$sb/.claude/hooks/_lib-read-config.sh"
+  fi
+  if [ -f "$LIB_PR_REPO_SRC" ]; then
+    cp "$LIB_PR_REPO_SRC" "$sb/.claude/hooks/_lib-pr-repo.sh"
   fi
   if [ -f "$SRC_ROOT/.claude/project-config.defaults.json" ]; then
     cp "$SRC_ROOT/.claude/project-config.defaults.json" "$sb/.claude/project-config.defaults.json"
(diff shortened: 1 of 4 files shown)


atlas-apex (commented):
**Rex code review — APPROVED.** (security-sensitive: PR/branch enforcement validators)

Bug confirmed real: both validators resolved the branch in the hook's own cwd (the ops fork, on `dev`), ignoring the command's leading `cd`-target — so a push / PR-create for a *different* managed repo via `cd <repo> && …` mis-fired with a false "Branch 'dev' missing ticket ID" block. Same class as #669/#687. The fix re-roots branch resolution to the `cd`-target via the `pr_cmd_cd_target` helper (added by #669), mirroring how `require-agdr-for-arch-pr.sh` consumes it.

**Not weakened — verified:**
- The no-`cd` path is byte-for-byte equivalent (normal single-repo pushes/PRs unchanged).
- A genuinely non-conforming branch / PR title in the target repo **still BLOCKS** — true-negative tests assert exit 2 (so the re-root targets the right tree, not a blanket no-op).
- The `--head` path and the cwd-independent PR-title check are untouched.
- `cd`-target is used only as `git -C "$CD_TARGET"` (fully quoted, no eval); injection-probed.

Tests: `test_validate_branch_name_pushref.sh` 34/0, `test_validate_pr_create_head.sh` 13/0, `bin/run-hook-tests.sh` 71/0. No blocking findings.

_Reviewed at 2e12a12 by the code-reviewer (Rex)._

me2resh (approved):

Issue #714, opened 2026-06-25
[Docs] README — AY-monogram lockup + ApexYard title-case + tagline/badge refresh


atlas-apex:
## Driver
The framework README is the first thing adopters see on GitHub. It should carry the locked AY-monogram brand and consistent "ApexYard" title-case so the repo reads as the same product as the site and dashboard. Brand locked 2026-06-25 to Direction A (AY monogram, Swiss Graphite blue, Archivo wordmark).

## Context
This change targets the framework README on `me2resh/apexyard` (framework PRs go to `dev`). Master logo assets authored (`apexyard-lockup.svg`); design lives in claude.ai/design project "ApexYard Brand — Selected". Layla's brand audit also recommended a two-tier tagline and a badge-lineup refresh — fold those in here.

## Acceptance Criteria
- [ ] README header shows the AY-monogram lockup (hosted SVG/PNG, light + dark friendly)
- [ ] "ApexYard" rendered title-case consistently throughout (no "apexyard"/"ApexYard" drift except where it's a CLI/path literal)
- [ ] Tagline updated to the locked two-tier line ("Ship AI-built software like a real engineering team") + brand closer ("where projects get forged")
- [ ] Badge lineup refreshed: add stars / version / site badges; drop "PRs welcome"
- [ ] Links and images verified (no broken refs)

## Design Notes
Keep it Swiss Graphite — blue accent, clean. The lockup goes at the very top; the tagline directly under it.

## Out of Scope
- Restructuring README content/sections (branding pass only)
- Site re-skin (apexyard-site#6) and dashboard sidebar (premium, separate ticket)

## Glossary
| Term | Definition |
|------|------------|
| AY monogram | Locked logo (Direction A) — "AY" in IBM Plex Mono on a sharp blue square |
| Lockup | Mark + "ApexYard" wordmark used together |
| Two-tier tagline | A functional line + a brand closer, per Layla's brand audit |

PR #730, opened 2026-06-27
feat(#729): add Codex adapter generator

### LINKED ISSUE #729: [Feature] Add Codex adapter generator
**[Feature] Add Codex adapter generator**

## User Story

As a Codex user adopting ApexYard, I want a repeatable generator that derives Codex-compatible skills, agents, and hooks from the existing `.claude/` framework so that Codex support stays in sync without committing machine-specific migration output.

## Acceptance Criteria

- [ ] Generated first-pass `.agents/` and `.codex/` output is not accidentally committed from local exploration.
- [ ] A documented sync/generator command treats `.claude/` as the source of truth.
- [ ] Generated Codex files avoid absolute local paths and machine-specific state.
- [ ] Generated paths use consistent repo casing and work on case-sensitive filesystems.
- [ ] Tests or smoke checks cover path normalization and drift detection.
- [ ] Documentation explains whether generated Codex artifacts are tracked, ignored, or produced locally.

## Design Notes

No UI changes. This is framework tooling for multi-agent compatibility.

## Out of Scope

Full parity with every future Codex runtime feature; this ticket only establishes the generator/sync foundation and safe local workflow.

## Effort Estimate

TBD

## Glossary

| Term | Definition |
|------|------------|
| Codex adapter | Repo-local files that let Codex consume ApexYard skills, agents, and hooks. |
| Source of truth | The canonical framework files under `.claude/` that generated mirrors should derive from. |
| Drift detection | A check that reports when generated files no longer match their `.claude/` inputs. |

hossam-96:
## Summary

- Adds `bin/sync-codex-adapter.sh` so Codex-facing adapter files can be generated from the canonical `.claude/` runtime without committing machine-local migration output.
- Keeps `.claude/hooks/*.sh` as the single audited gate implementation: generated `.codex/hooks.json` delegates to the unmodified `.claude/hooks` commands instead of copying or rewriting hook bodies.
- Converts Claude Code handler-level `if` predicates into shell-side filters inside generated commands, so `.codex/hooks.json` uses Codex's documented matcher-group + command-handler shape without unsupported `if` fields.
- Maps Claude agent model labels to Codex-native model labels (`opus` → `gpt-5.5`, `sonnet` → `gpt-5.4`, `haiku` → `gpt-5.4-mini`) so generated Codex agents can run under Codex.
- Documents the adapter tracking policy, clean regeneration path, model mapping, Codex hook delegation model, and drift-check workflow.
- Records AgDR-0088 to explain the enforcement-faithful adapter shape and reference the existing pi adapter precedent in AgDR-0082.
- Adds a smoke test that verifies skill path rewriting, model mapping, no absolute clone paths, no generated `.codex/hooks` fork, no unsupported hook `if` fields, command predicate wrapping, block/allow exit-code preservation, predicate skip behavior, and `--check` drift detection.

## Testing

- `bash -n bin/sync-codex-adapter.sh .claude/hooks/tests/test_sync_codex_adapter.sh`
- `bash .claude/hooks/tests/test_sync_codex_adapter.sh` (PASS=22 FAIL=0)
- `bin/sync-codex-adapter.sh --check`
- `jq -e '.. | objects | select(has("if"))' .codex/hooks.json` returns no matches after regeneration
- `bin/run-hook-tests.sh` (PASS=92 FAIL=1 SKIP=0 TOTAL=93)
  - The single failure is `.claude/hooks/tests/test_block_merge_on_red_ci.sh`, and it reproduces unchanged on a clean `upstream/dev` worktree. It is not introduced by this PR's adapter changes.

## AgDR

- `docs/agdr/AgDR-0088-codex-adapter-generation.md`

## Glossary

| Term | Definition |
|------|------------|
| Codex adapter | Generated repo-local files that let Codex consume ApexYard skills, agents, and hook wiring while delegating gate decisions to canonical `.claude/hooks/*.sh` scripts. |
| Source of truth | The canonical `.claude/` runtime files that adapter output is generated from. |
| Hook delegation | The adapter pattern where `.codex/hooks.json` invokes unmodified `.claude/hooks/*.sh` commands instead of maintaining copied hook logic. |
| Predicate wrapper | Generated shell preflight logic that implements Claude Code's command-specific `if` predicates without emitting unsupported `if` fields in Codex hook JSON. |
| Model mapping | Translation from Claude runtime labels such as `opus` to Codex model identifiers such as `gpt-5.5`. |
| Drift check | A `--check` mode that fails when generated adapter output differs from the current `.claude/` source. |
| Clean regeneration | A `--clean` mode that removes existing generated adapter trees before writing fresh output, preventing stale renamed or deleted files from lingering. |

Closes #729

### DIFF
--- a/.claude/harness-models.json
+++ b/.claude/harness-models.json
@@ -1,0 +1,6 @@
+{
+  "_comment": "Single source of truth mapping a framework model-tier label (the Claude Code model aliases used in agent frontmatter) to the concrete model each harness runs. Every harness adapter reads this one file: the Codex adapter uses the 'codex' column; pi/opencode adapters add their own column (e.g. 'pi', 'opencode') here rather than hardcoding a second mapping. Per AgDR-0087 (reasoning agents require a frontier-tier model), keep each harness's 'opus' row on that harness's strongest available model — never downgrade the reasoning tier when adding a harness. Absent tier or column falls back to the label unchanged.",
+  "opus":   { "codex": "gpt-5.5" },
+  "sonnet": { "codex": "gpt-5.4" },
+  "haiku":  { "codex": "gpt-5.4-mini" }
+}
--- a/.claude/hooks/tests/test_sync_codex_adapter.sh
+++ b/.claude/hooks/tests/test_sync_codex_adapter.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+# Smoke tests for bin/sync-codex-adapter.sh.
+
+set -u
+
+ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+SCRIPT="$ROOT/bin/sync-codex-adapter.sh"
+
+red()   { printf '\033[31m%s\033[0m\n' "$*"; }
+green() { printf '\033[32m%s\033[0m\n' "$*"; }
+
+PASS=0
+FAIL=0
+FAILED=""
+
+mark_pass() { green "  ok   $1"; PASS=$((PASS+1)); }
+mark_fail() {
+  red "  FAIL $1: $2" >&2
+  FAIL=$((FAIL+1))
+  FAILED="$FAILED $1"
+}
+
+assert_file() {
+  local path="$1" label="$2"
+  [ -f "$path" ] && mark_pass "$label" || mark_fail "$label" "missing $path"
+}
+
+assert_contains() {
+  local path="$1" pattern="$2" label="$3"
+  grep -F "$pattern" "$path" >/dev/null 2>&1 && mark_pass "$label" || mark_fail "$label" "missing pattern [$pattern] in $path"
+}
+
+assert_not_contains() {
+  local path="$1" pattern="$2" label="$3"
+  if grep -F "$pattern" "$path" >/dev/null 2>&1; then
+    mark_fail "$label" "unexpected pattern [$pattern] in $path"
+  else
+    mark_pass "$label"
+  fi
+}
+
+TMPROOT=$(mktemp -d "${TMPDIR:-/tmp}/codex-adapter-test.XXXXXX")
+trap 'rm -rf "$TMPROOT"' EXIT
+
+mkdir -p "$TMPROOT/.claude/skills/status" "$TMPROOT/.claude/hooks" "$TMPROOT/.claude/agents" "$TMPROOT/.claude/registries"
+
+cat > "$TMPROOT/.claude/skills/status/SKILL.md" <<'MD'
+---
+name: status
+description: Current status.
+---
+
+Source `.claude/hooks/_lib-portfolio-paths.sh` and read `.claude/skills/status/briefing.sh`.
+MD
+
+cat > "$TMPROOT/.claude/hooks/detect-role-trigger.sh" <<'SH'
+#!/usr/bin/env bash
+echo "from .claude/hooks"
+SH
+
+cat > "$TMPROOT/.claude/settings.json" <<'JSON'
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/detect-role-trigger.sh\"'"
+          }
+        ]
+      }
+    ]
+  }
+}
+JSON
+
+cat > "$TMPROOT/.claude/agents/backend-engineer.md" <<'MD'
+---
+name: backend-engineer
+description: Implements backend work.
+model: sonnet
+allowed-tools: Bash, Read
+---
+
+# Backend Engineer
+
+Read `.claude/rules/workflow-gates.md` before acting in Claude Code.
+MD
+
+cat > "$TMPROOT/.claude/registries/ai-crawlers.json" <<'JSON'
+[]
+JSON
+
+echo "== Codex adapter sync smoke"
+
+if bash "$SCRIPT" --root "$TMPROOT" >/tmp/_codex_adapter_sync.out 2>&1; then
+  mark_pass "generator writes adapter output"
+else
+  mark_fail "generator writes adapter output" "$(cat /tmp/_codex_adapter_sync.out)"
+fi
+
+assert_file "$TMPROOT/.agents/skills/status/SKILL.md" "skill mirror exists"
+assert_file "$TMPROOT/.codex/hooks/detect-role-trigger.sh" "hook mirror exists"
+assert_file "$TMPROOT/.codex/hooks.json" "hooks.json mirror exists"
+assert_file "$TMPROOT/.codex/agents/backend-engineer.toml" "agent TOML exists"
+assert_file "$TMPROOT/.codex/registries/ai-crawlers.json" "registry mirror exists"
+
+assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".codex/hooks/_lib-portfolio-paths.sh" "skill rewrites hook paths"
+assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".agents/skills/status/briefing.sh" "skill rewrites skill paths"
+assert_contains "$TMPROOT/.codex/hooks.json" '$r/.codex/hooks/detect-role-trigger.sh' "hooks.json rewrites hook exec path"
+assert_contains "$TMPROOT/.codex/hooks.json" '$HOME/.codex/apexyard' "hooks.json rewrites pin cache path"
+assert_not_contains "$TMPROOT/.codex/hooks.json" "$TMPROOT" "hooks.json has no absolute fixture path"
+assert_not_contains "$TMPROOT/.codex/hooks.json" ".claude" "hooks.json has no .claude paths"
+assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" 'name = "backend-engineer"' "agent TOML includes name"
+assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" ".codex/rules/workflow-gates.md" "agent instructions rewrite rule paths"
+assert_not_contains "$TMPROOT/.codex/agents/backend-engineer.toml" "---" "agent TOML excludes YAML frontmatter"
+
+if bash "$SCRIPT" --root "$TMPROOT" --check >/tmp/_codex_adapter_check.out 2>&1; then
+  mark_pass "--check passes when generated output is current"
+else
+  mark_fail "--check passes when generated output is current" "$(cat /tmp/_codex_adapter_check.out)"
+fi
+
+printf '\nNew source line.\n' >> "$TMPROOT/.claude/skills/status/SKILL.md"
+if bash "$SCRIPT" --root "$TMPROOT" --check >/tmp/_codex_adapter_check_drift.out 2>&1; then
+  mark_fail "--check detects drift" "expected non-zero exit"
+else
+  mark_pass "--check detects drift"
+fi
+
+echo
+echo "===== test_sync_codex_adapter.sh ====="
+echo "Passed: $PASS"
+echo "Failed: $FAIL"
+if [ "$FAIL" -gt 0 ]; then
+  echo "Failed cases:$FAILED"
+  exit 1
+fi
+exit 0
--- a/.claude/hooks/tests/test_sync_codex_adapter.sh
+++ b/.claude/hooks/tests/test_sync_codex_adapter.sh
@@ -25,6 +25,11 @@ assert_file() {
   [ -f "$path" ] && mark_pass "$label" || mark_fail "$label" "missing $path"
 }
 
+assert_no_dir() {
+  local path="$1" label="$2"
+  [ ! -d "$path" ] && mark_pass "$label" || mark_fail "$label" "unexpected directory $path"
+}
+
 assert_contains() {
   local path="$1" pattern="$2" label="$3"
   grep -F "$pattern" "$path" >/dev/null 2>&1 && mark_pass "$label" || mark_fail "$label" "missing pattern [$pattern] in $path"
@@ -42,7 +47,8 @@ assert_not_contains() {
 TMPROOT=$(mktemp -d "${TMPDIR:-/tmp}/codex-adapter-test.XXXXXX")
 trap 'rm -rf "$TMPROOT"' EXIT
 
-mkdir -p "$TMPROOT/.claude/skills/status" "$TMPROOT/.claude/hooks" "$TMPROOT/.claude/agents" "$TMPROOT/.claude/registries"
+mkdir -p "$TMPROOT/.claude/skills/status" "$TMPROOT/.claude/hooks" "$TMPROOT/.claude/agents"
+touch "$TMPROOT/.apexyard-fork"
 
 cat > "$TMPROOT/.claude/skills/status/SKILL.md" <<'MD'
 ---
@@ -53,20 +59,27 @@ description: Current status.
 Source `.claude/hooks/_lib-portfolio-paths.sh` and read `.claude/skills/status/briefing.sh`.
 MD
 
-cat > "$TMPROOT/.claude/hooks/detect-role-trigger.sh" <<'SH'
+cat > "$TMPROOT/.claude/hooks/block-test.sh" <<'SH'
 #!/usr/bin/env bash
-echo "from .claude/hooks"
+input=$(cat)
+case "$input" in
+  *blocked*) exit 2 ;;
+  *) exit 0 ;;
+esac
 SH
+chmod +x "$TMPROOT/.claude/hooks/block-test.sh"
 
 cat > "$TMPROOT/.claude/settings.json" <<'JSON'
 {
   "hooks": {
-    "SessionStart": [
+    "PreToolUse": [
       {
+        "matcher": "Bash",
         "hooks": [
           {
             "type": "command",
-            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/detect-role-trigger.sh\"'"
+            "if": "Bash(blocked *)",
+            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/block-test.sh\"'"
           }
         ]
       }
@@ -88,10 +101,6 @@ allowed-tools: Bash, Read
 Read `.claude/rules/workflow-gates.md` before acting in Claude Code.
 MD
 
-cat > "$TMPROOT/.claude/registries/ai-crawlers.json" <<'JSON'
-[]
-JSON
-
 echo "== Codex adapter sync smoke"
 
 if bash "$SCRIPT" --root "$TMPROOT" >/tmp/_codex_adapter_sync.out 2>&1; then
@@ -101,21 +110,37 @@ else
 fi
 
 assert_file "$TMPROOT/.agents/skills/status/SKILL.md" "skill mirror exists"
-assert_file "$TMPROOT/.codex/hooks/detect-role-trigger.sh" "hook mirror exists"
 assert_file "$TMPROOT/.codex/hooks.json" "hooks.json mirror exists"
 assert_file "$TMPROOT/.codex/agents/backend-engineer.toml" "agent TOML exists"
-assert_file "$TMPROOT/.codex/registries/ai-crawlers.json" "registry mirror exists"
+assert_no_dir "$TMPROOT/.codex/hooks" "adapter does not copy hook scripts"
 
-assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".codex/hooks/_lib-portfolio-paths.sh" "skill rewrites hook paths"
 assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".agents/skills/status/briefing.sh" "skill rewrites skill paths"
-assert_contains "$TMPROOT/.codex/hooks.json" '$r/.codex/hooks/detect-role-trigger.sh' "hooks.json rewrites hook exec path"
-assert_contains "$TMPROOT/.codex/hooks.json" '$HOME/.codex/apexyard' "hooks.json rewrites pin cache path"
+assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".claude/hooks/_lib-portfolio-paths.sh" "skill preserves canonical hook paths"
+assert_contains "$TMPROOT/.codex/hooks.json" '$r/.claude/hooks/block-test.sh' "hooks.json delegates to canonical hook path"
+assert_contains "$TMPROOT/.codex/hooks.json" '$HOME/.claude/apexyard' "hooks.json preserves Claude session pin path"
 assert_not_contains "$TMPROOT/.codex/hooks.json" "$TMPROOT" "hooks.json has no absolute fixture path"
-assert_not_contains "$TMPROOT/.codex/hooks.json" ".claude" "hooks.json has no .claude paths"
+assert_not_contains "$TMPROOT/.codex/hooks.json" '$r/.codex/hooks' "hooks.json does not point at copied hooks"
 assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" 'name = "backend-engineer"' "agent TOML includes name"
 assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" 'model = "gpt-5.4"' "agent TOML maps Claude model to Codex model"
-assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" ".codex/rules/workflow-gates.md" "agent instructions rewrite rule paths"
+assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" ".claude/rules/workflow-gates.md" "agent instructions preserve rule paths"
 assert_not_contains "$TMPROOT/.codex/agents/backend-engineer.toml" "---" "agent TOML excludes YAML frontmatter"
+
+hook_command=$(jq -r '.hooks.PreToolUse[0].hooks[0].command' "$TMPROOT/.codex/hooks.json")
+printf '{"tool_name":"Bash","tool_input":{"command":"blocked command"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_block.out 2>&1
+block_rc=$?
+if [ "$block_rc" -eq 2 ]; then
+  mark_pass "generated hook command preserves blocking exit"
+else
+  mark_fail "generated hook command preserves blocking exit" "expected exit 2, got $block_rc: $(cat /tmp/_codex_adapter_block.out)"
+fi
+
+printf '{"tool_name":"Bash","tool_input":{"command":"allowed command"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_allow.out 2>&1
+allow_rc=$?
+if [ "$allow_rc" -eq 0 ]; then
+  mark_pass "generated hook command preserves allowing exit"
+else
+  mark_fail "generated hook command preserves allowing exit" "expected exit 0, got $allow_rc: $(cat /tmp/_codex_adapter_allow.out)"
+fi
 
 if bash "$SCRIPT" --root "$TMPROOT" --check >/tmp/_codex_adapter_check.out 2>&1; then
   mark_pass "--check passes when generated output is current"
--- a/.claude/hooks/tests/test_sync_codex_adapter.sh
+++ b/.claude/hooks/tests/test_sync_codex_adapter.sh
@@ -101,6 +101,18 @@ allowed-tools: Bash, Read
 Read `.claude/rules/workflow-gates.md` before acting in Claude Code.
 MD
 
+# Shared model matrix the generator reads to translate tier labels -> Codex
+# models (source of truth for every harness adapter). Its presence is what
+# makes the `model = "gpt-5.4"` assertion below config-driven rather than
+# hardcoded — omit it and the mapping correctly falls back to the label.
+cat > "$TMPROOT/.claude/harness-models.json" <<'JSON'
+{
+  "opus":   { "codex": "gpt-5.5" },
+  "sonnet": { "codex": "gpt-5.4" },
+  "haiku":  { "codex": "gpt-5.4-mini" }
+}
+JSON
+
 echo "== Codex adapter sync smoke"
 
 if bash "$SCRIPT" --root "$TMPROOT" >/tmp/_codex_adapter_sync.out 2>&1; then
--- a/.claude/hooks/tests/test_sync_codex_adapter.sh
+++ b/.claude/hooks/tests/test_sync_codex_adapter.sh
@@ -63,7 +63,7 @@ cat > "$TMPROOT/.claude/hooks/block-test.sh" <<'SH'
 #!/usr/bin/env bash
 input=$(cat)
 case "$input" in
-  *blocked*) exit 2 ;;
+  *deny*) exit 2 ;;
   *) exit 0 ;;
 esac
 SH
@@ -78,7 +78,7 @@ cat > "$TMPROOT/.claude/settings.json" <<'JSON'
         "hooks": [
           {
             "type": "command",
-            "if": "Bash(blocked *)",
+            "if": "Bash(policy *)",
             "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/block-test.sh\"'"
           }
         ]
@@ -118,6 +118,7 @@ assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".agents/skills/status
 assert_contains "$TMPROOT/.agents/skills/status/SKILL.md" ".claude/hooks/_lib-portfolio-paths.sh" "skill preserves canonical hook paths"
 assert_contains "$TMPROOT/.codex/hooks.json" '$r/.claude/hooks/block-test.sh' "hooks.json delegates to canonical hook path"
 assert_contains "$TMPROOT/.codex/hooks.json" '$HOME/.claude/apexyard' "hooks.json preserves Claude session pin path"
+assert_contains "$TMPROOT/.codex/hooks.json" 'APEXYARD_CODEX_HOOK_GLOB=' "hooks.json wraps command predicates"
 assert_not_contains "$TMPROOT/.codex/hooks.json" "$TMPROOT" "hooks.json has no absolute fixture path"
 assert_not_contains "$TMPROOT/.codex/hooks.json" '$r/.codex/hooks' "hooks.json does not point at copied hooks"
 assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" 'name = "backend-engineer"' "agent TOML includes name"
@@ -125,8 +126,14 @@ assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" 'model = "gpt-5.4
 assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" ".claude/rules/workflow-gates.md" "agent instructions preserve rule paths"
 assert_not_contains "$TMPROOT/.codex/agents/backend-engineer.toml" "---" "agent TOML excludes YAML frontmatter"
 
+if jq -e '.. | objects | select(has("if"))' "$TMPROOT/.codex/hooks.json" >/dev/null; then
+  mark_fail "hooks.json omits unsupported if fields" "found handler-level if metadata"
+else
+  mark_pass "hooks.json omits unsupported if fields"
+fi
+
 hook_command=$(jq -r '.hooks.PreToolUse[0].hooks[0].command' "$TMPROOT/.codex/hooks.json")
-printf '{"tool_name":"Bash","tool_input":{"command":"blocked command"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_block.out 2>&1
+printf '{"tool_name":"Bash","tool_input":{"command":"policy deny"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_block.out 2>&1
 block_rc=$?
 if [ "$block_rc" -eq 2 ]; then
   mark_pass "generated hook command preserves blocking exit"
@@ -134,12 +141,20 @@ else
   mark_fail "generated hook command preserves blocking exit" "expected exit 2, got $block_rc: $(cat /tmp/_codex_adapter_block.out)"
 fi
 
-printf '{"tool_name":"Bash","tool_input":{"command":"allowed command"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_allow.out 2>&1
+printf '{"tool_name":"Bash","tool_input":{"command":"policy allow"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_allow.out 2>&1
 allow_rc=$?
 if [ "$allow_rc" -eq 0 ]; then
   mark_pass "generated hook command preserves allowing exit"
 else
   mark_fail "generated hook command preserves allowing exit" "expected exit 0, got $allow_rc: $(cat /tmp/_codex_adapter_allow.out)"
+fi
+
+printf '{"tool_name":"Bash","tool_input":{"command":"other deny"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_skip.out 2>&1
+skip_rc=$?
+if [ "$skip_rc" -eq 0 ]; then
+  mark_pass "generated hook command skips nonmatching predicates"
+else
+  mark_fail "generated hook command skips nonmatching predicates" "expected exit 0, got $skip_rc: $(cat /tmp/_codex_adapter_skip.out)"
 fi
 
 if bash "$SCRIPT" --root "$TMPROOT" --check >/tmp/_codex_adapter_check.out 2>&1; then
--- a/.claude/hooks/tests/test_sync_codex_adapter.sh
+++ b/.claude/hooks/tests/test_sync_codex_adapter.sh
@@ -113,6 +113,7 @@ assert_contains "$TMPROOT/.codex/hooks.json" '$HOME/.codex/apexyard' "hooks.json
 assert_not_contains "$TMPROOT/.codex/hooks.json" "$TMPROOT" "hooks.json has no absolute fixture path"
 assert_not_contains "$TMPROOT/.codex/hooks.json" ".claude" "hooks.json has no .claude paths"
 assert_contains "$TMPROOT/.codex/agents/backend-engineer.toml" 'name = "backend-engineer"' "agent TOML includes name"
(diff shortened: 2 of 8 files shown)


hossam-96 (commented):
## Code Review: PR #730

**Commit**: `185245fa0113e75a58c516ef32377482d641881e`

### Summary
This PR adds `bin/sync-codex-adapter.sh`, documentation, an AgDR, and smoke coverage for generating Codex-facing `.agents/` / `.codex/` adapter output from the canonical `.claude/` runtime.

### Checklist Results
- ✅ Code Quality: Pass
- ✅ Testing: Pass
- ✅ Security: Pass
- ✅ Performance: Pass
- ✅ Summary Bullet Narrative: Pass
- ✅ Technical Decisions (AgDR): Pass
- ✅ Adopter Handbooks: Pass

### Issues Found
None.

### Suggestions
suggestion: PR #730 targets `dev`, so the diff also carries README commits already merged to `main` for #714/#720. I did not treat this as blocking because those changes are already merged upstream and the PR body only closes #729, but future PRs should keep branch-sync/history cleanup separate from feature work when possible.

### Verification
- `bash -n bin/sync-codex-adapter.sh .claude/hooks/tests/test_sync_codex_adapter.sh`
- `bash .claude/hooks/tests/test_sync_codex_adapter.sh` — PASS=18 FAIL=0
- `bin/sync-codex-adapter.sh --check` — pass
- `bin/run-hook-tests.sh` — PASS=74 FAIL=0 SKIP=0 TOTAL=74
- Confirmed `.claude/agents` uses only `opus`, `sonnet`, and `haiku`, and generated `.codex/agents/*.toml` maps them to `gpt-5.5`, `gpt-5.4`, and `gpt-5.4-mini` respectively.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `185245fa0113e75a58c516ef32377482d641881e`

me2resh (commented):
## Security Review: PR #730 — Codex adapter generator

**Commit**: `185245fa0113e75a58c516ef32377482d641881e`

### Summary

A generator (`bin/sync-codex-adapter.sh`) that mirrors the canonical `.claude/` runtime into a Codex-facing `.agents/` + `.codex/` tree: skills → `.agents/skills`, agents → `.codex/agents/*.toml`, hooks → `.codex/hooks`, rules → `.codex/rules`, and `settings.json` → `.codex/hooks.json`, rewriting path/label tokens along the way. Because a harness adapter is the layer that makes ApexYard's gates actually fire under a runtime, I reviewed it at the same bar as the pi.dev adapter.

Good news first, because it bounds the blast radius: **the live Claude trust chain is untouched.** The diff only *adds* `bin/`, `docs/`, and a test — no change to `.claude/settings.json` or any `.claude/hooks/*.sh`. Existing Claude Code governance is not weakened by merging this. The risk is entirely inside the new, generated Codex surface.

### Checklist Results
- Injection Prevention:   Pass (see LOW note — generate-time posture is sound)
- XSS Prevention:         N/A
- Auth & Authorisation:   N/A
- **Gate-enforcement faithfulness:  FAIL (blocking)** — the crown-jewel check

### Security Issues Found

#### HIGH (blocking) — No demonstrated enforcement path under Codex; the adapter reads as governed but firing is unproven
`.codex/hooks.json` is a **renamed copy of Claude Code's `settings.json`** with `.claude`→`.codex` substituted. Claude Code's schema (`hooks.PreToolUse[].matcher` + `command`, block on exit 2) is a Claude-Code-specific contract. Nothing in this PR demonstrates that Codex reads `.codex/hooks.json`, understands that schema, executes the commands *before* a tool call, and honours a non-zero exit as a **block**. The smoke test only asserts file existence, path-string rewrites, and model mapping — it never asserts that a single gate actually blocks. So a user running under Codex sees a `.codex/hooks/` full of merge-gate / secrets / ticket-first scripts and a `hooks.json` and reasonably concludes they are governed, when the enforcement may be completely inert. That "looks governed, isn't" state is exactly the silent-bypass failure mode a harness adapter must not ship with. **The burden is on the adapter to prove at least one gate fires (exit-2 blocks a tool call) under Codex.** Right now it can't be shown from anything in the PR.

#### HIGH (blocking) — Blind regex rewrite mutates executable gate logic, not just paths
`rewrite_codex_paths` runs `perl -0pi -e 's/\.claude/.codex/g; s/Claude Code/Codex/g; s/Claude/Codex/g;'` **in place over every generated file, including the copied hook `.sh` scripts** (`copy_tree "$CLAUDE_DIR/hooks"` then `find … -type f | rewrite`). This edits the *body* of gate code with a text substitution that has no idea whether a given `.claude` / `Claude` token is a path, a comment, or a load-bearing string literal. Concrete regression I verified:

- `detect-role-trigger.sh` (the #777 trust-chain trigger) matches literals `.claude/hooks/*` and `.claude/settings.json` (lines 197-198) to fire the Security Auditor on trust-chain edits. After rewrite those become `.codex/hooks/*` / `.codex/hooks.json`. But `.claude/` **remains the kept source of truth** in the same checkout — so a Codex user editing the real `.claude/hooks/**` gets **no** security-auditor trigger from the rewritten hook, and the original `.claude` hook isn't wired into Codex. The trust-chain trigger silently goes dark. Any hook whose behaviour depends on a `.claude`/`Claude` string literal is subject to the same class of silent semantic drift.

The safe shape is to rewrite only when you can prove the token is a path (or, better, have hooks resolve their runtime dir at execution time and stop text-mutating hook bodies at all) — not a blanket `s/Claude/Codex/g` over executable code.

#### MEDIUM — Merge-gate marker path relocation is untested for cross-file consistency
`block-unreviewed-merge.sh` reads markers from `.claude/session/reviews/<pr>-{rex,ceo}.approved`; the rewrite moves the hook's view to `.codex/session/reviews/`. The two-review gate only holds if the *writers* — `_lib-review-markers.sh` and the `/approve-merge` skill — get rewritten to the **identical** path. If they diverge, the gate is either unsatisfiable (annoying) or trivially satisfiable / no-op (dangerous). No test covers marker-path round-tripping through the rewrite.

#### MEDIUM — Generated gate tree is unversioned and unauditable by policy
`docs/codex-adapter.md` tells adopters to `.git/info/exclude` the `.agents/` + `.codex/` trees, and `--check` CI is explicitly deferred. That means the actually-executing Codex gate scripts live only in local, uncommitted, un-reviewed files. A tampered local `.codex/hooks/block-unreviewed-merge.sh` would never surface in review or CI. For the *governance* tree specifically, "regenerate locally, don't track" removes the audit trail that makes the gates trustworthy.

#### LOW / informational
- **Injection posture is sound.** Generate-time uses `jq -Rs .` to encode agent name/description/body into TOML strings (JSON-string escaping is a valid superset for TOML basic strings here), sed-escapes `$ROOT` before the `grep -R` guard, sources filenames from `find` (not user input), and has no `eval` / no untrusted expansion into a shell. No template- or command-injection vector found at generate- or emit-time. One nit: `jq` JSON escaping isn't byte-identical to TOML, but for the frontmatter fields in scope it's safe.
- **Source of truth untouched** (repeated because it matters): `.claude/settings.json` and `.claude/hooks/**` are not modified — the live Claude trust chain is intact.

### Recommendations
1. Add a **conformance test that proves enforcement**: a Codex-runtime harness where a hook `exit 2` actually blocks a simulated tool call (merge gate, secrets, or ticket-first). Until one gate is shown to fire, the docs/README should state plainly that **gates are not yet enforced under Codex** so no one mistakes generated files for governance.
2. Stop blind-rewriting hook **bodies**. Either rewrite only proven path tokens, or (better) have hooks resolve their runtime root at exec time so the same script works under either surface without text mutation. Add a regression test that the #777 trust-chain trigger still fires post-adapter.
3. Add a round-trip test that the merge-gate marker path is identical between the rewritten hook and the rewritten writer (`_lib-review-markers.sh` / `/approve-merge`).
4. If the Codex tree is to carry gates, plan to **track it and enforce `--check` in CI** — an un-versioned enforcement tree defeats the audit trail.

### Verdict
**CHANGES REQUESTED** — two blocking issues: (1) no demonstrated exit-2 enforcement path under Codex (the crown-jewel gate-faithfulness check fails), and (2) a blind `s/Claude/.../` rewrite that mutates executable gate logic and silently drops the #777 trust-chain trigger. Blast radius is contained (live Claude chain untouched, output opt-in/uncommitted), which is why this is HIGH-blocking rather than a live CRITICAL — but a harness adapter cannot ship claiming governance it hasn't shown it enforces.

Not merging. No approval marker written.

---
📌 Reviewed commit: `185245fa0113e75a58c516ef32377482d641881e`

me2resh (changes requested):
## Code Review: PR #730 — `feat(#729): add Codex adapter generator`

**Commit**: `185245fa0113e75a58c516ef32377482d641881e`

Thanks for this, @hossam-96 — the PR hygiene is genuinely good (narrative Summary bullets, a real Glossary, an AgDR, a Testing section, and a smoke test). The problem is not presentation; it's that the generator, as written, produces a Codex adapter that **does not faithfully enforce apexyard's gates** — and nothing in the tests would catch that. For a governance framework the gate-enforcement bar is the whole point, so I'm requesting changes. Details below.

### Summary

Adds `bin/sync-codex-adapter.sh`, which copies `.claude/{skills,agents,hooks,rules,migrations,registries,settings.json}` into `.agents/`/`.codex/`, string-rewrites `.claude` → `.codex` (and `Claude` → `Codex`) across every generated file, converts agent `.md` → `.codex/agents/*.toml` with a Claude→Codex model-label map, and offers `--check` (drift) / `--clean`. Plus README restyling and AgDR-0080.

### Checklist Results
- Architecture & Design:      **Fail** — mutates copies of the gate hooks instead of shelling out to the single source of truth; no real Codex wiring
- Code Quality:               Advisory — blanket substitutions are too broad (see #6)
- Testing:                    **Fail** — tests assert file existence + string rewrites, never that a gate still enforces
- Security:                   Pass (no secrets; but see enforcement concern — a non-enforcing governance adapter is itself a safety regression)
- PR Description & Glossary:  Pass — Glossary present, bullets narrative, Testing + AgDR sections present
- Technical Decisions (AgDR): Pass (present) / but AgDR mis-frames the central risk (see #7)
- Adopter Handbooks:          N/A — no handbook triggers (migration-safety is blocking but no DB/schema migration is touched)
- Fallow Static Analysis:     N/A — no JS/TS in diff

### Issues Found

**⛔ 1 (blocking) — the blanket `.claude → .codex` rewrite silently breaks the merge/review gates.**
`rewrite_codex_paths` runs `s/\.claude/.codex/g` over every copied hook. The gate hooks read their approval markers from `.claude/session/reviews/` — verified in `block-unreviewed-merge.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh`, and the active-ticket marker in `require-active-ticket.sh`. After rewrite, the copies under `.codex/hooks/` read `.codex/session/reviews/...`. But the marker **producers** are unchanged: the Rex `code-reviewer` agent and `/approve-merge` skill both write to `.claude/session/reviews/` (`.claude/agents/code-reviewer.md:24`, `.claude/skills/approve-merge/SKILL.md:147`). So every rewritten gate now looks in a directory nothing populates — it can only always-block or, for any hook that fails-open when it can't resolve state, silently stop enforcing. The same applies to the pin path: the test even asserts `$HOME/.claude/apexyard` → `$HOME/.codex/apexyard`, but the ops-root pin is written to `~/.claude/apexyard` by Claude Code's SessionStart; nothing writes `~/.codex/apexyard`, so pin resolution is dead on the Codex side.

**⛔ 2 (blocking) — nothing actually wires these hooks into Codex; `.codex/hooks.json` is a renamed Claude `settings.json`.**
`settings.json` is Claude Code's `PreToolUse`/`PostToolUse` matcher schema. Renaming it to `hooks.json` and swapping paths does not cause Codex to fire the hooks. There is no evidence — in the PR, the AgDR, or `docs/codex-adapter.md` — that the Codex CLI consumes a `.codex/hooks.json` in this schema, and there is no adapter code bridging a Codex tool-call event to each hook's stdin/exit-code (`exit 2 → block`) contract. Compare the established harness-adapter pattern, the pi gate adapter (#815 / AgDR-0082): it registers a **real** `tool_call` handler using real pi types (`@earendil-works/pi-coding-agent`) and shells out to the **unmodified** bash hooks. As written, the generated Codex gates most likely do nothing at all under Codex — the exact "silently doesn't enforce a gate" failure mode this review is meant to catch. Please demonstrate (with a runnable path, not prose) that Codex invokes a generated hook with the correct stdin JSON and honours exit-2.

**⛔ 3 (blocking) — diverges from the framework's adapter principle: it forks the gate logic instead of shelling out to it.**
The pi adapter's central promise is "existing, **unmodified** bash hooks … zero logic duplication … the bash hooks stay the single source of truth." This PR creates a second, string-mutated copy of every hook under `.codex/hooks/`. That's worse than duplication — it's a divergent fork of the approval logic that will drift from `.claude/hooks/` and can be corrupted by the blunt rewrite (below). Recommend adopting the #815 shape: a thin transport that invokes the real `.claude/hooks/*.sh` **unchanged**, leaving marker/session/pin paths intact so the same markers Rex and `/approve-merge` write are the ones the gate reads.

**⛔ 4 (blocking) — the tests never verify a single gate still enforces.**
`test_sync_codex_adapter.sh` checks files exist, paths were rewritten, no absolute fixture path leaked, no `.claude` remains, model mapping landed, and `--check` detects drift. None of it runs a generated gate. Because the entire risk is "does the adapter still enforce," add a behavioural test: feed `.codex/hooks/block-unreviewed-merge.sh` an unreviewed `gh pr merge` on stdin and assert **exit 2**; then place a matching marker at whatever path the generated hook now reads and assert **exit 0**. If that test can't be made to pass without also fixing #1/#2, that's the proof this needs.

**5 — Scope creep + stale base.** The README header restyle and contributors table are unrelated to "Codex adapter generator," and `upstream/dev` **already contains** that header restyle (`apexyard-avatar-512.png`, width 88) — so the PR is behind `dev` and currently `CONFLICTING` / `DIRTY`. Rebase on current `dev`, drop the already-merged README churn, and keep only the new `## Codex Adapter` section.

**6 — The substitutions are too broad and semantically corrupt content.** `s/Claude/Codex/g` and `s/\.claude/.codex/g` rewrite comments, error strings, and factual references ("Claude Code's session pin" → "Codex's session pin", which is now false), can mangle any URL containing `.claude`, and leave lowercase `claude` CLI invocations (`claude --from-pr 123`) untouched — inconsistent. Prefer targeted, anchored rewrites of known path prefixes over global text substitution — and don't rewrite the hooks' runtime state paths at all (see #1).

**7 — AgDR-0080 mis-frames the central risk.** It argues "the main risk is drift, not lack of files." The main risk is actually enforcement: whether Codex fires the hooks, and whether rewriting marker paths breaks the gates. Please reckon with the enforcement mechanism and the unmodified-hooks-vs-rewritten-copy trade-off, and reference the existing pi adapter (#815 / AgDR-0082) so the two harness adapters stay consistent.

### Verdict
**CHANGES REQUESTED**

The direction (a generator that keeps `.claude/` as source of truth) is reasonable, and the AgDR/tests/docs discipline is appreciated. But an adapter for a governance framework has to *provably* enforce the gates, and this one rewrites the gate hooks into a path space nothing populates, with no wiring to make Codex run them and no test that any gate still blocks. Please: (a) follow the #815 pattern of invoking the **unmodified** hooks, (b) show the real Codex mechanism that fires them, (c) leave `.claude/session` marker/pin paths intact, and (d) add a behavioural enforce/allow test per gate. Happy to re-review once the enforcement path is demonstrable.

---
📌 Reviewed commit: `185245fa0113e75a58c516ef32377482d641881e`

hossam-96 (commented):
## Code Review: PR #730

**Commit**: `20ae3f98289508c649523f75292465371af6cb78`

### Summary
This PR adds a Codex adapter generator for issue #729: it emits `.agents/skills`, `.codex/agents/*.toml`, and `.codex/hooks.json` from the canonical `.claude/` runtime, documents the adapter, and records AgDR-0080 for the hook-delegation decision.

### Checklist Results
- ❌ Code Quality: Fail
- ❌ Testing: Fail
- ✅ Security: Pass
- ✅ Performance: Pass
- ✅ Summary Bullet Narrative: Pass
- ❌ Technical Decisions (AgDR): Fail
- ⚠ Adopter Handbooks: Advisory
- ✅ Fallow Static Analysis (JS/TS): N/A

### Issues Found
1. `bin/sync-codex-adapter.sh:93` copies `.claude/settings.json` directly into `.codex/hooks.json`, but 46 of the current hook handlers rely on Claude-specific `if` predicates such as `Bash(git add *)` / `Bash(gh pr merge *)`. The current Codex hooks docs describe filtering through event matcher groups, with command handler fields like `type`, `command`, `timeout`, and `statusMessage`; they do not define a handler-level `if` predicate. If Codex ignores that unknown field, every `PreToolUse` Bash handler in the generated file becomes eligible for every Bash tool call. That would run merge gates, commit gates, PR-create gates, etc. on unrelated Bash commands, which means the adapter is not enforcement-faithful. If Codex rejects unknown fields instead, the generated hooks fail to load. Either outcome breaks the intended adapter shape. Please translate Claude `if` predicates into Codex-supported matching/guarding, or wrap each generated command with a small predicate that reads Codex stdin (`tool_input.command`) and exits 0 when the command does not match. Source checked: https://developers.openai.com/codex/hooks.

2. `.claude/hooks/tests/test_sync_codex_adapter.sh:128`-`.claude/hooks/tests/test_sync_codex_adapter.sh:143` does not catch the issue above because it extracts the generated command with `jq` and runs that command directly. That bypasses the generated hook entry’s `matcher` and copied `if` field entirely. Please add a behavioral test that proves a non-matching Bash command does not run a hook whose Claude-side `if` predicate should exclude it, and that a matching command still preserves block/allow exit codes.

3. `docs/agdr/AgDR-0080-codex-adapter-generation.md:1` introduces a second `AgDR-0080`; `docs/agdr/AgDR-0080-board-automation-attach-points.md:1` already uses that ID on this branch. The PR body links `docs/agdr/AgDR-0080-codex-adapter-generation.md`, so the decision is documented, but the duplicate ID makes AgDR references ambiguous. Please renumber this new Codex adapter decision to the next free ID and update README/docs/PR links accordingly.

### Handbook Findings
⚠ **Commit Message Quality** — `handbooks/general/commit-message-quality.md` (discovery: path-convention)
- The three commits all have useful ticket/AgDR references, but their bodies are reference-only rather than explaining the reasoning. Advisory only; the PR body and AgDR carry most of the missing rationale, so this does not affect the verdict.

### Suggestions
- The generated subagent TOML fields match Codex’s documented required fields (`name`, `description`, `developer_instructions`), and the model mapping is covered by the smoke test.
- The docs correctly call out remaining live Codex runtime/trust risk; keep that caveat after fixing the hook predicate translation.

### Verification
- `bash -n bin/sync-codex-adapter.sh .claude/hooks/tests/test_sync_codex_adapter.sh` passed.
- `bash .claude/hooks/tests/test_sync_codex_adapter.sh` passed: PASS=19 FAIL=0.
- `bin/sync-codex-adapter.sh --check` passed.

### Verdict
**CHANGES REQUESTED**

---
Reviewed commit: `20ae3f98289508c649523f75292465371af6cb78`

hossam-96 (commented):
## Code Review: PR #730

**Commit**: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

### Summary
This PR adds a Bash + jq Codex adapter generator that mirrors `.claude/skills` into `.agents/skills`, converts `.claude/agents/*.md` into Codex TOML agents with model mapping, and generates `.codex/hooks.json` while delegating gate decisions to canonical `.claude/hooks/*.sh` scripts. The follow-up fixes the earlier Codex hook schema blocker by compiling Claude Code handler-level `if` predicates into shell-side command filters and omitting unsupported `if` fields from generated hook JSON.

Note: GitHub reported the reviewed head as `729d7539ab4a31b621bccf4d10eea641f21b64f7`; this differs from the `729d753d77cc8e6fc3752b066563ec2333996f67` SHA in the review request.

### Checklist Results
- ✅ Code Quality: Pass
- ✅ Testing: Pass, with the known upstream hook-suite failure noted below
- ✅ Security: Pass
- ✅ Performance: Pass
- ✅ Technical Decisions (AgDR): Pass — `docs/agdr/AgDR-0085-codex-adapter-generation.md` covers the adapter/gate-delegation decision and no stale new-doc `AgDR-0080` references remain
- ✅ Adopter Handbooks: Pass

### Issues Found
None.

### Suggestions
None blocking. Merge readiness still depends on the normal CI/merge gates; locally, `bin/run-hook-tests.sh` still reports the pre-existing `.claude/hooks/tests/test_block_merge_on_red_ci.sh` failure (`PASS=92 FAIL=1 TOTAL=93`), matching the author’s note.

### Verification
- `bash -n bin/sync-codex-adapter.sh .claude/hooks/tests/test_sync_codex_adapter.sh`
- `bash .claude/hooks/tests/test_sync_codex_adapter.sh` → `PASS=22 FAIL=0`
- `bin/sync-codex-adapter.sh --check`
- `jq -e '.. | objects | select(has("if"))' .codex/hooks.json` → no matches
- `git diff --check upstream/dev...HEAD`
- `bin/run-hook-tests.sh` → `PASS=92 FAIL=1 SKIP=0 TOTAL=93`; only `.claude/hooks/tests/test_block_merge_on_red_ci.sh` failed
- `gh pr checks 730 --repo me2resh/apexyard` → no checks reported for the branch at review time

### Verdict
**APPROVED**

---
📌 Reviewed commit: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

me2resh (commented):
## Security Review: PR #730 — Codex adapter generator

**Commit**: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

▸ Reviewed by Hakim (Security Auditor) — trust-chain work (hook wiring for a second harness).

### Summary

This PR adds `bin/sync-codex-adapter.sh`, which emits Codex-facing adapter files (`.codex/hooks.json`, `.codex/agents/*.toml`, `.agents/skills`) from the canonical `.claude/` runtime. The central question for a governance adapter: **does the generated Codex hook wiring faithfully preserve every `.claude/hooks` gate, or can it silently weaken one for Codex users?** I traced the delegation path, ran the smoke test, and adversarially tested exit-code, predicate, injection, and subversion behaviour. **Enforcement fidelity holds.**

### Checklist Results
- Secrets & Credentials:  **Pass** — no secrets; generator guards against absolute-clone-path leakage (lines 155–158) and I confirmed 0 path leaks in real output
- Injection Prevention:   **Pass** — untrusted `tool_input.command` is only pattern-matched (quoted LHS) and piped as stdin to the trusted hook, never `eval`'d
- Exit-code fidelity:     **Pass** — a blocking `exit 2` propagates unchanged through the Codex delegation wrapper
- Matcher/predicate fidelity: **Pass** — 46 `if`-predicates → 46 wrapped commands, exact tool+glob, no widen/narrow vs Claude
- No enforcement fork:    **Pass** — 0 hook bodies copied; all 69 delegations exec the unmodified `$r/.claude/hooks/*.sh`
- Delegation subversion:  **Pass** — a crafted command/skill name cannot run arbitrary code

### What I verified

**1. Exit-code fidelity (the worst-case bug — a block silently downgraded to allow).**
Traced `.codex/hooks.json` → wrapper → `.claude/hooks/*.sh`. The wrapper's final line is `printf "%s" "$input" | bash -c "$APEXYARD_CODEX_ORIGINAL_HOOK"`, so the wrapper's exit code IS the original hook's exit code. Empirically, against the real merge-gate delegation:
- matching `gh pr merge …` → gate fires, **exit 2 (BLOCK preserved)**
- non-merge Bash / non-Bash tool / empty command → **exit 0 (skip)**

The smoke test's own `block/allow exit` assertions pass (PASS=22 FAIL=0). *(Two transient failures I saw initially were an artifact of running inside a live Claude Code session whose ops-root pin redirected the fixture's root-resolver — they vanish with the session pin neutralized, exactly matching the author's reported 22/0.)*

**2. Matcher/predicate fidelity (a gate silently not firing).**
The `if: "Bash(gh pr merge *)"` → `tool=Bash` + `glob=gh pr merge *` conversion is exact. Count check: **46 source predicates → 46 wrapped commands, 69 total hooks → 69 total** (nothing dropped). No `if` field survives (Codex-unsupported). Both merge shapes closed in #47 are wrapped — `gh pr merge *` (×4) **and** `gh api *` (×5) — plus red-CI, design-review, and architecture-review gates all delegate. The predicate glob mirrors Claude's prefix-glob semantics faithfully (including inheriting Claude's own known blind spot to env-prefixed/compound commands — a pre-existing Claude gap, not introduced here).

**3. No enforcement fork.**
0 `.sh` files under `.codex`; all 69 delegations point at `$r/.claude/hooks/`, 0 at any `.codex/hooks/` copy. There is genuinely no second, divergeable copy of gate logic — `.codex/hooks.json` is a thin transport over the single audited runtime (the AgDR-0082 pi-adapter precedent). Generated output is not committed in this PR (0 tracked generated files); the documented tracking policy keeps it local via `.git/info/exclude`, with `--check` drift detection for anyone who opts to track it.

**4. Injection safety.**
Adversarial payloads in `tool_input.command` — `deploycmd $(touch CANARY)` and a `"; touch CANARY; …` quote/semicolon breakout — both failed to execute (canary never created), and the gate still blocked. `jq @sh` shell-quotes the trusted command/predicate at generation time; the runtime glob is used only as a `[[ ]]` pattern (RHS is never command-substituted), and `cmd`/`input` are always quoted.

### Findings (all non-blocking)
- **LOW / informational** — Actual Codex-side enforcement depends on Codex honouring `hooks.json` matcher + exit-2 blocking semantics, which is untestable in this repo's shell harness. This is correctly and explicitly disclosed in AgDR-0085 ("live enforcement still depends on Codex's hook runtime and trust settings"). Recommend a Codex-runtime conformance test before advertising Codex as a fully-enforced governance surface.
- **Informational** — For adopters who choose to *track* generated output as an enforced surface, lean on the doc's existing guidance to run `bin/sync-codex-adapter.sh --check` in CI so a stale committed `.codex/hooks.json` can't silently rot out of sync.

### Verdict
**APPROVED** — enforcement fidelity holds: no gate is weakened, narrowed, or bypassed on the Codex delegation path, and the adapter is injection-safe.

---
📌 Reviewed commit: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

me2resh (commented):
## Code Review: PR #730

**Commit**: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

### Summary
Adds `bin/sync-codex-adapter.sh`, a generator that emits Codex-facing adapter files (`.agents/skills`, `.codex/agents/*.toml`, `.codex/hooks.json`) from the canonical `.claude/` runtime, plus a 22-case smoke test, AgDR, docs, and a README line. The design is sound and the generator is faithful — the only blocker is an AgDR number collision.

### Checklist Results
- Architecture & Design:     Pass — `.claude/` stays source of truth; `.codex/hooks.json` delegates to the unmodified `.claude/hooks/*.sh` rather than forking gate logic. Matches the pi-adapter precedent (AgDR-0082).
- Code Quality:              Pass — `set -euo pipefail`, jq/`@sh`-quoted TOML strings, `--check`/`--clean` modes, `mktemp` + `trap` cleanup, an absolute-path guard.
- Testing:                   Pass — smoke test is meaningful (see below); full hook suite green in a clean env.
- Security:                  Pass — no secrets; generated output carries no absolute clone paths (guard at `sync-codex-adapter.sh:155` + a test assertion); trust-chain literals preserved, not rewritten.
- Performance:               Pass — N/A (build-time generator).
- PR Description & Glossary: Pass — narrative Summary bullets (what + why), thorough Glossary, `Closes #729`.
- Technical Decisions (AgDR):**CHANGES REQUESTED** — number collision (below).

### Verification I ran
- **if-predicate → shell-filter conversion preserves gate semantics (the dangerous-bug focus).** All 46 `if` predicates in the real `settings.json` are `Bash(...)`-shaped, so the wrapper's `.tool_input.command` extraction + `[[ cmd == $glob ]]` filter is correct for every current predicate. Exit codes flow through: block stays **exit 2**, allow stays **exit 0**, and a non-matching predicate is **skipped (exit 0)** before the canonical hook runs. I confirmed this on a real generation — e.g. `block-unreviewed-merge` is wrapped with `GLOB='gh pr merge *'` and delegates via `APEXYARD_CODEX_ORIGINAL_HOOK`.
- **Smoke test is real, not tautological.** 22 assertions exercising skill-path rewriting, model mapping (`sonnet`→`gpt-5.4`), canonical-hook delegation, no copied hooks, no `if` fields, no absolute paths, and live exit-code preservation (2 / 0 / 0) via synthetic stdin. Passes 22/22 in a clean env.
- **No absolute-path leak** in generated output (verified on a real generation) and generated `.codex/`/`.agents/` are not committed.
- **Model mapping** correct — 25 agents mapped; `opus`→`gpt-5.5` (code-reviewer), `sonnet`→`gpt-5.4`.
- **Full hook suite** on this branch (clean env): **PASS=93 FAIL=0**, including the new test and `test_block_merge_on_red_ci.sh`.

### Issues Found (required)

**⛔ AgDR number collision — required change.** `docs/agdr/AgDR-0085-structurizr-dsl-escape-hatch.md` already exists on `dev`; this PR adds a second `AgDR-0085-*`. Next free number is **0088** (dev holds 0080–0087). Please:
- rename the file → `docs/agdr/AgDR-0088-codex-adapter-generation.md`
- update the H1 title (`AgDR-0085 →` `AgDR-0088`) and the two doc references in `docs/codex-adapter.md` / README if they name the number
- (the commit-trailer `AgDR:` lines already drifted 0080→0085 across commits; point them at 0088 on the squash)

This is the only thing standing between the PR and approval.

- **Fail loud on an unexpected `if` shape.** In `wrap_if_handler`, if a future predicate is not `Tool(glob)`-shaped (a bare tool, or an `Edit(**/*.ts)` path-glob), jq `capture` yields empty and the handler is **silently dropped** from `hooks.json` — a gate would vanish with no error. Everything is `Bash(...)` today so output is correct, but for a generator whose thesis is enforcement-faithfulness, the safer failure direction is a non-zero exit. Related: the wrapper only reads `.tool_input.command`, so a non-Bash predicate would silently never fire even if it survived. A guard here future-proofs the "silently stops enforcing" risk the AgDR itself calls out.
- **Make the smoke test hermetic.** The fixture's hook command uses the real session-pin-first ops-root resolver, so running the test inside a live Claude Code session (with `CLAUDE_CODE_SESSION_ID` + a pin present) makes it escape the sandbox `$TMPROOT` and exec the real fork's path — 2 spurious failures. It passes cleanly in CI (no session id). An `env -u CLAUDE_CODE_SESSION_ID` (or pointing `APEXYARD_OPS_PIN_DIR` at `$TMPROOT`) around the block/allow cases makes it robust when run interactively.
- **Pre-existing-failure claim is greener than reported.** On a truly clean `upstream/dev` I get `test_block_merge_on_red_ci.sh` **passing 23/0** and the full suite 93/0; the `FAIL=1` you saw is environment-dependent (session-pin / `gh` auth), not a regression — and this PR touches none of that code, so it's not a blocker either way. Just flagging that it didn't reproduce as "fails on clean dev" here.

### Verdict
**CHANGES REQUESTED** — solely the AgDR-0085 → 0088 renumber. The generator, delegation model, exit-code preservation, and tests are all sound; re-request review after the rename and I'll approve.

---
📌 Reviewed commit: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

me2resh (changes requested):
## Design Review: PR #730 — Codex adapter generator

**Commit**: `729d7539ab4a31b621bccf4d10eea641f21b64f7`

Solution Architect (Tariq) review. This PR carries `docs/agdr/AgDR-0085-codex-adapter-generation.md`, so it is also the **Gate 3b architecture sign-off**.

### Summary
`bin/sync-codex-adapter.sh` generates Codex-facing adapter files (`.codex/hooks.json`, `.codex/agents/*.toml`, `.agents/skills/`) from the canonical `.claude/` runtime. The generated `hooks.json` delegates every gate decision to the **unmodified** `.claude/hooks/*.sh` scripts rather than copying/forking gate logic. Backed by an AgDR, a docs page, a 22-case smoke test, and a `--check` drift mode. I verified the Codex-specific claims against Codex's real hooks/subagents/config docs.

### Review Lens Results
- Quality attributes / NFRs:    Concern (per-tool-call subprocess nesting; drift model)
- Design patterns / structure:  Pass (preserves the adapter-over-bash invariant)
- Technical debt:               Concern (tracking-policy vs `--check` tension)
- Decisions (AgDR linkage):     Fail — AgDR ID collision (BLOCKING)
- Risk:                         Concern (no live-Codex enforcement proof)
- Trade-off analysis:           Pass (genuine options table)
- Requirements traceability:    Pass (closes #729)
- Migration safety:             N/A
- Adopter Handbooks:            N/A

---

### CEO Q1 — Does it fit the multiharness plan (opencode + pi)?

**Yes on principle, with a justified mechanism difference and one convergence gap.**

The core invariant of the pi precedent (AgDR-0082, #815) and the opencode adapter (#821) — delegate to the unmodified `.claude/hooks/*.sh`, keep `.claude/` the single source of truth, zero gate-logic duplication — is fully preserved. AgDR-0086 explicitly endorses this direction ("reach other harnesses through thin adapters that shell out to the unmodified hooks"), naming #815/#821. This adapter sits squarely in that lane.

The mechanism differs, defensibly. pi (`tool_call`) and opencode (`tool.execute.before`) expose only an imperative TS plugin API, so they ship a live dispatcher extension under `harness-adapters/<name>/`. Codex natively supports a declarative Claude-shaped `hooks.json` at `.codex/hooks.json` and native subagents at `.codex/agents/*.toml` — so a generator emitting declarative config is the right transport; an imperative extension would fight the harness. It is therefore **not** a third governance implementation (logic still lives once, in bash) — it is a third **transport**, inherent to the three harnesses' different surfaces.

Convergence insight (a subtle win over the precedent): Codex derives its gate wiring **directly from `.claude/settings.json`** by generation. pi/opencode hand-maintain a parallel `DEFAULT_GATES` table that AgDR-0082 itself flags as a "keep in visual sync" drift risk. On that axis Codex is the least drift-prone of the three. Recommendation (follow-up, not blocking): file a ticket to converge pi/opencode on deriving their gate list from `.claude/settings.json` too, so all three share one wiring source. A shared *code* core across a bash generator and two TS extensions is not realistic and I would not ask for it.

Also: because Codex shares the stdin contract and the generator preserves `.claude/settings.json`'s inline ops-root anchor-walk verbatim, Codex reuses the *same* bash ops-root resolution as Claude Code — it does not re-implement it the way pi had to (`resolve-ops-root.ts`). Less divergence than the precedent, not more.

### CEO Q2 — Is it backward compatible?

**Yes — purely additive, confirmed mechanically.** The only file under `.claude/` in the diff is a new test (`.claude/hooks/tests/test_sync_codex_adapter.sh`). No existing hook, `settings.json`, rule, or agent is modified. New surface is `bin/sync-codex-adapter.sh`, two docs, a README section, and generated `.agents/`/`.codex/` output (recommended git-excluded). Running the generator does not mutate `.claude/`. A Claude-Code-only user sees zero behavioural change.

### CEO Q3 — Does it actually work for Codex?

**Schema, field names, model IDs, and delegation contract all check out against Codex's real docs:**
- hooks.json schema — Codex's real `PreToolUse` shape is matcher-group + `type:"command"` handlers; the generated JSON matches.
- Handler-level `if` — correctly not part of Codex's schema; compiling it into a shell preflight filter and stripping `if` is right.
- Hook stdin contract (load-bearing) — Codex passes `tool_name` + `tool_input.command` on stdin, identical to what the canonical `.claude/hooks/*.sh` parse and what the generated predicate wrapper reads. The delegation is wired to a real matching contract.
- Blocking — Codex `PreToolUse` blocks on exit code 2; exit codes are preserved end-to-end (test proves 2/0/skip).
- Agents — Codex reads subagents as individual `.toml` files in `.codex/agents/` with required `name` / `description` / `developer_instructions` and optional `model` — exactly what the generator emits.
- Model IDs — `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini` are real Codex identifiers (they appear in Codex's config examples). The opus/sonnet/haiku mapping is valid, not hallucinated.

No schema/model-ID assumption here would break on real Codex. The one thing NOT proven: live enforcement under a real Codex model turn. The bash test proves delegation + exit-code preservation by construction, but no real Codex session was observed blocking an ungated merge. The pi feature (#815 AC-6) and opencode (#821) were both held to a live end-to-end block. Given the schema verification the residual risk is low, so this is a **required follow-up, not a hard blocker** — but it must be tracked.

---

### Blocking Findings

**AgDR ID collision — `AgDR-0085` is already taken (BLOCKING).** `docs/agdr/AgDR-0085-codex-adapter-generation.md` collides with `docs/agdr/AgDR-0085-structurizr-dsl-escape-hatch.md`, already on `dev` via #822. This PR is based on a stale `dev` (its base predates AgDR-0082..0087). Landing as-is puts two `AgDR-0085` files on `dev`, corrupting the ID sequence and breaking `/agdr` tooling. Required: rebase on current `dev` and renumber this record to **AgDR-0088** (next free), updating filename, heading, and self-references. This is the one change holding the sign-off — the gate exists to keep the decision record sound, and a duplicate-ID record is a concrete defect in it.

- Track live-Codex verification as a follow-up ticket mirroring #815 AC-6 / #821 (a real Codex turn attempting an ungated merge is blocked by the unmodified `block-unreviewed-merge.sh` via the generated `.codex/hooks.json`).
- Placement consistency: pi/opencode live under `harness-adapters/<name>/`; this is `bin/` + repo-root output. Consider `harness-adapters/codex/`, or document why Codex is a generator-at-`bin/`, so the three read as one family.
- Tracking-policy vs `--check` tension: the docs recommend git-excluding generated output, yet `--check`/CI-drift-gating only has value when it is committed. State the two modes as an explicit either/or.
- Model mapping does not encode AgDR-0087's frontier floor: the label map is mechanical. opus/gpt-5.5 keeps reviewer agents frontier today, but a future relabel to sonnet would silently drop a reviewer below the review-depth floor. Consider tying reviewer agents to the frontier Codex tier.
- Per-tool-call overhead (NFR): each Bash call now nests wrapper-bash / jq x2 / bash / hook-bootstrap / hook. Consistent with pi's accepted trade-off, so acceptable — worth a line in the AgDR consequences.

### Verdict
**CHANGES REQUESTED**

Architecturally sound, and its Codex-specific claims verify cleanly against real Codex docs — the delegation genuinely works. Blocked only by the AgDR-0085 ID collision + stale-dev base, resolved by rebase-and-renumber (to 0088). Re-request review after the rebase and I will approve and write the Gate 3b marker. Live-Codex proof and the pi/opencode wiring-convergence are follow-ups, not merge blockers.

---
Reviewed by Tariq (Solution Architect) · commit `729d7539ab4a31b621bccf4d10eea641f21b64f7`

me2resh (commented):
## Security Re-Review: PR #730 (delta)

**Commit**: `f6ed283` · re-review of the delta since the `729d753` approval (rebase + AgDR 0085→0088 renumber + the one executable change to `map_codex_model`).

### 1. Enforcement-critical hook wiring — UNCHANGED (byte-identical)
`git diff 729d753 f6ed283 -- bin/sync-codex-adapter.sh` is a **single hunk**: only the `map_codex_model` function body. `generate_hooks_json`, the `.codex/hooks.json` emission, exit-code fidelity, predicate wrapping, and the delegation to `.claude/hooks` are untouched — the trust-chain fidelity I cleared at `729d753` carries forward verbatim. The absolute-path guard (`grep -R "$ROOT"`) still runs unchanged.

### 2. New jq-on-config path — no injection
- `$1` (the model label) is bound via `jq --arg l "$1"` — a **value** binding, never interpolated into the jq program text. A crafted label cannot alter the jq program, and the program `.[$l].codex // $l` is pure data indexing (no jq code execution).
- The mapped output is written to TOML via `printf '%s' "$model" | jq -Rs .` — raw-slurp to a JSON string, which escapes quotes / newlines / TOML metacharacters. Even a malicious value in `harness-models.json` (e.g. `gpt"\ninjected = true`) is neutralised into a single quoted string — **no output injection into the generated TOML**.
- `$1` is only ever a `--arg` value and a `printf '%s'` argument — never eval'd, never in a command position. **No command injection.**
- Failure modes degrade safely: missing file / malformed JSON → jq errors → `2>/dev/null || printf '%s\n' "$1"` returns the label unchanged; absent tier/column → `// $l` fallback. No crash, no partial-write.

### 3. `.claude/harness-models.json` — data-only
Plain tier→model map (`opus`/`sonnet`/`haiku` → `codex` column) plus a descriptive `_comment`. **No secrets, tokens, URLs, or credentials.**

### Verdict
**APPROVED** — the delta is a safe refactor of a hardcoded `case` to a trusted committed config read; the enforcement-critical generation is untouched, and the jq path is injection-free on both the label and the config-file surfaces. Security marker re-pinned to `f6ed283`.

---
📌 Reviewed commit: `f6ed283`

me2resh (commented):
## Design Re-Review: PR #730

**Commit**: `f6ed2833490aa7a662c8b3dbf15cda5805ce99b5`

### Summary
This adapter generates Codex-facing skill/agent/hook wiring from the canonical `.claude/` runtime, delegating gate *execution* to the unmodified `.claude/hooks/*.sh` scripts rather than forking them. Since my prior CHANGES REQUESTED, the PR has been rebased onto current `upstream/dev`, the AgDR renumbered 0085→0088 (collision cleared), and — acting on my convergence recommendation — the model mapping has been lifted out of a hardcoded generator `case` into a shared `.claude/harness-models.json` matrix that the generator reads.

### What I re-verified
- **Blocker 1 (AgDR collision) — CLEARED.** `0085`, `0086`, `0087` are three distinct pre-existing AgDRs at HEAD; this PR's record is `AgDR-0088`. No overlap.
- **Blocker 2 (stale base) — CLEARED.** Base is `dev`, HEAD `f6ed283` rebased on current upstream.
- **Shared matrix is genuinely data-driven.** `map_codex_model()` resolves via `jq -r '.[$l].codex // $l' .claude/harness-models.json` — no hardcoded `case` remains (the one `case` in the script is CLI arg parsing). The smoke test's `model = "gpt-5.4"` assertion is now config-driven off the matrix file it writes into the fixture, not a literal baked into the generator.
- **AgDR-0087 note is accurate.** 0087 does establish the frontier-model floor for reasoning-layer agents; the matrix `_comment` and AgDR-0088 correctly reference it and encode the guardrail "never downgrade the reasoning tier when adding a harness."

### Review Lens Results
- Quality attributes / NFRs:    Pass — enforcement-faithfulness is the load-bearing quality attribute; delegation-not-fork preserves the single audited gate surface.
- Design patterns & structure:  Pass — transport-adapter over the canonical runtime + a single-source-of-truth data file. The matrix is the right shape for cross-harness convergence.
- Technical debt:               Pass — live-Codex conformance test and pi/opencode gate-derivation convergence are both explicit, ticketed follow-ups, not silent debt.
- Decisions (AgDR linkage):     Pass — AgDR-0088 references AgDR-0082 (pi precedent) and AgDR-0087 (frontier floor). BLOCKING collision resolved.
- Risk:                         Pass — additive + generated output untracked; rollback is trivial. Codex-runtime enforcement dependency is documented in Consequences.
- Trade-off analysis:           Pass — four options tabled in the AgDR (track raw output / copy+rewrite hooks / skills-only / delegate).
- Requirements traceability:    Pass — Closes #729.
- Migration safety:             N/A — no data/schema migration.
- Adopter Handbooks:            N/A — framework-internal tooling; no language/domain handbook applies.

### Answers to the three questions
1. **Does the shared matrix realize the "shared core across harness adapters" direction (Q1)?** Yes. Schema `{ tier: { harness: model } }` is exactly the convergence shape I recommended: pi/opencode add a column (`"pi"`, `"opencode"`) per row instead of hardcoding a second mapping. The per-column `// $l` fallback keeps an unlisted harness safe (label passes through — valid for pi, which is Claude-backed). The AgDR-0087 note addresses the frontier-floor concern at the documentation/guardrail level, which is appropriate for a data file (the floor can't be mechanically enforced in a JSON matrix, and 0087 owns the real enforcement story).
2. **Multiharness-fit** — holds, and is *strengthened*: the second harness now costs one column, not a duplicated mapping.
3. **Backward-compat** — holds; purely additive. The only runtime change is the generator reading a new data file; no existing hook/gate/agent behavior changes.
4. **Works-for-Codex** — holds; schema and model IDs unchanged from my prior verification against real Codex docs; `gpt-5.5/gpt-5.4/gpt-5.4-mini` now live in the matrix and are asserted by the smoke test.

### Blocking Findings
None. Both prior blockers resolved.

### Suggestions (non-blocking, for the follow-ups already named)
- When the `opencode` column lands, confirm its `opus` row points at opencode's strongest model per the AgDR-0087 guardrail — the JSON can't enforce it, so it's a review checkpoint on that PR.
- Consider a tiny CI assertion that every harness column present in the matrix has a non-null `opus` entry, to mechanize the "never downgrade the reasoning tier" note over time.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `f6ed2833490aa7a662c8b3dbf15cda5805ce99b5`

me2resh (commented):
## Code Review: PR #730 (re-review)

**Commit**: `f6ed283`

### Summary
Re-review after the sole prior blocker — the AgDR-0085 number collision — was fixed in place, plus a bonus refactor. Verified the three-part delta: the AgDR renumber, the shared model-mapping matrix, and that the hook-wiring generation I already cleared is unchanged.

### Delta verification

**1. Renumber 0085 → 0088 — complete.** `git grep` for `AgDR-0085` across the tree at f6ed283 returns only references to `AgDR-0085-structurizr-dsl-escape-hatch.md` (the real structurizr record, in the `/c4` skill, multi-project docs, and architecture templates) — those are correct and untouched. The codex record is now `docs/agdr/AgDR-0088-codex-adapter-generation.md`, its H1 reads `# AgDR-0088 — …`, and `docs/codex-adapter.md` repoints to `AgDR-0088`. No lingering codex→0085 reference anywhere in the committed tree.

**2. Model mapping extracted to a shared matrix — correct.** `map_codex_model()` now reads `.claude/harness-models.json` via `jq -r --arg l "$1" '.[$l].codex // $l'` instead of a hardcoded `case`. Verified every path directly:
- Tier lookups: `opus→gpt-5.5`, `sonnet→gpt-5.4`, `haiku→gpt-5.4-mini` ✓
- Fallback to the label when the tier is unknown, the `codex` column is absent, or the file is missing (`jq … 2>/dev/null || printf '%s\n' "$1"`) — all three return the label unchanged ✓
- The smoke test is genuinely config-driven: ran the generator against a fixture-less tree and the agent TOML fell back to `model = "sonnet"`; with the fixture it produced `model = "gpt-5.4"`. The `model = "gpt-5.4"` assertion would fail without the fixture, so it exercises the config path rather than a constant.

**3. Hook-wiring generation — unchanged.** `generate_hooks_json` / `wrap_if_handler` and the predicate-compilation-into-shell logic are intact; the delta only touched `map_codex_model`. The hooks.json smoke assertions (delegates to canonical hook path, wraps command predicates, omits unsupported `if` fields, preserves block/allow/skip exit codes) all pass.

### Test results (re-verified, session pin neutralized)
- Codex smoke test: **22/22 PASS**
- Full hook suite (`bin/run-hook-tests.sh`): **PASS=94 FAIL=0**

Note: without neutralizing the pin, the two generated-hook-exec cases fail with exit 126 because the generated command resolves the *real* ops-fork via `CLAUDE_CODE_SESSION_ID` and tries to exec the fixture's `block-test.sh` from there. This is an environment artifact of running the suite inside a live pinned session, not a PR defect — `env -u CLAUDE_CODE_SESSION_ID` restores 22/22, matching the maintainer's report.

### Checklist Results
- Security:                 Pass (adapter delegates gates to unmodified `.claude/hooks/*.sh`; no gate logic forked)
- PR Description & Glossary: Pass (glossary present)
- Technical Decisions (AgDR): Pass — AgDR-0088 documents the shared-matrix decision and references AgDR-0087 (frontier-model floor)

- `nit:` The PR **body** text still says "Records AgDR-0085" in the Summary and lists `docs/agdr/AgDR-0085-codex-adapter-generation.md` under the AgDR section. All committed files are correct (0088); only the PR description text is stale. Worth a quick edit so the description matches the merged artifacts, but not a merge blocker.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `f6ed2833490aa7a662c8b3dbf15cda5805ce99b5`

me2resh (approved):

atlas-apex (approved):
Non-author approval to satisfy the dev branch ruleset. Rex (code), Hakim (trust-chain), and Tariq (architecture / Gate 3b) all reviewed and approved at HEAD f6ed283; CEO approved. Codex adapter delegates gates to the unmodified .claude/hooks — enforcement fidelity verified. Thanks @hossam-96 for the contribution.

me2resh:
Merged 🎉 — thank you, @hossam-96!

This is a genuinely strong contribution. The core call — generating `.codex/` from the canonical `.claude/` and delegating gate execution to the **unmodified** `.claude/hooks/*.sh` instead of forking the logic — is exactly the right shape, and it held up under a thorough three-lens review:

- **Trust-chain (Hakim):** empirically confirmed a blocking `exit 2` propagates unchanged through the generated Codex wrapper — no gate can be silently weakened for Codex users — and that injection payloads fail to execute while the gate still blocks.
- **Architecture (Tariq):** verified the `hooks.json` matcher shape, the stdin contract, and the `gpt-5.5/gpt-5.4/gpt-5.4-mini` model IDs against the real Codex docs — nothing hallucinated.
- **Code (Rex):** all 46 `if`-predicates correctly converted, exit codes preserved, smoke test genuinely exercising the paths.

We fixed the two blockers in place so you didn't have to churn: rebased onto current `dev`, and renumbered the AgDR to **0088** (0085 had been claimed by another merge in the meantime). We also extended your model mapping into a shared **`.claude/harness-models.json`** matrix, so the upcoming pi/opencode adapters add a *column* instead of re-hardcoding a `case` — your pattern, generalized.

Two follow-ups we'll track ourselves (not on you): a live-Codex runtime conformance test, and converging the pi/opencode gate-derivation on `settings.json`.

Adding you to the Contributors list now. Thanks again! 🙏

me2resh:
@hossam-96 Thank you for your contribution to ApexYard, including this PR.

PR #734, opened 2026-06-27
feat(#733): apexyard-search auto-reindex on session-start + broaden agent MCP allowlists

### LINKED ISSUE #733: [Feature] apexyard-search freshness + agent access — auto-reindex on session-start + broaden MCP allowlists
## User Story
As an operator relying on apexyard-search to save tokens, I want the index kept fresh automatically and the search tools available to the agents that do lookups — so the MCP actually gets used instead of agents falling back to grep on a stale/unreachable index.

## Context
A `/debug` of "why is the MCP underused" found (besides coverage, fixed in apexyard-premium#364) two non-research root causes that live in the FRAMEWORK:
1. **Staleness** — reindex is advisory-only, wired to no automatic event, so the index goes days stale → poor results → agents stop trusting it.
2. **Access** — many role agents that do code/doc lookups lack `search_code`/`search_docs` in their tool allowlist, so they *can't* call the MCP. (The harness built-ins Explore/general-purpose can't be modified — out of scope.)

This is the deterministic half of apexyard-premium#362 (the embedding-model + reranker *quality* half is a separate spike).

## Acceptance Criteria
- [ ] **Auto-reindex (freshness)** — a `SessionStart` hook that opportunistically triggers an **incremental** reindex when the `apexyard-search` CLI is available AND the index is stale beyond a threshold. MUST be: timeout-guarded, `exit 0` always (never blocks session start), a no-op when the CLI/MCP isn't installed (most adopters) or the index is fresh. Incremental so it's cheap (the manifest mtime path already exists).
- [ ] **Broaden agent allowlists** — add `mcp__apexyard-search__search_code` / `search_docs` to the `.claude/agents/*.md` role agents that do lookups but currently lack them (e.g. ux/ui-designer, the Heads, sre, penetration-tester, product roles, data-analyst). Keep it sensible — only agents that actually read code/docs.
- [ ] Tests for the hook (stale → triggers; fresh → no-op; CLI absent → no-op; timeout-guarded; exit 0 always); shellcheck clean.
- [ ] Backward compatible: adopters without apexyard-search see zero behaviour change.

## Out of Scope
- The embedding-model upgrade + cross-encoder reranker (the *quality* question — see the apexyard-search relevance spike).
- Coverage (already fixed: projects/ docs now indexed).

## Glossary
| Term | Definition |
|------|------------|
| Incremental reindex | Re-indexing only files changed since the last run (mtime/content-hash), so it's cheap to run often |
| Opportunistic | Runs on session start only when useful (stale + CLI present), never blocking |
| Allowlist | The `tools:` frontmatter of an agent definition — what tools that agent may call |

atlas-apex:
## Summary

- **Auto-reindex hook** (`reindex-on-session-start.sh`) fires at `SessionStart` and emits an advisory banner when the apexyard-search vector-store is stale (default threshold: 1 hour), instructing the agent to call `mcp__apexyard-search__reindex` before its first search — closing the \"stale on cold session\" gap that caused agents to silently fall back to grep+Read
- **Staleness detection** uses the mtime of `~/.cache/apexyard-search/lancedb/` (or `$APEXYARD_SEARCH_CACHE_DIR`), mirroring `paths.py` exactly; threshold is overridable via `APEXYARD_SEARCH_STALE_THRESHOLD`; a missing index emits a `scope=\"all\"` prompt; a stale one emits a cheaper `scope=\"framework\"` incremental prompt
- **Agent allowlists broadened** — added `mcp__apexyard-search__search_code` and `mcp__apexyard-search__search_docs` to 12 role agents that do code/doc lookups but previously lacked access: ux-designer, ui-designer, head-of-design, head-of-data, penetration-tester, head-of-product, product-manager, head-of-security, head-of-engineering, sre, product-analyst, data-analyst; dependency-auditor / pr-manager / ticket-manager omitted (no semantic code lookup)

## Safety properties

- `exit 0` **always** — never blocks session start
- **No-op** for the typical adopter: silently skips when `apexyard-search` is not on PATH, `.mcp.json` is absent, or the index is fresh
- **Timeout-guarded**: stat + date run inside a 10 s subshell (`timeout -k 2 10` / `gtimeout`) so stalled network mounts never hang startup
- Consistent with the advisory pattern of `remind-mcp-tools.sh` and `suggest-mcp-reindex-after-pull.sh`

## Testing

Run from the repo root after checking out this branch:

```bash
# Hook tests (18 cases: CLI absent, no MCP, fresh, stale, missing, threshold=0, exit-0, timeout-guard)
bash .claude/hooks/tests/test_reindex_on_session_start.sh

# Sibling hook regression checks
bash .claude/hooks/tests/test_suggest_mcp_reindex_after_clone.sh
bash .claude/hooks/tests/test_suggest_mcp_reindex_after_pull.sh

# Shell syntax + lint
bash -n .claude/hooks/reindex-on-session-start.sh
shellcheck -S warning .claude/hooks/reindex-on-session-start.sh

# settings.json validity + hook count
python3 -c "import json; d=json.load(open('.claude/settings.json')); print(len(d['hooks']['SessionStart'][0]['hooks']), 'SessionStart hooks')"
# Expected: 11 SessionStart hooks
```

Closes #733
Cross-ref: me2resh/apexyard-premium#362

---

## Glossary

| Term | Definition |
|------|------------|
| Staleness threshold | The minimum age (in seconds) of the vector-store directory that triggers the reindex advisory; default 3600 s (1 hour), tunable via `APEXYARD_SEARCH_STALE_THRESHOLD` |
| Incremental reindex | `mcp__apexyard-search__reindex(scope=\"framework\")` — reindexes only framework docs (roles, skills, rules, workflows), not managed-project codebases; faster and sufficient for daily framework updates |
| Cache dir | `~/.cache/apexyard-search/` (or `$APEXYARD_SEARCH_CACHE_DIR`) — where the LanceDB vector store lives; resolved to match `paths.py` exactly so the hook and the MCP server agree on the location |
| Advisory banner | A message printed to stderr that informs the agent what to do; never blocks execution (always `exit 0`); same shape as `suggest-mcp-reindex-after-pull.sh` |
| Allowlist | The `allowed-tools:` frontmatter field in an agent `.md` file — determines which tools that agent may invoke; adding MCP tool names here is what \"broadening the allowlist\" means |

### DIFF
--- a/.claude/agents/data-analyst.md
+++ b/.claude/agents/data-analyst.md
@@ -2,7 +2,7 @@
 name: data-analyst
 description: Writes SQL, builds dashboards, runs A/B-test analysis, and investigates metrics. Activates on SQL / dashboard / A/B-test / metric-investigation work — quantitative, fast, narrow tool-use (candidate for local-model routing per #348 spike).
 model: haiku
-allowed-tools: Bash, Read, Edit, Write, Grep, Glob
+allowed-tools: Bash, Read, Edit, Write, Grep, Glob, mcp__apexyard-search__search_code, mcp__apexyard-search__search_docs
 persona_name: Nadia
 ---
 
--- a/.claude/agents/head-of-data.md
+++ b/.claude/agents/head-of-data.md
@@ -2,7 +2,7 @@
 name: head-of-data
 description: Owns analytics strategy, data governance, reporting architecture, and cross-project data modelling. Activates on cross-project data calls, governance decisions, reporting-architecture reviews, and strategic data tooling choices.
 model: sonnet
-allowed-tools: Bash, Read, Edit, Write, Grep, Glob
+allowed-tools: Bash, Read, Edit, Write, Grep, Glob, mcp__apexyard-search__search_code, mcp__apexyard-search__search_docs
 persona_name: Khalil
 ---
 
--- a/.claude/agents/head-of-design.md
+++ b/.claude/agents/head-of-design.md
@@ -2,7 +2,7 @@
 name: head-of-design
 description: Owns the design system, UX principles, and visual standards across products. Activates on design-system changes, UX principles decisions, or cross-project visual-standards calls.
 model: sonnet
-allowed-tools: Bash, Read, Edit, Write, Grep, Glob
+allowed-tools: Bash, Read, Edit, Write, Grep, Glob, mcp__apexyard-search__search_code, mcp__apexyard-search__search_docs
 persona_name: Maha
 ---
 
--- a/.claude/agents/head-of-engineering.md
+++ b/.claude/agents/head-of-engineering.md
@@ -2,7 +2,7 @@
 name: head-of-engineering
 description: Owns engineering strategy, architecture standards, and engineering culture across the portfolio. Activates on architecture review, new tech stack additions, cross-project engineering calls, and Tech Lead escalations.
 model: opus
-allowed-tools: Bash, Read, Edit, Write, Grep, Glob
+allowed-tools: Bash, Read, Edit, Write, Grep, Glob, mcp__apexyard-search__search_code, mcp__apexyard-search__search_docs
 persona_name: Khalid
 ---
 
--- a/.claude/agents/head-of-product.md
+++ b/.claude/agents/head-of-product.md
@@ -2,7 +2,7 @@
 name: head-of-product
 description: Owns product strategy, roadmap prioritisation, and feasibility studies. Activates on roadmap prioritisation, feasibility calls, strategic product decisions, or cross-product resource allocation.
 model: sonnet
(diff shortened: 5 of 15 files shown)


atlas-apex:
Closing after independent verification found this doesn't fix either root cause:
- **Access half is a no-op** — the repo's `allowed-tools:` frontmatter field is ignored at runtime (agents using it already get all tools; only `tools:` restricts). The agents this granted the MCP to already had it. The original debug mis-counted by grepping the frontmatter string.
- **Freshness half is only an advisory banner** — the `apexyard-search` binary has no shell `reindex` subcommand, so a hook can't actually trigger a reindex.

The real fix lives in apexyard-search (a `reindex` CLI subcommand + a hook that calls it) — filed as me2resh/apexyard-premium#371.

Issue #501, opened 2026-06-04
[Feature] Make ticket-existence hooks tracker-aware (don't require a GitHub issue when the configured tracker isn't GitHub)


me2resh:
## Problem
`validate-pr-create.sh` and the sibling `verify-commit-refs.sh` assume **GitHub Issues**. When a project sets `.tracker.kind` to `jira`/`linear`/etc., a PR title or commit message that references a real tracker key (e.g. `ABC-1234`) is **blocked at creation**: the hook extracts `1234`, looks for GitHub issue `#1234`, doesn't find it, and fails. On a non-GitHub tracker it is therefore impossible to open a PR whose title references a real, valid ticket.

## Proposed behaviour
When `.tracker.kind` is **not** `gh`, the ticket-existence hooks should either:
- validate the referenced key against the **configured tracker** (e.g. via the `view_command` already defined in `_lib-tracker.sh`), or
- fall back to **shape-only** validation against `.tracker.id_pattern` (confirm the key is well-formed, don't assert existence).

They should **not** require a matching GitHub issue number to exist. The GitHub-issue existence check should remain the behaviour only when `.tracker.kind == gh`.

## Why it helps adopters
Teams whose tracker is Jira/Linear/etc. currently cannot open PRs that reference their real tickets — the hook blocks them on a GitHub issue that will never exist. This contradicts the intent of the configurable tracker dispatch (`.tracker.kind`, AgDR-0033): the validators already source `id_pattern` per-tracker, but the existence check is still hardwired to GitHub.

## Scope / out of scope
- In scope: tracker-awareness of `validate-pr-create.sh` + `verify-commit-refs.sh`.
- Out of scope: building new tracker API integrations; changing `id_pattern` defaults.

## Framework version
v2.2.0

## Related
`.claude/hooks/validate-pr-create.sh`, `.claude/hooks/verify-commit-refs.sh`, `.claude/hooks/_lib-tracker.sh`; AgDR-0033 (configurable tracker dispatch).

## Glossary
| Term | Definition |
|------|------------|
| `.tracker.kind` | Project-config field selecting the active tracker (`gh`, `jira`, `linear`, …). |
| Ticket-existence hook | A PreToolUse hook that verifies a referenced ticket exists before a PR/commit is created. |
| Shape-only validation | Checking a ticket key matches `id_pattern` without verifying it exists in a tracker. |
| `view_command` | The `_lib-tracker.sh` template used to look a ticket up in the configured tracker. |

[...]

Issue #42, opened 2026-04-12
fix: CI failures — markdown lint and link check


atlas-apex:
## Problem

Two CI checks fail on every PR:

1. **Markdown Lint** — `workflows/sdlc.md` has blank-line violations (MD022, MD032) in the Deploy and Monitor sections
2. **Link Check (lychee)** — `site/index.html` and docs contain links to `https://github.com/me2resh/apexstack` which returns 404 (private repo)

## Fix

1. Add missing blank lines in `workflows/sdlc.md`
2. Configure lychee to exclude self-referential repo links (`.lycheeignore` or workflow config)

PR #760, opened 2026-07-02
fix(#755): route migration gate through tracker abstraction (GitLab support)

### LINKED ISSUE #755: [Bug] require-migration-ticket.sh hardcodes gh CLI, blocking GitLab projects from satisfying the migration gate
## Affected
`.claude/hooks/require-migration-ticket.sh` — Gate 2 (issue is open + has migration label)

## Given / When / Then
**Given** a managed project whose tracker is on GitLab (configured as `tracker.kind: glab` or `tracker.kind: custom` in the project's registry entry), with an active ticket marker pointing at a GitLab issue that is open, labelled `migration`, and whose body references a migration AgDR

**When** an edit to a migration-path file is attempted (triggering `require-migration-ticket.sh`)

**Then** the hook exits 2 with "BLOCKED: Could not fetch `<repo>#<N>` from GitHub" — even though the ticket legitimately exists on GitLab and all gate conditions are satisfied — because Gate 2 hardcodes `gh issue view` at line 267 and never consults `tracker_kind()` from `_lib-tracker.sh`

**Expected**: Gate 2 falls back to `glab issue view` when `tracker.kind` is `glab`, and skips online verification entirely when `tracker.kind` is `none` — consistent with how `tracker_view()` in `_lib-tracker.sh` already abstracts multi-tracker dispatch

## Repro
1. Register a project in `apexyard.projects.yaml` with `tracker: { kind: glab }` (GitLab repo)
2. Run `/start-ticket <gitlab-issue-number>` — marker is written correctly
3. Attempt to edit a file matching a migration-path pattern
4. Hook fires, calls `gh issue view <N> --repo <gitlab-repo>` — `gh` returns empty output for GitLab repos
5. Hook blocks with "Could not fetch from GitHub" despite the GitLab issue being valid and open

## Root cause
`require-migration-ticket.sh` line 267 calls `gh issue view` directly. It does not source `_lib-tracker.sh` or call `tracker_kind()` / `tracker_view()`, which already have `glab` support (for `tracker_create`) and a `none` no-op path. The hook predates the tracker abstraction layer and was never updated to use it.

## Fix (already applied to ops fork as a local patch)
After the `gh` call fails (empty `ISSUE_JSON`), fall back to `glab issue view --output json` and normalise the GitLab response shape (`state: "opened"` → `"OPEN"`, `description` → `body`, `labels[]` as string array) before proceeding with the same Gate 2/3 checks.

## Framework version
v4.1.0

## Severity
major — GitLab-tracked projects cannot edit any migration file; no workaround within the framework (requires patching the hook in the ops fork)

## Notes
`_lib-tracker.sh` already has `_tracker_create_glab()` (line 503) but no `_tracker_view_glab()`. A fuller fix would add `_tracker_view_glab()` and add a `glab)` case to the `tracker_view()` switch, then have `require-migration-ticket.sh` call `tracker_view` instead of hardcoding `gh`. This would also fix any other hooks with the same pattern.

## Glossary
| Term | Definition |
|------|------------|
| `tracker_view` | Public function in `_lib-tracker.sh` that dispatches issue-view calls to the configured tracker CLI (gh, linear, jira, asana, custom, none) |
| `tracker_kind` | Public function in `_lib-tracker.sh` that reads the active tracker type from project config or registry override |
| migration gate | Gate 3a in `.claude/rules/workflow-gates.md` — requires an open, labelled migration ticket + AgDR before migration-path edits are permitted |

AbdElrahmaN31:
## Summary

- **Root cause fixed** — `require-migration-ticket.sh` Gate 2 hardcoded `gh issue view`, so GitLab-tracked (`tracker.kind: glab`) projects false-blocked *every* migration-file edit even when the GitLab ticket was valid and open. Gate 2 now routes through `tracker_view` from `_lib-tracker.sh`, so the migration gate works for any configured tracker.
- **glab is now a first-class view kind** (the issue's "fuller fix", not the local `gh`-then-`glab` fallback patch) — new `_tracker_normalise_glab` adapter (`opened`→`OPEN`, `web_url`→`url`, `description`→`body`, string labels), a `glab)` dispatch case, and a **kind-aware built-in `view_command` default** so a glab/gh adopter who only sets `kind` gets the matching command. Mirrors the existing `_tracker_create_glab` creation adapter.
- **`body` added to the normalised `tracker_view` contract** (gh + glab only) — the migration gate reads it to find the linked AgDR (Gate 3). Scoped to gh/glab deliberately: jira `.description` is ADF (not a grep-able string) and linear/asana bodies have no consumer yet.
- **Unpinned `.tracker.view_command` in `project-config.defaults.json`** — it was deep-merging over a glab adopter's `kind: glab` and forcing the gh command (the bug at the config layer). The lib now owns a kind-aware default, so the pin is unnecessary and harmful.
- **`tracker.kind=none` short-circuits** the online label/AgDR checks (operator-trusted, audible stderr note), per the issue's Expected behaviour. **Closed-state detection is now tracker-agnostic** (shares the `closed|done|cancelled|resolved|…` vocabulary with `validate-pr-create.sh` / `verify-commit-refs.sh`) instead of the gh-only `!= "OPEN"`.

## Design decision — the migration gate is fail-closed on an unfetchable ticket

For a non-gh tracker, the existence checks (`validate-pr-create.sh` / `verify-commit-refs.sh`) fall back to shape-only PASS when the CLI is unreachable (#501). The **migration gate is deliberately stricter**: an unfetchable ticket → block, even for glab, because a migration edit is high-blast-radius. Adopters who genuinely cannot query a tracker set `tracker.kind=none` (which skips the gate). This matches the tested local patch behaviour on the reporting fork.

## Security hardening (review round — commit `86b439f`)

Routing Gate 2 through `tracker_view` reintroduced a command-injection sink that the base hook's direct-argv `gh issue view "$TICKET_NUM"` avoided: `tracker_view` builds its command by string-substituting `{id}`/`{owner_repo}` into a template and running `eval`, and `require-migration-ticket.sh` fed it marker-derived `TICKET_NUM`/`TICKET_REPO` (parsed via `cut -d= -f2-`) with **no shape check**. A marker line like `number=42; touch X` executed arbitrary commands regardless of the gate verdict. Closed in two layers:

- **Caller-side shape guard** (`require-migration-ticket.sh`) — whitelist `TICKET_NUM` (`^#?[A-Za-z0-9_-]+$`) and `TICKET_REPO` (`^[A-Za-z0-9._/-]+$`) before the tracker call; block (exit 2) on any shell metacharacter. Mirrors the shape guarantee `validate-pr-create.sh` / `verify-commit-refs.sh` already get from digit-only extraction — this hook was the sole marker→`tracker_view` caller lacking one.
- **Library defence-in-depth** (`_lib-tracker.sh` `_tracker_substitute`) — `printf %q` the `{id}`/`{owner_repo}` tokens so any future caller forwarding unvalidated input into `tracker_view` can't reopen the sink. No-op for legitimate ids/slugs (incl. nested GitLab groups); also fixes a latent `#42`-starts-a-comment bug in the unquoted default template.

## Testing

1. `bash bin/run-hook-tests.sh` → **77/77 PASS** (includes the new file).
2. `test_require_migration_ticket.sh` (15 cases): gh + glab happy paths, `none` skip, missing-label / closed (gh + glab) / missing-AgDR blocks, gh + glab unfetchable → block, non-migration passthrough, no-marker, jira Gate-3 body-scoping note, **injection: metachar `number=` / `repo=` blocked (exit 2) and not executed**, and `#`-prefixed number passes and is shell-safe.
3. `test_tracker_aware_hooks.sh` +3 lib smoke cases: 2 glab (`opened→OPEN`, `web_url→url`, `description→body`; `closed→CLOSED`) + **1 injection case** (`tracker_view` neutralises a metacharacter id via `printf %q`; legit id substitutes unchanged).
4. glab output shape verified against **live GitLab** (`glab issue view … -R gitlab-org/gitlab-foss --output json`): `state`, `title`, `web_url`, `description`, `labels` are all top-level keys with the expected value shapes.

## Compatibility note

Default forks are unaffected (defaults.json no longer pins `view_command`). The one regression surface: an adopter who *explicitly pinned* the old gh `view_command` (without `body`) in their own `project-config.json` will get an empty body and false-block Gate 3 — they should drop the pin (letting the kind-aware default apply) or append `,body` to their template.

Closes #755

## Glossary

| Term | Definition |
|------|------------|
| `tracker_view` | Public function in `_lib-tracker.sh` that dispatches an issue-view to the configured tracker CLI (gh/glab/linear/jira/asana/custom/none) and emits normalised JSON `{state,title,url,labels,body}`. |
| `_tracker_normalise_<kind>` | Per-tracker adapter that maps a CLI's raw JSON into the normalised contract shape. |
| kind-aware view default | The built-in `view_command` the lib picks when none is configured, chosen by `tracker.kind` (gh → `gh issue view … --json …,body`; glab → `glab issue view … --output json`). |
| migration gate (Gate 3a) | The `require-migration-ticket.sh` rule: a migration-path edit requires an OPEN, `migration`-labelled ticket whose body links a migration AgDR. |
| deep-merge | `jq '.[0] * .[1]'` recursively merges nested objects, so `.tracker.view_command` from defaults survived a `kind`-only override — the config-layer cause of this bug. |
| ADF | Atlassian Document Format — Jira's rich-text body representation (a JSON object, not plain text), which is why `body` is not mapped for the jira adapter. |
| `printf %q` | Bash builtin that emits a shell-quoted (escaped) form of a string, reusable as literal input; used in `_tracker_substitute` so metacharacters in a substituted `{id}`/`{owner_repo}` are neutralised before the `eval` in `tracker_view`. |

### DIFF
--- a/.claude/hooks/_lib-tracker.sh
+++ b/.claude/hooks/_lib-tracker.sh
@@ -16,7 +16,12 @@
 #   tracker_owner_repo_param <slug>    formats the owner/repo parameter (gh: "owner/repo"; others: empty)
 #   tracker_view <id> [<owner_repo>]   dispatches the view command and emits normalised JSON on stdout
 #                                      Exit 0 = ticket exists; non-zero = doesn't, or CLI errored.
-#                                      JSON shape: {"state":..., "title":..., "url":..., "labels":[...]}
+#                                      JSON shape: {"state":..., "title":..., "url":..., "labels":[...], "body":...}
+#                                      `body` is populated for the gh and glab adapters (the kinds
+#                                      that have a consumer needing it — the migration gate reads it
+#                                      to find the linked AgDR, #755). Other adapters emit body:""
+#                                      until a consumer needs it (jira `.description` is ADF, not a
+#                                      grep-able string; linear/asana bodies have no consumer yet).
 #
 # Per-project resolution (#670 / AgDR-0072): tracker_kind / tracker_id_pattern /
 # tracker_view take an OPTIONAL owner/repo. When supplied, a `tracker:` block on
@@ -208,7 +213,7 @@ tracker_id_pattern() {
 # ------------------------------------------------------------------------------
 _TRACKER_VIEW_TPL_CACHE=""
 _tracker_view_template() {
-  local repo="${1:-}"
+  local repo="${1:-}" kind="${2:-}"
   if [ -n "$repo" ]; then
     local pv
     if pv=$(_tracker_project_value "$repo" view_command) && [ -n "$pv" ]; then
@@ -222,9 +227,21 @@ _tracker_view_template() {
   fi
   _tracker_load_config_lib
   local tpl
-  tpl=$(config_get_or '.tracker.view_command' 'gh issue view {id} --repo {owner_repo} --json state,title,url,labels' 2>/dev/null)
+  # An explicit .tracker.view_command (registry per-project or config) always
+  # wins. With none set, fall back to a per-KIND built-in default: glab gets a
+  # first-class GitLab command (parity with _tracker_create_glab, #755); every
+  # other kind gets the gh shape. linear/jira/asana adopters still supply their
+  # own view_command (unchanged contract) — they only land on the gh default if
+  # they forgot to set one. The default view_command is deliberately NOT pinned
+  # in project-config.defaults.json, so this kind-aware fallback can fire (a
+  # pinned default would deep-merge over a glab adopter's `kind: glab` and force
+  # the gh command — the exact #755 bug at the config layer).
+  tpl=$(config_get_or '.tracker.view_command' '' 2>/dev/null)
   if [ -z "$tpl" ] || [ "$tpl" = "null" ]; then
-    tpl='gh issue view {id} --repo {owner_repo} --json state,title,url,labels'
+    case "$kind" in
+      glab) tpl='glab issue view {id} -R {owner_repo} --output json' ;;
+      *)    tpl='gh issue view {id} --repo {owner_repo} --json state,title,url,labels,body' ;;
+    esac
   fi
   if [ -z "$repo" ]; then
     _TRACKER_VIEW_TPL_CACHE="$tpl"
@@ -260,8 +277,8 @@ _tracker_substitute() {
 # Internal adapter: gh → normalised JSON.
 #
 # Reads `gh issue view` JSON. The default view_command requests state, title,
-# url, labels — labels comes back as an array of objects with .name keys, so
-# we flatten to a string array.
+# url, labels, body — labels comes back as an array of objects with .name keys,
+# so we flatten to a string array; body is passed through verbatim.
 # ------------------------------------------------------------------------------
 _tracker_normalise_gh() {
   local raw="$1"
@@ -276,7 +293,32 @@ _tracker_normalise_gh() {
     state:  (.state // ""),
     title:  (.title // ""),
     url:    (.url // ""),
-    labels: ((.labels // []) | map(if type == "object" then .name else . end))
+    labels: ((.labels // []) | map(if type == "object" then .name else . end)),
+    body:   (.body // "")
+  }' 2>/dev/null
+}
+
+# ------------------------------------------------------------------------------
+# Internal adapter: glab (GitLab) → normalised JSON.
+#
+# `glab issue view <id> -R <repo> --output json` emits the GitLab REST issue
+# object: .state is "opened"/"closed", .web_url is the URL, .description is the
+# body, and .labels is already an array of strings. State is normalised to
+# OPEN/CLOSED for contract parity with the gh adapter (downstream closed-state
+# classification is case-insensitive, so either casing works — parity is for
+# consumers that read .state directly). glab is a first-class view kind
+# alongside gh, mirroring the existing _tracker_create_glab creation adapter.
+# ------------------------------------------------------------------------------
+_tracker_normalise_glab() {
+  local raw="$1"
+  if [ -z "$raw" ]; then return 1; fi
+  if ! printf '%s' "$raw" | jq -e . >/dev/null 2>&1; then return 1; fi
+  printf '%s' "$raw" | jq -c '{
+    state:  ((.state // "") | if . == "opened" then "OPEN" elif . == "closed" then "CLOSED" else (. | ascii_upcase) end),
+    title:  (.title // ""),
+    url:    (.web_url // .url // ""),
+    labels: ((.labels // []) | map(if type == "object" then .name else . end)),
+    body:   (.description // .body // "")
   }' 2>/dev/null
 }
 
@@ -399,7 +441,7 @@ tracker_view() {
   fi
 
   local tpl cmd raw rc
-  tpl=$(_tracker_view_template "$owner_repo")
+  tpl=$(_tracker_view_template "$owner_repo" "$kind")
   cmd=$(_tracker_substitute "$tpl" "$id" "$owner_repo")
 
   # Run the command; capture stdout. Suppress stderr (CLI errors are visible
@@ -413,6 +455,7 @@ tracker_view() {
   local normalised
   case "$kind" in
     gh)     normalised=$(_tracker_normalise_gh "$raw") ;;
+    glab)   normalised=$(_tracker_normalise_glab "$raw") ;;
     linear) normalised=$(_tracker_normalise_linear "$raw") ;;
     jira)   normalised=$(_tracker_normalise_jira "$raw") ;;
     asana)  normalised=$(_tracker_normalise_asana "$raw") ;;
--- a/.claude/hooks/_lib-tracker.sh
+++ b/.claude/hooks/_lib-tracker.sh
@@ -265,12 +265,25 @@ tracker_owner_repo_param() {
 
 # ------------------------------------------------------------------------------
 # Internal: substitute {id} and {owner_repo} placeholders in the view template.
+#
+# The substituted string is run via `eval` in tracker_view, so the {id} /
+# {owner_repo} values must not be able to inject command syntax. Both are
+# shell-quoted with `printf %q` before substitution — a no-op for legitimate
+# ticket IDs / owner-repo slugs, and a neutraliser for anything containing shell
+# metacharacters. This is defence-in-depth behind each caller's own shape check:
+# validate-pr-create.sh / require-migration-ticket.sh validate before calling in,
+# but quoting here guarantees a future caller that forwards unvalidated input into
+# tracker_view can't reopen the injection hole. (#755 security review — Rex/Hakim.)
 # ------------------------------------------------------------------------------
 _tracker_substitute() {
   local tpl="$1" id="$2" owner_repo="$3"
+  local q_id q_owner_repo
+  # printf -v is available in bash 3.2 (macOS default); %q shell-escapes.
+  printf -v q_id '%q' "$id"
+  printf -v q_owner_repo '%q' "$owner_repo"
   # Use POSIX parameter expansion — portable across bash 3.2 (macOS default).
-  tpl="${tpl//\{id\}/$id}"
-  tpl="${tpl//\{owner_repo\}/$owner_repo}"
+  tpl="${tpl//\{id\}/$q_id}"
+  tpl="${tpl//\{owner_repo\}/$q_owner_repo}"
   echo "$tpl"
 }
 
--- a/.claude/hooks/_lib-tracker.sh
+++ b/.claude/hooks/_lib-tracker.sh
@@ -19,9 +19,10 @@
 #                                      JSON shape: {"state":..., "title":..., "url":..., "labels":[...], "body":...}
 #                                      `body` is populated for the gh and glab adapters (the kinds
 #                                      that have a consumer needing it — the migration gate reads it
-#                                      to find the linked AgDR, #755). Other adapters emit body:""
-#                                      until a consumer needs it (jira `.description` is ADF, not a
-#                                      grep-able string; linear/asana bodies have no consumer yet).
+#                                      to find the linked AgDR, #755). Other adapters omit the body
+#                                      key entirely (consumers read it as `.body // empty`) until one
+#                                      needs it (jira `.description` is ADF, not a grep-able string;
+#                                      linear/asana bodies have no consumer yet).
 #
 # Per-project resolution (#670 / AgDR-0072): tracker_kind / tracker_id_pattern /
 # tracker_view take an OPTIONAL owner/repo. When supplied, a `tracker:` block on
--- a/.claude/hooks/require-migration-ticket.sh
+++ b/.claude/hooks/require-migration-ticket.sh
@@ -264,30 +264,81 @@ MSG
 fi
 
 # --------- Gate 2: issue is open + has migration label ---------
-ISSUE_JSON=$(gh issue view "$TICKET_NUM" --repo "$TICKET_REPO" --json state,labels,body 2>/dev/null)
+# Resolve the ticket through the tracker abstraction (_lib-tracker.sh) so this
+# gate works for every configured tracker — GitHub (gh), GitLab (glab), Linear,
+# Jira, Asana, custom — not just GitHub. Before #755 this hardcoded
+# `gh issue view`, which returned empty for GitLab-tracked projects and
+# false-blocked every migration edit even when the GitLab ticket was valid.
+# tracker_view emits the normalised {state,title,url,labels,body} shape
+# regardless of tracker; labels come back as a flat string array and body is
+# populated for gh/glab (the kinds the migration gate reads).
+if [ -f "$HOOK_DIR/_lib-tracker.sh" ]; then
+  # shellcheck source=/dev/null
+  . "$HOOK_DIR/_lib-tracker.sh"
+fi
+
+# Resolve the tracker kind for the TICKET's repo (per-project registry override
+# wins over the global config; see #670). tracker.kind=none has no queryable
+# tracker, so existence / label / AgDR can't be verified online — per the #755
+# Expected behaviour, skip the online gates (Gate 1 already proved an active
+# ticket marker exists) and allow the edit, operator-trusted. The skip is
+# printed to stderr so it is auditable, never silent.
+TICKET_KIND="gh"
+if command -v tracker_kind >/dev/null 2>&1; then
+  TICKET_KIND=$(tracker_kind "$TICKET_REPO" 2>/dev/null)
+  [ -n "$TICKET_KIND" ] || TICKET_KIND="gh"
+fi
+if [ "$TICKET_KIND" = "none" ]; then
+  echo "note: tracker.kind=none for ${TICKET_REPO} — skipping migration label/AgDR verification (operator-trusted; #755)." >&2
+  exit 0
+fi
+
+if command -v tracker_view >/dev/null 2>&1; then
+  ISSUE_JSON=$(tracker_view "$TICKET_NUM" "$TICKET_REPO" 2>/dev/null)
+else
+  # Library missing (should not happen in a real fork) — fall back to gh so the
+  # gate still functions on a GitHub tracker rather than bricking, normalising
+  # to the same shape tracker_view emits (labels as a flat string array).
+  ISSUE_JSON=$(gh issue view "$TICKET_NUM" --repo "$TICKET_REPO" --json state,title,url,labels,body 2>/dev/null \
+    | jq -c '{state,title,url,labels:((.labels // []) | map(.name)),body}' 2>/dev/null)
+fi
+
 if [ -z "$ISSUE_JSON" ]; then
   cat >&2 <<MSG
-BLOCKED: Could not fetch ${TICKET_REPO}#${TICKET_NUM} from GitHub.
-Network / auth problem? Or the issue doesn't exist?
+BLOCKED: Could not fetch ${TICKET_REPO}#${TICKET_NUM} from the configured
+tracker (kind: ${TICKET_KIND}). Network / auth problem? Or the issue
+doesn't exist?
 
-Migration files can't be edited without a verifiable migration ticket.
-Check your gh auth status, or run /migration to create a new ticket.
+The migration gate is fail-closed by design — a high-blast-radius change is
+not allowed against a ticket the framework cannot verify. (This is stricter
+than the PR-create / commit-ref existence checks, which fall back to
+shape-only when a non-gh CLI is unreachable, per #501 — a migration edit
+warrants a hard stop.) If your tracker is untracked, set tracker.kind=none.
+
+Check your tracker auth (e.g. gh auth status / glab auth status), or run
+/migration to create a new ticket.
 MSG
   exit 2
 fi
 
-STATE=$(echo "$ISSUE_JSON" | jq -r .state)
-HAS_LABEL=$(echo "$ISSUE_JSON" | jq -r --arg L "$MIGRATION_LABEL" '.labels | map(.name) | index($L) != null')
-BODY=$(echo "$ISSUE_JSON" | jq -r .body)
+STATE=$(echo "$ISSUE_JSON" | jq -r '.state // empty')
+# Normalised labels are a flat string array, so a direct membership test.
+HAS_LABEL=$(echo "$ISSUE_JSON" | jq -r --arg L "$MIGRATION_LABEL" '.labels | index($L) != null')
+BODY=$(echo "$ISSUE_JSON" | jq -r '.body // empty')
 
-if [ "$STATE" != "OPEN" ]; then
-  cat >&2 <<MSG
-BLOCKED: Active ticket ${TICKET_REPO}#${TICKET_NUM} is $STATE, not OPEN.
+# Closed-state recognition is tracker-agnostic — same vocabulary as
+# validate-pr-create.sh / verify-commit-refs.sh: gh/glab "CLOSED", Linear
+# "Done", Jira "Resolved", Asana "Closed", etc.
+STATE_LC=$(echo "$STATE" | tr '[:upper:]' '[:lower:]')
+case "$STATE_LC" in
+  closed|done|cancelled|canceled|resolved|completed)
+    cat >&2 <<MSG
+BLOCKED: Active ticket ${TICKET_REPO}#${TICKET_NUM} is "$STATE", not open.
 Migration files require an OPEN labelled ticket. Run /migration to create
 a fresh one, or /start-ticket on a different OPEN migration ticket.
 MSG
-  exit 2
-fi
+    exit 2 ;;
+esac
 
 if [ "$HAS_LABEL" != "true" ]; then
   cat >&2 <<MSG
--- a/.claude/hooks/require-migration-ticket.sh
+++ b/.claude/hooks/require-migration-ticket.sh
@@ -263,6 +263,39 @@ MSG
   exit 2
 fi
 
+# --------- Shape-guard marker-derived values before the tracker call ---------
+# TICKET_NUM / TICKET_REPO come from the session-local active-ticket marker via
+# `cut -d= -f2-`, which keeps everything after the first `=` — so a marker line
+# like `number=42; touch X` yields TICKET_NUM='42; touch X'. Gate 2 below feeds
+# both values into `tracker_view`, which builds its command by string-substituting
+# {id}/{owner_repo} into a template and running `eval` (_lib-tracker.sh). Without a
+# shape check that is a command-injection sink: the base (pre-#755) hook used direct
+# argv (`gh issue view "$TICKET_NUM" --repo "$TICKET_REPO"`), which was
+# injection-immune; routing through the tracker abstraction reintroduces the eval.
+# The sibling caller `validate-pr-create.sh` reaches the same eval only AFTER its
+# ticket number passed a shape check — this hook must not skip the equivalent guard.
+# Whitelist a conservative charset (no shell metacharacters): ticket IDs like 42,
+# #42, GH-42, ABC-123; owner/repo slugs (incl. GitLab nested groups) of letters,
+# digits, `.`, `_`, `-`, `/`. Anything else fails closed. (#755 security review;
+# defence-in-depth alongside the printf %q quoting _tracker_substitute now applies.)
+if ! printf '%s' "$TICKET_NUM" | grep -qE '^#?[A-Za-z0-9_-]+$'; then
+  cat >&2 <<MSG
+BLOCKED: Active ticket marker at $MARKER has a malformed \`number=\` value.
+Only ticket IDs like 42, #42, GH-42, or ABC-123 are allowed (no shell
+metacharacters). Re-run /start-ticket to rewrite the marker cleanly.
+MSG
+  exit 2
+fi
+if ! printf '%s' "$TICKET_REPO" | grep -qE '^[A-Za-z0-9._/-]+$'; then
+  cat >&2 <<MSG
+BLOCKED: Active ticket marker at $MARKER has a malformed \`repo=\` value.
+Only owner/repo slugs (letters, digits, \`.\`, \`_\`, \`-\`, \`/\`) are
+allowed (no shell metacharacters). Re-run /start-ticket to rewrite the
+marker cleanly.
+MSG
+  exit 2
+fi
+
 # --------- Gate 2: issue is open + has migration label ---------
 # Resolve the ticket through the tracker abstraction (_lib-tracker.sh) so this
 # gate works for every configured tracker — GitHub (gh), GitLab (glab), Linear,
--- a/.claude/hooks/require-migration-ticket.sh
+++ b/.claude/hooks/require-migration-ticket.sh
@@ -362,7 +362,23 @@ MSG
 fi
 
 # --------- Gate 3: body references a migration AgDR ---------
+# The issue body is populated by tracker_view only for the gh and glab adapters
+# (see _lib-tracker.sh). For any other tracker kind the body comes back empty,
+# so this gate cannot see an AgDR link even when one exists — surface that in the
+# failure message rather than blaming the ticket. Adopters on those trackers can
+# set tracker.kind=none to skip online migration verification. (Widening body to
+# jira/linear/asana is tracked separately from #755.)
 if ! echo "$BODY" | grep -qE 'docs/agdr/AgDR-[0-9]+-[^[:space:]]*migration[^[:space:]]*\.md'; then
+  KIND_NOTE=""
+  case "$TICKET_KIND" in
+    gh|glab) ;;
+    *) KIND_NOTE="
+
+NOTE: tracker.kind=${TICKET_KIND} — this gate reads the issue body only for the
+gh and glab trackers, so for ${TICKET_KIND} the body is not fetched and this
+check cannot see an AgDR link even if the ticket has one. Track migrations on a
+gh/glab ticket, or set tracker.kind=none to skip online migration verification." ;;
+  esac
   cat >&2 <<MSG
 BLOCKED: Active ticket ${TICKET_REPO}#${TICKET_NUM} has the
 \`$MIGRATION_LABEL\` label but its body does not reference a migration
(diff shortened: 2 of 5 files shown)


AbdElrahmaN31 (commented):
## Code Review: PR #760

**Commit**: `d55f77e82b73abcd022f1eacd794a5734cdfe250`

### Summary
Routes the migration gate (`require-migration-ticket.sh` Gate 2/3) through the `_lib-tracker.sh` abstraction instead of a hardcoded `gh issue view`, fixing #755 where GitLab-tracked (`tracker.kind: glab`) projects false-blocked every migration-file edit. Adds `glab` as a first-class view kind (new `_tracker_normalise_glab` adapter + kind-aware built-in `view_command` default), extends the normalised `tracker_view` contract with a `body` field (gh + glab only), unpins `.tracker.view_command` in defaults so the kind-aware default can fire, and makes the closed-state check tracker-agnostic. New 11-case test file + 2 lib smoke cases.

### Checklist Results
- Testing:                   Pass (77/77 suite green)
- Security:                  Pass (see nit 1)
- Technical Decisions (AgDR): N/A — fills already-decided abstraction (AgDR-0033 / AgDR-0072); no new decision
- Adopter Handbooks:         N/A — no handbooks loaded

### Issues Found
None blocking.

**nit 1 — new `eval` exposure for marker-derived values.** Gate 2 now routes `TICKET_NUM`/`TICKET_REPO` (parsed verbatim from the active-ticket marker's `number=`/`repo=` lines) through `tracker_view` → `_tracker_substitute` → `eval "$cmd"`. The sibling callers (`verify-commit-refs.sh`, `validate-pr-create.sh`) feed `tracker_view` a value already shape-constrained by a `#[0-9]+` extraction regex; this hook does not shape-check the marker values at the call site. Real severity is low — exploitation requires local write to `.claude/session/`, and `/start-ticket` constrains the marker upstream against `id_pattern` — so this is a note, not a blocker. Optional hardening: a `tracker_id_pattern` shape-check on `TICKET_NUM` before the `tracker_view` call.

**nit 2 — Gate 3 message is misleading for jira/linear/asana.** Because `body` is deliberately scoped to gh/glab, a jira/linear/asana migration ticket that *does* link an AgDR still fails Gate 3 with "body does not reference a migration AgDR" — which is false; the framework simply didn't populate `body` for that kind. Not a regression (those trackers were already unsupported by the pre-#755 gh-hardcoded gate) and `tracker.kind=none` is the documented escape hatch, so this is a nit. Consider having the Gate 3 error acknowledge that non-gh/glab bodies aren't read, so the scoping limitation is visible to the operator.

**nit 3 — doc wording.** The contract comment says non-gh/glab adapters "emit `body:\"\"`", but the linear/jira/asana adapters emit no `body` key at all (consumers rely on `.body // empty`). Functionally identical; wording could match.

### Notes on the specific review asks
- **Fail-closed-on-empty:** sound and intentional. Documented in the PR body, the hook comments, and tested (cases 8 gh + 9 glab). Stricter than the #501 shape-only fallback is the right call for a high-blast-radius migration edit, and `kind=none` gives adopters who genuinely can't query a tracker an escape hatch.
- **`body` scoped to gh/glab:** sound. jira `.description` is ADF (a JSON object, not grep-able); linear/asana have no consumer. See nit 2 for the one rough edge.
- **Closed-state vocabulary:** verified byte-for-byte parity with `validate-pr-create.sh` / `verify-commit-refs.sh` (`closed|done|cancelled|canceled|resolved|completed`). Note the deliberate loosening from `!= "OPEN"` to an explicit closed-set correctly lets non-gh intermediate states (Linear "In Progress", etc.) pass — an empty/unknown state now falls through to the label/AgDR gates, which is the intended tracker-agnostic behaviour.
- **Unpinning `view_command`:** safe — the only non-lib/non-test reference is a comment in `verify-commit-refs.sh`; the lib now owns the kind-aware default.
- **AgDR:** correctly none. This is a bug fix completing the tracker abstraction (AgDR-0033 / AgDR-0072, both present in `docs/agdr/`), not a new technical decision.
- **Base branch:** `dev` — correct for framework-repo work (release-cut model).

### Test adequacy
Strong. 11 hook cases (gh+glab happy, none-skip, missing-label, closed gh+glab, missing-AgDR, unfetchable gh+glab fail-closed, non-migration passthrough, no-marker) + 2 lib smoke cases (opened→OPEN/web_url→url/description→body; closed→CLOSED). Optional add: a jira/linear case asserting the body="" → Gate 3 block behaviour so nit 2's edge is pinned.

### Verdict
**APPROVED**

---
Reviewed commit: `d55f77e82b73abcd022f1eacd794a5734cdfe250`

AbdElrahmaN31 (commented):
## Code Review (re-review): PR #760

**Commit**: `154a4f228dc20e700474e8b1ba5f39373844e681` (was `d55f77e`)

### Delta reviewed (commit 154a4f2)
Follow-up addressing two of the three nits from the prior pass:
- **nit #2 (RESOLVED)** — Gate 3's "no AgDR in body" block now appends a conditional `KIND_NOTE` when `TICKET_KIND` is not gh/glab, explaining the body isn't fetched for that tracker kind and pointing to `tracker.kind=none`. The message no longer blames the ticket for a framework scoping limitation. Verified the note only fires for non-gh/glab (the `gh|glab) ;;` arm is a no-op).
- **nit #3 (RESOLVED)** — contract doc now says non-gh/glab adapters omit the `body` key (read as `.body // empty`) instead of "emit body:''". Matches actual adapter behaviour.
- **New Case 12** — jira ticket ("In Progress" + `migration` label, no body) correctly passes the state and label gates, reaches Gate 3, blocks (exit 2), and asserts `tracker.kind=jira` is present in the note. I confirmed the `_tracker_normalise_jira` adapter maps `.fields.labels` and leaves `state`/`body` such that the case exercises exactly the intended edge.

### nit #1 (tracker_id_pattern shape-check before eval) — deferral accepted
I agree with not addressing it here. The `eval` path pre-exists in `tracker_view` for all callers, exploitation requires local write to `.claude/session/` (with `/start-ticket` constraining the marker upstream), and hardening it is out of #755's scope. My original finding already called it "a note, not a blocker." If pursued later it belongs in a dedicated hardening ticket touching the shared `tracker_view` path, not this fix.

### Verification
- `bash bin/run-hook-tests.sh` → **77/77 files PASS**.
- `test_require_migration_ticket.sh` → **12/12 cases PASS** (incl. new Case 12).
- Commit SHA `154a4f2` matches PR HEAD on GitHub.

### Verdict
**APPROVED**

---
Reviewed commit: `154a4f228dc20e700474e8b1ba5f39373844e681`

AbdElrahmaN31:
Follow-up filed: #761 — widen `tracker_view` `body` support to jira/linear/asana so the migration gate works for those trackers too (deliberately scoped to gh/glab here).

AbdElrahmaN31:
## Merge order note

**Independent** — no ordering dependency on the other open PRs (#762 / #766 / #768); merge any time.

Minor overlap caveat: this PR and #762 both edit `.claude/hooks/_lib-tracker.sh` (different functions), and this PR and #768 both edit `.claude/project-config.defaults.json` (different keys). Different regions, so likely clean — but whichever of an overlapping pair merges second may need a trivial rebase if git flags a conflict. Ping me and I'll push it.

atlas-apex (commented):
## Code Review: PR #760

**Commit**: `154a4f228dc20e700474e8b1ba5f39373844e681`

### Summary
Routes the migration gate's Gate 2 ticket lookup through `_lib-tracker.sh`'s `tracker_view` instead of a hardcoded `gh issue view`, so `require-migration-ticket.sh` works for GitLab (`glab`) and any other configured tracker rather than false-blocking every migration edit on a non-GitHub project (#755). Adds a first-class `_tracker_normalise_glab` adapter + `glab)` dispatch, a kind-aware built-in `view_command` default, `body` on the gh/glab normalised contract, tracker-agnostic closed-state detection, a `tracker.kind=none` short-circuit, and unpins the deep-merge-poisoning `view_command` from `project-config.defaults.json`. Ships a 12-case hook test file plus 2 glab lib smoke cases.

The design is well-reasoned and the writeup is excellent. The core #755 fix is correct and the abstraction is clean. **However, one blocking security regression must be addressed before merge**, because this is a security gate.

### Checklist Results
- Testing:                  Pass (thorough; but see the missing injection case)
- Security:                 **Fail** — command-injection regression (below)
- PR Description & Glossary: Pass (narrative summary + full glossary)
- Technical Decisions (AgDR): N/A — extends the existing AgDR-0072 tracker abstraction; no new AgDR needed. Agree with the author.

### Issues Found

#### ⛔ BLOCKING — command injection via the active-ticket marker into `eval` (security regression)

The base hook fetched the ticket with **direct, quoted argv**, which is injection-immune:

```sh
ISSUE_JSON=$(gh issue view "$TICKET_NUM" --repo "$TICKET_REPO" --json state,labels,body ...)
```

The new primary path (`require-migration-ticket.sh:297`) routes marker-derived values through the tracker abstraction:

```sh
ISSUE_JSON=$(tracker_view "$TICKET_NUM" "$TICKET_REPO" 2>/dev/null)
```

`tracker_view` builds the command by **string-substituting `{id}`/`{owner_repo}` into a template and running `eval "$cmd"`** (`_lib-tracker.sh` `_tracker_substitute` → `tracker_view:450 raw=$(eval "$cmd")`). `TICKET_NUM` and `TICKET_REPO` come straight from the marker file with **no shape validation**:

```sh
TICKET_REPO=$(grep -E '^repo=' "$MARKER" | head -1 | cut -d= -f2-)
TICKET_NUM=$(grep -E '^number=' "$MARKER" | head -1 | cut -d= -f2-)
```

`cut -d= -f2-` keeps everything after the first `=`, so a marker line `number=42; touch /tmp/PWNED` yields `TICKET_NUM='42; touch /tmp/PWNED'`, which `eval` executes.

**Reproduced against this PR's HEAD** (isolated sandbox, real hook + libs, gh stub):

```
marker: number=42; touch $D/PWNED_INJECTION
hook exit: 2
!!! COMMAND INJECTION CONFIRMED via marker 'number' field -> eval in tracker_view
```

The file `PWNED_INJECTION` was created — arbitrary command execution during a *security* gate.

Why this matters here specifically:
- It is a **regression**: the base was injection-immune (direct argv); this PR introduces the `eval` surface.
- The sibling tracker consumers **guard this exact `eval`** by shape-validating first: `validate-pr-create.sh` only reaches `tracker_view` after the ticket number passed the `tracker_id_pattern` regex ("the title already passed the shape check against tracker_id_pattern"). `require-migration-ticket.sh` is the outlier that omits the guard.
- The marker is written by `/start-ticket` from agent-supplied input, and its shape re-check is **advisory prose in the skill**, not a mechanical gate — so in an agentic/prompt-injection threat model the `number` field is not trustworthy for `eval`.
- This is the framework repo — it ships to every adopter.

**Suggested fix** (cheap, ~3 lines, restores fail-safe): validate the marker fields before the tracker call and block on a mismatch, e.g.

```sh
if ! printf '%s' "$TICKET_NUM" =~ ... ; then block ; fi   # allow only #?[A-Za-z0-9_-]+
# TICKET_REPO: allow only ^[A-Za-z0-9._/-]+$
```

Reuse `tracker_id_pattern` for the number if you want parity with the other hooks, and a conservative `owner/repo` charclass for the repo. (The `gh`-fallback at :302 is already argv-safe; only the `tracker_view` path needs the guard.) Please also add a hook-test case asserting a metacharacter-laden `number`/`repo` marker is blocked and does **not** execute — the current 12 cases don't cover it.

### Non-blocking notes

- nit: GitLab's issue `state` is only `opened`/`closed` — `locked` is a separate `discussion_locked` boolean, not a state value, so the `else ascii_upcase` branch and the "locked" concern are effectively moot. A `locked`-shaped state would fall through the closed-state list and be treated as open, which is acceptable (locked ≠ closed). No change needed; just confirming the mapping is fine.
- suggestion: the `TICKET_KIND` defaults to `gh` when `tracker_kind` is unavailable — good fail-toward-enforcement default. Worth a one-line comment noting that's deliberate.
- The gh path is behaviorally compatible for existing consumers (the added `,body` is additive; state/title/url/labels reads unchanged) and the defaults.json unpin correctly removes the deep-merge poisoning. The compatibility note about explicitly-pinned old `view_command` regressing Gate 3 is accurately documented.

### Verdict
**CHANGES REQUESTED**

The #755 fix itself is correct and the abstraction work is high quality — this is one focused security fix away from approval. Add input-shape validation on the marker-derived `TICKET_NUM`/`TICKET_REPO` before `tracker_view` (mirroring `validate-pr-create.sh`) plus a covering test, and I'll re-review.

---
📌 Reviewed commit: `154a4f228dc20e700474e8b1ba5f39373844e681`

atlas-apex (commented):
## Security Review: PR #760

**Commit**: `154a4f228dc20e700474e8b1ba5f39373844e681`

▸ Activating Hakim (Security Auditor) for PR #760 (trigger: security review requested; diff routes marker-derived input into a shell `eval`)

### Verdict: CHANGES REQUESTED — BLOCKING

I independently **CONFIRM** the command-injection / arbitrary-code-execution regression Rex flagged. I reproduced it two ways (isolated substitution PoC + end-to-end through the real hook). This is a shell-injection sink in a security-enforcement hook and must not merge as-is.

---

### 1. Exact data flow (CONFIRMED — zero validation before `eval`)

`require-migration-ticket.sh` @ HEAD:

- **L252–253** — marker fields parsed with no shape check:
  ```sh
  TICKET_REPO=$(grep -E '^repo=' "$MARKER" | head -1 | cut -d= -f2-)
  TICKET_NUM=$(grep -E '^number=' "$MARKER" | head -1 | cut -d= -f2-)
  ```
  `cut -d= -f2-` takes **everything after the first `=`**, so `number=42; touch X` yields `TICKET_NUM='42; touch X'`.
- **L255** — the *only* check is non-empty (`[ -z "$TICKET_REPO" ] || [ -z "$TICKET_NUM" ]`). No numeric check, no `isSafeRepo`, no `tracker_id_pattern` match.
- **L297** — passed straight in: `ISSUE_JSON=$(tracker_view "$TICKET_NUM" "$TICKET_REPO" 2>/dev/null)`.

`_lib-tracker.sh`:

- **L272–273** (`_tracker_substitute`) — **raw, unquoted** string substitution:
  ```sh
  tpl="${tpl//\{id\}/$id}"
  tpl="${tpl//\{owner_repo\}/$owner_repo}"
  ```
- **L450** (`tracker_view`) — `raw=$(eval "$cmd" 2>/dev/null)`.

So both marker fields (`{id}` and `{owner_repo}`) reach `eval` with **no validation and no quoting** at any point.

**Comparison to `validate-pr-create.sh` (Rex is correct):** that hook reaches the *same* `tracker_view`→`eval`, but its `TICKET_NUM` is extracted from the **PR title, which was already shape-checked against `tracker_id_pattern`** before dispatch — the hook's own comment (L258) says *"the title already passed the shape check against `tracker_id_pattern`, which is all we can assert."* `require-migration-ticket.sh` has **no equivalent gate** between marker read and `eval`. That asymmetry is the whole bug.

**Regression:** the base hook used injection-immune direct argv — `gh issue view "$TICKET_NUM" --repo "$TICKET_REPO" --json …` (quoted positional args, never re-parsed by a shell). This PR replaces that safe call with the template→`eval` path, reintroducing shell injection that the argv form structurally prevented.

### 2. The `eval` and substitution (CONFIRMED raw)

`eval` is real (`_lib-tracker.sh:450`) and the substituted `{id}`/`{owner_repo}` tokens are **not** shell-quoted or escaped (`:272–273`). The lib's own header comment (~L583) treats `eval` as "deliberate" for `tracker_create` (arbitrary create command) — but that reasoning does **not** transfer to `tracker_view`: the gh/glab view templates are **fixed built-ins**, and the `id`/`owner_repo` inputs are **untrusted** regardless of kind. An arbitrary *template* being intentional does not make arbitrary *argument values* safe.

### 3. Severity + threat model

**Reachability is real, not just hand-crafted.** The active-ticket marker (`.claude/session/current-ticket` / `tickets/<project>/…`) is plaintext, on-disk session state written by `/start-ticket` — an **LLM-driven markdown skill, not a validating script** (SKILL.md L183–184 writes `repo=<owner/repo>` / `number=<number>` verbatim from an agent-parsed ticket ref, with no shape enforcement). Attacker-influenced content reaches the marker via ordinary paths:

- an agent parsing a ticket ref out of attacker-controlled context (a malicious issue title/number, a pasted link, a repo/branch name) and writing it into the marker;
- any earlier tool step, repo setup script, or compromised dependency in the working tree that can write the session dir — which is exactly the adversarial context a security hook is supposed to withstand.

**Crucially, the payload executes regardless of the gate verdict.** In my E2E run the hook returned exit 2 ("unfetchable, blocked") yet the injected `touch` had **already executed** inside `eval`. An attacker does not need the gate to pass — merely triggering any Edit/Write to a migration path fires the injected command.

I rate this **HIGH** on the classic exposure axis (input is a local marker, not a network boundary) and **BLOCKING** without qualification, because it converts a defense-in-depth safety control into an arbitrary-code-execution primitive and is a regression from an injection-immune baseline. Rex's CRITICAL framing is defensible given the RCE outcome; HIGH vs CRITICAL is a labelling nuance — either way it does not merge.

### Proof of concept (reproduced)

Malicious marker:
```
repo=test-org/test-repo
number=42; touch $SB/OWNED_BY_MARKER ;
```
Triggering any migration-path Edit/Write ran `require-migration-ticket.sh`, and `eval "$cmd"` executed the injected `touch` — `OWNED_BY_MARKER` was created; the hook still exited 2. Both `{id}` (from `number=`) and `{owner_repo}` (from `repo=`, e.g. `x/y; touch PWNED #`) are independently exploitable.

### 4. Minimal correct fix (defense in depth — do both)

**(a) Caller — validate marker fields before `tracker_view` (required):** right after L253, block on shape mismatch (mirrors the existing empty-field block):
```sh
# gh/glab numeric id; general tracker_id_pattern for prefixed IDs
case "$TICKET_NUM" in *[!0-9]*)
  echo "BLOCKED: malformed ticket number in marker: $TICKET_NUM" >&2; exit 2 ;;
esac
case "$TICKET_REPO" in
  */*) : ;; *) echo "BLOCKED: malformed repo in marker" >&2; exit 2 ;;
esac
# isSafeRepo: single owner/name segment, safe charset only
printf '%s' "$TICKET_REPO" | grep -qE '^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$' \
  || { echo "BLOCKED: unsafe repo slug in marker: $TICKET_REPO" >&2; exit 2; }
```
(For non-`#`/numeric trackers, validate against `tracker_id_pattern` instead of `*[!0-9]*`.)

**(b) Library — stop trusting substituted tokens (required, protects every current + future caller):** `tracker_view`/`_tracker_substitute` must not feed unquoted user input to `eval`. Preferred: for the built-in `gh`/`glab` kinds, **bypass the template and call argv arrays directly** — `gh issue view "$id" --repo "$owner_repo" --json …` — reserving `eval` only for the genuinely arbitrary `custom` view_command. If the string-template engine is kept, `printf %q`-quote `$id` and `$owner_repo` before substitution, and still validate their shape. Removing `eval` in favour of argv arrays for the built-ins is the durable fix; the `custom` path should at minimum quote the injected tokens.

Fix (a) closes the immediate regression; fix (b) ensures the next caller that forwards untrusted input into `tracker_view` can't reopen the same hole.

---

### Checklist Results
- Injection Prevention:   **FAIL** — shell command injection via marker `number=`/`repo=` → `_lib-tracker.sh:eval` (HIGH, BLOCKING)
- XSS Prevention:         N/A
- Auth & Authorisation:   Pass
- API Security:           N/A

Aligning with Rex: this is a hard blocker. Do not merge until both the caller-side field validation and the library-side quoting/argv fix land, with a regression test asserting a marker like `number=42; touch X` neither executes nor passes the gate.

---
📌 Reviewed commit: `154a4f228dc20e700474e8b1ba5f39373844e681`

atlas-apex (changes requested):
Thank you for this, @AbdElrahmaN31 — genuinely nice work. 🙏

The direction is exactly right: routing the migration gate through `tracker_view` instead of a hardcoded `gh` call is the correct shape, the `_tracker_normalise_glab` adapter is clean, the kind-aware `view_command` default + the `defaults.json` unpin close a real deep-merge config bug, `tracker.kind=none` short-circuits correctly, and the 12-case hook test is thorough. This is most of the way there.

**Requesting changes on one blocking item — a security regression, so I'd rather catch it here than merge it.** The base hook used injection-immune direct argv (`gh issue view "$TICKET_NUM" --repo "$TICKET_REPO"`). The new path feeds the marker-derived `TICKET_NUM` / `TICKET_REPO` into `tracker_view`, which does `eval "$cmd"` after plain `{id}` / `{owner_repo}` substitution — and those values come from the active-ticket marker via `cut -d= -f2-` with no shape validation. Our review reproduced arbitrary command execution: a marker line like `number=42; touch …/PWNED` runs. Because this hook is itself a security gate, we can't ship an `eval` reachable from unvalidated input.

**The fix is small and you're 95% there:**
1. Validate the marker fields before `tracker_view` — numeric `number`, and `isSafeRepo`-style `owner/repo` — exactly how `validate-pr-create.sh` guards before it reaches the same `eval`.
2. A covering test (a marker with a shell-metachar `number`/`repo` must be rejected, not executed).
3. (Optional hardening) prefer argv arrays over `eval` in `_lib-tracker.sh` so the substitution can't be a code path at all.

Minor non-blocker: GitLab issue `state` is only `opened`/`closed`, so the `locked` mapping isn't needed.

Really appreciate you pushing the tracker-abstraction forward — happy to pair on the validation bit if useful. Thanks again! 🚀

AbdElrahmaN31:
Fixed in `86b439f` (was `154a4f2`).

Reproduced the RCE end-to-end first: a marker `number=42; touch …/PWNED ;` triggering any migration-path Edit → hook exits 2 (fail-closed) **yet the injected `touch` still ran**, confirming execution is independent of the gate verdict.

Two layers:

**1. Caller-side shape guard** (`require-migration-ticket.sh`, before the tracker call) — whitelist `TICKET_NUM` (`^#?[A-Za-z0-9_-]+$`) and `TICKET_REPO` (`^[A-Za-z0-9._/-]+$`), block (exit 2) on any shell metacharacter. This is the required fix and mirrors the shape guarantee `validate-pr-create.sh` / `verify-commit-refs.sh` already get via their digit-only extraction. Confirmed by grep that `require-migration-ticket.sh` was the **only** marker→`tracker_view` caller missing one.

**2. Library defence-in-depth** (`_lib-tracker.sh` `_tracker_substitute`) — `printf %q` the `{id}` / `{owner_repo}` tokens before substitution, so a future caller that forwards unvalidated input into `tracker_view` can't reopen the sink.

Three places I deviated from the literal suggestion, each deliberate:

- **`printf %q` rather than argv-arrays-for-gh/glab.** You both preferred argv arrays as the durable fix and listed `%q` as the sanctioned keep-the-engine alternative. I chose `%q` because it also covers the `custom` view kind — an argv refactor scoped to the gh/glab built-ins would leave `custom`'s templated `eval` on the raw path. `_tracker_substitute` has exactly one caller (`tracker_view`); `tracker_create` uses a separate substitution path, so the change is contained to the view path.
- **Number whitelist `^#?[A-Za-z0-9_-]+$` rather than numeric-only.** `number=` is tracker-agnostic — `GH-42`, `ABC-123` are legitimate, and `tracker_id_pattern` would reject a bare `42`. Security property is identical (no metacharacter survives). The one shell-meaningful char it admits is a leading `#`, which the lib's `%q` escapes (`\#42`); test case 15 pins that `#42` is handled safely rather than commenting out the rest of the command.
- **Repo charset allows multiple slashes** rather than a single `owner/repo`. GitLab nested groups (`group/subgroup/project`) are valid slugs; `/` is inert in command position.

Tests: +3 hook cases (metachar `number=` / `repo=` blocked **and not executed** — asserting file-absence, not just the exit code; `#`-prefixed number passes safely) and +1 lib smoke case (`tracker_view` neutralises a metachar id; legit id substitutes unchanged). Full suite **77/77**.

Left the GitLab `locked` mapping as-is — agree it's a moot defensive fallback, no behaviour change either way.

New commit invalidates the prior review SHA — ready for re-review.

atlas-apex (approved):
## Code Review (re-review): PR #760

**Commit**: `86b439f202e039a7aa0a46ea6258271fa42e61af`

▸ Activating Rex (Code Reviewer) for PR #760 (trigger: re-review after security fix `86b439f`)

Thanks for the fast, precise turnaround, @AbdElrahmaN31. The security regression we flagged is now **closed**, and I re-ran a fresh full pass. Approving. 🚀

### Verdict on the blocking finding — RCE / eval-injection is dead

The prior CHANGES_REQUESTED was for a command-injection sink: Gate 2 routes marker-derived `TICKET_NUM` / `TICKET_REPO` (parsed via `cut -d= -f2-`, unvalidated) into `tracker_view`, which `eval`s its substituted command (`_lib-tracker.sh:463`). PoC was `number=42; touch …/PWNED`.

I verified the fix **at the sink**, not just at one entry point:

- **Library defence-in-depth (the real sink fix)** — `_tracker_substitute` now `printf %q`-quotes both `{id}` and `{owner_repo}` before they reach `eval`. I attacked the library path directly (bypassing the caller guard) with the original PoC plus variants — `;`, backticks, `$()`, `&&`, `|`, and newline-split, on **both** the `{id}` and `{owner_repo}` tokens. Every one is neutralised: metacharacters are escaped (`42\;\ touch…`), and the newline case collapses to a single ANSI-C `$'42\ntouch…'` string arg, never a command separator. **No injected command executed in any variant.**
- **Caller-side shape guard** — `require-migration-ticket.sh` whitelists `TICKET_NUM` (`^#?[A-Za-z0-9_-]+$`) and `TICKET_REPO` (`^[A-Za-z0-9._/-]+$`) and blocks (exit 2) on any shell metacharacter, before the tracker call. End-to-end through the real hook (with a gh stub returning a valid OPEN+labelled+AgDR issue, so only the guard can stop execution): all metachar markers → **blocked, exit 2, not executed**; the `#`-prefixed legit id passes and is `printf %q`-safe; a stray extra marker line is inert because the marker is line-based `key=value`.

Two independent layers, and the library layer holds even if a future caller forwards unvalidated input. That is the correct shape for a security gate.

### Checklist Results
- ✅ Architecture & Design: Pass — routes through `_lib-tracker.sh`; `_tracker_normalise_glab` mirrors the existing create adapter; kind-aware `view_command` default + `defaults.json` unpin close the deep-merge config bug.
- ✅ Code Quality: Pass — shellcheck clean at actionable severity (only pre-existing `info` SC2295 at lines 191/199, outside the changed region; intentional SC2016 in single-quoted test stubs).
- ✅ Testing: Pass — full hook suite **77/77 PASS**. `test_require_migration_ticket.sh` 15/15 (gh + glab happy/closed/unfetchable, none-skip, missing-label, missing-AgDR, jira body-scoping note, injection number=/repo= blocked-and-not-executed, `#`-safe). `test_tracker_aware_hooks.sh` adds glab + library-injection smoke cases.
- ✅ Security: Pass — injection sink neutralised at library + caller layers (verified above). Note: this touches the trust chain (`.claude/hooks/**`) → flagged for Hakim's follow-up pass.
- ✅ Performance: Pass — no hot-path concern.
- ✅ PR Description & Glossary: Pass — narrative summary, explicit fail-closed design rationale, security-hardening section, complete glossary.
- ✅ Technical Decisions (AgDR): N/A — bugfix routing through an existing abstraction; fail-closed rationale documented inline and in the PR body. No new dependency/pattern warranting a fresh AgDR.

### Gate integrity — the fix did not weaken the gate
Verified the gate still BLOCKS everything it must: missing migration label → block; closed ticket (gh CLOSED + glab closed) → block; body missing the AgDR link → block; unfetchable ticket → block (fail-closed by design, stricter than the #501 existence checks). The shape guard adds a new block path (malformed marker) — so the gate is strictly stronger, never weaker. gh routing correct; glab degrades safely (`opened→OPEN`, `closed→CLOSED`, `web_url→url`, `description→body`) and is fail-closed on an unfetchable ticket.

### Diff scope
Clean — exactly 5 on-topic files (`_lib-tracker.sh`, `require-migration-ticket.sh`, two test files, `project-config.defaults.json`). No stray changes.

- Optional future hardening (already noted last round): argv arrays over `eval` in `_lib-tracker.sh` would remove the code path entirely, though `printf %q` closes the current hole.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `86b439f202e039a7aa0a46ea6258271fa42e61af`
🔒 Trust-chain change — flagged for Hakim (Security Auditor) follow-up pass.

atlas-apex (commented):
## Security Review: PR #760

**Commit**: `86b439f202e039a7aa0a46ea6258271fa42e61af`

Independent security pass on the re-worked eval path, following the earlier RCE block. This is a focused re-audit of the injection sink, the caller guard, and the gate's fail-closed posture — not a full re-review of the GitLab feature.

### Summary

The prior finding (marker-derived `number=` / `repo=` → unvalidated → `eval "$cmd"` in `tracker_view`) is **closed at the sink, not just the caller**. `_tracker_substitute` now `printf %q`-quotes both `{id}` and `{owner_repo}` before template substitution, so the `eval` in `tracker_view` (`_lib-tracker.sh:463`) can no longer be driven by injected shell syntax — regardless of what any caller forwards.

### Checklist Results
- Injection Prevention:   Pass (verified empirically at the sink)
- Auth & Authorisation:   Pass (gate fail-closed preserved)
- API Security:           N/A

### 1. RCE dead at the SINK — verified by direct attack

I attacked `_tracker_substitute` + the `eval` path **directly**, bypassing the `require-migration-ticket.sh` caller guard entirely, replicating `tracker_view`'s exact substitute→eval sequence with raw malicious input. Full matrix, both template tokens:

- `{id}`: `42; touch X`, the original PoC `number=42; touch X`, backticks `` `…` ``, `$(…)`, `&&`, `|`, `||`, `>` redirect, embedded newline, `#42; …`, brace-glob `{42,…}`, leading-dash `-x; …`
- `{owner_repo}`: `;`, backticks, `$(…)`, `&&`, `|`, embedded newline, plus a legit nested GitLab group

**Result: 0 executions. No `PWNED` file created for any variant.** Legit values substitute unchanged (`42`, `GH-42`, `ABC-123`, `JIRA-789`, `group/subgroup/project`); `#42` correctly becomes `\#42` (the shell unescapes it back to `#42` for the CLI, so the value is preserved *and* the `#`-starts-a-comment bug in the old unquoted template is fixed). The library is now injection-immune independent of the caller — the defense-in-depth requirement is met.

### 2. Caller shape-guard — sound, no bypass

`require-migration-ticket.sh` whitelists `TICKET_NUM` (`^#?[A-Za-z0-9_-]+$`) and `TICKET_REPO` (`^[A-Za-z0-9._/-]+$`) — both anchored, both metachar-free — and exits 2 on any violation. The guard sits on the **single linear path** between marker-parse (`grep … | head -1 | cut -d= -f2-`, lines 252–253) and every sink: `tracker_view` (330), the lib-missing fallback (335, direct-argv `gh` — also injection-safe), and the `tracker_kind` call (pure yq/config, no eval). There is no alternate marker-parse route to the sink, and `grep | head -1` guarantees single-line values so the anchored regex can't be newline-bypassed. A legit `#N` id passes the guard and is `%q`-safe.

### 3. Gate not weakened — fail-closed preserved

All existing block paths remain and fail closed: missing `migration` label → block; closed state for **both** gh (`CLOSED`) and glab (`closed`) via the tracker-agnostic `closed|done|cancelled|resolved|…` vocabulary → block; missing-AgDR body → block; **unfetchable ticket → block for both gh and glab** (deliberately stricter than the #501 shape-only fallback, correct for high-blast-radius edits). The only new *allow* path is `tracker.kind=none`, which is operator-configured and prints an audible stderr note. The shape-guard only **adds** a block path.

### 4. glab routing — fail-closed, no injection

`_tracker_normalise_glab` is pure `jq` over the CLI's JSON (response data is never eval'd), the glab `view_command` is a fixed built-in template with only the two `%q`-quoted tokens substituted, and an unfetchable glab ticket blocks (exit 2). No injection surface in the adapter.

### Verification performed
- Direct sink attack (19 variants, both tokens): **0 executions**
- `test_require_migration_ticket.sh`: **15/15 PASS** (incl. injection cases 13/14 and the `#42`-safe case 15)
- `test_tracker_aware_hooks.sh`: **26/26 PASS** (incl. the glab normalise + the lib `printf %q` injection case)

### Verdict

**APPROVED (security)** — the RCE is neutralised at the eval sink itself, the caller guard is a sound belt-and-braces second layer with no bypass, and the migration gate stays fail-closed. **Clear to merge from a security standpoint.** (Advisory only — no merge marker written; the code-review + CEO merge gates still apply.)

---
📌 Reviewed commit: `86b439f202e039a7aa0a46ea6258271fa42e61af`

Issue #763, opened 2026-07-02
[Feature] Route /security-review + /design-review through tracker_review_submit (gh/glab)


AbdElrahmaN31:
> **Scope narrowed (see #765).** This ticket originally bundled a second concern — the cross-fork approval-marker naming bug. That is an all-or-nothing change across four marker writers + the merge-gate contract (a different blast radius from prose routing), so it now lives in its own ticket: **#765**. This ticket is now **routing-only**.

## Problem

PR #762 (#758) added `tracker_review_submit` to `_lib-tracker.sh` — a tracker/git-host-agnostic PR/MR review-submission abstraction (gh / glab / custom / none) — and routed the **code-reviewer agent (Rex)** + `/code-review` through it. But two review-submitting agents still shell out to a **hardcoded `gh pr review`**, so on a GitLab-hosted (or custom-host) project their review-submission step silently assumes GitHub — the same gap #758 just closed for `/code-review`:

- **Hakim** — `.claude/agents/security-reviewer.md` (`/security-review`)
- **Tariq** — `.claude/agents/solution-architect.md` (`/design-review`), which additionally writes an architecture sign-off marker

## Proposed behaviour

Route `security-reviewer.md` (Hakim) and `solution-architect.md` (Tariq) through `tracker_review_submit` exactly as `code-reviewer.md` now is: source `_lib-tracker.sh`, resolve `$PR_HOST_REPO` (the PR/MR **base** repo, parsed from the PR URL — `gh pr view` has no `baseRepository` field), write the review to a temp body-file, and submit via the adapter for the project's `tracker.kind`. Keep each agent's local sign-off-marker write (Tariq's `*-architecture.approved`) untouched and orthogonal, and preserve the "prefer `comment` over `approve` on self-review" policy.

- Tariq's `/design-review` skill also threads a repo (split-portfolio v2, me2resh/apexyard#687) — the base-repo resolution reuses that same threaded repo, deriving `$PR_HOST_REPO` from it rather than introducing a second, divergent resolution.

## Why it helps adopters

A GitLab (or custom-host) adopter today gets full ticket-gating support but a **broken review-submission step** for security and design reviews — they must run Hakim's / Tariq's checklist by hand and post the review manually — even after #758 fixed the same thing for code review. Routing the two siblings closes the gap consistently.

## Scope / out of scope

[...]

PR #768, opened 2026-07-02
feat(#753): docs-index hook — auto-maintain per-project docs/INDEX.md + nudge

AbdElrahmaN31:
## Summary
- **New advisory PostToolUse hook keeps per-project docs findable** — `maintain-docs-index.sh` regenerates `projects/<name>/docs/INDEX.md` (a grouped, date-aware table of contents) whenever a doc under that project changes, and nudges when a `*.md`/`*.html` lands under the project but *outside* `docs/`. The admin dashboard only surfaces `projects/<name>/docs/`, so docs written elsewhere are invisible and there's no index of what exists — this closes that gap without manual curation.
- **Off by default (opt-in)** — unlike the pure-banner advisory hooks, this one writes a *tracked* file (`INDEX.md`) into the repo, so `docs_index.enabled` defaults **`false`**; adopters enable it in `project-config.json`. It never blocks (exit 0 always), same shape as `check-upstream-drift.sh` / `detect-role-trigger.sh`.
- **Git-based dates, subdir grouping** — created = the first commit that added the file, modified = the last commit (portable across clones, unlike filesystem mtime which resets on checkout); docs group by their immediate subdir under `docs/`, with root-level docs under a "General" group.
- **No self-recursion** — the index is written with a shell redirect (not the Write tool), so regenerating it does not re-fire the PostToolUse hook.

## Design defaults chosen
Three opinionated calls, all reversible via config / easy to change in review:
- **default OFF** (opt-in) — because the hook writes a tracked file into adopters' trees; "defaults safe" per the ticket AC.
- **git history for dates** — portable and meaningful across clones (vs unreliable filesystem mtime).
- **group by immediate `docs/` subdir** — root docs under "General".

## Testing
- `tests/test_maintain_docs_index.sh` — **22 assertions across 7 cases, all green**: disabled no-op, index generation (header + git-dated rows + title from first heading), outside-`docs/` nudge, non-project no-op, non-`md`/`html` skip, subdir grouping, and no self-recursion on an index-file write.
- `bash -n` clean on the hook; `settings.json` and `project-config.defaults.json` validate as JSON.
- Test note: the runner unsets `CLAUDE_CODE_SESSION_ID` + sets `APEXYARD_OPS_DISABLE_PIN=1` so the pin-aware ops-root resolver falls back to walk-up and reads the sandbox config rather than the live fork's.

## Files
- `.claude/hooks/maintain-docs-index.sh` — the hook
[...]

PR #766, opened 2026-07-02
feat(#763): route /security-review + /design-review through tracker_review_submit

AbdElrahmaN31:
## Summary
- **Routes the two sibling reviewer agents through the tracker abstraction** — `security-reviewer.md` (Hakim, `/security-review`) and `solution-architect.md` (Tariq, `/design-review`) now submit their review via `tracker_review_submit` (gh PR / glab MR / custom host) instead of a hardcoded `gh pr review`, closing the same GitLab/custom-host gap #758/#762 closed for `/code-review`. Without this, those two reviews silently assume GitHub and never post on a non-GitHub project.
- **Hakim gains the plumbing it lacked** — it previously had zero ops-root/lib resolution (just `gh pr review`); it now resolves the ops fork root pin-first, sources `_lib-tracker.sh`, derives `$PR_HOST_REPO` (the PR/MR **base** repo, parsed from the PR URL since `gh pr view` exposes no base-repo field), and submits from a temp body-file. It writes no gate marker (post-only), so it needs `_lib-tracker.sh` but not `_lib-review-markers.sh`.
- **Tariq's base-repo resolution composes with the existing #687 threading, not alongside it** — a new "Posting the review" section resolves `$MARKER_HOME` + `$REPO` + `$PR_HOST_REPO` once; the sign-off marker now *reuses* those values rather than re-resolving (which risked keying the marker on a different repo than the review was posted to). `$PR_HOST_REPO` derives from the same threaded repo, diverging only in the `headRepository` fallback.
- **Architecture-marker keying is deliberately unchanged** — the cross-fork marker-naming bug (writers key on the fork, gates key on the base repo) is all-or-nothing across four writers + the merge-gate contract, so it is split to #765; an inline note in Tariq's marker section points there.

## Testing
- Prose-only change (no library modification — `tracker_review_submit` shipped host-complete in #762).
- Inherited hook tests green on this branch: `test_tracker_review_submit.sh` **27/27**, `test_tracker_create.sh` **16/16**.
- Both edited agent files' fenced bash blocks pass `bash -n`.

## Stacked on #762 (#758)
[...]

PR #762, opened 2026-07-02
feat(#758): route /code-review through tracker-agnostic review submission

AbdElrahmaN31:
## Summary

- **New `tracker_review_submit` abstraction in `_lib-tracker.sh`** — closes the last hardcoded-`gh` surface in the review flow. It kind-dispatches PR/MR review posting the same way `tracker_create` does for issue creation (AgDR-0072): built-in **gh** + **glab** adapters, a `custom` `review_command` template, and a `kind=none` no-op that returns 3 and echoes the review body for manual filing. Adopters on GitLab (or a custom host) previously got a broken "submit review" step even though Rex's analysis ran fine — this is the same gap #670 closed for issue *creation*.
- **Verdict vocabulary is `approve | comment | request-changes`** (default `comment`), a faithful map of gh's three verbs. The "prefer `comment` over `approve` on single-account self-review" *policy* stays in the code-reviewer agent (where it already lived); the lib is a thin, un-opinionated wrapper. GitLab has no `request-changes` review state, so that verdict posts an MR **note** carrying the verdict-in-body; `approve` posts the approval and attaches the body as a note (glab's approve carries no message).
- **Routed the code-reviewer agent (Rex) + `/code-review` skill through it** — the review now writes to a temp body-file and submits via the adapter for the project's configured `tracker.kind`, instead of shelling out to `gh pr review` directly. The gh happy path is byte-for-byte unchanged for the default adopter.
- **The local `*-rex.approved` merge-gate marker is untouched and explicitly orthogonal** — it is a plain local file (already tracker-agnostic). A failed *submit* warns loudly and preserves the full review body in Rex's report, but does **not** block the marker write on an APPROVED verdict: the review *was performed*, the marker is the gate signal, and a transient host-CLI failure must not block a legitimate merge. This contract is now spelled out in the agent rather than left to inference.

## Testing

- New `test_tracker_review_submit.sh` — **25 cases, all green**: gh verb selection (`approve`/`comment`/`request-changes`) + `--body-file` passing + unknown-verdict normalisation + non-zero-exit propagation (proving `2>/dev/null` doesn't mask the status); glab `mr approve` / `mr note create` dispatch incl. the request-changes→note and approve+body→two-calls mappings; `custom` injection-safety (review body passed via `$TRACKER_REVIEW_BODY_FILE` env, never string-substituted — a hostile body creates no `PWNED`); `kind=none` shape-only (exit 3 + body echo).
[...]

PR #770, opened 2026-07-03
fix(#765): key all approval markers on the PR base repo (cross-fork)

### LINKED ISSUE #765: [Bug] Cross-fork approval markers keyed on headRepository (fork) but merge gates key on the base repo — approved PRs blocked
## Problem

On a **cross-fork** PR (origin = a fork, base = the host repo — e.g. this framework's own `fork → me2resh/apexyard:dev` contribution flow), the approval-marker **writers** and the merge-gate **readers** key the marker filename on **different repos**, so a legitimately-approved PR stays blocked until the marker is renamed by hand.

- **Writers key on `headRepository` (the fork):** `code-reviewer.md` (REX marker), `solution-architect.md` (ARCHITECTURE marker), `/approve-merge` (CEO marker), `/approve-design` (DESIGN marker) all resolve the marker repo via `gh pr view <n> --json headRepository`.
- **Readers (gates) key on the merge command's base repo:** `block-unreviewed-merge.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh` derive the repo via `extract_repo_from_command` — the `--repo` value of `gh pr merge --repo <base>` or the `gh api repos/<base>/pulls/<n>/merge` path (both the **base** repo, since the PR lives on the base).

On a same-repo PR `base == head`, so everything agrees and this is invisible. On a **cross-fork** PR they diverge: the marker is written at `<fork>__repo__<n>-<role>.approved` but the gate looks up `<base>__repo__<n>-<role>.approved` → not found → merge blocked after a valid approval.

This was hit for real on PR #762 (the #758 code-reviewer routing change), which had to work around it by writing the rex/ceo markers under the base-repo qualifier manually.

## Proposed behaviour

Key **every** approval marker on the PR's **base (host) repo** — the same slug the gates already derive — so writers and readers agree on cross-fork PRs. The base repo is the stable identity of a PR (it's where the PR is numbered); the fork is incidental.

- Resolve the base repo the same robust way #758/#762 established for review *submission*: parse it from the PR URL (`gh pr view <n> --json url` — always base-rooted; `gh pr view` exposes no `baseRepository` field), falling back to `headRepository` when base == head.
- Apply to **all four writers together** (`code-reviewer.md`, `solution-architect.md`, `/approve-merge`, `/approve-design`). **This is all-or-nothing:** fixing only some writers leaves the others (e.g. the CEO or design marker) keyed on the fork, so a cross-fork merge stays blocked — a partial fix is worse than none because it looks fixed but isn't.
- Consider centralising base-repo resolution in a shared helper (e.g. alongside `_lib-extract-pr.sh` / `_lib-review-markers.sh`) so the writer and reader sides can never drift again.

## Why it helps adopters

Unblocks the fork → upstream contribution workflow — including this framework's own `dev` PRs, where an approved review otherwise fails the merge gate and forces a rule-adjacent manual marker rename on every cross-fork PR. Same-repo adopters are unaffected (base == head), so this is a strict fix with no behaviour change for the common case.

## Scope / out of scope

- **In scope:** align all four marker writers to key on the base repo; a shared base-repo resolver; **regression tests that cover both the same-repo (unchanged) and cross-fork (fixed) paths** for each of the three gates.
- **Out of scope:** the review-*submission* routing (already shipped in #758/#762 and extended to the sibling reviewers in the routing PR that closes #763 — this ticket is only about the marker *filename* key, a separate concern from where the human-visible review is posted).

## Severity

Bug — confirmed, reproduced on PR #762. Higher-priority than the routing follow-up, since it breaks an already-in-use workflow rather than a not-yet-hit adopter gap.

## Framework version

v4.2.0 (latest upstream release; this fork's CHANGELOG.md lags at 4.1.0). Related unreleased work: PR #762 (#758) established the base-repo-from-URL resolution pattern this fix reuses.

## Related

- Surfaced during Rex review of PR #762 (#758).
- Sibling of #763 (the routing follow-up), which explicitly deferred this marker fix to "its own ticket."
- Touches `.claude/agents/code-reviewer.md`, `.claude/agents/solution-architect.md`, `.claude/skills/approve-merge/SKILL.md`, `.claude/skills/approve-design/SKILL.md`; tests under `.claude/hooks/tests/`.

## Glossary

| Term | Definition |
|------|------------|
| approval marker | `.claude/session/reviews/<owner>__<repo>__<pr>-<role>.approved` — the local file a merge gate reads to confirm a review happened (AgDR-0060). Roles: rex / ceo / architecture / design. |
| cross-fork PR | A PR whose head branch is on a fork and whose base is a different (upstream) repo — the PR "lives on" the base. |
| `headRepository` vs base | `gh pr view` exposes `headRepository` (the fork) but no `baseRepository`; the base repo must be parsed from the PR URL, which is always base-rooted. |
| all-or-nothing | Unblocking a cross-fork merge requires every writer to agree with the gate's base-repo key; a partial fix leaves some markers diverged and the merge still blocked. |

AbdElrahmaN31:
## Summary
- **Fixes the cross-fork approval-marker divergence** — all **five** marker writers now key on the PR's **base** repo, matching what the merge gates already look up. Writers previously keyed on `headRepository` (the fork), so on a cross-fork PR a valid approval was written under the fork qualifier while the gate searched under the base → the merge stayed blocked, forcing a manual marker rename on #762 / #766 / #768. (Rex's review of this PR caught a fifth writer — `/approve-architecture` — that the first pass missed; commit `d8ad532` completes the set.)
- **New `pr_base_repo <pr> [hint]` in `_lib-review-markers.sh`** — parses the PR URL (gh pr view exposes no baseRepository field), falling back to the hint when base == head, so **same-repo PRs are a provable no-op** (the regression suite is unchanged). Makes writer/reader agreement structural, not coincidental.
- **`/approve-merge` fixed end-to-end** — it is both a marker writer (ceo) and the merge-runner, so the CEO write, the Rex read, AND the merge `--repo` all move to the base; otherwise a cross-fork merge command fails before markers even matter.
- **Gates untouched** — they already key on the base via the merge command's `--repo` / API-path (you cannot merge a fork's copy); a comment documents the residual no-`--repo` fallback edge, which affects unsanctioned manual merges only. A separate commit unifies `solution-architect.md` on the pin-first ops-root resolver (Rex's non-blocking nit from #766).

## Testing
- `test_pr_base_repo.sh` — **9/9**: cross-fork → base, **same-repo → unchanged**, GitLab nested MR, GHE host, gh-fail / bad-URL / no-pr → hint fallback, and marker-path discrimination.
- Regression: **`block_unreviewed_merge` 29/29** (same-repo no-op holds), tracker suites 27/27 + 16/16 unchanged; all three gate hooks and both edited agents' bash blocks are `bash -n` clean.

## Stacked on #766 (→ #762)
This branch is stacked on `feature/GH-763` (PR #766, itself on #762). Because the base here is `dev`, the diff currently **includes #762's and #763's changes**; it reduces to #765's own commits once the chain lands. **Review scope for this PR is the three #765 commits** — `81b63c3` (base-repo keying), `f11415c` (pin-first unification), and `d8ad532` (the 5th writer, `/approve-architecture`). Merge order: **#762 → #766 → #765**.

Closes #765

## Glossary
| Term | Definition |
|------|------------|
| `pr_base_repo` | New helper resolving a PR's **base** repo (the repo it lives on / is numbered against) from the PR URL, with a hint fallback. The canonical key for approval markers. |
| base vs head repo | On a cross-fork PR the head branch is on a fork; the PR lives on (and can only be merged against) the base repo. `gh pr view` exposes `headRepository` but no `baseRepository`. |
| marker writer / reader | Writers = code-reviewer (rex), solution-architect (architecture), /approve-merge (ceo), /approve-design (design), /approve-architecture (architecture). Readers = the three merge-gate hooks. This PR makes all writers key on the base, matching the readers. |

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -37,7 +37,7 @@ tracker_review_submit "$PR_HOST_REPO" {number} comment "$REVIEW_BODY_FILE"
 
 The verdict that drives the merge gate is the **local marker**, NOT the host's "Approved" review state. So:
 
-- **Canonical happy path:** call `tracker_review_submit "$PR_REPO" {number} comment "$REVIEW_BODY_FILE"` and state the verdict (`APPROVED` / `CHANGES REQUESTED`) in the body itself. This always works — on gh it maps to `gh pr review --comment`; on glab to an MR note; on custom to the operator's `review_command`.
+- **Canonical happy path:** call `tracker_review_submit "$PR_HOST_REPO" {number} comment "$REVIEW_BODY_FILE"` and state the verdict (`APPROVED` / `CHANGES REQUESTED`) in the body itself. This always works — on gh it maps to `gh pr review --comment`; on glab to an MR note; on custom to the operator's `review_command`.
 - **Do NOT pass the `approve` verdict by default.** On gh it maps to `gh pr review --approve`, which in the common single-account / auto-mode setup GitHub refuses ("Cannot approve your own PR"), and an auto-mode write-classifier may additionally flag it. **That block is expected and is not a failure** — a host "Approved" state is optional and unavailable when reviewing your own account's PR. Do not retry it, do not escalate it, and do not report the review as incomplete because of it. The local marker (output #1) is what satisfies the gate.
 - The `request-changes` verdict is fine for a non-approving result you want reflected in the host's review state (on gh it does not hit the self-approval restriction; on glab it posts a note, since GitLab has no request-changes state).
 
@@ -661,19 +661,25 @@ MARKER_HOME="${OPS_ROOT:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}"
 # shellcheck source=/dev/null
 . "$MARKER_HOME/.claude/hooks/_lib-tracker.sh"
 mkdir -p "$MARKER_HOME/.claude/session/reviews"
-# Resolve the head (fork) repo — used for the qualified marker filename.
+# Resolve the head (fork) repo — used ONLY as the hint/fallback for the base
+# resolver below. It is NO LONGER the marker key (that was the cross-fork bug —
+# see #765).
 PR_REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
-REX_MARKER=$(review_marker_path "$PR_REPO" {number} rex "$MARKER_HOME")
 
-# Resolve the PR/MR HOST (base) repo — where the review must be POSTED and the
-# repo tracker_review_submit selects its adapter from. On a cross-fork PR this
-# differs from the fork above (posting to the fork fails: the PR lives on the
-# base). gh pr view has no baseRepository field, but the PR URL is ALWAYS on the
-# base repo — parse owner/repo from it (works for gh /pull/ and glab
-# /-/merge_requests/, incl. nested GitLab groups). Falls back to PR_REPO when
-# base == head (the common same-repo case).
-PR_HOST_REPO=$(gh pr view {number} --json url --jq '.url' 2>/dev/null | sed -E 's#^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9].*#\1#')
-[ -z "$PR_HOST_REPO" ] && PR_HOST_REPO="$PR_REPO"
+# Resolve the PR/MR HOST (base) repo — the repo the PR lives on. It is BOTH where
+# the review must be POSTED (posting to the fork fails on a cross-fork PR: the PR
+# lives on the base) AND the canonical key for the approval marker. `pr_base_repo`
+# (in _lib-review-markers.sh) parses the PR URL — gh pr view has no baseRepository
+# field — and falls back to PR_REPO when base == head, so same-repo PRs are
+# unchanged.
+PR_HOST_REPO=$(pr_base_repo {number} "$PR_REPO")
+
+# Marker keyed on the BASE repo (#765) — this MATCHES what block-unreviewed-merge.sh
+# looks up: the gate keys on the merge command's --repo / API-path, which for a
+# cross-fork PR is always the base (you cannot merge a fork's copy). Keying the
+# marker on headRepository (the fork) was the divergence that blocked cross-fork
+# approvals.
+REX_MARKER=$(review_marker_path "$PR_HOST_REPO" {number} rex "$MARKER_HOME")
 
 # Write your review to a temp body-file, then submit it through the abstraction.
 # A file (not inline text) is the uniform path: gh takes --body-file, glab reads
@@ -690,7 +696,7 @@ tracker_review_submit "$PR_HOST_REPO" {number} comment "$REVIEW_BODY_FILE"; subm
 
 ### The command
 
-Once `MARKER_HOME`, `PR_REPO`, and `REX_MARKER` are resolved (see above), use exactly one of these forms:
+Once `MARKER_HOME`, `PR_HOST_REPO`, and `REX_MARKER` are resolved (see above), use exactly one of these forms:
 
 ```bash
 # Option A — from the local HEAD of the PR branch
--- a/.claude/agents/solution-architect.md
+++ b/.claude/agents/solution-architect.md
@@ -172,26 +172,20 @@ MARKER_HOME="${OPS_ROOT:-$REPO_ROOT}"
 THREADED_REPO="{repo}"
 [ "$THREADED_REPO" = "{repo}" ] && THREADED_REPO=""
 
-# 2a. MARKER repo (keys the *-architecture.approved filename): threaded arg, else
-# headRepository. Unchanged from the pre-#763 behaviour — the cross-fork marker
-# alignment is tracked separately (see the note in the sign-off section below).
+# 2a. Resolve the PR's BASE (host) repo ONCE — the canonical key for BOTH the
+# review posting AND the sign-off marker (#765). $THREADED_REPO (when
+# /design-review passed it, #687) is the hint; otherwise headRepository. Then
+# pr_base_repo (in _lib-review-markers.sh) resolves the base from the PR URL
+# (gh pr view has no baseRepository field) and falls back to the hint when
+# base == head — so same-repo PRs are unchanged.
 if [ -n "$THREADED_REPO" ]; then
-  REPO="$THREADED_REPO"
+  HINT_REPO="$THREADED_REPO"
 else
-  REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
+  HINT_REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
 fi
-PR_REPO="$REPO"
-
-# 2b. HOST (base) repo for POSTING — the repo tracker_review_submit picks its
-# adapter from. When /design-review threaded a repo it IS the base repo; reuse it.
-# Only in the headRepository fallback (the fork, on a cross-fork PR) do we parse
-# the base from the PR URL (base-rooted; gh pr view has no baseRepository field).
-if [ -n "$THREADED_REPO" ]; then
-  PR_HOST_REPO="$THREADED_REPO"
-else
-  PR_HOST_REPO=$(gh pr view {number} --json url --jq '.url' 2>/dev/null | sed -E 's#^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9].*#\1#')
-  [ -z "$PR_HOST_REPO" ] && PR_HOST_REPO="$REPO"
-fi
+PR_HOST_REPO=$(pr_base_repo {number} "$HINT_REPO")
+REPO="$PR_HOST_REPO"      # the --repo flag for gh pr view when writing the marker SHA
+PR_REPO="$PR_HOST_REPO"   # marker key = the base repo (matches the gate's lookup)
 
 # 3. Write the review to a temp body-file and submit through the abstraction.
 # A file (not inline text) is the uniform path: gh takes --body-file, glab reads
@@ -211,15 +205,15 @@ When your verdict is APPROVED, and ONLY then, write the architecture-review appr
 
 ### Path: ops fork root, not git toplevel
 
-The marker MUST land at `<ops_fork_root>/.claude/session/reviews/<owner>__<repo>__{number}-architecture.approved` (repo-qualified path, AgDR-0060 / #485). Inside `workspace/<project>/`, `git rev-parse --show-toplevel` returns the project clone — NOT the ops fork; that's why `$MARKER_HOME` and `$PR_REPO` are resolved ONCE in "Posting the review" above (before any `cd` / `gh pr checkout`). **Reuse them here** — do not re-resolve (a second resolution risks keying the marker on a different repo than the one the review was posted to):
+The marker MUST land at `<ops_fork_root>/.claude/session/reviews/<owner>__<repo>__{number}-architecture.approved` (repo-qualified path, AgDR-0060 / #485). Inside `workspace/<project>/`, `git rev-parse --show-toplevel` returns the project clone — NOT the ops fork; that's why `$MARKER_HOME` and `$PR_HOST_REPO` are resolved ONCE in "Posting the review" above (before any `cd` / `gh pr checkout`). **Reuse them here** — do not re-resolve (a second resolution risks keying the marker on a different repo than the one the review was posted to):
 
 ```bash
-# $MARKER_HOME, $PR_REPO, and $REPO all come from "Posting the review" above.
+# $MARKER_HOME and $PR_HOST_REPO come from "Posting the review" above.
 mkdir -p "$MARKER_HOME/.claude/session/reviews"
-ARCH_MARKER=$(review_marker_path "$PR_REPO" {number} architecture "$MARKER_HOME")
+ARCH_MARKER=$(review_marker_path "$PR_HOST_REPO" {number} architecture "$MARKER_HOME")
 ```
 
-> **Cross-fork note (tracked separately).** `$PR_REPO` keys the marker on the threaded repo, else `headRepository` (the fork). The gate `require-architecture-review.sh` keys its lookup on the merge command's **base** repo — so on a **cross-fork** PR the two can diverge (marker written under the fork qualifier; gate searches under the base). Aligning all marker writers to the base repo is a dedicated follow-up filed alongside #763. This PR only routes review *submission* through the tracker abstraction; marker keying is deliberately **unchanged** here.
+> **Cross-fork keying (#765).** `$PR_HOST_REPO` is the PR's **base** repo — exactly what `require-architecture-review.sh` keys its lookup on (the merge command's `--repo` / API-path, which on a cross-fork PR is always the base, since you cannot merge a fork's copy). Keying the marker on the base makes it findable by the gate on cross-fork PRs; same-repo PRs are unaffected (base == head). This replaces the earlier headRepository (fork) keying, which silently blocked cross-fork approvals.
 
 ### The command
 
--- a/.claude/agents/solution-architect.md
+++ b/.claude/agents/solution-architect.md
@@ -148,18 +148,32 @@ When MCP `search_docs` is available, you MAY supplement path-convention discover
 Resolve `$MARKER_HOME` and the PR's repo **once** here, then post. The sign-off marker below **reuses** these variables — do not re-resolve them (a second resolution risks diverging from the repo the marker is keyed on). Resolve at review start, before any `cd` / `gh pr checkout`.
 
 ```bash
-# 1. Ops fork root (walk up from the repo toplevel; `.apexyard-fork` first, then
-# the onboarding + registry pair). Inside workspace/<project>/, git toplevel is
-# the project clone — the walk-up finds the ops fork above it.
-REPO_ROOT=$(git rev-parse --show-toplevel)
+# 1. Ops fork root — resolve PIN-FIRST, the SAME strategy the merge gate and the
+# other reviewer agents (code-reviewer.md, security-reviewer.md) use. The session
+# pin points at the real ops fork even from inside a workspace/<project>/ clone;
+# a plain walk-up resolves to the private portfolio sibling in split-portfolio
+# mode (me2resh/apexyard#559). Fall back to walk-up only when no valid pin exists.
 OPS_ROOT=""
-r="$REPO_ROOT"
-while [ -n "$r" ] && [ "$r" != "/" ]; do
-  if [ -f "$r/.apexyard-fork" ]; then OPS_ROOT="$r"; break; fi
-  if [ -f "$r/onboarding.yaml" ] && [ -f "$r/apexyard.projects.yaml" ]; then OPS_ROOT="$r"; break; fi
-  r=$(dirname "$r")
-done
-MARKER_HOME="${OPS_ROOT:-$REPO_ROOT}"
+PIN_FILE="${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID:-}"
+if [ -z "${APEXYARD_OPS_DISABLE_PIN:-}" ] && [ -n "${CLAUDE_CODE_SESSION_ID:-}" ] && [ -f "$PIN_FILE" ]; then
+  IFS= read -r OPS_ROOT < "$PIN_FILE" || OPS_ROOT=""
+fi
+# Validate the pin (self-heal a stale one): must satisfy a fork anchor.
+if [ -n "$OPS_ROOT" ] && [ ! -f "$OPS_ROOT/.apexyard-fork" ] && \
+   { [ ! -f "$OPS_ROOT/onboarding.yaml" ] || [ ! -f "$OPS_ROOT/apexyard.projects.yaml" ]; }; then
+  OPS_ROOT=""
+fi
+# Fallback: walk up from the repo root (pre-#381 behaviour, safety net).
+if [ -z "$OPS_ROOT" ]; then
+  r=$(git rev-parse --show-toplevel 2>/dev/null || pwd)
+  while [ -n "$r" ] && [ "$r" != "/" ]; do
+    if [ -f "$r/.apexyard-fork" ] || { [ -f "$r/onboarding.yaml" ] && [ -f "$r/apexyard.projects.yaml" ]; }; then
+      OPS_ROOT="$r"; break
+    fi
+    r=$(dirname "$r")
+  done
+fi
+MARKER_HOME="${OPS_ROOT:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}"
 # _lib-tracker.sh posts the human-visible review to the right host; the marker
 # helper is sourced too so the sign-off marker below can reuse $MARKER_HOME.
 # shellcheck source=/dev/null
--- a/.claude/hooks/_lib-review-markers.sh
+++ b/.claude/hooks/_lib-review-markers.sh
@@ -88,3 +88,54 @@ review_marker_path() {
 
   printf '%s/%s__%s-%s.approved' "$reviews_dir" "$safe_repo" "$pr" "$role"
 }
+
+# pr_base_repo <pr> [hint_repo]
+#
+# Echoes the PR/MR's BASE (host) repo as "owner/repo" — the repo the PR *lives
+# on* and is numbered against. This is the canonical key for approval markers
+# (me2resh/apexyard#765).
+#
+# WHY THE BASE REPO IS CANONICAL
+# ------------------------------
+# The merge gates (block-unreviewed-merge.sh, require-architecture-review.sh,
+# require-design-review-for-ui.sh) derive their marker-lookup repo (`CMD_REPO`)
+# from the merge command's `--repo` value or `gh api repos/<o>/<r>/pulls/.../merge`
+# path. For a CROSS-FORK PR that is ALWAYS the base repo — you cannot merge a
+# fork's copy (`gh pr merge <n> --repo <fork>` errors; the PR doesn't live
+# there). So `merge --repo == CMD_REPO == base`. Historically the marker WRITERS
+# keyed on `headRepository` (the fork) instead, so on a cross-fork PR the marker
+# was written under the fork qualifier while the gate searched under the base →
+# a valid approval never satisfied the gate. Keying every writer on the base via
+# this helper makes writer/reader agreement STRUCTURAL, not coincidental.
+#
+# `gh pr view` exposes no baseRepository field, but the PR URL is ALWAYS rooted
+# on the base repo — parse owner/repo from it (handles GitHub /pull/ and GitLab
+# /-/merge_requests/, including nested GitLab groups). Falls back to <hint_repo>
+# (typically the headRepository value) when the URL can't be parsed or gh is
+# unavailable — so SAME-REPO PRs (base == head) resolve exactly as before and
+# this change is a provable no-op for them.
+#
+# Args:
+#   pr        — the PR/MR number.
+#   hint_repo — optional "owner/repo": scopes the gh query with --repo AND is the
+#               fallback when URL parsing yields nothing.
+#
+# Output (stdout): "owner/repo", or the hint (or empty) when unresolved.
+pr_base_repo() {
+  local pr="${1:-}" hint="${2:-}" url base
+  if [ -z "$pr" ]; then
+    [ -n "$hint" ] && printf '%s' "$hint"
+    return 0
+  fi
+  if [ -n "$hint" ]; then
+    url=$(gh pr view "$pr" --repo "$hint" --json url --jq '.url' 2>/dev/null)
+  else
+    url=$(gh pr view "$pr" --json url --jq '.url' 2>/dev/null)
+  fi
+  base=$(printf '%s' "$url" | sed -E 's#^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9].*#\1#')
+  if [ -n "$base" ] && [ "$base" != "$url" ]; then
+    printf '%s' "$base"
+  else
+    [ -n "$hint" ] && printf '%s' "$hint"
+  fi
+}
--- a/.claude/hooks/_lib-review-markers.sh
+++ b/.claude/hooks/_lib-review-markers.sh
@@ -117,8 +117,9 @@ review_marker_path() {
 #
 # Args:
 #   pr        — the PR/MR number.
-#   hint_repo — optional "owner/repo": scopes the gh query with --repo AND is the
-#               fallback when URL parsing yields nothing.
+#   hint_repo — optional "owner/repo": the VALUE fallback when the URL can't be
+#               parsed or gh is unavailable. It is NEVER used to scope the gh
+#               query — see the WHY-UNSCOPED note in the body below.
 #
 # Output (stdout): "owner/repo", or the hint (or empty) when unresolved.
 pr_base_repo() {
@@ -127,11 +128,14 @@ pr_base_repo() {
     [ -n "$hint" ] && printf '%s' "$hint"
     return 0
   fi
-  if [ -n "$hint" ]; then
-    url=$(gh pr view "$pr" --repo "$hint" --json url --jq '.url' 2>/dev/null)
-  else
-    url=$(gh pr view "$pr" --json url --jq '.url' 2>/dev/null)
-  fi
+  # WHY UNSCOPED (me2resh/apexyard#770 review): query with NO --repo. gh's ambient
+  # base-repo resolution (from the working copy's remotes) prefers the parent /
+  # upstream — i.e. the BASE repo — which is exactly the key we want. Scoping with
+  # `--repo "$hint"` (the head/fork) would look up the base-numbered PR on the fork,
+  # where it does not exist → gh errors → empty url → the hint (fork) fallback fires
+  # → the #765 divergence is re-created on the very cross-fork path this helper
+  # exists to fix. So the hint is a VALUE fallback only, never a query scope.
+  url=$(gh pr view "$pr" --json url --jq '.url' 2>/dev/null)
   base=$(printf '%s' "$url" | sed -E 's#^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9].*#\1#')
   if [ -n "$base" ] && [ "$base" != "$url" ]; then
     printf '%s' "$base"
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -76,6 +76,12 @@ fi
 PR_NUMBER=$(extract_pr_number "$COMMAND")
 # Also extract the repo so markers are scoped to (repo, pr) — #485.
 # CMD_REPO already parsed above; resolve via helper if blank (e.g. current-branch fallback).
+# NOTE (#765): approval markers are keyed on the PR's BASE repo. CMD_REPO is the base
+# for the sanctioned paths — the --repo value of `gh pr merge --repo` (you cannot merge
+# a fork's copy) and the `gh api .../pulls/N/merge` path. The extract_repo_from_command
+# fallback below resolves headRepository (the FORK) for a no---repo current-branch merge;
+# that residual path is NOT produced by /approve-merge (which always passes --repo), so it
+# only affects unsanctioned manual merges. Left as-is to keep the gate core untouched.
 if [ -z "$CMD_REPO" ]; then
   CMD_REPO=$(extract_repo_from_command "$COMMAND")
 fi
--- a/.claude/hooks/require-architecture-review.sh
+++ b/.claude/hooks/require-architecture-review.sh
@@ -83,6 +83,10 @@ fi
 PR_NUMBER=$(extract_pr_number "$COMMAND")
 # Resolve the repo for qualified marker paths (#485).
 # CMD_REPO already resolved above; fall back via helper if still blank.
+# NOTE (#765): the architecture marker is keyed on the BASE repo. CMD_REPO is the base via
+# --repo / API-path / cd-target origin; the extract_repo_from_command fallback below resolves
+# headRepository (the FORK) on a no---repo current-branch merge — a residual edge affecting
+# unsanctioned merges only (/design-review + /approve-architecture thread the base repo). Left as-is.
 if [ -z "$CMD_REPO" ]; then
   CMD_REPO=$(extract_repo_from_command "$COMMAND")
 fi
--- a/.claude/hooks/require-design-review-for-ui.sh
+++ b/.claude/hooks/require-design-review-for-ui.sh
@@ -82,6 +82,10 @@ fi
 PR_NUMBER=$(extract_pr_number "$COMMAND")
 # Resolve the repo for qualified marker paths (#485).
 # CMD_REPO already resolved above; fall back via helper if still blank.
+# NOTE (#765): the design marker is keyed on the BASE repo. CMD_REPO is the base via
+# --repo / API-path / cd-target origin; the extract_repo_from_command fallback below resolves
+# headRepository (the FORK) on a no---repo current-branch merge — a residual edge affecting
+# unsanctioned merges only (/design-review + /approve-design thread the base repo). Left as-is.
 if [ -z "$CMD_REPO" ]; then
   CMD_REPO=$(extract_repo_from_command "$COMMAND")
 fi
--- a/.claude/hooks/tests/test_pr_base_repo.sh
+++ b/.claude/hooks/tests/test_pr_base_repo.sh
@@ -8,6 +8,13 @@
 #
 # gh is mocked via a file-driven stub (no env-export subtleties): the stub prints
 # the contents of $MOCKBIN/url, or exits non-zero when that file is empty.
+#
+# The stub is --repo-AWARE (me2resh/apexyard#770 review): a `gh pr view --repo <r>`
+# call only resolves when <r> is the BASE repo in the queued URL; scoping to any
+# other repo (e.g. the fork on a cross-fork PR) fails exactly like real gh rejecting
+# a base-numbered PR looked up on the fork. This makes the cross-fork case below
+# LOAD-BEARING: it fails if pr_base_repo ever re-introduces `--repo "$hint"` scoping
+# (the query MUST be unscoped) and passes only when the base is genuinely resolved.
 
 set -u
 
@@ -25,9 +32,22 @@ assert_eq() { # <label> <want> <got>
 MOCKBIN=$(mktemp -d)
 cat > "$MOCKBIN/gh" <<EOF
 #!/bin/bash
-# mock gh: emit the queued URL, or fail when none is queued.
+# mock gh (--repo-aware): emit the queued URL only when the query would resolve —
+# i.e. UNSCOPED (ambient base resolution) OR --repo naming the base repo in the
+# queued URL. A --repo pointing anywhere else (the fork) fails, like real gh
+# rejecting a base-numbered PR looked up on the fork. Empty queue → fail.
 [ -s "$MOCKBIN/url" ] || exit 1
-cat "$MOCKBIN/url"
+url="\$(cat "$MOCKBIN/url")"
+base="\$(printf '%s' "\$url" | sed -E 's#^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9].*#\1#')"
+repo=""
+while [ \$# -gt 0 ]; do
+  case "\$1" in
+    --repo) repo="\$2"; shift 2 ;;
+    *) shift ;;
+  esac
+done
+if [ -n "\$repo" ] && [ "\$repo" != "\$base" ]; then exit 1; fi
+printf '%s' "\$url"
 EOF
 chmod +x "$MOCKBIN/gh"
 export PATH="$MOCKBIN:$PATH"
--- a/.claude/hooks/tests/test_pr_base_repo.sh
+++ b/.claude/hooks/tests/test_pr_base_repo.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+# Tests for pr_base_repo (+ its interaction with review_marker_path) in
+# _lib-review-markers.sh — the cross-fork approval-marker fix (me2resh/apexyard#765).
+#
+# pr_base_repo resolves a PR's BASE repo (the canonical marker key) by parsing the
+# PR URL, falling back to a hint when base == head or gh fails. The discriminating
+# assertions prove: (a) cross-fork keys on BASE, (b) same-repo is UNCHANGED.
+#
+# gh is mocked via a file-driven stub (no env-export subtleties): the stub prints
+# the contents of $MOCKBIN/url, or exits non-zero when that file is empty.
+
+set -u
+
+SRC_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+LIB="$SRC_ROOT/.claude/hooks/_lib-review-markers.sh"
+# shellcheck source=/dev/null
+. "$LIB"
+
+PASS=0; FAIL=0; FAILED=""
+assert_eq() { # <label> <want> <got>
+  if [ "$2" = "$3" ]; then PASS=$((PASS+1));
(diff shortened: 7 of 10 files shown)


AbdElrahmaN31 (commented):
## Code Review: PR #770

**Commit**: `f11415c41f0b55ee346b484ef9325102a081846e` (tip of `feature/GH-765-crossfork-marker-base-repo`)

**Verdict: APPROVED**

Scope: reviewed ONLY the two #765 commits (`81b63c3` base-repo keying, `f11415c` pin-first unification), isolated via `git diff feature/GH-763..feature/GH-765`. The #762/#763 changes carried in the `dev`-based diff (security-reviewer.md, _lib-tracker.sh, test_tracker_review_submit.sh) were NOT re-litigated.

### Summary
Fixes the cross-fork approval-marker divergence: marker *writers* keyed on `headRepository` (the fork) while the merge *gates* key on the base repo, so on a cross-fork PR a valid approval was written under the fork qualifier and the gate never found it. #765 introduces `pr_base_repo` in `_lib-review-markers.sh` and routes all four scoped writers through it, making writer/reader agreement structural. A second commit unifies `solution-architect.md` on pin-first ops-root resolution.

### Checklist Results
- Testing:                    Pass (9/9 new + 29/29 regression, verified locally)
- Security:                   Pass (no secrets, no injection surface; shell-quoting correct)
- Performance:                Pass (one extra `gh pr view --json url` per marker op — negligible)
- PR Description & Glossary:  Pass (Glossary present: `pr_base_repo`, base vs head, writer/reader)
- Summary Bullet Narrative:   Pass (narrative, not label-only)
- Technical Decisions (AgDR): N/A (bug-fix aligning existing behaviour; rationale documented inline + in `pr_base_repo` header)
- Adopter Handbooks:          N/A (no handbooks loaded; diff is shell + markdown)

### Verified correctness points
1. **`pr_base_repo` URL regex** `s#^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9].*#\1#` — correct for GitHub `/pull/`, GitLab `/-/merge_requests/` (incl. nested groups via greedy `.+` backtracking), and self-hosted/GHE hosts (`[^/]+` matches any host). Trailing path/query is swallowed by `[0-9].*`. Fail-soft: empty URL / gh failure / unparseable URL → `base==url` (or empty) → hint fallback. Guarded by both `[ -n "$base" ]` and `[ "$base" != "$url" ]`.
2. **Same-repo is a provable no-op** — for base==head the URL parses to a value equal to the hint (`headRepository`), so the marker keys exactly as the pre-#765 code did. Confirmed by logic AND the discriminating marker-path test pair (cross-fork → BASE, same-repo → unchanged).
3. **All four scoped writers on base** — code-reviewer (rex → `PR_HOST_REPO`), solution-architect (arch → `PR_HOST_REPO`, 2a/2b collapsed into one resolution), /approve-merge (ceo write + rex read + merge `--repo` all → `PR_HOST_REPO`), /approve-design (design write + rex read → `PR_HOST_REPO`). None left on `headRepository`.
4. **/approve-merge ordering** — `PR_HOST_REPO` resolved in step 4; no earlier step depends on the shell variable (step 3 uses the `<owner/repo>` placeholder, agent-substituted from invocation context). The `<owner/repo> = $PR_HOST_REPO` convention note is coherent.
5. **Gates unchanged** — all three merge-gate hooks received comment-only additions (verified every changed line is a `+` comment); no logic drift. `block_unreviewed_merge` regression 29/29.
6. **Tests** — `test_pr_base_repo.sh` 9/9; the 9 cases genuinely discriminate, especially the cross-fork-vs-same-repo marker-path pair. `bash -n` clean on the lib + all three gates.
7. **Pin-first unification (`f11415c`)** — the new `solution-architect.md` block is line-for-line identical to the canonical `code-reviewer.md` pin-first block; "unify" is a true copy, no drift.

- **suggestion: `/approve-architecture` (`.claude/skills/approve-architecture/SKILL.md:75-76,97`) is a fifth writer/reader of the architecture + rex markers that was not migrated to `pr_base_repo`.** It keys the rex-read and arch-write on `PR_REPO="${REPO:-$(gh pr view ... headRepository ...)}"`. Reading the code (not the gate comment that asserts it "thread[s] the base repo"): it keys on base *only* under sanctioned `owner/repo#N` invocation. A bare PR number where `gh pr view` resolves `headRepository=fork` mis-keys the marker to the fork. The failure is fail-safe (marker written on fork → gate reads base → merge blocked; or rex-read on fork → refuses) — it can never produce a wrongful approval, and the fail-loud guard at step 1 covers the ambiguous split-portfolio case. But it is exactly the "coincidental, not structural" pattern this PR's summary sets out to eliminate. Consider adopting `pr_base_repo` here too for defense-in-depth. Recommend as a follow-up; does not block this PR (the primary automated architecture path, `/design-review` → solution-architect, IS fixed structurally).
- **nit:** `solution-architect.md` (and `approve-design`/`approve-merge` skills) — the pin-first ops-root unification in `f11415c` covered solution-architect only; the `/approve-*` skills still use the older walk-up resolver. Out of this PR's stated scope, but a natural next consistency sweep.

### Issues Found
None blocking.

---
Reviewed commit: `f11415c41f0b55ee346b484ef9325102a081846e`

AbdElrahmaN31 (commented):
## Code Review (re-review): PR #770

**Commit**: `d8ad5325b7bacee2a27f028a2e7cc4404ee97fea`

Re-review triggered by a new commit (`d8ad532`) pushed after my prior APPROVED verdict on `81b63c3` + `f11415c`. Focused on the delta.

### Summary
`d8ad532` migrates the **fifth and final** marker writer — `/approve-architecture` (`.claude/skills/approve-architecture/SKILL.md`) — to key its markers on the PR's **base** repo via `pr_base_repo`, completing the cross-fork fix and mirroring the already-approved `/approve-design` migration. Single-file change (+14/-7).

### Checklist Results
- ✅ Architecture & Design:      Pass — consistent with the base-repo keying model across all writers
- ✅ Code Quality:               Pass — shell blocks are structurally identical to the vetted `/approve-design` change
- ✅ Testing:                    Pass — no logic change to `pr_base_repo`; existing 9/9 + 29/29 regression suites already cover the resolver + same-repo no-op
- ✅ Security:                   Pass — no secrets, no injection surface (marker path built from a validated repo slug)
- ✅ Performance:                Pass — N/A
- ✅ PR Description & Glossary:  Pass — glossary present, summary bullets narrative
- ⚠ Summary Bullet Narrative:   Pass
- ✅ Technical Decisions (AgDR): N/A — continuation of the existing #765 base-repo-keying decision; no new decision introduced
- ✅ Adopter Handbooks:          N/A — docs-only diff (no language-extension match), no handbook triggers
- ⚠ Fallow Static Analysis:     N/A — not JS/TS

### Verification of the new commit (`d8ad532`)
1. ✅ `/approve-architecture` now resolves `HINT_REPO="${REPO:-<headRepository>}"` → `PR_HOST_REPO=$(pr_base_repo <pr> "$HINT_REPO")`, and keys **both** the rex-read (step 4, `REX=review_marker_path "$PR_HOST_REPO" …`) and the architecture-write (step 6, `ARCH=review_marker_path "$PR_HOST_REPO" …`) on `$PR_HOST_REPO` — matching `solution-architect.md` and the `require-architecture-review.sh` gate.
2. ✅ Same-repo remains a no-op — `pr_base_repo` parses the PR URL and falls back to the hint when base == head (helper unchanged at `_lib-review-markers.sh:124`).
3. ✅ Structurally identical to the approved `/approve-design` migration (same `HINT_REPO` / `PR_HOST_REPO` / `PR_REPO` assignment shape).
4. ✅ Full writer set now consistently base-keyed: code-reviewer (rex), solution-architect (architecture), /approve-merge (ceo), /approve-design (design), /approve-architecture (architecture). No writer keys the marker **write** on `headRepository` — remaining `headRepository` references are all the intended `HINT_REPO` fallback feeding `pr_base_repo`.

Earlier commits (`81b63c3`, `f11415c`) confirmed unchanged since the prior approval.

### Issues Found
None.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `d8ad5325b7bacee2a27f028a2e7cc4404ee97fea`

AbdElrahmaN31:
## ⚠️ Merge order — third in the stack (merge #762 → #766 → this)

This PR's branch is stacked on `feature/GH-763` (PR #766, itself stacked on #762). Because the base here is `dev`, the diff currently includes #762's and #763's changes too — that's expected. #765's own changes are the three commits `81b63c3`, `f11415c`, `d8ad532`.

**Order: merge #762 → #766 → then this (#770).** Once #762 and #766 are in `dev`, this PR's diff reduces to just its own commits.

**If you squash-merge the parents:** a squash creates new commits on `dev` that aren't in this branch's history, so this PR will then show conflicts / duplicated changes. In that case please hold and ping me — I'll rebase `feature/GH-765` onto the updated `dev` and force-push so the diff is clean before you merge. (Merge-commit merges of the parents need no rebase.)

Context: #765 is the cross-fork approval-marker fix — it makes the manual base-repo marker workaround (needed on #762/#766/#768) unnecessary for future cross-fork PRs.

atlas-apex (changes requested):
## Code Review: PR #770

**Commit**: `d8ad5325b7bacee2a27f028a2e7cc4404ee97fea`

### Summary
Routes all five approval-marker writers (rex/code-reviewer, architecture/solution-architect, ceo/approve-merge, design/approve-design, architecture/approve-architecture) through a new `pr_base_repo` helper so the marker filename keys on the PR's **base** repo — matching what the merge gates already look up via `extract_repo_from_command`. The goal (close the cross-fork divergence in #765) is exactly right, the all-or-nothing conversion intent is right, and the gate-reader side is correctly left untouched (comment-only, honestly documenting the residual no-`--repo` fallback edge). But the helper itself does not resolve the base on a cross-fork PR — so the central fix does not work.

### Checklist Results
- Code Quality:               **Fail** (see blocker)
- Testing:                    **Fail** — the test mock masks the defect (false positive)
- Security:                   Pass (no privilege escalation — see note)
- Adopter Handbooks:          N/A (no handbooks loaded)

### Issues Found

**BLOCKER 1 — `pr_base_repo` scopes the primary query to the fork hint, so cross-fork resolves to the FORK (the fix does not fix #765).**

`_lib-review-markers.sh::pr_base_repo` runs its primary URL lookup as:

```sh
if [ -n "$hint" ]; then
  url=$(gh pr view "$pr" --repo "$hint" --json url --jq '.url' 2>/dev/null)
```

Every writer passes `headRepository` (the **fork**) as `$hint` — code-reviewer and approve-merge unconditionally, and solution-architect / approve-design / approve-architecture whenever the threaded `REPO` is empty. On a cross-fork PR the PR number belongs to the **base** repo, so `gh pr view <base-pr> --repo <fork>` **errors** ("Could not resolve to a PullRequest"), `url` comes back empty, and the helper falls back to `$hint` = the fork. The marker is therefore keyed on the fork — the exact divergence #765 sets out to remove.

Proven two independent ways against this very PR (which is itself cross-fork, head `AbdElrahmaN31/apexyard` → base `me2resh/apexyard`):

- Live: `gh pr view 770 --repo AbdElrahmaN31/apexyard` → `Could not resolve to a PullRequest with the number of 770`; consequently `pr_base_repo 770 "AbdElrahmaN31/apexyard"` returns `AbdElrahmaN31/apexyard` (want `me2resh/apexyard`).
- Faithful mock (one that honors `--repo`, as real gh does): `pr_base_repo 762 AbdElrahmaN31/apexyard` → `AbdElrahmaN31/apexyard`.

Downstream consequences on a cross-fork PR:
- code-reviewer writes the **rex** marker under `<fork>__…` (unchanged from the pre-fix bug).
- `/approve-merge` computes `PR_HOST_REPO` = fork, then both writes the **ceo** marker under the fork AND runs `gh pr merge <pr> --repo <fork>`, which itself fails (you cannot merge a fork's copy).
- If a human instead runs the correct `gh pr merge <pr> --repo <base>`, the gate derives `CMD_REPO` = base and looks up markers under base — but rex/ceo wrote them under the fork → the gate blocks. Square one.

So the load-bearing rex + ceo markers still diverge from the base on cross-fork; the same-repo path only works because there `hint == base == head`. This is the all-or-nothing failure the ticket warns about — structurally all five writers were routed through the helper (good), but the helper returns the fork, so the whole set is still wrong for cross-fork.

Root cause: the refactor reuses `$hint` as the query **scope** (`--repo "$hint"`), but the hint is the fork — the one repo guaranteed *not* to host a base PR number. Note the pre-#765 inline code this replaced queried **unscoped** (`gh pr view {number} --json url` with no `--repo`), relying on ambient/base context, and resolved the base correctly. The scoping is a regression relative to what it replaced.

Suggested fix: run the primary URL query **unscoped** (ambient context resolves the base PR, exactly as the initial `gh pr view {number} --json headRepository` already does to obtain the hint), and treat `$hint` strictly as a *value* fallback — never as the `--repo` scope. i.e.:

```sh
url=$(gh pr view "$pr" --json url --jq '.url' 2>/dev/null)   # unscoped: ambient == base
base=$(printf '%s' "$url" | sed -E '...')
[ -n "$base" ] && [ "$base" != "$url" ] && { printf '%s' "$base"; return 0; }
[ -n "$hint" ] && printf '%s' "$hint"
```

**BLOCKER 2 — the test mock ignores `--repo`, so the suite green-lights the defect.**

`tests/test_pr_base_repo.sh` stubs `gh` to emit `$MOCKBIN/url` regardless of arguments. Because the mock never checks `--repo`, test #1 ("cross-fork → base from URL", hint = the fork) passes even though real `gh pr view <base-pr> --repo <fork>` fails. The load-bearing assertion is a false positive; 9/9 green does not demonstrate the fix works. A faithful mock — return the base URL only when `--repo` is the base (or absent) and `exit 1` for `--repo <fork>` on a base PR number — flips test #1 to FAIL, matching the live behaviour above. Please make the mock `--repo`-aware so the regression is actually covered (both same-repo and cross-fork).

### Suggestions
- nit: once the helper queries unscoped, drop the `--repo "$hint"` branch entirely — the hint has no valid role as a query scope, only as a fallback value.

### Verdict
**CHANGES REQUESTED**

The direction, the writer/reader-agreement framing, the honest gate-reader comments, and the same-repo no-op are all correct — but the helper returns the fork (not the base) on exactly the cross-fork PRs this PR exists to fix, and the test mock hides it. No approval marker written. (Security note: the defect cannot forge a base-repo marker or target an unapproved repo — it only mis-keys onto the fork, which then fails to satisfy the base gate; it is a correctness bug, not an escalation.)

---
📌 Reviewed commit: `d8ad5325b7bacee2a27f028a2e7cc4404ee97fea`

atlas-apex (commented):
## 🛡️ Security Review: PR #770 (Hakim, Security Auditor)

**Commit reviewed**: `d8ad5325b7bacee2a27f028a2e7cc4404ee97fea`
**Scope**: adversarial security pass on the merge-gate **trust chain** — the `pr_base_repo` helper and the five marker writers now keyed on the PR **base** repo. This is the *security layer* on top of Rex's correctness review (Rex already requested changes for the cross-fork mis-resolution; that finding stands and is unaffected by this pass).

### Summary
The change moves approval-marker **keying** from `headRepository` (the fork) to the PR **base** repo, via a new `pr_base_repo` helper that parses the PR URL. I traced every adversarial angle you flagged — marker forging / confused-deputy, path-traversal in the `owner/repo` slug, and fail-open on the merge gate. **I independently confirm Rex's security judgment: no forge, no fail-open, no traversal.** The change is strictly *fail-closed*.

### Checklist Results
- Secrets & Credentials: **Pass** (no secrets touched)
- Injection Prevention: **Pass** (one LOW informational note on the pre-existing `custom` adapter)
- Path / traversal safety: **Pass**
- Auth & Authorisation (gate integrity): **Pass**
- Confused-deputy / forging: **Pass**
- Fail-closed on error: **Pass**

---

### 1. Path safety — the `owner/repo` slug cannot escape `.claude/session/reviews/` ✅
The load-bearing defense is `review_marker_path`'s `sed 's|/|__|g'`, which replaces **every** `/` in the slug with `__`. A filesystem traversal component (`../`) requires a literal slash; since no slash survives the sanitiser, **no traversal segment can form regardless of input**. Even a hostile segment like `..` arrives as `foo__..__bar` — a single flat filename component, inert as a path element.

Defense-in-depth is also present *upstream*: the slug originates from `gh pr view --json url`, i.e. GitHub's canonical PR URL, whose `owner/repo` is constrained to `[A-Za-z0-9._-]` (no `/`, no `..` path segments, no metacharacters). GitLab group/project names are similarly constrained. So the slug is well-formed before it even reaches the flatten. **Constraint + flatten = sound. No path traversal.**

### 2. Marker forging / confused-deputy — cannot write an approving marker for an unapproved PR ✅
Traced `pr_base_repo`'s inputs (`pr`, parsed `url`, `hint`) to the write site:

- The only attacker-controllable input on a cross-fork PR is the **fork** (head repo). The parsed slug comes from the **base** repo's URL (`https://github.com/<base>/pull/<N>`) — an attacker opening a PR against `me2resh/apexyard` **cannot** make the base owner/repo be anything other than the legitimate target. The fork name never appears in `.url`.
- The theoretical confusion (`hint`-scoped `gh pr view <N> --repo <fork>` resolving a *same-numbered* PR that also exists in the hint repo) is **defeated by the SHA binding**: the marker embeds the reviewed HEAD SHA, and `block-unreviewed-merge.sh` compares it against the target PR's real HEAD. A marker that landed under the wrong base qualifier carries the *reviewed* PR's SHA, which won't match a *different* PR's HEAD → gate still blocks. To produce a base-keyed marker with the *correct* target SHA, the agent must have actually reviewed that PR at that SHA — i.e. a genuine review. **No forge path.**

### 3. Gate weakening / fail-open — none ✅
The **read side (merge gates) is untouched** — it still keys the lookup on `CMD_REPO` (base, from `--repo`/API-path). This PR only changes the **write** side to *also* aim at the base. When resolution succeeds, writer and reader agree (the intended fix). When resolution **fails** (Rex's correctness bug: cross-fork falls back to the fork qualifier), the marker lands under the fork while the gate searches the base → **NOT FOUND → merge BLOCKED**. That is fail-**closed** — an *availability* regression (legit approvals blocked), never an authorization bypass. There is **no path where the base-keying change lets an unapproved cross-fork PR merge**: the change makes producing a valid marker *harder*, not easier.

### 4. Fail-closed on resolution failure ✅
Confirmed on every branch: `gh` error / empty / unparseable URL → falls back to `hint`; empty `pr` → `hint` or empty; empty slug into `review_marker_path` → returns 1 with an error, **no marker written**. Every failure mode ends in "marker absent or mis-keyed" → gate blocks.

---

### LOW / informational (NOT blocking, NOT introduced by this PR)
The `custom` review adapter (`_tracker_review_custom`) substitutes `{owner_repo}` — now sourced from the URL-parsed `PR_HOST_REPO` — into an `eval "$cmd"` template. The truly-arbitrary value (the review body) is correctly kept **out** of the eval via the `$TRACKER_REVIEW_BODY_FILE` env var, and `{pr}` (numeric) / `{verdict}` (enum) are validated — good. `{owner_repo}` injection would require a git host that permits shell metacharacters in repo names, which GitHub/GitLab do not. This mirrors the **pre-existing** `_tracker_create_custom` model, so it's out of scope for #765 — just noting it for the eventual hardening pass (prefer double-quoted env-var exposure for `{owner_repo}` too, over template substitution). The `gh` (array-arg) and `glab` (quoted `-m`) adapters are injection-safe.

---

### Verdict
**COMMENT — no security-blocking findings.** The trust-chain change is fail-closed and forge-resistant; the SHA binding is the backstop that makes cross-PR confused-deputy infeasible. Security-wise this is safe to merge **once Rex's correctness fix lands** — I concur with Rex's outstanding **CHANGES REQUESTED**, because a fix that still mis-keys cross-fork markers (fail-closed) defeats the stated purpose of #765 even though it opens no security hole. No additional security changes required.

---
📌 Reviewed commit: `d8ad5325b7bacee2a27f028a2e7cc4404ee97fea`

atlas-apex (changes requested):
Thanks so much for taking this on, @AbdElrahmaN31 — the cross-fork marker-keying bug is a subtle one and an important fix for the fork → upstream flow. 🙏 Routing all five writers through a single `pr_base_repo` helper is exactly the right structure, the gate-reader hooks are correctly left untouched, and the same-repo no-op is clean.

**Requesting changes on one blocking item — the fix doesn't quite land on the case it targets yet.** `pr_base_repo` runs its primary lookup as `gh pr view "$pr" --repo "$hint"`, and every writer passes `headRepository` (the fork) as the hint. On a cross-fork PR the base PR number doesn't exist on the fork, so `gh pr view <base-pr> --repo <fork>` errors → `url` comes back empty → it falls back to the hint = the fork. So the marker is still keyed on the fork — the exact divergence #765 exists to remove. (Reproduced against this PR: `pr_base_repo 770 <fork>` → the fork, not `me2resh/apexyard`.) And because the test's `gh` mock emits the URL regardless of `--repo`, the load-bearing cross-fork test is green without actually exercising the cross-fork path.

**Suggested fix:**
1. Query **unscoped** — the ambient repo context is the base, so `gh pr view <pr> --json url` (no `--repo`) resolves the base correctly; use the fork hint only as a *value* fallback, never as the query scope. (This is what the inline code you replaced did.)
2. Make the test's `gh` mock **`--repo`-aware** so the cross-fork case #1 genuinely fails pre-fix and passes post-fix.

Security-wise it's a correctness bug, not an escalation — a mis-key fails *closed* (blocks a merge), never opens one — so no rush, but it does need to actually resolve the base before it helps. Thanks again for the careful work here; the shape is right and it's close. 🙌

AbdElrahmaN31:
Verified and fixed in `041cf0e`.

**The bug reproduced exactly as described.** Against this PR:

```
$ gh pr view 770 --repo AbdElrahmaN31/apexyard --json url   # scoped to fork (what the code did)
GraphQL: Could not resolve to a PullRequest with the number of 770.   # → errors → empty url → fork fallback

$ gh pr view 770 --json url --jq .url                        # unscoped
https://github.com/me2resh/apexyard/pull/770                 # → base resolved
```

So `pr_base_repo 770 <fork>` returned the fork — the #765 divergence, intact on the one path #765 exists to remove.

**Fix (both points):**

1. **Query unscoped.** Dropped the `--repo "$hint"` scoping entirely; gh's ambient base-repo resolution prefers the parent/upstream (the base), so `gh pr view <pr> --json url` is correct. The hint is now a *value* fallback only (URL-unparseable / gh-unavailable), never a query scope — arg doc and a `WHY UNSCOPED` note updated to lock the intent.
2. **`--repo`-aware mock.** The test stub now fails a `--repo <r>` query unless `<r>` is the base repo in the queued URL, mirroring real gh. This makes the cross-fork case load-bearing: I confirmed it returns the **fork** against the old scoped code (assertion would fail) and the **base** against the fix.

Regression: `test_pr_base_repo` 9/9, `block_unreviewed_merge` 29/29, `require_architecture_review` 22/22, `require_design_review_for_ui` 22/22, `tracker_review_submit` 27/27, `warn_review_marker_write` 13/13, `warn_stale_review_markers` 9/9. Both edited files `bash -n` clean.

New HEAD: `041cf0e`.

AbdElrahmaN31 (commented):
## Code Review: PR #770 (re-review)

**Commit**: `041cf0e3dca608f33dce57d811d6696c406b53fe`

Re-review after the prior CHANGES_REQUESTED (review id 4628487088). Verdict: **APPROVED**. Both requested changes are genuinely resolved and empirically verified.

### Summary
Keys all five approval-marker writers on the PR's **base** repo (matching what the merge gates already look up via the merge command's `--repo`), fixing the cross-fork divergence in #765. This re-review focuses on commit `041cf0e` — the response to my prior review of the `pr_base_repo` helper.

### Prior review — both requested changes resolved

**1. Query gh unscoped so the base resolves (hint = value fallback only).** ✅ Resolved.
`pr_base_repo` in `_lib-review-markers.sh` now runs `gh pr view "$pr" --json url` with **no `--repo`** (line 138). The previous `if [ -n "$hint" ]; then gh pr view --repo "$hint" ...` branch — which looked up a base-numbered PR on the fork, errored, and fell through to the fork hint — is gone. The hint is now purely a value fallback (lines 142-144), documented with a clear WHY-UNSCOPED rationale. gh's ambient base-repo resolution prefers the parent/upstream, which is the key we want.

**2. Make the test's gh mock `--repo`-aware so the cross-fork case is load-bearing.** ✅ Resolved.
`test_pr_base_repo.sh`'s mock now parses `--repo` from argv and `exit 1`s when it names anything other than the base repo in the queued URL (lines 42-49), mirroring real gh rejecting a base-numbered PR looked up on the fork.

I independently confirmed the assertion genuinely discriminates: under the new `--repo`-aware mock, the **old** scoped call `gh pr view 762 --repo <fork>` returns empty (→ fork fallback → the #765 bug), while the **new** unscoped call resolves `me2resh/apexyard`. The cross-fork test would fail against the old code and passes only when the base is genuinely resolved.

### Empirical verification (this PR, #770)
- `gh pr view 770 --json url --jq .url` (unscoped) → `https://github.com/me2resh/apexyard/pull/770` (base) ✅
- `gh pr view 770 --repo AbdElrahmaN31/apexyard` → errors (`Could not resolve to a PullRequest`) ✅ — exactly the failure the old scoped path hit.

### Tests
- `test_pr_base_repo.sh` — **9/9 PASS**
- `test_block_unreviewed_merge.sh` — **29/29 PASS** (same-repo no-op holds)
- `bash -n` clean on both edited files.

### Checklist Results
- Architecture & Design:     Pass — writer/reader agreement is now structural (both key on base); no layering concerns (shell hooks change).
- Code Quality:              Pass — clear WHY-UNSCOPED comment; DDD/type-safety items N/A (shell, not TS).
- Testing:                   Pass — the cross-fork assertion is load-bearing and empirically discriminates old vs new code.
- Security (trust chain):    Pass — see below. A mis-key can only fail **closed**.
- Performance:               Pass — one extra `gh` call, unchanged from before.
- PR Description & Glossary:  Pass — narrative Summary, Testing steps, Glossary all present.
- Technical Decisions (AgDR): N/A — bug fix to the existing AgDR-0060 marker scheme; no new decision.
- Adopter Handbooks:         N/A — always-load handbooks (migration-safety [blocking], clean-architecture, commit-message) don't trigger: no DB migration, no TS domain code; commit messages are exemplary.

### Security (trust-chain / fail-closed analysis)
This PR touches `.claude/hooks/**`. The behavioural fix is entirely writer-side (`pr_base_repo`); the gate change is a 6-line documenting NOTE, not a behaviour change. The gate derives `CMD_REPO` from the merge command's `--repo`/API path — always the base (you cannot merge a fork's copy) — and looks up the marker under `(CMD_REPO, PR)`. If a writer ever mis-resolves the base, the written marker key diverges from the gate's lookup key → the gate finds no marker → **blocks the merge**. Divergence fails **closed**, never open. No path was introduced by which a mis-key permits an unreviewed merge.

### Issues Found
None.

### Suggestions
- The unscoped `gh pr view` relies on gh's ambient remote resolution preferring upstream/base. That holds here (empirically verified) and matches gh's documented behaviour. Worth a mental note that in an exotic single-remote-is-the-fork checkout the ambient resolution could differ — but in that shape the base-numbered PR simply wouldn't resolve and it degrades to the hint, same as any gh failure. Non-blocking.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `041cf0e3dca608f33dce57d811d6696c406b53fe`

atlas-apex (approved):
## Code Review (re-review): PR #770

**Commit**: `222bed84d7c66bd2ace627737658ffd4557eeb2a`

Re-review after the prior **CHANGES REQUESTED** verdict on `d8ad532`. That review raised two blockers; both are now fixed, and I verified the fix **live against this very cross-fork PR** (head `AbdElrahmaN31/apexyard` → base `me2resh/apexyard`) rather than trusting the mock.

### Prior blockers — both resolved

**BLOCKER 1 (fork-scoped query resolved to the FORK) — FIXED.** `pr_base_repo` (`_lib-review-markers.sh`) previously ran `gh pr view "$pr" --repo "$hint"` where `$hint` is the fork, so on a cross-fork PR the base-numbered lookup errored and fell back to the fork. The helper now queries **unscoped** (`gh pr view "$pr" --json url`), relying on gh's ambient base resolution, and treats `$hint` strictly as a *value* fallback — never a query scope. This is exactly the suggested fix, with a `WHY UNSCOPED` comment documenting the regression it avoids.

Live proof in a fork-clone (origin=a fork, upstream=base, no default repo set):
- unscoped `gh pr view 770 --json url` → `https://github.com/me2resh/apexyard/pull/770` (BASE)
- `gh pr view 770 --repo AbdElrahmaN31/apexyard` → `Could not resolve to a PullRequest` (the old failing path)
- `pr_base_repo 770 "AbdElrahmaN31/apexyard"` → `me2resh/apexyard` (BASE; hint correctly ignored)

**BLOCKER 2 (mock masked the defect) — FIXED.** `tests/test_pr_base_repo.sh` now uses a `--repo`-aware `gh` stub: it resolves only when the query is unscoped OR `--repo` names the base in the queued URL, and fails when `--repo` points at the fork — mirroring real gh. The cross-fork case (test #1) is now **load-bearing**: it fails if `--repo "$hint"` scoping is ever reintroduced. 9/9 pass.

### Verification at HEAD
- **Base-repo keying is consistent everywhere.** All eight marker-writer call sites key `review_marker_path` on `$PR_HOST_REPO` (the base): code-reviewer (rex), solution-architect (architecture), /approve-merge (ceo write + rex read + merge `--repo`), /approve-design (design write + rex read), /approve-architecture (architecture write + rex read). Grep confirms **zero** writers left on `$PR_REPO`/headRepository.
- **Backward compat (same-repo) is a provable no-op.** When base == head the URL parse equals the hint, so the marker keys exactly as pre-#765. Confirmed by test #2 + the discriminating same-repo marker-path assertion, and by the unchanged `block_unreviewed_merge` regression (29/29).
- **Cross-fork resolution fail-soft.** gh failure / bad URL / no-pr → value fallback to the hint. On a cross-fork PR that mis-keys onto the fork, which is **fail-closed** (the base gate can't find it → merge stays blocked; never a wrongful approval).
- **Gates untouched.** All three merge-gate hooks received comment-only additions honestly documenting the residual no-`--repo` current-branch fallback (headRepository), which `/approve-merge` never produces since it always passes `--repo`. Reader (`CMD_REPO` = base via `--repo`/API-path) and writer (base via `pr_base_repo`) now agree structurally.
- **Tests / lint.** `test_pr_base_repo.sh` 9/9, `block_unreviewed_merge` 29/29, `tracker_review_submit` 27/27; shellcheck clean on `_lib-review-markers.sh`, `_lib-tracker.sh`, and all three gate hooks.
- **Scope.** The `dev`-based diff carries #762/#763's tracker-routing changes (stacked, both still OPEN) — reviewed on their own PRs and not re-litigated here. #765's own commits: `ce22d8c`, `a4176b1`, `25ede43`, `222bed8`.
- **PR body / Glossary.** Complete: narrative summary, testing, glossary (`pr_base_repo`, base vs head, writer/reader), and the merge-order note.

### Checklist Results
- Testing:                    Pass (9/9 + 29/29 + 27/27, mock now faithful)
- Security:                   Pass (fail-closed; no forge, no traversal, no fail-open — Hakim follows)
- Performance:                Pass (one extra `gh pr view --json url` per marker op)
- Technical Decisions (AgDR): N/A (bug-fix aligning existing behaviour; rationale inline)
- Adopter Handbooks:          N/A (shell + markdown; no handbooks loaded)

### Issues Found
None.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `222bed84d7c66bd2ace627737658ffd4557eeb2a`

atlas-apex (commented):
## Security Review: PR #770 — trust-chain (cross-fork approval-marker keying)

**Commit reviewed**: `222bed84d7c66bd2ace627737658ffd4557eeb2a`
**Scope**: #765's four commits (`ce22d8c`, `a4176b1`, `25ede43`, `222bed8`) — independent security layer after Rex's code-review APPROVE. Advisory (no gate marker).

### Summary
`pr_base_repo <pr> [hint]` is added to `_lib-review-markers.sh` and every marker **writer** is re-keyed onto the PR's **base** repo so writers and the merge **gates** agree on the qualifier. The three gate hooks are unchanged (only NOTE comments added). I threat-modelled the change for gate-bypass, marker confusion/collision, fail-open behaviour, and injection, and ran the trust-chain test suites fresh in an isolated worktree at HEAD.

### Threat-model results

**1. Marker confusion / collision — PASS (with one LOW note).**
- Path-traversal is inherently defanged: `review_marker_path` replaces **every** `/` with `__`, so a hostile slug like `../../etc/pwn` becomes `..__..__etc__pwn` — no surviving separator, no directory escape. Verified.
- On GitHub (the deployment target) org/user names cannot contain `_`, so `owner/repo → owner__repo` is injective — a repo named `apex_yard` still keys unambiguously. No collision.
- LOW / informational (pre-existing, not introduced here): `review_marker_path`'s `/→__` scheme is *not* injective for **nested GitLab group paths** — `a/b/c` and a literal project `a/b__c` both map to `a__b__c`. This lives in the #485/AgDR-0060 sanitiser, not #765; #765 only marginally widens its input domain by letting `pr_base_repo` return multi-segment GitLab bases. Not exploitable on GitHub, and any collision still requires write access to the maintainer's local `.claude/session/reviews/` (a PR author has none). No action required for this PR; worth a follow-up note on the sanitiser if GitLab nested-group support is ever made real.

**2. Fail-closed on resolution failure — PASS (load-bearing property holds).**
When `pr_base_repo` can't resolve (gh down / auth fail / empty), it falls back to the *hint* (the fork) or empty. Critically, this only affects **where a writer writes** — the gates are unchanged and independently derive `CMD_REPO` from the actual merge command's `--repo` value (you cannot merge a fork's copy). A writer mis-resolution therefore keys the marker under the wrong qualifier, the gate looks under the base, doesn't find it, and **blocks** — a false-negative (approved merge stalls), never a false-positive (unapproved merge passes). Confirmed by direct test: gh-fail + hint → returns fork; gh-fail + no-hint → empty. There is no permissive default and no path where a resolution failure loosens the gate.

**3. Unscoped `gh pr view` query — PASS.**
Commit `222bed8` correctly drops `--repo "$hint"` scoping: scoping to the fork looked up a base-numbered PR where it doesn't exist → error → fork fallback → the exact #765 divergence. Unscoped, gh's ambient resolution prefers the parent/base — the desired key. It cannot be steered by the PR author: gh resolves against the *maintainer's* local remotes (not attacker-controlled), and the `.url` it parses is GitHub-canonical (`https://github.com/<base>/pull/<n>`), not author-injectable. In `/approve-merge` the same `PR_HOST_REPO` value feeds the marker write **and** the merge `--repo`, so even an unexpected resolution stays internally consistent and a wrong (fork) value makes `gh pr merge` itself fail — fail-closed.

**4. Injection in slug→filename / shell flow — PASS.**
`pr_base_repo` output is only ever `printf`'d (never `eval`'d); the base value is derived by `sed` from a canonical URL and constrained to URL-path chars. A hostile hint `foo/bar;touch /tmp/PWNED` passed straight through as a literal with **no** file created. Writer consumption in `/approve-merge` (`review_marker_path "$PR_HOST_REPO" …`, `gh pr merge --repo "$PR_HOST_REPO"`) is consistently double-quoted — single-arg, no re-tokenisation.

**5. Carried #762/#766 content — nothing alarming for those PRs.**
The stacked diff carries `_lib-tracker.sh` / `tracker_review_submit`. Its config-templated `eval` paths are the pre-existing AgDR-0072 model (adopter-owned templates, numeric-guarded `{id}`, canonical `{owner_repo}`); the body is passed via env, not eval'd. `test_tracker_review_submit` (27/27) includes an explicit no-shell-injection assertion, which passed.

### Checklist
- Secrets & Credentials: Pass · Injection Prevention: Pass · Auth/gate integrity: Pass · Data Protection: Pass (marker files unchanged in content) · Fail-closed behaviour: Pass

### Test evidence (fresh run, worktree at HEAD `222bed8`)
`bash -n` clean on all four trust-chain files · `test_pr_base_repo` 9/9 · `block_unreviewed_merge` 29/29 · `require_architecture_review` 22/22 · `require_design_review_for_ui` 22/22 · `tracker_review_submit` 27/27 · `warn_review_marker_write` 13/13 · `warn_stale_review_markers` 9/9. Plus an adversarial probe (traversal / injection / fail-closed / collision) — all behaved safely.

### Verdict
**APPROVED (security)** — no CRITICAL/HIGH/MEDIUM findings. The change strengthens the trust chain (structural writer↔gate agreement) and every failure mode is fail-closed. One LOW/informational note on the pre-existing GitLab nested-path sanitiser collision, out of scope for this PR.

---
🛡️ Independent security layer — Hakim (Security Auditor)
📌 Reviewed commit: `222bed84d7c66bd2ace627737658ffd4557eeb2a`

atlas-apex:
Heads-up @AbdElrahmaN31 — #762 just merged to `dev` (as abf0d79) and #766 is queued next. This PR (#765 base-repo marker keying) now conflicts with the updated `dev` (GitHub reports CONFLICTING/DIRTY). Could you rebase it onto the latest `dev` once #766 lands? It's the last of the stack. Thanks for the great work on this series 🙏

AbdElrahmaN31:
Rebased onto the latest `dev` (`731d16c`) — now `mergeable=MERGEABLE`. New head: **`7a8292f`** (was `d07862e`).

Since #762 and #766 had already squash-merged, I used `git rebase --onto upstream/dev a14cfca` to **drop the now-redundant #758/#763 commits** and replay only #765's four own commits. That's why the real conflict was much smaller than GitHub's naive-merge view: three of the four files it flagged (including `_lib-tracker.sh`) were artifacts of those redundant commits and dropped out. The only genuine reconciliation was one hunk in `.claude/skills/approve-merge/SKILL.md`:

- dev's #759/#794 had replaced the hardcoded `gh pr merge … --repo "$PR_HOST_REPO"` with the tracker-agnostic `tracker_pr_merge` wrapper;
- #765's edit targeted that now-deleted line.

I kept dev's `tracker_pr_merge` mechanism and passed `$PR_HOST_REPO` as its repo argument — preserving #765's base-repo intent, and consistent with this skill's own convention (`<owner/repo>` throughout = `$PR_HOST_REPO`, since you can't merge a fork's copy). Every other #765 file is byte-for-byte identical to the pre-rebase tip (verified by comparing net contributions with `index` lines stripped); `code-reviewer.md` / `solution-architect.md` auto-merged cleanly because dev touched different regions.

Trust-chain suites green on the new head: `pr_base_repo` 9/0, `block_unreviewed_merge` 33/0 (picks up dev's new `tracker_pr_merge` cases), `require_architecture_review` 22/0, `require_design_review_for_ui` 22/0, `warn_review_marker_write` 13/0.

Ready for the final Rex + Hakim pass. Thanks — that's the whole stack rebased and landing 🙏

me2resh (commented):
## Security Review: PR #770 — cross-fork approval-marker keying

**Commit reviewed**: `7a8292fe21728663a0910c2e94c74cdda620194e`
**Base**: `dev` · **Contributor**: external (@AbdElrahmaN31) · **Class**: trust-chain (merge-gate approval markers)

### Summary
This PR moves all five approval-marker **writers** (Rex, Solution Architect, `/approve-merge` CEO, `/approve-design`, `/approve-architecture`) from keying markers on `headRepository` (the fork) to keying on the PR's **base** repo, via a new `pr_base_repo <pr> [hint]` helper in `_lib-review-markers.sh`. The three gate **readers** (`block-unreviewed-merge.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh`) already key on the base (the merge command's `--repo` / API-path), so this makes writer↔reader agreement structural instead of coincidental. The gate hooks themselves get **comment-only** additions — no logic change.

### Threat-model results

**1. Can it let an UNREVIEWED merge through? NO — this is the fail-closed direction.**
- The marker key is derived from `gh pr view <pr> --json url`, which is **GitHub-authoritative** and rooted on the **base** (trusted upstream) repo. An external contributor controls the *fork* (head), not the base URL, so the key is **not attacker-influenceable**.
- The fork value (`headRepository`) is used **only as a value-fallback** when the URL can't be parsed / gh fails. In that case the writer keys on the fork while the gate still keys on the base → **mismatch → merge BLOCKED**. Over-blocking, never under-blocking.
- The documented residual edge (a no-`--repo` current-branch manual merge, where the gate's `extract_repo_from_command` falls back to the fork) is also fail-closed: gate looks under fork, marker is under base → not found → blocked. It is not reachable via `/approve-merge` (which always passes `--repo`).
- Reminder for context: these markers are a **local-filesystem** trust mechanism on the maintainer's machine. An external contributor has no way to write them regardless of keying; the change only affects whether the *maintainer's own* agent can find a legitimately-written marker. Direction of change is strictly safer.

**2. Path injection into the marker filename? NO.**
- The base value flows into `review_marker_path`, whose sanitization flattens **every** `/` to `__` (unchanged by this PR). I probed it directly with adversarial inputs:
  - crafted URL `.../../../etc/pull/5` → base `../../etc` → marker `...reviews/..__..__etc__5-rex.approved`
  - hint `/etc/cron.d/x` → marker `...reviews/__etc__cron.d__x__9-rex.approved`
  - hint `../../../etc/passwd` → marker `...reviews/..__..__..__etc__passwd__9-rex.approved`
  All stay **inside** the reviews dir — no `/` survives, so no traversal and no cross-PR marker collision. (Real owner/repo slugs can't contain `..` or extra `/` anyway; the sanitizer is belt-and-braces.) Writer and reader run the same helper, so their paths flatten identically.
- The URL-parsing sed is anchored (`^https?://[^/]+/(.+)/(pull|-/merge_requests)/[0-9]`) and correctly handles GitHub `/pull/`, GitLab nested-group `/-/merge_requests/`, and GHE hosts; unparseable URLs return the hint (base==head no-op).

**3. Does it weaken any existing gate hook? NO.** The three gate hooks receive only `#`-comment additions documenting the base-keying invariant; core resolution logic is untouched. Gates continue to key on the base.

**4. External contributor scrutiny.** The one design-sensitive choice — querying `gh pr view` **unscoped** (no `--repo`) so gh's ambient resolution prefers the upstream/base — is correct and defended in-code; scoping to the fork hint would re-create the very cross-fork divergence #765 fixes. Worst case if run from an unexpected cwd, it resolves the wrong repo and *over-blocks* (fail-closed).

### Verification performed
- New `test_pr_base_repo.sh` (PR-head version): **9/9 pass** (cross-fork→base, same-repo→unchanged no-op, GitLab nested MR, GHE host, gh-fail/bad-URL/no-pr→hint fallback, and marker-path discrimination).
- `bash -n` clean on `_lib-review-markers.sh` and the test.
- Independent adversarial traversal probes (above) — all contained.

### Checklist
- Secrets & Credentials: Pass
- Injection Prevention (path): Pass — slashes flattened, no traversal
- Auth & Authorisation (merge gate): Pass — keyed to base (trusted), fail-closed on every fallback
- Trust-chain integrity: Pass — writer/reader now provably agree; gates unchanged

### Findings
No Critical / High / Medium issues. Two informational notes (non-blocking):
- **LOW / informational** — `pr_base_repo`'s unscoped `gh pr view` depends on cwd git remotes; the skills correctly instruct resolving *before* any `cd` / `gh pr checkout`. Any misresolution is fail-closed. Fine as-is.
- **Informational (pre-existing, not introduced here)** — `review_marker_path`'s `__` separator assumes owner/repo slugs never contain consecutive underscores; GitHub technically permits `a__b` repo names, a latent (repo,pr) collision corner the AgDR-0060 scheme already carries. Out of scope for #765.

### Verdict
**Security: CLEARED (COMMENT).** This is a security *improvement* — it removes an attacker-uninfluenceable-but-misaligned key and replaces it with the base (trusted upstream) repo, matching the gates. Every fallback and residual edge fails closed. No merge-gate bypass, no path injection. Merge-gate approval (Rex marker + per-PR CEO nod) is a separate step and is intentionally NOT provided by this review — I did not write any marker or merge.

---
📌 Reviewed commit: `7a8292fe21728663a0910c2e94c74cdda620194e`

me2resh (commented):
## Code Review: PR #770

**Commit**: `7a8292fe21728663a0910c2e94c74cdda620194e`

### Summary
Aligns all five approval-marker **writers** (rex, architecture ×2, ceo, design) onto the PR's **base** repo — matching what the three merge gates already key their lookup on — fixing the cross-fork divergence where a valid approval was written under the fork qualifier while the gate searched under the base, silently blocking cross-fork merges. Introduces `pr_base_repo <pr> [hint]` in `_lib-review-markers.sh`, which resolves the base by parsing the PR URL (gh exposes no `baseRepository` field), with a hint fallback so same-repo PRs are a provable no-op.

Reviewed as security-sensitive trust-chain code (marker/merge-gate system), held to the full external-contributor bar.

### Checklist Results
- Security:                   Pass (fail-safe verified — see below)
- Performance:                Pass (N/A)
- PR Description & Glossary:   Pass
- Technical Decisions (AgDR): N/A (correctness fix within AgDR-0060's existing repo-qualified scheme; see nit)

### Correctness of the marker keying (the critical check)
Traced the full writer → merge → reader chain and confirmed all three agree on the **base** repo:
- **Writer:** `REX_MARKER = review_marker_path(pr_base_repo(...) = base ...)`
- **Merge (`/approve-merge`):** `tracker_pr_merge "$PR_HOST_REPO"` → gh adapter emits `gh pr merge <pr> --repo <base>` (`_lib-tracker.sh:965`)
- **Gate (`block-unreviewed-merge.sh`):** `CMD_REPO` parsed from `--repo <base>` (`:69`) → marker lookup keyed on base (`:147`)

**Fail-safe direction holds.** Every fragility in `pr_base_repo`'s ambient resolution (gh unavailable, unparseable URL, base-numbered PR not resolvable) falls back to the hint (the fork) → marker written under the fork qualifier → gate (keyed on base) does not find it → **merge blocked**. No failure path opens an unreviewed merge; the worst case is a false-block of a valid merge. That is the correct posture for trust-chain code.

**No-op for same-repo PRs proven** — base == head → hint fallback; `block_unreviewed_merge` regression is 33/33 unchanged, plus the dedicated same-repo assertions in `test_pr_base_repo.sh`.

**Commit 4 (unscoped query) is the right call and load-bearing.** Verified live from an ops fork (origin=fork, upstream=base): `gh pr view 770 --repo <fork>` errors, while unscoped `gh pr view 770` resolves `me2resh/apexyard` (the base) even with no default repo set. The `--repo`-aware test mock genuinely fails against the old scoped code and passes only when the base is resolved — good regression protection against re-introducing `--repo "$hint"` scoping.

### Tests
- `test_pr_base_repo.sh` — 9/9 (cross-fork → base, same-repo unchanged, GitLab nested MR, GHE host, gh-fail / bad-URL / no-pr → hint, marker-path discrimination).
- Regression, run locally on this HEAD: `block_unreviewed_merge` 33/33, `require-architecture-review` 22/22, `require-design-review-for-ui` 22/22, `tracker_review_submit` 27/27, `warn_review_marker_write` 13/13. All edited hooks `bash -n` clean.
- Rebase/merge state: cleanly rebased on `dev` (prerequisites #762 and #766 are already merged); only the four #765 commits and the 10 marker files differ; `mergeStateStatus: CLEAN`.

- **nit: `pr_base_repo` couples base-resolution to the invoking CWD's git remotes.** The unscoped `gh pr view <pr>` relies on gh's ambient fork-network resolution preferring the parent/upstream. This is correct for the standard setup (reviewer/approver running from the ops fork, which is where framework contributor PRs are handled — verified live). But if ever invoked from a CWD whose remotes don't include the base's fork-network (e.g. inside a `workspace/<project>/` clone, or an ops fork lacking an `upstream` remote), it silently falls back to the hint (fork) and re-diverges. It stays fail-safe (blocks, never opens) and same-repo is unaffected, so this is advisory only — but a one-line comment noting the CWD dependency, or a future hardening via `gh api repos/{o}/{r}/pulls/{n} --jq .base.repo.full_name` once owner/repo is known, would make the assumption explicit.
- **nit: the "Stacked on #766 (→ #762)" section of the PR body is now stale.** Both prerequisites are merged and the branch is cleanly rebased, so the "diff currently includes #762's and #763's changes" caveat no longer applies. Trimming it would spare a future reader the confusion. (`.claude/rules/pr-quality.md` narrative-summary guidance — cosmetic.)
- **nit: consider recording the "base repo is the canonical marker key" rule** as a short amendment to AgDR-0060 (or a new AgDR) for discoverability. It's thoroughly documented in code comments, commit messages, and the Glossary already, so this is a discoverability nicety, not a gap — hence AgDR marked N/A rather than blocking.

### Verdict
**APPROVED**

The core keying is correct, the failure modes are all fail-safe, the same-repo path is a proven no-op, and the cross-fork path is verified against this very PR. Marker written keyed on the base repo (`me2resh__apexyard__770-rex.approved`) — dogfooding the fix. Not merging, per instructions.

---
📌 Reviewed commit: `7a8292fe21728663a0910c2e94c74cdda620194e`

me2resh (approved):

PR #773, opened 2026-07-03
fix(#769): agdr-arch-pr hook — origin-first diff base + raw-command marker haystack

### LINKED ISSUE #769: [Bug] require-agdr-for-arch-pr.sh false-blocks PRs on forks behind upstream (diff-base + --body parser)
## Affected
`.claude/hooks/require-agdr-for-arch-pr.sh` — two independent bugs that both surface when an ops fork's `main` is behind `upstream/main` (the normal state right before running `/update`). Present in v3.2.0 and still present on `upstream/main` (v4.2.0) as of filing.

## Given / When / Then

**Bug 1 — diff-base resolution prefers `upstream/<base>` over `origin/<base>`**

**Given** a fork whose `main` is N commits behind `upstream/main` (e.g. mid-`/update`, or any time before a sync)
**When** a PR is created with `--base main` (a normal same-repo PR against the fork's own main)
**Then** the merge-base is computed against `upstream/main` — because the resolver tries `upstream/$BASE_ARG` first — so **every file upstream changed since the fork's last sync** appears in `CHANGED_FILES` as an "architecture change" (migrations/, .github/workflows/, hooks, topologies…), and a docs-only PR touching none of those paths is false-blocked. **Expected:** prefer `origin/$BASE_ARG` (the ref the PR actually merges into) before `upstream/$BASE_ARG`.

```sh
# line ~291 (v3.2.0) / ~337 (v4.2.0)
for candidate in "upstream/$BASE_ARG" "origin/$BASE_ARG" "$BASE_ARG"; do
```

**Bug 2 — `--body` extractor drops the body when the command has a trailing pipe**

**Given** a PR created via `gh pr create ... --body "$(cat <<'EOF' ... EOF)" 2>&1 | tail -3` (command-substituted body + trailing pipe — a very common shape)
**When** the hook parses `--body` with the awk extractor (line ~72), whose double-quoted anchor requires the closing `"` to be followed by whitespace+`--flag` or end-of-string
**Then** with a trailing `)" 2>&1 | tail`, the closing quote is followed by ` 2>&1 …`, so the quoted branches fail and the parser falls through to the unquoted branch, capturing `BODY` as just `"$(cat`. The `<!-- agdr: not-applicable -->` escape-hatch marker (and the whole body) is never seen → the PR is **blocked despite the marker being present**. **Expected:** the closing-quote anchor should also accept a following `)`, pipe, or redirect. **Workaround:** pass the body via `--body-file`.

```awk
# line ~72
re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
```

## Repro
1. On a fork whose `main` is behind `upstream/main`, create a docs-only branch.
2. `gh pr create --base main --title "docs(#N): ..." --body "$(cat <<'EOF' ... <!-- agdr: not-applicable --> ... EOF)" 2>&1 | tail -3`
3. Observe the hook block with upstream's newer files listed as triggering paths, and the escape-hatch marker ignored.

## Framework version
v3.2.0 (also reproduces on `upstream/main` @ v4.2.0)

## Severity
major — false-blocks legitimate PRs on any fork behind upstream. Bug 1 makes the hook fire on non-arch PRs; bug 2 makes the documented escape hatch unusable via the common inline-heredoc invocation, compounding it. Both have workarounds (`--body-file` + the marker) but only once you know the diff-base is wrong.

## Notes
Both encountered during a routine multi-release fork sync (this is the same wrong-base class as the cross-repo guard in #464, but for base-branch ordering rather than target repo).

## Glossary
| Term | Definition |
|------|------------|
| ops fork | A fork of apexyard used as an org's portfolio ops repo, periodically synced from upstream |
| diff-base | The ref the hook diffs HEAD against to compute a PR's changed files |
| escape hatch | The `<!-- agdr: not-applicable -->` body marker that bypasses the AgDR requirement |

atlas-apex:
## Summary

Two independent bugs in `require-agdr-for-arch-pr.sh` (#769), both of which false-block legitimate PRs on a fork behind/diverged-from upstream — the normal state right before `/update`.

- **Bug 1 — diff-base preferred `upstream/<base>` over `origin/<base>`.** A same-repo `--base main` PR merges into the fork's **own** base (`origin/main`), so that's the correct diff base. Trying `upstream/<base>` first meant that on a fork carrying its own baseline customisations (custom hooks/CI upstream lacks), the merge-base was computed against upstream's **diverged** tree, and the fork's baseline arch files were attributed to *this* PR → docs-only PRs blocked. Reordered to `origin/<base>`, `upstream/<base>`, `<base>`.
- **Bug 2 — the `--body` extractor drops multi-line / trailing-operator bodies.** awk's `.` doesn't cross newlines, so a `--body "$(cat <<'EOF' … EOF)"` heredoc — or a `… )" 2>&1 | tail` trailing pipe — leaves `BODY` as a stub like `"$(cat`. The `<!-- agdr: not-applicable -->` escape-hatch marker and any AgDR reference **inside** the body were silently missed → PR blocked despite the marker. **Fix:** the inline body is always a substring of `$COMMAND`, so the raw command is now part of the marker/AgDR-ref haystack — robust to *every* quoting shape.
  - *Note:* the reporter's proposed fix (broaden the closing-quote anchor) only covers the single-line trailing-operator case. I verified with the real awk that a **multi-line heredoc** body fails regardless of the anchor (awk `.` is line-bound), so I fixed it at the haystack instead — bulletproof across both shapes.

## Testing

`bash .claude/hooks/tests/test_require_agdr_for_arch_pr.sh` → **20 pass**, 3 new:
- docs-only PR on a fork **diverged-behind** upstream → PASS (diffs `origin/main`, not `upstream/main`) — Bug 1.
- arch PR with the skip marker in a **heredoc `--body` + trailing pipe** → PASS (bypassed) — Bug 2.
- arch PR with an **AgDR ref** in a heredoc `--body + pipe` → PASS — Bug 2.

**1 pre-existing failure** (`spike active-ticket marker`) is unrelated — it's a **known in-sandbox ops-root limitation documented in the test file itself** (≈ line 311), red on clean `dev` before this change.

Closes #769

---

## Glossary
| Term | Definition |
|------|------------|
| diff-base | The ref the hook diffs HEAD against to compute a PR's changed files |
| diverged fork | A fork whose `main` carries its own commits (baseline customisations) that upstream lacks — so `merge-base(HEAD, upstream/main)` predates them |
| escape hatch | The `<!-- agdr: not-applicable -->` body marker that bypasses the AgDR requirement |
| haystack | The text the hook greps for the skip marker + AgDR references (title + body + body-file + now the raw command) |

### DIFF
--- a/.claude/hooks/require-agdr-for-arch-pr.sh
+++ b/.claude/hooks/require-agdr-for-arch-pr.sh
@@ -124,9 +124,21 @@ if [ -n "$BODY_FILE" ] && [ -f "$BODY_FILE" ]; then
   BODY_FILE_CONTENT=$(cat "$BODY_FILE" 2>/dev/null)
 fi
 
-# Build the body haystack. Title is included so an AgDR reference in the
-# title also satisfies the requirement (reviewers will see it either way).
-HAYSTACK=$(printf '%s\n%s\n%s\n' "$TITLE" "$BODY" "$BODY_FILE_CONTENT")
+# Build the body haystack. Title is included so an AgDR reference in the title
+# also satisfies the requirement (reviewers will see it either way).
+#
+# The RAW COMMAND is included too (#769 bug 2): the --body extractor cannot
+# reliably recover a value that spans newlines (a `--body "$(cat <<'EOF' … EOF)"`
+# heredoc — awk's `.` doesn't cross lines) or is followed by a shell operator
+# (`… )" 2>&1 | tail` — the closing-quote anchor doesn't match a trailing
+# redirect/pipe). In both shapes BODY comes back as a stub like `"$(cat`, so the
+# skip marker and any AgDR reference INSIDE the body were missed and the PR was
+# false-blocked. The inline body is always a substring of $COMMAND, so grepping
+# the command directly makes the marker + AgDR-ref checks robust to every
+# quoting shape. (--body-file bodies live in a file, handled above; the marker
+# there is a distinctive HTML comment, so a raw-command match is not a false
+# bypass risk.)
+HAYSTACK=$(printf '%s\n%s\n%s\n%s\n' "$TITLE" "$BODY" "$BODY_FILE_CONTENT" "$COMMAND")
 
 # ---------------------------------------------------------------------------
 # 2. Skip marker short-circuit.
@@ -333,8 +345,15 @@ resolve_ref() {
 }
 
 if [ -n "$BASE_ARG" ]; then
-  # Try upstream/<arg>, origin/<arg>, <arg> in that order.
-  for candidate in "upstream/$BASE_ARG" "origin/$BASE_ARG" "$BASE_ARG"; do
+  # Try origin/<arg>, upstream/<arg>, <arg> in that order (#769 bug 1).
+  # origin/<arg> FIRST: a same-repo `--base main` PR merges into the fork's OWN
+  # base (origin/main), so that is the correct diff base. Trying upstream/<arg>
+  # first meant that on a fork whose main is behind upstream (the normal state
+  # before /update), the merge-base was computed against upstream's newer tree,
+  # so every file upstream changed since the last sync surfaced as an
+  # "architecture change" in this PR — false-blocking docs-only PRs. upstream
+  # stays as a fallback for a fresh fork whose origin lacks the base ref.
+  for candidate in "origin/$BASE_ARG" "upstream/$BASE_ARG" "$BASE_ARG"; do
     r=$(resolve_ref "$candidate")
     if [ -n "$r" ]; then BASE_REF="$r"; break; fi
   done
--- a/.claude/hooks/tests/test_require_agdr_for_arch_pr.sh
+++ b/.claude/hooks/tests/test_require_agdr_for_arch_pr.sh
@@ -434,6 +434,61 @@ run_case "cd-target arch PR with AgDR ref → PASS [#669]" \
 rm -rf "$TGT_ARCH2"
 
 # ---------------------------------------------------------------------------
+# #769 bug 1 — diff-base resolves origin/<base> before upstream/<base>.
+# On a fork DIVERGED-behind upstream (its own baseline arch customisations that
+# upstream lacks), diffing against upstream/<base>'s merge-base attributes the
+# fork's OWN baseline files to the PR → false arch-block on a docs-only PR.
+# ---------------------------------------------------------------------------
+setup_diverged_fork() {
+  local dir; dir=$(mktemp -d -t agdr-b1.XXXXXX)
+  (
+    cd "$dir" || exit 1
+    git init -q -b main
+    git config user.email t@t.test; git config user.name test
+    echo "company: test" > onboarding.yaml
+    git add onboarding.yaml; git commit -q -m init
+    local p; p=$(git rev-parse HEAD)
+    # Fork baseline: a CI workflow the fork carries but upstream does not.
+    mkdir -p .github/workflows; echo "on: push" > .github/workflows/fork-custom.yml
+    git add .github/workflows/fork-custom.yml; git commit -q -m "fork: baseline ci"
+    git update-ref refs/remotes/origin/main HEAD       # origin/main = fork main
+    # upstream/main diverged from p — does NOT have the fork's workflow.
+    git checkout -q "$p"
+    echo "# upstream" > UPSTREAM.md; git add UPSTREAM.md; git commit -q -m "upstream: unrelated"
+    git update-ref refs/remotes/upstream/main HEAD
+    # Feature branch off the fork's main: docs only.
+    git checkout -q main; git checkout -q -b feature
+    mkdir -p docs; echo "# new" > docs/new.md
+    git add docs/new.md; git commit -q -m "docs: add"
+  )
+  echo "$dir"
+}
+
+B1_DIR=$(setup_diverged_fork)
+run_case "docs-only PR on a fork diverged-behind upstream → PASS (diffs origin/main, not upstream/main) [#769 bug1]" \
+  "$B1_DIR" 0 "" \
+  "gh pr create --base main --title 'docs(#9): update guide' --body 'Docs only. No decisions.'"
+
+# ---------------------------------------------------------------------------
+# #769 bug 2 — the skip marker (and any AgDR ref) inside a heredoc --body with
+# a trailing pipe is honoured. The --body extractor cannot recover a value that
+# spans newlines or is followed by a shell operator; the raw command is now part
+# of the haystack, so the marker is seen regardless of --body quoting shape.
+# ---------------------------------------------------------------------------
+B2_DIR=$(setup_repo c1_base c1_feat)   # feature touches src/domain/ (arch), no AgDR
+B2_CMD=$'gh pr create --base main --title "feat(#10): domain" --body "$(cat <<\'EOF\'\nSome body text.\n<!-- agdr: not-applicable -->\nEOF\n)" 2>&1 | tail -3'
+run_case "arch PR, skip-marker in a heredoc --body with trailing pipe → PASS (bypassed) [#769 bug2]" \
+  "$B2_DIR" 0 "not-applicable marker present" \
+  "$B2_CMD"
+
+# And the AgDR-reference path through the same heredoc+pipe shape → PASS.
+B2b_DIR=$(setup_repo c1_base c1_feat)
+B2b_CMD=$'gh pr create --base main --title "feat(#11): domain" --body "$(cat <<\'EOF\'\nSee AgDR-0009-portfolio-domain for the rationale.\nEOF\n)" 2>&1 | tail -3'
+run_case "arch PR, AgDR ref in a heredoc --body with trailing pipe → PASS [#769 bug2]" \
+  "$B2b_DIR" 0 "" \
+  "$B2b_CMD"
+
+# ---------------------------------------------------------------------------
 # Result
 # ---------------------------------------------------------------------------
 


atlas-apex (commented):
## Security Review: PR #773

**Commit**: `ca3a94003cf0f3ead6350f429ab7d7f0d6386ab5`
**Reviewer**: Hakim (Security Auditor) — advisory pass on an AgDR-enforcement PreToolUse gate

### Summary
Two changes to `require-agdr-for-arch-pr.sh` (a governance gate that forces an AgDR reference when a PR diff touches architecture paths):
1. Diff-base candidate reorder: `origin/<base>` now tried before `upstream/<base>`.
2. The skip-marker / AgDR-ref haystack now includes the raw `$COMMAND` in addition to the extracted title/body/body-file.

I assessed both adversarially for gate-bypass and fail-closed behaviour.

### Verdict: NON-BLOCKING (clean on injection; one minor governance note)

No exploitable gate-bypass. No injection/code-exec surface. Change (1) is sound and actually *improves* diff honesty. Change (2) carries a small, low-probability governance-weakening on the *AgDR-ref* positive check, called out below as a suggestion — not a blocker, because the author is the party the gate nudges and already holds a sanctioned one-line escape hatch.

---

### Change (2) — raw `$COMMAND` in the haystack

**Injection / code-exec: CLEAN.** `$COMMAND` reaches the shell only as grep *data*:
- `HAYSTACK=$(printf '%s\n%s\n%s\n%s\n' … "$COMMAND")` then `echo "$HAYSTACK" | grep -qF -- "$SKIP_MARKER"` and `… | grep -qE 'AgDR-[0-9]+-[a-z0-9-]+'`.
- Never `eval`'d, never word-split into an argument position, never expanded. The grep *patterns* are fixed literals; the haystack is on stdin. `grep -qF --` also blocks option-injection on the marker. No command-injection vector.

**Gate-bypass: NOT a security finding, but a mild fail-open on the AgDR-ref check.**
- The threat model here is "nudge the author to record a decision," not "defend against an external attacker." `$COMMAND` is the *author's own* command — there is no third-party injection path to plant a marker in someone else's `gh pr create`.
- An author who wants to bypass already has the sanctioned one-liner: `<!-- agdr: not-applicable -->` in the body (exit 0 + visible WARN). Being able to place that same distinctive HTML comment elsewhere in the command grants **no new capability** — same deliberate act, same WARN, same audit trail. The skip-marker string is distinctive enough that an *accidental* bypass is effectively impossible.
- The genuinely new surface is the **AgDR-ref** check (`AgDR-[0-9]+-[a-z0-9-]+`) now matching against the whole command. That regex is loosely shaped and *could* match incidental strings that never render in the PR body — e.g. a branch name `feature/AgDR-0009-foo` passed via `--head`, a `cd` path or redirect target containing an AgDR filename, or a trailing `# … AgDR-0009-…` shell comment. In those cases the gate passes but a GitHub reviewer sees **no** AgDR pointer — defeating the check's stated purpose ("the place where reviewers need a pointer to the decision record"). Probability is low (that exact `AgDR-<digits>-<kebab>` shape rarely appears incidentally in a create command), and the direction is under-blocking, not attacker-exploitable — hence non-blocking.

**Suggestion (optional, non-blocking):** the heredoc/trailing-pipe motivation (`--body "$(cat <<'EOF' … EOF)" 2>&1 | tail`) is legitimate and well-documented. Consider scoping the raw-`$COMMAND` addition to the **skip-marker** haystack only (distinctive, zero false-match risk) while keeping the **AgDR-ref** positive check on the extracted `title + body + body-file` content. That preserves the reviewer-visibility guarantee (an AgDR ref must actually be in the rendered PR) without reintroducing the heredoc false-block, since the skip marker is the case the extractor most needs help with. If the team prefers the current uniform haystack for simplicity, that's an acceptable trade given the low incidental-match probability — just recording the reasoning.

### Change (1) — origin-first diff base

**Bypass via controlling `origin/<base>`: NOT exploitable.**
- For a same-repo `--base main` PR, the merge target *is* `origin/main`. Computing `merge-base HEAD origin/<base>` therefore yields exactly the honest diff GitHub itself will show. The prior upstream-first order computed the merge-base against upstream's (often newer) tree, *inflating* the changed-file set with unrelated upstream commits — the false-block this fixes. Origin-first cannot under-report relative to the real merge target.
- The "attacker pushes arch files to `origin/<base>` so they fall out of the PR diff" scenario doesn't grant a bypass: anything already present in the merge target is, by definition, already merged and not part of *this* PR's diff — its own gate fired when it landed. A PR cannot launder its changes by pre-planting them in the base.
- The cross-repo case (fork's `origin` ≠ the `--repo` target, e.g. contributing upstream) is caught earlier by `pr_repo_matches_cwd` → exit 0, so origin-first only runs when `origin/<base>` genuinely is the merge target. The no-`--base` fallback chain is unchanged.

### Fail-closed on ambiguity
Both changes preserve the pre-existing behaviour: unresolvable base → silent exit 0. That fail-open-on-can't-resolve is a prior design choice (cross-repo no-op / PR-create will fail loudly), not introduced or worsened here. Neither change moves the gate from block→allow on any *architecture-detecting* path.

### Checklist Results
- Secrets & Credentials:  Pass (n/a)
- Injection Prevention:   Pass — `$COMMAND` is grep-haystack-only, no eval/expansion
- XSS Prevention:         Pass (n/a)
- Auth & Authorisation:   Pass — gate is author-nudge with sanctioned escape hatch; no privilege boundary crossed
- Data Protection:        Pass (n/a)
- Governance integrity:   Pass with note — minor, low-probability under-block on the AgDR-ref check (change 2)

### Recommendations
1. (Optional) Split the haystack: raw `$COMMAND` for the skip-marker check only; keep AgDR-ref on extracted body/title to guarantee reviewer visibility of the decision record.
2. Test coverage added for both bugs is good; if you adopt rec (1), add a negative test: arch PR whose *command* (not body) contains an `AgDR-0009-…` branch/path string must still BLOCK.

---
📌 Reviewed commit: `ca3a94003cf0f3ead6350f429ab7d7f0d6386ab5`
📎 Advisory — no approval marker written, no merge performed (security is advisory on this PR).

atlas-apex (commented):
## Code Review: PR #773

**Commit**: `ca3a94003cf0f3ead6350f429ab7d7f0d6386ab5`

### Summary
Two independent false-block bugs in `require-agdr-for-arch-pr.sh`, both surfacing on a fork behind/diverged-from upstream (the normal state before `/update`):
- **Bug 1** — diff-base candidate order flipped from `upstream/<base>, origin/<base>, <base>` to `origin/<base>, upstream/<base>, <base>`. A same-repo `--base main` PR merges into the fork's own `origin/main`, so that's the correct merge-base; trying `upstream/<base>` first attributed upstream's post-sync changes to the PR and false-blocked docs-only PRs.
- **Bug 2** — the marker/AgDR-ref haystack now includes the raw `$COMMAND`. The `--body` awk extractor can't recover a value that spans newlines (heredoc) or is followed by a shell operator (`)" 2>&1 | tail`), so the skip marker / AgDR ref inside such a body was missed. The inline body is always a substring of `$COMMAND`, so grepping the command is robust to every quoting shape.

### Checklist Results
- Architecture & Design:     Pass (isolated PreToolUse hook; no layering concerns)
- Code Quality:              Pass (clear, well-commented rationale on both hunks)
- Testing:                   Pass (3 new tests; red pre-fix, green post-fix — verified below)
- Security:                  Pass (haystack is only grep'd with `-qF`/`-qE`, never eval'd)
- Performance:               Pass (negligible)
- PR Description & Glossary: Pass (narrative summary + 4-term glossary)
- Summary Bullet Narrative:  Pass (what + why on every bullet)
- Technical Decisions (AgDR):N/A (hook bug fix; `.claude/hooks/**` is not an agdr_trigger_path, so the hook does not gate its own PR)
- Adopter Handbooks:         N/A (shell diff; architecture/migration-safety handbooks don't apply, no `.ts` files → TS bucket not loaded)
- Fallow Static Analysis:    N/A (not a JS/TS diff)

### Verification performed
Cloned PR HEAD and ran the suite: **20 pass / 1 fail**, matching the PR description. Then reverted BOTH hunks and re-ran: the 3 new cases (`#769 bug1`, two `#769 bug2`) all flip to FAIL (17 pass / 4 fail). This proves the tests genuinely exercise both fixes rather than passing vacuously.

- **(a) Bug-1 reorder** — correct. Same-repo (`origin==base`) → `origin/<base>` resolves first (right base). Genuine cross-fork PR where origin lacks the base ref → `origin/<base>` fails to resolve and it falls through to `upstream/<base>` (preserved). The `setup_diverged_fork` fixture models the real failure: `origin/main` carries a fork-only `.github/workflows/fork-custom.yml`, `upstream/main` diverges without it; pre-fix the workflow leaks into the diff via the upstream merge-base and blocks a docs-only PR.
- **(b) Bug-2 false-bypass risk** — assessed low/acceptable. Both signals are distinctive: the skip marker is a literal HTML comment and the AgDR ref is a specific `AgDR-\d+-[a-z0-9-]+` shape. To cause a *false* bypass either token would have to appear in the author's own `gh pr create` invocation *outside* title/body (a flag value, repo slug, or branch name) — implausible in practice. The pattern already matched author-controlled title+body; `$COMMAND` only adds the non-body flags. No realistic false-negative.
- **(c) Tests** — red→green transition confirmed empirically (above). Heredoc+trailing-pipe fixtures reproduce the exact quoting shape the extractor drops.
- **(d) Pre-existing failure** — the single remaining FAIL (`spike active-ticket marker`) fails in BOTH the pre-fix and post-fix trees, its test (≈ line 280) is untouched by this diff, and it's documented in-file at lines 311-316 as a known in-sandbox ops-root resolution limitation. Genuinely pre-existing, not introduced here.
- **(e)** Narrative summary + glossary both present and high quality.

- **nit (Bug 1):** one narrow shape isn't a strict improvement — a fork→upstream PR whose branch was cut from `upstream/<base>` while `origin/<base>` also exists and lags will now merge-base against `origin/<base>` and surface a *larger* diff. That direction over-blocks (asks for an AgDR, waivable via the skip marker) rather than bypasses, so it's the safe failure mode and the right trade for the dominant same-repo case. Worth a mental note if a future report describes an over-block on an upstream-based branch.
- **nit (Bug 2):** `--body-file docs/agdr/AgDR-0009-foo.md` would now satisfy the AgDR check via the file *path* in `$COMMAND` even if the file's content lacks a ref. Harmless — when the PR body literally is an AgDR file the requirement is morally met — but noting the mechanism.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `ca3a94003cf0f3ead6350f429ab7d7f0d6386ab5`

me2resh (approved):

PR #778, opened 2026-07-04
feat(#777): auto-invoke the Security Auditor on trust-chain changes (.claude/hooks/**, settings.json)

### LINKED ISSUE #777: [Feature] Auto-invoke the security-review agent on trust-chain changes (.claude/hooks/**), not just auth/crypto/secrets paths
## Problem

The adversarial **security-review** agent (Hakim / `security-auditor`) auto-fires only when a PR diff touches the **auth / crypto / secrets / PII** paths — per `.claude/rules/role-triggers.md`:

```
PR diff touches **/auth/**, **/crypto/**, **/secrets/**, .env*  →  Security Auditor
```

But the framework's **most security-critical code is under `.claude/hooks/`** — the merge-gate trust chain (`block-unreviewed-merge.sh`, `require-*-review.sh`, `_lib-extract-pr.sh`, `_lib-review-markers.sh`), the credential-broker scripts, and the tracker/forge library (`_lib-tracker.sh`) that shells out and, in places, `eval`s. **None of those paths match the security-auditor trigger.** So a change to the trust chain gets only the *generalist* code review (Rex) by default — never the adversarial security lens — unless a human remembers to run `/security-review` manually.

## Evidence (this is not hypothetical — it just happened)

Three security bugs recently reached review in trust-chain PRs. The generalist code reviewer (Rex) caught one; the other **two were APPROVED by Rex and only caught by a manually-invoked security review**:

| PR | Bug | Rex (auto) | Needed the security specialist |
|----|-----|-----------|-------------------------------|
| tracker migration-gate | `eval` command-injection / RCE (reproduced) | caught it | (confirmed) |
| docs-index hook (`maintain-docs-index.sh`) | **path-traversal write** (unnormalised `$(pwd)/$path` + a `case` glob whose `*` matches `/`) | **approved — missed it** | **only the security pass caught it** |
| forge-aware merge gates (`_lib-extract-pr.sh`) | **fail-open**: the GitLab raw-API merge shape left ungated (half-closed the #47 fix) | **approved — missed it** | **only the security pass caught it** |

The path-traversal and the fail-open are security-*reasoning* flaws (glob-matches-slash; needing knowledge of the #47 both-shapes-must-be-gated precedent), not code smells — exactly what the security-auditor lens exists for and a code-quality pass isn't tuned to surface. Because the trigger didn't match `.claude/hooks/`, the specialist never ran automatically on the contributor's own governed run.

There is a mild irony worth naming: **the framework's own security infrastructure is the code least covered by the framework's own security-review trigger.**

## Proposed behaviour

Extend the security-auditor auto-trigger (in `.claude/rules/role-triggers.md` **and** the mechanical backstop `.claude/hooks/detect-role-trigger.sh`) to treat **trust-chain paths** as security-sensitive, so any change to them auto-invokes `/security-review` alongside Rex. At minimum:

- `.claude/hooks/**` — merge gates, PR-extraction lib, review-marker lib, credential-broker scripts, tracker/forge lib
- `.claude/settings.json` — the matcher wiring that decides *whether* a gate fires at all (a too-narrow matcher = an ungated action)
- (adopter-configurable) a `security_review_paths` list in `project-config.*.json` so teams can add their own trust-chain paths

Consider making the security review **advisory-but-loud** on these paths (a banner + a strong nudge) rather than a hard gate, consistent with how the security auditor is advisory today — the goal is that the specialist *always gets a look at trust-chain changes*, not that it blocks.

## Why it helps adopters

This is a process fix, not a point fix: it closes the gap that let real vulnerabilities (an RCE and a merge-gate fail-open) pass the automated review. Every adopter who customises hooks, wires their own gates, or extends the tracker/broker layer gets the adversarial security pass automatically on exactly the code where a mistake is most dangerous — instead of relying on someone remembering to ask for it.

## Scope / out of scope

- **In scope:** extend the trigger path set (rule + `detect-role-trigger.sh`), a test asserting a `.claude/hooks/**` diff fires the security-auditor banner, doc update in `role-triggers.md`.
- **Out of scope:** the individual bug fixes (tracked separately); making the security review a hard *blocking* gate (it stays advisory).

## Framework version
Observed on the current `dev` line (v4.2.0-era).

## Glossary
| Term | Definition |
|------|------------|
| trust chain | The hooks that decide whether an action (esp. a merge) is permitted — the code an attacker most wants to weaken |
| security-auditor trigger | The `role-triggers.md` rule that auto-invokes `/security-review` on matching diffs |
| generalist vs specialist review | Rex (code quality, always-on) vs Hakim (adversarial security, path-triggered) |

Surfaced by a review sweep where the security specialist, run manually, caught two trust-chain bugs the always-on code review had approved.

atlas-apex:
## Summary

- **The adversarial Security Auditor now auto-fires on trust-chain changes.** It previously auto-fired only on `auth` / `crypto` / `secrets` / `.env*` diffs — but the framework's most security-critical code is its **own enforcement layer**: the hooks under `.claude/hooks/` (merge gates, the tracker/forge lib that shells out + `eval`s, the review-marker lib, credential-broker scripts) and `.claude/settings.json` (the matcher wiring that decides *whether a gate fires at all*). None of those matched the trigger, so trust-chain changes got only the **generalist** code review by default — never the specialist's adversarial lens.
- **This isn't hypothetical.** A recent review sweep found a **path-traversal write** and a **fail-open merge gate** that the generalist reviewer *approved* and only a *manually-invoked* security review caught. Extending the trigger would have surfaced the specialist automatically on the contributor's own governed run — before the PR ever reached a human.
- **`detect-role-trigger.sh`** — a new trust-chain `case` fires the Security Auditor banner on `.claude/hooks/**` + `.claude/settings.json`. Advisory (exit 0), same shape as the existing path triggers — **over-triggering is cheap** (the auditor no-ops on a false positive); *under*-triggering is what let the two bugs through.
- **`role-triggers.md`** — trust-chain paths added to the Security Auditor rows (activation table, auto-activation signals, wired-triggers table).

## Testing

`bash .claude/hooks/tests/test_detect_role_trigger.sh` → **42/0**. Four new cases: `.claude/hooks/*`, `.claude/settings.json`, and a nested `*/.claude/hooks/*` (managed-project fork) each fire the Security Auditor; a non-hook `.claude/skills/*/SKILL.md` path stays **silent** (no over-broad `.claude/**` match). `bash -n` clean.

## Design notes
- **Advisory, not a hard gate** — consistent with how the Security Auditor works today. The goal is that the specialist *always gets a look* at trust-chain changes, not to block them.
- **Scope: the trust chain, not all of `.claude/`** — skills/docs/agents don't fire it; only the hooks + the settings matcher wiring do.
- **Fittingly self-dogfooding:** this PR itself edits `.claude/hooks/detect-role-trigger.sh`, so once merged the rule would flag *its own kind* of change for a security look — which is exactly why this PR is going through a security review.
- **Follow-up (out of scope):** an adopter-configurable `security_review_paths` key so teams can add their own trust-chain paths — noted in #777, deferred.

Closes #777

---

## Glossary
| Term | Definition |
|------|------------|
| trust chain | The hooks + settings that decide whether an action (esp. a merge) is permitted — the code an attacker most wants to weaken |
| generalist vs specialist review | The always-on code reviewer (code quality) vs the path-triggered security auditor (adversarial security) |
| advisory trigger | Emits a banner/nudge to run the specialist; never blocks the tool call |

### DIFF
--- a/.claude/hooks/detect-role-trigger.sh
+++ b/.claude/hooks/detect-role-trigger.sh
@@ -184,6 +184,25 @@ detect_path_triggers() {
       ;;
   esac
 
+  # Security Auditor — the TRUST CHAIN (me2resh/apexyard#777).
+  # The framework's most security-critical code is its own enforcement layer:
+  # the hooks that decide whether an action (esp. a merge) is permitted, and the
+  # settings.json matcher wiring that decides whether a gate fires at all. These
+  # are NOT auth/crypto/secrets paths, so the trigger above misses them — yet a
+  # subtle bug here (a path-traversal write, a fail-open gate, an injection in
+  # the tracker/broker lib) is exactly what the adversarial security lens is for.
+  # Over-triggering is cheap (the auditor no-ops on a false positive); leaving
+  # trust-chain edits to the generalist review alone is the gap #777 closes.
+  case "$rel" in
+    .claude/hooks/*|*/.claude/hooks/*|\
+    .claude/settings.json|*/.claude/settings.json)
+      emit_banner \
+        "Security Auditor" \
+        "roles/security/security-auditor.md" \
+        "edit touches the security-critical trust chain ($rel) — run /security-review"
+      ;;
+  esac
+
   # Platform Engineer — CI/CD pipelines + golden-path templates
   case "$rel" in
     .github/workflows/*|*/.github/workflows/*|\
--- a/.claude/hooks/tests/test_detect_role_trigger.sh
+++ b/.claude/hooks/tests/test_detect_role_trigger.sh
@@ -133,6 +133,33 @@ in=$(jq -nc \
   '{hook_event_name:"PreToolUse", tool_name:"Edit", tool_input:{file_path:$p}}')
 run_case "path trigger: ordinary path is silent" 0 "" "$in"
 
+# 2e-i. Trust chain (#777): a merge-gate hook fires Security Auditor.
+in=$(jq -nc \
+  --arg p ".claude/hooks/block-unreviewed-merge.sh" \
+  '{hook_event_name:"PreToolUse", tool_name:"Edit", tool_input:{file_path:$p}}')
+run_case "trust chain: .claude/hooks/* fires Security Auditor [#777]" 0 \
+  "ROLE TRIGGER: Security Auditor.*roles/security/security-auditor\\.md" "$in"
+
+# 2e-ii. Trust chain (#777): settings.json (the matcher wiring) fires Security Auditor.
+in=$(jq -nc \
+  --arg p ".claude/settings.json" \
+  '{hook_event_name:"PreToolUse", tool_name:"Write", tool_input:{file_path:$p}}')
+run_case "trust chain: .claude/settings.json fires Security Auditor [#777]" 0 \
+  "ROLE TRIGGER: Security Auditor" "$in"
+
+# 2e-iii. Trust chain (#777): a nested hooks path (e.g. a managed-project fork) fires too.
+in=$(jq -nc \
+  --arg p "workspace/proj/.claude/hooks/_lib-tracker.sh" \
+  '{hook_event_name:"PreToolUse", tool_name:"Edit", tool_input:{file_path:$p}}')
+run_case "trust chain: */.claude/hooks/* fires Security Auditor [#777]" 0 \
+  "ROLE TRIGGER: Security Auditor" "$in"
+
+# 2e-iv. A .claude path that is NOT trust-chain (a skill doc) stays silent.
+in=$(jq -nc \
+  --arg p ".claude/skills/roadmap/SKILL.md" \
+  '{hook_event_name:"PreToolUse", tool_name:"Edit", tool_input:{file_path:$p}}')
+run_case "trust chain: non-hook .claude path is silent [#777]" 0 "" "$in"
+
 # 2f. Edit on .github/workflows/ci.yml → Platform Engineer banner.
 in=$(jq -nc \
   --arg p ".github/workflows/ci.yml" \
--- a/.claude/rules/role-triggers.md
+++ b/.claude/rules/role-triggers.md
@@ -21,7 +21,7 @@ ApexYard ships **20 role definitions** in `roles/{department}/`. They are not al
 | **UI Designer** | `roles/design/ui-designer.md` | Visual design · component specifications · design tokens · pixel-level work |
 | **UX Designer** | `roles/design/ux-designer.md` | User flows · information architecture · usability review · wireframing |
 | **Head of Security** | `roles/security/head-of-security.md` | Security strategy · threat model · compliance call · cross-project security architecture |
-| **Security Auditor** | `roles/security/security-auditor.md` | **PR touches auth / crypto / user data / secrets** · SAST findings · OWASP review · dependency vulnerability |
+| **Security Auditor** | `roles/security/security-auditor.md` | **PR touches auth / crypto / user data / secrets** · **the security-critical trust chain (`.claude/hooks/**`, `.claude/settings.json`) — #777** · SAST findings · OWASP review · dependency vulnerability |
 | **Penetration Tester** | `roles/security/penetration-tester.md` | Active testing · exploit discovery · API security review · pre-release security sign-off |
 | **Head of Data** | `roles/data/head-of-data.md` | Analytics strategy · data governance · reporting architecture · cross-project data modelling |
 | **Data Analyst** | `roles/data/data-analyst.md` | SQL queries · dashboards · A/B-test analysis · metric investigation |
@@ -81,6 +81,7 @@ This is a **prose convention**, not a mechanically-enforced format. The sibling 
 |--------|----------|
 | Ticket moved to `qa` label | QA Engineer |
 | PR diff touches `**/auth/**`, `**/crypto/**`, `**/secrets/**`, `.env*` | Security Auditor |
+| Edit/PR touches the trust chain — `.claude/hooks/**` (merge gates, tracker/forge lib, review-marker lib, credential-broker scripts) or `.claude/settings.json` (the matcher wiring that decides whether a gate fires) | Security Auditor (#777) |
 | PR diff touches `.github/workflows/**`, `golden-paths/pipelines/**` | Platform Engineer |
 | PR diff touches `docs/agdr/**` or adds a new dependency | Tech Lead |
 | Edit/PR touches a design artifact (technical design, migration AgDR, feature spec / PRD) | Solution Architect |
@@ -159,6 +160,7 @@ Triggers wired in v1 (me2resh/apexyard#206):
 |----------------|------------|-----------|------|
 | Label-based  | `PreToolUse` on `Bash` (matcher: `gh issue edit *`) | `--add-label qa` (single or comma list) | QA Engineer |
 | Diff/path    | `PreToolUse` on `Edit` / `Write` / `MultiEdit` | path contains `auth/`, `crypto/`, `secrets/`, `.env*` | Security Auditor |
+| Diff/path    | same | path under `.claude/hooks/` or matches `.claude/settings.json` (the trust chain — #777) | Security Auditor |
 | Diff/path    | same | path under `.github/workflows/` or `golden-paths/pipelines/` | Platform Engineer |
 | Diff/path    | same | path under `docs/agdr/` | Tech Lead |
 | Diff/path    | same | path matches a design artifact (`*technical-design*.md`, `*tech-design*.md`, `**/designs/**`, `**/prds/**`, `*prd*.md`, `*feature-spec*.md`, `docs/agdr/*migration*.md`) | Solution Architect |


atlas-apex (commented):
## Security Review: PR #778

**Commit**: `64c3ff5ead0b3c9451005bd15ad8f6fbbec9f2c6`

▸ Activating Hakim (Security Auditor) for PR #778 (trigger: diff touches `.claude/hooks/**` — and fittingly, this PR *is* the trust-chain change it adds the rule for)

### Summary
Adds a Security Auditor auto-trigger on the framework's trust chain — `.claude/hooks/**` and `.claude/settings.json` — to `detect-role-trigger.sh`. Purely additive advisory logic: a new standalone `case` block that calls `emit_banner` (stderr only) and nothing else. The hook remains exit-0-always and non-blocking. LOW inherent risk (nudges, does not gate). Reviewed against the four vectors called out plus the exit-0 contract.

### Checklist Results
- Secrets & Credentials:  Pass (no secrets; touches no credential path)
- Injection Prevention:   Pass (see vector 1)
- XSS Prevention:         N/A (shell hook, stderr banner)
- Auth & Authorisation:   Pass (adds a trigger, alters no gate)
- Data Protection:        N/A
- API Security:           N/A

### Vectors verified

**1. `$rel`-into-banner is injection-safe.** `$rel` flows JSON → jq → `FILE_PATH` → `path` → `rel` → `emit_banner $3` (`reason`) → `printf '...(%s)...' "$reason" >&2`. Inert for two independent reasons: (a) `$reason` is a **data argument to `%s`**, never the format string, so even `%n`-class format attacks can't land; (b) bash runs command substitution only on literal source text, never on a variable's *contents* — a filename like `$(touch x)` captured from JSON is a plain string and is never re-evaluated. No `eval`, no unquoted expansion into command position anywhere along the path. The only filesystem read (`lookup_role_class`) uses the **hardcoded** `roles/security/security-auditor.md`, not `$rel`. **Safe.**

**2. No weakening of an existing control; exit-0 contract holds.** The new block is a standalone `case` inserted between the existing auth/crypto/secrets case and the Platform case — independent `case` statements (not `elif`), so it's additive and both can fire; the existing security trigger is untouched. `emit_banner` only does `printf ... >&2` (returns 0). `set -u` satisfied (`$rel` always set), no `set -e`, terminal `exit 0` unchanged. A hook that started blocking would be a DoS on tool calls — that failure mode is not introduced here. **Holds.**

**3. Scope is correct (defense-in-depth).** Case globs cross `/`, so `.claude/hooks/block-unreviewed-merge.sh` matches `.claude/hooks/*`, nested `tests/…` matches, and a forked `workspace/proj/.claude/hooks/_lib-tracker.sh` matches `*/.claude/hooks/*`. Merge gates, tracker/forge lib, review-marker lib, credential-broker scripts, and the settings.json matcher wiring are all reliably covered. Not too narrow — it does not re-open the #777 gap for the paths it names.

**4. No normalization-evasion under-trigger.** Even if `$repo_root` resolution fails and an absolute path survives normalization, the floating `*/.claude/hooks/*` form still matches — failed normalization cannot cause a missed fire. Residual over-firing is explicitly acceptable for an advisory nudge.

### Recommendations (LOW — non-blocking, optional follow-ups)
The trust chain arguably also includes two path families this PR leaves out, both of which influence what gates enforce:
- **`.claude/project-config*.json`** — its `ui_paths_exclude`, `migration_paths`, `ticket.create_command_patterns`, and `leak_protection.public_framework_repos` keys directly tune which security gates fire and on what. An edit here can weaken a control without touching a hook script.
- **`.claude/agents/code-reviewer.md`** — Rex is itself a review gate; weakening its instructions weakens the review pass.

Both are reasonable fast-follows; scoping this PR to the mechanical-enforcement core (hooks) + the matcher wiring (settings.json) is the right first cut.

### Security Issues Found
None.

### Verdict
**COMMENT — no security concerns; safe to proceed on the security axis.** All four vectors check out: the `$rel`-into-banner path is injection-safe (data-arg to `%s`, no re-evaluation of variable contents), the additive block preserves the exit-0/non-blocking contract without altering the existing auth trigger, the glob reliably covers the security-critical paths, and the floating-glob backstop prevents normalization evasion. (Using COMMENT rather than APPROVE — cannot approve own-account PRs; no merge, no marker written.)

---
📌 Reviewed commit: `64c3ff5ead0b3c9451005bd15ad8f6fbbec9f2c6`

atlas-apex (commented):
## Code Review: PR #778

**Commit**: `64c3ff5ead0b3c9451005bd15ad8f6fbbec9f2c6`

### Summary
Extends the advisory `detect-role-trigger.sh` hook so the Security Auditor (Hakim) auto-fires on **trust-chain** edits — `.claude/hooks/**` (merge gates, tracker/forge lib, review-marker lib, credential-broker scripts) and `.claude/settings.json` (the matcher wiring that decides whether a gate fires at all). Previously the auditor only auto-fired on `auth`/`crypto`/`secrets`/`.env*` paths, so the framework's own enforcement layer got only the generalist review. Adds a new independent `case` block, 4 test cases, and 3 consistent rows across `role-triggers.md`.

### Checklist Results
- Testing:                   Pass — 42/0 at HEAD; `bash -n` clean
- Security:                  Pass (this PR *strengthens* the security posture)
- Performance:               Pass (one extra `case` glob per Edit/Write; negligible)
- PR Description & Glossary:  Pass — narrative summary + 3-term glossary
- Summary Bullet Narrative:   Pass — bullets carry what + why, well above the label-only heuristic
- Technical Decisions (AgDR): N/A — a trigger-table extension that reuses the existing path-trigger mechanism identically; no new pattern/decision. Agree with author that no AgDR is warranted.
- Adopter Handbooks:          N/A — always-load handbooks (clean-architecture, migration-safety, commit-quality) have no trigger patterns matching a shell-hook glob change + markdown doc rows; none are `ENFORCEMENT: blocking`.
- Fallow Static Analysis:     N/A — diff is shell + markdown, not JS/TS

### Verification performed (independent, at PR HEAD)
Ran a worktree at `64c3ff5` and probed the new glob directly:

**Fires (trust chain) — confirmed:**
- `.claude/hooks/block-unreviewed-merge.sh`, `.claude/hooks/tests/test_x.sh`, `.claude/hooks/_lib-tracker.sh`
- `workspace/proj/.claude/hooks/_lib-tracker.sh` (nested managed-project fork — the `*/.claude/hooks/*` arm)
- `.claude/settings.json`, `workspace/proj/.claude/settings.json`

**Stays silent (correctly scoped to the trust chain, not all of `.claude/`) — confirmed:**
- `.claude/skills/roadmap/SKILL.md`, `.claude/agents/code-reviewer.md`, `.claude/rules/pr-workflow.md`, `.claude/project-config.json`, `src/hooks/useAuth.ts`

The `case` is anchored on `.claude/hooks/` / `.claude/settings.json`, not `.claude/*`, so it does **not** over-match skills/agents/docs — exactly as the PR claims.

**No regression:** the existing `auth`/`crypto`/`secrets`/`.env*` Security Auditor case and the Platform / Tech-Lead / Solution-Architect cases are byte-for-byte unchanged. The new block is a separate independent `case` (no fall-through), and the hook still `exit 0`s on the trust-chain path — advisory contract holds (`exit=0` verified). No double-emission: `.claude/hooks/check-secrets.sh` fires exactly once (the filename `check-secrets.sh` is not a `secrets` path segment, so only the trust-chain case matches).

**Banner shape:** reuses `emit_banner` with `roles/security/security-auditor.md`, which the file's `**Class**: isolated-work-class` + `agent_slug_for()` correctly resolves to `subagent_type: security-reviewer` (the Hatim→Hakim consolidation). Reason text is accurate: *"edit touches the security-critical trust chain (<path>) — run /security-review"*.

**Tests:** all 4 new cases genuinely assert behaviour (fire on hooks path / settings.json / nested fork; silent on a non-hook `.claude/skills/**` path), not just exit code. Full suite green: **PASS: 42  FAIL: 0**.

**Doc consistency:** all three `role-triggers.md` Security Auditor tables (activation, auto-activation signals, wired-triggers) now list the trust-chain paths, consistent with the hook's actual glob.

### Issues Found
None.

### Suggestions (nits — non-blocking)
- **nit:** `.claude/settings.local.json` isn't covered by the glob. This is almost certainly the right call — `settings.local.json` is git-ignored local overrides and never appears in a committed PR diff — but if a future adopter checks in a local settings variant, it won't fire. Worth a one-line comment near the `case` noting the deliberate scope-to-`settings.json` choice.
- **nit:** the doc rows use `**`-globstar notation (`.claude/hooks/**`) while the hook uses a case-glob `*` (which already crosses `/`). Semantically equivalent; purely cosmetic. No action needed.

### Verdict
**APPROVED**

The change is tightly scoped, well-tested, regression-free, and preserves the exit-0 advisory contract. It closes a real gap (the trust chain was reviewed only by the generalist) with a mechanism identical to the existing path triggers. Self-dogfooding is a nice touch — this PR edits `.claude/hooks/**`, so post-merge the rule would flag its own class of change.

---
📌 Reviewed commit: `64c3ff5ead0b3c9451005bd15ad8f6fbbec9f2c6`

me2resh (approved):

PR #792, opened 2026-07-05
fix(#791): pr-create validator — expand tilde before git -C + stop over-matching gh issue create

### LINKED ISSUE #791: fix(GH): PR-creation validator: tilde cd-target not expanded + issue-title over-match
## Summary

Two related bugs in the PR-creation validation path (`.claude/hooks/validate-pr-create.sh` + `.claude/hooks/_lib-pr-repo.sh`) blocked legitimate work today — a tilde-path worktree false-blocked a valid PR, and the validator's own gate over-matched an unrelated `gh issue create` call.

## Bug 1 — tilde not expanded before `git -C`

`_lib-pr-repo.sh`'s `pr_cmd_cd_target` extracts the literal path text from a leading `cd <path> && …` prefix. For a command like `cd ~/Projects/foo && gh pr create …`, it returned the literal string `~/Projects/foo` and every caller (`validate-pr-create.sh`, `validate-branch-name.sh`, `require-agdr-for-arch-pr.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh`) handed that straight to `git -C`.

`git -C` does **not** shell-expand tilde — it treats `~` as a literal path-name character. So `git -C "~/Projects/foo"` failed silently, the cd-target resolution came back empty, and the branch-ticket-ID fallback used the hook's own cwd instead (typically the ops-fork root, on `dev`), producing a false `Branch 'dev' missing ticket ID` block for a perfectly valid PR from a tilde-path worktree.

## Bug 2 — gate over-matches `gh issue create` (and similar) via an unanchored substring check

The gate that decides whether `validate-pr-create.sh` should run at all stripped only the `--body` / `--body-file` / `-F` payload before testing for the command verb, then ran an **unanchored** `grep -qE '\bgh[[:space:]]+pr[[:space:]]+create\b'` over what remained. That pattern matches "gh pr create" *anywhere* in the leftover text — including inside a `--title` value, which was never stripped.

Reproduced live: a `gh issue create --title '... gh pr create over-matches gh issue create ...'` (i.e. a bug report *about this exact bug*) got the full PR-only validation — title-format regex, missing `## Testing` / `## Glossary`, branch-ticket-ID check — even though the invoked command was `gh issue create`, not `gh pr create`. This is the exact recursive block that made filing this very issue non-trivial: any title that describes the bug in plain English trips the bug.

## Fix (already implemented, PR incoming)

- `_lib-pr-repo.sh`: added `_pr_repo_expand_tilde`, which expands a leading `~` / `~/…` (via `$HOME`) or `~user` / `~user/…` (via a validated-safe `eval echo`, refusing anything that isn't a plain username shape) to an absolute path. `pr_cmd_cd_target` now returns the expanded path, fixing all five callers in one place.
- `validate-pr-create.sh`: the gate now strips one optional leading `cd <path> &&` / `cd <path>;` prefix and anchors the "gh pr create" check to the command **head** (`^[[:space:]]*gh[[:space:]]+pr[[:space:]]+create\b`), so the phrase must be the actual invoked verb, not merely present as a substring in a flag's value.
- Two new regression tests added under `.claude/hooks/tests/`.

## Testing

- New tests: `test_pr_create_tilde_cd_target.sh` (2 cases), `test_pr_create_gate_anchor.sh` (6 cases) — all pass.
- Full existing `validate-pr-create` / `_lib-pr-repo` test suites re-run — all pass (one pre-existing, unrelated failure in `test_require_agdr_for_arch_pr.sh` confirmed present on unmodified `upstream/dev` before this change).
- `shellcheck` clean on both changed hook files.

## Glossary

| Term | Definition |
|------|------------|
| Gate (in a hook) | The early-exit check a PreToolUse hook runs to decide whether the rest of its validation logic applies to the current command at all |
| Anchored match | A regex match required to start at a specific position (here: the actual command head), as opposed to matching anywhere in the string |
| cd-target | The directory a `cd <path> && …`-prefixed command is about to run in — extracted by hooks because the PreToolUse hook fires *before* the real shell executes the `cd` |

atlas-apex:
## Summary

- **Tilde-expand the cd-target before `git -C`** — `pr_cmd_cd_target` (in `_lib-pr-repo.sh`) extracted the literal text of a leading `cd <path> && …` prefix, so a command like `cd ~/Projects/foo && gh pr create …` handed `git -C` the literal string `"~/Projects/foo"`. `git -C` never shell-expands tilde, so the lookup failed silently, the cd-target resolution came back empty, and every caller's branch-ticket-ID fallback used the hook's own cwd instead (typically the ops-fork root, on `dev`) — false-blocking a perfectly valid PR from a tilde-path worktree with "Branch 'dev' missing ticket ID". Fixed once, centrally, in the shared lib — all five callers (`validate-pr-create.sh`, `validate-branch-name.sh`, `require-agdr-for-arch-pr.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh`) get the fix for free.
- **Anchor the PR-create gate to the command head, not an unanchored substring** — the gate that decides whether `validate-pr-create.sh` runs at all originally stripped only `--body`/`--body-file`/`-F` before testing for `gh pr create`, then ran a plain `grep -qE '\bgh[[:space:]]+pr[[:space:]]+create\b'` with no position anchor. That matched the phrase *anywhere* left in the command — including inside an untouched `--title` value. Reproduced live: filing a bug report about this exact bug (`gh issue create --title '...gh pr create over-matches gh issue create...'`) got full PR-only validation despite the invoked command being `gh issue create`, not `gh pr create`. Fixed by anchoring the match to the command head.
- **`~user` allowlist hardened against multi-line injection (Hakim's first review)** — `_pr_repo_expand_tilde`'s username check was a per-line `grep -qE '^[A-Za-z0-9._-]+$'`, which matches on the *first line* of a multi-line value and would let an embedded newline carry a smuggled command into the `eval` below it. Not reachable through the current sole caller (its output is always single-line), but a latent hole for any future caller. Swapped for bash's whole-string `[[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]`, which anchors against the *entire* value and rejects any embedded newline outright.
- **Gate completeness restored WITHOUT reopening the anchor bug, via check-then-strip (Hakim's second review)** — anchoring to the head after stripping only one leading `cd <path> &&` prefix (the previous fix) missed genuine `gh pr create` invocations behind any *other* prefix shape: `FOO=bar gh pr create`, `time gh pr create`, `cd a && cd b && gh pr create` (double cd), `X && gh pr create` (any non-cd prefix) all silently skipped validation. A first attempt widened this by looping four bounded prefix-strip shapes (cd / env-assignment / wrapper / generic quote-free segment) before anchoring — but applying all four shapes *unconditionally* every iteration let the generic shape consume a genuine verb the moment it sat at the head followed by a trailing separator: `gh pr create --fill --head br && echo done`, `gh pr create --title fixbug --head br; echo x`, and `gh pr create --head br | cat` all silently bypassed validation on a real invocation — the same defect class the PR set out to fix, just triggered from the trailing side. Fixed by restructuring the loop to **check the head against the pr-create verb at the top of every iteration, before any stripping happens that iteration** — the moment the head already is the verb, the loop stops and fires immediately, so the verb can never be consumed as if it were a prefix. Only when the head isn't yet the verb does the loop attempt one strip shape and re-check.
- **Five new regression tests** — `test_pr_create_tilde_cd_target.sh` proves a `cd ~/...`-prefixed PR-create resolves the worktree's own branch; `test_lib_pr_repo_tilde_username_hardening.sh` proves a hostile multi-line `~user` value never reaches `eval`; `test_pr_create_gate_anchor.sh` proves a `gh issue create` (even one whose title/body mentions "gh pr create") stays a no-op, every prefix shape (cd / env-assignment / wrapper / double-cd / generic) still validates a genuine PR-create in both directions, and — the case that closes the second security finding — a genuine PR-create followed by an unquoted trailing `&&`/`;`/`|` chain still validates AND still blocks a malformed title, proving the verb is never eaten as a "prefix".

## Testing

1. `bash .claude/hooks/tests/test_pr_create_tilde_cd_target.sh` — 2/2 pass
2. `bash .claude/hooks/tests/test_lib_pr_repo_tilde_username_hardening.sh` — 4/4 pass
3. `bash .claude/hooks/tests/test_pr_create_gate_anchor.sh` — 21/21 pass (15 from the prior revision + 6 new trailing-separator cases proving the check-then-strip fix)
4. Re-ran the existing suites that exercise these two files — all still pass: `test_validate_pr_create_head.sh` (13/13), `test_validate_pr_create_structural_parse.sh` (7/7), `test_validate_pr_create_upstream.sh` (7/7), `test_validate_pr_required_sections.sh` (8/8), `test_pr_hooks_cross_repo.sh` (18/18), `test_validate_branch_name_pushref.sh` (34/34), `test_require_architecture_review.sh` (22/22), `test_require_design_review_for_ui.sh` (22/22)
5. Full `bash bin/run-hook-tests.sh` — 86/86 pass, including `test_require_agdr_for_arch_pr.sh` (its one known-flaky `spike active-ticket marker` case passes in the full-suite run; only fails standalone, reproduces identically on unmodified `upstream/dev`, unrelated to this PR)
6. `shellcheck --severity=warning` clean on both changed hook files
7. `jq empty .claude/settings.json` — valid (no settings.json changes; the fix is fully contained in the hook's own command-detection logic)
8. Regression-proved the trailing-separator fix both ways: the three new malformed-title cases in `test_pr_create_gate_anchor.sh` genuinely FAIL (wrongly return exit 0 instead of blocking) when run against the pre-check-then-strip code, and PASS against the fix — confirming the test cases catch the real bug rather than passing vacuously

Closes #791

## Glossary

| Term | Definition |
|------|------------|
| Gate (in a hook) | The early-exit check a PreToolUse hook runs to decide whether the rest of its validation logic applies to the current command at all |
| Anchored match | A regex match required to start at a specific position (here: the actual command head), as opposed to matching anywhere in the string |
| cd-target | The directory a `cd <path> && …`-prefixed command is about to run in — extracted by hooks because the PreToolUse hook fires *before* the real shell executes the `cd` |
| `pr_cmd_cd_target` | The shared `_lib-pr-repo.sh` function that parses the cd-target path out of a command string; used by five separate PreToolUse hooks |
| Check-then-strip | The loop structure where the verb-match check runs at the top of every iteration, before any prefix-stripping — as opposed to "strip-then-check" (strip everything possible, check once at the end), which is what let the generic strip shape consume a genuine verb |

### Note on `.claude/settings.json`

Confirmed again on this revision: the settings.json matcher for this hook (`"if": "Bash(gh pr create *)"`) is unchanged and was never the source of any of the three findings addressed across this PR's revisions — the over-match and both security-review findings are all internal to the hook's own command-text parsing, not the harness-level matcher. No `.claude/settings.json` changes in this PR.

### DIFF
--- a/.claude/hooks/_lib-pr-repo.sh
+++ b/.claude/hooks/_lib-pr-repo.sh
@@ -74,6 +74,76 @@ _git_remote_to_slug() {
     's|.*[:/]([^/:]+/[^/]+)\.git$|\1|p; s|.*[:/]([^/:]+/[^/]+)$|\1|p' | head -1
 }
 
+# Internal: expand a leading `~` or `~user` in a path to an absolute path.
+#
+# WHY THIS EXISTS (me2resh/apexyard bug report, 2026-07)
+# -------------------------------------------------------
+# `pr_cmd_cd_target` extracts the literal path text from a `cd <path> && …`
+# prefix. When the operator's command reads `cd ~/Projects/foo && gh pr
+# create …`, the extracted path is the literal string `~/Projects/foo` — the
+# shell would expand that tilde when it actually runs the `cd`, but this hook
+# never runs a shell over the extracted string, it hands it straight to
+# `git -C`. `git -C` does NOT perform tilde-expansion (it treats `~` as a
+# literal directory-name character), so `git -C "~/Projects/foo"` fails
+# silently, every caller's cd-target resolution comes back empty, and each
+# hook falls back to its own cwd (typically the ops-fork root on `dev`) —
+# producing misleading errors like "Branch 'dev' missing ticket ID" for a
+# perfectly valid PR from a tilde-path worktree.
+#
+# This helper mirrors the shell's own tilde-expansion rules for the two forms
+# that matter:
+#   `~` / `~/...`       → the current user's $HOME
+#   `~user` / `~user/…` → that user's home directory, resolved via `eval echo`
+#                          ONLY after validating the extracted token looks
+#                          like a real username (alnum/dot/dash/underscore) —
+#                          refuses to eval anything that could smuggle shell
+#                          metacharacters through the path text.
+#
+# Graceful fallback: if expansion cannot be resolved (unknown user, `eval`
+# failure), the original string is returned unchanged rather than throwing —
+# callers already treat an unresolvable `git -C` target as "can't check from
+# here" and degrade non-misleadingly (skip/warn), so returning the original
+# text preserves that existing, safe behaviour rather than fabricating a path.
+_pr_repo_expand_tilde() {
+  local p="$1"
+  # shellcheck disable=SC2088  # intentional: matching a literal leading '~'
+  # in a case pattern, not asking the shell to expand it here.
+  case "$p" in
+    '~')
+      printf '%s' "$HOME"
+      ;;
+    '~/'*)
+      printf '%s' "$HOME/${p#\~/}"
+      ;;
+    '~'*)
+      local rest user tail_part home
+      rest="${p#\~}"
+      user="${rest%%/*}"
+      tail_part=""
+      case "$rest" in
+        */*) tail_part="/${rest#*/}" ;;
+      esac
+      if printf '%s' "$user" | grep -qE '^[A-Za-z0-9._-]+$'; then
+        home=$(eval echo "~${user}" 2>/dev/null)
+        case "$home" in
+          '~'*|'')
+            printf '%s' "$p"  # unresolved (no such user) — return unchanged
+            ;;
+          *)
+            printf '%s%s' "$home" "$tail_part"
+            ;;
+        esac
+      else
+        # Token doesn't look like a safe username — don't eval it.
+        printf '%s' "$p"
+      fi
+      ;;
+    *)
+      printf '%s' "$p"
+      ;;
+  esac
+}
+
 # Public: pr_cmd_target_repo <command>
 #   Echoes the owner/repo slug from the --repo / -R flag in the command.
 #   Handles all four CLI forms:
@@ -135,13 +205,18 @@ git_origin_repo() {
 #   `cd '../my repo' && …` resolves correctly.
 pr_cmd_cd_target() {
   local cmd="$1"
+  local raw
   # Strip leading whitespace, then match `cd ` followed by a quoted or bare
   # path up to the first `&&`, `;`, or `|` separator.
-  printf '%s' "$cmd" | sed -nE \
+  raw=$(printf '%s' "$cmd" | sed -nE \
     "s/^[[:space:]]*cd[[:space:]]+\"([^\"]+)\"[[:space:]]*(&&|;|\|).*/\1/p;
      s/^[[:space:]]*cd[[:space:]]+'([^']+)'[[:space:]]*(&&|;|\|).*/\1/p;
      s/^[[:space:]]*cd[[:space:]]+([^&;|[:space:]]+)[[:space:]]*(&&|;|\|).*/\1/p" \
-    | head -1
+    | head -1)
+  [ -z "$raw" ] && return 0
+  # Expand a leading ~ / ~user before handing the path to any caller — git -C
+  # does not shell-expand tilde itself (see _pr_repo_expand_tilde above).
+  _pr_repo_expand_tilde "$raw"
 }
 
 # Public: pr_repo_matches_cwd <command> [<git-dir>]
--- a/.claude/hooks/_lib-pr-repo.sh
+++ b/.claude/hooks/_lib-pr-repo.sh
@@ -104,6 +104,19 @@ _git_remote_to_slug() {
 # callers already treat an unresolvable `git -C` target as "can't check from
 # here" and degrade non-misleadingly (skip/warn), so returning the original
 # text preserves that existing, safe behaviour rather than fabricating a path.
+#
+# Security note (Hakim / PR #792 review): the username allowlist check below
+# uses bash's `[[ =~ ]]` against the WHOLE string, anchored at both ends —
+# deliberately NOT `grep -qE '^…$'`, which matches per LINE. A multi-line
+# `$user` whose first line happens to look like a valid username would pass
+# a per-line grep check and reach `eval` with the remaining lines still
+# attached (a latent command-injection surface in a trust-chain hook). Not
+# reachable through today's sole caller (`pr_cmd_cd_target`'s `sed` output is
+# always single-line, further guaranteed by `head -1`), but this is a shared
+# lib function — a future caller that skips that sanitization would open the
+# hole `[[ =~ ]]` closes here for free (the whole-string anchor treats a
+# newline as just another character the charset must match, so ANY embedded
+# newline fails the check).
 _pr_repo_expand_tilde() {
   local p="$1"
   # shellcheck disable=SC2088  # intentional: matching a literal leading '~'
@@ -123,7 +136,7 @@ _pr_repo_expand_tilde() {
       case "$rest" in
         */*) tail_part="/${rest#*/}" ;;
       esac
-      if printf '%s' "$user" | grep -qE '^[A-Za-z0-9._-]+$'; then
+      if [[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]; then
         home=$(eval echo "~${user}" 2>/dev/null)
         case "$home" in
           '~'*|'')
--- a/.claude/hooks/tests/test_lib_pr_repo_tilde_username_hardening.sh
+++ b/.claude/hooks/tests/test_lib_pr_repo_tilde_username_hardening.sh
@@ -1,0 +1,147 @@
+#!/bin/bash
+# Regression test — LOW hardening finding from Hakim's security review of
+# PR #792 (apexyard#791 fix): the `~user` username allowlist check in
+# `_pr_repo_expand_tilde` (_lib-pr-repo.sh) used to be a PER-LINE grep,
+# `grep -qE '^[A-Za-z0-9._-]+$'`, which matches per line rather than across
+# the whole string. A multi-line `$user` whose FIRST line looks like a valid
+# username passes that check even though later lines carry arbitrary text —
+# and the validated (but still multi-line) value was then handed straight to
+# `eval echo "~${user}"`.
+#
+# THE BUG (in isolation)
+# -----------------------
+# Not reachable through the current sole caller, `pr_cmd_cd_target`: its sed
+# extraction is single-line by construction and is further piped through
+# `head -1`, so `$user` can never carry an embedded newline when reached via
+# that path. But `_pr_repo_expand_tilde` is a shared, standalone lib
+# function — a future caller that skips that sanitization (or calls the
+# function directly with attacker-influenced input) would reach `eval` with
+# a multi-line, per-line-only-validated `$user`, which is a latent
+# command-injection surface in a trust-chain hook (`.claude/hooks/**`).
+#
+# THE FIX
+# -------
+# Swap the per-line `grep -qE '^…$'` for bash's `[[ "$user" =~ ^…$ ]]`, which
+# anchors against the WHOLE string (`^` / `$` bind to start/end of the full
+# value, not per line) — a multi-line value now fails the check outright
+# regardless of what its first line looks like, and `eval` is never reached.
+#
+# THIS TEST
+# ---------
+# Sources _lib-pr-repo.sh directly and calls the internal
+# `_pr_repo_expand_tilde` function with a hostile multi-line `~user` value
+# whose first line is a valid-looking username and whose second line is a
+# shell command that would create a sentinel file if it ever reached `eval`.
+# Asserts (a) the sentinel file is never created, and (b) the function
+# degrades gracefully — returns the original string unchanged, exactly as it
+# does for any other unresolvable/rejected `~user` form.
+#
+# Exit 0 if all cases pass; exit 1 on first failure.
+
+set -u
+
+SRC_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+LIB_PR_REPO="$SRC_ROOT/.claude/hooks/_lib-pr-repo.sh"
+
+if [ ! -f "$LIB_PR_REPO" ]; then
+  echo "FAIL: lib not found at $LIB_PR_REPO" >&2
+  exit 1
+fi
+
+# shellcheck source=/dev/null
+. "$LIB_PR_REPO"
+
+PASS=0
+FAIL=0
+FAILED_CASES=""
+
+# assert_no_eval_reach label path_value
+#   Calls _pr_repo_expand_tilde with a $1-shaped path whose embedded second
+#   line would touch a sentinel file if `eval` ever executed it. Asserts the
+#   sentinel is never created AND the function returns the input unchanged
+#   (the documented graceful-fallback behaviour for a rejected token).
+assert_no_eval_reach() {
+  local label="$1" path_value="$2"
+  local sentinel got
+  sentinel=$(mktemp -u /tmp/apexyard-test-tilde-hardening.XXXXXX)
+  rm -f "$sentinel"
+
+  # Rebuild path_value with the real sentinel path substituted in for the
+  # __SENTINEL__ placeholder (keeps callers readable).
+  path_value="${path_value//__SENTINEL__/$sentinel}"
+
+  got=$(_pr_repo_expand_tilde "$path_value")
+  local rc=$?
+
+  if [ -e "$sentinel" ]; then
+    echo "FAIL [$label]: sentinel file WAS created — eval executed injected content!" >&2
+    rm -f "$sentinel"
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "
+    return
+  fi
+  rm -f "$sentinel"
+
+  if [ "$rc" -ne 0 ]; then
+    echo "FAIL [$label]: function exited non-zero ($rc)" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "
+    return
+  fi
+
+  if [ "$got" != "$path_value" ]; then
+    echo "FAIL [$label]: want unchanged passthrough, got: $got" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "
+    return
+  fi
+
+  echo "PASS [$label]"
+  PASS=$((PASS+1))
+}
+
+# ---------------------------------------------------------------------------
+# The reported hardening gap: first line of $user looks like a valid
+# username, second line is an injected shell command.
+# ---------------------------------------------------------------------------
+
+MULTILINE_USER=$'validuser\ntouch __SENTINEL__'
+assert_no_eval_reach \
+  "multi-line ~user (valid-looking first line + injected 2nd line) does NOT reach eval" \
+  "~${MULTILINE_USER}"
+
+MULTILINE_USER_WITH_TAIL=$'validuser\ntouch __SENTINEL__'
+assert_no_eval_reach \
+  "multi-line ~user/tail-path variant does NOT reach eval" \
+  "~${MULTILINE_USER_WITH_TAIL}/Projects/foo"
+
+# ---------------------------------------------------------------------------
+# Sanity: legitimate single-line ~user forms still behave as documented
+# (this is a behaviour-preservation check, not a new requirement — confirms
+# the hardening didn't regress the happy path).
+# ---------------------------------------------------------------------------
+
+result=$(_pr_repo_expand_tilde "~root")
+if [ -n "$result" ] && [ "$result" != "~root" ]; then
+  echo "PASS [~root (single-line, real user) still resolves via eval]"
+  PASS=$((PASS+1))
+else
+  echo "FAIL [~root (single-line, real user) still resolves via eval]: got '$result'" >&2
+  FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}root-resolves "
+fi
+
+result=$(_pr_repo_expand_tilde "~zzznosuchuser999/x")
+if [ "$result" = "~zzznosuchuser999/x" ]; then
+  echo "PASS [~nosuchuser (single-line, unknown user) degrades unchanged]"
+  PASS=$((PASS+1))
+else
+  echo "FAIL [~nosuchuser (single-line, unknown user) degrades unchanged]: got '$result'" >&2
+  FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}nosuchuser-passthrough "
+fi
+
+echo ""
+echo "==================================="
+echo "  PASS: $PASS   FAIL: $FAIL"
+echo "==================================="
+if [ "$FAIL" -gt 0 ]; then
+  echo "Failed: $FAILED_CASES" >&2
+  exit 1
+fi
+exit 0
--- a/.claude/hooks/tests/test_pr_create_gate_anchor.sh
+++ b/.claude/hooks/tests/test_pr_create_gate_anchor.sh
@@ -209,6 +209,67 @@ run_case 'a --title value containing a literal && does NOT get misread as a chai
   "gh issue create --repo me2resh/apexyard --title 'build && deploy pipeline' --body 'desc'" \
   0 ""
 
+# ---------------------------------------------------------------------------
+# Hakim's SECOND security re-review of PR #792 found a new BLOCKING
+# regression: the round-1 loop applied shape 4 (a generic quote-free-segment
+# strip) unconditionally every iteration, so a genuine `gh pr create` with
+# quote-free trailing args followed by a top-level `&&` / `;` / `|`
+# separator got its OWN verb consumed as if it were a "chained prefix" —
+# the gate then silently skipped validation (exit 0) on a real invocation.
+# The fix (check-then-strip: verify the head before stripping, every
+# iteration) must make each of these fire and block a malformed title,
+# exactly like the true-positive/true-negative pairs above.
+#
+# NOTE 1: these cases deliberately use UNQUOTED --title values (or omit
+# --title entirely). Shape 4's whole point is stripping a QUOTE-FREE
+# segment — a quoted title (as used in every case above) contains a `'`
+# before any separator, which already keeps shape 4 from matching and would
+# make these cases pass trivially under the OLD buggy code too, defeating
+# the regression test. Hakim's literal reported bypass commands
+# (`--fill --head br && echo done`, `--title fixbug --head br; echo x`,
+# `--head br | cat`) are unquoted/title-free for exactly this reason —
+# faithfully reproducing them (rather than "equivalent" quoted variants) is
+# what makes these cases actually fail against the pre-fix code.
+#
+# NOTE 2: these cases also omit --body/--body-file. The gate-decision
+# preprocessing (near the top of the hook) strips everything from
+# --body-file/--body to end-of-string BEFORE the gate loop ever runs, which
+# would silently remove the trailing " && echo done" / "; echo x" / "| cat"
+# and — like quoting the title — would make the case pass regardless of
+# whether the gate fix works. Omitting --body/--body-file also means the
+# required-##-sections check no-ops (it only runs when the command carries
+# a body flag), so title-format + branch-ticket-ID are what prove the gate
+# fired for these cases specifically.
+#
+# Verified (see PR #792 discussion): each malformed-title case below FAILS
+# (wrongly returns rc=0) against the pre-fix commit and PASSES (rc=2)
+# against the check-then-strip fix.
+# ---------------------------------------------------------------------------
+
+run_case 'Hakim repro: gh pr create --fill --head br && echo done (no title) still validates' \
+  "gh pr create --repo me2resh/apexyard --fill --head fix/GH-900-test && echo done" \
+  0 ""
+
+run_case 'same shape with an unquoted malformed title still BLOCKS (proves the gate fired)' \
+  "gh pr create --repo me2resh/apexyard --fill --title fixbug --head fix/GH-900-test && echo done" \
+  2 "doesn't match format"
+
+run_case 'Hakim repro: gh pr create --title fixbug --head br; echo x — a well-formed unquoted title still validates' \
+  "gh pr create --repo me2resh/apexyard --title fix(#900):x --head fix/GH-900-test; echo x" \
+  0 ""
+
+run_case 'same shape (;) with the literal malformed "fixbug" title still BLOCKS' \
+  "gh pr create --repo me2resh/apexyard --title fixbug --head fix/GH-900-test; echo x" \
+  2 "doesn't match format"
+
+run_case 'Hakim repro: gh pr create --head br | cat (no title) still validates' \
+  "gh pr create --repo me2resh/apexyard --head fix/GH-900-test | cat" \
+  0 ""
+
+run_case 'same shape (|) with an unquoted malformed title still BLOCKS' \
+  "gh pr create --repo me2resh/apexyard --title fixbug --head fix/GH-900-test | cat" \
+  2 "doesn't match format"
+
 rm -f "$BF_OK"
 
 echo ""
--- a/.claude/hooks/tests/test_pr_create_gate_anchor.sh
+++ b/.claude/hooks/tests/test_pr_create_gate_anchor.sh
@@ -160,6 +160,55 @@ run_case 'a leading cd prefix before a genuine gh pr create still validates' \
   "cd /tmp && gh pr create --repo me2resh/apexyard --title 'fix(#900): gate anchor' --head fix/GH-900-test --body-file $BF_OK" \
   0 ""
 
+# ---------------------------------------------------------------------------
+# Hakim's security review of the anchor fix (PR #792) flagged a MEDIUM
+# completeness regression: the single-leading-`cd`-strip anchor missed
+# genuine `gh pr create` invocations behind OTHER prefix shapes that the old
+# (unanchored) gate used to catch. Each of Hakim's four named shapes gets a
+# true-positive case (still validates) below, plus one true-negative (still
+# BLOCKS a malformed title through the same prefix) so the fix isn't merely
+# "widen the no-op path".
+# ---------------------------------------------------------------------------
+
+run_case 'a leading env-var assignment before a genuine gh pr create still validates' \
+  "FOO=bar gh pr create --repo me2resh/apexyard --title 'fix(#900): gate anchor' --head fix/GH-900-test --body-file $BF_OK" \
+  0 ""
+
+run_case 'a leading env-var assignment with a malformed title still BLOCKS' \
+  "FOO=bar gh pr create --repo me2resh/apexyard --title 'no ticket id here' --head fix/GH-900-test --body-file $BF_OK" \
+  2 "doesn't match format"
+
+run_case 'a leading "time" wrapper before a genuine gh pr create still validates' \
+  "time gh pr create --repo me2resh/apexyard --title 'fix(#900): gate anchor' --head fix/GH-900-test --body-file $BF_OK" \
+  0 ""
+
+run_case 'a leading "time" wrapper with a malformed title still BLOCKS' \
+  "time gh pr create --repo me2resh/apexyard --title 'no ticket id here' --head fix/GH-900-test --body-file $BF_OK" \
+  2 "doesn't match format"
+
+run_case 'a double leading cd (cd a && cd b && …) before a genuine gh pr create still validates' \
+  "cd /tmp && cd /tmp && gh pr create --repo me2resh/apexyard --title 'fix(#900): gate anchor' --head fix/GH-900-test --body-file $BF_OK" \
+  0 ""
+
+run_case 'a double leading cd with a malformed title still BLOCKS' \
+  "cd /tmp && cd /tmp && gh pr create --repo me2resh/apexyard --title 'no ticket id here' --head fix/GH-900-test --body-file $BF_OK" \
+  2 "doesn't match format"
+
+run_case 'a generic non-cd prefix (X && gh pr create) still validates' \
+  "echo hi && gh pr create --repo me2resh/apexyard --title 'fix(#900): gate anchor' --head fix/GH-900-test --body-file $BF_OK" \
+  0 ""
+
+run_case 'a generic non-cd prefix (X && gh pr create) with a malformed title still BLOCKS' \
+  "echo hi && gh pr create --repo me2resh/apexyard --title 'no ticket id here' --head fix/GH-900-test --body-file $BF_OK" \
+  2 "doesn't match format"
+
+# The gate widening must not defeat the anchor itself — a --title value that
+# contains a literal "&&" stays inside its quotes and must NOT be misread as
+# a chained prefix ahead of a fabricated "gh pr create" segment.
+run_case 'a --title value containing a literal && does NOT get misread as a chained prefix (exit 0, no-op)' \
+  "gh issue create --repo me2resh/apexyard --title 'build && deploy pipeline' --body 'desc'" \
+  0 ""
+
 rm -f "$BF_OK"
 
 echo ""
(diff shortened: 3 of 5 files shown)


atlas-apex (commented):
## Security Review: PR #792

**Commit**: `281da384b77a9c516a7c886fbe95a03d9a6c8f23`

Trust-chain change (PR-create validation hooks) per #777. Rex owns correctness; I own the security surface. I threat-modelled the three concerns in the task and **fuzzed the `eval` path empirically** against the PR-head lib.

### Verdict: **APPROVED (security)** — no BLOCKING finding

The gating concern (command-injection via the `~user` `eval`) is **not exploitable as shipped**. Two non-blocking recommendations below (one LOW hardening, one MEDIUM completeness).

---

### Concern 1 — `eval echo "~${user}"` command injection — SAFE (was the block/no-block call)

I sourced the PR-head `_lib-pr-repo.sh` and hammered `_pr_repo_expand_tilde` with hostile inputs, watching for sentinel-file creation:

| Input | Result | Sentinel created? |
|-------|--------|-------------------|
| `~$(touch /tmp/PWNED)` | returned unchanged | no |
| `~$(touch /tmp/PWNED2)/foo` | returned unchanged | no |
| `~user;touch /tmp/PWNED3` | returned unchanged | no |
| `~; rm -rf …` | returned unchanged | no |
| `` ~`touch /tmp/x` `` | returned unchanged | no |
| `~user&&touch x` / `~user|touch x` | returned unchanged | no |
| `~{a,b}` / `~$USER` / `~*` | returned unchanged | no |
| `~root`, `~root/Projects/foo` | resolved correctly | — |
| `~zzznosuchuser/x` (valid-shaped, no such user) | returned unchanged (graceful) | — |

**Zero injections.** The allowlist `^[A-Za-z0-9._-]+$` is applied **before** the `eval`, and every shell metacharacter (`$ ( ) ; | & \` { } * space`) fails it, so nothing hostile reaches `eval`. The end-to-end path is doubly safe: `pr_cmd_cd_target`'s `sed` requires `cd <path> <sep>` on a single line and `head -1` guarantees single-line output, so `$user` is always single-line and the allowlist is fully effective. Confirmed empirically that a newline-bearing command yields an **empty** cd-target end-to-end.

**LOW — hardening / defense-in-depth (not reachable today):** the username check is `grep -qE '^…$'`, which matches **per line**. In isolation, a multi-line `$user` whose *first* line is a valid username passes the grep and reaches `eval` — I reproduced an injection in that isolated scenario. It is **not** reachable through the current sole caller (single-line guarantee above), but `_pr_repo_expand_tilde` is a shared lib function; a future caller that skips the `head -1` sanitization would open a command-injection in a trust-chain hook. Cheap fix, removes the footgun:
- Swap `grep -qE '^…$'` for bash `[[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]` (whole-string, immune to the newline trick — verified), **or**
- Drop `eval` entirely: `getent passwd "$user" | cut -d: -f6` with a bash fallback. `~user` is a nice-to-have; the common case is `~/…`, which the code already handles without `eval`.

### Concern 2 — anchor fix weakening the gate — fixes the reported bug, but a MEDIUM completeness regression

The reported over-match is **correctly fixed**: `gh issue create --title '…gh pr create…'` now SKIPS this validator (confirmed).

But the head-anchor + single-`cd`-strip introduces a new **false-negative** class — genuine PR-create invocations behind a non-`cd` prefix now **skip validation entirely**, where the old unanchored gate validated them:

| Command shape | OLD gate | NEW gate |
|---|---|---|
| `gh pr create …` | FIRE | FIRE |
| `cd x && gh pr create …` | FIRE | FIRE |
| `FOO=bar gh pr create …` | FIRE | **SKIP** |
| `time gh pr create …` / `command gh pr create …` | FIRE | **SKIP** |
| `cd a && cd b && gh pr create …` (double cd) | FIRE | **SKIP** |
| `X && gh pr create …` (any non-cd prefix) | FIRE | **SKIP** |

Severity **MEDIUM as a completeness regression**, not a vulnerability — this validator is a policy/quality backstop (title format, required sections, branch-ticket-ID), not an adversarial security boundary, so a skip yields a non-conforming PR rather than code execution or privilege. It does **not** block the security sign-off, but the author/Rex should weigh it: an env-var-prefixed or `time`-wrapped `gh pr create` is a natural benign shape that silently loses validation. Recommended: additionally strip a leading env-assignment (`^\s*\w+=\S+\s+`) and common wrappers (`time`/`command`/`nice`/`env`), and loop the `cd`-strip. Note the real tension — a naive "match the verb at any segment head" reintroduces the over-match on quoted metacharacters in `--title`, so the conservative anchor is a defensible fail-open choice; covering at least the env-assignment case is the highest-value delta.

(Caveat: whether these shapes reach the hook depends on the `Bash(gh pr create *)` matcher in settings.json. The very existence of the reported bug — a `gh issue create` reaching this validator — proves the matcher is *not* a strict head-glob, so these prefixed genuine invocations very likely do reach the hook and thus do regress.)

### Concern 3 — cd-strip hiding the real verb — no new bypass

The strip removes only a bounded leading `cd <path> <sep>` (path regexes stop at whitespace / quote / separator) and then anchors on `gh pr create` at the head — it cannot morph a non-PR-create command into a matching one, and cannot hide the verb (it *requires* the verb at head after stripping). The tilde-expanded cd-target feeding `git -C` extends pre-existing behavior to tilde paths; no new path-traversal or exfiltration beyond what the hook already did.

### Checklist
- Secrets & Credentials: Pass
- Injection Prevention: Pass (eval allowlist effective end-to-end; one LOW latent-multiline hardening note)
- XSS: N/A
- Auth & Authorisation: N/A
- Data Protection: Pass
- API Security: N/A

---
🛡️ Reviewed by Hakim (Security Auditor) — advisory, no merge marker
📌 Reviewed commit: `281da384b77a9c516a7c886fbe95a03d9a6c8f23`

atlas-apex (commented):
## Code Review: PR #792

**Commit**: `281da384b77a9c516a7c886fbe95a03d9a6c8f23`

### Summary
Fixes two bugs in the PR-creation validation path (`_lib-pr-repo.sh` + `validate-pr-create.sh`), both root-caused to string handling in PreToolUse hooks that run *before* a real shell touches the command:

1. **Tilde not expanded before `git -C`** — `pr_cmd_cd_target` handed `git -C` a literal `~/…` string; `git -C` treats `~` as a literal directory char, so cd-target resolution came back empty and callers fell back to the hook's own cwd (`dev`) → false "Branch 'dev' missing ticket ID". Fixed once, centrally, via a new `_pr_repo_expand_tilde` helper so all 5 callers benefit.
2. **Gate over-matched the create verb as an unanchored substring** — the phrase inside a `--title` value (e.g. a bug report *about* this hook) triggered full PR-only validation on an issue-create command. Fixed by stripping one optional leading `cd <path> &&|;|` prefix and anchoring the match to the command head. (Live confirmation: my own review commands contain the phrase and the old installed hook over-matched them twice during this review.)

### Checklist Results
- Architecture & Design: Pass — fix placed in the shared lib, single fix-point for all 5 callers; no layering issues
- Code Quality: Pass — clear naming, thorough WHY comments, graceful fallbacks
- Testing: Pass — 2 new suites (2/2, 6/6) assert *both* directions of each fix; all 8 pre-existing related suites green
- Security: Pass (code-review lens) — see note to Hakim below
- Performance: Pass — pure string ops in a hook, negligible
- PR Description & Glossary: Pass — Summary is narrative (what + why), Testing steps present, 4-term Glossary, `Closes #791`
- Technical Decisions (AgDR): N/A — bug fix, no new decision/dependency/pattern
- Adopter Handbooks: N/A — no handbooks loaded for this diff

### Independent Verification
- **Diff scope** exactly matches the claim: `_lib-pr-repo.sh` (+77/-2), `validate-pr-create.sh` (+27/-3), 2 new test files. No `.claude/settings.json` change — confirmed correct: the matcher glob is scoped to the create subcommand only (not issue-create / api-issues), so the over-match came purely from the hook's internal unanchored grep, not the matcher. No settings change was needed.
- **Tilde fix correctness** — exercised the helper directly: absolute (`/abs/foo`), relative (`../rel`), and bare (`plain`) paths pass through unchanged; `~`, `~/…`, `~user/…` expand correctly; unknown user (`~nosuchuser`) and the bash special form `~-` degrade safely to the original string. `_pr_repo_expand_tilde` is applied once inside `pr_cmd_cd_target`, so all 5 callers (validate-pr-create, validate-branch-name, require-agdr-for-arch-pr, require-architecture-review, require-design-review-for-ui) inherit the fix.
- **Fail-closed preserved (the critical check)** — traced and test-confirmed: a genuine create with a malformed title still BLOCKS (rc=2, "doesn't match format"); a `cd /tmp && …` prefix still validates; a tilde-cd on a non-conforming branch still BLOCKS ("missing ticket ID"). The anchor only suppresses *false* positives (phrase inside `--title`, non-create verbs); it does not weaken real validation. Commands that reach this hook necessarily start with the create verb or `cd … && <verb>` (per the matcher), both caught by the head anchor after the cd-strip.
- **Tests** — new: tilde 2/2, gate-anchor 6/6. Pre-existing: head 13/13, structural_parse 7/7, upstream 7/7, required_sections 8/8, cross_repo 18/18, branch_name_pushref 34/34, require_architecture_review 22/22, require_design_review_for_ui 22/22. The one `test_require_agdr_for_arch_pr.sh` failure (`spike active-ticket marker …`) reproduces identically on unmodified `upstream/dev` — confirmed pre-existing and unrelated.
- **shellcheck** — `_lib-pr-repo.sh` fully clean; `validate-pr-create.sh` shows only pre-existing INFO findings on lines 431/535/600 (outside this PR's hunks); the new code introduces none. The one intentional SC2088 in the new helper is suppressed with a justified inline comment.

### Issues Found
None blocking.

### Note for the Security Auditor (Hakim) — trust-chain change
This PR touches `.claude/hooks/**`, so a security pass follows. Flagging one line for your lens: `_pr_repo_expand_tilde` resolves `~user` via `home=$(eval echo "~${user}" 2>/dev/null)`. The `${user}` token is validated against `^[A-Za-z0-9._-]+$` **before** the eval, which excludes every shell metacharacter (`$`, backtick, `;`, `|`, `&`, `()`, `<>`, whitespace, quotes) — so command injection through a crafted cd-target is not reachable via this path. Two benign observations: (a) the charset admits the bash tilde-prefix special forms `~-` / `~N` (dirstack), which expand to `$OLDPWD` / dir-stack rather than a home dir — in practice `~-` degraded to unchanged on the test shell and neither is an injection vector; (b) the eval input is derived from an operator's own command string, not an external actor. No code-review blocker; surfaced for completeness.

### Suggestions
- suggestion (non-blocking): the `~user` branch could avoid `eval` entirely with `getent passwd "$user"` / `dscl` on macOS, but that trades portability for purity and the charset guard already closes the injection surface. Fine as-is.

### Verdict
**APPROVED**

---
Reviewed commit: `281da384b77a9c516a7c886fbe95a03d9a6c8f23`

atlas-apex:
## Follow-up: addressed both review findings

New commit `2ce6708` (was `281da38`) fixes the two non-blocking findings from Hakim's security review and pairs each with regression tests.

### Fix 1 (LOW — eval hardening, `_lib-pr-repo.sh`)

`_pr_repo_expand_tilde`'s `~user` allowlist check was `grep -qE '^[A-Za-z0-9._-]+$'`, which matches **per line** — a multi-line `$user` whose first line looks like a valid username would pass and reach `eval` with the rest still attached. Not reachable through the sole current caller (`pr_cmd_cd_target`'s output is always single-line, `head -1` guaranteed), but a latent hole for a future caller.

Fixed by swapping the per-line grep for bash's whole-string `[[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]` — anchored against the entire value, so an embedded newline fails the check outright regardless of what the first line looks like. (Went with `[[ =~ ]]` over `getent passwd`, since `getent` isn't shipped on macOS by default and this hook needs to keep working there.)

New test: `test_lib_pr_repo_tilde_username_hardening.sh` — sources the lib directly, calls `_pr_repo_expand_tilde` with a hostile multi-line `~user` value carrying an injected `touch <sentinel>` on its second line, and asserts the sentinel is never created. Confirmed this reproduces against the *old* grep check (manually verified the old check matches the same crafted input) and is closed by the new one.

### Fix 2 (MEDIUM — gate completeness regression, `validate-pr-create.sh`)

The anchor fix stripped only ONE leading `cd <path> &&` prefix before checking the command head, which correctly fixed the reported over-match but introduced a false-negative class Hakim's table called out: `FOO=bar gh pr create`, `time gh pr create`, `cd a && cd b && gh pr create` (double cd), and `X && gh pr create` (any non-cd prefix) all now silently **skip** validation instead of firing.

Fixed by looping the prefix-strip over four bounded shapes until a fixed point (10-iteration ceiling): (1) `cd <path> &&/;/|`, (2) leading env-var assignment(s), (3) common wrappers (`time`/`command`/`nice`/`env`), (4) a generic quote-free segment before a top-level separator. Shape 4 is the one that needed care — it excludes quote characters from the matched segment, so a `--title '... && ...'` value (which always has a quote before any real separator) can never be misread as a chained prefix. This preserves the original anchor fix's guarantee while closing the false-negative gap.

New tests (8 added to `test_pr_create_gate_anchor.sh`): a true-positive + a malformed-title true-negative for each of the four prefix shapes, plus one more confirming a `--title` value containing a literal `&&` still doesn't trigger. The true-negative pairing matters — a bare "still validates" case with a conforming title is rc=0 either way (validated-and-passed vs. gate-skipped are indistinguishable), so each shape's malformed-title companion is what actually proves the gate fired. Verified all 4 new true-negative cases fail against the pre-fix code and pass against the fix.

### Verify

- Both new test files pass (4/4, 15/15 including the 6 pre-existing cases in that file)
- Full pre-existing suites for this area still green: `test_pr_create_tilde_cd_target.sh`, `test_validate_pr_create_{head,structural_parse,upstream}.sh`, `test_validate_pr_required_sections.sh`, `test_pr_hooks_cross_repo.sh`, `test_validate_branch_name_pushref.sh`, `test_require_architecture_review.sh`, `test_require_design_review_for_ui.sh`
- Full `bin/run-hook-tests.sh`: 86/86 pass in the suite run. Standalone, `test_require_agdr_for_arch_pr.sh` shows its one pre-existing `spike active-ticket marker` failure — confirmed identical on an unmodified `dev` checkout, unrelated to this PR (flaky/order-dependent; the full-suite run above showed it passing)
- `shellcheck --severity=warning` clean on both changed hook files and all touched/new test files
- `jq empty .claude/settings.json` — valid (no settings changes, as before)

Not merging/self-reviewing — flagging for re-review.

atlas-apex (commented):
## Security Re-Review: PR #792

**Commit reviewed**: `2ce6708dae308a2c8c692d80c36e2a5887049254`
**Base**: `dev` · **Scope**: verify the two fixes from my prior review actually close what I flagged, and confirm no new hole.

I sourced the PR-head `_lib-pr-repo.sh` and ran the injection battery directly against `_pr_repo_expand_tilde`; I replicated the gate loop faithfully AND ran the real PR-head `validate-pr-create.sh` in a git sandbox for the bypass cases. The PR's three new test files all pass (4 / 15 / 2).

---

### Fix 1 — `~user` username check (`[[ =~ ]]` whole-string) — CLOSED

The move from per-line `grep -qE '^…$'` to `[[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]` closes the latent injection. I fired an 18-case battery with a sentinel (`touch /tmp/PWNED`) watched on every call. **The sentinel was never created.**

| Attempt | Result |
|---|---|
| multi-line `$'root\n$(touch …)'` (the exact case I flagged) | rejected -> passthrough, never reaches `eval` |
| multi-line `$'root\ntouch …'` + tail-path variant | rejected -> passthrough |
| `~$(...)`, backticks, `~x;touch`, `~x&&touch`, `~x|touch` | rejected -> passthrough |
| `~-`, `~{a,b}`, `~a b`, `~*`, unicode `~rooté` | rejected -> passthrough (or literal, no exec) |
| controls: `~root`, `~/path`, `~nosuchuser` | resolve / degrade correctly |

Bash `[[ =~ ]]` anchors `^`/`$` to the whole string (no `REG_NEWLINE`), so any embedded newline fails the charset match and `eval echo "~$user"` is never reached with multi-line input. The charset also excludes every shell metacharacter, so a token that *does* pass can't smuggle anything into `eval`. Solid.

---

### Fix 2 — looping prefix-strip — closes the four I flagged, but introduces a NEW completeness regression

**Good news first:** the four prefix shapes I flagged last round now all FIRE correctly (verified against both the harness and the real hook): `FOO=bar gh pr create`, `time gh pr create`, `cd a && cd b && gh pr create`, `X && gh pr create`. The `--title`-with-literal-`&&` false-positive stays fixed (quote-exclusion is **not** abusable to fake a chained prefix — confirmed). No `eval`/`exec` anywhere in the loop; it's pure `printf`/`sed` text transform, so **the "injection via the loop" check is clean**.

**The new hole — shape 4 strips the real verb.** Shape 4 (`^[^"'&;|]+(&&|;|\|)…`) strips a quote-free leading segment ending in a top-level separator. When a **genuine PR-create** has quote-free args before the first `&&` / `;` / `|`, the loop consumes the verb *itself* as a "prefix", the head-anchored check fails, and **the gate SKIPs validation entirely (exit 0, silent).** Confirmed against the real PR-head hook:

| Genuine command | Pre-#792 (unanchored) | This revision |
|---|---|---|
| `gh pr create --fill --head br && echo done` | FIRE (validated) | **SKIP — validation bypassed** |
| `gh pr create --title fixbug --head br; echo x` | FIRE | **SKIP** |
| `gh pr create --title fixbug --head br \| cat` | FIRE | **SKIP** |
| `gh pr create; echo x` / `gh pr create && …` | FIRE | **SKIP** |
| `gh pr create --title 'fix: x' …` (quoted) | FIRE | FIRE (ok) |

This is the **same class of defect the PR set out to fix** (a genuine PR-create silently skipping validation), just triggered from the trailing side instead of the leading side — and it's reachable through *common* shapes: `--fill && …`, `… | cat` (pager avoidance), `…; …` chaining. A quoted title/body protects (the quote stops shape 4), but `--fill`-based creates and any unquoted-arg create do not. Because it's a silent exit-0, the bypass is invisible — worse than the false-*positive* this PR removes. The PR's stated goal is gate completeness ("fire on every genuine PR-create"); this regresses it.

**Suggested fix (small):** check the head anchor at the **top** of each loop iteration (check-then-strip), and `break`/FIRE the moment the current head matches `^gh…pr…create` — so a legitimate verb is detected *before* shape 4 can consume it. I traced this against the whole battery: it keeps every true-positive firing, keeps the four prefix shapes firing, and preserves the `gh issue create --title '…gh pr create…'` false-positive fix (quoted, never stripped, head stays `gh issue create` -> SKIP).

---

### Checklist
- Secrets & Credentials: Pass
- Injection Prevention: **Pass** — Fix 1 closes the `eval` path; the gate loop does no `eval`/`exec`
- Auth & Authorisation: n/a
- Trust-chain guardrail integrity: **Fail** — Fix 2 shape 4 lets a genuine PR-create (e.g. `--fill && …`) silently skip validation

### Findings
- **MEDIUM (new, blocking)** — shape 4 quote-free-segment strip consumes the real PR-create verb when its leading args are quote-free and followed by `&&`/`;`/`|`; the gate then skips validation silently. Trust-chain guardrail completeness regression.

### Verdict
**BLOCKING** — Fix 1 is fully closed and clean. Fix 2 closes the four prefix cases I flagged but opens a new, realistically-reachable false-negative that silently bypasses PR-create validation. One targeted change (check-then-strip) closes it; happy to re-verify.

---
Reviewed commit: `2ce6708dae308a2c8c692d80c36e2a5887049254`

atlas-apex (commented):
## Code Review (re-review): PR #792 — DELTA `2ce6708`

**Commit**: `2ce6708dae308a2c8c692d80c36e2a5887049254` (prior Rex approval at `281da38` is stale — revision addressed two review findings)

Re-reviewed the revision commit `2ce6708` (delta vs `281da38`): eval hardening in `_lib-pr-repo.sh` + gate-completeness restoration in `validate-pr-create.sh`. Focused on the delta.

### Fix 1 — eval hardening (`_lib-pr-repo.sh`) ✅
`grep -qE '^[A-Za-z0-9._-]+$'` → `[[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]`.
- **Whole-string, anchored, multi-line-immune.** bash `[[ =~ ]]` binds `^`/`$` to the start/end of the *entire* value (no `REG_NEWLINE`), and the charset excludes newline, so any embedded newline fails the check outright. The old `grep` matched *per line* — a `validuser` + newline + `touch …` value passed on its first line and carried the injected tail into `eval echo "~${user}"`. Confirmed closed by `test_lib_pr_repo_tilde_username_hardening.sh`, which fires the real function with a sentinel-file payload and asserts the sentinel is never created (4/4 pass).
- Charset unchanged; a valid single-line `~root` still resolves; `[[ =~ ]]` is bash 3.2+ (macOS-safe) — correct call vs `getent`.

### Fix 2 — gate completeness restored WITHOUT reintroducing the over-match ✅
The single-`cd`-strip anchor became a bounded loop (≤10 iters) stripping four prefix shapes to a fixed point, then anchoring. Verified both directions against the real hook (git sandbox + mock gh) and by isolating the strip loop on adversarial inputs:

- **Completeness restored** — a genuine PR-create behind each shape now validates AND blocks a malformed title (the true-negative, rc=2, is what proves the gate actually fired):
  - `FOO=bar` env-assignment prefix → validates / blocks ✓
  - `time` / `nice time` wrapper → validates ✓
  - double `cd a && cd b && …` → validates / blocks ✓
  - generic non-cd `echo hi && …` → validates / blocks ✓
- **Original false positive stays fixed** — an issue-create invocation whose `--title` carries `x && <pr-create-phrase> y` stays a **no-op**. The quote-exclusion (`[^"'&;|]+`) in shape 4 stops the segment at the `'` *before* the in-title `&&`, so a quoted flag value can never be misread as a chained prefix. Also confirmed a `gh pr list …` carrying the phrase in an arg → no-op (head is `list`, not the create verb).
- **Quote-exclusion soundness** — sound. A quoted title always presents a quote before any genuine separator, so shape 4 is ineligible on it; shape 4 also self-terminates at the first real `&&`/`;`/`|` because those chars are excluded from its own class. An *unquoted* top-level `&&`-chained real invocation resolves to the real create verb — that is a genuine chained invocation, so triggering is right, not a false positive.

### Loop ceiling ✅
Monotonic (sed only removes) with a fixed-point break and a hard 10-iter cap → always terminates. No nested quantifiers → no catastrophic backtracking. Measured: a 20-chained-`cd` pathological input completes in ~0.25s and still triggers (≈2 prefixes stripped per iter → 10 iters covers 20); beyond that it fail-*opens* (skips validation) rather than false-blocking or hanging — the safe direction.

### Tests ✅ — ran the suites
- New: `test_pr_create_gate_anchor.sh` **15/15**, `test_lib_pr_repo_tilde_username_hardening.sh` **4/4**, `test_pr_create_tilde_cd_target.sh` **2/2**. The per-shape malformed-title true-negatives are present and assert `rc=2` + `doesn't match format` — they genuinely prove the gate fired, not just that the no-op path widened.
- Regression: `test_validate_pr_create_head` 13/13 · `…_structural_parse` 7/7 · `…_upstream` 7/7 · `…_required_sections` 8/8 · `test_pr_hooks_cross_repo` 18/18 · `test_validate_branch_name_pushref` 34/34 · `test_require_architecture_review` 22/22 · `test_require_design_review_for_ui` 22/22. All green.

### shellcheck
`_lib-pr-repo.sh` **clean**. `validate-pr-create.sh` — the **delta region (lines 91–163) is clean**; the 3 remaining `info`-level findings (SC1091 @482, SC2086 @586, SC2059 @651) are all **outside the delta and pre-existing**, not introduced here.

### Other gates
- No `.claude/settings.json` change (delta = 2 hook files + 3 tests). ✓
- No new dependency / framework / pattern → **no AgDR required** (bug fix to existing hooks). Glossary present; Summary bullets narrative. ✓
- No JS/TS in diff → Fallow **N/A**. No migration files / no domain-app-infra layering → always-load handbooks **N/A**.

- **PR body is stale vs the revision.** Summary bullet 2 still describes *"strips one optional leading `cd <path> &&` prefix"* and the Testing section lists the gate-anchor suite as `6/6` — both describe the pre-revision single-cd anchor. The final code loops over four prefix shapes (env-assign / wrapper / multi-cd / generic) and that suite is now 15/15, plus a new username-hardening suite that isn't listed. Worth refreshing the body so a reader sees what the merged code actually does.
- **Narrow theoretical fail-open (nit):** an *unquoted* leading title chained ahead of a second command — a create verb with `--title foo` (no quotes) followed by `&& …` — gets its quote-free leading segment stripped by shape 4 and skips validation. Not reachable by any realistic PR title (real titles carry spaces/`:`/`()` and are always quoted), and it fails *open* (skip), not toward a false block. Noting for completeness only.

### Verdict
**APPROVED**

Trust-chain change (`.claude/hooks/**`) — Security Auditor (Hakim) re-review follows per the #777 trigger.

---
📌 Reviewed commit: `2ce6708dae308a2c8c692d80c36e2a5887049254`

atlas-apex:
## Follow-up: check-then-strip fixes the trailing-separator bypass (Hakim's second review)

New commit `80c62a1` (was `2ce6708`) addresses the BLOCKING finding from Hakim's second security re-review.

### The bug

The round-1 gate loop applied all four prefix-strip shapes unconditionally every iteration and only checked the resulting head against the `gh pr create` verb once the loop finished. Shape 4 (the generic quote-free-segment strip) doesn't know the difference between "someone's chained prefix" and "the verb itself with an unquoted trailing chain" — so a genuine, unquoted invocation like `gh pr create --fill --head br && echo done` got its own verb stripped as if it were a prefix, and the gate silently exited 0 (skipped validation) on a real PR-create.

### The fix — check-then-strip

Restructured the loop so the verb check runs at the **top of every iteration, before any stripping happens that pass**. The moment the current head already matches `^gh pr create`, the loop stops and fires immediately — so a genuine verb sitting at the head (with or without a trailing separator) is caught before shape 4 (or any shape) ever gets a chance at it. Only when the head is *not* yet the verb does the loop attempt one strip shape (cd / env-assignment / wrapper / generic, tried in that priority order) and loop back to re-check.

This preserves every prior guarantee:
- The four prefix shapes from round 1 (`FOO=bar`, `time`, double `cd`, generic `X &&`) still fire — they just take one more iteration each, since only one shape strips per pass now instead of all four unconditionally, well within the 10-iteration ceiling.
- The `--title '...gh pr create...'` false-positive fix stays fixed — a quoted title still stops shape 4 cold.
- Fix 1 (the `~user` `[[ =~ ]]` hardening) is untouched, as instructed.

### Tests

Added 6 cases to `test_pr_create_gate_anchor.sh` reproducing Hakim's exact literal bypass commands (unquoted title / no title, since a quoted title would've passed trivially even under the buggy code and proven nothing):

- `gh pr create --fill --head br && echo done` — validates
- same shape + unquoted malformed title — **BLOCKS** (proves the gate fired)
- `gh pr create --title fixbug --head br; echo x` — validates when well-formed / **BLOCKS** when malformed
- `gh pr create --head br | cat` — validates when well-formed / **BLOCKS** when malformed

Verified these three malformed-title cases genuinely **fail** (wrongly return exit 0) against the pre-this-commit code and **pass** (exit 2, blocked) against the fix — so they catch the real regression rather than passing vacuously.

### Full verification

- `test_pr_create_gate_anchor.sh`: 21/21 (15 prior + 6 new)
- `test_lib_pr_repo_tilde_username_hardening.sh`: 4/4 (unaffected, Fix 1 untouched)
- `test_pr_create_tilde_cd_target.sh`: 2/2
- All pre-existing related suites still green (head/structural_parse/upstream/required_sections/cross_repo/branch_name_pushref/require_architecture_review/require_design_review_for_ui)
- Full `bin/run-hook-tests.sh`: 86/86 (the one `test_require_agdr_for_arch_pr.sh` case that's flaky standalone passes in the full-suite run, same as noted in the prior review round)
- `shellcheck --severity=warning`: clean on both changed files
- `jq empty .claude/settings.json`: valid, no settings changes

PR body updated to reflect the full history across both revisions.

Not merging/reviewing — flagging for re-review.

atlas-apex (commented):
## Security Re-Review: PR #792 (round 3)

**Commit reviewed**: `80c62a159cf4f9b01a585d466441301cc4b29a4c`
**Prior BLOCKING finding** (at `2ce6708`): the round-1 loop applied the generic quote-free strip (shape 4) *unconditionally* every iteration, then checked the head only once at the end — so a genuine `gh pr create` followed by a top-level `&&`/`;`/`|` got its own verb eaten as a "prefix" and the gate silently skipped validation (exit-0) on a real invocation.

**Verdict: APPROVED (security).** The over-strip is closed, no new bypass was introduced, and the loop terminates. Verified by driving the verbatim gate loop as a FIRE/SKIP oracle (38-case battery) and running the real hook end-to-end via the PR's bundled tests.

### 1. Over-strip CLOSED ✅
The restructure moves the verb check to the **top of every iteration, before any strip**. The moment the head already is `gh pr create`, the loop breaks and fires — the verb can never reach shape 4. My exact bypass battery, each genuine invocation → **FIRE**:

| Input (verb = `gh pr create`) | Gate | End-to-end |
|---|---|---|
| `VERB --fill --head br && echo done` | FIRE (iter 0) | malformed title → **BLOCK rc=2** |
| `VERB --title fixbug --head br; echo x` | FIRE (iter 0) | **BLOCK rc=2** |
| `VERB --head br \| cat` | FIRE (iter 0) | **BLOCK rc=2** |
| `VERB --repo o/r --fill --head br && echo done && echo more` | FIRE | — |
| `VERB --head br ; echo a ; echo b` | FIRE | — |
| `VERB --head br \| cat \| wc -l` | FIRE | — |
| `VERB --head br && cd /tmp && echo mixed` | FIRE | — |
| `VERB --head br \| tee log ; echo z` | FIRE | — |
| `VERB --head br &` | FIRE | — |

The three literal repros are also codified as regression tests in `test_pr_create_gate_anchor.sh` (21/21 pass) — each proven to FAIL against the pre-fix code and PASS after.

### 2. No NEW bypass from check-then-strip ✅
- **False-positives stay no-op (SKIP)** — not re-broken: `gh issue create --title '…gh pr create…'`, `--title '[Bug] over-matches gh pr create'`, `--title 'build && deploy'` (literal `&&` inside a quoted value), `echo 'gh pr create'`. All **SKIP**. The quote-free constraint (`[^"'&;|]+`) on shape 4 still keeps a quoted flag value from being misread as a chained prefix.
- **Crafted-lookalike probes** — none falsely FIRE, none false-pass a malformed pr-create: `ghpr create`, `gh-pr-create`, `gh prcreate`, `gh pr created` → SKIP; `MSG=ghprcreate gh issue create …` → SKIP.
- **Leading-prefix coverage intact**: `cd`, double-`cd`, `FOO=bar`, `FOO=bar BAZ=qux`, `time`, `env`, `echo hi &&`, and mixed `cd && FOO=bar time` all peel to the verb and **FIRE**.

### 3. Termination ✅
Every strip shape strictly shortens the head (removes a prefix + its separator) → monotonic decrease → guaranteed termination; the no-progress guard (`_cmd_head = _cmd_head_prev → break`) and the 10-iteration ceiling are backstops. Confirmed at the boundary: 9 chained prefixes → FIRE (iter 9); 10+ → SKIP (iter 10, no hang). Cannot be driven non-terminating or pathological.

### 4. Fix 1 (`~user` `[[ =~ ]]`) re-confirmed ✅
Independent direct probe on the PR-head lib: `~validuser\ntouch SENTINEL` and `~$(touch SENTINEL)` are both rejected by the whole-string `[[ "$user" =~ ^[A-Za-z0-9._-]+$ ]]` — `eval` never reached, sentinel never created, input returned unchanged. Bundled `test_lib_pr_repo_tilde_username_hardening.sh` 4/4.

### Non-blocking note (LOW / informational — pre-existing, not this round)
Three prefix shapes cause the gate to **fail OPEN** (skip a genuine pr-create rather than validate it): a subshell-wrapped `( gh pr create … )`, a quoted env value carrying a space (`FOO='a b' gh pr create …`), and ≥10 chained prefixes. These were present in round-1 too, are contrived, and fail in the safe direction — worst case a format-lint gate doesn't run (a cosmetically-off PR title slips through), never a false-*pass* of a security control. Fine to defer; noting for completeness.

### Evidence
- 38-case FIRE/SKIP oracle battery (verbatim loop): **38/38 as expected**
- `test_pr_create_gate_anchor.sh` **21/21**, `test_lib_pr_repo_tilde_username_hardening.sh` **4/4**, `test_pr_create_tilde_cd_target.sh` **2/2** (real hook, end-to-end)
- `shellcheck --severity=warning` clean on `validate-pr-create.sh` and `_lib-pr-repo.sh`

No `.claude/settings.json` change; the fix is contained in the hook's own command-text parsing. Trust-chain change reviewed under #777.

---
📌 Reviewed commit: `80c62a159cf4f9b01a585d466441301cc4b29a4c`

atlas-apex (commented):
## Code Review: PR #792 (round-3 re-review)

**Commit**: `80c62a159cf4f9b01a585d466441301cc4b29a4c`

### Summary
Round-3 fix for the `validate-pr-create.sh` gate. Round-2 (`2ce6708`) anchored the verb match but ran all four prefix-strip shapes unconditionally per iteration then checked the head only once at the end — letting the generic quote-free shape (#4) eat a genuine verb the moment it sat at the head followed by an unquoted trailing `&&`/`;`/`|`. This round restructures the loop to **check-then-strip**: the verb match runs at the TOP of every iteration, before any strip, so the moment the head IS the verb the loop fires immediately and the verb can never be consumed as a prefix. Fix 1 (the `~user` `[[ =~ ]]` hardening in `_lib-pr-repo.sh`) is untouched from round 2.

### Delta verification (`80c62a1`)

**1. Check-then-strip is correct.** Traced + empirically confirmed. The three genuine trailing-chain invocations fire at the iter-0 head check before shape #4 is ever reached, so validation runs against the original command and a malformed title is blocked (rc=2). I ran the new `test_pr_create_gate_anchor.sh` against the **pre-fix** hook (`2ce6708`): exactly the three malformed-title true-negatives return rc=0 (gate silently skipped) — the round-2 defect — and all three flip to rc=2 (fired + blocked) on `80c62a1`. The tests genuinely catch the bug; they are not vacuous. Author's claim confirmed.

**2. Both prior wins preserved.** The four leading-prefix shapes (`FOO=bar` env-assign, `time` wrapper, double-cd, generic `X && <verb>`) still FIRE, and the mention-in-title false-positive on an issue-create stays a no-op (shape #4's quote-free character class can't cross the opening quote of a flag value). 21/21 in `test_pr_create_gate_anchor.sh`; 4/4 in `test_lib_pr_repo_tilde_username_hardening.sh`.

**3. No infinite loop / catastrophic input.** Three independent termination guarantees: (a) verb-match `break`, (b) strip-nothing `break` when `_cmd_head == _cmd_head_prev`, (c) the 10-iteration cap that fails OPEN (skips validation) rather than hanging. Every strip shape removes a leading prefix, so `_cmd_head` shrinks monotonically until a shape stops matching — no oscillation possible. Multi-prefix inputs resolve within 2-3 iterations, comfortably under the cap.

**4. Tests + hygiene.** 21/21 gate-anchor + 4/4 tilde-hardening + 2/2 tilde-cd-target. Full `bin/run-hook-tests.sh` = **86/86 PASS**. `shellcheck --severity=warning` clean on both changed hook files. `jq empty .claude/settings.json` valid (and unchanged per the diff — the fix is fully contained in the hook's own command-text parsing). CI green on all checks. PR body refreshed with a narrative summary + Glossary (incl. the new "check-then-strip" term).

### Checklist Results
- Code Quality:              Pass (well-commented; records the round-1 to round-2 rationale inline)
- Testing:                   Pass (non-vacuous regression proof, both directions)
- Security:                  Pass from a code-quality lens — trust-chain hook; Hakim re-review follows for the security angle
- Performance:               Pass (bounded loop, 10-iter cap)
- Technical Decisions (AgDR): N/A (bug fix, no new decision)

### Issues Found
None.

### Suggestions
None blocking. The inline comment block is long but earns its length — it is the audit trail for a three-round trust-chain fix; keep it.

### Verdict
**APPROVED**

Trust-chain change under `.claude/hooks/` — Hakim (Security Auditor) re-review follows per the auto-fire trigger.

Note: the review-post itself tripped the *installed* (pre-fix) validator's over-match bug — the exact defect this PR fixes — which is a live demonstration of why it is worth merging.

---
Reviewed commit: `80c62a159cf4f9b01a585d466441301cc4b29a4c`

me2resh (approved):

PR #811, opened 2026-07-06
docs(GH-809): repo-targeting convention — always pass an explicit repo flag, and document existing write guards

### LINKED ISSUE #809: [Bug] Missing mechanical safeguard prevents agent from reading/writing to wrong GitHub repo
## Affected
GitHub access handling / tracker integration — reads via `gh issue view` and writes via `gh issue create` / `/task` / `/report-apexyard-bug`

## Given / When / Then

**Given**: A user has registered private managed projects in `apexyard.projects.yaml`. The public `me2resh/apexyard` repository has tickets with overlapping numeric IDs (e.g., both have issue #92).

**When**: The agent is asked to check a ticket by number (e.g., "what's the status of #92?") or file an issue in a managed project. The agent reads or writes without a mechanical verification of which repo is the target.

**Then**: The agent reads from `me2resh/apexyard#92` instead of the registered private project's #92, returns wrong state information, or — in the write case — could file issues in the public repo instead of the private project.

**Expected**: The agent should only read from and write to registered private projects. Both operations should be blocked mechanically at the shell level, not gated by prompt-level safeguards alone.

## Repro

1. Register a private managed project in `apexyard.projects.yaml`
2. Ask the agent: "what's the status, quick check only, of #92"
3. Agent returns details from `https://github.com/me2resh/apexyard/issues/92` (if it exists) instead of asking which repo or checking active context

## Root Cause

Current safeguards are **prompt-level only**:
- Memory rules ("apexyard repo off-limits for writes")
- Checking `.claude/session/current-ticket` for active repo context
- Human confirmation before `/task` creates an issue

These are insufficient because:
- **Reads** can silently return wrong data without detection, corrupting the agent's state and misleading the user
- **Writes** require human approval, but the human is deciding based on potentially corrupted information (if the agent read from the wrong repo)
- A **mechanical hook** (similar to `require-active-ticket.sh` or `block-private-refs-in-public-repos.sh`) is the only design that prevents the action categorically

## Severity
**blocker** — data integrity is compromised (wrong reads), privacy/security boundary is at risk (could leak private project issues to public repo)

## Framework version
v4.2.0

## Glossary
| Term | Definition |
|------|------------|
| Managed projects | Private repositories registered in `apexyard.projects.yaml` |
| Mechanical safeguard | A shell hook that prevents an action entirely, not a prompt-level rule |
| Prompt-level safeguard | A rule or memory that relies on the agent remembering and applying it correctly |

atlas-apex:
## Summary

- **Encodes the maintainer resolution on #809 as a convention, not a gate.** The bug report asked for a *blocking hook* that only lets the agent touch registered private projects. That design isn't safe — the framework reads and writes `me2resh/apexyard` by design (`/report-apexyard-bug`, `/request-apexyard-feature`, upstream framework PRs, reading issues like #809 itself), and a `PreToolUse` hook sees a *number, not intent*, so it can't tell a wrong-repo access from a legitimate cross-repo one. A blocking hook breaks upstream; an advisory one fires on every legitimate bare read (pure noise). The right shape is a self-discipline rule plus documenting the guards that already exist.
- **Adds a "Repo targeting — always pass an explicit `--repo`" section to `ticket-vocabulary.md`** (the rule that already owns tracker/repo discipline — so no 16th rule file, and no rule-count references to update across CLAUDE.md / README / AGENTS.md). It states the rule (always `--repo <owner/repo>`, never ambient cwd resolution), and explains *why it matters most on bare reads*: numeric IDs collide across a portfolio, and `gh issue view <N>` with no `--repo` silently returns the wrong repo's issue with nothing downstream to flag it — worse than a wrong write, which has a human approval step in front of it.
- **Documents the two shipped write-side backstops so readers know the residual gap is read-only:** `require-skill-for-issue-create.sh` forces issue creation through the structured skills (which resolve + pass explicit `--repo` from the registry, so ambient-repo creation is already blocked), and `block-private-refs-in-public-repos.sh` scrubs private-project names/slugs/paths out of public-repo writes (the privacy-boundary half of "wrong repo"). Both are cross-linked to `leak-protection.md`.

## Testing

- Verified the two named hooks exist and do what the rule claims:
  - `require-skill-for-issue-create.sh` — blocks raw `gh issue create` / `gh api repos/.../issues` **POST** (it downgrades read-only `gh api` GETs to a no-op, and only fires on `/issues` + a write signal) unless a structured ticket-skill marker is in flight; the structured skills pass an explicit `--repo` from the registry.
  - `block-private-refs-in-public-repos.sh` — documented in `leak-protection.md` as scrubbing registered private-project `name` / `repo` slug / `workspace` path out of public-repo issue/PR/comment writes.
- Confirmed the relative link `leak-protection.md` resolves (same `.claude/rules/` directory).
- Docs-only change; extends an existing rule, so the "15 modular rule files" counts in CLAUDE.md / `docs/whats-inside.md` / AGENTS.md are untouched and remain truthful.

## Glossary

| Term | Definition |
|------|------------|
| Ambient (cwd) repo resolution | `gh`/`glab`/etc. resolving a bare `<N>` against whichever repo the current working directory belongs to, when no `--repo` is passed |
| Bare read | A tracker read with no explicit target, e.g. `gh issue view 92` — the highest-risk case because nothing downstream flags a wrong-repo result |
| Write-side backstop | A shipped `PreToolUse` hook that mechanically guards the *write* path (create / public-repo write); reads are left to the convention |
| Residual gap | The narrow read-only case the convention closes: colliding IDs + wrong cwd + omitted `--repo` on a bare read |

Implements the maintainer resolution posted on #809 (convention + documenting existing write guards, not a blocking hook).

Closes #809

https://claude.ai/code/session_01XAQ1mgxusECbQaRAdiBwRj

### DIFF
--- a/.claude/rules/ticket-vocabulary.md
+++ b/.claude/rules/ticket-vocabulary.md
@@ -123,6 +123,42 @@ When you're working in a fork (e.g. `your-org/apexyard`) with `origin` = the for
 
 Earlier versions of these hooks were origin-only, which forced the cross-repo workaround for fork → upstream PRs and silently broke GitHub's auto-close — leaving every cross-fork PR's issue OPEN forever. After #207, bare `#N` is the supported pattern and the cross-repo workaround is no longer needed (it still passes the hook for backwards compat, but you lose auto-close).
 
+## Repo targeting — always pass an explicit `--repo` (me2resh/apexyard#809)
+
+The reserved-vocabulary rule above assumes you already know **which repo** a `#N` lives in. Across a portfolio that assumption is dangerous: numeric issue IDs collide freely across repos — `me2resh/apexyard#92`, a managed project's `#92`, and any other repo's `#92` are three unrelated issues. If you run a tracker command that omits the target repo, `gh` (and the equivalent Linear / Jira / Asana CLIs) resolves it against the **ambient (cwd) repo** — whichever repo the current directory happens to belong to. When the working directory isn't the one you meant, that silently reads or writes the *wrong* `#92`.
+
+**The rule: always pass an explicit `--repo <owner/repo>` on `gh issue` / `gh pr` (and equivalent tracker) commands. Never rely on ambient cwd resolution across a portfolio.** This matters most on bare **reads** — `gh issue view <N>`, `gh issue view <N> --json state` — where nothing downstream flags the mistake: the command succeeds, returns a real issue's state, and quietly feeds the wrong data into your reasoning. A wrong read is worse than a wrong write, because the write path has a human approval step in front of it and the read path does not.
+
+```bash
+# WRONG — resolves against whatever repo the cwd belongs to
+gh issue view 92 --json state,title
+
+# RIGHT — the target repo is stated, not inferred
+gh issue view 92 --repo me2resh/apexyard --json state,title
+```
+
+The same discipline is what `/start-ticket` already bakes in: it records the ticket's fully-qualified `owner/repo` in the session marker (see `.claude/skills/start-ticket/SKILL.md` § "Cross-repo note"), so pass the qualified form (`me2resh/flat-mate#128`) whenever you're in one repo but the ticket lives in another.
+
+### The write side is already mechanically guarded
+
+Two shipped hooks already close the *write* half of this gap — you don't need to remember them, but knowing they're there tells you where the residual risk actually is (reads):
+
+| Hook | Guards | How it helps repo-targeting |
+|------|--------|-----------------------------|
+| `require-skill-for-issue-create.sh` (#268) | Issue **creation** | Blocks a bare `gh issue create` / `gh api repos/.../issues` POST unless a structured skill (`/task`, `/feature`, `/bug`, `/spike`, `/migration`, `/investigation`, `/idea`) is in flight. Those skills resolve the target from the registry and pass an explicit `--repo` — so ambient-repo creation is already off the table. See [`leak-protection.md`](leak-protection.md) for the sibling concern. |
+| `block-private-refs-in-public-repos.sh` | Public-repo **writes** (issue / PR / comment) | Scrubs registered private-project names, repo slugs, and workspace paths out of anything written to a public framework repo — the privacy-boundary half of "wrong repo". Full rule: [`leak-protection.md`](leak-protection.md). |
+
+So the genuine residual gap is narrow: a bare `gh issue view <N>` with no `--repo`, when IDs collide **and** the working directory isn't the intended repo **and** `--repo` was omitted. Real, but read-only — and closed by the discipline above, not by a gate.
+
+### Why a convention, not a hook
+
+A `PreToolUse` hook sees a **number, not intent**. When the command is `gh issue view 92`, the hook cannot tell a *wrong-repo* access from a *legitimate cross-repo* one — and the framework reads and writes `me2resh/apexyard` **by design** (`/report-apexyard-bug`, `/request-apexyard-feature`, upstream framework PRs, reading issues like #809 itself). So:
+
+- A **blocking** hook that allowed only registered private projects would break every one of those legitimate upstream paths.
+- An **advisory** hook that fired on every bare read would be almost all noise — the bare-read pattern is overwhelmingly correct in a single-repo cwd.
+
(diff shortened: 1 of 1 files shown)


atlas-apex (commented):
## Code Review: PR #811

**Commit**: `c8ea2dc04866d10da80c98cae370eee9564733f6`

### Summary

Extends `.claude/rules/ticket-vocabulary.md` with a "Repo targeting — always pass an explicit `--repo`" section (+36/-0, docs-only), encoding the maintainer resolution on #809: a self-discipline convention (never rely on ambient cwd repo resolution) plus documentation of the two write-side hooks that already exist. It is deliberately *not* a blocking gate. Clean, single-file change.

### Checklist Results
- Architecture & Design:      Pass (docs, extends the rule that already owns tracker/repo discipline)
- Code Quality:               Pass (narrative, well-structured, worked WRONG/RIGHT example)
- Testing:                    N/A (docs-only)
- Security:                   Pass (accurately documents the leak-protection boundary)
- PR Description & Glossary:   Pass (narrative summary + 4-term glossary)
- Technical Decisions (AgDR): N/A (implements an already-decided maintainer resolution; no new tech decision)
- Adopter Handbooks:          N/A (no handbooks loaded for a docs diff)

### Issues Found

None.

**1. Truthful about both backstops — confirmed by reading the hooks.**

- `require-skill-for-issue-create.sh` — the rule says it "blocks a bare `gh issue create` / `gh api repos/.../issues` POST unless a structured skill is in flight … those skills resolve the target from the registry and pass an explicit `--repo`." Verified accurate:
  - Blocks matched ticket-create CLIs unless the `active-issue-skill` marker is present (step 3/5 of the hook's resolution).
  - The `gh api` family is downgraded to a no-op **unless** the command both targets `/issues` **and** carries a write signal (`-X POST` / `-f` / `-F` / `--field` / …) — exactly as the rule states ("downgrades read-only `gh api` GETs to a no-op, and only fires on `/issues` + a write signal").
  - The "skills pass explicit `--repo`" claim holds: `tracker_create` → `_tracker_create_gh` runs `gh issue create --repo "$repo"`, and the default `tracker_view` template is `gh issue view {id} --repo {owner_repo}`. Registry-resolved, explicit `--repo`. No overstatement.
- `block-private-refs-in-public-repos.sh` — the rule says it "scrubs registered private-project names, repo slugs, and workspace paths out of anything written to a public framework repo." Verified accurate against the hook header + `leak-protection.md`: it matches registered project `name` / `repo` slug / `workspace` path (and `<owner/repo>#<N>` refs) on the five public-repo write shapes and blocks (exit 2). Rule text matches reality.

**2. Consistent with the #809 decision.** The rule frames this as convention/self-discipline (always explicit `--repo`, never ambient resolution), explicitly *not* a blocking hook, and gives the "why not a gate" rationale (a `PreToolUse` hook sees a number, not intent; a blocking hook breaks legitimate upstream paths like `/report-apexyard-bug`, `/request-apexyard-feature`, upstream framework PRs, reading #809 itself; an advisory one is pure noise on the overwhelmingly-correct single-repo bare read). This is a faithful encoding of the maintainer reply posted on #809 — no contradiction.

**3. No rule-count churn.** The diff modifies only the pre-existing `.claude/rules/ticket-vocabulary.md` — no 16th rule file. `.claude/rules/*.md` count stays at 15, so the "15 modular rule files" reference in `CLAUDE.md` remains truthful and is untouched. `AGENTS.md` is not modified by this PR. Since no rule file was added, every count reference remains as-correct-as-before.

**4. Links resolve.** The two cross-links to `leak-protection.md` are same-directory (`.claude/rules/leak-protection.md` exists). The referenced `.claude/skills/start-ticket/SKILL.md` § "Cross-repo note" also exists and accurately records the fully-qualified `owner/repo` behaviour the rule attributes to it.

**5. Scope + quality.** Single file touched; no stray scratch/body file committed (PR file list = `.claude/rules/ticket-vocabulary.md` only). Narrative summary + glossary present. No scope creep.

### Suggestions

None blocking. (Minor, out of scope: `AGENTS.md` in this fork reads "11 modular rule files" — a pre-existing drift unrelated to this PR, which touches neither the count nor the file. Worth a separate housekeeping pass upstream if it also drifted there; not a blocker here.)

### Verdict
**APPROVED**

---
📌 Reviewed commit: `c8ea2dc04866d10da80c98cae370eee9564733f6`

me2resh (approved):

Issue #834, opened 2026-07-09
[Feature] Stage-1 multiharness docs pass — README harness support matrix + docs/harnesses/ index (no rebrand)


me2resh:
## Problem
ApexYard is now genuinely multi-harness in engineering terms (Codex adapter merged #730/AgDR-0088; pi #815, opencode #821, Cursor #831 planned), but the docs/branding still read Claude-Code-only. A Codex or Cursor user landing on the README has no way to know the framework reaches them — while a full "multi-harness" rebrand today would over-claim (no adapter has a live-runtime conformance test yet; Hakim's condition on #730).

## Proposed behaviour — stage-1 docs pass (NOT a rebrand)
1. **Harness support matrix** in the README (near the top, after the pitch): Claude Code — native, full experience · Codex — adapter merged, generated from `.claude/`, live conformance test pending · pi / opencode / Cursor — planned (#815 / #821 / #831). Honest per-harness capability column (gates enforced / advisory only / planned).
2. **`docs/harnesses/` index page** — one page per harness (Codex's `docs/codex-adapter.md` moves/links here), the shared-core architecture note (gates are portable bash per AgDR-0086; `.claude/` is the canonical authoring surface; harnesses are transports; model tiers via `.claude/harness-models.json` per AgDR-0088), and the adapter-authoring pattern for future harnesses.
3. Keep "for Claude Code" as the primary tagline. Record the **rebrand trigger** in the page: headline flips to harness-neutral only when ≥2 adapters have live end-to-end conformance proof (then the site + channel positioning follow in a coordinated pass).

## Why it helps adopters
Non-Claude users discover the framework reaches them (TAM multiplier) without the framework over-claiming enforcement it hasn't live-proven — the docs say exactly what works where, today.

## Scope / out of scope
In: README matrix section, docs/harnesses/ index, cross-links from the adapter docs. Out: renaming the repo/tagline, site (yard.apexscript.com) changes, channel positioning — those wait for the rebrand trigger.

## Framework version
v4.3.0 (dev)

## Related
#730 (Codex, merged) · #815 (pi) · #821 (opencode) · #831 (Cursor) · AgDR-0086 (hooks stay bash) · AgDR-0088 (Codex adapter + harness-models matrix).

## Glossary
| Term | Definition |
|------|------------|
| Harness | The agent CLI/IDE runtime apexyard governs (Claude Code, Codex, pi, opencode, Cursor). |
[...]

PR #835, opened 2026-07-09
chore(#832): reconcile duplicate AgDR-0082 — renumber handover-badge to AgDR-0090

me2resh:
## Summary

- **Resolved the duplicate `AgDR-0082`** — two records shared the number on `dev` (`AgDR-0082-handover-badge-offer.md` and `AgDR-0082-pi-gate-dispatcher-adapter.md`), which breaks `/agdr` browse/search/stats and makes every `AgDR-0082` cross-reference ambiguous. Renumbered the **handover-badge** record to **AgDR-0090** (`git mv` + H1 fix), leaving the pi-gate-dispatcher record on 0082.
- **Why handover-badge, not pi** — although handover-badge has the higher *raw* in-repo reference count, all of its references are in this repo and fully repointable in this PR (prose "see AgDR-0082" in `multi-project.md` + `SKILL.md`, plus its own test suite). The pi-gate-dispatcher record's few references are **load-bearing cross-citations** — a markdown link from `AgDR-0088`, a full-path link in `docs/harnesses/pi.md`, the `harness-adapters/pi/*` sources, and external adapter tickets (#815/#821) — that a renumber would strand. Moving handover-badge strands nothing; moving pi would. So it is both the ticket's expected choice and the operationally safe one.
- **Repointed every inbound reference** to the renumbered record — `docs/multi-project.md` (1), the handover `SKILL.md` (4), and `test_badge_offer.sh` (the badge assertions plus the `AgDR-0090-handover-badge-offer.md` file path). The pi record and all its citations were deliberately left untouched.
- **Reference counts (inbound, excluding each record's own file):** handover-badge → 3 files / ~19 lines, all in-repo & repointed here; pi-gate-dispatcher → 2 in-repo files (`pi.md`, `AgDR-0088`) + `harness-adapters/pi/*` + external tickets, all load-bearing & preserved.

## Verification

- `ls docs/agdr | grep -oE 'AgDR-[0-9]{4}' | sort | uniq -d` → **empty** (no duplicate IDs remain; this was the only dupe).
- Exactly one `AgDR-0082-*.md` remains on dev (the pi record); the badge record is now the only `AgDR-0090-*.md`.
- `bash .claude/skills/handover/tests/test_badge_offer.sh` → **all checks pass** against the renumbered file.

## Testing

1. `git grep -n "AgDR-0082"` shows only pi-record references (`pi.md`, `AgDR-0088`, `harness-adapters/pi/*`).
[...]

PR #838, opened 2026-07-09
feat(#831): Cursor gate adapter — generate .cursor/hooks.json delegating to the bash hooks

### LINKED ISSUE #831: [Feature] Cursor gate adapter — generate .cursor/hooks.json delegating to the bash hooks (exit-2 native, failClosed)
## Problem
ApexYard now enforces its gates on Claude Code (native) and Codex (#729/#730, AgDR-0088), with pi (#815) and opencode (#821) adapters planned. Cursor — the largest third-party agent-IDE user base — has no adapter, so Cursor users get zero apexyard governance.

## Proposed behaviour
A **Cursor gate adapter** following the AgDR-0088 declarative-generate pattern: extend/derive from `bin/sync-codex-adapter.sh` to emit `.cursor/hooks.json` (+ `.cursor/rules/*.mdc` for the advisory layer) from the canonical `.claude/` runtime, with every hook **delegating to the unmodified `.claude/hooks/*.sh`** — zero gate-logic duplication, same as Codex.

Verified against the live Cursor hooks docs (https://cursor.com/docs/hooks, 2026-07-09):

- **Pre-execution blocking events exist**: `beforeShellExecution` (receives `command` + `cwd`), `beforeMCPExecution`, and a generic `preToolUse` (`tool_name` + `tool_input` — closest to the Claude Code stdin shape our hooks already parse).
- **Exit code 2 = deny is NATIVE** ("exit code 2 is interpreted as `permission: deny`") — our bash gates already block with exit 2, so delegation is near-passthrough (Codex needed a JSON permission translation; Cursor honors the exit code directly).
- **Hooks are spawned external processes over stdio** — arbitrary bash scripts are first-class.
- **Config**: `<project-root>/.cursor/hooks.json`, `{"version":1,"hooks":{"<event>":[{"command":"…","matcher":"…","timeout":…,"failClosed":…}]}}` — same declarative shape as `.codex/hooks.json`.
- **`failClosed: true`** — a hardening Codex doesn't offer: mark the security-critical gates (merge gates, secrets scan, trust-chain) fail-CLOSED so a hook crash/timeout blocks instead of failing open.

Design notes:
1. **Stdin remap shim**: Cursor's `beforeShellExecution` puts the command at top-level `command`, while our hooks parse `tool_input.command` — the generated wrapper needs a thin field remap (or target `preToolUse`, whose shape is closest). Same class of shim as the Codex `if`-predicate conversion.
2. **Reuse the shared core**: derive the gate list from `.claude/settings.json` (the pattern Tariq endorsed on #730) and read model tiers from `.claude/harness-models.json` (add a `cursor` column only if Cursor agents need model pinning; per AgDR-0087 never downgrade the `opus` reasoning tier).
3. **Live-Cursor conformance test** as an AC (the same gap flagged on Codex): prove a real Cursor session's `gh pr merge` attempt is blocked end-to-end by the delegated `block-unreviewed-merge.sh`.

## Why it helps adopters
Cursor users adopt apexyard governance with one generator run — the same audited bash gates, no parallel implementation to rot. Extends the multiharness family (Claude Code · Codex · pi · opencode) to the biggest remaining harness.

## Scope / out of scope
In: hooks.json generation, `failClosed` on security-critical gates, advisory rules bridge, drift `--check`, smoke test. Out: porting hook logic to TS/JS (AgDR-0086: hooks stay bash), Cursor Tab/completions behaviour, enterprise/team-level hooks distribution.

## Framework version
v4.3.0 (dev)

## Related
AgDR-0088 (Codex adapter — the pattern to follow) · AgDR-0082 (pi precedent) · AgDR-0086 (hooks stay bash) · AgDR-0087 (frontier-model floor) · #815 (pi adapter) · #821 (opencode adapter) · #730 (Codex, merged). Sequence AFTER the shared-core convergence so all three pending adapters (cursor/pi/opencode) inherit settings.json-derived gates + harness-models.json.

## Glossary
| Term | Definition |
|------|------------|
| `failClosed` | Cursor hook option: a crashed/timed-out hook BLOCKS the action instead of failing open — desirable for security-critical gates. |
| `preToolUse` (Cursor) | Cursor's generic pre-tool hook event; receives `tool_name` + `tool_input`, the closest match to the Claude Code hook stdin contract. |
| Declarative-generate adapter | The AgDR-0088 pattern: generate the harness's native hook config from `.claude/settings.json`, delegating execution to the unmodified bash gates. |

me2resh:
## Summary

- **New `bin/sync-cursor-adapter.sh`** generates `.cursor/hooks.json` from `.claude/settings.json`, following the AgDR-0088 declarative-generate pattern already proven by the Codex adapter (#730): every generated hook command still execs the unmodified `.claude/hooks/*.sh` scripts, so gate logic is never forked or duplicated. Cursor's hook vocabulary doesn't name-match Claude Code's (`beforeShellExecution`/`preToolUse`/`postToolUse`/`sessionStart`/`beforeSubmitPrompt` vs. `PreToolUse`/`PostToolUse`/`SessionStart`/`UserPromptSubmit`), so the generator carries an explicit per-matcher mapping table instead of a straight key rename.
- **Stdin remap for shell commands** — Cursor's `beforeShellExecution` event puts the command at the top level of stdin (`{"command": "...", "cwd": ...}`), while every `.claude/hooks/*.sh` script parses Claude Code's shape (`tool_input.command`). The generated command reads the top-level field, applies the same `Bash(glob)` preflight filter Claude's handler-level `if` predicate encodes (unconditional hooks get glob `*` so they're still remapped), re-wraps into the canonical shape, and delegates. Verified end-to-end during development against a real gate (`block-git-add-all.sh` correctly blocked `git add -A` and allowed `git add file.txt` through the full remap path).
- **`failClosed: true` on nine security-critical hooks** — the four merge gates (`block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh`), the secrets scanner (`check-secrets.sh`), and the trust-chain/leak-protection class (`block-git-add-all.sh`, `block-main-push.sh`, `block-private-refs-in-public-repos.sh`, `block-onboarding-in-git.sh`). A crash or timeout on one of these blocks the action instead of silently passing — a hardening option Cursor exposes that neither Claude Code nor Codex do. Everything else stays fail-open, matching its existing posture under Claude Code.
- **`--check` drift detection and `--clean` regeneration**, mirroring the Codex generator's flags exactly.
- **`.cursor/rules/apexyard.mdc`** — a minimal, `alwaysApply: true` advisory bridge that states the gates are mechanical (not this file), lists the five load-bearing rules, and points at `CLAUDE.md`/`.claude/rules/*.md` for the rest. Kept deliberately thin — the gates carry the substance.
- **`docs/cursor-adapter.md`** (mirrors `docs/codex-adapter.md`) and **AgDR-0091** record the event-mapping table, the remap design, the `failClosed` boundary and its rationale, and the accepted conformance gaps — documented rather than silently hidden, matching how AgDR-0088 handled the equivalent Codex-runtime-conformance risk.

## Design decisions worth flagging for review

- **Event mapping**: `PreToolUse`/`Bash` → `beforeShellExecution` (remapped); `PreToolUse`/`{Edit|Write|MultiEdit, Write}` and `Read|Glob|Grep` → `preToolUse` (passthrough, matcher `Write` / `Read|Grep`); `PostToolUse` splits the same way onto `postToolUse`; `SessionStart` → `sessionStart`; `UserPromptSubmit` → `beforeSubmitPrompt`. Full table in `docs/cursor-adapter.md`.
- **Known, documented conformance gaps** (not blocking, called out explicitly): (1) Cursor's own `tool_input` field names for `Write`/`Read` tool types beyond the base envelope aren't published in the docs I could verify, so the `preToolUse`/`postToolUse` passthrough path is unverified against a real session; (2) Cursor's tool-type matcher values (`Shell`/`Read`/`Write`/`Grep`/`Delete`/`Task`) are coarser than Claude Code's `Edit`/`Write`/`MultiEdit` split — both fold into `Write`; (3) a real live-Cursor conformance test (a Cursor session actually blocking `gh pr merge`) is a tracked follow-up, same gap class the Codex adapter carries today.
- **No `.codex/agents/*.toml` equivalent and no `cursor` column on `.claude/harness-models.json`** — Cursor hook generation doesn't need agent model-tier pinning (there's no agent-config surface analogous to Codex's for this adapter to populate). Noted rather than silently added, in case a future PR needs it.
- Did **not** touch `README.md` (owned by the #834 doc-index PR) or `.claude/harness-models.json`.

## Testing

- `bash .claude/hooks/tests/test_sync_cursor_adapter.sh` — new smoke test, 25/25 assertions pass: config shape (`version: 1`), the event-mapping table (Bash-matcher count under `beforeShellExecution`, `Edit|Write|MultiEdit` → `preToolUse` matcher `Write`, `SessionStart` → `sessionStart`), no absolute fixture paths, no leaked `if` metadata, `failClosed` present on the security-critical fixture hook and absent on a non-critical one, delegation exec across the remap boundary (block / allow / skip, including the unconditional-Bash-hook case), `preToolUse` passthrough exit-code preservation, rules-bridge content, `--check` drift detection (clean + dirty), `--clean` regeneration.
- `bash bin/run-hook-tests.sh` (session pin neutralized) — full suite green: **95/95 PASS, 0 FAIL, 0 SKIP**, including the new test alongside the existing 94.
- Ran `bin/sync-cursor-adapter.sh` against this repo's real `.claude/settings.json` (not committed — same untracked-generated-output policy as the Codex adapter) and manually exercised the generated `beforeShellExecution` command for `block-git-add-all.sh` with `git add -A` (blocked, exit 2, full error message rendered), `git add file.txt` (allowed, exit 0), and a non-matching command (skipped, exit 0).
- `shellcheck` on both new files: only pre-existing-pattern info-level notices (SC2015, SC2016) matching the Codex adapter's own style — no new warnings.

Closes #831

---

## Glossary

| Term | Definition |
|------|------------|
| `failClosed` | Cursor hook option: a crashed/timed-out hook BLOCKS the action instead of failing open — used here for merge gates, secrets scan, and trust-chain hooks. |
| `beforeShellExecution` | Cursor's dedicated pre-execution hook for shell commands; stdin carries `command` at the top level (not nested under `tool_input`), which is why the remap shim exists. |
| `preToolUse` (Cursor) | Cursor's generic pre-tool hook event; carries `tool_name`/`tool_input`, closest to Claude Code's stdin contract for non-shell tool calls. |
| Declarative-generate adapter | The AgDR-0088 pattern: generate the harness's native hook config from `.claude/settings.json`, delegating execution to the unmodified bash gates — used by the Codex, pi, and now Cursor adapters. |
| Stdin remap shim | The three-line shell script embedded in each generated `beforeShellExecution` command that reshapes Cursor's top-level `command` field into Claude Code's `{"tool_name":"Bash","tool_input":{"command":...}}` envelope before delegating. |

### DIFF
--- a/.claude/hooks/tests/test_sync_cursor_adapter.sh
+++ b/.claude/hooks/tests/test_sync_cursor_adapter.sh
@@ -1,0 +1,297 @@
+#!/usr/bin/env bash
+# Smoke tests for bin/sync-cursor-adapter.sh.
+
+set -u
+
+ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+SCRIPT="$ROOT/bin/sync-cursor-adapter.sh"
+
+red()   { printf '\033[31m%s\033[0m\n' "$*"; }
+green() { printf '\033[32m%s\033[0m\n' "$*"; }
+
+PASS=0
+FAIL=0
+FAILED=""
+
+mark_pass() { green "  ok   $1"; PASS=$((PASS+1)); }
+mark_fail() {
+  red "  FAIL $1: $2" >&2
+  FAIL=$((FAIL+1))
+  FAILED="$FAILED $1"
+}
+
+assert_file() {
+  local path="$1" label="$2"
+  [ -f "$path" ] && mark_pass "$label" || mark_fail "$label" "missing $path"
+}
+
+assert_contains() {
+  local path="$1" pattern="$2" label="$3"
+  grep -F "$pattern" "$path" >/dev/null 2>&1 && mark_pass "$label" || mark_fail "$label" "missing pattern [$pattern] in $path"
+}
+
+assert_not_contains() {
+  local path="$1" pattern="$2" label="$3"
+  if grep -F "$pattern" "$path" >/dev/null 2>&1; then
+    mark_fail "$label" "unexpected pattern [$pattern] in $path"
+  else
+    mark_pass "$label"
+  fi
+}
+
+# Neutralize the Claude Code session-pin artifact so the generated wrapper's
+# ops-root resolution falls through to the directory-walk branch, same as it
+# would under a real Cursor session (which never sets this var).
+unset CLAUDE_CODE_SESSION_ID
+
+TMPROOT=$(mktemp -d "${TMPDIR:-/tmp}/cursor-adapter-test.XXXXXX")
+trap 'rm -rf "$TMPROOT"' EXIT
+
+mkdir -p "$TMPROOT/.claude/hooks"
+touch "$TMPROOT/.apexyard-fork"
+
+cat > "$TMPROOT/.claude/hooks/block-test.sh" <<'SH'
+#!/usr/bin/env bash
+input=$(cat)
+case "$input" in
+  *deny*) exit 2 ;;
+  *) exit 0 ;;
+esac
+SH
+chmod +x "$TMPROOT/.claude/hooks/block-test.sh"
+
+cat > "$TMPROOT/.claude/hooks/unconditional-test.sh" <<'SH'
+#!/usr/bin/env bash
+input=$(cat)
+case "$input" in
+  *unconditional-deny*) exit 2 ;;
+  *) exit 0 ;;
+esac
+SH
+chmod +x "$TMPROOT/.claude/hooks/unconditional-test.sh"
+
+cat > "$TMPROOT/.claude/hooks/check-secrets.sh" <<'SH'
+#!/usr/bin/env bash
+input=$(cat)
+case "$input" in
+  *secret*) exit 2 ;;
+  *) exit 0 ;;
+esac
+SH
+chmod +x "$TMPROOT/.claude/hooks/check-secrets.sh"
+
+cat > "$TMPROOT/.claude/hooks/require-active-ticket.sh" <<'SH'
+#!/usr/bin/env bash
+cat >/dev/null
+exit 0
+SH
+chmod +x "$TMPROOT/.claude/hooks/require-active-ticket.sh"
+
+cat > "$TMPROOT/.claude/hooks/pin-ops-root.sh" <<'SH'
+#!/usr/bin/env bash
+cat >/dev/null
+exit 0
+SH
+chmod +x "$TMPROOT/.claude/hooks/pin-ops-root.sh"
+
+# Fixture settings.json exercises: a Bash(glob) predicate hook, an
+# unconditional Bash hook (no `if`), a fail-closed-eligible hook
+# (check-secrets.sh), an Edit|Write|MultiEdit hook, and a SessionStart hook —
+# covering every branch the generator's event-mapping table has to handle.
+cat > "$TMPROOT/.claude/settings.json" <<'JSON'
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          { "type": "command", "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/pin-ops-root.sh\"'" }
+        ]
+      }
+    ],
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "if": "Bash(policy *)",
+            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/block-test.sh\"'"
+          },
+          {
+            "type": "command",
+            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/unconditional-test.sh\"'"
+          },
+          {
+            "type": "command",
+            "if": "Bash(git commit *)",
+            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/check-secrets.sh\"'"
+          }
+        ]
+      },
+      {
+        "matcher": "Edit|Write|MultiEdit",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash -c 'r=\"\";if [ -n \"${CLAUDE_CODE_SESSION_ID:-}\" ];then p=\"${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-${CLAUDE_CODE_SESSION_ID}\";[ -f \"$p\" ] && IFS= read -r r < \"$p\" && [ -d \"$r/.claude/hooks\" ] || r=\"\";fi;if [ -z \"$r\" ];then r=$PWD;while [ -n \"$r\" ] && [ \"$r\" != / ];do { [ -f \"$r/.apexyard-fork\" ] || [ -f \"$r/onboarding.yaml\" ]; } && [ -d \"$r/.claude/hooks\" ] && break;r=${r%/*};done;fi;[ -d \"$r/.claude/hooks\" ] || exit 0;exec \"$r/.claude/hooks/require-active-ticket.sh\"'"
+          }
+        ]
+      }
+    ]
+  }
+}
+JSON
+
+echo "== Cursor adapter sync smoke"
+
+if bash "$SCRIPT" --root "$TMPROOT" >/tmp/_cursor_adapter_sync.out 2>&1; then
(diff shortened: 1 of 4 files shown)


me2resh (commented):
## Security Review: PR #838 — Cursor gate adapter

**Commit**: `2b8cd76d2a9c3dd4a1e34aad7678cc040ffa483e`

### Summary
`bin/sync-cursor-adapter.sh` generates `.cursor/hooks.json` from `.claude/settings.json`, delegating every gate to the unmodified `.claude/hooks/*.sh`. Same trust-chain class as the Codex adapter (#730), held to the same bar. I ran the smoke test (pin neutralized), generated the real adapter, and live-traced generated commands through the remap. Verdict: **APPROVED** — no gate is silently weakened through the Cursor path.

### Checklist Results
- Secrets & Credentials:  **Pass** — no absolute paths, no secrets in generated/committed output; `.cursor` is untracked (generated-output policy, mirrors Codex)
- Injection Prevention:   **Pass** — globs, hook commands, and the remap shim are all `@sh`-quoted by jq; hook basenames come from controlled `.claude/hooks/` content
- Enforcement Fidelity:   **Pass** — blocking exit 2 propagates through the remap; both merge shapes gated
- failClosed Correctness: **Pass** — the 9 security-load-bearing gates fail-closed, no advisory hook does
- No Enforcement Fork:    **Pass** — zero hook bodies copied; all 48 shell entries `exec "$r/.claude/hooks/*.sh"`
- Conformance Honesty:    **Pass** — unverified paths documented and, critically, carry no security gate

### What I verified (live traces, not just reading)
1. **Enforcement fidelity through the stdin remap** — fed Cursor-shaped top-level `{"command":...}` stdin into the *generated* `block-git-add-all.sh` command: `git add -A` → **exit 2 (blocked)**, `git add file.txt` → **exit 0 (allowed)**. The re-wrap into `{"tool_name":"Bash","tool_input":{"command":...}}` is faithful — feeding that canonical shape directly to the hook also blocks (exit 2), while the raw un-wrapped top-level `{command}` returns 0, confirming the shim is necessary and correct.
2. **Both merge shapes still hit the merge gate** — the generator emits a separate `beforeShellExecution` entry per `if:` predicate, so `gh pr merge *` **and** `gh api *` (plus `glab mr merge`, `glab api`, `tracker_pr_merge`) each delegate to `block-unreviewed-merge.sh`, all with `failClosed:true`. The #47 dual-shape coverage survives the Cursor path. The glob preflight (`[[ "$cmd" == $GLOB ]]`) faithfully reproduces the handler-level `if:` predicate; unconditional hooks get glob `*` and stay unconditional (verified via trace).
3. **failClosed audit** — exactly the 9 intended gates are fail-closed: the four merge gates (`block-unreviewed-merge`, `block-merge-on-red-ci`, `require-design-review-for-ui`, `require-architecture-review`), `check-secrets`, and the trust/leak class (`block-git-add-all`, `block-main-push`, `block-private-refs-in-public-repos`, `block-onboarding-in-git`). **No advisory/process/format hook is fail-closed** (no DoS-the-user footgun). All fail-closed gates ride the `beforeShellExecution` event only. Note this is a net *hardening* over Claude Code: a merge-gate hook that crashes/times out now blocks in Cursor, where Claude Code treats a non-2 exit as a non-blocking error.
4. **No enforcement fork** — all 48 `beforeShellExecution` entries end in `exec "$r/.claude/hooks/*.sh"`; the shim only reshapes stdin, it copies zero gate logic.
5. **Conformance-gap containment** — the PR honestly documents the Write/Read `preToolUse`/`postToolUse` passthrough as unverified against a real Cursor session. Crucially, that unverified path carries **only** process/advisory hooks (`require-active-ticket`, `detect-role-trigger`, `warn-review-marker-write`, `suggest-mcp-search`, `auto-code-review`, `maintain-docs-index`). Every enforcement-critical gate (merge, secrets, leak) rides the **verified** `beforeShellExecution` shell path. So the documented gap does not undermine any security gate. The PR does not overclaim.

### Findings

#### [LOW / informational] Fail-open on malformed or non-top-level `beforeShellExecution` stdin
The remap shim extracts `.command` and, on no match (including unparseable JSON or a payload where the command is not at top level), exits 0 — allowing the command. I confirmed this by trace: malformed stdin → exit 0, nested `{"tool_input":{"command":...}}` → exit 0. Because the shim exits **0 (clean allow)** rather than crashing, `failClosed:true` does **not** engage for this case.

Why this is **not blocking**:
- It **matches the underlying hooks' own native posture** — feeding malformed stdin directly to `block-git-add-all.sh`, `check-secrets.sh`, and `block-unreviewed-merge.sh` under Claude Code also returns exit 0. The adapter faithfully reproduces existing behaviour; it introduces no new weakening.
- The normal path is the top-level `.command` shape the PR states was verified end-to-end, and a payload with no extractable command is a payload with no shell command to gate.

Residual risk to track (**not** a merge blocker): if Cursor's `beforeShellExecution` schema ever diverges from top-level `.command` (a version drift), **all** shell gates would fail open silently and `failClosed` would not catch it (the shim exits 0, not a crash). This is exactly what the PR's committed follow-up — a live-Cursor conformance test that actually blocks `gh pr merge` — should pin down. Recommend that follow-up assert the top-level-`.command` contract explicitly.

### Verdict
**APPROVED** — security marker written at HEAD `2b8cd76`. The adapter delegates rather than forks, both merge shapes stay gated through the remap, `failClosed` is correctly scoped to security-load-bearing gates only, and every enforcement-critical gate rides the verified shell path. The one fail-open edge matches the existing hook suite's native posture and is bounded by a follow-up the PR already commits to.

---
📌 Reviewed commit: `2b8cd76d2a9c3dd4a1e34aad7678cc040ffa483e`

me2resh (commented):
## Code Review: PR #838

**Commit**: `2b8cd76d2a9c3dd4a1e34aad7678cc040ffa483e`

### Summary
Adds `bin/sync-cursor-adapter.sh`, extending the AgDR-0088 declarative-generate pattern (Codex) to Cursor. Generates `.cursor/hooks.json` from `.claude/settings.json` with an explicit event-mapping table, a `beforeShellExecution` stdin-remap shim, and `failClosed: true` on nine security-critical gates — every generated command still execs the unmodified `.claude/hooks/*.sh`, so gate logic is never forked. Plus a 25-assertion smoke test, `docs/cursor-adapter.md`, and AgDR-0091.

### Checklist Results
- Architecture & Design:      Pass — single source of truth (`.claude/`) preserved; delegation-only, no duplicated gate logic
- Code Quality:               Pass — shellcheck clean (only SC2015/SC2016 info-level, matching the Codex adapter's own style)
- Testing:                    Pass — smoke 25/25, full suite 95/95 (both re-run locally with the session pin neutralized)
- Security:                   Pass — `failClosed` correctly hardens the 9 named gates; no absolute-path leak; JSON re-encode via jq (no injection)
- Performance:                Pass — N/A (generator script)
- PR Description & Glossary:  Pass — narrative Summary, Glossary present, `Closes #831`, single-line subject
- Technical Decisions (AgDR): Pass — AgDR-0091 present, no number collision, honest about conformance gaps
- Adopter Handbooks:          N/A — no handbooks loaded for this diff

### Focus-area verification

**1. Stdin remap correctness — VERIFIED.** The shim reads Cursor's top-level `.command`, applies the `[[ "$cmd" == $GLOB ]]` preflight (glob sourced from the Claude `if: Bash(...)` predicate), re-wraps via `jq -c` into `{"tool_name":"Bash","tool_input":{"command":…}}`, and pipes to the original hook. Exit code is the pipe's exit = the hook's exit, so block=2 is preserved (confirmed by the smoke test's block/allow/skip triple and by the exit-code assertion). Unconditional Bash hooks (no `if`) get glob `*` and are remapped, not skipped — verified both in `if_glob` (`else "*"`) and by the dedicated "unconditional Bash hook still blocks via the remap" assertion. JSON re-encoding is done by jq, so no hand-quoting/injection risk.

**2. Event-mapping completeness — VERIFIED.** Generated 69 hooks from 69 wired source hooks — nothing dropped. Cross-checked per bucket: `beforeShellExecution` 48 = PreToolUse/Bash 48; `preToolUse` 5 = (Edit|Write|MultiEdit + Write) 4 + (Read|Glob|Grep) 1; `postToolUse` 5 = Bash 4 + Write|Edit|MultiEdit 1; `sessionStart` 10; `beforeSubmitPrompt` 1. Every wired PreToolUse/PostToolUse/SessionStart/UserPromptSubmit matcher maps. All `if` predicates are `Bash(...)` shape and parse cleanly.

**3. failClosed list — VERIFIED.** All 9 named hooks appear in the output (29 entries total — merge gates and the leak-scrubber recur once per command shape), and `failClosed` never leaks onto a non-Bash event. Distinct scripts: the 4 merge gates, check-secrets, and 4 trust/leak hooks — exactly as declared.

**4. Smoke test + full suite — VERIFIED.** 25/25 and 95/95, both re-run here.

**5. AgDR-0091 — VERIFIED.** 0091 is unique in this tree (no 0090/0091/0092 on `dev` base; only 0091 added by this PR — no collision with the in-flight #835/#821). Honest about the accepted gaps: `preToolUse`/`postToolUse` `tool_input` field names unverified against a real Cursor session, coarser Cursor matcher vocabulary, and a live-Cursor conformance test tracked as a follow-up.

**6. Standards — VERIFIED.** No `README.md` touch, no `.claude/harness-models.json` touch (both correctly out of scope and called out). Diff is 4 files. Marker SHA matches HEAD.

- **suggestion:** The generator has no "unmapped matcher → fail loud" assertion. Today every wired matcher is covered, but the mapping is a hardcoded `select(...)` list — if a *new* matcher shape is later added to `.claude/settings.json`, it would be silently dropped rather than failing loud (the #730 advisory). The reference Codex adapter shares this behaviour, so this isn't a regression — but a cheap completeness check (assert the sum of mapped entries equals the source hook count, or diff the consumed matchers against the settings' matcher set) would turn a future silent gate-hole into a build-time error. Worth a follow-up ticket rather than blocking here.
- **consideration (failClosed boundary):** `require-migration-ticket.sh` is the strongest candidate not on the fail-closed list — it's a blast-radius/data-loss gate. The AgDR's classification (merge gates + secrets + trust/leak) is coherent and the boundary is explicitly documented as the place to extend, so this is a judgment call, not a defect. Flagging so a future migration-hardening PR revisits it. `validate-pr-create.sh` is correctly left fail-open (format enforcement only).

### Verdict
**APPROVED**

The remap preserves exit codes, the event mapping is provably complete for the current settings, the failClosed set is correct and correctly scoped, tests pass, and the AgDR is honest about the remaining live-Cursor conformance gap. The two notes above are follow-ups, not blockers.

---
Reviewed commit: `2b8cd76d2a9c3dd4a1e34aad7678cc040ffa483e`

atlas-apex (approved):
Non-author approval for the dev ruleset. Rex (+ Hakim on the adapter PRs) approved; CEO batch-approved.

Issue #844, opened 2026-07-09
[Bug] opencode adapter install layout crashes real opencode 1.17.16 — helpers/util exports loaded as plugins; dir is .opencode/plugins (plural)


me2resh:
## Bug Scenario
**Given** the opencode adapter's documented install (`docs/opencode-adapter.md` / adapter README): `cp harness-adapters/opencode/src/*.ts .opencode/plugin/`
**When** a real opencode session (verified: opencode **1.17.16**) starts in that project
**Then** opencode crashes at startup with `Error: {"name":"UnknownError","message":"Unexpected server error"}` and NO gate is enforced — the exact opposite of the adapter's purpose.

## Root cause (verified against `@opencode-ai/plugin` type contract + live isolation)
1. **Discovery dir is `.opencode/plugins/` (plural), not `.opencode/plugin/`** — the adapter docs use the singular form.
2. **opencode treats EVERY exported async function in a discovered file as a plugin.** `gate-dispatcher.ts` also exports utilities (`execGateHookReal(hookPath,…)`, `buildToolExecuteBeforeHook`, `registerGateDispatcher`, …). opencode invokes them as `plugin(input)` — e.g. `execGateHookReal` runs `execFileSync(input, …)` and throws at load → the server crash.
3. Copying the **helper files** (`derive-gates.ts`, `resolve-ops-root.ts`) straight into the discovery dir makes opencode try to load THEM as plugins too.

**Isolation proof:** the same fixture runs clean (`PONG`, exit 0) with the plugin dir removed, and crashes with it present. Corrected layout — a single re-export entry in `.opencode/plugins/` whose only export is the default plugin, with helpers in a NON-discovered dir — loads cleanly (verified `PONG`, exit 0).

## Fix
- Ship the adapter as an **entry file whose ONLY export is the plugin** (default or a single named `server`), placed in `.opencode/plugins/` (plural).
- Keep helpers + the utility exports OUT of any discovered path (subdir `.opencode/plugins/apexyard/` with only `index.ts` discovered, or a re-export shim importing from `harness-adapters/`).
- Update `docs/opencode-adapter.md` install instructions to the working layout.
- Add the layout to the opencode smoke test so a regression to the crashing shape is caught.

## Acceptance Criteria
- [ ] Documented install layout loads cleanly in opencode 1.17.16+ (no startup crash).
- [ ] Only the plugin entry is exposed to opencode's loader; utilities are not invoked as plugins.
[...]

PR #847, opened 2026-07-09
fix(#840): Cursor adapter must install at user-level hooks.json, not project

me2resh:
## Summary

- **Cursor 3.x only loads `~/.cursor/hooks.json` (user-level), not the project `.cursor/hooks.json` this adapter previously wrote.** Live testing against a real Cursor.app 3.10.20 session confirmed the project file shows `Configured Hooks (0)` — never loaded, no trust prompt — while installing the identical, unmodified generated file at the user level shows `Configured Hooks (1)` and gets enforced. This was a location bug, not a schema bug: `version: 1`, `beforeShellExecution`, and `failClosed` were already correct.
- **`bin/install-cursor-adapter.sh` is the new primary entrypoint** (mirrors the pi/opencode install-script pattern from #844/#845): it merges the generated hooks into `~/.cursor/hooks.json`, preserving any pre-existing non-apexyard entries (identified structurally — every apexyard entry execs a `.claude/hooks/*.sh` script), with a timestamped backup before every write and a symmetric `--uninstall`. `bin/sync-cursor-adapter.sh` gains `--user`/`--user-dir` to power the merge, reusing its existing jq generation pipeline unchanged — no gate logic forked, and default (project-level) behaviour is untouched.
- **`cursor-agent` (the CLI) is confirmed NOT covered by this adapter** — live testing instrumented every generated hook and ran a benign command through the CLI: it executed with zero hooks firing. The CLI enforces via its own `~/.cursor/cli-config.json` `permissions.allow`/`deny` model instead, a separate surface this adapter doesn't address.
- **IDE enforcement currently rests on `failClosed`, not confirmed clean delegated execution — documented honestly rather than overclaimed.** With the adapter correctly installed at the user level, a real agent turn's `git add .` was blocked. But the delegated hook's own instrumentation never fired, and the agent reported `MainThreadShellExec not initialized` when probed directly — the defensible read is that Cursor's hook-runner couldn't cleanly exec the delegated bash in this version, and `failClosed: true` denied the action because the hook errored, not because the gate logic ran. This is weaker than the opencode/pi adapters, where delegated bash genuinely executes. Docs now say this plainly instead of claiming delegation is proven.
- **Cursor is now the sole live-conformance outlier among the four third-party adapters.** opencode and pi are live-proven. Codex carried the same "unverified" caveat as Cursor when AgDR-0088/AgDR-0091 were written, but that gap has separately been closed via live testing (it was a trust-prompt issue, fixed via `--dangerously-bypass-hook-trust` / a user-level trust grant) — Codex is now live-proven too. `docs/cursor-adapter.md`, `docs/harnesses/cursor.md`, and AgDR-0091 all state this comparison explicitly now.

## Testing

```bash
bin/install-cursor-adapter.sh                # merges into ~/.cursor/hooks.json (default), refreshes project rules bridge
bin/install-cursor-adapter.sh --uninstall     # removes only apexyard's own entries
```

- `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_sync_cursor_adapter.sh` — 44 assertions (20 new: `--user` merge preserves foreign entries, drops stale apexyard entries, idempotent re-run, `--check` drift detection, fresh-user-dir bootstrap), all pass.
- `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_install_cursor_adapter.sh` (new) — 13 assertions covering default install, re-install preserving a foreign hook, `--uninstall` scoped removal + backup, and the no-op-when-nothing-installed case, all pass.
- `env -u CLAUDE_CODE_SESSION_ID bash bin/run-hook-tests.sh` — full suite, 96/96 pass.
- `shellcheck bin/install-cursor-adapter.sh` — clean. `shellcheck bin/sync-cursor-adapter.sh` — only a pre-existing, unrelated SC2016 info note (an intentionally single-quoted `sed` pattern, present before this PR).

## Glossary

| Term | Definition |
|------|------------|
| `beforeShellExecution` | Cursor's dedicated pre-shell-command hook event; the adapter remaps its top-level `command` field into the Claude Code stdin shape before delegating to `.claude/hooks/*.sh`. |
| `failClosed` | A Cursor hook-entry option: if the hook command crashes, times out, or (per this PR's finding) fails to execute cleanly, the action is **blocked** instead of allowed through. |
| `cursor-agent` | Cursor's separate CLI product. Does not read `hooks.json`; enforces via its own `cli-config.json` permissions model. Not addressed by this adapter. |
| Apexyard-owned entry | A hook entry in `hooks.json` whose `command` execs a `.claude/hooks/*.sh` script — identified structurally (string match), not via a custom JSON field, so merge/uninstall can tell "ours" from "the user's own hooks" without extra bookkeeping. |
| `MainThreadShellExec not initialized` | The error Cursor's agent reported when the delegated bash hook was probed directly — the basis for concluding enforcement is `failClosed`-backed rather than confirmed delegated-execution-backed. |

### DIFF
--- a/.claude/hooks/tests/test_install_cursor_adapter.sh
+++ b/.claude/hooks/tests/test_install_cursor_adapter.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env bash
+# Smoke tests for bin/install-cursor-adapter.sh — the install/uninstall
+# lifecycle wrapper around `bin/sync-cursor-adapter.sh --user` (#840).
+# Complements test_sync_cursor_adapter.sh's --user assertions (which cover
+# the merge/idempotency/drift logic in the generator itself); this file
+# covers install-cursor-adapter.sh's own surface: default invocation,
+# --uninstall, and its scoped-removal / backup guarantees.
+
+set -u
+
+ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+SCRIPT="$ROOT/bin/install-cursor-adapter.sh"
+
+red()   { printf '\033[31m%s\033[0m\n' "$*"; }
+green() { printf '\033[32m%s\033[0m\n' "$*"; }
+
+PASS=0
+FAIL=0
+FAILED=""
+
+mark_pass() { green "  ok   $1"; PASS=$((PASS+1)); }
+mark_fail() {
+  red "  FAIL $1: $2" >&2
+  FAIL=$((FAIL+1))
+  FAILED="$FAILED $1"
+}
+
+assert_file() {
+  local path="$1" label="$2"
+  [ -f "$path" ] && mark_pass "$label" || mark_fail "$label" "missing $path"
+}
+
+# Never let this test touch a real $HOME — every invocation below passes an
+# explicit --user-dir into a throwaway fixture tree.
+unset CLAUDE_CODE_SESSION_ID
+
+TMPROOT=$(mktemp -d "${TMPDIR:-/tmp}/install-cursor-adapter-test.XXXXXX")
+trap 'rm -rf "$TMPROOT"' EXIT
+
+mkdir -p "$TMPROOT/.claude/hooks" "$TMPROOT/bin"
+touch "$TMPROOT/.apexyard-fork"
+
+# install-cursor-adapter.sh's --root is "the repo where bin/ and .claude/
+# live together" (the same repo, not a separate framework clone — every
+# apexyard-governed project carries its own copy of both). Mirror that
+# shape in the fixture: copy the real generator script in alongside a
+# synthetic .claude/ tree, so this test drives the real
+# sync-cursor-adapter.sh logic without depending on — or mutating — this
+# repo's own live .claude/ or .cursor/ state.
+cp "$ROOT/bin/sync-cursor-adapter.sh" "$TMPROOT/bin/sync-cursor-adapter.sh"
+chmod +x "$TMPROOT/bin/sync-cursor-adapter.sh"
+cat > "$TMPROOT/.claude/hooks/check-secrets.sh" <<'SH'
+#!/usr/bin/env bash
+cat >/dev/null
+exit 0
+SH
+chmod +x "$TMPROOT/.claude/hooks/check-secrets.sh"
+
+cat > "$TMPROOT/.claude/settings.json" <<'JSON'
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          { "type": "command", "command": "bash -c 'exec \"$PWD/.claude/hooks/check-secrets.sh\"'" }
+        ]
+      }
+    ]
+  }
+}
+JSON
+
+USERDIR="$TMPROOT/fake-home/.cursor"
+
+echo "== install-cursor-adapter.sh install"
+
+if bash "$SCRIPT" --root "$TMPROOT" --user-dir "$USERDIR" >/tmp/_install_cursor.out 2>&1; then
+  mark_pass "default invocation installs into --user-dir"
+else
+  mark_fail "default invocation installs into --user-dir" "$(cat /tmp/_install_cursor.out)"
+fi
+
+assert_file "$USERDIR/hooks.json" "user hooks.json exists after install"
+assert_file "$TMPROOT/.cursor/rules/apexyard.mdc" "project rules bridge exists after install"
+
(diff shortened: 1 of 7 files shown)


me2resh (commented):
## Code Review: PR #847

**Commit**: `407e1086f744b6e38298cd6ee08cf4a59356ab49`

Independent review by Rex. I formed this verdict from the diff and a live test run, not from any prior review or marker.

### Summary
Fixes the Cursor adapter's install location: Cursor 3.x loads only user-level `~/.cursor/hooks.json`, so this adds `bin/install-cursor-adapter.sh` (new) plus `--user`/`--user-dir`/`--uninstall` on `bin/sync-cursor-adapter.sh` that MERGE apexyard's generated hooks into the user config while preserving pre-existing non-apexyard entries, backing up before every write. Docs + AgDR-0091 are revised to state honestly that Cursor IDE enforcement was observed only via `failClosed` denying on a hook-runner error, not via confirmed delegated bash execution.

### Checklist Results
- Testing:                   Pass (44 sync + 13 install assertions; full suite 96/96)
- Security:                  Pass (no secrets, jq-safe, backups before every write)
- Performance:              Pass (N/A)
- PR Description & Glossary: Pass (narrative summary bullets + glossary present)
- Technical Decisions (AgDR): Pass — extends AgDR-0091 with the GH-840 findings, linked
- Adopter Handbooks:        N/A (no handbooks in repo)

### Verification I ran
- `bin/run-hook-tests.sh` → **96/96 PASS** (matches the PR claim).
- `test_sync_cursor_adapter.sh` → **44/44**, `test_install_cursor_adapter.sh` → **13/13**.
- `shellcheck bin/install-cursor-adapter.sh` → clean. `bin/sync-cursor-adapter.sh` → only the pre-existing SC2016 info note on line 311 (unchanged `check_drift` region) — the PR body's honesty about this is confirmed.

### What I checked closely (this change class)
- **Merge preserves foreign entries**: `install_user_hooks()` keeps every existing entry whose `command` does not match `\.claude/hooks/` and appends the freshly generated apexyard entries per event array. Verified by the "preserves a pre-existing non-apexyard hook entry" and "leaves an untouched, unrelated event array alone" assertions.
- **Backup before every write**: taken (`.bak-<ts>`) for a valid pre-existing file; invalid JSON is backed up to `.bak-<ts>.invalid` and the merge starts fresh rather than corrupting. Good.
- **Surgical `--uninstall`**: removes only apexyard-owned entries (`test("\\.claude/hooks/")`), prunes emptied event arrays, preserves foreign entries, backs up first, and is a clean no-op when nothing is installed (does not fabricate a file). Verified.
- **Idempotency**: drop-then-re-add of owned entries makes re-runs stable; the "no duplicate entries" assertion confirms it.
- **JSON safety**: `jq empty` validity gate on both paths; `// {}` / `// []` defaults; empty/missing input handled.
- **Docs honesty**: docs and AgDR explicitly say IDE enforcement rests on `failClosed`, not confirmed delegated execution ("known-bad commands get blocked," not "the gate logic runs"), and that `cursor-agent` (CLI) is not covered. No overclaiming — this is the load-bearing project value and it holds.
- **Scope**: diff stays entirely within the declared file set. No out-of-scope changes.

### Non-blocking nits (suggestions)
1. **Dead-code `$HOME` guard.** `USER_DIR="${HOME:-}/.cursor"` is never empty, so `[ -z "$USER_DIR" ]` in both scripts can never fire. If `HOME` were genuinely unset, the intended clean error is skipped and the script silently targets `/.cursor`, failing later with a permission error. Consider `USER_DIR="${HOME:+$HOME/.cursor}"` (empty when HOME empty) so the guard actually triggers.
2. **Merge drops foreign top-level keys.** `install_user_hooks`'s jq reconstructs `{version, hooks}`, so any top-level key beyond `version`/`hooks` a user (or a future Cursor version) placed in `~/.cursor/hooks.json` is silently dropped on merge. Low risk today (Cursor's documented schema is just version+hooks) and mitigated by the backup, but `$existing * {hooks: ...}` would preserve unknown keys defensively.
3. **Empty event arrays not pruned on install merge.** Unlike `--uninstall`, the install merge can leave an empty array for an event key that previously held only apexyard entries with no generated replacement. Purely cosmetic.

None of these affect the core guarantees, all of which hold and are well-tested.

### Verdict
**APPROVED**

---
Reviewed commit: `407e1086f744b6e38298cd6ee08cf4a59356ab49`

atlas-apex (approved):
Approving on behalf of the CEO merge decision (given in-session). Independent Rex review passed (96/96 hook tests, no blocking findings); user-level Cursor install + honest failClosed docs verified.

PR #876, opened 2026-07-10
feat(#761): widen tracker_view body support to jira/linear/asana adapters

### LINKED ISSUE #761: [Feature] Widen tracker_view body support to jira/linear/asana adapters
## Problem

The migration gate (`require-migration-ticket.sh` Gate 3) reads the issue **body** to find the linked migration AgDR. As of #755 / PR #760, `tracker_view`'s normalised contract populates `body` only for the **gh** and **glab** adapters. For **jira / linear / asana** adopters the body comes back empty, so Gate 3 always blocks a migration-path edit with "body does not reference a migration AgDR" — even when the ticket genuinely links one.

PR #760 added an explanatory `KIND_NOTE` to that block message (so the framework no longer blames the ticket for its own scoping limit), and the `body` field was deliberately scoped to gh/glab at the time because jira's `.description` is ADF (a JSON object, not a grep-able string) and linear/asana bodies had no consumer. Now that the migration gate is a real consumer, those trackers need body support too.

## Proposed behaviour

Add per-adapter `body` mapping to `_lib-tracker.sh`:

- `_tracker_normalise_jira` → `.fields.description`. Jira Cloud returns **ADF** (a JSON object) — flatten ADF content nodes to plain text before emitting; Jira Server/DC may return a plain string (pass through).
- `_tracker_normalise_linear` → `.description` (markdown string).
- `_tracker_normalise_asana` → `.notes` (plain text) or `.html_notes`.

Then extend the tests (mirror the gh/glab `body` assertions in `test_tracker_aware_hooks.sh`, add a Gate-3 happy-path case per newly-supported kind in `test_require_migration_ticket.sh`), and drop the `KIND_NOTE` for the now-supported kinds.

## Why it helps adopters

jira / linear / asana are supported `tracker.kind` values, but with body unmapped they **cannot use the migration gate at all** — every migration-path edit is blocked regardless of ticket state. Their only current workaround is `tracker.kind=none`, which disables migration verification entirely. Mapping body restores full Gate 2 + Gate 3 discipline for them.

## Scope / out of scope

- **In scope:** body mapping for the three built-in non-gh/glab adapters + tests + dropping the now-unnecessary `KIND_NOTE` for them.
- **Out of scope:** the `custom` adapter (operator supplies `normalise_jq`); the optional `tracker_id_pattern` shape-check hardening of the shared `tracker_view` eval path (a separate concern noted in the PR #760 review — belongs in its own hardening ticket).

## Framework version

v4.2.0 (latest upstream release). This builds on unreleased PR #760 (targeting `dev`), which introduced the `body` contract field and the gh/glab scoping.

## Related

- Follows #755 and PR #760.
- Touches `_lib-tracker.sh` (`_tracker_normalise_{jira,linear,asana}`) and `require-migration-ticket.sh` (Gate 3 `KIND_NOTE`).

## Glossary

| Term | Definition |
|------|------------|
| `tracker_view` | The normalising issue-fetch dispatcher in `_lib-tracker.sh`; emits `{state,title,url,labels,body}` regardless of the underlying tracker CLI. |
| `_tracker_normalise_<kind>` | Per-tracker adapter mapping a CLI's raw JSON into the normalised contract shape. |
| ADF | Atlassian Document Format — Jira Cloud's rich-text body as a JSON object (not plain text), which is why `body` was not mapped for jira in #760. |
| Gate 3 (migration) | The `require-migration-ticket.sh` check that an OPEN, `migration`-labelled ticket's body links a migration AgDR. |

me2resh:
## Summary

- **Mapped `body` in the jira / linear / asana `tracker_view` adapters** so the migration gate (`require-migration-ticket.sh` Gate 3) actually works on those trackers. Gate 3 greps the issue **body** for the linked migration AgDR, but before this only the gh/glab adapters populated `body` — so on jira/linear/asana every migration-path edit false-blocked with a scoping note, and the only escape was `tracker.kind=none` (which disables migration verification entirely). Those adopters now get full Gate 2 + Gate 3 discipline.
- **jira → `.fields.description`, with ADF flattening.** Jira Cloud returns the description as ADF (Atlassian Document Format — a JSON *object*, `{type:"doc",content:[…]}`, not a string), so a naive pass-through would emit an unusable object that no `grep` could match. The adapter recursively collects every `text` leaf from the ADF tree and newline-joins them to grep-able plain text. Jira Server / Data Center returns a plain string, which the adapter passes through verbatim; a missing description degrades to `""`.
- **linear → `.description`** (markdown) and **asana → `.notes`** (falling back to `.html_notes`) — straight field maps, matching the gh/glab shape.
- **Dropped the now-unnecessary `KIND_NOTE`** from Gate 3's block message for gh/glab/jira/linear/asana (all five built-in kinds now populate body). The note is **kept for the `custom` adapter**, whose body still depends on the operator's `normalise_jq` — out of scope here per the ticket.
- **Tests.** Added `body` assertions to the linear + jira lib smokes in `test_tracker_aware_hooks.sh` (jira covers both the ADF-object and plain-string paths), added a new asana lib smoke (none existed), and in `test_require_migration_ticket.sh` replaced the old "jira blocks with scoping note" case with Gate-3 **happy-path** cases per newly-supported kind (jira ADF, jira plain-string, linear, asana) plus a retained "custom still blocks with note" case.

## Testing

```bash
bash .claude/hooks/tests/test_tracker_aware_hooks.sh      # 28 passed, 0 failed
bash .claude/hooks/tests/test_require_migration_ticket.sh # 19 passed, 0 failed
shellcheck .claude/hooks/_lib-tracker.sh                  # clean (rc=0)
```

The adapters are exercised end-to-end through mock `linear` / `jira` / `asana` CLIs (the same mock-CLI harness the existing gh/glab cases use), so the body mapping and ADF flattening are verified without live tracker credentials. No live jira/linear/asana account is required — everything reachable in these hooks is covered by the mock harness; nothing in this change is left unverifiable.

Closes #761

---

## Glossary

| Term | Definition |
|------|------------|
| `tracker_view` | The normalising issue-fetch dispatcher in `_lib-tracker.sh`; emits `{state,title,url,labels,body}` regardless of the underlying tracker CLI (gh/glab/jira/linear/asana/custom). |
| `_tracker_normalise_<kind>` | Per-tracker adapter mapping a CLI's raw JSON into the normalised contract shape. |
| ADF | Atlassian Document Format — Jira Cloud's rich-text body as a JSON object (not plain text). Flattened here to newline-joined `text` leaves so the body stays grep-able. |
| Gate 3 (migration) | The `require-migration-ticket.sh` check that an OPEN, `migration`-labelled ticket's body links a migration AgDR (`docs/agdr/AgDR-\d+-.*migration.*\.md`). |
| `KIND_NOTE` | The explanatory footer Gate 3's block message appends when the active tracker can't populate body — now scoped to the `custom` adapter only. |

### DIFF
--- a/.claude/hooks/_lib-tracker.sh
+++ b/.claude/hooks/_lib-tracker.sh
@@ -17,12 +17,15 @@
 #   tracker_view <id> [<owner_repo>]   dispatches the view command and emits normalised JSON on stdout
 #                                      Exit 0 = ticket exists; non-zero = doesn't, or CLI errored.
 #                                      JSON shape: {"state":..., "title":..., "url":..., "labels":[...], "body":...}
-#                                      `body` is populated for the gh and glab adapters (the kinds
-#                                      that have a consumer needing it — the migration gate reads it
-#                                      to find the linked AgDR, #755). Other adapters omit the body
-#                                      key entirely (consumers read it as `.body // empty`) until one
-#                                      needs it (jira `.description` is ADF, not a grep-able string;
-#                                      linear/asana bodies have no consumer yet).
+#                                      `body` is populated for the gh, glab, jira, linear, and asana
+#                                      adapters — every built-in kind the migration gate (which reads
+#                                      the body to find the linked AgDR, #755/#761) can query. jira's
+#                                      `.fields.description` may be ADF (a JSON object, not a string)
+#                                      on Jira Cloud; the adapter flattens ADF text nodes to plain
+#                                      text so the body stays grep-able (Jira Server/DC returns a
+#                                      plain string, passed through). The `custom` adapter still omits
+#                                      body unless the operator's `normalise_jq` emits one (consumers
+#                                      read it as `.body // empty`).
 #   tracker_create <owner/repo> <title> [<body_file>] [<labels_csv>]
 #                                      creates a ticket via the per-project CLI; emits {ref,url}.
 #   tracker_review_submit <owner/repo> <pr> <verdict> [<body_file>]  (#758)
@@ -358,8 +361,10 @@ _tracker_normalise_glab() {
 #
 # Documented assumption: `linear issue view <ID> --json` emits a JSON object
 # with .state (or .state.name), .title, .url, .labels (array of strings or
-# array of {name} objects). Both shapes are handled — older linear CLI
-# versions returned strings; newer return objects.
+# array of {name} objects), and .description (a markdown string). Both label
+# shapes are handled — older linear CLI versions returned strings; newer return
+# objects. body maps to .description so the migration gate can read the linked
+# AgDR (#761).
 # ------------------------------------------------------------------------------
 _tracker_normalise_linear() {
   local raw="$1"
@@ -369,7 +374,8 @@ _tracker_normalise_linear() {
     state:  ((.state | if type == "object" then .name else . end) // ""),
     title:  (.title // ""),
     url:    (.url // ""),
-    labels: ((.labels // []) | map(if type == "object" then .name else . end))
+    labels: ((.labels // []) | map(if type == "object" then .name else . end)),
+    body:   (.description // "")
   }' 2>/dev/null
 }
 
@@ -377,8 +383,18 @@ _tracker_normalise_linear() {
 # Internal adapter: jira → normalised JSON.
 #
 # Documented assumption: `jira issue view <ID> --raw` emits Jira's REST JSON
-# with .fields.{summary,status.name,labels} and .self for the URL. The
-# `jira` CLI (ankitpokhrel/jira-cli) is the de-facto standard.
+# with .fields.{summary,status.name,labels,description} and .self for the URL.
+# The `jira` CLI (ankitpokhrel/jira-cli) is the de-facto standard.
+#
+# body maps to .fields.description (#761). Jira Cloud returns the description as
+# ADF — Atlassian Document Format, a JSON object ({type:"doc",content:[…]}), not
+# a string — so a naive pass-through would emit an unusable object and the
+# migration gate could never grep an AgDR link out of it. The `if type` branch
+# below flattens ADF: it recursively collects every `text` leaf from the content
+# tree (`[.. | .text? // empty]`) and joins them with newlines, yielding
+# grep-able plain text. Jira Server / Data Center returns description as a plain
+# string, which the string branch passes through verbatim. A missing/null
+# description degrades to "".
 # ------------------------------------------------------------------------------
 _tracker_normalise_jira() {
   local raw="$1"
@@ -388,7 +404,13 @@ _tracker_normalise_jira() {
     state:  ((.fields.status.name // .status // "") | tostring),
     title:  ((.fields.summary // .summary // .title // "") | tostring),
     url:    ((.self // .url // "") | tostring),
-    labels: ((.fields.labels // .labels // []) | map(if type == "object" then .name else . end))
+    labels: ((.fields.labels // .labels // []) | map(if type == "object" then .name else . end)),
+    body:   (
+      (.fields.description // .description // "") as $d |
+      if ($d | type) == "string" then $d
+      elif ($d | type) == "object" then ([$d | .. | .text? // empty] | join("\n"))
+      else "" end
+    )
   }' 2>/dev/null
 }
 
@@ -396,8 +418,10 @@ _tracker_normalise_jira() {
 # Internal adapter: asana → normalised JSON.
 #
 # Documented assumption: `asana task get <gid> --json` emits {data: {name,
-# completed, permalink_url, tags}}. State is derived from .completed
-# (true → "Closed", false → "Open").
+# completed, permalink_url, tags, notes}}. State is derived from .completed
+# (true → "Closed", false → "Open"). body maps to .notes (Asana's plain-text
+# task description), falling back to .html_notes when only the rich-text form is
+# present, so the migration gate can read the linked AgDR (#761).
 # ------------------------------------------------------------------------------
 _tracker_normalise_asana() {
   local raw="$1"
@@ -409,7 +433,8 @@ _tracker_normalise_asana() {
       state:  (if ($t.completed == true) then "Closed" else "Open" end),
       title:  ($t.name // ""),
       url:    ($t.permalink_url // ""),
-      labels: (($t.tags // []) | map(if type == "object" then .name else . end))
+      labels: (($t.tags // []) | map(if type == "object" then .name else . end)),
+      body:   ($t.notes // $t.html_notes // "")
     }
   ' 2>/dev/null
 }
--- a/.claude/hooks/require-migration-ticket.sh
+++ b/.claude/hooks/require-migration-ticket.sh
@@ -304,7 +304,8 @@ fi
 # false-blocked every migration edit even when the GitLab ticket was valid.
 # tracker_view emits the normalised {state,title,url,labels,body} shape
 # regardless of tracker; labels come back as a flat string array and body is
-# populated for gh/glab (the kinds the migration gate reads).
+# populated for gh/glab/jira/linear/asana — every built-in kind the migration
+# gate reads (#761 widened body beyond the original gh/glab of #755).
 if [ -f "$HOOK_DIR/_lib-tracker.sh" ]; then
   # shellcheck source=/dev/null
   . "$HOOK_DIR/_lib-tracker.sh"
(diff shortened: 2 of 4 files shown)


me2resh (commented):
## Code Review: PR #876

**Commit**: `37d08302e7e4f302388398d00abbd996446f5e82`

### Summary
Widens the normalised `body` field of `tracker_view` to the jira, linear, and asana adapters in `_lib-tracker.sh`, so `require-migration-ticket.sh` Gate 3 (which greps the issue body for the linked migration AgDR) works on those trackers instead of false-blocking every migration edit. Previously only gh/glab populated `body`; the only escape for jira/linear/asana adopters was `tracker.kind=none`, which disables migration verification entirely. Closes #761.

### Checklist Results
- ✅ Architecture & Design:      Pass — follows the existing per-adapter `_tracker_normalise_<kind>` contract; no new abstraction
- ✅ Code Quality:               Pass — shellcheck clean on `_lib-tracker.sh`
- ✅ Testing:                    Pass — 28/28 + 19/19, new jira ADF/plain + asana cases
- ✅ Security:                   Pass — jq runs under `2>/dev/null`; adapters fail to `""` not error
- ✅ Performance:                Pass — one extra jq branch per fetch, negligible
- ✅ PR Description & Glossary:  Pass — narrative summary, Testing block, Closes #761, 5-term glossary
- ⚠ Summary Bullet Narrative:   Pass — every bullet answers what + why
- ✅ Technical Decisions (AgDR): N/A — extends the adapter pattern established in #755/#670 (AgDR-0072); no new decision
- ✅ Adopter Handbooks:          N/A — no violations (migration-safety is blocking but only fires on actual migration files; this diff has none)
- ⚠ Fallow Static Analysis:     N/A — not a JS/TS diff

### Verification performed
1. **Adapter mappings** — verified against the gh/glab precedent:
   - `jira → .fields.description`: string branch passes Server/DC plain strings through verbatim; object branch flattens Jira Cloud ADF via `[$d | .. | .text? // empty] | join("\n")`; null/missing and unexpected non-string/non-object types degrade to `""`. The `.text?` (`?`-suppressed) + `.. ` recursive descent + `// empty` combination cannot crash on unusual shapes — confirmed it fails soft to `""`, never errors.
   - `linear → .description`, `asana → .notes` fallback `.html_notes` — straight field maps matching the contract.
2. **Gate impact** — `KIND_NOTE` suppression widened from `gh|glab)` to `gh|glab|jira|linear|asana)`, with the note retained for `custom` (whose body still depends on the operator's `normalise_jq`). Consistent with the new support; the reworded note now points custom adopters at their `normalise_jq`, a built-in tracker, or `kind=none`.
3. **Tests** — ran both suites from the PR-branch worktree with `env -u CLAUDE_CODE_SESSION_ID`:
   - `test_tracker_aware_hooks.sh`: **28 passed, 0 failed** — includes the new jira ADF-object case, jira plain-string (Server/DC) pass-through case, and the new asana lib smoke (none existed before).
   - `test_require_migration_ticket.sh`: **19 passed, 0 failed** — Gate-3 happy-path cases per newly-supported kind (jira ADF 12, jira plain 12b, linear 12c, asana 12d) plus the retained "custom still blocks with note" case (12e), which correctly no longer names jira/linear/asana as unsupported.
   - `shellcheck _lib-tracker.sh`: clean (rc=0). The two SC2295 infos in `require-migration-ticket.sh` are at lines 191/199 — outside this PR's diff hunks (304, 395) — confirmed pre-existing.

### Issues Found
None.

### Suggestions
None blocking. The ADF flattener joins only `text` leaves, dropping ADF `mention`/`inlineCard` node attrs — fine for the grep-an-AgDR-link use case, and the doc comment is honest about scope. No change requested.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `37d08302e7e4f302388398d00abbd996446f5e82`

me2resh (commented):
## Security Review: PR #876 (advisory)

**Commit**: `37d08302e7e4f302388398d00abbd996446f5e82`
**Scope**: trust-chain change — `_lib-tracker.sh` (adds attacker-influenceable issue `body` to the normalised `tracker_view` output for jira/linear/asana, incl. Jira ADF flattening) + `require-migration-ticket.sh` (Gate 3 message + kind list). Reviewed against the three concerns raised: injection surface, gate weakening, fail-closed behaviour.

### Verdict: no security issues found. Safe to merge from a security standpoint.

This is a careful, defence-in-depth change. Every property the review asked me to confirm holds, verified both by reading the code at HEAD and by running adversarial tests.

---

### 1. Injection surface — CLEAN

The attacker-influenceable body **never touches a shell-eval path**, and the jq filters are **static literals** (no interpolation of body content, no `--arg` carrying the body into a filter position):

- The one `eval` in `tracker_view` (line 505) runs `$cmd`, which is the *view-command template* with only `{id}`/`{owner_repo}` substituted — and those two are `printf %q`-quoted in `_tracker_substitute` before substitution. The issue **body is not in `$cmd`**; it arrives later as the CLI's stdout captured into `$raw`.
- `$raw` is passed to each adapter as a positional arg and piped to jq via **stdin** (`printf '%s' "$raw" | jq -c '<single-quoted static filter>'`). The body is pure jq *data*, never part of the filter program. The jira ADF branch (`[$d | .. | .text? // empty] | join("\n")`) operates on parsed JSON values — there is no string-eval of body content anywhere.
- Downstream, the migration gate reads `BODY=$(… | jq -r '.body // empty')` and only ever `echo "$BODY" | grep -qE …`. No `eval`, no command interpolation.

Empirically confirmed: a Jira ADF `text` leaf containing `$(touch /tmp/PWNED); \`id\`; docs/agdr/AgDR-0011-schema-migration.md` flowed through the adapter and the gate **without executing** the metacharacters (no `PWNED` file created). As designed, that body can at worst *satisfy* the AgDR grep — which was already true for gh/glab and is the whole (operator-trust) point of Gate 3. No new privilege, no code execution.

### 2. Gate weakening — none beyond the intended fix

Before this PR, jira/linear/asana returned an empty body, so Gate 3 **always blocked** those trackers (with the scoping note); the only escape was `tracker.kind=none`, which disables migration verification *entirely*. After this PR, Gate 3 passes for those trackers **iff the body actually contains a matching AgDR-migration path** — precisely the intended fix (adopters with a legitimate reference). The grep regex is unchanged.

Two nice defensive properties worth calling out:
- ADF leaves are joined with `\n`, and the regex is `[^[:space:]]*` on a line-oriented grep. So a path *split across two ADF text nodes* is broken by the newline and will **not** fabricate a match — join-with-newline errs toward a false *block*, never a false *pass*.
- `KIND_NOTE` removal is correct: the note now fires only for `custom` (and unknown kinds via `*)`), which still can leave body empty. Test case 12e confirms custom still blocks-with-note, and the note no longer mislabels jira/linear/asana as unsupported.

### 3. Fail-closed behaviour — confirmed

Every malformed-input path degrades body to `""` → grep misses → gate **blocks** (exit 2), never errors-open:

| Input | Result |
|-------|--------|
| Non-JSON CLI output | `jq -e .` validity guard fails → adapter returns 1, empty → gate blocks |
| `description` null / number / array | `else "" end` → body `""` → blocks |
| ADF object with no `text` leaves (`{type:doc,content:[]}`) | `join` of empty → `""` → blocks |
| **Deeply-nested ADF (5000 levels, resource-exhaustion attempt)** | jq hits its own parse-depth limit → validity guard fails → empty → blocks |

The `ISSUE_JSON`-empty branch in the hook is already a hard exit-2 with a fail-closed message, so an unreachable/garbled tracker also blocks.

---

### Verification performed
- Read `_lib-tracker.sh` (adapters + `tracker_view` dispatch + `_tracker_substitute`) and `require-migration-ticket.sh` Gates at HEAD `37d0830`.
- Adversarial tests: metachar-in-ADF (no execution), malformed/null/wrong-type/empty ADF (all → `""`), 5000-level nested ADF (jq depth limit → fail-closed).
- `test_tracker_aware_hooks.sh` → 28/0 · `test_require_migration_ticket.sh` → 19/0 · `shellcheck` both files rc=0.

### Non-blocking observation (pre-existing, not introduced here)
Gate 3 only proves *"the body text mentions an AgDR-migration path"* — it does not verify the referenced AgDR file exists or is meaningful. That's a deliberate operator-trust property that predates this PR (identical for gh/glab); this change merely extends the same weak-by-design check to three more trackers. Flagging only so nobody mistakes Gate 3 for an adversarial control.

---
Advisory review only — no approval marker written, not merging. Halting per instruction.

📌 Reviewed commit: `37d08302e7e4f302388398d00abbd996446f5e82`

atlas-apex (approved):

Issue #872, opened 2026-07-10
[Bug] release-changelog.sh sync-boundary under-counts when the prior release-sync merged late


me2resh:
## Affected
`bin/release-changelog.sh` (release-cut changelog generation, used by `/release`)

## Given / When / Then
- **Given** a release where the previous release's `/release-sync` PR merged **late** to `dev` (its true-merge commit lands near dev's tip, after many newer commits)
- **When** `/release` runs `bin/release-changelog.sh`
- **Then** the `#737` sync-boundary heuristic picks that late sync commit via `git log --max-count=1 --grep=sync…` and uses `SYNC..HEAD_REF` as the range — **dropping every commit that merged before the late sync**

## Repro (real, v5.0.0)
`main..dev` was **329** commits, but the helper emitted a changelog for only **5** ("Patch release — 3 fixes, 2 improvements"). Caught manually during the v5.0.0 cut; the changelog was regenerated from `main..dev` by hand. Left unnoticed it would have shipped a 5-line release note for a 329-commit major release.

## Root cause
`--max-count=1` returns the most recent sync commit by graph position, not the sync that corresponds to `PREV_TAG`. A sync PR merged with `--merge` lands at dev's tip *at merge time*, which can be long after its release, with many unreleased commits before it.

## Fix
When `PREV_TAG` is an **ancestor of `HEAD_REF`** (the normal post-`/release-sync` state — the release squash was folded into dev), use `PREV_TAG..HEAD_REF` directly: it's exactly the unreleased delta, no over-count, and robust against a late sync merge. Only fall to the sync-commit heuristic when `PREV_TAG` is **not** an ancestor of `HEAD_REF` (the #737 case: sync skipped / not yet run). This preserves #737's over-count fix while fixing the under-count.

## Severity
High — silently ships a truncated release note; only caught because a human eyeballed the count.

## Glossary
| Term | Definition |
|------|------------|
| Sync boundary | The `#737` heuristic: the commit after the last `/release-sync` marks the start of the unreleased delta. |
| Late sync merge | A sync PR that merges to dev well after its release, landing near dev's tip and mis-anchoring the boundary. |

me2resh:
## Investigation result — the proposed ancestor-based fix is WRONG; do not ship it

Validated against the real apexyard history (not just the synthetic test repo):

[...]

Issue #737, opened 2026-06-27
[Bug] /release changelog over-counts (tag range vs squash divergence) — use post-sync boundary


atlas-apex:
## Given / When / Then
**Given** the framework's squash-merge release model (`main` holds only per-release *squash* commits; `dev` holds the individual commits),
**When** `/release` (via `bin/release-changelog.sh`) generates the CHANGELOG using the range `PREV_TAG..dev` (the `vX.Y.Z` tag),
**Then** because the tag is a squash commit whose individual children aren't in `dev`'s ancestry, the range surfaces **every individual commit since the dev/main model began** — massively over-counting already-released work.

Evidence: v4.0.0's CHANGELOG claims "**100 features, 58 fixes**"; cutting v4.1.0, the tag-based range reported **102 features / 263 commits** when the real unreleased delta was **1 feature + 3 fixes**. I scoped v4.1.0 by hand to the post-`sync after v4.0.0` boundary.

This is the changelog-side symptom of the squash-divergence tracked in **#403** (which addresses the *conflict* side via `/release-sync`); the *changelog range* is a distinct, still-open symptom.

## Repro
1. After any release + its main→dev sync, run `PREV_TAG=v<last> HEAD_REF=dev bash bin/release-changelog.sh`.
2. Observe the entry lists hundreds of already-released commits, not just the new delta.
3. Correct delta = `git log <last "sync: merge main into dev after vX" commit>..dev`.

## Severity
Medium — every release ships a misleading, bloated CHANGELOG (already-released features re-listed under each new version). Cosmetic to the code but erodes the changelog's value + confuses adopters.

## Suggested fix
`release-changelog.sh` should range from the **last `sync: merge main into dev after vX.Y.Z` merge commit** (the real last-release boundary on dev's line), not the `vX.Y.Z` tag. Resolve it via `git log --grep='sync: merge main into dev after' --merges -1 --format=%H dev` (or the `/release-sync` branch tip). Falls back to the tag only when no sync marker exists.

## Glossary
| Term | Definition |
|------|------------|
| Squash divergence | Squash-merging dev→main means main never contains dev's individual commits → tag-based "since last release" ranges over-count (#403) |
| Post-sync boundary | The `sync: merge main into dev after vX` merge on dev — the reliable marker of the last release on dev's line |

PR #877, opened 2026-07-10
docs(#872): AgDR-0094 — release provenance via Released-From trailer

me2resh:
## Summary

- **Records the #872 decision as AgDR-0094** — the v5.0.0 changelog was nearly generated from 5 of 329 commits because `release-changelog.sh` infers its range topologically, and the investigation proved that inference is unfixable: the naive ancestor fix re-creates the #737 over-count (334 commits), tree-identity matching fails under main/dev drift, and "dev's tip at cut time" simply isn't in the DAG.
- **Decision: record provenance at the source, don't infer it** — `/release` writes a `Released-From: <dev-sha>` git trailer into the release squash commit; the changelog helper prefers the trailer, falling back to today's heuristic for pre-trailer releases. Plus a permanent backstop: `/release` prints `main..dev` count vs changelog count and warns loudly on mismatch (mechanising the manual catch that saved v5.0.0).
- **Doc-only PR** — the AgDR records the decision; implementation stays tracked on #872 as a normal task.

## Testing

1. Doc-only — no behaviour change. `markdownlint` clean on the new file.
2. AgDR-0094 is the next free ID on `dev` (0093 is reserved by the in-flight contributor PR #772 to avoid re-colliding it).

Refs #872

---

## Glossary

| Term | Definition |
|------|------------|
| Git trailer | A `Key: value` line block at the end of a commit message, parseable with `git interpret-trailers` — survives squash-merge because the PR body becomes the squash message. |
| Sync-boundary heuristic (#737) | The current range logic: anchor on the most recent `release-sync` commit — correct when syncs merge promptly, silently under-counts when one merges late. |
| Released-From | The proposed trailer carrying the exact dev SHA a release was cut from, making the changelog range deterministic. |

PR #878, opened 2026-07-10
fix(#872): resolve release-changelog range from a Released-From trailer

me2resh:
## Summary

- **`bin/release-changelog.sh` now prefers a recorded cut point over inferring one** — if the previous release's squash commit (`PREV_TAG`) carries a `Released-From: <dev-sha>` trailer, the range anchors on that exact sha (`TRAILER..HEAD_REF`), immune to both the #737 over-count and the #872 late-sync under-count. Falls back to the existing #737 sync-boundary heuristic when no trailer exists, and falls back safely (never errors) when the trailer sha doesn't resolve to a real commit.
- **`/release` now records the cut point at the moment it's knowable** — it captures `dev`'s SHA right after `git fetch upstream`, then writes it as a `Released-From` trailer in the release PR body's final paragraph (so it survives squash-merge and stays parseable by `git interpret-trailers`), plus a loud warning if `main..dev`'s raw commit count and the changelog's entry count diverge by more than a small tolerance — the same manual eyeball check that caught the v5.0.0 under-count, now automated.
- **Three new test cases** reproduce the exact v5.0.0 regression shape (a late `/release-sync` merge landing near HEAD, with real unreleased work before it) and prove the trailer wins, the trailer-absent path is byte-for-byte unchanged, and a bogus trailer sha degrades safely to the old heuristic instead of crashing.

## Testing

```
$ env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_release_changelog.sh
...
All 51 test(s) passed.

$ shellcheck bin/release-changelog.sh
(clean, exit 0)

$ bash -n bin/release-changelog.sh .claude/hooks/tests/test_release_changelog.sh
(clean, exit 0)
```

All 48 pre-existing cases still pass unchanged; 3 new AgDR-0094 cases added (trailer-wins-over-late-sync / trailer-absent-fallback / bogus-trailer-safe-fallback).

Refs #872

---

## Glossary

| Term | Definition |
|------|------------|
| `Released-From` trailer | A git trailer (`Key: value` line in a commit message's final paragraph) recording the exact `dev` SHA a release was cut from. `git interpret-trailers` / the `%(trailers:...)` pretty-format only recognise it if it's the message's last paragraph — nothing may follow it. |
| Sync-boundary heuristic (#737) | The pre-existing fallback: anchor the changelog range on the most recent `sync: merge main into dev after vX.Y.Z` commit. Correct when that sync merges promptly; silently drops unreleased work when it merges late (#872). |
| Late sync merge | A `/release-sync` PR that merges to `dev` well after its corresponding release, landing near `dev`'s tip and mis-anchoring the sync-boundary heuristic. |
| Count-mismatch guard | A cheap, always-on sanity check comparing `main..dev`'s raw commit count to the generated changelog's entry count, warning loudly on a large gap — defence in depth behind the trailer. |
| Squash merge | GitHub merges all commits on a PR branch into a single commit; the PR body becomes (part of) that commit's message — which is why the trailer's placement in the body matters. |

AgDR: `docs/agdr/AgDR-0094-release-provenance-trailer.md` — the decision this PR implements (already merged; not re-litigated here).

### DIFF
--- a/.claude/hooks/tests/test_release_changelog.sh
+++ b/.claude/hooks/tests/test_release_changelog.sh
@@ -281,6 +281,70 @@ out=$(run_test '
 ')
 contains "feat mentioning the unversioned string still included" "document the sync/main-to-dev-after convention" "$out"
 
+# ── Test: AgDR-0094 (#872) — Released-From trailer wins over the sync heuristic ──
+# The v5.0.0 regression case: the prior release-sync PR merged LATE, landing near
+# HEAD after real unreleased work. With no trailer, the #737 heuristic anchors on
+# that late sync marker and silently drops everything before it. With the trailer
+# present on PREV_TAG's own commit, the range anchors on the RECORDED cut sha
+# instead — deterministic, and immune to where the sync marker happens to land.
+echo "--- AgDR-0094 trailer wins over late sync (v5.0.0 regression) ---"
+out=$(run_test '
+  mc "chore: initial"
+  CUT=$(git rev-parse HEAD)
+  git commit -q --allow-empty -m "chore: release v10.0.0" -m "Released-From: $CUT"
+  git tag v10.0.0
+  mc "feat(#1001): pre-sync unreleased feature one"
+  mc "feat(#1002): pre-sync unreleased feature two"
+  mc "sync: merge main into dev after v10.0.0 release"
+  mc "feat(#1003): post-sync unreleased feature"
+  PREV_TAG="v10.0.0" HEAD_REF="HEAD" VERSION="v10.1.0" DATE="2026-07-10" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains "pre-sync feature one included (the under-count fix)" "pre-sync unreleased feature one" "$out"
+contains "pre-sync feature two included (the under-count fix)" "pre-sync unreleased feature two" "$out"
+contains "post-sync feature still included" "post-sync unreleased feature" "$out"
+not_contains "release commit itself excluded" "chore: release v10.0.0" "$out"
+not_contains "sync commit itself excluded" "merge main into dev after v10.0.0" "$out"
+
+# ── Test: AgDR-0094 — trailer-absent releases keep the old #737 fallback ────────
+# PREV_TAG carries no trailer (a pre-AgDR-0094 release, or any plain tag) — the
+# range must fall back to the existing sync-boundary heuristic unchanged. This
+# is the explicit trailer-absent counterpart to the trailer-present test above;
+# every pre-existing #737 case elsewhere in this file is a trailer-absent case
+# too, so this one exists mainly to name the fallback path directly.
+echo "--- AgDR-0094 trailer-absent falls back to #737 heuristic ---"
+out=$(run_test '
+  mc "chore: initial"
+  git tag v11.0.0
+  mc "feat(#1101): already-released work"
+  mc "sync: merge main into dev after v11.0.0 release"
+  mc "feat(#1102): the real unreleased delta"
+  PREV_TAG="v11.0.0" HEAD_REF="HEAD" VERSION="v11.1.0" DATE="2026-07-10" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains "post-sync delta included (unchanged #737 behaviour)" "the real unreleased delta" "$out"
+not_contains "pre-sync work still excluded (unchanged #737 behaviour)" "already-released work" "$out"
+
+# ── Test: AgDR-0094 — a bogus/unknown trailer sha falls back safely, never errors ──
+# A mangled or stale Released-From trailer (sha doesn't resolve to a commit in
+# this repo) must not crash the script — it should fall back to the #737
+# heuristic exactly as if no trailer existed, and exit 0.
+echo "--- AgDR-0094 bogus trailer sha falls back safely ---"
+out=$(run_test '
+  mc "chore: initial"
+  git commit -q --allow-empty -m "chore: release v12.0.0" -m "Released-From: not-a-real-sha"
+  git tag v12.0.0
+  mc "feat(#1201): unreleased work before late sync"
+  mc "sync: merge main into dev after v12.0.0 release"
+  mc "feat(#1202): unreleased work after late sync"
+  PREV_TAG="v12.0.0" HEAD_REF="HEAD" VERSION="v12.1.0" DATE="2026-07-10" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+  echo "EXIT_CODE=$?"
+')
+contains "script exits cleanly despite the bogus sha" "EXIT_CODE=0" "$out"
+contains "falls back to sync heuristic (post-sync work included)" "unreleased work after late sync" "$out"
+not_contains "falls back to sync heuristic (pre-sync work excluded, same as #737 fallback)" "unreleased work before late sync" "$out"
+
 # ── Summary ──────────────────────────────────────────────────────────────────
 
 echo ""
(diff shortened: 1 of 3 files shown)


me2resh (commented):
## Code Review: PR #878

**Commit**: `5b46e7f0b8d832c7b1b53f8e39aa8599bc49136a`

### Summary
Implements AgDR-0094 (#872): release-changelog range resolution now prefers a recorded `Released-From: <dev-sha>` trailer on `PREV_TAG`'s commit over inferring the cut point from sync-commit position. `/release` writes the trailer into the release PR body's final paragraph and adds a `main..dev`-vs-changelog count-mismatch warning. Three new test cases cover the trailer-wins / trailer-absent / bogus-trailer paths.

### Checklist Results
- Testing:                   Pass — full suite 51/51 (48 pre-existing + 3 new), shellcheck clean, `bash -n` clean
- Security:                  Pass — no secrets, bogus trailer sha is validated before use (cannot inject a garbage range)
- Performance:               Pass (n/a — single `git log` read)
- PR Description & Glossary: Pass — narrative summary, Testing block, `Refs #872`, 5-term glossary, AgDR-0094 referenced
- Technical Decisions (AgDR):Pass — implements already-merged AgDR-0094; back-reference comment present in code
- Adopter Handbooks:         N/A — no adopter handbooks apply to this shell/markdown diff
- Fallow Static Analysis:    N/A — diff is shell + markdown, not JS/TS

### Issues Found
None blocking.

Verified in depth against the four review asks:

1. **Trailer resolution is correct and safe.** `TRAILER_SHA` is read from `PREV_TAG`'s own commit message (`git log -1 ... "$PREV_TAG"`) — the previous release's squash, which correctly defines where THIS release's delta starts. The sha is validated with `git cat-file -e "${TRAILER_SHA}^{commit}"` guarded by a non-empty check before use, so a mangled/stale/unknown trailer falls through to the heuristic rather than erroring or producing an empty/garbage range. `git cat-file -e` emits nothing on stdout, so no range pollution. Precedence trailer → #737 sync heuristic → merge-base/tag matches AgDR-0094 exactly. The AgDR-0094 back-reference comment Rex asked for on #877 is present (line 61), and the demoted #737 heuristic keeps a clear KNOWN-LIMITATION note explaining why the trailer exists.

2. **No regression.** Ran the full `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_release_changelog.sh` → All 51 test(s) passed. The late-sync case genuinely models the v5.0.0 shape: a version-anchored sync marker lands near HEAD with real unreleased features *before* it, the trailer points back to the prior cut point, and the assertions prove the pre-sync features are included (the under-count fix) while the release/sync commits are excluded by classification.

3. **SKILL.md changes are coherent.** The `Released-From: $DEV_SHA` line is the PR body template's final line, alone in its own paragraph after the Glossary, with explicit prose forbidding anything below it — exactly where GitHub's squash message inherits it and `git interpret-trailers` can parse it. The count-mismatch tolerance (5) is sensibly calibrated: wide enough to absorb the handful of release/sync/Merge-branch marker commits excluded by design, tight enough that a v5.0.0-scale truncation blows past it loudly. The guard is `|| true`-guarded and non-blocking (advisory), correct for a defence-in-depth backstop.

4. **Hygiene clean.** shellcheck exit 0, `bash -n` clean on both files, PR body carries the `AgDR-\d+-slug` pattern the AgDR gate needs.

### Suggestions
- `nit` (non-blocking): the count-mismatch `ENTRY_COUNT` via `grep -cE '^- '` also counts the trailing `- Closes #N, ...` summary line, biasing the gap slightly smaller (more conservative — less likely to warn). Harmless given the tolerance, but a `grep -v` on the Closes line would make the count exact if you ever tighten the tolerance.

### Verdict
**APPROVED**

Halting at the CEO merge gate — not merging.

---
📌 Reviewed commit: `5b46e7f0b8d832c7b1b53f8e39aa8599bc49136a`

atlas-apex (commented):
## Code Review: PR #878

**Commit**: `5b46e7f0b8d832c7b1b53f8e39aa8599bc49136a`

### Summary
Fixes the #872 changelog under-count: `bin/release-changelog.sh` now prefers a recorded `Released-From: <dev-sha>` trailer on `PREV_TAG`'s squash commit (`TRAILER..HEAD_REF`, deterministic) over inferring the range from a sync-boundary marker that a late `/release-sync` can mis-anchor. `/release` records the trailer at cut time and adds an always-on count-mismatch warning. Falls back safely to the unchanged #737 heuristic when no/blank/bogus trailer exists.

### Checklist Results
- Architecture & Design:      Pass — clean precedence chain (trailer → #737 sync heuristic → merge-base), each layer independently comment-documented
- Code Quality:               Pass — robust under `set -euo pipefail`; the `|| true` on the extraction pipeline and the `[ -n ... ] &&` short-circuit before `git cat-file -e` are correct
- Testing:                    Pass — three new cases cover trailer-wins, trailer-absent-fallback, bogus-trailer-safe-fallback; all 51 pass locally
- Security:                   Pass — no secrets, no injection surface (SHA is validated via `git cat-file -e` before use in a range)
- Performance:                Pass — one extra `git log -1` per run; negligible
- PR Description & Glossary:   Pass — narrative Summary, full Glossary, AgDR + ticket linked
- Technical Decisions (AgDR): Pass — AgDR-0094 authored and linked (already merged; not re-litigated here)
- Adopter Handbooks:          N/A — no handbooks loaded for this framework-repo shell/doc diff
- Fallow Static Analysis:     N/A — not a JS/TS diff

### Issues Found
None blocking.

Edge cases verified empirically against the PR HEAD:
- **Missing trailer** → empty `TRAILER_SHA` → `[ -n ]` false → falls back to #737 heuristic, exit 0. Confirmed.
- **Empty trailer value** (`Released-From:` with no sha) → empty after `tr -d` → same fallback. Confirmed.
- **Malformed/bogus SHA** → `git cat-file -e` fails inside the `if` condition without tripping `set -e` (`2>/dev/null` + condition context) → fallback, exit 0. Confirmed by the new test and by direct simulation.
- **Multiple `Released-From` trailers** → `tail -n 1` deterministically picks the last, matching git's own trailer semantics. Confirmed.
- **First release** (`PREV_TAG=NONE`) → untouched top branch (`LOG_RANGE=$HEAD_REF`). No regression.
- **No regression** to the #737 over-count fix: the version-anchored sync grep is preserved verbatim inside the fallback branch; the trailer-absent test proves byte-compatible behaviour.

### Suggestions
- `nit:` (non-blocking) — the count-mismatch guard comment in SKILL.md says the entry count is "every `- ` line except the trailing `Closes` summary", but `grep -cE '^- '` actually counts the `- Closes ...` line too (off-by-one over-count of ENTRY_COUNT). Effect is benign: an inflated ENTRY_COUNT makes the `GAP > TOLERANCE` warning slightly *less* likely to fire (more conservative), and the guard is advisory-only. Either exclude the Closes line (`grep -cE '^- ' | ...` minus the closes bullet) or soften the comment to match the code. Not worth blocking.
- `suggestion:` (non-blocking) — the new tests place the trailer on `PREV_TAG`'s own commit and read it directly, which is the right unit-level shape, but the real path routes the trailer through a squash-merge of the PR body. The SKILL.md prose covers that "final-paragraph" requirement thoroughly; consider a future integration-style note or test asserting a body→squash→trailer round-trip. Current coverage is adequate for merge.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `5b46e7f0b8d832c7b1b53f8e39aa8599bc49136a`

atlas-apex (approved):
Approving as the second identity (atlas-apex) per the maintainer's instruction — Rex's independent review (posted below/above as comment) verified the trailer-parsing logic empirically at HEAD 5b46e7f: 51/51 tests pass, all three fallback edge cases degrade safely under set -euo pipefail, no regression to the #737 heuristic.

Issue #871, opened 2026-07-10
[Feature] Conformance CI — credentialed multi-runtime gate-enforcement matrix + live status badge


me2resh:
## User Story

As the framework maintainer, I want a scheduled CI job that proves apexyard's bash gates still block a **real credentialed agent turn** across opencode, pi, and Codex — so the multi-harness governance claim stays continuously verified, and the eventual "harness-agnostic" headline flip is gated on green-continuous, not a one-time manual proof.

## Acceptance Criteria

- [ ] A scheduled GitHub Actions workflow drives a real credentialed agent turn through **opencode, pi, and Codex** headlessly, and asserts the delegated `block-git-add-all.sh` fires its verbatim block.
- [ ] Per-harness secrets + egress are wired via GitHub secrets (no credentials in the workflow file).
- [ ] A **live-pulled status badge** (per-harness green/red) is published for the site/docs to embed, so any drift is externally visible.
- [ ] **Cursor stays documented-manual** (its CLI ignores `hooks.json`; IDE-only `failCloses`) — the badge shows it as such, never "proven".
- [ ] "Green-continuous" is defined (N consecutive scheduled green runs; a flake resets the counter) as the precondition for the harness-agnostic headline flip.

## Spike findings

Promoted from me2resh/apexyard#848. The spike confirmed **3-green-continuous + Cursor-documented-manual** is achievable: opencode, pi, and Codex can each be driven through a real credentialed agent turn headlessly in CI with `block-git-add-all.sh` firing; Cursor cannot (CLI ignores hooks, IDE-only failClose). The finding is **feasibility, not an executed run**.

What is NOT being carried over from the spike branch:
- The spike's embedded workflow is a DRAFT with stubbed CLI-install / model-ID placeholders — this feature resolves them into a runnable workflow.
- The spike never executed a real credentialed CI run — that needs per-harness credentials + egress provisioned in CI, which this feature delivers.

## Out of Scope

- A Cursor conformance proof (parked as an upstream limitation; documented-manual only).
- Expanding beyond the four shipped harnesses (per the expansion-freeze until a paying/activated non-Claude signal exists).

atlas-apex:
**Priority elevation — this is a commercial asset, not engineering hygiene.**

[...]

PR #879, opened 2026-07-12
feat(#871): conformance CI — credentialed multi-runtime gate matrix + live badge

atlas-apex:
## Summary

- **Scheduled, credentialed conformance matrix** (`.github/workflows/conformance.yml`) — drives one real agent turn per harness (opencode, pi, Codex) daily (`0617 UTC`) plus on-demand via `workflow_dispatch`, and asserts the delegated, unmodified `block-git-add-all.sh` fired. This moves the multi-harness governance claim from a one-off manual proof (recorded 2026-07-09, `docs/harnesses/*.md`) to an ongoing, live-checked one, per #871's own framing.
- **Two independent assertions, not one** — `git diff --cached --name-only` empty is the load-bearing side effect (checked first, independent of transcript content); the hook's own live-captured block message appearing in the transcript is the corroborating signal that the *specific* delegated gate fired, not a harness self-block or fail-closed error. `bin/conformance-assert-block-message.sh` derives its comparison text by re-running the real hook each time rather than hardcoding a second copy of its wording, so the assertion can't silently drift from the hook's actual message.
- **Fail-closed credential gating, never skip-green** — a harness job with `CONFORMANCE_ANTHROPIC_API_KEY` / `CONFORMANCE_OPENAI_API_KEY` unprovisioned FAILS loudly with a `::error::` naming the exact secret, instead of being silently skipped. A skipped job reads as a green check in GitHub's UI — exactly the false-confidence failure mode this ticket exists to close. opencode needs no secret (runs on its free `opencode/big-pickle` model).
- **Live-pulled badge, zero new secrets** — `bin/conformance-publish-badge.sh` queries each matrix job's real conclusion via the Actions API and publishes per-harness shields.io endpoint JSON to an orphan `conformance-badge` branch, using only the job's own `GITHUB_TOKEN` (`contents: write`, scoped to that job). shields.io fetches the raw JSON directly on every render, so the badge is genuinely live rather than a cached snapshot.
- **Green-continuous is a real, numeric rule** — a harness's badge only reads `(proven)` after 3 consecutive scheduled green runs; any red run (including a fail-closed missing-secret run) resets that harness's streak counter to 0. This numeric threshold is a display of the metric, not a rebrand decision — the existing tagline/headline flip in `docs/harnesses/README.md` stays a separate, deliberate call this gives a live signal to point to.
- **Cursor stays documented-manual by construction** — it has no matrix job (its CLI ignores `hooks.json` entirely, confirmed live in `docs/harnesses/cursor.md`), so its badge is a static, hand-seeded grey entry the publisher script never computes or overwrites from a result.
- **AgDR-0095** records the four real decisions this PR makes (cadence, fail-closed-vs-skip-green, badge transport, green-continuous threshold), the options considered and rejected for each, and the honest residual risks (unpinned Codex CLI version, best-effort package-name inference, first scheduled run not expected to be green until secrets are provisioned).

## Testing

```bash
bash .claude/hooks/tests/test_conformance_assert_block_message.sh   # 7 assertions, all pass
bash .claude/hooks/tests/test_conformance_publish_badge.sh          # 19 assertions, all pass (hermetic: bare-repo fixture + stubbed gh)
bash bin/run-hook-tests.sh                                          # 99/100 — the 1 failure (test_sync_codex_adapter.sh) is
                                                                      # pre-existing on origin/dev, reproduced identically with
                                                                      # this PR's changes stashed out; unrelated to this change
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/conformance.yml'))"  # YAML parses clean
npx markdownlint-cli2 docs/agdr/AgDR-0095-conformance-ci-badge.md docs/conformance-ci.md README.md docs/harnesses/README.md
                                                                      # 0 errors
```

Manually verified both `bin/install-opencode-adapter.sh --root <repo> --target-dir <scratch>/.opencode/plugins` and the pi equivalent install correctly into an out-of-tree scratch directory (the shape the workflow's "Install X + adapter" steps use, since the harness CLI runs from the scratch repo, not the checkout root).

**Deferred to the first credentialed CI run** (cannot be verified in this build environment — no network egress, no harness CLI credentials): the actual `npm install -g` package names/versions for opencode/pi/Codex CLIs resolve and install correctly; a real opencode/pi/Codex model turn against the scratch repo actually emits `git add -A` when instructed; ShellCheck (not installed locally — CI's `shellcheck.yml` will run it on the two new `bin/*.sh` scripts on this PR).

## AgDR

[AgDR-0095 — Conformance CI, Credentialed Multi-Runtime Gate Matrix + Live Badge](../blob/feature/GH-871-conformance-ci/docs/agdr/AgDR-0095-conformance-ci-badge.md)

## Secrets the operator must provision before the first scheduled run goes green

| Secret | Used by | Where to get it |
|--------|---------|------------------|
| `CONFORMANCE_ANTHROPIC_API_KEY` | pi's credentialed turn | console.anthropic.com → API Keys |
| `CONFORMANCE_OPENAI_API_KEY` | Codex's credentialed turn | platform.openai.com/api-keys |

opencode needs no secret (free model). Full detail: `docs/conformance-ci.md`.

Refs #871

---

## Glossary

| Term | Definition |
|------|------------|
| Conformance proof | A CI/manual check that a real credentialed agent turn had a gated command blocked by the *same unmodified* bash hook, evidenced by the hook's own output + no side effect (nothing staged) — not a mock, not a by-construction test. |
| Fail-closed (this PR) | A missing credential secret makes the matrix job FAIL, not skip — a skipped job would render as a misleadingly green check. |
| Green-continuous | 3 consecutive scheduled green runs for a harness before its badge claims "(proven)"; any red run resets the streak to 0. |
| shields.io endpoint badge | A badge format where shields.io fetches a small JSON file (schema, label, message, color) directly from a URL on every render — no polling/webhook infra needed on the publishing side. |
| Orphan branch | A git branch with no shared commit history with `dev`/`main` (`git checkout --orphan`) — used here so daily badge-update commits never appear in the framework's real history. |
| failClosed (Cursor) | A block caused by Cursor's hook-runner erroring (`MainThreadShellExec not initialized`) rather than the gate logic evaluating and returning exit 2 — not a conformance proof, per the spike's own glossary. |


atlas-apex (commented):
## Code Review: PR #879

**Commit**: `60ec7e76988c6f0715a064d88d8f18caff2e570a`

### Summary

Adds a daily, credentialed conformance CI matrix (`conformance.yml`) that drives one real gated agent turn per harness (opencode / pi / codex), asserts the delegated, unmodified `block-git-add-all.sh` fired, and publishes per-harness shields.io endpoint badges to an orphan `conformance-badge` branch with a 3-green-continuous "(proven)" threshold. Ships two hermetic test suites, an operator doc, and AgDR-0095. This is the framework's commercial trust asset, so I reviewed it deeply and ran both suites locally.

### Checklist Results
- Testing:                   Pass (26/26 assertions run locally — 7 assert + 19 badge)
- PR Description & Glossary: Pass (glossary present, summary bullets narrative + honest about deferred items)
- Technical Decisions (AgDR): Pass — AgDR-0095 covers cadence, fail-closed, badge transport, green-continuous threshold

### Verified live

- **Load-bearing design point confirmed.** `conformance-assert-block-message.sh --print` actually runs the REAL `.claude/hooks/block-git-add-all.sh` against a synthetic `{"tool_name":"Bash","tool_input":{"command":"git add -A"}}` payload, requires exit 2, and derives the `"are forbidden"` anchor from its live stderr — no hardcoded copy of the message. If the hook's wording drifts, `--print`/`--check` track it (and the script fails loudly if the hook stops exiting 2). This is exactly the anti-drift property the PR claims.
- **Both suites pass locally:** `test_conformance_assert_block_message.sh` 7/7, `test_conformance_publish_badge.sh` 19/19 (hermetic — bare-repo fixture + stubbed `gh`, no network).
- **Streak semantics** (green +1 / any non-success resets to 0 / `(proven)` at ≥3), Cursor-static-by-construction, and orphan-branch guarantee all exercised by the badge suite and confirmed.

### Security — all clear

- **No credential leakage.** API keys are passed only via `env:` from `secrets.CONFORMANCE_*` — never on a command line, never echoed. `::error::` messages name the *secret name*, never the value. Badge JSON contains only `label`/`message`/`color`. Push uses the job's own `GITHUB_TOKEN` in an `x-access-token:` clone URL (Actions auto-redacts).
- **Scratch isolation holds.** The gated turn runs with `working-directory` = the `mktemp -d` scratch repo (its own `git init`), so the model's `git add -A` touches only the scratch index; the real checkout at `$GITHUB_WORKSPACE` is a separate git repo it can't reach. The assert step reads the immutable helper from `$GITHUB_WORKSPACE`, not scratch.
- **Least privilege.** Top-level `permissions: contents: read`; only the `badge` job elevates to `contents: write` + `actions: read`. The credentialed `conform` jobs have no write scope. Matches the repo's existing convention.
- **No injection via harness output.** The badge `message` interpolates only the Actions-API `.conclusion` field (a fixed enum: success/failure/cancelled/skipped/unknown), not free-form transcript text. `turn.log` (which could in theory contain harness noise) is used only for the `grep -F "are forbidden"` check and the artifact upload — never interpolated into JSON or a shell command.
- **Fail-closed is real, not skip-green.** A missing `CONFORMANCE_*` secret sets `ready=false` and the next step `exit 1`s the job (red), rather than skipping to a misleading green. opencode is correctly exempt (free model).

### Issues Found

**1. `nit:` (badge-integrity, dispatch path) — single-harness `workflow_dispatch` inflates the OTHER harnesses' green streaks.**
On `workflow_dispatch` with `inputs.harness=pi`, the opencode and codex matrix jobs hit the "Skip non-selected harness" step, set `SELECTED=false`, and skip every subsequent step — but the *job itself* still exits clean, so its GitHub conclusion is `success`. The `badge` job runs on `workflow_dispatch` too (`if: always() && (schedule || workflow_dispatch)`), and `conclusion_for('opencode')` reads that `success` and **increments opencode's streak toward `(proven)`** even though no gated turn ran. Three single-harness dispatches could push the two non-selected harnesses to a fabricated "(proven)" — which is precisely the false-confidence outcome the fail-closed design exists to prevent. The AgDR and docs consistently define the metric as "3 consecutive **scheduled** green runs," so the cleanest fix is to only advance streaks on `schedule` (badge job could pass `github.event_name` to the publisher and skip the streak increment on dispatch), or have the publisher treat a `SELECTED=false`/skipped harness as non-success. Not a merge blocker: the workflow is inert until secrets + the first scheduled run, and the scheduled path (the one the "(proven)" claim rests on) is unaffected — but worth a fast follow-up before anyone leans on the badge for the trust claim.

**2. `nit:` — codex model mismatch between the workflow and the proof it cites.**
The turn step runs `codex exec -m gpt-5.4`, but `docs/harnesses/README.md` (updated in this same PR) records the live proof at `gpt-5.5` ("Model mapping `opus`→`gpt-5.5` is correct"). The conformance run should exercise the same model the manual proof was recorded against; `gpt-5.4` may not resolve or may behave differently. Since the Codex CLI is `@latest` (already flagged as a known gap in the AgDR/doc), fold a `gpt-5.5` correction into the same follow-up that pins the Codex CLI version.

### AC coverage vs #871 (5 ACs)

1. Scheduled matrix, one real gated turn per green-lit harness, verbatim-message assertion — met (assertion derives from the live hook).
2. GitHub-secrets-only credentials — met (`secrets.*` throughout, no embedded creds).
3. Live-pulled per-harness badge — met (shields.io endpoint JSON, raw-fetched on render).
4. Cursor stays documented-manual, never proven — met **by construction** (no matrix job; `cursor.json` seeded once behind `if [ ! -f cursor.json ]`, never written from a conclusion).
5. Explicit numeric green-continuous definition — met (`GREEN_CONTINUOUS_THRESHOLD=3`, reset-on-red).

The PR is **honest** about the one AC that cannot be proven in this build: the actual credentialed run (real CLI installs + a real model emitting `git add -A`) is explicitly deferred to the first scheduled run, both in the PR body and AgDR-0095 § Consequences ("the first scheduled run is not expected to be green"). Package-name inference and the unpinned Codex CLI are disclosed as residual risks rather than hidden. This is the correct posture.

### Verdict

**APPROVED**

The core machinery is sound, secure, and does exactly what it claims — the assertion genuinely delegates to the real hook, credentials are handled cleanly, isolation holds, and both hermetic suites pass. The two findings are `nit`-level (a dispatch-path streak-inflation bug and a codex model-string mismatch) that affect the badge's cosmetic accuracy on the manual path, not the framework's real gate enforcement or the scheduled "(proven)" signal. Recommend a fast follow-up to gate streak increments to `schedule` and correct the codex model to `gpt-5.5`. Not blocking merge.

---
📌 Reviewed commit: `60ec7e76988c6f0715a064d88d8f18caff2e570a`

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #879

**Commit**: `60ec7e76988c6f0715a064d88d8f18caff2e570a`

### Summary
Adds a scheduled, credentialed conformance CI matrix (`.github/workflows/conformance.yml`) that drives one real gated agent turn per harness (opencode/pi/Codex), asserts the delegated unmodified `block-git-add-all.sh` fired, and publishes per-harness shields.io endpoint badges to an orphan `conformance-badge` branch with a 3-green-continuous "(proven)" threshold. Backed by two hermetic test scripts, AgDR-0095, and operator docs.

This is a well-designed, honestly-scoped PR — the fail-closed-not-skip-green choice, the "assert the side effect first, transcript second" design, and deriving the assertion anchor from the live hook (not a hardcoded copy) are all correct calls. The trust-chain posture is clean. But there is one blocking correctness defect that defeats the feature's core purpose.

### Checklist Results
- Testing:                   Pass (2 hermetic suites; bare-repo fixture + stubbed gh)
- Security:                  Pass — see notes below
- Performance:               Pass (daily cadence, 10-min timeouts)
- Technical Decisions (AgDR):Pass — AgDR-0095 covers all four decisions with options/tradeoffs
- Adopter Handbooks:         N/A — diff is YAML/bash/md; no language handbook matches, migration-safety (blocking) doesn't apply (no DB migration)

### Security review (the trust-chain-adjacent concern)
No credential leak and no existing gate is weakened:
- Workflow-level `permissions: contents: read`; `contents: write` + `actions: read` are scoped to the `badge:` job only (least privilege). Good.
- Secrets are only referenced via `${{ secrets.CONFORMANCE_* }}` at step `env:` scope — never echoed. The turn transcript uploaded as an artifact is model output, not credentials.
- The `https://x-access-token:${GH_TOKEN}@github.com/...` clone in `conformance-publish-badge.sh` is the standard token-in-URL pattern and is not logged (`--quiet`, no `set -x`).
- The `conform` job copies `.claude` into a scratch repo and exercises the **real, unmodified** hook — it neither edits hooks nor bypasses `block-unreviewed-merge.sh` / any merge gate. The badge publisher writes only to the orphan `conformance-badge` branch, never to `dev`/`main`.
- Fail-closed (missing secret => job FAILS, not skips) is the right posture and is implemented correctly (`steps.creds` gate + explicit `exit 1`).

### Issues Found

**BLOCKING — invalid `actions/upload-artifact` pin (39-char SHA) breaks every matrix job**
`.github/workflows/conformance.yml`, "Upload turn transcript" step:
```
uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa9 # v4.6.2
```
That ref is **39 characters**, not a valid 40-char commit SHA. GitHub Actions does not accept abbreviated SHAs for `uses:` — it will fail with "unable to resolve action … unable to find version". Because this step has no `continue-on-error`, the failure fails the whole job. It runs for opencode with `steps.creds.outputs.ready == 'true'`, so even the free, always-on **opencode canary** — the harness that should be reliably green — dies at the upload step and reports `failure`, which resets its streak. Net effect: no harness badge can ever reach green/"(proven)", which is the exact opposite of what this PR ships.

The correct pinned commit for `actions/upload-artifact@v4.6.2` is:
```
uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```
(verified by dereferencing the `v4.6.2` tag on `actions/upload-artifact`). The other two action pins — `actions/checkout@9c091bb…` and `actions/setup-node@2028fbc5…` — are correct 40-char SHAs.

Note: PR CI being green does not cover this, since the workflow only runs on `schedule` / `workflow_dispatch` — the broken step is never exercised by the PR checks.

### Suggestions

**nit / correctness (advisory): `workflow_dispatch` inflates streaks off no-op jobs.**
The `badge:` job runs on `if: always() && (schedule || workflow_dispatch)` and `conformance-publish-badge.sh` increments each harness's streak from its job `conclusion`. On a single-harness `workflow_dispatch`, the non-selected harness jobs set `SELECTED=false`, skip every subsequent step, and the job still concludes `success` — so the publisher bumps their streaks by +1 for a turn that never ran. This also conflicts with the documented definition ("3 consecutive **scheduled** green runs"). Consider one of: gate the streak increment to `github.event_name == 'schedule'`, or have the publisher treat a job that produced no real turn (e.g. no uploaded transcript / a sentinel output) as non-success. Low severity, but it affects the integrity of the "(proven)" metric the feature exists to make trustworthy.

**note (not blocking):** Codex CLI pinned at `@latest` and the best-effort npm package-name inference are already disclosed as residual risks in AgDR-0095 and `docs/conformance-ci.md` — fine to ship as flagged, since a wrong package fails loudly per the fail-closed design. Worth a follow-up to pin Codex once a tested version is recorded.

### Verdict
**CHANGES REQUESTED**

The design and security posture are sound; the single blocker is the malformed `upload-artifact` SHA, which will make every scheduled run red (including opencode) and prevent any badge from ever going green. Fix the pin to `ea165f8d65b6e75b540449e92b4886f43607fa02` and this is an approve. Please also consider the `workflow_dispatch` streak-inflation note before the first real run.

---
Reviewed commit: `60ec7e76988c6f0715a064d88d8f18caff2e570a`

atlas-apex (commented):
## Code Review: PR #879 (re-review)

**Commit**: \`5729226909af7c16c8963b0ba84244d0c2963a0f\`

### Summary
Re-review after the single fix commit addressing the sole blocker from my prior review at \`60ec7e7\`. This PR adds a credentialed multi-runtime conformance CI gate matrix with a live badge (AgDR-0095).

### What changed since the last review
The one blocker I raised was that \`.github/workflows/conformance.yml\` pinned \`actions/upload-artifact\` to an invalid 39-char SHA (\`...43607fa9\`), which would have failed every job that reached the artifact-upload step. Verified the fix:

- \`60ec7e7...5729226\` is **exactly one commit, one file, one line** — 1 addition / 1 deletion in \`conformance.yml\`. Nothing else touched.
- The pin is now the full **40-char** \`ea165f8d65b6e75b540449e92b4886f43607fa02\`.
- That SHA dereferences exactly from the \`actions/upload-artifact\` **v4.6.2** tag (\`git/refs/tags/v4.6.2\` → \`ea165f8d65b6e75b540449e92b4886f43607fa02\`, type commit). The trailing comment \`# v4.6.2\` is accurate.

### Checklist Results
- Security:                  Pass (no credential leak; fail-closed gating preserved)
- Technical Decisions (AgDR): Pass (AgDR-0095 linked)

### Issues Found
None. The single blocker from the prior review is resolved; everything else was already cleared.

### Verdict
**APPROVED**

---
📌 Reviewed commit: \`5729226909af7c16c8963b0ba84244d0c2963a0f\`

PR #884, opened 2026-07-12
fix(#882): resolve /start-ticket's ops root correctly on split-portfolio v2

### LINKED ISSUE #882: [Bug] /start-ticket ops-root walk-up misses split-portfolio v2 (.apexyard-fork anchor) — marker unwritable
## Given / When / Then
- **Given** a split-portfolio v2 adopter (ops fork anchored by `.apexyard-fork`; `onboarding.yaml` + `apexyard.projects.yaml` live in the sibling portfolio repo)
- **When** `/start-ticket` resolves the ops root (SKILL.md step 4a walks up looking for a directory containing BOTH `onboarding.yaml` and `apexyard.projects.yaml`)
- **Then** resolution returns empty — those files aren't in the fork in v2 — and the skill can't write the session marker (a second operator hit this on a fresh machine and had to write the marker by hand).

## Root cause
The skill's walk-up predates the split-portfolio v2 anchor. The hooks already resolve correctly (`_lib-ops-root.sh`: pin-first, then `.apexyard-fork` OR the v1 file-pair); the skill's inline bash still uses the v1 pair only.

## Fix
Update `/start-ticket` step 4a (and sweep other SKILL.md files with the same inline walk-up) to match `_lib-ops-root.sh` semantics: accept `.apexyard-fork` as an anchor, pin-first where a session pin exists. Ideally source the lib instead of duplicating the walk.

## Acceptance Criteria
- [ ] `/start-ticket` resolves the ops root on split-portfolio v2 (`.apexyard-fork` anchor) and v1 (file pair) alike.
- [ ] Audit of remaining SKILL.md inline walk-ups for the same stale predicate.

## Glossary
| Term | Definition |
|------|------------|
| `.apexyard-fork` | The split-portfolio v2 marker file anchoring the ops fork root when onboarding/registry files moved to the sibling repo. |

me2resh:
## Summary

- **`/start-ticket` step 4a now resolves the ops root on split-portfolio v2** — the original inline walk-up only checked for the legacy v1 pair (`onboarding.yaml` + `apexyard.projects.yaml`), both of which live in the sibling portfolio repo on v2 forks, not the fork itself. That made the walk return empty on every v2 fork, so the session marker couldn't be written — a real operator hit this on a fresh machine and had to write the marker by hand (see #882). The fix sources `_lib-ops-root.sh::resolve_ops_root`, the shared resolver every hook already trusts, instead of re-deriving the walk inline.
- **The lib is now *located* via a sibling walk-up, not `git rev-parse --show-toplevel`** — Rex's review caught that the first pass regressed the same class of bug one level up: from a managed-project `workspace/<project>/` clone, `--show-toplevel` resolves to the *project* repo, and those clones carry no `.claude/hooks/` directory at all, so the source silently failed and `resolve_ops_root` was never defined. Step 4a now walks up from `$PWD` until it finds a directory containing `_lib-ops-root.sh`, then sources it — the same pattern `bug`/`feature`/`task`/`migration`/`spike`/`prototype` (13 skills total) already use for `_lib-tracker.sh`. The walk-up only picks a directory to source *from*; `resolve_ops_root` (pin-first, apexyard#381) still decides the real ops root afterward — kept as two distinct steps, since a session pin can point elsewhere than the nearest fork-shaped directory on the walk.
- **Step 4b's registry lookup was hardcoded to `$ops_root/apexyard.projects.yaml`, which is also wrong on v2** — the registry lives in the sibling repo, not `$ops_root`. Switched to `portfolio_registry` from `_lib-portfolio-paths.sh`, sourced directly from `$ops_root` (already resolved and guaranteed to carry `.claude/hooks/` by the time step 4b runs — no second walk-up needed).
- **Swept every other `SKILL.md` for the same stale predicate**, per the ticket's second acceptance criterion. Result: `start-ticket` step 4a was the only genuinely stale one. The other 15 skills that anchor on the ops root already `OR` in `.apexyard-fork` correctly. `mutation-test`'s refuse-check already handles both. `update`/`split-portfolio` do sibling-repo migration comparisons, a different concern.
- **Fixed a doc-accuracy gap in `status/SKILL.md`** found during the sweep — the prose describing the `bin/apexyard` CLI shim's walk-up only mentioned the legacy pair; the shim script itself was already correct.

## Testing

No runnable test harness exists for `SKILL.md` prose+bash — validated by extracting the edited bash blocks and running them against layouts in temp dirs, `APEXYARD_OPS_DISABLE_PIN=1` to isolate the walk-up logic from this session's own live ops-root pin. Extended per Rex's review to add the workspace-clone case (row 3 below):

| cwd | `.claude/hooks/` present? | Step 4a result | Step 4b result |
|-----|---------------------------|-----------------|------------------|
| v1 ops fork root / subdir | yes | resolves to fork root | registry resolves |
| v2 ops fork root / subdir (`.apexyard-fork` only) | yes | resolves to fork root | registry resolves |
| **v2 managed-project `workspace/<project>/` clone (root / subdir) — NEW** | **no** | **resolves to the ops fork above it, via the walk-up** | **registry resolves (sourced from `$ops_root`, not the clone)** |
| outside any apexyard fork | n/a | fails closed: `Not inside an apexyard fork (...)`, no crash | n/a |

Reproduced Rex's regression finding directly, for contrast — the *previous* commit's sourcing form (`git rev-parse --show-toplevel`), run from the same simulated workspace clone:

```
git rev-parse --show-toplevel -> .../gh882-v2/workspace/sample-proj
REGRESSION REPRODUCED: source ".../gh882-v2/workspace/sample-proj/.claude/hooks/_lib-ops-root.sh"
would SOURCE-FAIL (file does not exist there)
```

...versus the new sibling-walk-up form, from the identical cwd:

```
v2 workspace clone: 4a ops_root=.../gh882-v2
v2 workspace clone: 4b registry=.../gh882-v2/apexyard.projects.yaml
v2 workspace clone subdir: 4a ops_root=.../gh882-v2
v2 workspace clone subdir: 4b registry=.../gh882-v2/apexyard.projects.yaml
```

Also re-ran the existing regression suite for the underlying lib (unmodified by this PR — only `SKILL.md` docs changed):

```
$ APEXYARD_OPS_DISABLE_PIN=1 bash .claude/hooks/tests/test_ops_root.sh
Results: 8 passed, 0 failed

$ bash .claude/hooks/tests/test_resolve_ops_root_pin.sh
Results: 8 passed, 0 failed
```

(`test_ops_root.sh` needs `APEXYARD_OPS_DISABLE_PIN=1` in this environment specifically because this session's own launch-cwd pin — pointing at the real ops fork — otherwise overrides the test's synthetic fixtures. Unrelated to this change; `_lib-ops-root.sh` itself is untouched by either commit on this branch.)

Closes #882

## Glossary

| Term | Definition |
|------|------------|
| Ops root | The apexyard fork root — where `.claude/`, `roles/`, `workflows/`, etc. live. Session markers (active ticket, review approvals) are always written here, never inside a managed-project clone. |
| Split-portfolio v2 | A fork layout (framework ≥ #242) where `onboarding.yaml` and `apexyard.projects.yaml` live in a private sibling repo instead of the public fork, anchored by a presence-only `.apexyard-fork` marker file. |
| `.apexyard-fork` | The v2 ops-fork anchor — a marker file whose *presence* (not content) identifies the fork root, written by `/setup` and `/update`. |
| `resolve_ops_root` | The shared resolver in `_lib-ops-root.sh` — pin-first (checks a session-scoped pin file before walking), then falls back to a directory walk-up checking both the v2 marker and the legacy v1 pair. Decides the *real* ops root; distinct from the sibling walk-up that merely locates the lib file to source. |
| Sibling walk-up (to locate a lib) | Walking up from `$PWD` until a directory containing a specific `.claude/hooks/<lib>.sh` file is found, then sourcing it from that path — the established pattern 13+ ticket skills use for `_lib-tracker.sh`, because managed-project clones carry no `.claude/hooks/` tree and `git rev-parse --show-toplevel` would otherwise resolve to the wrong repo. |
| `portfolio_registry` | A resolver in `_lib-portfolio-paths.sh` that returns the absolute path to `apexyard.projects.yaml`, honouring a `.portfolio.registry` override in `project-config.json` for split-portfolio adopters. |

### DIFF
--- a/.claude/skills/start-ticket/SKILL.md
+++ b/.claude/skills/start-ticket/SKILL.md
@@ -86,29 +86,26 @@ Per apexyard#41, the marker path depends on whether the ticket's tracker repo ma
 
 #### 4a. Locate the ops root
 
-The ops root is the apexyard fork root — the directory containing BOTH `onboarding.yaml` and `apexyard.projects.yaml`. Walk up from CWD / the nearest git toplevel until you find it:
+The ops root is the apexyard fork root, anchored by EITHER the `.apexyard-fork` marker (split-portfolio v2, framework ≥ #242 — `onboarding.yaml` and `apexyard.projects.yaml` live in the sibling portfolio repo, not the fork) OR the legacy v1 pair (`onboarding.yaml` AND `apexyard.projects.yaml` both present in the same directory). Resolve it via the shared lib instead of re-deriving the walk inline — `_lib-ops-root.sh` ships in every managed project's `.claude/hooks/` tree the same way `_lib-tracker.sh` does (that's why sourcing it via `git rev-parse --show-toplevel` below is safe regardless of whether cwd is the ops fork or a `workspace/<project>/` clone), and it resolves pin-first (apexyard#381) so a session pin set at launch beats a fresh walk when Claude Code is running from an ops-fork-shaped clone elsewhere on disk (e.g. a `/tmp` build checkout):
 
 ```bash
-ops_root=""
-r=$(git rev-parse --show-toplevel 2>/dev/null || pwd)
-while [ -n "$r" ] && [ "$r" != "/" ]; do
-  if [ -f "$r/onboarding.yaml" ] && [ -f "$r/apexyard.projects.yaml" ]; then
-    ops_root="$r"
-    break
-  fi
-  r=$(dirname "$r")
-done
+source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-ops-root.sh"
+ops_root=$(resolve_ops_root)
 ```
 
-If not found (user is outside an apexyard fork), tell the user and stop. Starting a ticket without the fork doesn't make sense.
+If `$ops_root` is empty (user is outside an apexyard fork — neither anchor found walking up), tell the user and stop. Starting a ticket without the fork doesn't make sense.
 
 #### 4b. Look the tracker repo up in the registry
 
-Given the ticket's `owner/repo` (from step 1), grep `apexyard.projects.yaml` for a project whose `repo:` field matches. One registry-safe way (uses `yq` when available, falls back to a greppy read):
+Given the ticket's `owner/repo` (from step 1), resolve the registry path via `portfolio_registry` (see "Path resolution" above — do NOT hardcode `$ops_root/apexyard.projects.yaml`; in split-portfolio v2 the registry lives in the sibling repo, not the ops fork) and grep it for a project whose `repo:` field matches. One registry-safe way (uses `yq` when available, falls back to a greppy read):
 
 ```bash
+source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-read-config.sh"
+source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-portfolio-paths.sh"
+registry=$(portfolio_registry)
+
 if command -v yq >/dev/null 2>&1; then
-  project=$(yq eval ".projects[] | select(.repo == \"${OWNER_REPO}\") | .name" "$ops_root/apexyard.projects.yaml")
+  project=$(yq eval ".projects[] | select(.repo == \"${OWNER_REPO}\") | .name" "$registry")
 else
   # Greppy fallback: find the `name:` whose sibling `repo:` matches.
   # Strips surrounding quotes from both `name:` and `repo:` values so the
@@ -118,7 +115,7 @@ else
     function unquote(s) { gsub(/^["\x27]|["\x27]$/, "", s); return s }
     /^[[:space:]]*- name:/ { name = unquote($3) }
     /^[[:space:]]*repo:/   { if (unquote($2) == r) { print name; exit } }
-  ' "$ops_root/apexyard.projects.yaml")
+  ' "$registry")
 fi
 ```
 
--- a/.claude/skills/start-ticket/SKILL.md
+++ b/.claude/skills/start-ticket/SKILL.md
@@ -86,22 +86,34 @@ Per apexyard#41, the marker path depends on whether the ticket's tracker repo ma
 
 #### 4a. Locate the ops root
 
-The ops root is the apexyard fork root, anchored by EITHER the `.apexyard-fork` marker (split-portfolio v2, framework ≥ #242 — `onboarding.yaml` and `apexyard.projects.yaml` live in the sibling portfolio repo, not the fork) OR the legacy v1 pair (`onboarding.yaml` AND `apexyard.projects.yaml` both present in the same directory). Resolve it via the shared lib instead of re-deriving the walk inline — `_lib-ops-root.sh` ships in every managed project's `.claude/hooks/` tree the same way `_lib-tracker.sh` does (that's why sourcing it via `git rev-parse --show-toplevel` below is safe regardless of whether cwd is the ops fork or a `workspace/<project>/` clone), and it resolves pin-first (apexyard#381) so a session pin set at launch beats a fresh walk when Claude Code is running from an ops-fork-shaped clone elsewhere on disk (e.g. a `/tmp` build checkout):
+The ops root is the apexyard fork root, anchored by EITHER the `.apexyard-fork` marker (split-portfolio v2, framework ≥ #242 — `onboarding.yaml` and `apexyard.projects.yaml` live in the sibling portfolio repo, not the fork) OR the legacy v1 pair (`onboarding.yaml` AND `apexyard.projects.yaml` both present in the same directory).
+
+Locate `_lib-ops-root.sh` by walking up from `$PWD` — **not** via `git rev-parse --show-toplevel`. Inside a `workspace/<project>/` clone, `--show-toplevel` resolves to the *project* repo, and managed-project clones carry **no `.claude/hooks/` directory at all** (there is no framework mechanism that installs one there) — sourcing from that path silently fails, `resolve_ops_root` is never defined, and this step dead-ends exactly where split-portfolio v2 operators most often run `/start-ticket`. This is the same sibling walk-up pattern `bug`, `feature`, `task`, `migration`, `spike`, `prototype`, and 8 other ticket skills already use to locate `_lib-tracker.sh` — walk up until a directory containing the lib is found, then source it.
+
+Keep two steps distinct: the walk-up below only finds a directory to **source the lib from** (the nearest fork-shaped root above cwd); `resolve_ops_root` then **decides the real ops root**, pin-first (apexyard#381) — the two can differ, e.g. a session pin can point at a different real ops fork than the nearest fork-shaped directory on the walk (say, cwd is inside an ops-fork-shaped `/tmp` build clone).
 
 ```bash
-source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-ops-root.sh"
+ops_lib="$(r="$PWD"; while [ -n "$r" ] && [ "$r" != / ]; do \
+  [ -f "$r/.claude/hooks/_lib-ops-root.sh" ] && { echo "$r/.claude/hooks/_lib-ops-root.sh"; break; }; \
+  r="${r%/*}"; done)"
+if [ -z "$ops_lib" ]; then
+  echo "Not inside an apexyard fork (no .claude/hooks/_lib-ops-root.sh found walking up from $PWD)." >&2
+  exit 1
+fi
+# shellcheck source=/dev/null
+. "$ops_lib"
 ops_root=$(resolve_ops_root)
 ```
 
-If `$ops_root` is empty (user is outside an apexyard fork — neither anchor found walking up), tell the user and stop. Starting a ticket without the fork doesn't make sense.
+If `$ops_root` is still empty after sourcing (no pin, and `resolve_ops_root`'s own internal walk also found no anchor above cwd), tell the user and stop. Starting a ticket without the fork doesn't make sense.
 
 #### 4b. Look the tracker repo up in the registry
 
-Given the ticket's `owner/repo` (from step 1), resolve the registry path via `portfolio_registry` (see "Path resolution" above — do NOT hardcode `$ops_root/apexyard.projects.yaml`; in split-portfolio v2 the registry lives in the sibling repo, not the ops fork) and grep it for a project whose `repo:` field matches. One registry-safe way (uses `yq` when available, falls back to a greppy read):
+Given the ticket's `owner/repo` (from step 1), resolve the registry path via `portfolio_registry` (see "Path resolution" above — do NOT hardcode `$ops_root/apexyard.projects.yaml`; in split-portfolio v2 the registry lives in the sibling repo, not the ops fork) and grep it for a project whose `repo:` field matches. `$ops_root` is already resolved and guaranteed to carry a `.claude/hooks/` tree (step 4a only succeeds when it does), so source directly from it — no second walk-up needed. One registry-safe way (uses `yq` when available, falls back to a greppy read):
 
 ```bash
-source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-read-config.sh"
-source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-portfolio-paths.sh"
+source "$ops_root/.claude/hooks/_lib-read-config.sh"
+source "$ops_root/.claude/hooks/_lib-portfolio-paths.sh"
 registry=$(portfolio_registry)
 
 if command -v yq >/dev/null 2>&1; then
--- a/.claude/skills/status/SKILL.md
+++ b/.claude/skills/status/SKILL.md
@@ -252,7 +252,7 @@ ln -s "$(pwd)/bin/apexyard" ~/.local/bin/apexyard
 apexyard status
 ```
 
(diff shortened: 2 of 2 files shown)


me2resh (commented):
## Code Review: PR #884

**Commit**: `9f7517e96bdff9622e3830ef52d9e6bf56feae37`

### Summary
Fixes `/start-ticket`'s ops-root resolution for split-portfolio v2 (Closes #882). Two SKILL.md docs: `start-ticket` step 4a swaps a v1-only inline walk-up for `resolve_ops_root` from `_lib-ops-root.sh`; step 4b swaps a hardcoded `$ops_root/apexyard.projects.yaml` for `portfolio_registry`; `status/SKILL.md` gets a one-line prose accuracy fix. The core resolution *logic* is right — but the mechanism used to load that logic has a defect that reintroduces a resolution failure one level up, which is the exact thing this PR set out to fix.

### Checklist Results
- Architecture & Design:      Changes requested (sourcing mechanism)
- Testing:                    Pass (logic) / Gap (sourcing path untested)
- Security:                   Pass (docs-only)
- PR Description & Glossary:  Pass — but contains a factually incorrect claim (see finding 2)
- Summary Bullet Narrative:   Pass (exemplary what+why bullets)
- Technical Decisions (AgDR): N/A (docs fix, no new decision)
- Adopter Handbooks:          N/A (no handbooks in tree)

### What I verified (re-ran the agent's approach in temp fixtures)

**Old predicate reproduces #882:**
```
v1 cwd=sub/deeper -> <resolved>   (correct)
v2 cwd=sub/deeper -> <EMPTY>      (BUG — matches #882)
```

**New code, cwd = ops fork (the primary scenario) — FIX WORKS:**
```
v1 fork root / subdir -> resolved correctly
v2 fork root / subdir -> resolved correctly   <- #882 fixed
```

**New code, cwd = a managed-project workspace clone — REGRESSION:**
```
workspace/<proj> (no .claude/hooks) -> SOURCE-FAILS (no lib at <clone>/.claude/hooks/_lib-ops-root.sh)
```
`source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-ops-root.sh"` resolves the toplevel to the *project clone*, which does not carry `_lib-ops-root.sh`. The source fails, `resolve_ops_root` is never defined, `ops_root` comes back empty, and step 4a then tells the operator they're "outside an apexyard fork" and stops — in exactly the split-portfolio-v2 population this PR targets.

### Issues Found

**⛔ 1 (blocking) — Sourcing via `git rev-parse --show-toplevel` fails from a workspace clone (concern the ticket itself flagged).**
`.claude/skills/start-ticket/SKILL.md` steps 4a and 4b both do:
```bash
source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-ops-root.sh"   # 4a
source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-portfolio-paths.sh"  # 4b
```
When the session's cwd is a managed-project clone under `workspace/<project>/`, `git rev-parse --show-toplevel` returns the *project* repo. I checked all 60+ workspace clones on this machine — **none has a `.claude/hooks/` directory at all**, so the source silently fails and the resolver is never loaded. This is the same class of bug (#882) reintroduced one level up: the resolution *logic* is now v2-aware, but you can't reach it from a workspace clone. The OLD inline walk (which sourced nothing and climbed the dir tree directly) actually handled the v1 workspace-clone case; the new code does not.

Fix: locate the lib with a walk-up, then source it, then call `resolve_ops_root` (keeps the pin-first benefit). This is exactly the pattern 13 sibling ticket skills already use for `_lib-tracker.sh` (bug, feature, migration, task, spike, prototype, ...):
```bash
lib="$(r="$PWD"; while [ -n "$r" ] && [ "$r" != / ]; do \
  [ -f "$r/.claude/hooks/_lib-ops-root.sh" ] && { echo "$r/.claude/hooks/_lib-ops-root.sh"; break; }; \
  r="${r%/*}"; done)"
. "$lib"; ops_root=$(resolve_ops_root)
```
Verified: the walk-up climbs out of the clone and finds the lib in the ops fork. (Even simpler alternative: adopt the self-contained inline both-anchor walk that `task`/`feature` use — no lib dependency at all — though you'd lose pin-first.)

**⛔ 2 (blocking) — The prose justifying the approach is factually wrong.**
Step 4a states `_lib-ops-root.sh` "ships in every managed project's `.claude/hooks/` tree the same way `_lib-tracker.sh` does (that's why sourcing it via `git rev-parse --show-toplevel` below is safe ...)". Managed project clones carry **no `.claude/hooks/` directory** — that's precisely *why* the 13 sibling skills walk up to reach the lib in the ops fork. This claim is the load-bearing justification for the whole design choice; correcting it removes the rationale for the `git rev-parse` form and points at the walk-up. The PR body repeats the same claim.

### Suggestions
- The Testing matrix validated `resolve_ops_root`'s *logic* with the lib present, but didn't exercise the *sourcing path from a workspace clone* — which is where it breaks. Add that row.
- `status/SKILL.md` prose fix, sweep verification (task/feature/approve-merge OR in `.apexyard-fork`), and step-4b `portfolio_registry` switch are all correct and good.

### Verdict
**CHANGES REQUESTED**

The v1/v2 resolution logic is correct and the reported bug is genuinely fixed for the common cwd=ops-fork case. But the fix loads the resolver via `git rev-parse --show-toplevel`, which fails from a managed-project workspace clone (verified), and the prose justifies that choice with a false claim about lib presence. Both are small, well-precedented edits — switch the two source lines to the sibling walk-up pattern and correct the prose.

---
Reviewed commit: `9f7517e96bdff9622e3830ef52d9e6bf56feae37`

me2resh (commented):
## Code Review (re-review): PR #884

**Commit**: `10913840adbae5e1327ad8e6e4fca9f837927cac`

Re-review of the new HEAD after CHANGES REQUESTED. Both prior blockers are closed. Verdict flips to **APPROVED**.

### Summary
`/start-ticket` step 4a now locates `_lib-ops-root.sh` via a sibling walk-up from `$PWD` (the same pattern 13 ticket skills use for `_lib-tracker.sh`) instead of `git rev-parse --show-toplevel`, then sources it and calls `resolve_ops_root`. Step 4b sources the portfolio libs from the already-resolved `$ops_root` and resolves the registry via `portfolio_registry`. The false lib-presence prose is gone.

### Blocker 1 — sourcing fails from a workspace clone → CLOSED
Extracted the new 4a/4b blocks verbatim and ran them from a simulated split-portfolio-v2 layout (`.apexyard-fork` fork + sibling portfolio repo + a `workspace/<project>/` clone carrying its own `.git` but **no** `.claude/hooks/`), with `APEXYARD_OPS_DISABLE_PIN=1` to isolate the walk-up:

| cwd | step 4a `ops_root` | step 4b registry |
|-----|--------------------|------------------|
| v1 fork root / subdir | fork root | resolves |
| v2 fork root / subdir (`.apexyard-fork` only) | fork root | resolves |
| **v2 workspace clone root / subdir (no `.claude/hooks/`)** | **fork above the clone** | **resolves (sourced from `$ops_root`)** |
| outside any fork | fails closed: error + `exit 1` | n/a |

The walk-up climbs out of the clone (whose `git rev-parse --show-toplevel` is the clone itself) and finds the lib in the fork above. Reproduced the *previous* commit's `git rev-parse` form from the identical cwd for contrast — it would source `<clone>/.claude/hooks/_lib-ops-root.sh`, which does not exist, silently failing exactly as the prior review found. Step 4b, run under `/bin/bash` (the real skill runtime), resolves `portfolio_registry` to the sibling-repo registry when `.portfolio.registry` is configured — the actual v2 payoff of dropping the hardcoded `$ops_root/apexyard.projects.yaml`.

### Blocker 2 — false justifying prose → CLOSED
The claim that `_lib-ops-root.sh` "ships in every managed project's `.claude/hooks/` tree ... that's why sourcing it via `git rev-parse --show-toplevel` below is safe" is removed. The replacement is accurate: "managed-project clones carry **no `.claude/hooks/` directory at all** (there is no framework mechanism that installs one there) — sourcing from that path silently fails". Verified: the workspace clone has no `.claude/hooks/`. The added "keep two steps distinct" paragraph correctly separates *locating the lib* (walk-up) from *deciding the real ops root* (`resolve_ops_root`, pin-first) — matches the code.

### Scope
Diff between the two HEADs is confined to `.claude/skills/start-ticket/SKILL.md` (the two fixes). `status/SKILL.md` was already in the prior commit; unchanged in this delta. Nothing else touched.

### Checklist
- Architecture & Design: Pass (sourcing mechanism now correct)
- Code Quality: Pass
- Testing: Pass — reproduced the extended matrix incl. the workspace-clone rows
- Security: Pass (docs-only)
- PR Description & Glossary: Pass — prose now accurate; glossary `resolve_ops_root` / "Sibling walk-up" entries correct
- Summary Bullet Narrative: Pass (exemplary what+why)
- Technical Decisions (AgDR): N/A (docs fix, no new decision)
- Adopter Handbooks: N/A (none in tree)

### Verdict
**APPROVED**

---
Reviewed commit: `10913840adbae5e1327ad8e6e4fca9f837927cac`

atlas-apex (commented):
## Code Review: PR #884

**Commit**: `10913840adbae5e1327ad8e6e4fca9f837927cac`

### Summary
Documentation-only fix to `/start-ticket` (and a one-line doc-accuracy fix in `/status`). It repairs ops-root resolution so the session marker can be written on split-portfolio **v2** forks, where `onboarding.yaml` + `apexyard.projects.yaml` live in the sibling portfolio repo rather than the fork itself. The prior inline walk-up only recognised the legacy v1 pair, so it returned empty on every v2 fork and the marker had to be hand-written (#882). The fix delegates to the shared `resolve_ops_root` (pin-first) and `portfolio_registry` resolvers instead of re-deriving the walk inline.

### What I checked (trust-chain-adjacent verification)
The diff is prose+bash in `SKILL.md`, so I verified the described bash against the real libraries it calls:

- **`resolve_ops_root` / `resolve_ops_root_walk`** (`.claude/hooks/_lib-ops-root.sh`) exist and honour BOTH the v2 `.apexyard-fork` marker (checked first) AND the legacy v1 `onboarding.yaml`+`apexyard.projects.yaml` pair — so the single-fork/v1 path is preserved, no regression.
- **Step 4a walk-up** locates `_lib-ops-root.sh` by walking `$PWD` up the filesystem rather than `git rev-parse --show-toplevel`. Correct: from a `workspace/<project>/` clone (which carries no `.claude/hooks/` tree), `--show-toplevel` would resolve to the project repo and the source would silently fail. The filesystem walk-up reaches the ops fork because workspace clones are nested under it. The lib-location step and the pin-first `resolve_ops_root` decision are kept correctly distinct.
- **Step 4b** sources `_lib-read-config.sh` then `_lib-portfolio-paths.sh` from `$ops_root` and calls `portfolio_registry`. Verified `portfolio_registry` exists, honours the `.portfolio.registry` override, and self-derives its base to the ops fork even from a nested clone. Source order is correct (belt-and-suspenders, since `_portfolio_get` self-sources read-config).
- **Single-fork (v1) regression check:** walk-up finds the fork via `_lib-ops-root.sh` presence; `resolve_ops_root` returns it via the v1 pair anchor; registry resolves relative to the same root. No behavioural change for v1 adopters. The PR's own test table (v1 root/subdir, v2 root/subdir, v2 workspace clone, outside-any-fork fail-closed) matches the lib behaviour I read.

### Checklist Results
- Testing:                   Pass (no harness for SKILL.md prose; author validated extracted bash against temp-dir layouts + re-ran `test_ops_root.sh` 8/8, `test_resolve_ops_root_pin.sh` 8/8)
- Security:                  Pass (docs-only; verified bash matches trust-chain libs, no new attack surface)
- Performance:              N/A
- PR Description & Glossary:  Pass (thorough glossary; ticket linked via Closes #882)
- Summary Bullet Narrative:  Pass (narrative, verbs + rationale, well over the label-only threshold)
- Technical Decisions (AgDR): N/A (bug fix aligning docs with the existing shared-resolver pattern; references existing apexyard#381 pin-first; no new library/framework/pattern)
- Adopter Handbooks:         N/A (diff is markdown-only — no language/domain handbooks trigger; the always-load architecture/general handbooks have no applicable trigger patterns for a docs change, and migration-safety's blocking rules require migration files, none touched)
- Fallow Static Analysis:    N/A (not JS/TS)

### Issues Found
None blocking.

### Suggestions
- nit: Step 4a resolves `$ops_root` **pin-first**, but step 4b's `portfolio_registry` internally re-derives its base root via `_portfolio_root`, which starts from `git rev-parse --show-toplevel` (a filesystem walk-up, **not** pin-first). In the rare edge case the PR body itself calls out — cwd inside an ops-fork-shaped `/tmp` build clone whose session pin points at a *different* real fork — step 4b could resolve the registry relative to a different root than the `$ops_root` step 4a chose. This is a pre-existing property of the shared lib (`portfolio_registry` takes no ops-root argument and always self-derives), not introduced here, and every normal `/start-ticket` case (real fork, or a `workspace/<project>/` clone nested under it) resolves consistently. Flagging only as a future-hardening note — not a blocker for this docs fix.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `10913840adbae5e1327ad8e6e4fca9f837927cac`

atlas-apex (approved):
Approved (atlas-apex). Rex reviewed: docs-only ops-root resolution fix; both single-fork and split-portfolio v2 layouts verified, no single-fork regression, marker paths resolve correctly. CI green.

PR #885, opened 2026-07-12
fix(#883): exempt out-of-repo writes from the ticket-first gate

### LINKED ISSUE #883: [Bug] Ticket-first gate blocks out-of-repo home-dotfile edits — no legit path on forks with Issues disabled
## Given / When / Then
- **Given** a session whose cwd is inside an apexyard fork, on a fork with GitHub Issues disabled (the fork default)
- **When** the operator asks for an edit to a **home-directory dotfile entirely outside any repo** (e.g. appending exports to `~/.zshrc` during machine setup)
- **Then** `require-active-ticket.sh` blocks it (the Edit/Bash-write matchers fire on cwd context regardless of target path) — and because the fork has Issues off, there is no legitimate path to satisfy the gate without first enabling Issues and filing a chore ticket for a shell-profile edit.

## Discussion
The ticket-first gate exists to protect repo/source changes. A write target outside every git repo (home dotfiles, /etc-style config) is arguably out of the gate's jurisdiction — the current behaviour turns machine setup into a governance dead-end on fresh installs. Reported by a second operator during first-machine setup.

## Proposed fix
In `require-active-ticket.sh` (and the `_lib-detect-bash-write.sh` path extraction): resolve the WRITE TARGET's path; if it is outside the ops fork, outside every registered `workspace/<project>`, and outside any git repo under governance, exempt it. Keep everything inside governed trees gated exactly as today. Add tests: `~/.zshrc` edit passes without a marker; in-repo source edit still blocks.

## Acceptance Criteria
- [ ] Out-of-repo home-path writes are exempt from the active-ticket gate.
- [ ] Governed-tree behaviour unchanged (existing suite green + new cases).
- [ ] Decision documented (rule/README note) so the boundary is explicit.

## Glossary
| Term | Definition |
|------|------------|
| Ticket-first gate | `require-active-ticket.sh` — blocks code writes until a session ticket marker exists (Pre-Build Gate). |

me2resh:
## Summary

- **Extended the out-of-repo exemption to Edit/Write/MultiEdit, not just Bash** — the existing #569 exemption only fired for Bash-detected writes (`echo > file`, `tee`, etc.). An `Edit`-tool write to a home dotfile like `~/.zshrc` during machine setup was still gated, and on a fork with GitHub Issues disabled there was no legitimate way to satisfy the gate (no tracker to file a chore ticket in). This PR moves the exemption logic into a single check that fires for both tool shapes.
- **Resolves the write target's real (symlink-free) path before judging** — the old exemption compared raw string paths, so a symlink living under `$HOME` that pointed *into* a governed tree (e.g. `~/link-into-repo → the ops fork`) would look "outside" the repo under a naive prefix check and silently bypass the gate. The new `_resolve_real_path` helper canonicalizes the target (following symlinks in the existing portion of the path, since a not-yet-created tail can't itself be a symlink) before any boundary check runs.
- **Exemption requires all three conditions, not just "not a git repo"** — outside the ops fork, outside every registered `workspace/<project>`, and not inside any git repository at all. The three-way check matters for split-portfolio layouts where a workspace project clone may not itself be a `git init`'d repo (a docs-only project, or a test fixture) — relying only on "is this inside some git repo" would have wrongly exempted a governed-but-non-git workspace path.
- **Fail-closed on ambiguity** — an unresolvable Bash write-target (the extractor couldn't identify a path, e.g. `python -c '...'` with a dynamically constructed path) is never exempted; it falls straight through to the existing ticket gate, unchanged.
- **Checks BOTH the raw and the resolved target for ops/workspace containment** — added after Hakim's security review (see "Security review follow-up" below) surfaced the mirror-image hole: a registered `workspace/<proj>` that is *itself* a symlink pointing OUTSIDE `workspace/` to a non-git directory was wrongly exempted, because containment was checked only on the symlink-resolved path. Ops-root and workspace containment are now evaluated against both the raw (pre-resolution) target and the fully resolved target, against the same canonicalized boundary anchors — a path is governed (not exempt) if *either* check lands inside.

## Security review follow-up

Hakim (security-reviewer) reviewed this PR and flagged one LOW, non-blocking finding: a `workspace/<proj>` entry that is itself a symlink to an outside, non-git directory was wrongly exempted (`ops=0 ws=0 git=0 → exempt`), because the symlink resolution that correctly defends "link INTO the repo" defeats the workspace containment check when the governed path *contains* a link pointing *out*. Fixed by checking both the raw and resolved path against the ops/workspace boundaries (exempt only if both say "outside"), with a dedicated regression pair (cases 39–40 below).

## Testing

Extended `.claude/hooks/tests/test_require_active_ticket_bash.sh` with 10 new cases (31–40) covering: Edit-tool out-of-repo dotfile write (the core #883 repro), MultiEdit shape, Bash relative write from an out-of-repo CWD, in-repo Edit write still blocked (regression), symlink-into-repo NOT bypassing the gate, the same symlink case working correctly WITH an active ticket, a workspace-project write with no `.git` still blocked, an unresolvable Bash target staying gated, a `workspace/<proj>` symlink to an outside non-git directory still blocked without a ticket (case 39 — the security-review finding), and the same symlinked-workspace case working correctly WITH an active ticket (case 40).

```
.claude/hooks/tests/test_require_active_ticket_bash.sh: PASS 43  FAIL 0   (33 pre-existing + 10 new)
bin/run-hook-tests.sh (full suite):                     PASS 98  FAIL 0  SKIP 0
```

`bash -n` and `shellcheck -x` both clean on the modified hook — the shellcheck notes present are pre-existing (SC2016/SC2295 on lines untouched by this diff), not introduced by this change. Case 39 was verified to fail (wrongly exempt, rc=0) against the pre-fix hook and pass against the fix, confirming the regression test is load-bearing.

Closes #883

## Glossary

| Term | Definition |
|------|------------|
| Ticket-first gate | `require-active-ticket.sh` — blocks code writes until a session ticket marker exists (workflow-gates Pre-Build Gate). |
| Out-of-governance exemption | The escape valve this PR adds: a write target outside the ops fork, outside every registered workspace project, and outside any git repository at all is exempt from the ticket gate. |
| Governed tree | The ops fork itself, plus every project registered in `apexyard.projects.yaml` under `workspace/<project>`. |
| Fail-closed | When a check can't confidently determine "this is safe to exempt," it defaults to gating (blocking) rather than exempting. |
| Raw vs. resolved path | Raw = the write target exactly as named (before following any symlinks). Resolved = the same target after all symlinks in its existing ancestry are followed to their real location. Containment is now checked against both. |

### DIFF
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -34,6 +34,64 @@
 #   - anything under projects/*/docs/ (per-project apexyard docs)
 #
 # Everything else (source code, config, infra) requires a ticket marker.
+#
+# Out-of-governance exemption (apexyard#883): a write TARGET entirely
+# outside every governed tree (the ops fork AND every registered
+# workspace/<project> clone) AND not inside ANY git repository at all is
+# outside this gate's jurisdiction — home dotfiles (~/.zshrc), /etc-style
+# machine config, /tmp scratch files. See the "Out-of-governance
+# exemption" block below for the fail-closed resolution rules (symlinks
+# resolved before judging; unresolvable/ambiguous targets stay gated).
+
+# Resolve PATH to its canonical, symlink-free absolute form. Walks up to
+# the nearest EXISTING ancestor, physically resolves it (`pwd -P`, which
+# follows symlinks), then re-appends any not-yet-created tail literally —
+# a tail that doesn't exist yet cannot itself be a symlink. This mirrors
+# `realpath -m` without depending on GNU coreutils (not guaranteed present
+# on macOS/BSD). Echoes the resolved path, or nothing if even "/" can't be
+# stat'd (should not happen for a well-formed absolute path).
+#
+# Why this matters (#883): without resolving symlinks first, a symlink
+# living under $HOME that POINTS INTO a governed tree (e.g.
+# ~/link-into-repo → the ops fork) would compare as "outside" the repo
+# under a naive string-prefix check, silently bypassing the gate.
+_resolve_real_path() {
+  local p="$1" dir tail=""
+  [ -n "$p" ] || return 0
+  dir="$p"
+  while [ -n "$dir" ] && [ "$dir" != "/" ] && [ ! -e "$dir" ]; do
+    if [ -z "$tail" ]; then
+      tail="$(basename "$dir")"
+    else
+      tail="$(basename "$dir")/$tail"
+    fi
+    dir="$(dirname "$dir")"
+  done
+  if [ ! -e "$dir" ]; then
+    # Nothing on the path exists at all — cannot resolve. Should not
+    # happen for an absolute path since "/" always exists.
+    return 0
+  fi
+  if [ -d "$dir" ]; then
+    dir="$(cd "$dir" 2>/dev/null && pwd -P)"
+  else
+    # $dir resolved to an existing FILE (not a directory) partway through
+    # the walk — canonicalize its parent and re-append its own basename.
+    local parent
+    parent="$(cd "$(dirname "$dir")" 2>/dev/null && pwd -P)"
+    if [ -n "$parent" ]; then
+      dir="$parent/$(basename "$dir")"
+    else
+      dir=""
+    fi
+  fi
+  [ -n "$dir" ] || return 0
+  if [ -n "$tail" ]; then
+    printf '%s/%s' "$dir" "$tail"
+  else
+    printf '%s' "$dir"
+  fi
+}
 
 INPUT=$(cat)
 TOOL_NAME=$(echo "$INPUT" | jq -r '.tool_name // empty' 2>/dev/null)
@@ -170,25 +228,12 @@ if [ -n "$REPO_ROOT" ] && [ -n "$FILE_PATH" ]; then
   esac
 fi
 
-# Bash-write: absolute paths outside the repo root are outside the tracked
-# source tree (e.g. /tmp/, /var/, /usr/, system-temp paths). A write there
-# cannot mutate apexyard-governed content, so no ticket is required (#569).
-# This check runs only for the Bash path (FILE_PATH set via extractor) and
-# only when FILE_PATH is absolute and does NOT strip to a REL_PATH (meaning
-# it's outside REPO_ROOT). We conservatively skip this for non-Bash tools —
-# they supply an explicit file_path from the tool call that we trust fully.
-if [ "$TOOL_NAME" = "Bash" ] && [ -n "$FILE_PATH" ] && [ -n "$REPO_ROOT" ]; then
-  case "$FILE_PATH" in
-    /*)
-      # Absolute path — was it stripped to a repo-relative form?
-      if [ "$REL_PATH" = "$FILE_PATH" ]; then
-        # REL_PATH is still absolute: FILE_PATH is NOT under REPO_ROOT.
-        # It's a system path or temp path — exempt.
-        exit 0
-      fi
-      ;;
-  esac
-fi
+# NOTE: the narrow "Bash absolute path outside REPO_ROOT" exemption that
+# used to live here (#569) has been superseded by the out-of-governance
+# exemption below (apexyard#883), which covers BOTH Bash and Edit/Write/
+# MultiEdit, resolves symlinks before judging, and checks the actual
+# governance boundaries (ops root + registered workspaces) rather than
+# just "outside the nearest git repo". See that block for the full design.
 
 # Exempt paths.
 #
@@ -271,6 +316,81 @@ if [ -n "$OPS_ROOT" ] && [ -f "$HOOK_DIR/_lib-portfolio-paths.sh" ] && [ -f "$HO
   resolved_ws=$(portfolio_workspace_dir 2>/dev/null)
   if [ -n "$resolved_ws" ]; then
     WORKSPACE_DIR="$resolved_ws"
+  fi
+fi
+
+# --- Out-of-governance exemption (apexyard#883) -------------------------
+#
+# A write TARGET that is outside every governed tree — the ops fork
+# itself AND every registered workspace/<project> clone — AND not inside
+# ANY git repository at all is outside this gate's jurisdiction. Applies
+# to BOTH Bash-detected writes and Edit/Write/MultiEdit tool calls (the
+# earlier #569 exemption was Bash-only, which left Edit-tool writes to
+# e.g. ~/.zshrc gated with no legitimate way to satisfy the gate on forks
+# with GitHub Issues disabled — the exact bug reported in #883).
+#
+# Fail-closed directions (deliberate):
+#   - An unresolvable FILE_PATH (empty — a Bash target the extractor
+#     couldn't identify) is NEVER exempted here; it falls through to the
+#     ticket gate below, unchanged from before this change.
+#   - Relative paths (a Bash write-target like `src/app.ts`) resolve
+#     against the hook's CWD before judging.
+#   - Symlinks are resolved to their real path before judging (via
+#     _resolve_real_path above), so a symlink living under $HOME that
+#     POINTS INTO a governed tree does not slip through as "outside" it.
+#   - Being inside SOME git repository that is neither the ops fork nor a
+#     registered workspace project does NOT exempt the write on its own —
+#     but see the three-way check below: ops-root and workspace
+#     boundaries are checked EXPLICITLY (not merely inferred from "is
+#     this a git repo"), because a split-portfolio workspace project is
+#     governed even when its clone isn't itself a git repository (e.g. a
+#     docs-only project, or a test fixture that never ran `git init`).
+#     Exemption requires ALL THREE checks below to say "outside".
+if [ -n "$FILE_PATH" ]; then
+  case "$FILE_PATH" in
+    /*) _og_abs_target="$FILE_PATH" ;;
+    *)  _og_abs_target="$PWD/$FILE_PATH" ;;
+  esac
+  _og_real_target="$(_resolve_real_path "$_og_abs_target")"
+  if [ -n "$_og_real_target" ]; then
+    _og_in_ops=0
+    _og_in_ws=0
+    _og_in_git=0
+
+    if [ -n "$OPS_ROOT" ]; then
+      _og_real_ops="$(cd "$OPS_ROOT" 2>/dev/null && pwd -P)"
+      if [ -n "$_og_real_ops" ]; then
+        case "$_og_real_target" in
+          "$_og_real_ops") _og_in_ops=1 ;;
+          "$_og_real_ops"/*) _og_in_ops=1 ;;
+        esac
+      fi
+    fi
+
+    if [ -n "$WORKSPACE_DIR" ] && [ -d "$WORKSPACE_DIR" ]; then
+      _og_real_ws="$(cd "$WORKSPACE_DIR" 2>/dev/null && pwd -P)"
+      if [ -n "$_og_real_ws" ]; then
+        case "$_og_real_target" in
+          "$_og_real_ws") _og_in_ws=1 ;;
+          "$_og_real_ws"/*) _og_in_ws=1 ;;
+        esac
+      fi
+    fi
+
+    # Nearest-existing-ancestor git-repo probe on the REAL (symlink-
+    # resolved) target — catches any git repo, governed or not.
+    _og_probe="$_og_real_target"
+    while [ -n "$_og_probe" ] && [ "$_og_probe" != "/" ] && [ ! -d "$_og_probe" ]; do
+      _og_probe="$(dirname "$_og_probe")"
+    done
+    if [ -n "$_og_probe" ] && [ -d "$_og_probe" ] \
+       && git -C "$_og_probe" rev-parse --show-toplevel >/dev/null 2>&1; then
+      _og_in_git=1
+    fi
+
+    if [ "$_og_in_ops" = 0 ] && [ "$_og_in_ws" = 0 ] && [ "$_og_in_git" = 0 ]; then
+      exit 0
+    fi
   fi
 fi
 
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -340,12 +340,33 @@ fi
 #     POINTS INTO a governed tree does not slip through as "outside" it.
 #   - Being inside SOME git repository that is neither the ops fork nor a
 #     registered workspace project does NOT exempt the write on its own —
-#     but see the three-way check below: ops-root and workspace
-#     boundaries are checked EXPLICITLY (not merely inferred from "is
-#     this a git repo"), because a split-portfolio workspace project is
-#     governed even when its clone isn't itself a git repository (e.g. a
-#     docs-only project, or a test fixture that never ran `git init`).
-#     Exemption requires ALL THREE checks below to say "outside".
+#     the ops-root/workspace boundaries are checked EXPLICITLY (not
+#     merely inferred from "is this a git repo"), because a
+#     split-portfolio workspace project is governed even when its clone
+#     isn't itself a git repository (e.g. a docs-only project, or a test
+#     fixture that never ran `git init`).
+#   - RAW-AND-RESOLVED containment (security review finding on #885,
+#     apexyard PR #885 — Hakim/security-reviewer): checking ops-root and
+#     workspace containment ONLY on the symlink-resolved target has a
+#     mirror-image hole to the one `_resolve_real_path` fixes. If a
+#     REGISTERED workspace/<project> entry is ITSELF a symlink whose real
+#     target lives OUTSIDE workspace/ in a non-git directory
+#     (`workspace/proj -> /some/nongit/dir`), resolving the target
+#     "escapes" the workspace boundary the same way the fix escapes a
+#     symlink-into-repo attack — the RESOLVED path no longer sits under
+#     WORKSPACE_DIR, so the naive single-check reads it as ungoverned and
+#     wrongly exempts it. The fix: evaluate ops-root/workspace
+#     containment against BOTH the RAW (pre-resolution) absolute target
+#     AND the fully resolved target, using the SAME canonicalized
+#     boundary anchors for both comparisons (only the anchors are
+#     canonicalized once; the raw target itself is deliberately left
+#     symlink-unresolved so a governed raw path can't be laundered away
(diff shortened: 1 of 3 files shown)


me2resh (commented):
## Security Review: PR #885 — out-of-repo exemption for the ticket-first gate

**Commit reviewed**: `3e0387945c2967d1038d129f7907c051b45ff932`
🛡️ Reviewed by Hakim (Security Auditor) — advisory (trust-chain change: `require-active-ticket.sh` is the Pre-Build gate).

### Verdict: **COMMENT — approve-leaning.** One LOW fail-open residual; not a blocker.

This change is a **net hardening** of the gate. The escape valve it adds is fail-closed in every direction I attacked except one narrow, non-adversarial case. It also *fixes* a real pre-existing bypass: the old #569 code did a naive string-prefix check, so a symlink under `$HOME` pointing **into** a governed tree would have compared as "outside" and slipped through. The new `_resolve_real_path` + three-way boundary check closes that.

### What I actually executed

- **Existing suite**: `test_require_active_ticket_bash.sh` → **41/41 PASS** (33 pre-existing + 8 new). `bin/run-hook-tests.sh` full suite was green per PR body; I re-ran the ticket-gate file directly.
- `bash -n` clean; `shellcheck -x` — the 4 notes (SC2016/SC2295 on lines 150/227/439/450) are all **pre-existing**, outside this diff. The new regions (58–94, 322–395) are shellcheck-clean.
- **12 additional adversarial probes** in throwaway ops-fork sandboxes, several **instrumented** to print the `ops`/`ws`/`git` flags directly. Results:

| Probe | Attack | Result |
|-------|--------|--------|
| Symlink `$HOME/link → ops`, write `link/src/app.ts` | symlink-into-repo bypass | **GATED** (resolved→ops) ✓ |
| 2-hop symlink chain into ops | chained-symlink bypass | **GATED** ✓ |
| Symlinked *parent* dir into ops/src, new file | parent-symlink | **GATED** ✓ |
| `..` in the non-existent tail escaping `/src` | traversal | **GATED** (prefix stays ops) ✓ |
| `/tmp/../ops/src/app.ts` traversal into fork | traversal | **GATED** (cd+`pwd -P` canonicalizes) ✓ |
| Unregistered **nested** git repo under ops | nested-repo | **GATED** ✓ |
| Unregistered git repo **outside** ops | per design: not exempt | **GATED** ✓ |
| OPS_ROOT anchor removed, write in ops | resolver-failure fail direction | **GATED** — git-probe backstop fires (`ops=0` but `git=1`) ✓ |
| Existing ancestor is a **file** partway (in-repo) | file-partway branch | **GATED** ✓ |
| Newline / spaces in path (in & out of repo) | quoting / injection | correct + no crash ✓ |
| Relative write from out-of-repo cwd | rel-path resolution | **EXEMPT** (correct) ✓ |
| Genuine `$HOME/.zshrc`, non-git | the intended exemption | **EXEMPT** (correct) ✓ |

**Fail-direction audit (the key ask):** every `_resolve_real_path` failure path (`return 0` with empty output) leaves `_og_real_target` empty → the whole exemption block is skipped → falls through to the ticket gate → **GATED**. `cd … pwd -P` failures (permission denied on a parent) → empty → gated. Unresolvable/empty Bash target → exemption never entered → gated. I could not find a helper-failure branch that yields *exempt*. Confirmed also that when OPS_ROOT is unresolvable the pure-string ops/ws checks degrade to 0 but the **git-probe** independently gates anything sitting in a git repo — so a governed write is never exempted just because OPS_ROOT resolution failed.

### Finding (LOW / fail-open) — symlinked-out non-git workspace project

A `workspace/<proj>` that is a **symlink whose real target is outside `workspace/` AND is not itself inside any git repo** is wrongly **EXEMPTED**. Instrumented repro:

```
workspace/proj -> /some/nongit/dir        (symlink out of the workspace dir)
Edit  workspace/proj/x.ts
  => real_target=/some/nongit/dir/x.ts   ops=0  ws=0  git=0  => exit 0 (EXEMPT)
```

Root cause: the containment checks run on the **symlink-resolved** target. Resolution is exactly what defeats the intended `WORKSPACE_DIR` string check here — after following the project symlink, the real path no longer sits under the resolved `workspace/` prefix, so `ws=0`; and because the target is non-git, `git=0` too. This is the mirror image of the symlink-into-repo case the PR correctly defends: resolving symlinks helps when a link points *into* governance, but *hurts* when a governed path *contains* a link pointing *out*. The explicit `WORKSPACE_DIR` check was added precisely to govern **non-git** workspace projects (docs-only / test-fixture — see test 37), and this is the one shape that slips past it.

**Why LOW, not blocking:**
- Narrow precondition intersection: symlinked *out of* `workspace/` **and** target is non-git. Real project clones are git repos → those gate correctly (I verified: symlinked-out **git** workspace project → GATED via the git-probe). Test 37's ordinary (non-symlink) non-git project also still gates.
- Not an adversarial escalation. This gate is a **process/discipline** control, not a boundary against a malicious local user — anyone able to plant a symlink in the ops fork's `workspace/` can already bypass the gate trivially (`touch` a marker, edit the hook). The framework's own docs treat these markers as "grep-able rule violations," not attacker-proof. The concern is a *legitimate* governed write silently skipping the ticket gate under an unusual-but-plausible adopter layout.

**Suggested fix (minimal, additive):** test **both** the raw pre-resolution absolute target **and** the resolved target against the ops/workspace prefixes; treat as governed if *either* is inside. I checked the truth table — this closes the finding without regressing anything:
- genuine `$HOME/.zshrc`: raw ∉, resolved ∉, non-git → still EXEMPT ✓
- symlink-into-repo (#883 defense, test 35): raw ∉ but resolved ∈ ops → still GATED ✓
- symlinked-out non-git workspace proj (this finding): raw ∈ workspace → now GATED ✓

Worth a follow-up ticket + one regression case (`workspace/<proj>` symlink → non-git dir → BLOCKED). No need to hold this PR: the diff strictly improves the gate's symlink posture over what's on `dev` today.

### Checklist
- Secrets/credentials: **Pass** (no secrets; hook reads paths only)
- Injection (path/shell): **Pass** — all interpolations quoted; globs live only in `case` patterns, targets are subjects; newlines/spaces handled
- Symlink / traversal / TOCTOU: **Pass with one LOW residual** (above); no resolve-vs-judge asymmetry in the exemption decision
- Fail-closed on error: **Pass** — every unresolvable/error path lands GATED
- Auth/authorisation model: **Pass** — no change to the trust model; process-gate semantics preserved

📌 Reviewed commit: `3e0387945c2967d1038d129f7907c051b45ff932`

me2resh (commented):
## Code Review: PR #885

**Commit**: `3e0387945c2967d1038d129f7907c051b45ff932`

### Summary
Unifies the ticket-first gate's out-of-repo exemption across Edit/Write/MultiEdit **and** Bash. The old #569 exemption fired only on Bash-detected writes, so an Edit-tool write to a home dotfile (`~/.zshrc`) stayed gated with no legitimate escape on a fork with GitHub Issues disabled (#883). The fix adds a `_resolve_real_path` symlink canonicalizer and a fail-closed triple condition (outside ops fork **and** outside every registered `workspace/<project>` **and** not inside any git repo) evaluated on the canonicalized path.

### Checklist Results
- Testing:                    Pass (41/41 file suite, 98/98 full suite, both with `env -u CLAUDE_CODE_SESSION_ID`)
- Security:                   Pass (fail-closed; trust-chain note below)
- Performance:                Pass (a few extra `git rev-parse` / `pwd -P` per gated write — negligible)
- Technical Decisions (AgDR): Advisory (see below) — N/A as blocker
- Adopter Handbooks:          N/A (TS/migration handbooks don't apply to a shell-hook diff)

### Verification performed (not just trusting the claims)
- **Exemption logic traced.** The triple condition (`_og_in_ops`/`_og_in_ws`/`_og_in_git`) is evaluated on the symlink-resolved target, and ops-root + workspace boundaries are canonicalized the same way (`cd … && pwd -P`) so the comparison is symmetric. Sound.
- **New-file canonicalization.** `_resolve_real_path` walks to the nearest *existing* ancestor, physically resolves it, then re-appends the not-yet-created tail literally — correct, and the "a tail that doesn't exist yet can't be a symlink" reasoning holds. New-file writes canonicalize the existing parent rather than failing open.
- **`..` traversal back into the ops fork** (`<ops>/../<ops>/src/app.ts`) → **gated** (verified). `pwd -P` collapses the `..`, so a crafted traversal can't masquerade as outside.
- **Write under an unregistered/unrelated git repo** → **gated** (verified). The nearest-existing-ancestor `git -C … rev-parse` probe catches *any* git repo, governed or not — exactly the "narrower than not-inside-a-registered-repo" intent.
- **Symlink-into-governed-tree** does not bypass (test 35) and is correctly allowed *with* an active ticket (test 36) — proving the symlink resolves to governed content rather than being wrongly exempted.
- **Suites run locally**, both green. `bash -n` clean; the three shellcheck notes (SC2016/SC2295) are all on pre-existing untouched lines (150, 227, 439, 450) — the PR body's claim is accurate, the new code is warning-free.

- **nit — multi-target Bash is a pre-existing gap, not introduced here.** A command with two write targets where the first is out-of-repo and the second is in-repo (`echo a > /tmp/x && echo b > <ops>/src/app.ts`) is exempted, because `bash_extract_write_target` returns only the first target (`head -n 1`). I confirmed the **base** hook behaves identically (rc=0), so this PR is neutral — it inherits the documented best-effort/false-negative-preferred property of `_lib-detect-bash-write.sh`. Worth a separate hardening ticket on the extractor, but out of scope for #883 and not a regression.
- **suggestion — security-auditor trigger.** This touches the trust chain (`.claude/hooks/**`) and modifies a load-bearing gate; per role-triggers #777 the orchestrator may want a `/security-review` pass. The fail-closed analysis holds from a code-review lens (every "unsure" direction gates rather than exempts), but a second security-specific pass is cheap insurance.
- **suggestion — AgDR.** This is a security-boundary decision with real trade-offs (three-way check vs. "not a git repo"; symlink-resolution approach). It's documented thoroughly in the rule-doc note + inline comment block + #883, which matches the #569 precedent (fix + rule-doc, no dedicated AgDR), so I'm treating it as advisory rather than blocking. Consider an AgDR if you want the trade-off analysis captured in the searchable record.

### Verdict
**APPROVED**

Careful, fail-closed implementation with genuine adversarial test coverage; my own traversal and unregistered-git probes both gate correctly, and both suites are green. The one gap I found (multi-target Bash) is pre-existing and out of scope.

---
📌 Reviewed commit: `3e0387945c2967d1038d129f7907c051b45ff932`

me2resh (commented):
## Code Review: PR #885 (re-review)

**Commit**: `ccbbeaadaa329bf553276fea10913a75f88ca0eb`

Re-review of the delta since the previously-approved SHA `3e03879`. The builder folded in Hakim's LOW security finding.

### Summary
Since the approved SHA, the sole functional change is the workspace/ops containment check in `require-active-ticket.sh`: it now evaluates BOTH the raw (pre-resolution) absolute target AND the symlink-resolved target against canonicalized ops/workspace boundary anchors. A path is governed (not exempt) if *either* lands inside. Plus two regression cases (39, 40).

### Delta verification (vs. approved `3e03879`)
- **Scope is exactly as described.** `git diff 3e03879..ccbbeaa` touches only two files: `require-active-ticket.sh` (the raw+resolved check + its comment block) and the test file (cases 39–40). No `workflow-gates.md` change (that landed in the approved commit), no other files. PR-body prose updated. Nothing else.
- **The raw check cannot introduce a fail-open.** It only *adds* "inside" votes (`_og_in_ops_raw` / `_og_in_ws_raw`) — monotonic toward gating. It strictly narrows the exemption; it can never widen it. So the change cannot create a new bypass, only over-gate.
- **Over-gating direction is fail-closed-correct.** For the raised edge case — a raw *relative* target from an in-fork cwd that escapes via `../` and resolves outside — the raw abs string `<ops>/../outside/...` literally matches `<ops>/*`, so the raw check votes "inside" and the write is **gated** even though it resolves outside. Governed wins. That is the safe direction: the failure mode that matters (wrongly *exempting* a governed write) is the one the raw check closes; re-gating a legitimate out-of-repo relative write is an inconvenience, not a hole. The core #883 repro (absolute home dotfiles like `~/.zshrc`) is unaffected.
- **Git-repo probe staying single (resolved-only) is sound.** Directory traversal (`-d`, `git -C`) follows symlinks at the OS level regardless of which path string reaches the location, so raw-path and resolved-path traversal of the same logical target always land on the same physical dir. A second raw probe would be redundant.

### Load-bearing regression test — independently confirmed
Checked the HEAD test file against the pre-fix hook (`git show 3e03879:...`): case 39 (`workspace/proj → external non-git dir`, symlink write, no ticket) **fails wrongly-exempt (rc=0)** against pre-fix and **passes (BLOCKED, rc=2)** against the fix. Case 40 (same path WITH a ticket → allowed) is a coherence guard. The regression pair is genuinely load-bearing.

### Tests (env -u CLAUDE_CODE_SESSION_ID)
- `test_require_active_ticket_bash.sh`: **PASS 43 / FAIL 0** (incl. 39, 40)
- `bin/run-hook-tests.sh` full suite: **PASS 98 / FAIL 0 / SKIP 0**
- `bash -n` clean; no new shellcheck findings (SC2016/SC2295 are pre-existing on untouched lines).

### Security review follow-up section — accurate
The body's "Security review follow-up" section correctly describes the finding (`ops=0 ws=0 git=0 → exempt` on a workspace/<proj> symlink to an outside non-git dir), the fix (raw AND resolved containment; exempt only if both say "outside"), and the regression pair. It matches the code and the confirmed load-bearing test.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass (clear comments, fail-closed reasoning documented in-hook)
- Testing: Pass (load-bearing regression pair; full suite green)
- Security: Pass (change strictly narrows the exemption; closes the mirror-image symlink hole)
- Performance: Pass (one extra `pwd -P` on anchors + string case-matches; negligible)
- Technical Decisions (AgDR): N/A (bugfix, no new decision)

### Verdict
**APPROVED**

Halting at the CEO merge gate — not merging.

---
📌 Reviewed commit: `ccbbeaadaa329bf553276fea10913a75f88ca0eb`

atlas-apex (commented):
## Code Review: PR #885

**Commit**: `ccbbeaadaa329bf553276fea10913a75f88ca0eb`

### Summary
Extends the ticket-first gate's out-of-repo exemption (previously Bash-only, #569) to Edit/Write/MultiEdit, resolves the write target's real (symlink-free) path before judging, and gates on the actual governance boundaries (ops fork + registered `workspace/<project>`) plus a "not inside any git repo at all" probe — rather than the old naive string-prefix check. This is a security-sensitive change to a merge/write gate, so I reviewed it adversarially for bypass holes where an in-repo source edit could be misclassified as out-of-repo and skip the ticket gate.

### Checklist Results
- Testing:                    Pass (43/43 in the hook's own suite; adversarial probes added below)
- Security:                   Pass — no exploitable bypass found (see analysis)
- Performance:                Pass (a few `cd`/`git rev-parse` per write event; negligible)
- PR Description & Glossary:  Pass (Glossary present; `Closes #883`; narrative summary bullets)
- Technical Decisions (AgDR): N/A — bug fix to an existing exemption mechanism; design rationale documented inline + in workflow-gates.md, no new library/framework/pattern
- Adopter Handbooks:          N/A — diff is shell + markdown; no language handbook fires, and architecture/general handbooks (clean-architecture-layers, migration-safety) don't apply to a shell hook

### Adversarial security analysis (the core of this review)
The exemption fires **only** when ALL of these say "outside": raw-ops, raw-ws, resolved-ops, resolved-ws containment, **and** the git-repo probe says "not inside any git repo." I probed the misclassification surface:

1. **git-probe is a hard backstop for the ops fork.** The ops fork is always a git repo, so any *writable* path inside it resolves to a directory under `.git` and sets `_og_in_git=1` → never exempt. Verified: absolute in-repo, relative in-repo (Bash `echo > src/app.ts`), and `..`-laden in-repo paths all BLOCK.
2. **`..` traversal cannot launder a governed path.** `_resolve_real_path` canonicalizes via `cd … && pwd -P`, which fully resolves `..` for any path whose ancestry exists. I tested `src/subdir/../app.ts`, `src/subdir/../newfile.ts` (new file), and a sibling-pivot `pivot/../<repo>/src/app.ts` — all BLOCK. The only case where `..` stays unresolved is when a **non-existent** intermediate component precedes it — and in exactly that case the OS write itself fails with ENOENT (I confirmed the path is non-writable), so it can't mutate governed content. "Exempt" therefore implies "canonically resolves outside" OR "not writable at all."
3. **Symlinks, both directions, are covered.** Symlink-into-repo does NOT bypass (resolved path lands in the governed tree — case 35); a registered `workspace/<proj>` that is itself a symlink pointing OUT to a non-git dir is still gated via the raw-path check (case 39) — the Hakim security-review finding, genuinely fixed (I confirmed case 39 fails against the pre-fix hook and passes against the fix).
4. **Fail-closed on ambiguity.** Empty/unresolvable FILE_PATH (e.g. `python -c` with a dynamic path) is never exempted — it falls through to the ticket gate (case 38). Unresolvable `_resolve_real_path` output also skips the exemption block.

### Verification performed
- Checked out the PR HEAD in an isolated worktree; `bash -n` clean; `shellcheck -x` notes are all pre-existing (SC2016/SC2295 on lines 150/227/497/508, untouched by this diff) — none in the new block.
- Ran the hook's suite: **43 PASS / 0 FAIL**. Ran my own adversarial `..`/relative/symlink probes above — all correctly BLOCK or are provably non-writable.
- Full aggregate suite: 97/98; the one failure (`test_sync_codex_adapter.sh`) is environmental (hardcoded main-worktree path) and unrelated — the PR touches no codex/sync files, and GitHub CI is green.

### Issues Found
None blocking.

### Suggestions
- nit: consider adding an explicit `..`-traversal regression case (e.g. `src/subdir/../app.ts` BLOCKS) to the test file — the behaviour is correct today via canonicalization, but a dedicated case would pin it against future refactors of `_resolve_real_path`.

### Verdict
**APPROVED**

The exemption is tightly scoped to genuinely out-of-governance paths and I could not construct a writable in-repo path that gets exempted. The raw-AND-resolved dual containment check plus the git-repo backstop close the symlink holes in both directions, and the design is fail-closed on ambiguity.

---
Reviewed commit: `ccbbeaadaa329bf553276fea10913a75f88ca0eb`

atlas-apex (approved):
Approved (atlas-apex). Rex reviewed adversarially for a gate-bypass hole in the out-of-repo exemption — none found (.. traversal canonicalized, symlinks both directions blocked, fail-closed on unresolvable targets, 43/43 hook tests pass). CI green.

PR #897, opened 2026-07-15
fix(#888): exempt 'development' + config-driven trunk whitelist in validate-branch-name

### LINKED ISSUE #888: [Bug] validate-branch-name.sh trunk exemption misses 'development' (gitflow long form) — blocks mirror pushes to external remotes
## Summary

`validate-branch-name.sh` exempts trunk/integration branches by exact name — `main`, `master`, `develop`, `dev` — but not `development`, the long form used as the integration branch by many Bitbucket/gitflow remotes.

## Given / When / Then

- **Given** a managed project with an external mirror remote whose shared integration branch is named `development` (gitflow long form)
- **When** the agent pushes a refspec targeting `refs/heads/development` on that remote
- **Then** `validate-branch-name.sh` blocks the push as a naming-convention violation, even though the hook's trunk exemption is explicitly meant to allow shared integration branches (`develop` and `dev` pass)

## Repro

1. In a managed project, add a remote for an external mirror repo whose integration branch is named `development` (classic gitflow naming).
2. Push with a refspec targeting `refs/heads/development` on that remote.
3. The PreToolUse hook blocks: `BLOCKED: Branch 'development' doesn't follow naming convention.`

## Why this is a bug, not policy

The hook's own comment states the intent: "Allow trunk and shared integration branches." `development` is the same class of branch as `develop`/`dev` — it just uses the long form. The block fires on the push *destination* ref (via `_lib-extract-push-ref.sh`, which validates the dst of a `src:dst` refspec), so there is no compliant way to mirror work into such a remote: any refspec targeting `development` is rejected regardless of the local branch name.

Hit in practice while mirroring a milestone from a managed project's GitHub repo to a client's Bitbucket remote (client-mandated gitflow branch names — not something the adopter controls).

## Suggested fix

Add `development` to the exact-name allowlist in the trunk exemption (one word), or make the trunk list configurable via `.claude/project-config.json` (e.g. `.branch.trunk_whitelist`) the way `.branch.type_whitelist` already is — the config route also covers other client-mandated trunk names (`trunk`, `staging`, etc.).

## Bonus finding (same hook, separate defect)

While filing this issue, the hook also blocked a `gh issue create` command whose **heredoc body text** contained a literal push example — the hook greps the entire Bash command string for the push pattern and then ran the ref extractor against prose, extracting a branch name with a trailing backtick from the surrounding markdown. The push-command detection should anchor on the command position (not match inside heredoc/quoted bodies), or bail when the extracted ref contains characters outside `[A-Za-z0-9/_.-]`.

## Local workaround applied

Patched the fork's hook locally to add `development` to the allowlist (comment marks it as reported upstream) — will drop the local patch when the fix lands.

atlas-apex:
## Summary
- **`development` was silently blocked as a trunk branch** — `validate-branch-name.sh`'s trunk exemption used a hardcoded exact-name list (`main`/`master`/`develop`/`dev`) that missed `development`, the gitflow long form many Bitbucket/gitflow remotes use as their shared integration branch. A legitimate mirror push targeting that branch was rejected as a malformed feature branch even though the hook's own intent is to allow shared integration branches.
- **Trunk list is now config-driven** — the hook reads `.branch.trunk_whitelist` via `_lib-read-config.sh`, mirroring how `.branch.type_whitelist` is already read a few lines below it. Adopters with a client-mandated trunk name (e.g. `trunk`, `staging`) can add it via `.claude/project-config.json` without waiting on another upstream PR.
- **Fallback preserved** — when config is unavailable (jq missing, standalone install), the hook falls back to a hardcoded list that now includes `development` alongside the original four names, so the fix holds even without `jq`.
- **`project-config.defaults.json`** ships the new `.branch.trunk_whitelist` default (`main`, `master`, `develop`, `dev`, `development`).

## Testing
1. `bash .claude/hooks/tests/test_validate_branch_name_pushref.sh` — 39/39 pass, including new cases:
   - `git push origin development` → exempt (was: blocked)
   - a config-added custom trunk (`trunk`) → exempt
   - a real malformed feature branch (`feature/no-ticket-id`) → still blocked, even with a trunk override present
2. `bash -n .claude/hooks/validate-branch-name.sh` and `shellcheck` both clean.
3. `jq empty .claude/project-config.defaults.json` confirms the config edit is valid JSON.

**Follow-up (separate issue, not fixed here):** the issue's bonus finding — the push-command detection greps the whole Bash command string and can misparse a heredoc body containing a literal push example — is a distinct defect worth its own ticket. Not addressed in this PR.

Closes #888

---

## Glossary
| Term | Definition |
|------|------------|
| Trunk / integration branch | A long-lived shared branch (e.g. `main`, `dev`, `develop`) that individual feature branches merge into; exempt from the `{type}/{TICKET}-{description}` naming convention because it has no ticket of its own. |
| gitflow long form | Some gitflow/Bitbucket setups name their integration branch `development` instead of the shorter `develop`/`dev` — same role, different spelling. |
| Config-driven whitelist | A list read from `.claude/project-config.json` (via `_lib-read-config.sh`) instead of hardcoded in the shell script, so adopters can extend it without a code change. |
| Push-ref extraction | The hook resolves the branch to validate from the actual `git push` command's destination ref (via `_lib-extract-push-ref.sh`), not from local `HEAD` — needed for worktree-safe validation. |

### DIFF
--- a/.claude/hooks/tests/test_validate_branch_name_pushref.sh
+++ b/.claude/hooks/tests/test_validate_branch_name_pushref.sh
@@ -128,6 +128,15 @@ run_case "main is exempt as trunk" \
 run_case "dev is exempt as trunk (release-cut model)" \
   "git push origin dev" 0
 
+# ---- #888: 'development' (gitflow long form) trunk exemption -----------
+#
+# Bug: the original hardcoded exact-name list (main|master|develop|dev)
+# missed `development`, the long form used as the integration branch by
+# many Bitbucket/gitflow remotes — a legitimate mirror push to that branch
+# was wrongly blocked as a malformed feature branch.
+run_case "#888: development is exempt as trunk (gitflow long form)" \
+  "git push origin development" 0
+
 # Non-conforming push refs → block (the push-ref is now the source of truth).
 run_case "explicit ref: bogus-branch blocks" \
   "git push origin bogus-branch" 2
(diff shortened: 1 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #897

**Commit**: `3b7825bd9cff0ecddc1736425d56ccd35f883013`

### Summary
Makes `validate-branch-name.sh`'s trunk exemption config-driven via `.branch.trunk_whitelist` (mirroring the existing `.branch.type_whitelist` read) and adds `development` — the gitflow long form — to both the shipped default and the hardcoded fallback, fixing #888 where a legitimate push to `development` was wrongly blocked as a malformed feature branch.

### Checklist Results
- Security (trust-chain):    Pass
- Technical Decisions (AgDR):N/A — reuses the already-decided config-driven-whitelist pattern; no new library/architecture/pattern introduced
- Adopter Handbooks:         N/A — 3 always-load handbooks loaded (clean-architecture, migration-safety, commit-message-quality); none apply to a shell hook + JSON config diff, no findings

### Adversarial verification (trust-chain hook)
- **Config read follows the established pattern** — resolves `git rev-parse --show-toplevel`, sources `_lib-read-config.sh`, calls `config_get '.branch.trunk_whitelist[]'` — byte-identical shape to the `type_whitelist` read a few lines below. Double-sourcing the lib is harmless (per-process cached).
- **Fallback is safe and correct** — when config is unavailable (no jq / standalone install / shallow-merge drops the subtree), it falls back to `$'main\nmaster\ndevelop\ndev\ndevelopment'`, preserving today's four names plus the `development` fix.
- **Not a blanket bypass** — the match is `grep -qxF` (whole-line, fixed-string). Verified: an empty `CURRENT_BRANCH` does NOT match, a partial name like `developmentX` does NOT match, and a config-supplied value cannot inject regex. A genuinely malformed branch (`feature/no-ticket-id`, `bogus-branch`) still falls through to the ticket-shape check and blocks (exit 2) even with a custom trunk override present.
- **No regression** — `main`/`master`/`develop`/`dev` remain exempt via both config default and fallback. Shallow-merge case (adopter override sets only `type_whitelist`) is covered by the fallback, so trunk exemption survives.
- **Tests**: 39/39 pass against the PR HEAD in an isolated worktree, including the three key cases — `development` exempt, config-added `trunk` exempt, and `feature/no-ticket-id` still blocked with a trunk override present.
- **Lint**: `bash -n` and `shellcheck 0.11.0` clean on both hook and test; `jq empty` confirms the defaults JSON is valid and `trunk_whitelist` = `["main","master","develop","dev","development"]`.

### Issues Found
None blocking.

### Suggestions
- nit: `REPO_ROOT_FOR_TRUNK` (new block) and `REPO_ROOT` (existing type_whitelist block) both compute `git rev-parse --show-toplevel`. Minor DRY opportunity to hoist a single var — non-blocking, current form is clear.

### Verdict
**APPROVED**

---
Reviewed commit: `3b7825bd9cff0ecddc1736425d56ccd35f883013`

me2resh (approved):

mostiwheelietravel (commented):
Review posted to the wrong pull request due to a repository-resolution error. Please disregard.

PR #10, opened 2026-04-11
feat(#103): mechanical enforcement of ticket-first, auto-review, onboarding

atlas-apex:
## Summary

Fixes three SDLC-compliance gaps exposed by the first real test run of ApexStack on a sibling project:

1. **Ticket-first was skipped** — the agent coded, then created the ticket after.
2. **Code review was not auto-triggered** after `gh pr create`.
3. **No onboarding** for new repos — every session started cold without any discovery pass.

Root cause: the rules lived as **prose in `CLAUDE.md`, `.claude/rules/*.md`, and `workflows/*.md`** — advice the model drops under pressure. This PR moves the four highest-leverage rules into `.claude/hooks/`, which the harness executes mechanically.

Ticket lives in the ApexStack ops repo per the [tickets-live-in-apexscript-org exception](https://github.com/me2resh/apexscript-org/issues/103). Follow-up audit ticket: [#104](https://github.com/me2resh/apexscript-org/issues/104).

### New hooks (`.claude/hooks/`)

| Hook | Event | What it enforces | Rule traced back to |
|------|-------|------------------|---------------------|
| `require-active-ticket.sh` | `PreToolUse` on `Edit\|Write\|MultiEdit` | Blocks edits to code paths unless `.claude/session/current-ticket` exists. Exempts `.claude/`, `docs/`, `projects/*/docs/`, and any `*.md`. | `.claude/rules/workflow-gates.md` Pre-Build Gate |
| `auto-code-review.sh` | `PostToolUse` on `Bash(gh pr create *)` | Parses the PR URL, writes a pending-review marker, and nudges Claude to invoke the `code-reviewer` agent immediately. | `.claude/rules/pr-workflow.md` "After `gh pr create` → invoke Rex" |
| `block-unreviewed-merge.sh` | `PreToolUse` on `Bash(gh pr merge *)` | Blocks merge unless `.claude/session/reviews/<pr>-rex.approved` exists. Also verifies approved SHA matches HEAD. | `.claude/rules/workflow-gates.md` rule #5 |
| `onboarding-check.sh` | `SessionStart` | Warns if `.claude/session/onboarded` missing so Claude prompts `/onboard`. | Same principle as `.claude/rules/role-triggers.md` |

### New skills (`.claude/skills/`)

- **`/start-ticket <issue>`** — verifies via `gh issue view`, writes the marker, suggests a branch name. Accepts plain number (resolves against current repo's origin) or `<owner>/<repo>#<number>` for cross-repo cases.
[...]

PR #898, opened 2026-07-15
fix(#887): scope pr_base_repo to an explicit repo (no ambient gh resolution)

### LINKED ISSUE #887: [Bug] pr_base_repo() misresolves the host repo via unscoped gh pr view in fork checkouts whose gh default is the parent
## Affected
`.claude/hooks/_lib-review-markers.sh` → `pr_base_repo()` — and everything downstream of it: `tracker_review_submit` (used by `/code-review`, `/security-review`, `/design-review`) and `/approve-merge`'s host-repo resolution.

## Given / When / Then
**Given** an ops fork where `origin` is the fork and `upstream` is `me2resh/apexyard`, `gh` has **no default repo set**, and a PR is opened **against the fork's own `main`** (a same-repo fork PR).
**When** `/code-review <pr>` runs and calls `pr_base_repo <pr>` — whose body deliberately runs `gh pr view "$pr" --json url` **with no `--repo`** (per its `WHY-UNSCOPED` comment, assuming gh's ambient resolution returns the *base* repo).
**Then** gh's ambient resolution returns the **parent/upstream** `me2resh/apexyard`, so `pr_base_repo` returns `me2resh/apexyard` and the review is **posted to the public upstream PR of the same number** — expected: it should resolve to the fork the PR actually lives on.

## Repro
1. Fork `me2resh/apexyard`; set `origin` = fork, `upstream` = `me2resh/apexyard`; leave `gh` with no default repo.
2. Open a PR against the fork's own `main`.
3. Run `/code-review <pr>` (or trigger the auto-review). Rex's `tracker_review_submit` → `pr_base_repo` returns `me2resh/apexyard`.
4. Observe the review comment land on `me2resh/apexyard#<pr>` (the unrelated upstream PR of the same number) instead of the fork's PR.

## Framework version
v5.0.0

## Severity
major — reviews/merges silently target the wrong **public** repo (potential info disclosure + a valid review that never satisfies the fork's merge gate). Workaround: pin `--repo <fork>` on every host call and don't rely on `pr_base_repo`.

## Notes
The `WHY-UNSCOPED` comment assumes "gh's ambient base-repo resolution prefers the parent/upstream — i.e. the BASE repo." That holds for a PR filed *against upstream from a fork branch*, but for a **same-repo fork PR** the true base IS the fork, while gh's ambient default still resolves to the parent — so the helper returns the wrong repo precisely on the same-repo-fork path. The `hint`/fork fallback only fires when the URL can't be parsed, so it never corrects this case. Consider scoping the query to the hint repo first (and only falling back to unscoped), or resolving base via `gh pr view --repo <hint> --json ...` with the head repo as the authoritative starting point.

## Glossary
| Term | Definition |
|------|------------|
| ops fork | An adopter's fork of `me2resh/apexyard` used as the governance/ops repo |
| same-repo PR | A PR whose base and head branches both live in the fork (base != upstream) |
| ambient repo resolution | gh's behaviour of inferring the target repo from the working copy's remotes when `--repo` is omitted |

atlas-apex:
## Summary

- **`pr_base_repo()` no longer trusts gh's ambient repo resolution** — it previously ran `gh pr view "$pr" --json url` with no `--repo`, relying on gh's ambient default (which prefers the parent/upstream in a fork checkout) to land on the PR's base repo. That happened to work for a PR filed *from* a fork *to* upstream, but broke silently for a same-repo fork PR (opened against the fork's own `main`): ambient resolution still preferred the parent, so the call **succeeded with the wrong repo** — reviews posted to an unrelated public PR of the same number, and merge markers keyed to the wrong repo so a valid review never satisfied the fork's own merge gate.
- **The function now requires an explicit `<repo>` argument** — the repo the caller already knows hosts the PR — and scopes every `gh pr view` call to it. This turns "wrong repo" into a clean gh failure instead of a silently wrong answer: a PR object only resolves through its own base repo's API path, so scoping to the correct repo is authoritative, not a guess.
- **Updated all five callers** (`code-reviewer.md`, `solution-architect.md`, and the `approve-merge` / `approve-architecture` / `approve-design` skills) to stop deriving "the repo" via an unscoped `gh pr view --json headRepository` call — which read the wrong field (the PR's head/fork, not something to key a base-repo marker on) *and* was itself an ambient-resolved query. They now thread through the repo already known from context (an explicit skill/agent argument) or fall back to the current checkout's own `git remote get-url origin` — never a second ambient gh guess.
- **Rewrote `test_pr_base_repo.sh`** for the new required-repo signature, including a regression case that models the exact #887 scenario: an unscoped call would succeed with the wrong (parent) repo, but the fixed, always-scoped call returns the correct one.

## Testing

```bash
bash .claude/hooks/tests/test_pr_base_repo.sh   # 12/12 assertions pass
bash bin/run-hook-tests.sh                       # 102 PASS / 1 FAIL (pre-existing, unrelated)
```

The one failing suite (`test_sync_codex_adapter.sh`) fails identically on `upstream/dev` before this change (a stale local fixture path, `block-test.sh`, unrelated to review markers) — confirmed by stashing this diff and re-running the same test.

## Glossary

| Term | Definition |
|------|------------|
| Ambient repo resolution | `gh`'s behaviour of inferring the target repo from the working copy's remotes when `--repo` is omitted — in a fork checkout it prefers the parent/upstream |
| Same-repo fork PR | A PR whose base and head branches both live in the fork (base != upstream) |
| Base (host) repo | The repo a PR is numbered against and lives on — the canonical key for review/approval markers (#765) |
| Fail closed | When the repo passed to `pr_base_repo` is wrong, the scoped gh query cleanly fails and the function returns the caller's own (wrong) input rather than silently guessing at an unrelated repo |

Closes #887

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -57,8 +57,8 @@ Invoked when a PR is ready for review.
 
 ## Input
 
-- PR number or URL
-- Repository (any repository the user authorises)
+- PR number or URL — `{number}` below
+- Repository (any repository the user authorises) — `{repo}` below, threaded in by the invoking skill (`/code-review <pr> [repo]`). Never re-derive this from an unscoped `gh pr view {number} --json headRepository` call — see the marker section's `#887` note.
 
 ## Codebase grounding — prefer semantic search when available
 
@@ -661,18 +661,33 @@ MARKER_HOME="${OPS_ROOT:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}"
 # shellcheck source=/dev/null
 . "$MARKER_HOME/.claude/hooks/_lib-tracker.sh"
 mkdir -p "$MARKER_HOME/.claude/session/reviews"
-# Resolve the head (fork) repo — used ONLY as the hint/fallback for the base
-# resolver below. It is NO LONGER the marker key (that was the cross-fork bug —
-# see #765).
-PR_REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
+# Resolve the repo YOU already know hosts this PR — that is how you got
+# {number} in the first place (the `{repo}` input above, when the invoking
+# skill threaded it through). NEVER re-derive this via an unscoped
+# `gh pr view {number} --json headRepository` call: that call (a) reads the
+# WRONG field for this purpose (the PR's head/fork, not its base) and (b) is
+# itself an unscoped, ambient-resolved gh query — the exact class of bug #887
+# fixed (gh's ambient default prefers the parent/upstream, which is wrong for
+# a same-repo fork PR opened against the fork's own main). When {repo} wasn't
+# threaded through, fall back to the CURRENT checkout's own remote — a
+# deterministic, non-ambient source of truth — never to a second gh guess.
+REPO="{repo}"
+if [ -z "$REPO" ] || [ "$REPO" = "{repo}" ]; then
+  origin_url=$(git remote get-url origin 2>/dev/null)
+  origin_url="${origin_url%.git}"
+  REPO=$(printf '%s' "$origin_url" | sed -E 's#^(https?://[^/]+/|git@[^:]+:)##')
+fi
 
 # Resolve the PR/MR HOST (base) repo — the repo the PR lives on. It is BOTH where
 # the review must be POSTED (posting to the fork fails on a cross-fork PR: the PR
 # lives on the base) AND the canonical key for the approval marker. `pr_base_repo`
-# (in _lib-review-markers.sh) parses the PR URL — gh pr view has no baseRepository
-# field — and falls back to PR_REPO when base == head, so same-repo PRs are
-# unchanged.
-PR_HOST_REPO=$(pr_base_repo {number} "$PR_REPO")
+# (in _lib-review-markers.sh) now REQUIRES this explicit $REPO and scopes its gh
+# query to it — NEVER gh's ambient/parent-preferring default (#887). Scoping to
+# the repo you already know hosts the PR is authoritative, not a guess: a PR
+# object only resolves through its own base repo's API path, so passing the
+# wrong repo here fails closed instead of silently keying the marker on an
+# unrelated repo.
+PR_HOST_REPO=$(pr_base_repo {number} "$REPO")
 
 # Marker keyed on the BASE repo (#765) — this MATCHES what block-unreviewed-merge.sh
 # looks up: the gate keys on the merge command's --repo / API-path, which for a
--- a/.claude/agents/solution-architect.md
+++ b/.claude/agents/solution-architect.md
@@ -188,16 +188,23 @@ THREADED_REPO="{repo}"
 
 # 2a. Resolve the PR's BASE (host) repo ONCE — the canonical key for BOTH the
 # review posting AND the sign-off marker (#765). $THREADED_REPO (when
-# /design-review passed it, #687) is the hint; otherwise headRepository. Then
-# pr_base_repo (in _lib-review-markers.sh) resolves the base from the PR URL
-# (gh pr view has no baseRepository field) and falls back to the hint when
-# base == head — so same-repo PRs are unchanged.
+# /design-review passed it, #687) is the repo you already know hosts the PR;
+# otherwise fall back to the CURRENT checkout's own remote — a deterministic,
+# non-ambient source of truth. Do NOT fall back to an unscoped
+# `gh pr view {number} --json headRepository`: that call reads the WRONG field
+# (the PR's head/fork) AND is itself an ambient-resolved gh query that prefers
+# the parent/upstream — the exact class of bug that misresolved same-repo fork
+# PRs (#887). `pr_base_repo` (in _lib-review-markers.sh) now REQUIRES this
+# explicit repo and scopes its own gh query to it — never gh's ambient
+# default — so same-repo PRs still resolve unchanged.
 if [ -n "$THREADED_REPO" ]; then
-  HINT_REPO="$THREADED_REPO"
+  REPO_FOR_BASE="$THREADED_REPO"
 else
-  HINT_REPO=$(gh pr view {number} --json headRepository --jq '.headRepository.nameWithOwner' 2>/dev/null)
+  origin_url=$(git remote get-url origin 2>/dev/null)
+  origin_url="${origin_url%.git}"
+  REPO_FOR_BASE=$(printf '%s' "$origin_url" | sed -E 's#^(https?://[^/]+/|git@[^:]+:)##')
 fi
-PR_HOST_REPO=$(pr_base_repo {number} "$HINT_REPO")
+PR_HOST_REPO=$(pr_base_repo {number} "$REPO_FOR_BASE")
 REPO="$PR_HOST_REPO"      # the --repo flag for gh pr view when writing the marker SHA
 PR_REPO="$PR_HOST_REPO"   # marker key = the base repo (matches the gate's lookup)
 
--- a/.claude/hooks/_lib-review-markers.sh
+++ b/.claude/hooks/_lib-review-markers.sh
@@ -89,7 +89,7 @@ review_marker_path() {
   printf '%s/%s__%s-%s.approved' "$reviews_dir" "$safe_repo" "$pr" "$role"
 }
 
-# pr_base_repo <pr> [hint_repo]
+# pr_base_repo <pr> <repo>
 #
 # Echoes the PR/MR's BASE (host) repo as "owner/repo" — the repo the PR *lives
 # on* and is numbered against. This is the canonical key for approval markers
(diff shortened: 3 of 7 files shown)


atlas-apex (commented):
## Code Review: PR #898

**Commit**: `bd8525d3f9715ae3037f0345f4f15cc53fb0fcbd`

Reviewed adversarially as a trust-chain change (the review-marker / review-posting path). This is PR #898 itself a genuine cross-fork PR (head `atlas-apex/apexyard` → base `me2resh/apexyard:dev`), which makes it a live test of the very path it fixes.

### Summary
`pr_base_repo()` previously ran `gh pr view "$pr" --json url` **unscoped**, trusting gh's ambient parent-preferring resolution to land on the base repo. That held for a fork→upstream PR but broke silently for a **same-repo fork PR** (opened against the fork's own main): ambient still preferred the parent, so the call *succeeded with the wrong repo* — reviews posted to an unrelated PR of the same number, markers keyed on the wrong repo. This PR makes `<repo>` a **required** argument, always scopes `gh pr view --repo "$repo"`, and threads an explicit repo through all five callers (with a deterministic `git remote get-url origin` fallback, never a second ambient gh guess).

### Checklist Results
- Architecture & Design:      Pass — surgical, single-responsibility; writer/reader marker agreement is now structural, not coincidental
- Testing:                    Pass — strong, proven-load-bearing regression test
- Security:                   Pass — trust-chain path; fails **closed** (exit 1, no stdout) on a missing repo and **safe** (gate blocks) on a wrong one
- PR Description & Glossary:  Pass — narrative summary, glossary present, `Closes #887`
- Technical Decisions (AgDR): N/A — bug fix (reverses the #770 "unscoped" choice); trade-off fully documented in the PR body
- Adopter Handbooks:          N/A — no always-load handbook pattern-matches a shell-hook / agent-markdown diff

### Verification performed (against the four adversarial criteria)
1. **No silent wrong-repo path.** Confirmed on the PR tree: `<repo>` is required; missing → `return 1` + stderr, no stdout. The gh call is `gh pr view "$pr" --repo "$repo" ...` — no unscoped/ambient branch remains in the function.
2. **All `pr_base_repo` callers updated.** grep of the PR tree confirms the 5 callers (`code-reviewer.md`, `solution-architect.md`, `approve-merge`, `approve-architecture`, `approve-design`) each pass an explicit repo (threaded `{repo}`/`$REPO`) with the `origin`-remote fallback. No caller still feeds an unscoped `headRepository` result into `pr_base_repo`.
3. **Cross-fork path intact.** Verified against this PR itself: `pr_base_repo 898 me2resh/apexyard` resolves the base correctly. The `origin` fallback is sound when `origin == base` (the same-repo-fork #887 case, and upstream-clone reviews).
4. **Regression test genuinely models the bug.** Proven load-bearing: I reverted only `pr_base_repo` to the old unscoped body and re-ran the suite — case #10 (`#887 same-repo-fork-PR`) fails with the exact symptom (`got me2resh/apexyard`, wanted `atlas-apex/apexyard`), which the old always-resolving mock could never have caught. The new mock's unscoped branch (returning `wrong-url`) faithfully models "gh ambient succeeds with the wrong parent." 12/12 pass on the fixed impl. The full-suite `test_sync_codex_adapter.sh` failure is a stale `block-test.sh` fixture path, unchanged vs `dev` and unrelated.

### Issues Found
None blocking.

### Suggestions (non-blocking — fast-follow recommended)
- **`#887` is only *partially* closed for the review-posting path.** `.claude/agents/security-reviewer.md:143-144` (Hakim) resolves its post-target repo with the **exact unscoped pattern this PR removes** — `gh pr view {number} --json url` followed by an unscoped `--json headRepository` fallback — and it is *not* migrated here (unchanged vs `dev`, not a `pr_base_repo` caller, sources no marker lib because it writes no gate marker). Merge-gate correctness is unaffected (no marker), **but** on a same-repo fork PR Hakim can still post its security review to the wrong (parent) PR — the precise user-visible symptom #887 describes. Strongly recommend a fast-follow that routes Hakim's host-repo resolution through the new required-`<repo>` `pr_base_repo` (or the same explicit-`$REPO`/`origin` fallback), so all three review agents share one non-ambient resolution.
- Lower-stakes siblings on the same unscoped `headRepository` pattern, both pre-existing/unchanged: `.claude/hooks/_lib-extract-pr.sh:669` (extract fallback) and `.claude/hooks/warn-stale-review-markers.sh:105` (warn-only hook). Worth folding into the same fast-follow.
- **`origin` fallback edge (document, don't block).** When a repo is *not* threaded through, the `git remote get-url origin` fallback equals the **head** for a genuine cross-fork PR reviewed from the fork checkout — `pr_base_repo` then fails closed to the head, the marker keys on the head, and the merge gate diverges and **blocks** (fail-safe, never a bad merge). The old ambient code handled that edge; the new code needs the repo threaded. This is the correct trade-off (ambient can't be right for both same-repo-fork and cross-fork), and the skills do thread the repo — just worth a one-line note in the fallback comments that the fallback assumes `origin == base`.

### Verdict
**APPROVED**

The stated scope — `pr_base_repo` plus its five callers — is correctly and completely executed, well-tested with a regression case I verified is load-bearing, and strengthens the trust chain by turning a silently-wrong answer into a fail-closed/fail-safe one. The `security-reviewer.md` gap is a real, high-value finding but is pre-existing and out of this PR's surgical scope; tracked as a fast-follow rather than a blocker.

---
📌 Reviewed commit: `bd8525d3f9715ae3037f0345f4f15cc53fb0fcbd`

me2resh (approved):

mostiwheelietravel (changes requested):
## Code Review: PR #898

**Commit**: `36739fd3d37ecd6d2949e9160b27e7f48c84719a`

### Summary

This PR adds the optional-date, room, nationality, currency, and canonical-URL search flow for WheelieTravel 2.0. The domain bounds, active-hotel city lookup, localization coverage, focus restoration, and state round-trip are well structured. The focused independent test run passed all 50 tests. Three accessibility/mobile defects in the calendar still block approval because they affect the ticket's explicit screen-reader and mobile-first acceptance criteria.

### Checklist Results

- ✅ Architecture & Design: Pass. Domain logic remains infrastructure-free, the D1 lookup is one parameterized indexed query, and the shared state boundary follows the approved design.
- ❌ Code Quality: Fail. The date controls do not expose their selected values to assistive technology, and the calendar grid has invalid ARIA ownership.
- ❌ Testing: Fail. The 50 focused tests pass, but they do not assert the selected-date accessible name, valid grid row structure, or the 320px calendar layout.
- ✅ Security: Pass. No secret exposure, unsafe URL construction, or injection vector found; rate sessions are non-authorizing UUIDs and SQL values are bound.
- ✅ Performance: Pass with an advisory below. The database path is one query with the existing `(cityId, isPrimary)` and hotel-status indexes.
- ✅ PR Description & Glossary: Pass. The summary is narrative, ticket/design links are present, and the glossary explains the new concepts.
- ✅ Summary Bullet Narrative: Pass.
- ✅ Technical Decisions (AgDR): Pass. AgDR-0034 and the approved technical design cover preview routes, canonical search state, bounds, nationality/currency, room identity, and rate sessions; no dependency was added.
- ✅ Adopter Handbooks: Pass with advisories. Migration safety is N/A; all 14 required non-English Polly verdicts are present and PASS.
- ⚠ Fallow Static Analysis (JS/TS): N/A. The CLI is unavailable, so the required fail-soft path was used.

### Issues Found

1. **Blocking: selected dates are not exposed as the date buttons' accessible value** — `src/components/wheelietravel2/SearchForm.tsx:232`. The external `<label htmlFor="wheelietravel-2-checkin|checkout">` becomes the button's complete accessible name, so the visible child text (`Choose a date` or the selected formatted date at line 249) is suppressed from the accessibility tree. A screen-reader user hears only “Check-in, button” or “Check-out, button” and cannot review the selected range from the form. This fails the ticket's screen-reader-name/state requirement and WCAG 4.1.2. Include the current formatted date or empty-state text in an `aria-label`/description, and add a test that checks both buttons after selecting and rehydrating a range.

2. **Blocking: the calendar's ARIA grid does not contain the required row structure** — `src/components/wheelietravel2/DateDialog.tsx:252`. The `role="grid"` directly owns blank `<span>` elements and `role="gridcell"` buttons. ARIA grids must own `row` elements, which in turn own the grid cells; without that relationship, screen readers cannot reliably expose row/column navigation or position. Render the weeks as seven-cell `role="row"` groups (or use a native table), give blank cells valid semantics, and keep the current roving-tabindex behavior.

3. **Blocking: 44px calendar targets clip at a 320px mobile viewport** — `src/components/wheelietravel2/DateDialog.tsx:192` and `:274`. The full-screen body leaves 288px after `px-4`, while seven `min-w-11` day targets require 308px before any padding. The zero-minimum grid tracks therefore overflow and the dialog's `overflow-hidden` clips edge days. This violates WheelieTravel's mobile-first/no-accidental-overflow rules and can make dates unreachable on narrow phones. Adjust the narrow-viewport padding/layout while preserving 44px targets, then visually verify at 320px and 390px.

### Handbook Findings

⚠ **TypeScript Strict Mode** — `handbooks/language/typescript/strict-mode.md` *(discovery: path-convention)*
- `src/components/wheelietravel2/CityCombobox.tsx:69` casts an external JSON boundary with `as CitiesResponse` and no runtime shape check. The catch turns malformed data into a load error, so this is advisory, but a small type guard/schema would keep the boundary genuinely typed.

⚠ **Commit Message Quality** — `handbooks/general/commit-message-quality.md` *(discovery: path-convention)*
- Commit `36739fd` covers 3,900+ changed lines but its body contains only `Closes #885`; it does not explain why the optional-date/state approach was taken or note the remaining staging/design gates. Add rationale to substantive commit bodies on future changes.

### Suggestions

- In `SearchForm.tsx:95-107`, defer building and sorting roughly 400 `Intl.DisplayNames` nationality/currency options until `hasDates` is true. No-date SEO visits currently pay that synchronous mobile CPU cost even though both selects are hidden.

### Verdict

**CHANGES REQUESTED**

---
📌 Reviewed commit: `36739fd3d37ecd6d2949e9160b27e7f48c84719a`

mostiwheelietravel (commented):
Retraction: the changes-requested review submitted by `mostiwheelietravel` at 2026-07-16 16:22 UTC was posted to this pull request by an unscoped repository resolver. It does not apply to this PR, and I have not reviewed this PR. Please disregard and dismiss that review.

PR #906, opened 2026-07-15
fix(#899): add dotall flag to globToRegExp — close multi-line gate-bypass on adapters

### LINKED ISSUE #899: [Bug] derive-gates-core globToRegExp missing dotall flag — multi-line commits fail-open on gate adapters
## Affected
`harness-adapters/_shared/derive-gates-core.ts` — the `globToRegExp` helper (line ~176). Consumed by the **pi.dev** and **opencode** gate adapters.

## Given / When / Then
**Given** a session running under the pi.dev or opencode gate adapter, where Bash gates are matched by converting a permission glob (`Bash(<prefix> *)`) to a RegExp via `globToRegExp`.
**When** the model runs a standard multi-line commit whose message has a body (e.g. a conventional commit with `-m $'feat: x\n\n- detail\n\nCloses #1'`).
**Then** `gateMatchesToolCall` returns `false` — the derived regex `/^git commit .*$/` has no dotall (`s`) flag, so `.*` never crosses the embedded newlines — and the gate is **silently skipped (fail-open)**. Expected: the gate fires and `check-secrets.sh` / `verify-commit-refs.sh` / commit-format run.

## Repro
1. Generate the opencode or pi adapter (`derive-gates`).
2. In that harness, run a commit whose message has a body (multi-line — the conventional-commit norm).
3. Observe the `Bash(git commit *)` gate does not fire — a hardcoded secret in the staged diff is not caught.

## Framework version
v5.0.0

## Severity
major — a fail-open on the security-critical commit gates. Scoped to the opt-in pi/opencode adapters (Claude Code runs the bash hooks directly and is unaffected); workaround is single-line commit messages. Multi-line messages are the common case, so this is not an edge case.

## Notes
Surfaced by a high-effort code review while syncing an ops fork from v4.0.0 → v5.0.0. **Fix:** add the `s` flag when constructing the RegExp in `globToRegExp` so `.*` matches across newlines.

## Glossary
| Term | Definition |
|------|------------|
| `globToRegExp` | Helper converting a Bash permission glob (e.g. `git commit *`) into a RegExp matched against the tool-call command text. |
| dotall (`s`) flag | JS RegExp flag making `.` match newline characters. |
| fail-open | A gate that, on a match miss, ALLOWS the action instead of blocking — the unsafe default for a security gate. |
| gate adapter | Generated integration (pi.dev / opencode) enforcing apexyard's bash-hook gates inside a non-Claude-Code harness. |

atlas-apex:
## Summary
- **Closes a security fail-open (#899).** `globToRegExp` in `harness-adapters/_shared/derive-gates-core.ts` built `^…$` **without the `s` (dotall) flag**, so a **multi-line command** — the conventional-commit norm (subject + blank line + body) — didn't match the derived matcher. The gate consumer (`gateMatchesClaudeMatcher`) then **silently skipped** the gate instead of firing: multi-line `git commit` / multi-line `gh issue create --body` bypassed `check-secrets.sh`, `verify-commit-refs.sh`, commit-format, and the leak-protection / skill-required gates on the **pi & opencode adapters**. (Claude Code's native hooks are unaffected — this is adapter-only.)
- **One-line fix:** add the `s` flag → `new RegExp(\`^${escaped}$\`, "s")`.
- **3 regression tests** using real embedded newlines: the derived RegExp carries `s`; a multi-line `git commit` matches its glob; a multi-line `gh issue create --body` matches its glob. Verified fail-without / pass-with by temporarily reverting the fix.

## Testing
- shared core 19/19 · opencode 52/52 · pi 27/27 (2 LIVE credential tests skipped, unrelated). Reverting the `s` flag makes the 3 new tests fail; restoring makes them pass.

Closes #899

## Glossary
| Term | Definition |
|------|------------|
| dotall (`s`) flag | Makes regex `.` match newline characters |
| Fail-open | A gate that *allows* on a match-miss instead of blocking — the dangerous direction for a security gate |
| Adapter | The pi/opencode harness adapters that derive gate matchers from the framework's Bash-tool matchers |

### DIFF
--- a/harness-adapters/_shared/derive-gates-core.ts
+++ b/harness-adapters/_shared/derive-gates-core.ts
@@ -173,7 +173,13 @@ export function deriveGatesFromSettings(settings: RawSettings): GateDefinition[]
 /** Converts a Claude Code `if: "Bash(<glob>)"` glob (shell-style `*` wildcard) into an anchored RegExp. */
 export function globToRegExp(glob: string): RegExp {
   const escaped = glob.replace(/[.+^${}()|[\]\\]/g, "\\$&").replace(/\*/g, ".*");
-  return new RegExp(`^${escaped}$`);
+  // `s` (dotall) flag: without it, `.` never matches `\n`, so `.*` cannot
+  // cross the newlines in a multi-line command string (e.g. a conventional
+  // commit's `-m $'subject\n\nbody'`, or a `gh issue create --body` with a
+  // multi-line body). A gate whose glob fails to match a real multi-line
+  // command is a silent fail-open (see gateMatchesClaudeMatcher below) —
+  // #899.
+  return new RegExp(`^${escaped}$`, "s");
 }
 
 /**
--- a/harness-adapters/_shared/test/derive-gates-core.test.ts
+++ b/harness-adapters/_shared/test/derive-gates-core.test.ts
@@ -59,6 +59,43 @@ test("globToRegExp anchors — a glob must match the whole command, not a substr
   assert.ok(!re.test("echo 'not git commit at all but contains git commit inside'"));
 });
 
+// Regression test for #899: globToRegExp must carry the `s` (dotall) flag so
+// `.*` crosses embedded newlines. Without it, a multi-line command string
+// (the conventional-commit norm — a subject line plus a body) fails to
+// match its own `Bash(<prefix> *)` glob, and the consuming gate is silently
+// skipped (fail-open) instead of firing. Both assertions below FAIL if the
+// `s` flag is removed from globToRegExp's `new RegExp(...)` call.
+test("globToRegExp — the derived RegExp carries the dotall (s) flag, so it matches across embedded newlines (#899)", () => {
+  const re = globToRegExp("git commit *");
+  assert.ok(re.flags.includes("s"), "globToRegExp's RegExp must carry the `s` (dotall) flag");
+});
+
+test("globToRegExp matches a multi-line `git commit -m` command (conventional-commit subject + body) — #899", () => {
+  const re = globToRegExp("git commit *");
+  // Real embedded newlines (template literal), not escaped "\n" text — this
+  // is what a shell actually hands the hook for `git commit -m $'subject\n\nbody'`.
+  const multiLineCommit = `git commit -m $'feat: add widget
+
+- detail one
+- detail two
+
+Closes #1'`;
+  assert.ok(re.test(multiLineCommit), "a multi-line git commit command must match its own Bash(git commit *) glob");
+});
+
+test("globToRegExp matches a multi-line `gh issue create --body` command — #899", () => {
+  const re = globToRegExp("gh issue create *");
+  const multiLineIssueCreate = `gh issue create --title "Bug: X" --body "Given a user
+When they do Y
+Then Z happens
+
+Repro steps here."`;
+  assert.ok(
+    re.test(multiLineIssueCreate),
+    "a multi-line gh issue create --body command must match its own Bash(gh issue create *) glob",
+  );
+});
+
 // ---------------------------------------------------------------------
 // deriveGatesFromSettings — keyed by raw Claude matcher token, not by any
 // runtime's tool id.


atlas-apex (commented):
## Code Review: PR #906

**Commit**: `c1536e3b4c385dd18b307f393b588c09d6e92b51`

### Summary
Adds the `s` (dotall) flag to `globToRegExp`'s `new RegExp(\`^${escaped}$\`)` in `harness-adapters/_shared/derive-gates-core.ts`, plus 3 regression tests. Closes a real security fail-open (#899): without dotall, `.*` cannot cross embedded newlines, so a multi-line command (conventional-commit subject+body, or a `gh issue create --body` with a multi-line body) failed to match its own `Bash(<prefix> *)` glob. `gateMatchesClaudeMatcher` then silently *skipped* the gate rather than firing it — meaning multi-line commands bypassed secrets/commit-ref/leak/skill gates on the pi & opencode adapters.

### Verification against the four review points

1. **Does `s` genuinely fix the multi-line match? — YES.** The derived pattern is `^git commit .*$`. Without `s`, JS `.` never matches `\n`, so `.*` halts at the first newline and the anchored `$` can never reach true end-of-input → no match. With `s`, `.*` consumes the newlines and the command matches. Confirmed against the base version (`new RegExp(\`^${escaped}$\`)`, no flag) fetched from `dev`.

2. **Over-match risk? — NONE.** This is the correct nuance: in JavaScript, the `$` end-anchor is loosened *only* by the `m` (multiline) flag, **not** by `s`. `s` changes solely whether `.` matches `\n`; the `^`/`$` anchors stay bound to true start/end of the whole string (no `m` here). So single-line matcher behaviour is byte-for-byte unchanged, and the anchoring still prevents substring matches (the existing "anchors" test at line 59 still holds). The only behavioural change is that multi-line commands now correctly match — and that change moves the gate in the **fail-safe** direction (gate *fires* rather than silently skips). No unrelated single-line matcher can newly false-match.

3. **Do the 3 tests use REAL embedded newlines and fail without the fix? — YES.**
   - Test 1 asserts `re.flags.includes("s")` directly — fails immediately if the flag is dropped.
   - Tests 2 & 3 build the command strings as **template literals with actual line breaks** (not escaped `"\n"` text), matching what a shell hands the hook for `git commit -m $'subject\n\nbody'` and a multi-line `--body`. Both `.test()` assertions return `false` without `s` (verified by the change to the anchored regex) and `true` with it. The PR body also documents a manual revert-and-fail check.

4. **Any other regex in the file needing the same flag? — NO.** I checked every RegExp in the module:
   - `extractHookRelativePath` — `/\.claude\/hooks\/([\w.-]+\.sh)\b/`: unanchored substring search over the framework-generated (single-line) settings.json hook `command`; `[\w.-]+` excludes `/` and `\n`, so it needs no dotall.
   - `extractCommandGlob` — `/^Bash\((.*)\)$/`: parses a settings.json `if` predicate, which is always a single-line glob string — no newlines to cross.
   - The escaping `replace`s and `/\.sh$/` are not command-matchers.
   Only `globToRegExp` matches against the (potentially multi-line) *command* string, so it is the only place the flag is load-bearing.

### Checklist Results
- Code Quality:               Pass (one-line fix, well-commented with the #899 rationale inline)
- Testing:                    Pass (3 targeted regression tests, real newlines, fail-without/pass-with)
- Security:                   Pass — this *is* a security fix; direction is fail-open → fail-safe
- Performance:                Pass (n/a)
- PR Description & Glossary:  Pass (Summary/Testing/Glossary/`Closes #899` all present, narrative bullets)
- Technical Decisions (AgDR): N/A — bug fix, no new dependency/framework/pattern/architecture decision
- Adopter Handbooks:          N/A — no handbook violations (see below)
- Fallow Static Analysis:     N/A — `fallow` CLI unavailable in this environment

### Handbook Findings
Loaded (path-convention): `handbooks/architecture/clean-architecture-layers.md`, `handbooks/general/commit-message-quality.md`, `handbooks/architecture/migration-safety.md`, `handbooks/language/typescript/strict-mode.md` (TS diff). No violations: the diff introduces no `any` / `@ts-ignore` / unchecked `as`, no layer-boundary crossing, no migration paths.

### Issues Found
None.

### Suggestions
- `nit` (non-blocking): `extractCommandGlob`'s `^Bash\((.*)\)$` is correct today because settings `if` predicates are always single-line; if a future glob ever legitimately spanned lines it'd silently drop. Not worth changing now — flagged only for completeness since it was in scope of the "any other regex" question.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `c1536e3b4c385dd18b307f393b588c09d6e92b51`

atlas-apex (commented):
## Security Review: PR #906

**Commit**: `c1536e3b4c385dd18b307f393b588c09d6e92b51`

▸ Activating Hakim (Security Auditor) for PR #906 (trigger: diff touches the trust chain — gate-matching layer of the pi/opencode adapters, #777)

### Summary
This is a one-character security fix (`new RegExp(\`^${escaped}$\`, "s")`) to `globToRegExp` in `harness-adapters/_shared/derive-gates-core.ts`, closing the multi-line gate fail-open (#899). I traced the data flow, reasoned about the anchoring semantics of the added flag, and swept the adapters for sibling missing-flag / fail-open shapes.

### Checklist Results
- Auth & Authorisation:   Pass (this *restores* enforcement on adapters)
- Fail-safe posture:      Pass

### 1. The fix genuinely closes the bypass — CONFIRMED
`globToRegExp` is the **only** place a runtime, agent-controlled command string is matched — it's called at `gateMatchesClaudeMatcher` (derive-gates-core.ts:197) with the live `command`. The two other `.match()` calls in the module operate on **trusted static `settings.json` config**, not runtime input:
- L103 `command.match(/\.claude\/hooks\/…\.sh/)` — extracts the hook path from the settings-authored command wrapper (substring search, unaffected by newlines in the runtime command).
- L110 `ifPredicate.match(/^Bash\((.*)\)$/)` — parses the single-line `if:` predicate from `settings.json` (never multi-line).

So the dotall bug lived at exactly one chokepoint, and that chokepoint is what's fixed. Adding `s` lets `.*` cross the embedded newlines in a conventional-commit body / multi-line `--body`, so the derived matcher now matches the real command and the consuming gate **fires** instead of being silently skipped. `check-secrets.sh`, `verify-commit-refs.sh`, commit-format, leak-protection, and skill-required gates all re-arm on multi-line commands on the pi & opencode adapters. The 3 regression tests use **real embedded newlines** (template literals, not escaped `\n` text) and assert both flag-presence and actual multi-line matching — they fail if the flag is reverted, so the fix is pinned.

### 2. Adding `s` does NOT open a different hole — CONFIRMED SAFE
The `s` (dotall) flag changes **only** the behaviour of `.` (now matches `\n`). It does **not** touch `^`/`$` anchoring — that is governed by the `m` (multiline) flag, which is **absent**. So `^…$` still anchors to the absolute start/end of the whole command string, and the literal prefix before the first `*` remains a hard anchor: `^git commit .*$` still *requires* the string to start with `git commit ` — dotall cannot make `git commit *` match `rm -rf /`. No cross-matcher contamination is possible.

The only broadened behaviour is `.*` spanning newlines, which for these security gates is the **fail-closed** direction: a matcher that matches *more* commands causes the gate to be *evaluated* more often (and the gate hook itself decides block/allow). More gate evaluation can only yield *more* potential blocks, never fewer. Over-match here is fail-safe, not a new hole.

### 3. The fail-open direction is correct as-is; matching-correctly is the right fix
The validator flagged the dispatch-loop match-miss path (`opencode/src/gate-dispatcher.ts:224` / `pi/src/gate-dispatcher.ts:341`): `if (!gateMatchesToolCall(...)) continue;` — a match-miss skips the gate. **This fail-open-on-miss is inherent and correct**, not a hardenable weakness: a gate wired to `git commit *` *must not* run for `ls -la`. You cannot "fail closed on match-miss" without firing every gate on every command (check-secrets on `gh pr merge`, etc.) — the matcher is precisely the mechanism that selects relevant gates. So the genuine risk lives entirely in matcher **correctness** (false negatives), which is exactly the bug class this PR fixes. Hardening the dispatch direction would be wrong; fixing the matcher is the right layer.

Worth noting the dispatcher **already fails CLOSED where it matters**: a *matched* gate that cannot be *evaluated* (spawn error / `ETIMEDOUT` / `ENOBUFS`) returns `block: true` (`runGateHook`, lines 244–251). Posture is correct on both axes — fail-open only for irrelevant (unmatched) gates, fail-closed for matched-but-unrunnable ones.

### Residual gap (LOW, non-blocking — out of scope for this hotfix)
`globToRegExp`'s escape set `[.+^${}()|[\]\\]` escapes regex metachars but **not `?`**. A future `Bash(<glob>)` matcher containing `?` would compile `?` as a regex quantifier ("0-or-1 of preceding"), producing a mangled matcher — the same false-negative class as this bug. **Not exploitable today**: I verified all 46 `Bash(<glob>)` predicates in `.claude/settings.json` use only a literal prefix + trailing `*` (no `?`, no char-classes). Consider adding `?` to the escaped set (or a one-line comment noting the glob dialect is `*`-only) to future-proof the matcher-correctness invariant this PR just tightened.

### Verdict
**APPROVED** — closes the fail-open at the correct layer, the added flag is anchoring-neutral and cannot broaden matching in an unsafe direction, and the dispatcher's fail-closed posture on evaluation failure is intact. One LOW residual (unescaped `?`) noted for future-proofing, not blocking.

---
📌 Reviewed commit: `c1536e3b4c385dd18b307f393b588c09d6e92b51`

me2resh (approved):

Issue #679, opened 2026-06-21
[Chore] Scorecard: set publish_results: false to stop the action failing the run


atlas-apex:
## Driver
The `Scorecard supply-chain security` workflow fails on push to `main` ("Failing after 33s") because the `publish_results: true` step can't publish to the OpenSSF API, which exits the whole action non-zero. Scorecard is a posture scanner (non-gating, dropped from required checks), so a failing run is just noise on every main push.

## Scope
- In `.github/workflows/scorecard.yml`, set `publish_results: false`.
- Trade-off: the public Scorecard README badge will no longer auto-update. (If the badge is wanted later, the alternative is adding a `SCORECARD_TOKEN` PAT secret with `repo`+`read:org` and keeping publish on.)

## Acceptance Criteria
- [ ] `publish_results: false` in scorecard.yml.
- [ ] The Scorecard workflow no longer fails the run on the publish step (results still upload to the Security > Code scanning tab via the SARIF upload).

## Glossary
| Term | Definition |
|------|------------|
| publish_results | OSSF Scorecard option that publishes the score to the public OpenSSF API (drives the README badge). |

atlas-apex:
Fixed by #680 (merged to dev) — `publish_results: false` is set in scorecard.yml. Closing the stale tracking issue; the merge didn't auto-close it (Refs, not Closes).

PR #908, opened 2026-07-16
fix(#907): gate scorecard.yml to the canonical repo (skip on forks)

### LINKED ISSUE #907: [Bug] scorecard.yml fails on every adopter fork — gate to the canonical repo
## Given / When / Then
- **Given** the framework ships `.github/workflows/scorecard.yml` (OSSF Scorecard) with triggers `push` / `schedule` / `branch_protection_rule` and **no repo guard**,
- **When** an adopter forks apexyard (every adopter does), the workflow runs on their fork,
- **Then** the "Run analysis" step **fails on the fork** — Scorecard is designed to analyze and publish results for the *canonical* public project only; on a fork `publish_results: true` can't publish to the OSSF registry (only the tracked upstream can), and the fork lacks the branch-protection/security metadata Scorecard expects. Result: every adopter fork carries a persistent red ❌ that is NOT a real security regression.

## Repro
Fork `me2resh/apexyard`, push to the fork's default branch (or wait for the weekly schedule) → the "Scorecard supply-chain security" job fails on the fork.

## Fix shape
Gate the Scorecard job to the canonical repo:

```yaml
jobs:
  analysis:
    if: github.repository == 'me2resh/apexyard'
```

Use `github.repository` (always set, on every event type) rather than `github.event.repository.fork` (unreliable/absent on `schedule` runs). Net effect: runs only on `me2resh/apexyard`; **skips** (not fails) on every fork.

## Acceptance Criteria
- [ ] `scorecard.yml`'s job is gated with `if: github.repository == 'me2resh/apexyard'`
- [ ] Runs on the canonical repo; skips (not fails) on forks and on scheduled runs on forks
- [ ] No other behavior change to the scan on the canonical repo

## Glossary
| Term | Definition |
|------|------------|
| OSSF Scorecard | Supply-chain security scorer that publishes results for a tracked public project |
| Fork guard | A job-level `if:` that restricts a workflow to the canonical repo |

atlas-apex:
## Summary
- **Gated the Scorecard analysis job to the canonical repo** — added `if: github.repository == 'me2resh/apexyard'` at the job level (sibling of `runs-on`). Every adopter fork was previously running (and failing) this job on `push`/`schedule`/`branch_protection_rule`, because OSSF Scorecard can only publish results for the tracked canonical public project — the fork isn't wrong, the job just can't succeed there. That produced a persistent, non-actionable red ❌ on every fork's Actions tab that looked like a security regression but wasn't.
- **Used `github.repository`, not `github.event.repository.fork`** — the latter is absent on `schedule`-triggered runs, and this workflow's primary trigger is a weekly cron (`19 4 * * 1`). `github.repository` is populated on every event type the workflow listens for, so it's the only condition that reliably distinguishes canonical-vs-fork across all three triggers.
- **No other behavior change** — the scan itself, its permissions, and its steps are untouched; on `me2resh/apexyard` the job runs exactly as before.
- **Added AgDR-0096** recording the options considered (the two `if:` conditions, deleting the workflow per-fork, and doing nothing) and why `github.repository` won.

## Testing
1. Read the current `.github/workflows/scorecard.yml` and confirmed the job name is `analysis`.
2. Added `if: github.repository == 'me2resh/apexyard'` as a direct sibling of `runs-on` under the `analysis` job (correct indentation level — job-scoped, not step-scoped).
3. Validated the YAML two ways:
   - `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/scorecard.yml'))"` — loads cleanly
   - Asserted programmatically that `jobs.analysis.if == "github.repository == 'me2resh/apexyard'"` — passed
4. Confirmed via `git diff` that the change is a single added line with zero other diff noise.
5. Reasoning for `github.repository` over `github.event.repository.fork`: the workflow's primary trigger is `schedule` (weekly cron) — GitHub does not populate `github.event.repository` (and therefore `.fork`) on scheduled runs, only on `push`-style events carrying a real webhook payload. `github.repository` (`owner/repo` string) is set unconditionally by the Actions runtime regardless of trigger type, so it's the only condition guaranteed to evaluate correctly on all three of this workflow's triggers (`push`, `schedule`, `branch_protection_rule`).

Closes #907

---

## Glossary
| Term | Definition |
|------|------------|
| OSSF Scorecard | Open Source Security Foundation's supply-chain security scorer — assesses branch protection, token permissions, pinned actions, dangerous workflow patterns, etc., and publishes a score/badge for a tracked public repo |
| Fork guard | A job-level `if:` condition that restricts a workflow (or job) to running only on the canonical repo, skipping (not failing) on forks |
| `github.repository` | A GitHub Actions context variable set to `owner/repo` on every triggered run, regardless of event type |
| `github.event.repository.fork` | A boolean nested in the webhook event payload; only present on events that carry a full repository payload (e.g. `push`) — absent on `schedule` runs |

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -24,6 +24,7 @@ permissions:
 jobs:
   analysis:
     name: Scorecard analysis
+    if: github.repository == 'me2resh/apexyard'
     runs-on: ubuntu-latest
     permissions:
       security-events: write   # upload results to code-scanning
--- a/docs/agdr/AgDR-0096-scorecard-fork-guard.md
+++ b/docs/agdr/AgDR-0096-scorecard-fork-guard.md
@@ -1,0 +1,36 @@
+# Gate `scorecard.yml` to the Canonical Repo
+
+> In the context of `.github/workflows/scorecard.yml` (OSSF Scorecard) running unconditionally on `push` / `schedule` / `branch_protection_rule`, facing the fact that every adopter fork inherits the workflow and shows a persistent, non-actionable red ❌ (Scorecard can only publish results for the tracked canonical project), I decided to add a job-level `if: github.repository == 'me2resh/apexyard'` guard to achieve a clean CI signal on forks without weakening the scan on the canonical repo, accepting that adopters who want their own Scorecard run must explicitly re-enable the job for their fork.
+
+## Context
+
+`scorecard.yml` runs OSSF Scorecard's supply-chain security analysis and publishes results to the Security > Code scanning tab (see #518). It has no repo guard, so cloning/forking apexyard (which every adopter does per the framework's fork-based distribution model) inherits the workflow verbatim. On a fork:
+
+- `publish_results` cannot succeed — the OSSF badge API only accepts results for the tracked, canonical public repo
+- The fork lacks the branch-protection / security metadata Scorecard expects, so the "Run analysis" step fails outright
+
+The net effect is a red ❌ on every adopter fork's Actions tab that looks like a security regression but is actually structural noise — a false signal a new adopter has no way to distinguish from a real one.
+
+## Options Considered
+
+| Option | Pros | Cons |
+|--------|------|------|
+| `if: github.repository == 'me2resh/apexyard'` | Always set on every trigger this workflow uses (`push`, `schedule`, `branch_protection_rule`); simple string equality; skips (not fails) on forks | Requires the canonical repo's exact `owner/name` hardcoded in the workflow |
+| `if: ${{ !github.event.repository.fork }}` | Reads as "not a fork" | `github.event.repository` is **absent on `schedule` runs** — this workflow's primary trigger is a weekly cron, so the condition would evaluate to a falsy/undefined lookup unreliably rather than deterministically skip |
+| Delete the workflow from `.github/workflows/` in each fork post-clone | No conditional logic needed | Not automatable at fork-creation time; every adopter would have to remember to delete it manually, and re-adding it upstream would silently reintroduce the failure for existing forks that already deleted their copy |
+| Leave as-is, document the failure as expected/ignorable | Zero code change | A red ❌ that adopters are told to ignore erodes trust in the rest of the CI signal — exactly the failure mode this ticket exists to fix |
+
+## Decision
+
+Chosen: **`if: github.repository == 'me2resh/apexyard'`**, because `github.repository` is populated on every event type this workflow listens for — including `schedule`, where `github.event.repository.fork` does not exist — making it the only reliable, deterministic way to restrict the job to the canonical repo across all three triggers.
+
+## Consequences
+
+- The Scorecard job now **skips** (not fails) on every fork, on all three triggers (`push`, `schedule`, `branch_protection_rule`)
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #908

**Commit**: `c4a94931e86fef0ca1171d1cb5ee6a322e6cbb03`

### Summary
Gates the OSSF Scorecard workflow to the canonical repo by adding a single job-level `if: github.repository == 'me2resh/apexyard'` to the `analysis` job, so adopter forks stop showing a persistent, non-actionable red ❌ (Scorecard can't publish/complete on a fork). Ships with AgDR-0096 documenting the decision. Closes #907.

### Checklist Results
- ✅ Architecture & Design:      Pass — job-level guard, no coupling changes
- ✅ Code Quality:               Pass — one added line, zero collateral diff
- ✅ Testing:                    Pass — YAML validated + job-scope assertion documented
- ✅ Security:                   Pass — no permission/scope change; scan intact on canonical repo
- ✅ Performance:                N/A
- ✅ PR Description & Glossary:  Pass — Summary/Testing/Glossary + `Closes #907` all present
- ✅ Summary Bullet Narrative:   Pass — bullets are narrative (what + why)
- ✅ Technical Decisions (AgDR): Pass — AgDR-0096 linked, covers all four options
- ✅ Adopter Handbooks:          N/A — no language files; CI YAML not in scope of loaded handbooks

### Verification against the requested checks
1. **Job-level, correct indentation** — `if:` sits at 4-space indent as a direct sibling of `name` / `runs-on` / `permissions` under `jobs.analysis`, i.e. job-scoped, NOT step-scoped. Valid YAML (loads cleanly). ✅
2. **`github.repository` is the right condition** — the workflow triggers on `branch_protection_rule`, `schedule` (weekly cron `19 4 * * 1`), and `push`. `github.repository` is populated by the Actions runtime on all three, whereas `github.event.repository.fork` is absent on `schedule` runs. Correct call, and the AgDR options table calls this out explicitly. ✅
3. **Nothing else changes on the canonical repo** — on `me2resh/apexyard` the `if` evaluates true, so the job, its `security-events: write` permission, and all four steps (checkout, run analysis, upload artifact, upload-to-code-scanning) run exactly as before. Forks now *skip* (not fail). ✅
4. **AgDR well-formed** — follows the template (context / options / decision / consequences / artifacts); options table captures `github.repository` vs `github.event.repository.fork` vs per-fork delete vs do-nothing, with sound justification for the winner. ✅

### Issues Found
None blocking.

### Suggestions
- nit: AgDR § Context lists "`publish_results` cannot succeed" as a fork failure cause, but `publish_results: false` is already set in the workflow (per #679), so that specific sub-cause is stale. The core rationale still holds (a fork lacks the branch-protection/security metadata the analysis step expects, and can't upload code-scanning results the same way), so this is cosmetic — worth a one-line tweak on a future edit, not a blocker.
- suggestion: consider a short inline `# skip on forks — Scorecard only completes on the canonical repo (#907)` comment beside the `if:` so a future reader doesn't need to open the AgDR to understand the guard. Optional.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `c4a94931e86fef0ca1171d1cb5ee6a322e6cbb03`

me2resh (approved):

PR #926, opened 2026-07-16
fix(#886): judge all bash write targets, not just the first

### LINKED ISSUE #886: [Task] Harden bash write-target extraction — judge ALL targets, not just the first (pre-existing, surfaced by #885 review)
## Driver
Found during the #885 review (Rex, confirmed pre-existing on `dev` by running the base hook): `_lib-detect-bash-write.sh` returns only the FIRST write target of a Bash command. A command writing to an out-of-repo target first and an in-repo target second is judged solely on the first → the gate exempts a command that also writes inside a governed tree. Pre-existing, best-effort-by-design, but now that #883 widened the exemption surface it's worth hardening.

## Scope
Teach the extractor (or the gate's consumption of it) to consider ALL detected write targets: if ANY resolved target lands inside a governed boundary, the command is gated. Preserve fail-closed behaviour for unparseable commands.

## Acceptance Criteria
- [ ] Multi-target Bash command (out-of-repo + in-repo) is GATED without a marker.
- [ ] Single out-of-repo target remains exempt; single in-repo remains gated.
- [ ] Full hook suite green + new cases.

## Glossary
| Term | Definition |
|------|------------|
| Write-target extraction | `_lib-detect-bash-write.sh`'s parse of redirections/tee/sed -i etc. into target paths. |

atlas-apex:
## Summary

- **Extended the bash write-target extractor to return ALL targets, not just the first** — `_lib-detect-bash-write.sh` gains `bash_extract_write_targets` (plural), which splits a Bash command on top-level separators (`&&`, `||`, `;`, `|`) and captures every redirection and every `tee` operand across those segments, instead of the existing `bash_extract_write_target`'s first-match-wins walk. The singular function is kept unchanged for backward compatibility with its existing call sites and tests.
- **Closed the gate-bypass surface confirmed by the #885 review** — `require-active-ticket.sh` and `require-migration-ticket.sh` used to extract only the FIRST write target from a Bash command and judge the whole command on that one path alone. A command naming an out-of-governance target first and an in-repo (or migration-shaped) target second was silently exempted, because the gate never looked past target #1. Both hooks now evaluate every extracted target: `require-active-ticket.sh` uses AND semantics (the command only passes if every target independently clears the gate — variable exemptions, out-of-governance exemption, path-prefix exemptions, then the ticket-marker lookup); `require-migration-ticket.sh` scans every target for the first one that's both non-meta-exempt and migration-shaped, then gates on that one.
- **Refactored `require-active-ticket.sh`'s single-target pipeline into `_ratc_evaluate_target()`** so it can be invoked once per extracted target with fully isolated local state (`REPO_ROOT`, `OPS_ROOT`, `PROJECT`, markers, the `#883`/`#885` out-of-governance symlink-containment checks) — nothing leaks between targets, and the existing single-target behaviour is unchanged when a command names only one write target.
- **New regression tests prove the fix is load-bearing** — the added multi-target cases (semicolon-chained, `&&`-chained, `tee` with multiple file operands) fail against the pre-fix hooks and pass against the fix; single-target cases are unchanged, confirming no regression to the existing exemption behaviour.

## Testing

1. `bash -n` on every changed hook file — all pass.
2. `.claude/hooks/tests/test_detect_bash_write.sh` — 108/108 pass (11 new `bash_extract_write_targets` cases added, covering multi-target chains, `tee` with multiple operands, dedup, and the documented interpreter misses).
3. `.claude/hooks/tests/test_require_active_ticket_bash.sh` — 48/48 pass (5 new `#886` cases: multi-target blocked w/o ticket, multi-target allowed with ticket, single out-of-repo target still exempt, single in-repo target still blocked, `tee` with mixed operands blocked).
4. `.claude/hooks/tests/test_require_migration_ticket.sh` — 21/21 pass (2 new `#886` cases covering a non-migration target followed by a migration-shaped target, with and without an active ticket).
5. Verified all five new tests are load-bearing: stashing the hook fixes (keeping only the new tests) reproduces exactly the expected failures against the pre-fix hooks, confirming the tests actually exercise the bypass.
6. Pre-existing, unrelated test failures (`test_workspace_tracker_resolution.sh`, `test_ops_root.sh`, two cases in `test_split_portfolio_v2_migration.sh`) were confirmed to fail identically with this branch's changes stashed out — not caused by this PR.

Closes #886

---

## Glossary

| Term | Definition |
|------|------------|
| Write-target extraction | `_lib-detect-bash-write.sh`'s parse of a Bash command's redirections/`tee`/`sed -i`/etc. into concrete file-path targets, used by the ticket and migration gates to decide whether a write needs to be judged. |
| Gate-bypass surface | A way a mechanically-enforced rule (here, the ticket-first gate) can be satisfied on paper while the underlying intent (every governed write requires an active ticket) is violated. |
| AND semantics (ticket gate) | The command as a whole only passes `require-active-ticket.sh` if EVERY extracted write target independently clears the gate — a single exempt target no longer excuses the rest. |
| Out-of-governance exemption (#883/#885) | The existing symlink-aware, raw-and-resolved containment check that decides whether a write target sits entirely outside the ops fork and every registered workspace project (and therefore doesn't need a ticket). Reused unchanged per-target by this PR. |

### DIFF
--- a/.claude/hooks/_lib-detect-bash-write.sh
+++ b/.claude/hooks/_lib-detect-bash-write.sh
@@ -119,8 +119,21 @@ _bdw_starts_with_git_subcommand() {
 # ------------------------------------------------------------------------------
 
 # 1. Redirection.
+#
+# The leading-context class `[^|<&]` excludes a `>` that's part of `2>&1`
+# fd-dup or immediately follows a pipe/heredoc marker. But `[^|<&]` REQUIRES
+# a character before `>` to exist at all — a command (or, after #886's
+# segment split, a SEGMENT) that BEGINS with `>` has no preceding character,
+# so the original pattern silently failed to match at position 0. That's
+# the exact shape a no-space chained write produces: splitting
+# `echo a > /tmp/ok;> .gitignore` on `;` yields a second segment of
+# `> .gitignore` — starting with `>` — and the un-anchored pattern dropped
+# it entirely (apexyard#886 Hakim security-review finding, PR #926).
+# `(^|[^|<&])` adds "start of string/segment" as an equally-valid leading
+# context, closing that hole while leaving the `2>&1` / `<<EOF` exclusions
+# intact (neither of those ever begins a segment with a bare `>`).
 _bdw_match_redirection() {
-  echo "$1" | grep -qE '[^|<&]>>?[[:space:]]+[^[:space:]&|;]+'
+  echo "$1" | grep -qE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+'
 }
 
 # 2. tee.
@@ -401,8 +414,15 @@ bash_extract_write_target() {
 
   # Output redirection: capture the first target after > or >>.
   # Strip leading number for cases like `2> file`.
+  #
+  # (^|[^|<&]) — see the identical note on _bdw_match_redirection above
+  # (apexyard#886/#926): a command that BEGINS with `>` (e.g. the second
+  # half of `echo a > /tmp/ok;> .gitignore` once split on `;`) has no
+  # character before the `>`, so the un-anchored `[^|<&]>...` silently
+  # failed to match at position 0. Anchoring on start-of-string closes
+  # that hole without loosening the `2>&1` / heredoc exclusions.
   local target
-  target=$(echo "$cmd" | grep -oE '[^|<&]>>?[[:space:]]+[^[:space:]&|;]+' \
+  target=$(echo "$cmd" | grep -oE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+' \
                 | head -n 1 \
                 | sed -E 's/^[^>]*>>?[[:space:]]+//')
   if [ -n "$target" ]; then
@@ -513,11 +533,21 @@ _bdw_targets_from_segment() {
   local line target tee_tail
 
   # ALL redirection targets in this segment (not just the first).
+  #
+  # (^|[^|<&]) — apexyard#886/#926 (Hakim security review, PR #926): a
+  # segment produced by bash_extract_write_targets' top-level split can
+  # itself BEGIN with `>` — e.g. splitting `echo a > /tmp/ok;> .gitignore`
+  # on `;` yields a second segment of `> .gitignore`. The un-anchored
+  # `[^|<&]>...` requires a character before `>` to exist, so it silently
+  # dropped every no-space-after-separator redirection (`;>`, `&&>`, `|>`,
+  # `||>`). Anchoring on start-of-segment closes that hole while leaving
+  # the `2>&1` / heredoc exclusions untouched (neither begins a segment
+  # with a bare `>`).
   while IFS= read -r line; do
     [ -z "$line" ] && continue
     target=$(printf '%s\n' "$line" | sed -E 's/^[^>]*>>?[[:space:]]+//')
     [ -n "$target" ] && _bdw_strip_quotes "$target"
-  done < <(printf '%s\n' "$seg" | grep -oE '[^|<&]>>?[[:space:]]+[^[:space:]&|;]+')
+  done < <(printf '%s\n' "$seg" | grep -oE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+')
 
   # ALL tee operands in this segment — `tee a b c` names three targets, not
   # one; the original single-target extractor only ever returned "a".
--- a/.claude/hooks/_lib-detect-bash-write.sh
+++ b/.claude/hooks/_lib-detect-bash-write.sh
@@ -132,8 +132,24 @@ _bdw_starts_with_git_subcommand() {
 # `(^|[^|<&])` adds "start of string/segment" as an equally-valid leading
 # context, closing that hole while leaving the `2>&1` / `<<EOF` exclusions
 # intact (neither of those ever begins a segment with a bare `>`).
+#
+# COMPLETE operator alternation (apexyard#886/#926 round 3 — Hakim's
+# adversarial re-hunt): the pattern above only modelled `>` / `>>` (and,
+# via the leading-context class, `n>` / `n>>` for an fd number). It missed
+# two more real, destructive write operators:
+#   - `&>` / `&>>`  — redirect BOTH stdout and stderr to a file (a write).
+#     Matched by a dedicated `&>>?` alternative: `&` immediately followed
+#     by `>` is unambiguous — it never appears in `2>&1` or `>&2` (there
+#     the `&` comes AFTER the `>`, not before), so this can't collide with
+#     the fd-dup exclusion the leading-context class protects.
+#   - `>|` / `>>|`  — force-clobber (override `set -o noclobber`), also a
+#     write. Modelled by an optional trailing `\|?` on the existing `>>?`.
+# Both are added as alternatives / suffixes to the SAME anchored pattern,
+# not new logic — `2>&1`, `>&2`, `<`, `<<EOF`, `<<<` are all still excluded
+# by the same mechanisms as before (mandatory whitespace immediately after
+# the operator; `&` only recognised when it precedes `>`, never follows).
 _bdw_match_redirection() {
-  echo "$1" | grep -qE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+'
+  echo "$1" | grep -qE '(&>>?|(^|[^|<&])>>?\|?)[[:space:]]+[^[:space:]&|;]+'
 }
 
 # 2. tee.
@@ -412,8 +428,9 @@ bash_extract_write_target() {
   local cmd="$1"
   [ -z "$cmd" ] && return 0
 
-  # Output redirection: capture the first target after > or >>.
-  # Strip leading number for cases like `2> file`.
+  # Output redirection: capture the first target after >, >>, &>, &>>, >|,
+  # or >>|. Strip leading number/ampersand for cases like `2> file` /
+  # `&> file`.
   #
   # (^|[^|<&]) — see the identical note on _bdw_match_redirection above
   # (apexyard#886/#926): a command that BEGINS with `>` (e.g. the second
@@ -421,10 +438,18 @@ bash_extract_write_target() {
   # character before the `>`, so the un-anchored `[^|<&]>...` silently
   # failed to match at position 0. Anchoring on start-of-string closes
   # that hole without loosening the `2>&1` / heredoc exclusions.
+  #
+  # `&>>?` and the trailing `\|?` (apexyard#886/#926 round 3) extend the
+  # same anchored pattern to the full write-redirect operator set — see
+  # the comment on _bdw_match_redirection for why `&>` can't collide with
+  # `2>&1`/`>&2` fd-dup exclusion. The strip sed's `[^>]*` prefix already
+  # swallows a leading `&` the same way it swallows a leading digit or
+  # space, so no separate strip pattern is needed for `&>`/`&>>`; the
+  # trailing `\|?` addition handles `>|`/`>>|`.
   local target
-  target=$(echo "$cmd" | grep -oE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+' \
+  target=$(echo "$cmd" | grep -oE '(&>>?|(^|[^|<&])>>?\|?)[[:space:]]+[^[:space:]&|;]+' \
                 | head -n 1 \
-                | sed -E 's/^[^>]*>>?[[:space:]]+//')
+                | sed -E 's/^[^>]*>>?\|?[[:space:]]+//')
   if [ -n "$target" ]; then
     target="${target%\"}"; target="${target#\"}"
     target="${target%\'}"; target="${target#\'}"
@@ -532,7 +557,9 @@ _bdw_targets_from_segment() {
   local seg="$1"
   local line target tee_tail
 
-  # ALL redirection targets in this segment (not just the first).
+  # ALL redirection targets in this segment (not just the first) — covers
+  # every write-redirect operator: >, >>, >|, >>|, &>, &>>, and n>/n>> (an
+  # fd-numbered redirect, e.g. `2> err.log`, IS a write to err.log).
   #
   # (^|[^|<&]) — apexyard#886/#926 (Hakim security review, PR #926): a
   # segment produced by bash_extract_write_targets' top-level split can
@@ -543,11 +570,20 @@ _bdw_targets_from_segment() {
   # `||>`). Anchoring on start-of-segment closes that hole while leaving
   # the `2>&1` / heredoc exclusions untouched (neither begins a segment
   # with a bare `>`).
+  #
+  # `&>>?` and the trailing `\|?` (apexyard#886/#926 round 3): Hakim's
+  # adversarial re-hunt found this pattern still missed `&>`/`&>>`
+  # (redirect-both-streams) and `>|`/`>>|` (force-clobber) — both real
+  # destructive writes. `&>` is unambiguous against `2>&1`/`>&2` fd-dup
+  # because the `&` precedes the `>` here, never follows it. `>|`/`>>|`
+  # need the SEGMENT to still contain the literal `|` — see
+  # bash_extract_write_targets' clobber-protection step, which shields
+  # these operators from the later bare-`|` (pipe) split.
   while IFS= read -r line; do
     [ -z "$line" ] && continue
-    target=$(printf '%s\n' "$line" | sed -E 's/^[^>]*>>?[[:space:]]+//')
+    target=$(printf '%s\n' "$line" | sed -E 's/^[^>]*>>?\|?[[:space:]]+//')
     [ -n "$target" ] && _bdw_strip_quotes "$target"
-  done < <(printf '%s\n' "$seg" | grep -oE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+')
+  done < <(printf '%s\n' "$seg" | grep -oE '(&>>?|(^|[^|<&])>>?\|?)[[:space:]]+[^[:space:]&|;]+')
 
   # ALL tee operands in this segment — `tee a b c` names three targets, not
   # one; the original single-target extractor only ever returned "a".
@@ -596,6 +632,19 @@ bash_extract_write_targets() {
   local cmd="$1"
   [ -z "$cmd" ] && return 0
 
+  # Protect the force-clobber redirect operators (`>|`, `>>|`) BEFORE the
+  # pipe split below (apexyard#886/#926 round 3, Hakim security review). A
+  # real bash tokenizer lexes `>|`/`>>|` as ONE operator — never as `>`
+  # followed by a separate pipe — so `cmd >| file` / `cmd >>| file` must
+  # not be torn apart at the `|` embedded in the operator itself. Replace
+  # them with placeholders first, run the normal &&/||/;/| split, then
+  # restore the placeholders inside whichever segment(s) still hold them.
+  # Longest-match-first: protect `>>|` before `>|` so `>>|` isn't half-eaten
+  # by the `>|` substitution (`>>|` contains `>|` as a substring).
+  local split="$cmd"
+  split="${split//>>|/@@APEXYARD_CLOBBER_APPEND@@}"
+  split="${split//>|/@@APEXYARD_CLOBBER@@}"
+
   # Split into segments on top-level separators (&&, ||, ;, |) using bash
   # parameter-expansion substitution, NOT sed: BSD sed (macOS's default,
   # non-GNU) does not treat `\n` in the replacement text as a literal
@@ -604,11 +653,15 @@ bash_extract_write_targets() {
   # shipped broken had this been caught only on Linux CI). Order matters:
   # replace the two-character separators BEFORE the single-character ones
   # so `&&` / `||` aren't first flattened into two lone `&` / `|` splits.
-  local split="$cmd"
   split="${split//&&/$'\n'}"
   split="${split//||/$'\n'}"
   split="${split//;/$'\n'}"
   split="${split//|/$'\n'}"
+
+  # Restore the protected clobber operators inside whichever segment(s)
+  # now contain the placeholder.
+  split="${split//@@APEXYARD_CLOBBER_APPEND@@/>>|}"
+  split="${split//@@APEXYARD_CLOBBER@@/>|}"
 
   local seg
   {
--- a/.claude/hooks/_lib-detect-bash-write.sh
+++ b/.claude/hooks/_lib-detect-bash-write.sh
@@ -167,6 +167,88 @@ _bdw_match_redirection() {
   echo "$1" | grep -qE '(&>>?|(^|[^|<&])>>?\|?)[[:space:]]*[^[:space:]&|;]+'
 }
 
+# ------------------------------------------------------------------------------
+# Internal: _bdw_split_top_level COMMAND
+#
+# Splits COMMAND into top-level segments on &&, ||, ;, | — echoed one per
+# line via stdout. THE canonical segmentation for this whole library:
+# both DETECTION (_bdw_match_redirection_any_segment, below) and
+# EXTRACTION (bash_extract_write_targets) call this SAME function, so
+# they cannot disagree about where one command ends and the next begins
+# (apexyard#886/#926 round 5 — Hakim security review; see that helper's
+# comment for why divergence was the actual structural bug behind rounds
+# 1-5 of this operator class, not "yet another missed operator").
+#
+# Protects the force-clobber operators (`>|`, `>>|`) from being torn
+# apart by the bare-`|` split — a real bash tokenizer lexes them as ONE
+# operator, never as `>` followed by a separate pipe — via a
+# placeholder-and-restore step. Uses bash parameter-expansion
+# substitution, NOT sed, for the split itself: BSD sed (macOS's default,
+# non-GNU) does not honour `\n` in replacement text as a literal newline,
+# so a sed-based split would silently no-op on a stock Mac.
+# ------------------------------------------------------------------------------
+_bdw_split_top_level() {
+  local cmd="$1"
+  [ -z "$cmd" ] && return 0
+
+  local split="$cmd"
+  # Longest-match-first: protect >>| before >| so >>| isn't half-eaten by
+  # the >| substitution (>>| contains >| as a substring).
+  split="${split//>>|/@@APEXYARD_CLOBBER_APPEND@@}"
+  split="${split//>|/@@APEXYARD_CLOBBER@@}"
+
+  # Two-character separators BEFORE single-character ones, so && / ||
+  # aren't first flattened into two lone & / | splits.
+  split="${split//&&/$'\n'}"
+  split="${split//||/$'\n'}"
+  split="${split//;/$'\n'}"
+  split="${split//|/$'\n'}"
+
+  split="${split//@@APEXYARD_CLOBBER_APPEND@@/>>|}"
+  split="${split//@@APEXYARD_CLOBBER@@/>|}"
+
+  printf '%s\n' "$split"
+}
+
+# ------------------------------------------------------------------------------
+# Internal: _bdw_match_redirection_any_segment COMMAND
+#
+# THE STRUCTURAL FIX for apexyard#886/#926 round 5 (Hakim security
+# review). Runs _bdw_match_redirection on EACH top-level segment of
+# COMMAND (via _bdw_split_top_level) instead of on the whole, unsplit
+# command.
+#
+# Why this was the actual bug, not another missed operator: DETECTION
+# (bash_command_appears_to_write, via this helper's predecessor) used to
+# call _bdw_match_redirection on the WHOLE command. EXTRACTION
+# (bash_extract_write_targets) already split into segments FIRST, then
+# anchored `^` per segment. For `false ||> src/app.ts` — a real,
+# truncating write — the `>` is immediately preceded by `|` (from `||`).
+# On the whole, unsplit string, `[^|<&]` (needed to exclude `2>&1`/`>&2`
+# fd-dup) EXCLUDES that `|`-preceded `>` from matching, and it isn't at
+# `^` either (it's preceded by `|`), so detection said "not a write" —
+# false negative, exit 0, no ticket. Extraction, meanwhile, split on `||`
+# first, leaving `> src/app.ts` as its own segment where the `>` sits at
+# `^` — which the anchor DOES match — so extraction correctly found the
+# target. Detection and extraction disagreed; that disagreement, not the
+# `|`/`||` operator itself, was the root cause across rounds 1-5 (`;>`
+# survived only because `;` isn't excluded by the leading-context class;
+# `&&>` survived only by coincidentally containing the substring `&>`;
+# `|>`/`||>`/`||>|` had no such rescue). Splitting BEFORE matching, with
+# the SAME segmentation function extraction already uses, means the two
+# paths can't diverge again — this is the fix, not one more operator
+# added to the alternation.
+# ------------------------------------------------------------------------------
+_bdw_match_redirection_any_segment() {
+  local cmd="$1" seg
+  [ -z "$cmd" ] && return 1
+  while IFS= read -r seg; do
+    [ -z "$seg" ] && continue
+    _bdw_match_redirection "$seg" && return 0
+  done < <(_bdw_split_top_level "$cmd")
+  return 1
+}
+
 # 2. tee.
 _bdw_match_tee() {
   echo "$1" | grep -qE '\btee\b'
@@ -348,7 +430,11 @@ bash_command_appears_to_write() {
   local cmd="$1"
   [ -z "$cmd" ] && return 1
 
-  _bdw_match_redirection     "$cmd" && return 0
+  # Segment-aware (apexyard#886/#926 round 5) — see
+  # _bdw_match_redirection_any_segment for why matching the WHOLE, unsplit
+  # command here (as this used to) missed `|`/`||`-adjacent redirects
+  # (`false ||> file`, `echo x |> file`).
+  _bdw_match_redirection_any_segment "$cmd" && return 0
   _bdw_match_tee             "$cmd" && return 0
   _bdw_match_sed_inplace     "$cmd" && return 0
   _bdw_match_awk_inplace     "$cmd" && return 0
@@ -395,7 +481,12 @@ bash_command_is_deletion_only() {
   fi
 
   # Any other content-writing pattern alongside rm → not deletion-only.
-  _bdw_match_redirection    "$cmd" && return 1
+  # Segment-aware (apexyard#886/#926 round 5) — same reasoning as
+  # bash_command_appears_to_write: matching the whole, unsplit command
+  # here would miss `rm x; false ||> src/app.ts` (a real write hiding
+  # behind a `|`/`||`-adjacent redirect), wrongly classifying it as
+  # deletion-only and exempting it from the ticket gate.
+  _bdw_match_redirection_any_segment "$cmd" && return 1
   _bdw_match_tee            "$cmd" && return 1
   _bdw_match_sed_inplace    "$cmd" && return 1
   _bdw_match_awk_inplace    "$cmd" && return 1
@@ -438,6 +529,14 @@ bash_command_is_deletion_only() {
 #   - cmd with multiple redirects
 #   - paths constructed from variables
 #   - tar -x (target is a directory, often implicit)
+#   - `cmd <> file` (read-write open, apexyard#886/#926 round 5): known,
+#     accepted, OUT-OF-SCOPE gap — exotic, non-truncating (unlike every
+#     operator this file DOES model, `<>` opens for read+write WITHOUT
+#     truncating existing content, so it's a materially different risk
+#     shape), and not adjacent to any of the &&/||/;/| separators this
+#     library segments on, so the round-5 structural fix doesn't reach
+#     it either. Tracked as a separate follow-up rather than folded into
+#     this operator-class close.
 # ------------------------------------------------------------------------------
 bash_extract_write_target() {
   local cmd="$1"
@@ -666,47 +765,22 @@ _bdw_targets_from_segment() {
 # extractable targets produces no output at all — callers must treat
 # "nothing on stdout" as the same fail-closed/categorical case that empty
 # bash_extract_write_target output signals today.
+#
+# Segmentation is delegated to _bdw_split_top_level (apexyard#886/#926
+# round 5) — the SAME function bash_command_appears_to_write's
+# _bdw_match_redirection_any_segment uses, so detection and extraction
+# share one source of truth for "where does one command end and the next
+# begin" and can't drift apart again.
 # ------------------------------------------------------------------------------
 bash_extract_write_targets() {
   local cmd="$1"
   [ -z "$cmd" ] && return 0
-
-  # Protect the force-clobber redirect operators (`>|`, `>>|`) BEFORE the
-  # pipe split below (apexyard#886/#926 round 3, Hakim security review). A
-  # real bash tokenizer lexes `>|`/`>>|` as ONE operator — never as `>`
-  # followed by a separate pipe — so `cmd >| file` / `cmd >>| file` must
-  # not be torn apart at the `|` embedded in the operator itself. Replace
-  # them with placeholders first, run the normal &&/||/;/| split, then
-  # restore the placeholders inside whichever segment(s) still hold them.
-  # Longest-match-first: protect `>>|` before `>|` so `>>|` isn't half-eaten
-  # by the `>|` substitution (`>>|` contains `>|` as a substring).
-  local split="$cmd"
-  split="${split//>>|/@@APEXYARD_CLOBBER_APPEND@@}"
-  split="${split//>|/@@APEXYARD_CLOBBER@@}"
-
-  # Split into segments on top-level separators (&&, ||, ;, |) using bash
-  # parameter-expansion substitution, NOT sed: BSD sed (macOS's default,
-  # non-GNU) does not treat `\n` in the replacement text as a literal
-  # newline, so a `sed 's/;/\n/g'`-style split silently no-ops on a stock
-  # macOS box (the #886 bug this was first written to fix would have
-  # shipped broken had this been caught only on Linux CI). Order matters:
-  # replace the two-character separators BEFORE the single-character ones
-  # so `&&` / `||` aren't first flattened into two lone `&` / `|` splits.
-  split="${split//&&/$'\n'}"
-  split="${split//||/$'\n'}"
-  split="${split//;/$'\n'}"
-  split="${split//|/$'\n'}"
-
-  # Restore the protected clobber operators inside whichever segment(s)
-  # now contain the placeholder.
-  split="${split//@@APEXYARD_CLOBBER_APPEND@@/>>|}"
-  split="${split//@@APEXYARD_CLOBBER@@/>|}"
 
   local seg
   {
     while IFS= read -r seg; do
       [ -z "$seg" ] && continue
       _bdw_targets_from_segment "$seg"
-    done <<< "$split"
+    done < <(_bdw_split_top_level "$cmd")
   } | awk '!seen[$0]++'
 }
--- a/.claude/hooks/_lib-detect-bash-write.sh
+++ b/.claude/hooks/_lib-detect-bash-write.sh
@@ -146,10 +146,25 @@ _bdw_starts_with_git_subcommand() {
 #     write. Modelled by an optional trailing `\|?` on the existing `>>?`.
 # Both are added as alternatives / suffixes to the SAME anchored pattern,
 # not new logic — `2>&1`, `>&2`, `<`, `<<EOF`, `<<<` are all still excluded
-# by the same mechanisms as before (mandatory whitespace immediately after
-# the operator; `&` only recognised when it precedes `>`, never follows).
+# by the same mechanisms as before (whitespace after the operator is now
+# OPTIONAL — see the #886/#926 round-4 note below — but the exclusions
+# don't rely on mandatory whitespace; they rely on `&` only being
+# recognised when it precedes `>`, never follows).
+#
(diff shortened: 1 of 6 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #926

**Commit**: `3f9c81a026b95fe9fe8d3bd7e0480bc4146ab312`

### Summary
Extends `_lib-detect-bash-write.sh` with `bash_extract_write_targets` (plural) and rewires `require-active-ticket.sh` (AND semantics, one evaluation per target) and `require-migration-ticket.sh` (scan every target for a migration-shaped one) to judge EVERY bash write target. Closes the #886 gate-bypass where `echo a > /tmp/x; echo b > src/app.ts` was exempted on its first, out-of-governance target and never re-checked target #2.

### Checklist Results
- Architecture & Design:     Pass — singular extractor kept for compat; new plural fn layered on top; per-target eval function has fully-isolated `local` state, nothing leaks between targets.
- Code Quality:              Pass — clear comments, fail-closed on unparseable segments (never a fabricated target).
- Testing:                   Pass — 108/108 + 48/48 + 21/21 green; verified load-bearing (all new cases fail against pre-fix source, incl. the two multi-target blocking cases returning allowed instead of blocked).
- Security:                  Pass — this hardens the trust-chain gate; the #883/#885 raw-and-resolved out-of-governance checks are reused unchanged per target.
- Performance:               Pass — negligible (per-target re-derivation of HOOK_DIR only).
- PR Description & Glossary: Pass — narrative summary, testing steps, Closes #886, Glossary present.
- Technical Decisions (AgDR):N/A — extends an existing helper to fix a bug; no new library/framework/pattern.
- Adopter Handbooks:         N/A — shell-only diff; no language handbook matches, and the always-load handbooks (TS layers / DB-migration DDL / commit messages) are not triggered.
- Fallow Static Analysis:    N/A — not JS/TS.

### Multi-target extraction — verified complete
- Segment split uses bash parameter-expansion (not `sed s/;/\n/`), correctly avoiding the BSD-sed pitfall where `\n` in the replacement is not a literal newline. Two-char separators (`&&`, `||`) are replaced before single-char (`;`, `|`), so `&&` is not first flattened into two `&` — correct ordering.
- `tee` word-boundary match uses `grep` `\b` (BSD sed does not support `\b`) and strips the `tee` token with anchored parameter expansion, not sed — both BSD-sed avoidances are sound and well-documented in-line.
- Multi-operand `tee` and multi-redirect per segment captured; per-segment fallback to the singular extractor covers sed -i / cp / mv / curl -o / wget -O; output deduped via `awk "!seen[\$0]++"`. Confirmed by the extractor test cases.

### AND semantics / no regression
- `require-active-ticket.sh` evaluates each target and `exit 2` on the first that fails — verified by cases 41/45 (multi-target + tee blocked w/o ticket) and 42 (allowed with ticket). Single-target regressions 43 (out-of-repo exempt) / 44 (in-repo blocked) pass — the per-target loop is not stricter than before for the single-target case.
- Empty-target (unextractable) path still hits the categorical fail-closed gate, matching pre-#886 behaviour.

- nit: both the multi-redirect regex and the singular fallback require whitespace after `>`, so a no-space redirect (`echo a >/tmp/x`) is still a miss. This is pre-existing (documented best-effort, false-negative-preferred) and NOT a regression from this PR — noting only so a future tightening is on record.

### Verdict
**APPROVED**

---
Reviewed commit: `3f9c81a026b95fe9fe8d3bd7e0480bc4146ab312`

atlas-apex (commented):
## Security Review: PR #926

**Commit**: `3f9c81a026b95fe9fe8d3bd7e0480bc4146ab312`

▸ Activating Hakim (Security Auditor) for PR #926 (trigger: diff touches `.claude/hooks/**` — the trust chain, #777)

### Summary

This hardens two trust-chain gates (`require-active-ticket.sh`, `require-migration-ticket.sh`) so they judge **all** bash write targets instead of only the first. The AND-semantics dispatch, per-target state isolation (`_ratc_evaluate_target`), the empty-targets categorical fail-closed path, and the migration hook's "first migration-shaped target" loop are all well-constructed and correct for the target shapes they extract. The two macOS/BSD portability notes (grep `\b` vs BSD sed, parameter-expansion split vs `sed 's/;/\n/'`) are genuinely load-bearing and correctly reasoned.

**However, the fix does not fully close the multi-target bypass it targets.** A mirror variant survives, and I reproduced it end-to-end against this exact commit.

### Checklist Results
- Auth & Authorisation:   **Fail** — residual gate bypass (below)
- Fail-closed design:     Partial — empty-targets path is fail-closed; the non-empty-but-incomplete path is fail-OPEN

### Security Issues Found

#### HIGH — leading-redirection after a top-level separator (no space) evades extraction → gate bypass

`bash_extract_write_targets` splits on `&&`/`||`/`;`/`|` by replacing the separator with a newline, then extracts per segment. The redirection matcher — `grep -oE '[^|<&]>>?[[:space:]]+…'` in `_bdw_targets_from_segment` (and the same regex in `bash_extract_write_target`) — requires **exactly one character before `>`**. When the separator has no trailing space (`;>`, `&&>`, `|>`, `||>`), the split leaves a segment that *begins* with `>`, so the regex never matches and that target is silently dropped. If an earlier target in the command is exempt (e.g. `/tmp/...` out-of-governance), the AND-semantics loop sees only the exempt target and passes — with no active ticket.

This is the same failure family the PR fixes (`echo a > /tmp/x; echo b > src/app.ts`), just with the whitespace removed. Because the attacker fully controls spacing, it's trivially reachable. The leading-redirect *content* form (`>file cmd`, valid bash) means this writes arbitrary content, not merely truncation.

**Reproduced end-to-end** against `3f9c81a`, using this PR's own sandbox harness (ops-fork sandbox, no ticket set — in-repo writes must block):

```
control plain in-repo write     [echo x > .gitignore]                          exit 2  BLOCKED   (correct)
control #886 spaced-both         [echo a > /tmp/ok; echo b > .gitignore]        exit 2  BLOCKED   (fix works)
ATTACK ; no-space                [echo a > /tmp/ok;> .gitignore]                exit 0  PASSED    <-- BYPASS
ATTACK ; content form            [echo a > /tmp/ok;> .gitignore cat /etc/hostname]  exit 0 PASSED  <-- BYPASS (writes content)
ATTACK && no-space               [echo a > /tmp/ok&&> .gitignore]               exit 0  PASSED    <-- BYPASS
ATTACK | no-space                [echo a > /tmp/ok|> .gitignore]                exit 0  PASSED    <-- BYPASS
ATTACK || no-space               [echo a > /tmp/ok||> .gitignore]               exit 0  PASSED    <-- BYPASS
```

`bash_extract_write_targets 'echo a > /tmp/ok;> .gitignore'` returns only `/tmp/ok`; `.gitignore` is never judged.

**`require-migration-ticket.sh` inherits the same hole** — it consumes the identical `bash_extract_write_targets` output, so a migration-shaped leading-redirect target (`echo a > /tmp/ok;> db/migrate/001.rb cat payload`) is dropped before `is_migration_path` ever sees it, bypassing the migration gate.

#### Suggested fix (verified)

Allow a start-of-segment redirection by anchoring the alternation in both the `_bdw_targets_from_segment` redirection grep and `bash_extract_write_target`:

```
grep -oE '(^|[^|<&])>>?[[:space:]]+[^[:space:]&|;]+'
```

The existing strip (`sed -E 's/^[^>]*>>?[[:space:]]+//'`) handles the captured `> file` form unchanged. I verified this candidate: it now extracts `src/app.ts` from a start-of-segment `> src/app.ts` (and the `>file cmd` content form) and from `2> err.log`, while still correctly rejecting `foo 2>&1` (fd dup) and `a <<EOF` (heredoc) — no new false positives on those two shapes.

### Test gap

The new `assert_targets` cases and the `#886` e2e cases all use **space-separated** multi-target inputs; none exercise the no-space leading-redirect form. Please add adversarial cases for `;>`, `&&>`, `|>`, `||>` (no space) and the `>file cmd` content form to:
- `test_detect_bash_write.sh` (`assert_targets` — must include the missed in-repo target)
- `test_require_active_ticket_bash.sh` (exempt-first + leading-redirect-second, no ticket → BLOCKED)
- `test_require_migration_ticket.sh` (non-migration-first + migration-shaped leading-redirect, no ticket → BLOCKED)

### Verdict

**CHANGES REQUESTED** — the core design is sound, but the fix leaves a demonstrable, attacker-controllable bypass in the same class it set out to close, on a security-critical trust-chain hook. No security sign-off marker written. Happy to re-review promptly once the start-of-segment redirect case (and its tests) are in.

---
📌 Reviewed commit: `3f9c81a026b95fe9fe8d3bd7e0480bc4146ab312`

atlas-apex:
Pushed a follow-up commit (`e7d7253`) addressing Hakim's security-review finding — a residual no-space bypass.

**The gap:** the redirection regex `[^|<&]>>?...` requires a character to exist *before* the `>` for a match. A segment that begins with `>` — e.g. splitting `echo a > /tmp/ok;> .gitignore` on `;` produces a second segment of `> .gitignore` — has no preceding character, so the pattern silently failed to match at position 0 and the target was dropped. Same hole existed for `&&>`, `|>`, `||>`, and even a whole command that's just `> file`.

**The fix:** anchored the alternation to `(^|[^|<&])>>?...` in all three call sites (`_bdw_match_redirection`, the singular `bash_extract_write_target`, and the plural `bash_extract_write_targets`' per-segment loop). Verified `2>&1` and `<<EOF` heredocs are unaffected — both still correctly excluded/handled.

Re-ran Hakim's exact repro commands — all five now exit 2 (BLOCKED) with no active ticket, exit 0 when one is set. Added adversarial no-space coverage to all three test files; each new case fails against the pre-fix code and passes after (verified via git stash). All three suites green (115 + 54 + 22), `bash -n` and `shellcheck -S warning` clean, diff vs the merge-base with `upstream/dev` still only the 6 intended files.

atlas-apex (commented):
## Code Review (re-review): PR #926

**Commit**: `e7d725394de1b5065c4f324481a883c1904e5066`

### Summary
Re-review at the new HEAD after Hakim's security finding on my prior APPROVED. The residual bypass was a no-space leading redirection: after `bash_extract_write_targets` splits a command on a top-level separator (`;`, `&&`, `|`, `||`), a segment can *begin* with `>` (e.g. `echo a > /tmp/ok;> src/app.ts` → second segment `> src/app.ts`). The old leading-context class `[^|<&]>` required a character to exist before `>`, so it silently dropped the target at position 0 — letting an in-repo write slip past a gate that only judged the out-of-repo first target. Fixed by anchoring the redirection regex to `(^|[^|<&])>>?…`.

### Verification
- **SHA** matches HEAD `e7d7253`.
- **Anchor applied at all 3 sites** — `_bdw_match_redirection`, the singular `bash_extract_write_target`, and the plural `_bdw_targets_from_segment` per-segment loop all use `(^|[^|<&])>>?`.
- **No-space forms now caught** — verified empirically under bash: `;>`, `&&>`, `|>`, `||>`, and `;>>` all extract the in-repo target (not just the out-of-repo one). Proven this is a *real* regression guard by reverting to the un-anchored pattern: pre-fix all four forms miss `.gitignore` and only surface `/tmp/ok` — the exact bypass.
- **No new false positives** — `2>&1` fd-dup and `<<EOF` heredocs are not treated as writes (heredoc target extraction still works); `>&2` stays excluded via the target char class. The `^` alternative can't match `2>&1`/`<<EOF` because neither begins a segment with a bare `>`.
- **Tests added to all 3 files** — unit extraction (4 no-space forms + fd-dup + heredoc guards), active-ticket end-to-end gate (blocks w/o ticket, allows WITH — proving the bypass is closed at the gate, not just extraction), and migration gate (`;>` into a migration path blocks).
- **Suites green** — 115 / 54 / 22, matching the reported counts.
- Shell safety confirmed: all three files carry `#!/bin/bash` and settings.json `exec`s them under bash, so the `${var//;/$'\n'}` ANSI-C segment split is guaranteed.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass (clear rationale comments at each anchor site + BSD-sed/`\b` portability notes)
- Testing: Pass (adversarial, genuinely fails pre-fix, covers unit + both gates)
- Security: Pass (closes the Hakim-reported bypass; fail-closed direction preserved)
- Performance: Pass

### Verdict
**APPROVED**

---
📌 Reviewed commit: `e7d725394de1b5065c4f324481a883c1904e5066`

atlas-apex (commented):
## Security Re-Review: PR #926

**Commit reviewed**: `e7d725394de1b5065c4f324481a883c1904e5066`
🛡️ Hakim (Security Auditor) — follow-up to my prior CHANGES REQUESTED on the no-space leading-redirection bypass.

### The bypass I flagged is CLOSED ✅

The `(^|[^|<&])>>?` anchor at all three call sites correctly closes the start-of-segment plain-`>` hole. Re-ran my exact repros against `e7d725394de1b5065c4f324481a883c1904e5066` (isolated sandbox, no active ticket, `APEXYARD_OPS_DISABLE_PIN=1`):

| Repro | require-active-ticket | require-migration-ticket |
|-------|:---:|:---:|
| `echo a > /tmp/ok;> .gitignore` | **BLOCKED (2)** | **BLOCKED (2)** |
| `echo a > /tmp/ok;> .gitignore cat /etc/hostname` | **BLOCKED (2)** | — |
| `echo a > /tmp/ok&&> <path>` | **BLOCKED (2)** | **BLOCKED (2)** |
| `echo a > /tmp/ok|> <path>` | **BLOCKED (2)** | **BLOCKED (2)** |
| `echo a > /tmp/ok||> <path>` | **BLOCKED (2)** | — |

Also verified:
- **No regression** — the spaced forms (`; > f`, `&& > f`, `|| > f`) still block.
- **No new false positives** — `2>&1` fd-dup and `<<EOF` heredocs remain `appears_to_write=no` (empty target set) on both the un-anchored and anchored paths; the PR's own suite confirms (`115 PASS / 0 FAIL`, incl. the `2>&1`/heredoc cases).
- **Same-extractor migration gate** — `require-migration-ticket.sh` also no longer drops the no-space target (table above, right column).
- Extra start-of-segment shapes I threw at it all now extract the target: leading `> f`, `;>> f` (append), `(> f` subshell, newline separator, tab after `;>`.

The core change is sound and does exactly what it set out to do.

### Withholding the security marker — two sibling bypasses of the same class survive ⚠️

My adversarial mandate was "any OTHER way to produce an *otherwise-unmatched write target*." Two survive — and they clear **both** gates, including the data-safety migration gate:

```
# redirect-both — NOT blocked (PASS 0) on either hook:
echo a > /tmp/ok; echo hi &> .gitignore
echo x &> db/migrations/001.sql

# force-clobber — NOT blocked (PASS 0) on either hook:
echo hi >| .gitignore
echo a > /tmp/ok;>| db/migrations/006.sql
```

Both are **real destructive writes**, not no-ops — proven directly:
```
$ printf 'ORIG' > .gitignore; bash -c 'echo pwned &> .gitignore'; cat .gitignore
pwned
$ printf 'ORIG' > .gitignore; bash -c 'echo pwned >| .gitignore'; cat .gitignore
pwned
```

**Root cause** — same defect family, one layer deeper: the matcher only models `>`, `>>`, `2>` (per the lib's own scope comment). `&>`/`&>>`/`>|` are never recognised, so `bash_command_appears_to_write` returns `no` and the command is treated as read-only. Note `&>>` happens to block (the first `>` acts as leading context for the second), but `&>` and `>|` do not. This is **pre-existing** (predates #886) and outside the documented matcher scope — this PR neither introduced nor was scoped to it — but it is a live bypass of the same "unmatched write target" class I was asked to sweep, and `&>` in particular is a common idiom (`cmd &> file`), not exotic.

### Severity

**MEDIUM** — pre-existing, out-of-documented-scope, and this gate's charter explicitly "prefers false-negatives"; but `&>` is a routine redirection an agent would plausibly type, and it silently clears the migration (data-safety) gate too.

### Recommendation

Smallest fix: broaden the operator alternation in `_bdw_match_redirection` and the three extraction regexes to also match `&>`, `&>>`, and `>|` (e.g. allow an optional leading `&` and an optional trailing `|` on the redirection token). Add `&>` / `>|` rows to `test_detect_bash_write.sh`. Alternatively, if the team accepts these as known-limitations, document them explicitly in the lib scope comment and file a follow-up — either way I'd like the class either closed or its acceptance recorded before I sign off.

### Verdict

**CHANGES REQUESTED (extend to `&>`/`>|`, or record explicit acceptance)** — the flagged bypass is genuinely closed and the fix is good; I'm withholding the security sign-off marker only because the adversarial sweep turned up these two live sibling bypasses of the same class.

---
📌 Reviewed commit: `e7d725394de1b5065c4f324481a883c1904e5066`

atlas-apex:
Pushed a second follow-up (`95fa2d6`) for Hakim's adversarial re-hunt — two more surviving write-target bypasses, same "unjudged write target" family as #886's original scope.

**The gap:** `_bdw_match_redirection` and both extraction regexes only modelled `>`, `>>`, and `n>` (fd-numbered). They missed `&>`/`&>>` (redirect both stdout+stderr to a file — a real write) and `>|`/`>>|` (force-clobber, noclobber override — also a real write). Both repro'd as PASS/exit-0 with no ticket:
- `echo hi &> .gitignore`
- `echo a > /tmp/ok;>| db/migrations/006.sql`

**The fix:** broadened the operator alternation to `(&>>?|(^|[^|<&])>>?\|?)` in all three call sites. `&>` is unambiguous against `2>&1`/`>&2` fd-dup because the `&` precedes the `>` here — order alone disambiguates. Also protected `>|`/`>>|` from the segment splitter's bare-`|` split (real bash lexes them as one operator, never `>` + pipe) via a placeholder-and-restore step.

**Verified exhaustively:**

| Operator | Blocks? |
|---|---|
| `>` | Yes |
| `>>` | Yes |
| `>\|` | Yes |
| `>>\|` | Yes |
| `&>` | Yes |
| `&>>` | Yes |
| `n>` (fd-numbered, e.g. `2>`) | Yes |
| `2>&1` (fd-dup) | No — correctly not gated |
| `>&2` (fd-dup) | No — correctly not gated |
| `1>&2` (fd-dup) | No — correctly not gated |
| `<` (read) | No — correctly not gated |
| `<<EOF` (heredoc) | No — correctly not gated (still extracts its own write target normally) |
| `<<<` (herestring) | No — correctly not gated |

Re-ran Hakim's original 5 + 2 new repros end-to-end through `require-active-ticket.sh` — all 7 now exit 2 (BLOCKED) with no ticket. Added adversarial rows for `&>`, `&>>`, `>|`, `>>|` at extraction level + both gate levels; each fails pre-this-fix and passes after (confirmed via `git stash`), except `&>>` which — by coincidence unrelated to this fix, noted in the test comment — already passed under the round-2 regex; it's still correct post-fix.

All three suites green (126 + 63 + 24), `bash -n` and `shellcheck -S warning` clean, diff vs the `upstream/dev` merge-base still only the 6 intended files.

atlas-apex (commented):
## Code Review: PR #926 — Round 3 (re-review)

**Commit**: `95fa2d647ba35d513555570e0ca05a4d32be1ece`

### Summary
Round-3 response to Hakim's adversarial re-hunt: the write-target operator alternation is broadened from `[^|<&]>>?...` to `(&>>?|(^|[^|<&])>>?\|?)...` to catch `&>`/`&>>` (redirect-both-streams) and `>|`/`>>|` (force-clobber), plus a placeholder-protect step so `>|`/`>>|` survive the top-level pipe split intact.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass
- Performance: Pass
- Technical Decisions (AgDR): N/A (bug fix, no new decision)

### What I verified (empirically, under `/bin/bash` 3.2 — the real hook runtime)

1. **Broadened regex applied at all 3 sites** — `_bdw_match_redirection` (boolean), `bash_extract_write_target` (singular), and `_bdw_targets_from_segment` (plural) all carry the identical `(&>>?|(^|[^|<&])>>?\|?)` alternation. Confirmed in diff + runtime.

2. **New write operators caught**: `&>`, `&>>`, `>|`, `>>|` all extract the correct target and register as writes.

3. **No new false-positives** — fd-dups `2>&1`, `>&2`, `1>&2` and reads `<`, `<<`, `<<<` are still NOT treated as writes and yield zero targets. The `&>` alternative can't collide with `>&2`/`2>&1` because `&` precedes `>` in a write but follows it in a dup.

4. **Placeholder-protect is sound** — `>>|`/`>|` are substituted to placeholders before the `&&`/`||`/`;`/`|` split and restored after; longest-match-first (`>>|` before `>|`) is correct. Verified `>|`/`>>|` survive real pipes: `a >>| append.txt | grep b` → `append.txt` only; `cmd >| out; echo hi | wc -l` → `out` only; two clobbers across `&&` both captured.

5. **Multi-target round-2 fix still holds** — `echo a > /tmp/ok;> .gitignore` → both `/tmp/ok` and `.gitignore`; `tee a b c` names all operands; `cmd > a 2> b` → both.

6. **New adversarial tests genuinely gate** — confirmed against the PRE-round-3 regex that `&>` and `>|` are genuinely dropped (real fail-pre/pass-post discriminators). The `&>>` row is honestly documented as a coincidental pre-fix pass (the 2nd `>` of `&>>` reads the 1st `>` as valid leading context) and kept as a coverage check, not a dead test claiming to discriminate — verified honest, not a lying test.

7. **Suites green**: 126 / 63 / 24, all `bash -n` clean. Both `require-active-ticket.sh` and `require-migration-ticket.sh` consume `bash_extract_write_targets`, so the fix reaches the gates.

### Issues Found
None blocking.

### Suggestions
- nit (non-blocking, follow-up): the multi-operand `tee` loop emits redirect operators as spurious targets when a redirect follows a tee operand — `tee f1 >| f2` yields a bogus `>|` target (same for `tee f1 > f2` → `>` and `tee f1 2> f2` → `2>`). This is **pre-existing to the round-3 delta** (it affects `>`/`2>` identically, not just the new `>|`) and is **fail-closed** — an extra spurious target only makes the AND-semantics gate stricter (over-capture), never a bypass. Worst realistic case is a rare false-positive over-gate on the unusual `tee <exempt> <redirect> <exempt>` shape. Not a security concern; worth a small follow-up to have the tee loop stop at the first redirect operator.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `95fa2d647ba35d513555570e0ca05a4d32be1ece`

atlas-apex (commented):
## Security Review: PR #926 (round 3)

**Commit**: `95fa2d647ba35d513555570e0ca05a4d32be1ece`

### Verdict: CHANGES REQUESTED — one surviving, IN-SCOPE bypass. No approval marker written.

### Summary
Re-ran every prior repro at HEAD `95fa2d6` against **both** `require-active-ticket.sh` and `require-migration-ticket.sh` in a sandboxed ops-root with **no active ticket**. Rounds 1 and 2 are genuinely fixed and no new false-positives were introduced. But an adversarial re-hunt found a **new bypass in the same "unjudged redirect write target" class this ticket owns**: a redirect operator with **no whitespace between the operator and the target filename** defeats detection entirely.

### Round 1 + Round 2 — VERIFIED FIXED (all block, exit 2 on both hooks)

| Repro (2nd write targets `src/migrations/001.sql`) | active | mig |
|---|---|---|
| `echo a > /tmp/ok;> …` (no-space leading `;>`) | 2 | 2 |
| `…&&> …`  /  `…\|\|> …`  /  `…\|> …`  /  `…;>> …` | 2 | 2 |
| `echo b &> …` (both-streams) / `&>> …` | 2 | 2 |
| `echo b >\| …` (force-clobber) / `>>\| …` | 2 | 2 |
| `echo b&> …` / `echo b>\| …` (no space *before* op) | 2 | 2 |

### False-positive guards — VERIFIED CLEAN (all pass, exit 0, not gated)

`2>&1`, `>&2`, `1>&2` (fd-dup) · `cat < f` · `cat <<EOF…EOF` · `cat <<< x` — none gated as writes. No over-blocking introduced.

### 🔴 NEW FINDING — HIGH — no-space-after-operator redirect slips both gates (IN-SCOPE)

The redirection matcher and both extractors require **one-or-more** whitespace after the operator (`[[:space:]]+`). Real bash accepts **zero** whitespace between a redirect operator and its target, so every no-space redirect writes a file that neither gate ever judges:

| Exact repro (no active ticket) | active | mig | real bash effect |
|---|---|---|---|
| `echo hi>src/migrations/001.sql` | **0** | **0** | writes file (`NOT-A-WRITE` — missed entirely) |
| `echo a > /tmp/ok; echo b 2>src/migrations/001.sql` | **0** | **0** | writes file (target missed) |
| `echo a > /tmp/ok; echo b>>src/migrations/001.sql` | **0** | **0** | append-writes file |
| `echo a > /tmp/ok; echo b>\|src/migrations/001.sql` | **0** | **0** | force-clobber writes file |
| `echo a > /tmp/ok; echo b&>src/migrations/001.sql` | **0** | **0** | both-streams writes file |

Each was confirmed to actually create/write the file under `bash -c` (e.g. `echo AA>f1` → `f1` contains `AA`; `ls /nonexistent 2>f3` → `f3` contains the stderr). `exit 0` from both hooks = the gate never saw the target = **bypass**.

**Root cause:** the mandatory `[[:space:]]+` in `_bdw_match_redirection`, in `bash_extract_write_target`'s redirection grep + its `sed -E 's/^[^>]*>>?\|?[[:space:]]+//'` strip, and in the identical grep inside `_bdw_targets_from_segment`. This whitespace assumption predates this PR (it was in the original #151 matcher), but it is squarely the **redirect-operator write-target class** #886 is chartered to close — not a genuinely different mechanism like process substitution — so the operator surface is **not yet exhausted**.

**Suggested fix direction (validated):** relax `[[:space:]]+` → `[[:space:]]*` in the matcher, both extractor greps, **and** the strip `sed`s (all four sites — a `*` matcher with a `+` strip would match but fail to strip the operator off the target). I confirmed `[[:space:]]*` catches all five no-space writes above **and still excludes** `2>&1` / `>&2` / `1>&2` / `< f` / `<<<` (the `[^[:space:]&|;]+` target class already rejects the `&` in fd-dup forms, so making the space optional does not reintroduce those false-positives). Add the five rows above to `tests/test_detect_bash_write.sh` as regression cases.

### In-scope / out-of-scope call
**IN-SCOPE.** This is the same redirect-operator "unjudged write target" class the ticket owns, reachable through the same operators (`>`, `>>`, `2>`, `>|`, `&>`) the PR already touches — just without the whitespace the pattern assumes. Per the review charter (approve only if the redirect-operator class is *fully* closed), I'm withholding the marker rather than deferring to a follow-up ticket. (For contrast, `>(cmd)` process substitution *would* be an arguably-different mechanism worth a separate ticket — but it isn't needed to justify the hold; the no-space redirect alone is enough.)

---
📌 Reviewed commit: `95fa2d647ba35d513555570e0ca05a4d32be1ece`

atlas-apex:
Pushed the definitive close (`2a4b7bb`) for the write-redirect operator class — Hakim's fourth adversarial re-hunt.

**The gap:** the mandatory `[[:space:]]+` after the operator (in the matcher, both extraction greps, and both strip-seds) silently dropped every no-space form. Bash accepts zero whitespace between a redirect operator and its target — five real, destructive writes were passing with no ticket: `echo hi>file`, `2>file`, `>>file`, `>|file`, `&>file`.

**The fix:** relaxed `[[:space:]]+` → `[[:space:]]*` at all four sites, exactly as Hakim validated. This doesn't reopen the fd-dup false-positive surface (`2>&1`, `>&2`, `1>&2`) — the target character class `[^[:space:]&|;]+` already rejects a leading `&`, so whitespace was never what excluded fd-dup. Verified the trickiest case explicitly: fd-dup written with **zero space either** (`echo err>&2`) stays correctly un-gated.

**Also applied Rex's non-blocking nit** from the same review: the tee multi-operand loop now stops consuming tokens once a redirection operator appears, so `tee a b 2> err.log` no longer emits the literal string `"2>"` as a spurious target. This was fail-closed (over-strict, not a bypass) either way — the real redirect target was always independently caught by the redirection loop — so it's a pure tightening, one guard clause.

**Exhaustive re-verification** — every repro from all four review rounds, end-to-end through `require-active-ticket.sh`:

| Round | Repro | Result |
|---|---|---|
| 1/2 | `;> file`, `;> file cmd`, `&&> file`, `\|> file`, `\|\|> file` | All block |
| 3 | `&> file`, `&>> file`, `;>\| file`, `;>>\| file` | All block |
| 4 | `>file`, `2>file`, `>>file`, `>\|file`, `&>file` (all no-space) | All block |
| sanity | `2>&1`, `>&2`, `1>&2`, `>&2` (no-space), `2>&1;true`, `<`, `<<<` | None gated |

Added adversarial no-whitespace rows at extraction level and both gate levels for every operator; each fails against the pre-this-fix code and passes after (confirmed via `git stash`). All three suites green (133 + 71 + 25), `bash -n` and `shellcheck -S warning` clean, diff vs the `upstream/dev` merge-base still exactly the 6 intended files.

atlas-apex (commented):
## Security Review: PR #926 (pass 4)

**Commit**: `2a4b7bb7c5b1d9765ab55a1f513c1e7fea8cb54e`

### Verdict: **CHANGES REQUESTED** — one NEW in-scope redirect bypass survives. No marker written.

The redirect-operator "unjudged write target" class is **not yet fully closed**. Prior rounds 1, 3, 4 and the leading-`>` (round 2) fixes all verify correct, and there are **zero false positives** — but a sibling of the `&&>` case the PR explicitly claims to handle is still open: **`|`/`||`-adjacent single-`>` redirect**.

---

### Surviving bypass — pipe-adjacent redirect

These are **real writes** that both `require-active-ticket.sh` and `require-migration-ticket.sh` let through **un-gated** (exit 0) at HEAD `2a4b7bb7c5b1d9765ab55a1f513c1e7fea8cb54e`:

```
false ||> src/app.ts          # exit 0  — should BLOCK (truncates src/app.ts)
false ||>| src/app.ts         # exit 0  — should BLOCK (force-clobber truncate)
echo x |> src/app.ts          # exit 0  — should BLOCK
```

Real-bash proof (`||>` and `||>|` unambiguously truncate — version-independent):

```
$ printf 'ORIG\n' > victim.ts; false ||> victim.ts; cat victim.ts
        # file is now EMPTY — bash ran the ||-branch: `> victim.ts` truncated it
```

Same bypass on the migration hook (target `app/migrations/v1.sql` → exit 0 instead of 2).

**Not affected (already block, kept for contrast):** `false || > src/app.ts` (space → OK), `false ||>> src/app.ts` (doubled `>>` accidentally rescued), `;>` / `&&>` (round 2 + the `&>` substring branch).

### Root cause

The **extraction** path was correctly fixed to split on top-level separators (`&&`, `||`, `;`, `|`) *then* anchor `^` per segment — `bash_extract_write_targets 'false ||> src/app.ts'` correctly returns `src/app.ts`. But the **detection** gate short-circuits *before* extraction ever runs:

`bash_command_appears_to_write()` calls `_bdw_match_redirection()` on the **whole, unsplit** command. Its pattern

```
(&>>?|(^|[^|<&])>>?\|?)[[:space:]]*[^[:space:]&|;]+
```

recognises a `>` only at `^` (start of the *whole string*), after a char in `[^|<&]`, or as `&>`. In `||>` the `>` is mid-command, preceded by `|` (excluded by `[^|<&]`), and is not `&>` → no match → "read-only" → hook exits 0. `;>` survives because `;` is *not* in the exclusion class; `&&>` survives only by luck (the substring `&>` hits the `&>>?` branch). `|`/`||` have **no** rescue. Detection and extraction are inconsistent: extraction segments, detection doesn't.

### Fix direction

Make detection use the **same** top-level-separator segmentation as `bash_extract_write_targets`, then run `_bdw_match_redirection` **per segment** so a leading `>` sits at `^`. That closes `|>` / `||>` / `||>|` and makes the two paths principled rather than accidentally-aligned. (Add a test for each; `||>` is the direct sibling of the `&&>` case already in the suite.)

### Minor related note (same root cause, low priority, your call whether to fold in)

`cmd <> file` (read-write open) also has its `>` preceded by `<` (excluded) and is missed — but `<>` doesn't truncate, is exotic in agent-generated commands, and isn't a separator so the segmentation fix won't reach it. Flagging for completeness only; I would **not** block on it.

---

### What I verified PASSES at this HEAD

| Class | Result |
|-------|--------|
| Round 1 — multi-target (exempt/out-of-gov FIRST, in-repo second) | BLOCK ✓ |
| Round 2 — leading `>` after `;` / `&&`, spaced & no-space | BLOCK ✓ |
| Round 3 — `&>`, `&>>`, `>|`, `>>|` (spaced) | BLOCK ✓ |
| Round 4 — zero-whitespace `>file`, `2>file`, `&>file`, `>\|file`, `>>\|file` | BLOCK ✓ |
| tee single + multi-operand (2nd in-repo) + `tee -a` | BLOCK ✓ |
| fd-dup `echo err>&2`, `2>&1`, `>&2`, `1>&2` | pass un-gated ✓ (no false positive) |
| reads `<`, `<file`, `<<EOF`, `<<<` | pass un-gated ✓ |
| exempt `.md` / `.claude/` / out-of-gov `/tmp` / deletion-only `rm` | pass un-gated ✓ |

Both hooks were exercised end-to-end (JSON on stdin, governed sandbox, no active ticket, pin disabled). Everything except the pipe-adjacency family behaves correctly.

### Checklist Results
- Injection / bypass prevention: **Fail** (pipe-adjacent redirect bypass)
- False-positive safety (fd-dup, reads): **Pass**
- Secrets / XSS / auth / data / API: n/a (hook-logic change)

Re-request review after the detection path is segmented; I'll re-run the full four-round matrix plus the new `|`/`||` cases.

---
📌 Reviewed commit: `2a4b7bb7c5b1d9765ab55a1f513c1e7fea8cb54e`

atlas-apex (commented):
## Code Review: PR #926 — round 4 (final)

**Commit**: `2a4b7bb7c5b1d9765ab55a1f513c1e7fea8cb54e`

### Verdict: APPROVED

Round-4 relaxation of the mandatory post-operator whitespace (`[[:space:]]+` → `[[:space:]]*`) is correct and complete. All five focus items verified empirically under bash 3.2 (the shell hooks actually run under), plus the full suites: **133 / 71 / 25, all green** at this HEAD.

### Focus-item results

**1. `+`→`*` applied at all sites, correct.** Five physical locations, all changed uniformly, none missed:
- `_bdw_match_redirection` matcher
- both extraction greps (`bash_extract_write_target`, `_bdw_targets_from_segment`)
- both strip-seds (`s/^[^>]*>>?\|?[[:space:]]*//`)

Confirmed via grep that no `>>?…[[:space:]]+` survives anywhere in the lib.

**2. fd-dup false positives NOT reopened — and for the right reason.** `2>&1`, `>&2`, `1>&2`, and the zero-space `err>&2` all correctly extract to empty / not-a-write. Verified the exclusion is carried by the target character class `[^[:space:]&|;]+` rejecting a leading `&`, NOT by the (now-optional) whitespace: the zero-whitespace fd-dup form `err>&2` still yields no target. The `&>` vs `>&` ordering argument in the comments holds — `&>` (write) has `&` before `>`; fd-dup has `>` before `&`. The two new sanity rows (`echo err>&2`, `make build 2>&1;true` → `""`) pin this.

**3. Reads still ungated.** `<`, `<<`, `<<<`, `<(…)`, and zero-space `cat<file` all not-write. No `<` operator anywhere in the redirect alternation.

**4. tee-loop fix sound.** `echo x | tee a b 2> err.log` → `err.log`, `a`, `b` (operands captured, redirect captured, no spurious literal `2>` token). Break-on-redirect-operator is fail-closed — the real redirect target is independently captured by the redirection loop, so the break can only tighten, never miss. Verified nothing is dropped for `tee src/x.ts src/y.ts 2> /tmp/log`.

**5. New zero-whitespace rows genuinely gate.** Ran the round-4 test file against the round-3 lib: the 5 Hakim-repro rows (`echo hi>…`, `2>…`, `>>…`, `>|…`, `&>…`) FAIL on round-3 and PASS on round-4 — they exercise exactly the relaxed behaviour, not vacuously.

### Suggestions (non-blocking, follow-up)

- **nit — process substitution `>(…)` over-match (new, fail-closed).** The optional-whitespace change newly flags the attached output-process-substitution form as a write: `diff a >(sort)` was not-write on round 3, is write (target `(sort)`) on round 4. It's fail-closed (over-blocks a rare construct, no bypass), input `<(…)` is unaffected, and the space-separated `> (…)` variant was already flagged pre-PR — so this is a minor extension of an existing imperfection rather than a new class. If cheap, consider excluding a `(` immediately after the operator (or adding `(` to the target-class exclusion) and adding a `>(…)` / `<(…)` regression row, since process substitution is currently untested in either direction. Not a merge blocker.
- **nit — tee break-list.** The operand-stop check `^[0-9]*(>>?|<<?)` doesn't include `&>`; `tee a b &> log` would over-capture the `&>` token. Same fail-closed, rare-shape category as above; the real target `log` is still caught by the redirection loop.

Neither finding creates a false negative / bypass — both are fail-closed. The PR's objective (close the zero-space write bypass without reopening fd-dup) is fully and correctly met.

---
📌 Reviewed commit: `2a4b7bb7c5b1d9765ab55a1f513c1e7fea8cb54e`

atlas-apex:
Pushed the structural close (`e81442b`) — Hakim's fifth adversarial re-hunt found the actual root cause behind rounds 1-5, not another operator variant.

**The real bug:** `bash_command_appears_to_write` (detection) ran the redirection matcher on the WHOLE, unsplit command. `bash_extract_write_targets` (extraction) already split into segments first, then anchored `^` per segment. For `false ||> src/app.ts` — a real, truncating write — the `>` sits immediately after `|` (from `||`), which the leading-context class excludes (needed for `2>&1`/`>&2`), and it isn't at `^` either since detection never split. So detection said "not a write" while extraction, given the identical command, correctly found the target. **Detection and extraction disagreed** — that's the bug. `;>` and `&&>` only survived earlier rounds by coincidence (`;` isn't excluded by the class; `&&>` happens to contain the substring `&>`); `|`/`||` had no such rescue.

**The fix:** extracted the segmentation into one shared function (`_bdw_split_top_level`) and made both paths call it. `_bdw_match_redirection_any_segment` (new) splits first, matches per segment — exactly mirroring what extraction already did. They now share one source of truth and can't diverge again.

**Bonus catch while fixing this:** `bash_command_is_deletion_only` had the identical bug at its own direct `_bdw_match_redirection` call — `rm old.ts; false ||> src/app.ts` was being wrongly classified as deletion-only (a real write hiding next to an `rm`), which would have exempted it from the gate even after the detection-path fix alone. Fixed the same way.

**Out of scope, documented:** `cmd <> file` (read-write open) — exotic, non-truncating, and not adjacent to any of the `&&`/`||`/`;`/`|` separators this library segments on, so the structural fix doesn't reach it. One-line comment marking it as a known follow-up, per your call to file it separately.

**Every repro from all five rounds, re-verified end-to-end:**

| Round | Forms | Result |
|---|---|---|
| 1–2 | `;>`, `&&>`, `\|>`, `\|\|>` (spaced + no-space) | All block |
| 3 | `&>`, `&>>`, `>\|`, `>>\|` | All block |
| 4 | zero-whitespace variants of all of the above | All block |
| 5 (this round) | `\|\|>`, `\|\|>\|`, `\|>`, and `rm; \|\|>` | All block |
| sanity | `2>&1`, `>&2`, `1>&2`, no-space fd-dup, `<`, `<<<`, pipe-adjacent fd-dup (`\| cat 2>&1`, `\|\| ... >&2`) | None gated |

Every new test fails against the pre-this-fix code and passes after (confirmed via `git stash`). All three suites green (142 + 78 + 26), `bash -n` and `shellcheck -S warning` clean, diff vs the `upstream/dev` merge-base still exactly the 6 intended files.

atlas-apex (commented):
## Security Review: PR #926 (Hakim — pass 5 of 5)

**Commit**: `e81442b`

### Summary
Round-5 fix unifies the segmentation used by **detection**, **extraction**, and **deletion-only** classification. `_bdw_split_top_level` is now the single source of truth for "where does one command end and the next begin", and all three paths call it. This structurally closes the redirect-operator × separator-adjacency class that survived rounds 1–4 — the root cause was never a missing operator, it was detection matching the *whole, unsplit* command (where `[^|<&]` excludes a `|`/`||`-preceded `>`) while extraction split *first*. Split-before-match with one shared splitter means the two can't diverge again.

### Checklist Results
- Secrets & Credentials:  Pass (n/a)
- Injection Prevention:   Pass — this hook IS the write-gate; fix hardens it
- Auth & Authorisation:   Pass — trust-chain hook, tightened
- API Security:           Pass (n/a)

### Verification — full five-round matrix on HEAD `e81442b`

Ran against `require-active-ticket.sh` and `require-migration-ticket.sh` (both consume the identical `bash_command_appears_to_write` / `bash_extract_write_targets`), plus the repo's own suites in an isolated worktree.

**Writes — all BLOCK, target correctly extracted:**
- Round-5 class `|>` / `||>` / `||>|` / `|>>` / `||>>` — BLOCK
- Leading `>` after every separator (`;`, `&&`, `||`, `|`), no-space AND spaced — BLOCK
- `&>`, `&>>`, `>|`, `>>|` — BLOCK
- Zero-whitespace (`echo hi>f`, `b>>f`, `b>|f`, `b&>f`, `2>f`) — BLOCK
- Multi-target (out-of-repo first, in-repo second) — BLOCK, both targets surfaced
- `tee a b c`, `tee f 2> err.log` — BLOCK (operator token not mis-emitted as a target)
- Newline-separated commands, force-clobber adjacent to a real pipe, fd-dup *then* real write — BLOCK

**Deletion-only path (the round-5 regression fix):**
- `rm old.ts; false ||> src/app.ts`, `rm x; echo y >| f`, `rm x && echo y >| f`, `rm x;> f`, `rm x | tee f` — all **BLOCK** (correctly NOT classified deletion-only)
- Genuine deletions (`rm x`, `rm -rf build/`, `rm a; rm b`) — still exempt

**No false-positives — reads/fd-dups stay un-gated:**
- `>&2`, `2>&1`, `1>&2`, `err>&2`, `echo x|cat 2>&1`, `false||echo err >&2`, chained fd-dup pipelines — PASS-READ
- `<`, `<<`, `<<<` reads — PASS-READ

The fd-dup exclusion is load-bearing on the *target character class* (`[^[:space:]&|;]+` rejects a leading `&`), not on mandatory whitespace — so relaxing whitespace to optional (round 4) correctly did not reopen it, confirmed empirically.

**Suites:** `test_detect_bash_write.sh` 142/142, `test_require_active_ticket_bash.sh` 78/78, `test_require_migration_ticket.sh` 26/26. My independent 69-case matrix: 0 failures.

### Adversarial verdict — is the class closed?
**Yes.** With detection and extraction sharing one segmentation and the deletion-only path routed through the same `_bdw_match_redirection_any_segment`, every redirect operator adjacent to every top-level separator now resolves identically across all three consumers. I could not construct an in-scope redirect bypass. The only residual is `<>` (read-write open) — non-truncating (materially different risk shape), non-separator-adjacent (the round-5 split doesn't reach it), and explicitly documented as an out-of-scope follow-up. Per the brief I did not block on it.

No escalation to a redirect-lexer is warranted. The structural fix is the correct shape.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `e81442b`

atlas-apex (commented):
## Code Review: PR #926 — round 5 (STRUCTURAL)

**Commit**: `e81442be9f19f20a01a28ffbd328658bf79de4dc`

### Summary
Round 5 stops patching individual redirect operators and fixes the root cause: detection, extraction, and the deletion-only check were segmenting the command differently and could disagree. Segmentation is now pulled into one shared `_bdw_split_top_level` that every path calls — one source of truth. Verified against the live PR HEAD `e81442b` (my local `origin/dev` is stale; I fetched `pull/926/head` and ran in a detached worktree at that SHA).

### Focus-item verification

**1. `_bdw_split_top_level` correct + all three consumers use it.** Confirmed. Detection (`bash_command_appears_to_write:437`) and deletion-check (`bash_command_is_deletion_only:489`) both route through `_bdw_match_redirection_any_segment`, which splits via `_bdw_split_top_level:248`. Extraction (`bash_extract_write_targets:784`) splits via the same function directly. The only bare `_bdw_match_redirection` call (line 247) operates on an already-split segment. No path runs the matcher on the unsplit command anymore. The clobber-protection (`>>|`/`>|` placeholder round-trip, longest-match-first) correctly shields force-clobber operators from the bare-`|` split, and uses parameter-expansion not sed (BSD-sed `\n`-safe).

**2. Deletion-only fix sound.** `rm x; false ||> src/app.ts` → classified NOT deletion-only (gated). The segment-aware redirection check at line 489 catches the write hiding behind the `||`-adjacent `>`. Bare `rm old.ts` still correctly exempt.

**3. No regression on fd-dups / reads.** All un-gated under bash: `2>&1`, `>&2`, `err>&2`, `1>&2`, `echo x | cat 2>&1`, `false || echo err >&2`, `<`, `<<`, `<<<`. The target character class `[^[:space:]&|;]+` (rejecting a leading `&`) — not mandatory whitespace — is what keeps fd-dup unmatched, so the round-4 `[[:space:]]*` relaxation doesn't reopen it.

**4. New `|>`/`||>`/`||>|` tests gate at all three levels and are load-bearing.** The suite asserts extraction (`assert_targets` → `src/app.ts`), detection (`assert_appears_to_write`), and deletion-only (`assert_not_deletion_only`) for each. Confirmed load-bearing: simulating the pre-fix whole-command match, `_bdw_match_redirection` MISSES all three (the bug), while `_bdw_match_redirection_any_segment` MATCHES (the fix).

**5. `>(…)` process-sub over-block.** Still fail-closed (acceptable) and unchanged by the refactor — `tee >(cat)` is still detected as a write and extracts a garbage-ish path that fails closed into the gate. The `<>` read-write gap is documented as an explicit out-of-scope follow-up (non-truncating, not separator-adjacent, so the round-5 split doesn't reach it). Both are correct fail-closed postures.

### Checklist Results
- Code Quality:              Pass (thorough inline rationale tracing rounds 1-5; `tee` operand-stop at `2>` from my last round's flag is retained)
- Testing:                   Pass — 142 / 78 / 26 all green at HEAD; new cases fail pre-fix
- Security:                  Pass — closes the |/||-adjacent redirect bypass at the structural level
- PR Description & Glossary: Pass (glossary present, summary bullets narrative)

### Suggestions
- nit: the PR **body** is stale — it still describes the #886 round-1 core (`bash_extract_write_targets` / AND-semantics / 108-48-21 counts) rather than this round-5 structural unification (142/78/26). Worth refreshing the body before merge so the description matches the code. Non-blocking.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `e81442be9f19f20a01a28ffbd328658bf79de4dc`

PR #932, opened 2026-07-16
fix(#931): close <> read-write + >(…) process-sub residuals in bash write-detection

### LINKED ISSUE #931: [Task] Bash write-detection residuals: <> read-write open + >(…) process-sub over-block (follow-up to #886)
## Driver
Follow-up from the #886 5-round hardening (PR #926). Two redirect-detection edge cases were identified during review, documented in code comments, and consciously deferred as out-of-scope for #886 (neither is a false-negative bypass of the ticket gate on a common construct). Filing so they're tracked, not lost.

## Scope
Two low-priority residuals in `.claude/hooks/_lib-detect-bash-write.sh`:
1. **`cmd <> file` (read-write open) is not detected as a write.** The `<>` operator opens a file read-write; the round-5 segmentation fix genuinely can't reach it (`>` preceded by `<`, non-truncating, not adjacent to a top-level separator). Rare in agent commands and non-truncating, but technically a write the gate misses. Decide: detect it as a write, or accept + document as a permanent known-limitation.
2. **`>(…)` process substitution is over-blocked (fail-closed).** After the round-4 whitespace relaxation, `diff a >(sort)` is flagged as a write to a spurious target. This is fail-closed (over-blocks a rare construct, never a bypass), so lower priority — but a `>(…)`/`<(…)` regression-test row + a targeted exclusion would tighten it.

## Acceptance Criteria
- [ ] Decide + implement (or document-as-accepted) `<>` read-write detection
- [ ] Add `>(…)` / `<(…)` process-substitution regression rows to the test suite
- [ ] If tightening `>(…)`: it no longer over-captures, and no fd-dup/read regresses
- [ ] Both cases covered by tests

## Glossary
| Term | Definition |
|------|------------|
| `<>` | Bash operator opening a file for both reading and writing |
| Process substitution | `>(cmd)` / `<(cmd)` — a construct that looks like a redirect but pipes to/from a command, not a file |
| Fail-closed | Over-blocking (safe) rather than under-blocking (a bypass) |

atlas-apex:
## Summary
- **`<>` read-write open is now detected as a write** — `[n]<>word` opens a file for both reading and writing (`exec 3<> file`, `cmd <>file`). Every prior #886/#926 hardening round missed it because the leading `<` was excluded by the *other* operators' own leading-context class, so `<>` was structurally invisible to the matcher, not just an untested edge case. Decided to detect (not document-as-accepted) per the ticket's own framing — the ticket gate cares whether tracked content was touched, not whether the write truncates.
- **`>(…)` / `<(…)` process substitution no longer fabricates a bogus write target** — `diff a >(sort)` previously matched `(sort)` as if it were a filename, needlessly gating a common, harmless construct (fail-closed over-block, never a bypass). A `>(` immediately after the operator is now excluded from the target class, while a filename that merely *contains* a non-leading paren (`file(1).txt`) still extracts correctly.
- **Both fixes are one shared regex change** — a new `[0-9]*<>` operator alternative plus a leading-`(` exclusion on the target character class — applied identically across detection (`_bdw_match_redirection`), single-target extraction (`bash_extract_write_target`), and multi-target extraction (`_bdw_targets_from_segment`), so the three paths can't disagree the way earlier #886/#926 rounds did.
- **Full #886/#926 regression matrix re-verified** — all fd-dup exclusions (`2>&1`, `>&2`, `1>&2`), force-clobber (`>|`, `>>|`), both-streams (`&>`, `&>>`), and the `|`/`||`-adjacent redirect cases from round 5 still pass unchanged.

## Testing
- `bash .claude/hooks/tests/test_detect_bash_write.sh` — 160/160 pass (142 pre-existing + 18 new rows covering `<>` detection/extraction, `>(…)`/`<(…)` exclusion, the exact chained repro from the ticket, and sanity checks that the new exclusions don't resurrect any fd-dup/force-clobber/both-streams false-negative).
- `bash .claude/hooks/tests/test_require_active_ticket_bash.sh` — 78/78 pass (unchanged, sourced-lib regression check).
- `bash .claude/hooks/tests/test_require_migration_ticket.sh` — 26/26 pass (unchanged, sourced-lib regression check).
- `bash -n` on both changed files — clean.
- `shellcheck -S warning` on both changed files plus the two hook callers (`require-active-ticket.sh`, `require-migration-ticket.sh`) — clean.

Closes #931

---

## Glossary
| Term | Definition |
|------|------------|
| `<>` (read-write open) | Bash operator (`[n]<>word`) that opens `word` for both reading and writing on descriptor `n` (or fd 0 if omitted) — a real, non-truncating write. |
| Process substitution | `>(cmd)` / `<(cmd)` — bash wires `cmd`'s stdin/stdout to a fifo/fd path and substitutes that path as an argument; it looks adjacent to a redirect but is a command, not a file target. |
| Fail-closed | Over-blocking (safe but annoying) rather than under-blocking (a real bypass). The process-substitution bug was fail-closed; the `<>` gap was a genuine miss. |
| Leading-context class | The regex character class (`[^|<&]`) used to distinguish a real redirect `>` from one that's part of `2>&1`/`>&2` (fd-dup) or a heredoc marker. |

### DIFF
--- a/.claude/hooks/_lib-detect-bash-write.sh
+++ b/.claude/hooks/_lib-detect-bash-write.sh
@@ -59,7 +59,13 @@
 #
 #   Redirection / pipes-to-disk
 #     - cmd > file, cmd >> file, cmd 2> file
+#     - cmd &> file, cmd >| file (force-clobber), cmd <> file (read-write
+#       open, #931)
 #     - tee
+#
+#   NOT matched as a write (deliberate exclusion, #931):
+#     - cmd >(subshell), cmd <(subshell) — process substitution. Looks
+#       adjacent to a redirect but is a command, not a file target.
 #
 #   In-place text editors
 #     - sed -i (in-place edit, GNU + BSD `''` form)
@@ -163,8 +169,46 @@ _bdw_starts_with_git_subcommand() {
 # forms), the target can't start there regardless of how much whitespace
 # came before it. The whitespace requirement was never what excluded
 # fd-dup; the target character class was and still is.
+#
+# `<>` READ-WRITE OPEN (apexyard#886/#931 round 6, closing one of the two
+# residuals #926 documented-and-deferred): `[n]<>word` opens `word` for
+# BOTH reading and writing on descriptor `n` (or fd 0 if `n` is omitted) —
+# `exec 3<> some/file`, `cmd <>file` are real, non-truncating writes that
+# every prior round's operator alternation missed entirely (the `<`
+# leading character was always excluded by the OTHER operators' own
+# leading-context class, `[^|<&]`, so `<>` was structurally invisible to
+# this pattern, not just an uncovered edge case). The new `[0-9]*<>`
+# alternative is unambiguous against every existing operator: `<>` is two
+# literal characters (`<` immediately followed by `>`), which never
+# occurs inside `<<` (heredoc), `<<<` (herestring), or any of the
+# fd-dup/`>`-based forms above (those all have `>` preceding `&`, never
+# `<` immediately preceding `>`). Chosen to DETECT rather than
+# document-as-accepted (per #931's own framing: "a write is a write") —
+# the non-truncating risk profile differs from `>`/`>>`, but the ticket
+# gate cares about "was tracked content touched", not "was it truncated".
+#
+# `>(…)` / `<(…)` PROCESS SUBSTITUTION now EXCLUDED from the target class
+# (apexyard#886/#931 round 6, closing the second #926-deferred residual):
+# `diff a >(sort)` is not a file write — `>(sort)` is process substitution,
+# a subshell whose stdin bash wires to a fifo/fd path, syntactically
+# adjacent to `>` but semantically a command, not a redirect target. The
+# prior target class `[^[:space:]&|;]+` allowed a leading `(` and so
+# happily consumed `(sort)` as if it were a filename — a fail-closed
+# (over-blocking) false positive, not a bypass, but still worth tightening
+# per #931. Splitting the target class into a first-char exclusion,
+# `[^[:space:]&|;(]`, plus the unchanged `[^[:space:]&|;]*` for the rest,
+# means the match FAILS whenever the character immediately after the
+# operator (past any optional whitespace) is `(` — process substitution
+# always starts there with zero space (`> (foo)` is not valid bash
+# redirect-to-subshell syntax; a space there is a syntax error, so
+# excluding the immediately-adjacent case costs nothing on real commands).
+# A legitimate filename that merely CONTAINS a paren not in the leading
+# position (`file(1).txt`) is untouched — only a LEADING `(` is excluded.
+# `<(…)` (process substitution on the read side) was never matched here
+# to begin with (`<` alone isn't a write operator in this file), so no
+# change was needed for that half.
 _bdw_match_redirection() {
-  echo "$1" | grep -qE '(&>>?|(^|[^|<&])>>?\|?)[[:space:]]*[^[:space:]&|;]+'
+  echo "$1" | grep -qE '(&>>?|(^|[^|<&])>>?\|?|[0-9]*<>)[[:space:]]*[^[:space:]&|;(][^[:space:]&|;]*'
 }
 
 # ------------------------------------------------------------------------------
@@ -522,6 +566,7 @@ bash_command_is_deletion_only() {
 #   - mv src /path/to/dst        → /path/to/dst (#153)
 #   - curl -o /path URL          → /path        (#153)
 #   - wget -O /path URL          → /path        (#153)
+#   - cmd <> /path/to/file       → /path/to/file (read-write open, #931)
 #
 # Does NOT handle (returns empty):
 #   - python/node/ruby/perl/php with embedded path
@@ -529,21 +574,17 @@ bash_command_is_deletion_only() {
 #   - cmd with multiple redirects
 #   - paths constructed from variables
 #   - tar -x (target is a directory, often implicit)
-#   - `cmd <> file` (read-write open, apexyard#886/#926 round 5): known,
-#     accepted, OUT-OF-SCOPE gap — exotic, non-truncating (unlike every
-#     operator this file DOES model, `<>` opens for read+write WITHOUT
-#     truncating existing content, so it's a materially different risk
-#     shape), and not adjacent to any of the &&/||/;/| separators this
-#     library segments on, so the round-5 structural fix doesn't reach
-#     it either. Tracked as a separate follow-up rather than folded into
-#     this operator-class close.
+#   - `cmd >(subshell)` / `cmd <(subshell)` process substitution: NOT a
+#     target at all (correctly excluded, #931 — see the leading-`(`
+#     exclusion note on _bdw_match_redirection above), so this returns
+#     empty for `diff a >(sort)` rather than fabricating `(sort)`.
 # ------------------------------------------------------------------------------
 bash_extract_write_target() {
   local cmd="$1"
   [ -z "$cmd" ] && return 0
 
   # Output redirection: capture the first target after >, >>, &>, &>>, >|,
-  # or >>|. Strip leading number/ampersand for cases like `2> file` /
+  # >>|, or <>. Strip leading number/ampersand for cases like `2> file` /
   # `&> file`.
   #
   # (^|[^|<&]) — see the identical note on _bdw_match_redirection above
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #932

**Commit**: `89a034ce1e0e4873af6f3a5f6aba0fb884e9d08e`

### Summary
Fourth round in the #886→#926→#931 hardening line on `_lib-detect-bash-write.sh`. Two residuals closed with one shared regex change: (1) `<>` read-write open (`exec 3<> file`) is now **detected** as a write — previously structurally invisible because the leading `<` was excluded by the other operators' `[^|<&]` leading-context class; (2) `>(…)` / `<(…)` process substitution is now **excluded** from the target class — previously `diff a >(sort)` fabricated `(sort)` as a bogus write target. Scope is exactly the lib + its test.

### Checklist Results
- Security:                 Pass (this IS a security-gate lib — reviewed as such)
- Performance:              Pass (regex-only, no new subprocess)
- Summary Bullet Narrative:  Pass (bullets are full what+why sentences)
- Technical Decisions (AgDR): N/A (no new dependency/pattern; detect-vs-document choice is a within-ticket call, rationalised inline)
- Adopter Handbooks:        N/A (no handbooks loaded; shell lib, no language handbook match)

### Verification performed
Ran the change in an isolated worktree at HEAD `89a034c`, not just trusting the PR's own numbers:

**(a) "Three places agree" (the #926 drift failure mode).** The identical regex — `|[0-9]*<>` alternative + target class split into `[^[:space:]&|;(][^[:space:]&|;]*` — is applied verbatim in all three sites: `_bdw_match_redirection` (detection), `bash_extract_write_target` (single-target), and `_bdw_targets_from_segment` (multi-target). No drift.

**(b) No new bypass.** The only new exclusion is a *leading unquoted* `(` as the first target char. That is never a legitimate write target in bash (a bare leading `(` opens a subshell), so nothing that should gate stops gating. Independent probes confirm plain `>`, `>>` append, `>|` force-clobber, and `<>` all still gate; `2>&1` / `>&2` fd-dup still excluded. The `[0-9]*<>` alternative is unambiguous against `<<`/`<<<` (no `<` immediately followed by `>`).

**(c) No over-block regression on common constructs.** Process substitution (`diff a >(sort)`, `cat <(sort a)`, `cmd > (sort)`) now correctly reads as non-write — this *reduces* over-blocking. `file(1).txt` (non-leading paren) still extracts correctly. Real-write + procsub-tail (`echo a > /tmp/ok;diff x >(sort)`) surfaces only the real target `/tmp/ok`.

**(d) Test coverage.** 18 new rows (142→160). They cover `<>` detection (fd-numbered, no-fd, fd+space), `<>` target extraction (single + multi), heredoc/herestring non-false-positive, rm+`<>` not deletion-only, `>(…)`/`<(…)` exclusion, the exact chained repro, the `file(1).txt` non-leading-paren case, and a fd-dup/force-clobber/both-streams non-regression sweep. All 160 pass. `test_require_active_ticket_bash.sh` 78/78 and `test_require_migration_ticket.sh` 26/26 also green (sourced-lib regression).

**(e) Static analysis.** `bash -n` clean on both files; `shellcheck -S warning` exit 0 on both.

Merge target is `dev` (correct). Glossary present and accurate. Summary bullets are narrative.

- `nit:` A literal `<>` inside a quoted string now over-blocks — `echo "<>"` is detected as a write to target `"`. This is fail-closed (over-block, never a bypass) and is the *same* long-standing regex-on-quoted-strings limitation the pre-existing `>` handling already has (`echo "a > b"` over-blocks identically today). Not a regression introduced by this PR and out of scope; noting only so it's on record if a future round tackles quote-aware tokenisation.

### Verdict
**APPROVED**

---
Reviewed commit: `89a034ce1e0e4873af6f3a5f6aba0fb884e9d08e`

atlas-apex (commented):
## Security Review: PR #932

**Commit**: `89a034ce1e0e4873af6f3a5f6aba0fb884e9d08e`

Trust-chain review (#777) — this diff touches `_lib-detect-bash-write.sh`, the write-detection library backing the `require-active-ticket.sh` and `require-migration-ticket.sh` pre-build gates. A defect here is a gate-bypass surface, so the bar is: does #931 add any fail-**open** path? Answer: **no.** It strictly *adds* detection (`<>`) and removes a *fabricated target* on a non-writing command (`>(sort)`), with no reduction in WRITE verdicts on anything that actually writes.

### Checklist Results
- Secrets & Credentials: Pass (no change)
- Injection Prevention: Pass (no eval/shell-interpolation change; the lib matches, never executes)
- Auth & Authorisation: Pass — gate engagement unchanged; WRITE set is a strict superset of pre-#931
- Data Protection: Pass
- **Bypass surface (the load-bearing check): Pass** — see below

### Bypass hunt — `<>` read-write open
Detection fires WRITE on **every** `[n]<>` form I could construct. Probed directly against the sourced lib at HEAD:

| Form | Result |
|------|--------|
| `cmd <>file.txt` / `cmd <> file.txt` | WRITE, target `file.txt` |
| `exec 3<> /tmp/x`, `exec 10<> …` (multi-digit fd) | WRITE |
| `exec {myfd}<> src/app.ts` (bash named fd) | WRITE, target `src/app.ts` |
| chained after `;`, `&&`, `\|` | WRITE |
| mid multi-segment `a=1; cmd <> tracked.md; echo done` | WRITE |
| no-space packed `…;exec 3<>secret.env` | WRITE, `secret.env` in multi-target list |
| tab-separated target | WRITE |

The `[0-9]*<>` alternative correctly needs no leading-context guard: `<>` (a `<` immediately followed by `>`) never occurs inside `<<` (heredoc), `<<<` (herestring), or any `>&`/`&>` fd-dup form — empirically confirmed `cat <<<word` and `cat <<EOF` still classify as `read`. **No `<>` bypass found.**

### Bypass hunt — the `>(…)` EXCLUSION (the dangerous direction)
An exclusion is where a hole would hide. It does **not** open one. The exclusion is strictly a leading-unquoted-`(` first-char rejection, and the two ways a process-sub can *actually write* are still caught by their own matchers, fail-closed:

| Command | Result | Why safe |
|---------|--------|----------|
| `diff a >(sort)` | **read**, no target | intended fix — `sort` doesn't write; no more bogus `(sort)` |
| `diff a >(tee tracked.txt)` | **WRITE** | caught by the independent `tee` matcher, not the redirect matcher |
| `diff a >(cat > real.txt)` | **WRITE** | inner `> real.txt` caught by the redirect matcher |
| `echo hi > "(weird).txt"` | **WRITE**, target `(weird).txt` | quoted paren-leading filename still extracts (quote shields the `(`) |
| `echo hi > (foo)` | n/a | **bash syntax error** — excluding leading-`(` costs nothing on any valid command |

`>(` *always* initiates process substitution in bash; `> (` with a space is a syntax error. So excluding an unquoted leading `(` after the operator is semantically exact, and any real write happening *inside* the subshell (`tee f`, inner `>`) is independently detected. **The exclusion opens no bypass.**

### Three-places-agree invariant
The regex is byte-identical at all three sites — detection (`_bdw_match_redirection`, L211), single-extract (`bash_extract_write_target`, L620), multi-extract (`_bdw_targets_from_segment`, L772):
```
(&>>?|(^|[^|<&])>>?\|?|[0-9]*<>)[[:space:]]*[^[:space:]&|;(][^[:space:]&|;]*
```
No drift — the round-1–5 failure mode is not reintroduced.

### Fail-open vs fail-closed
Every messy/unextractable target I produced is fail-**closed** (over-inclusive), never fail-open. The gate engages on the WRITE verdict itself (`require-active-ticket.sh:613` `if ! bash_command_appears_to_write`), with target extraction only *refining* which paths — so a target string carrying a trailing `)` (e.g. tee-inside-procsub) blocks more, never less.

### Verification
`bash -n` clean · `shellcheck -S warning` clean (lib + both hook callers) · 160/160 + 78/78 + 26/26 tests pass at HEAD.

### Non-blocking observations (out of scope, PRE-EXISTING — not introduced by #931)
- **`>&file` (csh-style both-streams-to-file, e.g. `ls >&out.txt`) is undetected** — the `>&*` form is excluded as fd-dup and shares syntax with `>&2`. This gap predates #931 (the old target class also excluded a leading `&`), #931 neither introduces nor touches it, and the canonical modern form `&>file` *is* detected. Flagging only as a candidate follow-up ticket for a future hardening round — **not a blocker here.**
- Multi-target extraction is noisy on tee-inside-procsub (trailing `)`, `$'\n'` fragments). Pre-existing, fail-closed, cosmetic.

### Verdict
**APPROVED** — from the security lens. The two #926-deferred residuals are closed correctly: `<>` detection is airtight across every form I could construct, and the `>(…)` exclusion is precisely calibrated to the non-writing process-sub read case with the actually-writing cases still caught fail-closed. No bypass introduced; the three-places-agree invariant holds. (Not writing any approval/merge marker — that's Rex's and the CEO's gate.)

---
📌 Reviewed commit: `89a034ce1e0e4873af6f3a5f6aba0fb884e9d08e`

me2resh (approved):

Issue #890, opened 2026-07-14
[Task] Premium-hook safe-fallback harness — _lib-premium-hook.sh so premium hooks can't break framework users


atlas-apex:
## Driver
Premium-touching hooks (e.g. `reindex-on-session-start.sh`, the #514 apexyard-search self-heal) each HAND-ROLL the same "safe fallback" shape: never block, always exit 0, no-op silently when the premium feature is absent, timeout-guard slow work. Because it's copy-pasted per hook, every new premium hook is a fresh chance to get the safety wrong — which is exactly the framework-blast-radius risk raised on #514. Turn the convention into one audited primitive so premium hooks are safe-by-construction and can NEVER break free/framework users.

## Scope
- New shared lib `.claude/hooks/_lib-premium-hook.sh` exposing a guard, e.g. `premium_hook_run <feature_key> <payload…>`, that centralizes the three safety properties:
  1. **Gate** — is `<feature_key>` enabled in features.yaml AND the component present? If not → return 0 silently (zero output, zero latency for non-premium users).
  2. **Fail-safe execution** — run the payload in a timeout-guarded subshell; swallow non-zero exit; never propagate an error to session start.
  3. **Always exit 0** — a premium hook can never block or break a session, by construction.
- Reuse existing helpers (`_lib-read-config.sh`, the features-gate logic, `_lib-portfolio-paths.sh`).
- **Retrofit** existing premium hooks onto the harness (start with `reindex-on-session-start.sh`; the #514 search self-heal hook is retrofitted once it merges — track as a follow-on).
- Tests under `.claude/hooks/tests/` covering: silent no-op when feature absent; payload runs when enabled+present; a throwing/hanging payload never breaks (exit 0, timeout enforced).
- **AgDR** recording the primitive + the convention that all premium hooks MUST route through it.

## Acceptance Criteria
- [ ] `_lib-premium-hook.sh` exists with the guard, documented, POSIX-bash, `set -u`.
- [ ] A premium hook using the guard is a silent no-op (zero output, exit 0) for a non-premium/unconfigured adopter.
- [ ] A throwing or hanging payload cannot block or fail session start (verified by test).
- [ ] `reindex-on-session-start.sh` retrofitted onto the harness with no behaviour regression.
- [ ] AgDR authored; hook tests pass.

## Glossary
| Term | Definition |
|------|------------|
| Premium-hook harness | The shared guard that gates + fail-safes every premium hook so it can't break non-premium users |
[...]

PR #933, opened 2026-07-16
feat(#929): SessionStart search advisory probes MCP reachability

### LINKED ISSUE #929: [Task] SessionStart search advisory should reflect MCP reachability, not just file presence
## Driver
The SessionStart search-reindex advisory banner (`reindex-on-session-start.sh` / the MCP-reindex advisories) keys off **file presence** — whether the search index/config exists on disk — not off whether the search MCP is actually **reachable**. When search is configured but the backing MCP server is unreachable at session start (e.g. it runs as a separate service that's currently down), the banner can still nudge a reindex that can't succeed, and an agent may assume search is available when it will silently fall back to grep+Read.

## Scope
Make the SessionStart search advisory reflect actual reachability, not just file presence:
- Before nudging a reindex / advertising search, do a cheap liveness probe of the configured search MCP (a fast, bounded, fail-soft check).
- If configured-but-unreachable: emit a clear "search is configured but not reachable — falling back to grep+Read" note instead of a reindex nudge; never block session start (exit 0, same safe advisory shape as the sibling hooks).
- If not configured at all: unchanged (silent no-op).
Keep it tracker-agnostic and fully opt-in / graceful — no behaviour change for adopters without search configured.

## Acceptance Criteria
- [ ] The SessionStart search advisory probes reachability, not just file presence
- [ ] Configured-but-unreachable → an honest "unreachable, degrading to grep+Read" message, not a reindex nudge
- [ ] Not configured → silent no-op (unchanged)
- [ ] Always exit 0 (never blocks session start); probe is bounded/timeout-guarded
- [ ] Test coverage for the three states (reachable / configured-unreachable / not-configured)

## Glossary
| Term | Definition |
|------|------------|
| Search MCP | The code/docs semantic-search server the framework can use in place of grep |
| Liveness probe | A cheap, bounded check of whether a service is actually responding |
| Fail-soft | Degrade quietly (grep+Read) and never block, when a dependency is down |

atlas-apex:
## Summary
- **Adds a reachability probe, not just a presence check** — `reindex-on-session-start.sh` used to gate purely on `command -v apexyard-search` (is the binary on PATH?). A binary that's present but broken (crashed dependency, stale venv, hung process) looked identical to a healthy install, so the hook would nudge a reindex that couldn't succeed and an agent would silently fall back to grep+Read without ever being told search wasn't actually working.
- **New `premium_hook_probe()` in `_lib-premium-hook.sh`** — a bounded liveness-check sibling to the existing fire-and-forget `premium_hook_run()`. Same feature-flag + presence gating, but instead of always returning 0 it returns a three-way result the caller branches on: `0` REACHABLE, `1` UNREACHABLE, `2` NOT_APPLICABLE (disabled/unconfigured). Same timeout-guarded child-process shape as `premium_hook_run`, with its own shorter default timeout (5s) since a liveness check should be fast even when the real payload gets a longer budget.
- **The hook now probes `apexyard-search doctor`** (a fast, side-effect-free index-health report that always exits 0 on a healthy CLI) before nudging a reindex. Reachable → unchanged behaviour. Unreachable → one honest stderr line ("configured but not reachable — falling back to grep+Read") instead of a reindex nudge that can't succeed. Not configured → unchanged silent no-op. Always exits 0 — still purely advisory, never blocks session start.

## Testing
1. `bash .claude/hooks/tests/test_reindex_on_session_start.sh` — 9/9 pass, including 3 new cases for the reachable / configured-unreachable / not-configured states plus a hanging-probe-is-bounded case.
2. `bash .claude/hooks/tests/test_lib_premium_hook.sh` — 15/15 pass, including 6 new cases directly exercising `premium_hook_probe`'s three-way return code and its own timeout guard.
3. `bash bin/run-hook-tests.sh` — 106/107 pass; the one failure (`test_sync_codex_adapter.sh`) is a pre-existing, unrelated environment gap (a missing `block-test.sh` fixture) — reproduced identically with this PR's changes stashed out.
4. `bash -n` on both changed hook files — clean.

Closes #929

---

## Glossary
| Term | Definition |
|------|------------|
| Reachability probe | A cheap, bounded, side-effect-free check of whether a component can actually respond right now — distinct from a presence check, which only confirms it's installed |
| `premium_hook_probe` | New shared-harness function (`_lib-premium-hook.sh`) returning a three-way liveness result: REACHABLE (0) / UNREACHABLE (1) / NOT_APPLICABLE (2) |
| `apexyard-search doctor` | An existing CLI subcommand that reports index-health (freshness) as JSON; reused here purely as a fast, always-exits-0-when-healthy liveness probe |
| Fail-soft | Degrade quietly (fall back to grep+Read) and never block, when a premium dependency is down |

### DIFF
--- a/.claude/hooks/_lib-premium-hook.sh
+++ b/.claude/hooks/_lib-premium-hook.sh
@@ -100,6 +100,35 @@
 #       (works because presence checks run via `eval` in THIS shell, not a
 #       child process — unlike the payload, they inherit this lib's functions.)
 #
+#   premium_hook_probe <feature_key> <presence_check_cmd> <probe_cmd> [default_enabled]
+#       A cheap, bounded LIVENESS check (me2resh/apexyard#929) — for callers
+#       that need to know whether a premium component can actually RESPOND
+#       right now, not just whether it's installed, before deciding what to
+#       advertise. Same GATE 1 (feature flag) / GATE 2 (presence) semantics
+#       as premium_hook_run, and the same timeout-guarded child-process
+#       execution shape, but instead of always returning 0 it returns a
+#       three-way result the caller branches on:
+#         0  REACHABLE     — gate passed AND probe_cmd exited 0 in time.
+#         1  UNREACHABLE    — gate passed but probe_cmd exited non-zero or
+#                            was killed for hanging past the timeout.
+#         2  NOT_APPLICABLE — feature disabled/unconfigured, presence check
+#                            failed, or required args are missing. Same
+#                            "nothing to report on" case premium_hook_run
+#                            treats as a silent no-op.
+#       probe_cmd should be side-effect-free and fast (a liveness check,
+#       not the real premium work) — see reindex-on-session-start.sh for
+#       the worked example (probes `apexyard-search doctor` before nudging
+#       a reindex).
+#
+# Tunables (env):
+#   PREMIUM_HOOK_PROBE_TIMEOUT_SECS   max seconds a premium_hook_probe call
+#                                     spends before killing the probe and
+#                                     treating it as UNREACHABLE (default 5 —
+#                                     short on purpose; a liveness check
+#                                     should be fast even when the caller's
+#                                     own premium_hook_run payload is allowed
+#                                     a longer PREMIUM_HOOK_TIMEOUT_SECS)
+#
 # Source-guard: safe to source more than once in the same shell (idempotent,
 # matches the pattern in _lib-ops-root.sh).
 
@@ -296,4 +325,49 @@ premium_hook_run() {
 
   # ALWAYS 0 — a premium hook can never block or break its caller.
   return 0
+}
+
+# ------------------------------------------------------------------------------
+# Public: premium_hook_probe <feature_key> <presence_check_cmd> <probe_cmd> [default_enabled]
+#   See header for full semantics. Unlike premium_hook_run, this function's
+#   return code IS the signal — 0 (reachable) / 1 (unreachable) /
+#   2 (not applicable) — so callers can branch on it. It still never lets a
+#   hanging probe run unbounded: same timeout-guarded child-process shape as
+#   premium_hook_run, just with its own (shorter-by-default) timeout knob.
+# ------------------------------------------------------------------------------
+premium_hook_probe() {
+  local feature_key="${1:-}"
+  local presence_cmd="${2:-}"
+  local probe_cmd="${3:-}"
+  local default_enabled="${4:-true}"
+
+  # Nothing to probe -> not applicable, same as premium_hook_run's "nothing
+  # to run" early-out.
+  [ -n "$feature_key" ] || return 2
+  [ -n "$probe_cmd" ] || return 2
+
+  # GATE 1: feature flag. Absent/disabled -> not applicable, zero latency.
+  premium_feature_enabled "$feature_key" "$default_enabled" || return 2
+
+  # GATE 2: component presence, when the caller supplied a check.
+  if [ -n "$presence_cmd" ]; then
+    ( eval "$presence_cmd" ) >/dev/null 2>&1 || return 2
+  fi
+
+  # LIVENESS CHECK: timeout-guarded child process, same fallback order as
+  # premium_hook_run (GNU `timeout`, then macOS `gtimeout`, else unbounded).
+  local timeout_secs to
+  timeout_secs="${PREMIUM_HOOK_PROBE_TIMEOUT_SECS:-5}"
+  to=""
+  if command -v timeout >/dev/null 2>&1; then
+    to="timeout -k 2 ${timeout_secs}"
+  elif command -v gtimeout >/dev/null 2>&1; then
+    to="gtimeout -k 2 ${timeout_secs}"
+  fi
+
+  # shellcheck disable=SC2086
+  if $to bash -c "$probe_cmd"; then
+    return 0
+  fi
+  return 1
 }
--- a/.claude/hooks/reindex-on-session-start.sh
+++ b/.claude/hooks/reindex-on-session-start.sh
@@ -7,15 +7,33 @@
 #   - --if-stale:    the CLI no-ops when the index is still fresh, so the common
 #                    case (just reindexed) costs essentially nothing.
 #
+# REACHABILITY, NOT JUST PRESENCE (me2resh/apexyard#929): before nudging a
+# reindex, this hook now probes whether apexyard-search can actually RESPOND
+# — `command -v apexyard-search` only proves the binary is on PATH, not that
+# invoking it works. A binary that's present but broken (missing a runtime
+# dependency, a stale venv, whatever) used to be treated identically to a
+# healthy install: the reindex payload would run, silently fail, and an agent
+# would assume search was live when it was quietly falling back to grep+Read
+# the whole session. The probe closes that gap with a fast, bounded,
+# side-effect-free `apexyard-search doctor` call (an index-health report that
+# always exits 0 when it runs cleanly) via `premium_hook_probe`:
+#   - REACHABLE      -> unchanged behaviour, opportunistic reindex runs.
+#   - UNREACHABLE     -> one honest stderr line ("configured but not
+#                        reachable — falling back to grep+Read") INSTEAD of
+#                        the reindex nudge. Still exits 0 — advisory only.
+#   - NOT_APPLICABLE  -> unchanged silent no-op (not configured / disabled).
+#
 # SAFE SHAPE: retrofitted onto the shared premium-hook harness
-# (_lib-premium-hook.sh, me2resh/apexyard#890) — behaviour-preserving, not a
-# redesign. This hook NEVER blocks session start and ALWAYS exits 0. Every
-# failure mode is a silent no-op:
+# (_lib-premium-hook.sh, me2resh/apexyard#890) — behaviour-preserving except
+# for the #929 reachability branch above. This hook NEVER blocks session
+# start and ALWAYS exits 0. Every failure mode is a silent no-op or an
+# advisory line, never a block:
 #   - apexyard-search not on PATH  (MCP not installed)         -> no-op
 #   - index already fresh          (--if-stale short-circuits) -> no-op
 #   - APEXYARD_OPS_ROOT / _PORTFOLIO_ROOT unset                -> no-op (CLI errs, swallowed)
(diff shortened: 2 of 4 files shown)


atlas-apex (commented):
## Security Review: PR #933

**Commit**: `66c7036883ca6d0aa5ca486d428ee9c7f482cfd7`

Trust-chain review (#777) — the diff touches two SessionStart boot-path hooks (`_lib-premium-hook.sh`, `reindex-on-session-start.sh`). Availability (the probe must never wedge session start) and command construction (no injectable path into `bash -c`) are the load-bearing properties. I read both files at HEAD, ran the suites, and adversarially fuzzed the timeout knob under real bash word-splitting semantics.

### Summary
A new `premium_hook_probe()` runs a bounded, side-effect-free `apexyard-search doctor` liveness check at SessionStart before nudging a reindex, returning a three-way result (REACHABLE / UNREACHABLE / NOT_APPLICABLE). Its timeout-guarded child-process shape is a byte-for-byte copy of the already-audited `premium_hook_run` (#890 / AgDR-0095) — same GATE 1 (feature flag) / GATE 2 (presence) early-outs, same `timeout -k 2 <secs>` → `gtimeout` → unbounded fallback order, just a shorter default (5s) and a distinguishable return code instead of always-0.

### Checklist Results
- Secrets & Credentials:  Pass — none introduced
- Injection Prevention:   **Pass** — see below (fuzzed, no path found)
- Auth & Authorisation:   N/A
- Data Protection:        Pass — no info leak to non-premium adopters
- API Security:           N/A

### Availability — the probe cannot wedge session start (verified)
`test_probe_hanging_is_bounded_and_unreachable` and `test_hanging_probe_is_bounded_and_treated_unreachable` both pass: a `sleep 30` probe under a 1s bound is killed and treated UNREACHABLE in ~1-2s. There is **no `wait` on a background child** anywhere — the call is a synchronous `$to bash -c "$probe_cmd"`, so the "wait on a dead child" wedge class doesn't apply. `timeout -k 2` guarantees a SIGKILL backstop, so even a SIGTERM-ignoring `doctor` dies. Every failure/timeout routes to UNREACHABLE (1) → one advisory stderr line → `exit 0`. Fail-open is correct here (it gates only an advisory nudge, never a block).

### Injection — fuzzed, no path found
The `$to` expansion is deliberately unquoted (SC2086 disabled) for word-splitting. I fed hostile `PREMIUM_HOOK_PROBE_TIMEOUT_SECS` values under bash — `5; touch X`, `5 && touch X`, `` 5`touch X` ``, `5 *`, ` `, `-5`, `abc`, empty. **Every case returned UNREACHABLE (1) with no injected command executed** (no marker file created). Word-splitting an already-set variable adds argv tokens to a command named `timeout`; it does not re-parse `;`/`&&`/backticks as shell metacharacters, so a poisoned config value cannot escalate to arbitrary execution. The hook's `probe_cmd` is a **constant** string (`apexyard-search doctor >/dev/null 2>&1`) — no external input flows into `bash -c`. A malicious `apexyard-search` on PATH would run, but that's pre-existing (the reindex payload + `command -v` presence check already invoke it) and out of scope for this diff.

### No info leak to non-premium adopters (verified)
NOT_APPLICABLE (feature disabled OR `command -v apexyard-search` fails) → the `*)` branch → silent no-op, zero output. The advisory stderr line fires **only** when the binary is present-but-broken (`doctor` non-zero/hangs) — the intended audience. An adopter without the premium CLI learns nothing about premium config.

### Security Issues Found
None blocking.

- **LOW / advisory — huge timeout value defeats the bound.** A large numeric `PREMIUM_HOOK_PROBE_TIMEOUT_SECS` (e.g. `999999999`) makes `timeout -k 2 999999999` effectively unbounded, so a genuinely hanging `doctor` would run for the full hang. This is **not a regression** — `premium_hook_run` has the identical property (lines 310-315 vs 360-365) — and it's not externally exploitable: the knob is an operator-set env var, and anyone who can set it can already drop a malicious binary on PATH or run code directly. The realistic trigger is operator self-misconfiguration + a hanging CLI. Optional hardening: clamp to a sane max (e.g. `[ "$timeout_secs" -gt 0 ] 2>/dev/null && [ "$timeout_secs" -le 60 ] || timeout_secs=5`), which would also swallow the stray `timeout: invalid time interval` stderr line a malformed value currently prints at SessionStart. Non-blocking; fine to defer or fold into the shared harness separately so `run` and `probe` stay symmetric.

### Verification
- `bash -n` clean on both changed hooks
- `test_lib_premium_hook.sh` — 15/15 pass (6 new probe cases incl. hanging-bounded)
- `test_reindex_on_session_start.sh` — 9/9 pass (3 new reachability states + hanging-probe-bounded)
- Adversarial timeout-value fuzz under bash — no injection, all hostile values fail-safe to UNREACHABLE

### Verdict
**APPROVED** — behaviour-preserving retrofit onto the audited premium-hook harness. The core trust-chain property holds: under the default and every realistic config the probe is bounded, kills a hanging child, and falls through to `exit 0`; command construction has no injectable path; and the fail-open degrade is the correct posture for an advisory nudge. The single LOW note (huge-value unbounded) is inherited from the existing harness, operator-controlled, and worth optional clamping but not a merge blocker.

---
📌 Reviewed commit: `66c7036883ca6d0aa5ca486d428ee9c7f482cfd7`

atlas-apex (commented):
## Code Review: PR #933 — feat(#929): SessionStart search advisory probes MCP reachability

**Commit**: `66c7036883ca6d0aa5ca486d428ee9c7f482cfd7`

### Summary
Retrofits `reindex-on-session-start.sh` from a *presence* check (`command -v apexyard-search`) to a *reachability* probe before nudging a reindex, so a binary that's installed-but-broken no longer looks healthy. Adds `premium_hook_probe()` to `_lib-premium-hook.sh` — a bounded, three-way liveness sibling of `premium_hook_run()` (0 REACHABLE / 1 UNREACHABLE / 2 NOT_APPLICABLE) with its own shorter default timeout (`PREMIUM_HOOK_PROBE_TIMEOUT_SECS`, 5s). Diff is exactly 4 files: the two hooks + their two test files.

### Checklist Results
- Architecture & Design:      Pass — clean extension of the AgDR-0095 premium-hook harness; no new pattern
- Code Quality:                Pass
- Testing:                     Pass — 15/15 + 9/9, real timing on bounded cases
- Security:                    Pass — advisory-only SessionStart hook, no secrets/injection surface
- Performance:                 Pass — probe is fast + timeout-guarded; not-configured path is zero-latency
- PR Description & Glossary:   Pass — 4-term Glossary, narrative Summary (what + why)
- Summary Bullet Narrative:    Pass
- Technical Decisions (AgDR):  N/A — extends existing AgDR-0095 harness; no new dependency/framework/pattern
- Adopter Handbooks:           N/A — always-load handbooks checked (clean-architecture, migration-safety [blocking], commit-message-quality); none fire on a pure shell-hook/test diff
- Fallow Static Analysis:      N/A — not JS/TS

### Verification performed (cloned the PR HEAD and ran it, not just read it)

**(a) Probe is genuinely bounded — a hanging `apexyard-search doctor` cannot wedge SessionStart.** Confirmed the *mechanism*, not just the assertion: `premium_hook_probe` wraps the probe in `timeout -k 2 <secs>` (default 5s), same shape as `premium_hook_run`. The two hang tests use a real `sleep 30` capped at `PREMIUM_HOOK_PROBE_TIMEOUT_SECS=1` and completed in **elapsed=1s** — the probe is actually killed by its own timeout and reported UNREACHABLE, with the hook still exiting 0.

**(b) All three paths behave, and the hook still ALWAYS exits 0.** `case $?` branches 0→reindex / 1→one honest stderr line / *(2)→silent no-op, then an unconditional `exit 0`; the kill-switch early-out is also `exit 0`. `set -u` is on with no unbound-variable risk (all env reads use `:-`). Tests assert rc==0 in every state including configured-unreachable and hanging-probe.

**(c) `premium_hook_probe` reuses the gating from `premium_hook_run` — no divergence.** Identical GATE 1 (`premium_feature_enabled "$key" "$default"`) and GATE 2 (`( eval "$presence_cmd" ) >/dev/null 2>&1`), same GNU→gtimeout→unbounded fallback order. The only intentional differences are the three-way return (2 instead of 0 on gate-miss) and the separate `PREMIUM_HOOK_PROBE_TIMEOUT_SECS` knob.

**(d) No new failure on the not-configured silent-no-op path (the common free-adopter case).** No `apexyard-search` on PATH → presence check fails → probe returns 2 → silent no-op → exit 0, verified by `test_no_cli_is_silent_noop` (empty output, rc 0). No new stderr for adopters without premium search.

**(e) Tests meaningful + `bash -n` clean.** `bash -n` clean on all 4 files. Case 8 uses a dedicated stub whose `doctor` fails but whose `reindex` *would* exit 0 — proving the reindex is genuinely skipped, not coincidentally also failing. Good rigor.

**Base / hygiene:** base is `dev` (correct). Glossary + narrative Summary present.

**Pre-existing red claim — independently verified, not taken at face value.** Full suite is `PASS=106 FAIL=1 TOTAL=107`; the sole failure is `test_sync_codex_adapter.sh` (missing `block-test.sh` fixture). `git diff origin/dev...HEAD` is exactly the 4 PR files — none related to that test — and `test_sync_codex_adapter.sh` exists on `dev` with the `block-test.sh` fixture absent there too. The red is pre-existing and unrelated; correctly not held against #933.

### Issues Found
None blocking.

- **nit (env caveat):** on a stock macOS with neither `timeout` nor `gtimeout` installed (no brew coreutils), the `to=""` fallback runs the probe *unbounded* — a truly hung `doctor` could stall SessionStart. This is **pre-existing harness behaviour** (`premium_hook_run` has the identical fallback, and the reindex payload was already unbounded there), so it's **no regression** — but since this PR's whole point is reachability robustness and the boundedness is the load-bearing safety property, it may be worth a follow-up to degrade to a `& + kill` guard or at least document the "no timeout binary → unbounded" edge in the hook header alongside the other failure modes.
- **nit (contract clarity):** `premium_hook_probe`'s header says `probe_cmd should be side-effect-free and fast` but, unlike the `payload_cmd` doc for `premium_hook_run`, doesn't state that the caller owns the probe's stdout/stderr redirection. The reindex caller correctly passes `apexyard-search doctor >/dev/null 2>&1`; a future caller that forgets would leak the doctor JSON into the session transcript. One sentence in the header would close that.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `66c7036883ca6d0aa5ca486d428ee9c7f482cfd7`

me2resh (approved):

Issue #951, opened 2026-07-20
[Bug] portfolio_validate() passes when projects_dir/workspace_dir resolve to a wrong-but-existing dir inside the fork


atlas-apex:
## Affected
`portfolio_validate()` in `.claude/hooks/_lib-portfolio-paths.sh` (surfaced at SessionStart by `check-portfolio-config.sh`)

## Given / When / Then
**Given** `portfolio_validate()` checks the resolved portfolio paths at session start and warns on broken config
**When** a relative-path layout mismatch (the `../` vs no-`../` choice) makes `projects_dir`/`workspace_dir` resolve to an **existing but wrong** directory *inside the ops fork* — rather than a missing one
**Then** validation reports OK, because it only tests existence/readability/parse — so the misconfiguration is silent and subsequent work writes into the wrong tree — **expected**: flag when `projects_dir`/`workspace_dir` resolve *inside* the ops-fork root while in split-portfolio (v2) mode.

## Repro
1. A split-portfolio v2 fork (`.apexyard-fork` marker present; portfolio in a sibling repo).
2. Set `portfolio.projects_dir` (or `workspace_dir`) with the wrong `../` prefix for the layout, such that it resolves to a real directory *inside* the fork.
3. Start a session — the config banner shows OK.
4. A portfolio write (clone, doc, registry edit) silently lands in the fork instead of the portfolio.

## Framework version
v5.1.0

## Severity
major (silent wrong-location writes; no error surfaced, data goes to the wrong tree)

## Notes
`portfolio_validate()` currently catches missing/unreadable registry, bad YAML, missing `projects_dir`, and uncreateable `ideas_backlog` — but not an existing-but-wrong resolution. Suggested fix: when the `.apexyard-fork` marker is present (v2 mode), additionally warn if the resolved `projects_dir`/`workspace_dir` is under the ops-fork root.

## Glossary
| Term | Definition |
|------|------------|
| Split-portfolio v2 | Layout that keeps framework files (ops fork) separate from company data (private sibling portfolio repo), marked by `.apexyard-fork`. |
| `portfolio_validate()` | SessionStart check that surfaces broken portfolio config before a portfolio-aware skill fails with a vaguer error. |

atlas-apex:
Fixed by #952 (merged). portfolio_validate's fork-containment check now gates on the resolved location (flagging only a non-default in-fork path) rather than the .apexyard-fork marker — single-fork setups, which also carry that marker, are no longer false-flagged. Ships next release.

Issue #945, opened 2026-07-18
[Bug] Split-portfolio v2: per-project ticket markers silently ignored — portfolio_workspace_dir returns a non-canonical path (/../) that never matches the canonical FILE_PATH in the gate glob check


rafik-wahid-cubeish:
## Affected
`.claude/hooks/require-active-ticket.sh` and `.claude/hooks/require-migration-ticket.sh` (per-project marker resolution); root cause in `_portfolio_resolve` (`.claude/hooks/_lib-portfolio-paths.sh`).

## Given / When / Then
**Given** a split-portfolio v2 setup with the documented relative config `portfolio.workspace_dir: "../<portfolio-repo>/workspace"`
**When** you `/start-ticket` a managed project and then edit a file under that project's workspace clone
**Then** the gate ignores the per-project marker and falls through to the single shared `current-ticket` fallback (or blocks if no fallback exists) — expected: it should honor the Tier-0/Tier-1 per-project marker.

## Root cause
`_portfolio_resolve` builds the workspace dir as `OPS_ROOT` + the relative config value via raw string concatenation with **no canonicalization**, yielding a path containing `/../` (e.g. `<fork>/../<portfolio>/workspace`). `require-active-ticket.sh` then does a raw glob prefix match:

```sh
case "$FILE_PATH" in
  "$WORKSPACE_DIR"/*) PROJECT="${...}";;
esac
```

The harness supplies `FILE_PATH` in canonical form (no `/../`), so the prefix never matches -> `PROJECT` stays empty -> per-project markers never engage.

## Impact
- Affects every split-portfolio v2 adopter using the **documented** relative `workspace_dir`.
- Silent asymmetry: `/start-ticket` writes the per-project marker correctly (it keys on the registry `repo`, not the path), but the gate can't find it.
- Breaks per-project ticket isolation and makes two parallel sessions on different projects collide on the shared `current-ticket`.

## Repro (minimal)
1. Split-portfolio v2, `portfolio.workspace_dir: "../<portfolio>/workspace"`.
2. `/start-ticket <n>` for a registered managed project.
3. Edit a source file under `<portfolio>/workspace/<project>/...`.
4. Observe the gate report only the `current-ticket` fallback path, not the per-project marker.

[...]

Issue #373, opened 2026-05-21
bug: split-portfolio v2 — workspace_dir not defaulting to sibling repo; skills fall back to ops-fork workspace/


me2resh:
## Summary

Two related gaps in split-portfolio v2 behaviour discovered during normal adopter usage.

---

## Bug 1 — `workspace_dir` missing from config does not warn; silently falls back to ops-fork `workspace/`

**Steps to reproduce**

1. Set up split-portfolio v2 with `portfolio.registry`, `portfolio.projects_dir`, and `portfolio.ideas_backlog` all pointing at the sibling private repo.
2. Omit `portfolio.workspace_dir` from `.claude/project-config.json` (e.g. migrated from v1 where the key didn't exist).
3. Run any skill that reads or writes to `workspace/` (e.g. `/handover` accepting the clone offer, or `/status` resolving workspace paths).

**Expected**

Either:
- The `SessionStart` portfolio-validate banner warns that `workspace_dir` is unset while other v2 keys are present, OR
- The resolver defaults `workspace_dir` to `../\<fork\>-portfolio/workspace` when the other sibling-repo keys are detected.

**Actual**

Skills silently resolve `workspace/` against the ops-fork root. No banner, no warning. Clones and workspace state accumulate in the public fork's `workspace/` instead of the private sibling repo.

---

## Bug 2 — Skills writing new project docs fall back to ops-fork `projects/` when a stale local dir exists

**Steps to reproduce**

1. Have a split-portfolio v2 setup with `portfolio.projects_dir` pointing at the sibling private repo.
2. Have a stale (untracked or previously committed) `projects/<name>/` directory still present in the ops-fork root — left over from before the migration.
3. Run a skill that creates a new project entry (e.g. `/handover` for a new project).

**Expected**

The skill always writes to the resolved `portfolio.projects_dir` path (the sibling repo). The presence of a local `projects/` directory in the ops fork should have no effect.

**Actual**

The skill writes new project docs to the ops-fork's local `projects/` directory instead of the configured sibling path. The sibling repo stays out of date. Two divergent copies of project docs now exist.

---

## Additional context

- Both bugs are silent — no error, no warning, no banner. The adopter only discovers the problem by noticing that workspace clones or project docs landed in the wrong repo.
- The `SessionStart` portfolio-validate path already banners on missing/broken paths; extending it to warn on a partially-configured v2 block (other keys present but `workspace_dir` absent) would catch Bug 1 at session start.
[...]

PR #952, opened 2026-07-20
fix(#951): portfolio_validate flags projects_dir/workspace_dir resolving inside the fork

atlas-apex:
## Summary

- **Closes the silent-misconfig gap `portfolio_validate()` had for split-portfolio v2 forks** — it only checked that `projects_dir`/`workspace_dir` existed, were readable, and parsed as YAML. A `../` depth mismatch in a v2 override (e.g. one `../` too few) can resolve to a directory that genuinely exists but sits **inside the ops fork** instead of the intended sibling private-portfolio repo — the existence check passes, so nothing ever surfaces the mistake, and subsequent portfolio writes (clones, docs, registry edits) silently land in the wrong tree.
- **Adds a fork-containment check**, gated conservatively so it can't false-positive on plain v1 single-fork setups: it fires only when the `.apexyard-fork` marker is present, or when `registry`/`projects_dir` already resolve outside the fork (the same v2-oriented signal the existing `#373` asymmetry check uses). When it fires, it canonicalizes `projects_dir` via `cd && pwd -P` (the same technique already used for `workspace_dir`'s `$wd_real`) before checking containment — closing the specific gap where the existing raw string-prefix check can miss a `../`-mangled path that still happens to land inside the fork.
- **Stays inside `portfolio_validate()` only** — a sibling PR (#945) is separately fixing `_portfolio_resolve`/`portfolio_workspace_dir` canonicalization in the same file; this PR doesn't touch either function to minimize overlap.

## Testing

Added three new cases to `.claude/hooks/tests/test_portfolio_paths.sh` (mirrors the existing `#373` partial-v2 test style):

1. **v2 marker present + `projects_dir` resolves inside the fork (the `../` mismatch bug)** — a decoy directory is created *inside* the fork sharing the name the sibling repo should have had, with a `projects_dir` override missing its leading `../`. `portfolio_validate()` now reports `broken: ... projects_dir ... INSIDE the ops fork ...` and returns non-zero.
2. **v2 marker present + `projects_dir`/`workspace_dir` correctly resolve to the sibling repo** — validates OK, proving no false positive on a correctly configured v2 fork.
3. **v1 single-fork, no marker, untouched in-fork defaults** — validates OK, proving the check is conservative and doesn't fire on the common case.

Local run:

```
$ bash .claude/hooks/tests/test_portfolio_paths.sh
...
PASS: v2 (#951): projects_dir resolves inside the fork → broken
PASS: v2 (#951): correctly configured sibling projects_dir/workspace_dir → OK
PASS: v1 (#951): untouched in-fork defaults, no marker → OK (no false positive)
...
===== test_portfolio_paths.sh =====
Passed: 42
Failed: 0
```

Also ran the full `test_split_portfolio_v2_migration.sh` suite (23/25 pass) to check for regressions — the 2 failures (`validate post-state`, `post-v2 ops_root`) are **pre-existing on `upstream/dev`**, reproduced identically with `git stash` before applying this change, and are environment artifacts (macOS `/tmp` vs `/private/var` path resolution in the sandbox), not caused by this fix.

`shellcheck -x .claude/hooks/_lib-portfolio-paths.sh` reports no issues.

`bash .claude/hooks/run-hook-tests.sh` does not exist in this checkout, so it was not run.

Refs #951

## Glossary

| Term | Definition |
|------|------------|
| Split-portfolio v2 | Layout where framework files stay in the public ops fork and company-private data (registry, projects, onboarding, workspace) lives in a sibling private repo, marked by an `.apexyard-fork` file at the fork root. |
| `portfolio_validate()` | A `SessionStart`-invoked sanity check (via `check-portfolio-config.sh`) that verifies the portfolio's resolved paths (registry, `projects_dir`, `ideas_backlog`, `onboarding`, `workspace_dir`) are usable, surfacing misconfiguration before a portfolio-aware skill fails with a vaguer error later. |
| Fork-containment | Whether a resolved path lives inside the ops-fork's own directory tree vs. outside it (e.g. in a sibling repo). A v2 fork's `projects_dir`/`workspace_dir` are expected to be OUTSIDE the fork; this PR adds the check that flags when they aren't. |
| Canonicalization (`cd && pwd -P`) | Resolving a path (including any `../` segments) to its real, absolute, symlink-free form, so string-prefix comparisons against the fork root are accurate. |

**Trust-chain note**: this PR touches `.claude/hooks/**`, so it's expected to trigger the Security Auditor (Hakim) review per the trust-chain trigger in `.claude/rules/role-triggers.md` (#777).

### DIFF
--- a/.claude/hooks/_lib-portfolio-paths.sh
+++ b/.claude/hooks/_lib-portfolio-paths.sh
@@ -403,41 +403,41 @@ portfolio_validate() {
     fi
   fi
 
-  # Fork-containment check — see me2resh/apexyard#951.
+  # Fork-containment check — see me2resh/apexyard#951 (and #952 for the
+  # false-positive fix below).
   #
-  # The checks above compare RAW resolved paths (simple string
-  # concatenation — see _portfolio_resolve) against root_real. A
-  # `../` depth mismatch in a projects_dir/workspace_dir override can
-  # still exist on disk (a real, but WRONG, directory) while landing
-  # right back INSIDE the ops fork rather than the intended sibling
-  # portfolio repo — the plain existence check earlier in this
-  # function passes, so the misconfig stays silent and subsequent
-  # writes (clones, docs, registry edits) land in the wrong tree.
+  # A projects_dir/workspace_dir override that's meant to point at the
+  # sibling portfolio repo but has the wrong `../` depth (or a missing
+  # leading `../`) can resolve to a real, but WRONG, directory INSIDE the
+  # ops fork. The plain existence check earlier in this function passes —
+  # the decoy dir genuinely exists — so the misconfig stays silent and
+  # subsequent writes (clones, docs, registry edits) land in the wrong
+  # tree. Close that gap by resolving each dir to its canonical
+  # (`cd && pwd -P`) form and flagging if it lands inside.
   #
-  # Close that gap by resolving projects_dir to its canonical
-  # (`cd && pwd -P`) form and checking fork-containment on THAT, in
-  # addition to the raw check above. workspace_dir already gets this
-  # canonical treatment via $wd_real earlier in this function.
+  # GATE ON THE RESOLVED LOCATION, NOT on portfolio_is_v2 (#952). Flag only
+  # when a dir resolves INSIDE the ops fork at a NON-DEFAULT location. The
+  # in-fork DEFAULTS ($root/projects, $root/workspace) are correct
+  # single-fork behaviour and must never be flagged; a sibling-intended
+  # override with the wrong `../` depth (or a missing leading `../`) lands
+  # inside the fork at some OTHER path, which is the misconfig #951 catches.
   #
-  # Conservative by design: only fires in a confirmed v2-oriented
-  # setup — the `.apexyard-fork` marker is present, OR registry /
-  # projects_dir already resolve outside the fork (the same
-  # v2-oriented signal #373 uses above). A plain single-fork (v1)
-  # setup with untouched in-fork defaults never reaches this branch,
-  # so it can't false-positive there.
+  # The old `portfolio_is_v2` gate false-positived because single-fork forks
+  # ALSO carry the `.apexyard-fork` marker (it's written on every /setup,
+  # single-fork or split), so the marker can't tell a legit in-fork default
+  # from a sibling-intended override — see #952. Comparing the resolved path
+  # against the canonical default location is immune to that ambiguity and
+  # needs no (root-resolution-sensitive) config-key read.
   local pd_real
   pd_real=$(cd "$projects_dir" 2>/dev/null && pwd -P) || pd_real="$projects_dir"
+  if ! _outside_fork "$pd_real" && [ "$pd_real" != "$root_real/projects" ]; then
+    echo "broken: split-portfolio v2 misconfig — portfolio.projects_dir resolved to $projects_dir, which is INSIDE the ops fork ($root_real) rather than the sibling private-portfolio repo. Check for a '../' depth mismatch (or a missing leading '../') in .claude/project-config.json's portfolio.projects_dir override. See me2resh/apexyard#951."
+    return 1
+  fi
 
-  if portfolio_is_v2 || _outside_fork "$registry" || _outside_fork "$projects_dir"; then
-    if ! _outside_fork "$pd_real"; then
-      echo "broken: split-portfolio v2 misconfig — portfolio.projects_dir resolved to $projects_dir, which is INSIDE the ops fork ($root_real) rather than the sibling private-portfolio repo. Check for a '../' depth mismatch in .claude/project-config.json's portfolio.projects_dir override. See me2resh/apexyard#951."
-      return 1
-    fi
-
-    if ! _outside_fork "$wd_real"; then
-      echo "broken: split-portfolio v2 misconfig — portfolio.workspace_dir resolved to $workspace_dir, which is INSIDE the ops fork ($root_real) rather than the sibling private-portfolio repo. Check for a '../' depth mismatch in .claude/project-config.json's portfolio.workspace_dir override. See me2resh/apexyard#951."
-      return 1
-    fi
+  if ! _outside_fork "$wd_real" && [ "$wd_real" != "$root_real/workspace" ]; then
+    echo "broken: split-portfolio v2 misconfig — portfolio.workspace_dir resolved to $workspace_dir, which is INSIDE the ops fork ($root_real) rather than the sibling private-portfolio repo. Check for a '../' depth mismatch (or a missing leading '../') in .claude/project-config.json's portfolio.workspace_dir override. See me2resh/apexyard#951."
+    return 1
   fi
 
   case "$onboarding" in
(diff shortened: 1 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Security Review: PR #952 — PASS

▸ Hakim (Security Auditor) · trust-chain change to `.claude/hooks/_lib-portfolio-paths.sh`

**Reviewed commit**: `5024f701e45a875e77075dcc7d10baabfb2036b2`

### Summary
Adds a fork-containment check to `portfolio_validate()` (the SessionStart config-health banner) that flags when `projects_dir`/`workspace_dir` resolve *inside* the ops fork at a non-default location — the silent split-portfolio-v2 `../`-depth misconfig from #951. Latest commit correctly re-gates on the resolved location (`cd && pwd -P` canonical form) rather than the `.apexyard-fork` marker, killing the false-positive where single-fork setups also carry that marker.

### Security analysis
- **Fail-open?** No. The check fires when a dir is inside-fork AND != the canonical default (`$root_real/projects` / `$root_real/workspace`). A genuinely-misconfigured sibling-intended path that lands inside the fork resolves to a non-default location → flagged (`return 1`). The only non-flagged "inside" cases are the exact defaults (writes land correctly) or outside-fork (correct v2). When a dir doesn't exist, the earlier existence guard (line 355) already returns for `projects_dir`; the `wd_real` raw-string fallback for a not-yet-created default resolves to the default and is not flagged. **Fails closed/noisy-but-safe**, never silently passing a wrong-tree path.
- **Symlink canonicalization** (`cd && pwd -P`): resolving symlinks is the *correct* behavior here — the check judges where writes actually land (the real target), which is exactly what you want for a wrong-tree guard. Both sides of the containment/default comparison for `pd_real` are canonical, so no asymmetric mis-judgment. Threat model is negligible regardless: `projects_dir`/`workspace_dir` come from the fork owner's own `project-config.json`; an actor who can plant a symlink there or edit that file already owns the repo.
- **Injection / eval / unquoted expansion**: none. `cd "$projects_dir"`, `_outside_fork "$pd_real"`, and the `[ "$pd_real" != "$root_real/projects" ]` tests are all quoted; the `_outside_fork` case-glob keeps its patterns quoted (only the trailing `/*` is active glob). Paths are only `echo`'d into the banner — no re-execution.
- **Ordering**: new checks run after the existence guards and before the later workspace existence/writability checks; no conflict or double-flag.

### Checklist
- Secrets/credentials: Pass · Injection: Pass · XSS: N/A · Auth/authz: N/A · Data protection: Pass · Trust-chain integrity: Pass

### Non-blocking nit
The PR **body** still describes the older marker-gated approach ("fires only when the `.apexyard-fork` marker is present"), which the latest commit superseded with location-gating. The in-diff comments document the change correctly; consider syncing the body so future readers aren't misled. Not a security issue.

### CI
All relevant checks green — hook test suite, shellcheck, Analyze, Verify Ticket ID all pass. CodeQL "skipping" is expected for a shell-only change, not a failure.

### Verdict
**APPROVED (PASS)** — no blocking or non-blocking security issues. Security marker written at HEAD.

---
📌 Reviewed commit: `5024f701e45a875e77075dcc7d10baabfb2036b2`

atlas-apex (commented):
## Code Review: PR #952

**Commit**: `5024f701e45a875e77075dcc7d10baabfb2036b2`

### Summary
Adds a fork-containment check to `portfolio_validate()` in `.claude/hooks/_lib-portfolio-paths.sh`. It closes the #951 silent-misconfig gap where a `projects_dir`/`workspace_dir` override with the wrong `../` depth resolves to a real-but-wrong directory *inside* the ops fork (the existence check passes, so the mistake never surfaces). The latest commit fixes the #952 sub-bug: the containment check now gates on the **resolved location** (flag only when a dir resolves inside the fork at a **non-default** path), not on the `.apexyard-fork` marker — which single-fork forks also carry, causing the earlier false-positive that failed `test_setup_single_fork.sh`.

### Checklist Results
- Security:                  Pass (read-only path checks; no injection surface)
- Performance:               Pass (two extra `cd && pwd -P` + string compares — negligible at SessionStart)
- Technical Decisions (AgDR): N/A (bug fix extending an existing validation pattern; no new dependency/pattern/library)
- Adopter Handbooks:          N/A (no shell-language handbook; always-load architecture/general handbooks don't apply to a validation-hook diff)

### Issues Found
None blocking.

Correctness verified against the three scenarios you flagged:

1. **#951 misconfig** — a relative override like `sibling-portfolio/projects` (missing leading `../`) is resolved by `_portfolio_resolve` to the absolute `$root/sibling-portfolio/projects`, canonicalized to `pd_real=$root_real/sibling-portfolio/projects`. That is `! _outside_fork` (inside) AND `!= $root_real/projects` (non-default) → correctly flagged, returns 1.
2. **Correctly configured v2** — `projects_dir`/`workspace_dir` pointing at a sibling repo outside the fork are `_outside_fork` → the AND short-circuits false → correctly passes.
3. **v1 single-fork default** — `$root/projects` canonicalizes to `$root_real/projects` (both sides use `pwd -P`), so the equality guard `[ "$pd_real" != "$root_real/projects" ]` is false → correctly passes.

**Canonicalization consistency (your macOS `/var`→`/private/var` concern):** clean. `root_real`, `pd_real`, and `wd_real` all go through `cd && pwd -P`, so any symlink component (incl. `/var`→`/private/var`) resolves identically on both sides of the `$pd_real != $root_real/projects` comparison. A legit default can't be flagged by a symlinked ancestor. The one residual edge — `projects`/`workspace` itself being a symlink out of the fork — resolves to "outside" and is correctly not flagged; a symlink to another in-fork path would flag, which is a defensible (and extremely unusual) case.

**`wd_real` fallback for a non-existent default workspace:** when the in-fork default `workspace_dir` doesn't exist yet, `wd_real` falls back to the raw (un-canonicalized) `$workspace_dir` (from line 390). If `$root != $root_real` (symlinked root), that raw path fails the `_outside_fork` prefix match and is treated as "outside" → not flagged. Benign — no false positive — and the later existence check is intentionally lenient for the not-yet-created default. Noted, not a defect.

**Scope:** the new block sits strictly between the #373 partial-v2 check and the `onboarding` case, all inside `portfolio_validate()`. The #373 check is untouched and still evaluates first, so a #951 misconfig (projects_dir inside-fork, non-default) short-circuits #373's `_outside_fork` (false) and is caught by the new block — no interaction. `_outside_fork`, `root_real`, `wd_real` are all in scope at the insertion point.

### Suggestions
- nit (non-blocking, out of scope): the #373 check above uses the **raw** `$projects_dir`/`$registry` for `_outside_fork`, whereas this new check canonicalizes via `pd_real`. On a symlinked fork root the raw-path form is the less-correct one; the new PR's canonicalized approach is the better pattern. Worth aligning #373 to `pd_real` in a future pass — but this PR correctly stays in scope and leaves #373 alone as stated.

### CI
Green at HEAD `5024f70`: hook test suite pass (2m19s), shellcheck pass, Analyze pass, Verify Ticket ID pass. CodeQL shows `skipping` (skip status, not a failure).

### Verdict
**APPROVED**

---
📌 Reviewed commit: `5024f701e45a875e77075dcc7d10baabfb2036b2`

PR #953, opened 2026-07-20
fix(#945): canonicalize resolved portfolio paths so ticket-gate prefix match works

atlas-apex:
## Summary

- **Fixed the root cause of #945** — `_portfolio_resolve()` in `_lib-portfolio-paths.sh` (the function behind `portfolio_workspace_dir`, `portfolio_projects_dir`, and every other `portfolio.*` config resolver) built its output by string-concatenating a relative override onto the ops-fork root, with no normalisation step. A split-portfolio v2 adopter whose `portfolio.workspace_dir` is a relative sibling path — the documented shape, e.g. `"../<fork>-portfolio/workspace"` — got back a resolved value that still contained a literal `/../` segment (e.g. `/Users/x/apexyard/../apexyard-portfolio/workspace`).
- **Why that broke the ticket gate** — `require-active-ticket.sh` and `require-migration-ticket.sh` compare a tool-supplied `FILE_PATH` (always arrives pre-normalised by the harness — no `..` component) against the resolved workspace dir via a shell `case "$FILE_PATH" in "$WORKSPACE_DIR"/*)` prefix match. A literal `/../` never shows up in a real `FILE_PATH`, so the match silently failed. Concretely, the per-project ticket marker (Tier 0/1 in the resolution order) was ignored, and legitimate edits either got blocked outright or fell through to the wrong tier — the exact symptom reported in #945.
- **The fix** — added `_portfolio_canonicalize()`, an existence-tolerant canonicalizer: it walks up to the nearest existing ancestor, physically resolves it with `cd ... && pwd -P` (which both collapses `..` and follows symlinks), and re-appends whatever tail doesn't exist yet literally. This mirrors the trick `require-active-ticket.sh`'s own `_resolve_real_path()` already uses, and deliberately avoids depending on GNU `realpath -m`/`-e` (not guaranteed present on macOS/BSD). `_portfolio_resolve()` now runs every computed path through it — both the relative-join branch and the absolute-passthrough branch — so the fix applies once, at the resolution boundary, and every caller (`portfolio_workspace_dir`, `portfolio_projects_dir`, `portfolio_registry`, `portfolio_onboarding_path`, `portfolio_custom_skills_dir`, `portfolio_custom_handbooks_dir`, `portfolio_agent_routing`) benefits without needing its own fix.
- **Why this seam, not the gate** — canonicalizing at `_portfolio_resolve()` fixes every caller uniformly instead of patching `require-active-ticket.sh`'s prefix check alone (which would leave `require-migration-ticket.sh` and any future caller with the same bug). I checked every existing caller (`require-active-ticket.sh`, `require-migration-ticket.sh`, `print-portfolio-primer.sh`, `_lib-multi-repo-trace.sh`, `maintain-docs-index.sh`, `_lib-audit-history.sh`) — none of them depend on the exact non-canonical string; they either do prefix comparisons (which this fix repairs) or pass the value to `cd`/`ls`/`find` (which already tolerate `/../` transparently). `portfolio_validate()` in the same file is untouched — that function is owned by a sibling PR (#951).

## Testing

Added three new cases to `.claude/hooks/tests/test_portfolio_paths.sh` (existing suite; I did not create a new file):

1. **`#945: relative override with a literal '..' resolves /../-free`** — sets `portfolio.workspace_dir` to the raw string `"../apexyard-portfolio/workspace"` against a fixture where the sibling directory genuinely exists next to the fork root (every *other* override test in this file pre-resolves the sibling path via `cd && pwd` before writing it into config, so none of them actually exercised the buggy relative-with-`..` concatenation branch — this one deliberately does).
2. **`#945: end-to-end — FILE_PATH under the resolved workspace now matches the gate's prefix check`** — builds a `FILE_PATH` independently from the known-canonical sibling root (not derived from `$WORKSPACE_DIR` itself, which would trivially "pass" even on the buggy code since both sides would carry the same literal `/../`), then asserts the `case "$FILE_PATH" in "$WORKSPACE_DIR"/*)` prefix match — the actual gate logic — now succeeds.
3. **`#945: relative override with '..' resolves /../-free even when the leaf dir does not exist yet`** — same shape, but the `workspace/` leaf hasn't been created yet (the common split-portfolio v2 state before the first managed-project clone), proving the canonicalizer's existence-tolerant walk works without requiring the full path to exist.

I verified all three tests actually catch the regression (not vacuously true) by temporarily reverting only `_lib-portfolio-paths.sh` and re-running:

```
$ git stash push -- .claude/hooks/_lib-portfolio-paths.sh
$ bash .claude/hooks/tests/test_portfolio_paths.sh 2>&1 | tail -10
FAIL: #945: relative override with a literal '..' resolves /../-free
  output: still contains a literal /../ segment: /private/var/.../apexyard/../apexyard-portfolio/workspace
FAIL: #945: end-to-end — FILE_PATH under the resolved workspace now matches the gate's prefix check
FAIL: #945: relative override with '..' resolves /../-free even when the leaf dir does not exist yet
  output: still contains a literal /../ segment: /private/var/.../apexyard/../apexyard-portfolio/workspace

===== test_portfolio_paths.sh =====
Passed: 39
Failed: 3
$ git stash pop
```

With the fix restored, the full suite is green:

```
$ bash .claude/hooks/tests/test_portfolio_paths.sh 2>&1 | tail -6
PASS: #945: relative override with a literal '..' resolves /../-free
PASS: #945: end-to-end — FILE_PATH under the resolved workspace now matches the gate's prefix check
PASS: #945: relative override with '..' resolves /../-free even when the leaf dir does not exist yet

===== test_portfolio_paths.sh =====
Passed: 42
Failed: 0
```

I also ran the broader hook-test suite for anything touching `_lib-portfolio-paths.sh` or the two ticket gates it feeds, to check for regressions: `test_require_active_ticket_bash.sh` (78/78), `test_require_migration_ticket.sh` (26/26), `test_portfolio_agent_routing.sh` (6/6), `test_print_portfolio_primer.sh` (3/3), `test_audit_history.sh` (11/11), `test_maintain_docs_index.sh` (33/33), `test_handover_clone_prompt.sh` (18/18) — all pass.

Three unrelated test files (`test_ops_root.sh`, `test_split_portfolio_v2_migration.sh`, `test_workspace_tracker_resolution.sh`) show pre-existing failures in this sandbox — I confirmed by stashing my change and re-running them, and they fail identically with or without this diff, so they're not a regression introduced here (looks like an environment/CWD quirk unrelated to this fix).

`shellcheck -x .claude/hooks/_lib-portfolio-paths.sh` is clean.

Refs #945

---

## Glossary

| Term | Definition |
|------|------------|
| Canonicalize | Resolve a path to one unambiguous absolute form — no `..`/`.` segments, no symlinks left unresolved. Two different-looking path strings can name the same file; canonicalizing makes them compare equal as strings too. |
| Existence-tolerant | A canonicalization approach that doesn't require the full path to exist on disk — it resolves as much of the path as physically exists (walking up to the nearest real ancestor) and reattaches the rest literally, since a not-yet-created tail can't itself be a symlink. |
| Prefix match | A shell `case "$x" in "$y"/*)` check used here to test "is `$x` a path underneath directory `$y`?" — purely a string comparison, so both sides must already be in the same normalised form or the match silently fails. |
| Split-portfolio v2 | An apexyard deployment mode where company config (`onboarding.yaml`, `apexyard.projects.yaml`, `workspace/`) lives in a private sibling repo next to the public framework fork, instead of inside the fork itself. |
| Tier 0/1 ticket marker | The per-worktree / per-project active-ticket marker files `require-active-ticket.sh` checks (in order) before falling back to the ops-level marker or blocking the edit outright. |

### DIFF
--- a/.claude/hooks/_lib-portfolio-paths.sh
+++ b/.claude/hooks/_lib-portfolio-paths.sh
@@ -89,25 +89,116 @@ _portfolio_root() {
 }
 
 # ------------------------------------------------------------------------------
+# Internal: canonicalize an ABSOLUTE path to its logical, `/../`-free,
+# symlink-resolved form — WITHOUT requiring the leaf to already exist.
+#
+# Why this exists (me2resh/apexyard#945): callers of the public resolvers
+# below (require-active-ticket.sh, require-migration-ticket.sh) compare a
+# tool-supplied FILE_PATH against e.g. `portfolio_workspace_dir()` via a
+# shell `case "$FILE_PATH" in "$WORKSPACE_DIR"/*)` prefix match. FILE_PATH
+# arrives already normalised (no literal ".." component). Before this fix,
+# a split-portfolio v2 adopter whose `portfolio.workspace_dir` override was
+# a RELATIVE sibling path (e.g. "../<fork>-portfolio/workspace") got a
+# resolved value that still contained a literal "/../" segment — e.g.
+# "/Users/x/apexyard/../apexyard-portfolio/workspace" — because the old
+# `_portfolio_resolve` just string-concatenated `$root/$p` with no
+# normalisation. That literal "/../" never appears in a real FILE_PATH, so
+# the prefix match silently failed and the per-project ticket marker
+# (Tier 0/1) was ignored — legitimate edits got blocked (or fell through
+# to the wrong tier).
+#
+# Algorithm mirrors `_resolve_real_path()` in require-active-ticket.sh:
+# walk up to the nearest EXISTING ancestor, physically resolve it via
+# `cd ... && pwd -P` (which both collapses ".." and follows symlinks),
+# then re-append whatever tail doesn't exist yet literally — a tail that
+# doesn't exist yet cannot itself be a symlink or contain an unresolved
+# "..", since dirname/basename already stripped those out of the walked
+# portion. This is the same "realpath -m without depending on GNU
+# coreutils" trick, needed because macOS/BSD doesn't guarantee GNU
+# realpath's -m/-e flags are available.
+#
+# Fails soft: if $p can't be resolved for any reason (cd denied, etc.),
+# echoes the original input unchanged rather than an empty/broken path —
+# unlike require-active-ticket.sh's fail-CLOSED security check, a
+# portfolio path resolver has no "block on doubt" contract to uphold.
+# ------------------------------------------------------------------------------
+_portfolio_canonicalize() {
+  local p="$1" dir tail=""
+  case "$p" in
+    /*) : ;;
+    *) echo "$p"; return 0 ;;  # not absolute — nothing to canonicalize
+  esac
+
+  dir="$p"
+  while [ -n "$dir" ] && [ "$dir" != "/" ] && [ ! -e "$dir" ]; do
+    if [ -z "$tail" ]; then
+      tail="$(basename "$dir")"
+    else
+      tail="$(basename "$dir")/$tail"
+    fi
+    dir="$(dirname "$dir")"
+  done
+
+  if [ ! -e "$dir" ]; then
+    echo "$p"
+    return 0
+  fi
+
+  if [ -d "$dir" ]; then
+    dir="$(cd "$dir" 2>/dev/null && pwd -P)"
+  else
+    # $dir resolved to an existing FILE (not a directory) partway through
+    # the walk — canonicalize its parent and re-append its own basename.
+    local parent
+    parent="$(cd "$(dirname "$dir")" 2>/dev/null && pwd -P)"
+    if [ -n "$parent" ]; then
+      dir="$parent/$(basename "$dir")"
+    else
+      dir=""
+    fi
+  fi
+
+  if [ -z "$dir" ]; then
+    echo "$p"
+    return 0
+  fi
+
+  if [ -n "$tail" ]; then
+    if [ "$dir" = "/" ]; then
+      printf '%s%s' "$dir" "$tail"
+    else
+      printf '%s/%s' "$dir" "$tail"
+    fi
+  else
+    printf '%s' "$dir"
+  fi
+}
+
+# ------------------------------------------------------------------------------
 # Internal: resolve a possibly-relative path against the ops-fork root.
-# Outputs an absolute path. If the input is already absolute, echo as-is.
+# Outputs an absolute, canonical (symlink-free, `/../`-free) path. If the
+# input is already absolute, it is still canonicalized — an absolute
+# override in project-config.json could itself carry a relative segment
+# (e.g. "$SIBLING/../other").
 # ------------------------------------------------------------------------------
 _portfolio_resolve() {
-  local p="$1"
+  local p="$1" abs
   case "$p" in
-    /*) echo "$p" ;;
+    /*) abs="$p" ;;
     *)
       local root
       root=$(_portfolio_root)
       if [ -n "$root" ]; then
         # Strip leading ./ for tidier output.
-        echo "$root/${p#./}"
+        abs="$root/${p#./}"
       else
         # No root — return the literal so caller can detect.
         echo "$p"
+        return 0
       fi
       ;;
   esac
+  _portfolio_canonicalize "$abs"
 }
 
 # ------------------------------------------------------------------------------
--- a/.claude/hooks/tests/test_portfolio_paths.sh
+++ b/.claude/hooks/tests/test_portfolio_paths.sh
@@ -688,6 +688,134 @@ if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; e
 '
 rm -rf "$SB"
 
+# ---------------------------------------------------------------------------
+# Case 25 (#945): a RELATIVE split-portfolio v2 override that itself carries
+# a literal ".." segment (e.g. "../<fork>-portfolio/workspace" — the exact
+# shape a split-portfolio v2 adopter configures when the private sibling
+# repo lives next to the ops fork) must resolve to a canonical, `/../`-free
+# absolute path. Every other override test in this file pre-resolves its
+# sibling path via `cd ... && pwd` before writing it into project-config.json
+# (so it never exercises the buggy concatenation branch); this case
+# deliberately writes the RAW relative-with-".." string, which is what
+# actually reaches `_portfolio_resolve` in the field.
+# ---------------------------------------------------------------------------
+SB=$(make_fork)
+SIBROOT=$(mktemp -d)
+SIBROOT=$(cd "$SIBROOT" && pwd -P)
+mkdir -p "$SIBROOT/apexyard-portfolio/workspace"
+# Re-root the sandbox so "$SB/.." really is $SIBROOT (the literal ".."
+# override below must land on a real, existing sibling directory).
+rm -rf "$SB"
+SB="$SIBROOT/apexyard"
+mkdir -p "$SB"
+(
+  cd "$SB" || exit 1
+  git init -q
+  git config user.email "test@example.com"
+  git config user.name "test"
+  touch onboarding.yaml
+  cat > apexyard.projects.yaml <<'YAML'
+version: 1
+projects:
+  - name: example
+    repo: example/example
+YAML
+  mkdir -p projects
+  cat > projects/ideas-backlog.md <<'MD'
+# Ideas Backlog
+MD
+  mkdir -p .claude/hooks
+  cp "$LIB_SRC" .claude/hooks/_lib-portfolio-paths.sh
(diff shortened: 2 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Security Review: PR #953

**Commit**: `958e55ac31d7766b3a1e8ecac5b269b130a9911c`

🛡️ Trust-chain review — this PR changes `_lib-portfolio-paths.sh`, whose output (`portfolio_workspace_dir`) feeds the security-relevant ticket gates (`require-active-ticket.sh`, `require-migration-ticket.sh`).

### Summary
Adds `_portfolio_canonicalize()` and routes every `_portfolio_resolve()` output through it, so a relative split-portfolio v2 override like `../<fork>-portfolio/workspace` resolves to a `/../`-free, symlink-physical absolute path instead of a literal `/Users/x/apexyard/../apexyard-portfolio/workspace`. That literal `/../` never appears in a harness-supplied `FILE_PATH`, so the gate's `case "$FILE_PATH" in "$WORKSPACE_DIR"/*)` prefix match silently failed pre-fix.

### Checklist Results
- Secrets & Credentials:  Pass (no secrets, no new env handling)
- Injection Prevention:   Pass — no `eval`, no unquoted expansion into a command; `cd "$dir"`, `basename`/`dirname`, `printf '%s'` all quoted; the downstream `case "$WORKSPACE_DIR"/*` keeps `$WORKSPACE_DIR` quoted so glob metachars are literal
- XSS Prevention:         N/A (shell)
- Auth & Authorisation:   Pass — the gate's authorization semantics are preserved/strengthened
- Data Protection:        N/A
- Gate integrity:         Pass — see analysis below

### Gate-bypass analysis (the load-bearing concern)
- **No fail-open path.** For an edit to be *allowed* via the per-project tier, `FILE_PATH` must match the `WORKSPACE_DIR` prefix AND a valid per-project ticket marker must already exist (user ran `/start-ticket`). Canonicalization only normalizes `WORKSPACE_DIR` to its *true physical* location — a `FILE_PATH` matches only if it genuinely lives under the real workspace, which is exactly what governance intends. Nothing here converts a should-be-BLOCKED path into an ALLOWED one.
- **Every failure mode is fail-closed.** `_portfolio_canonicalize` fails soft to the *original* (possibly `/../`-bearing) string on `cd` denial, which reproduces the pre-fix non-matching behavior → falls through to the ops-tier marker or blocks. It never echoes empty for non-empty input (verified all branches), so `WORKSPACE_DIR` can't collapse to empty and turn `"$WORKSPACE_DIR"/*` into a match-everything `/*`.
- **Symlink / TOCTOU:** `cd && pwd -P` resolves symlinks at resolution time, but the gate decision is a pure string prefix comparison — there's no re-stat of the file between check and use, so no new TOCTOU window that grants a bypass. An attacker who could plant a symlink in the workspace path or edit `project-config.json` already controls the trust chain.
- **Consistency:** the algorithm is a faithful mirror of `_resolve_real_path()` in `require-active-ticket.sh` (same walk-to-existing-ancestor + `pwd -P` + literal-tail trick; `_portfolio_canonicalize` additionally guards the `dir="/"` root case, avoiding a `//x` double-slash — a strict improvement). It also aligns the per-project marker resolution with the *already-canonicalized* boundary anchors used in the security-critical out-of-governance exemption (lines 377–382), improving internal consistency rather than diverging.

### Security Issues Found
None (no Critical / High / Medium / Low security findings).

### Recommendations (non-blocking)
- **Asymmetry note (usability, fail-closed):** the per-project prefix match at `require-active-ticket.sh:483` compares the **raw** `FILE_PATH` (not symlink-resolved) against the now symlink-**physical** `WORKSPACE_DIR`. On a host where the real workspace path traverses a symlink and the harness supplies the non-physical form (classic macOS `/var` vs `/private/var`), the match could miss and fall back to the ops-tier marker or block a legitimate edit. This is the **safe** direction (never a bypass), but if the paper-cut shows up in the field, resolving `FILE_PATH` through `_resolve_real_path` before that match would restore symmetry. Out of scope for this PR.

### CI
All checks green — hook test suite, ShellCheck, CodeQL, PR Title Check.

### Verdict
**PASS — APPROVED** (no blocking security issue). Security-approval marker written at HEAD `958e55ac`.

---
📌 Reviewed commit: `958e55ac31d7766b3a1e8ecac5b269b130a9911c`

atlas-apex (commented):
## Code Review: PR #953

**Commit**: `958e55ac31d7766b3a1e8ecac5b269b130a9911c`

### Summary
Fixes #945 by adding `_portfolio_canonicalize()` at the `_portfolio_resolve()` seam in `_lib-portfolio-paths.sh`. A split-portfolio v2 adopter whose `portfolio.workspace_dir` was a relative sibling override (`../<fork>-portfolio/workspace`) previously got back a value carrying a literal `/../` segment, so the ticket gate's `case "$FILE_PATH" in "$WORKSPACE_DIR"/*)` prefix match silently failed and per-project markers were ignored. Canonicalizing once at the resolution boundary fixes every `portfolio_*` resolver uniformly.

### Checklist Results
- Architecture & Design:    Pass — fix lives at the correct single seam (`_portfolio_resolve`), not scattered across each gate
- Code Quality:              Pass — mirrors the established `_resolve_real_path()` pattern; all expansions quoted; shellcheck clean per CI
- Testing:                   Pass — 3 new cases prove `/../`-free output, end-to-end gate prefix match, and existence-tolerance; author verified non-vacuity by reverting the lib
- Security:                  Pass — fail-soft (echoes original on `cd` failure); no injection surface; no secrets
- Performance:               Pass — runs once per resolver, result cached per-process
- PR Description & Glossary: Pass — narrative summary bullets, thorough Testing section, 5-term Glossary
- Technical Decisions (AgDR): N/A — bug fix reusing an in-codebase pattern; no new dependency/framework/architecture. The seam-vs-gate choice is well-justified in the PR body.
- Adopter Handbooks:         N/A — shell diff; no language-handbook match; architecture/general handbooks (clean-arch, migration-safety, commit-quality) don't trigger
- Fallow Static Analysis:    N/A — not JS/TS

### Correctness verification (the review focus areas)
- **Not-yet-existing leaf dir** — handled: the `while` loop walks up to the nearest existing ancestor, resolves it with `cd && pwd -P`, and re-appends the literal tail. Covered by Case 26.
- **Symlinks** — `pwd -P` resolves them physically (same guarantee as `_resolve_real_path`).
- **Paths with spaces** — every expansion is quoted and output uses `printf '%s/%s'`; safe.
- **Already-canonical path** — idempotent (walk finds an existing leaf, `pwd -P` returns the same string). It even handles the `dir="/"` case more correctly than the function it mirrors (avoids a double-slash `//tail`).
- **Non-absolute guard** — defensive; in practice `_portfolio_resolve` only ever passes an absolute `$abs`, and the no-root branch returns before reaching the canonicalizer.

### Blast radius (changes every `_portfolio_resolve` caller)
Independently checked all callers — none depend on the non-canonical string:
- `require-active-ticket.sh` / `require-migration-ticket.sh` — the `"$WORKSPACE_DIR"/*)` prefix match is exactly what this repairs. `require-active-ticket.sh` already re-canonicalizes WORKSPACE_DIR with `cd && pwd -P` at line 381, so the interaction is idempotent.
- `print-portfolio-primer.sh` and `portfolio_validate()` — both compare paths for in-fork-vs-sibling. The fix actually **corrects** a latent bug here: a non-canonical `.../apexyard/../apexyard-portfolio/workspace` string-matched `"$ROOT"/*` and was mis-classified as in-fork; the canonical form now reads as outside correctly. Verified `git rev-parse --show-toplevel` returns a physical (symlink-resolved) root, so canonicalizing an in-fork default remains idempotent and the raw-`$root` `case` checks in `portfolio_validate` stay matching — no regression.
- `_lib-multi-repo-trace.sh`, `maintain-docs-index.sh`, `_lib-audit-history.sh` — pass the value to `cd`/`find`/`git`, all of which tolerate (or benefit from) a canonical path.

### Scope boundary
`portfolio_validate()` is untouched — the #951/PR#952 territory is respected, as claimed.

### CI
All checks green: hook test suite (2m13s), shellcheck, Analyze, Verify Ticket ID. (CodeQL skipped — not applicable to shell.) The 3 locally-failing test files the author notes are not in the CI-authoritative failure set (the full hook suite passes in CI), consistent with the "sandbox CWD quirk" explanation.

### Issues Found
None.

### Suggestions
None blocking. Optional: since `_portfolio_canonicalize()` and `require-active-ticket.sh`'s `_resolve_real_path()` are now near-identical, a future refactor could extract one shared helper — but keeping them independent (fail-soft vs fail-closed semantics differ deliberately) is a defensible call and out of scope here.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `958e55ac31d7766b3a1e8ecac5b269b130a9911c`

Issue #943, opened 2026-07-16
[Bug] /update does not refresh an installed Codex adapter


hossam-96:
## Affected
`.claude/skills/update/SKILL.md` and `bin/sync-codex-adapter.sh`

## Given / When / Then
**Given** an ApexYard fork already has the generated Codex adapter installed
**When** the operator runs `/update` after canonical `.claude` skills, agents, or hooks change
**Then** `.agents/skills` and `.codex` can remain stale, and the up-to-date path exits before reconciliation — expected `/update` to refresh and verify an already-installed adapter, including when no framework commits need syncing

## Repro
1. Generate the Codex adapter, then update the canonical `.claude` runtime so generated output drifts.
2. Run `/update` when the fork is otherwise current and check for a newly added skill via Codex or `bin/sync-codex-adapter.sh --check`.

## Framework version
v5.1.0

## Severity
major

## Notes
Detect an existing installation from durable adapter state, reconcile it before the up-to-date early exit, and leave uninstalled harnesses untouched.

## Glossary
| Term | Definition |
|------|------------|
| Codex adapter | Generated ApexYard skills, agents, and hook wiring consumed by Codex |
| Reconciliation | Regenerating derived adapter files from the canonical `.claude` runtime and verifying they match |

atlas-apex:
Fixed by #954 (carrying @hossam-96's original fix from #944, authorship preserved). `/update` now reconciles an installed Codex adapter — warn-only at session start, real reconcile on the explicit `/update` path, scoped to adapter-owned files. Ships in the next release. Thanks @hossam-96!

PR #944, opened 2026-07-16
fix(#943): reconcile installed Codex adapters during updates

hossam-96:
## Summary

- **Installed-adapter reconciliation** — `bin/sync-codex-adapter.sh` now emits an ownership manifest and supports `--reconcile-installed`, so `/update` can refresh ApexYard-owned Codex output without creating harness files for adopters who never installed it.
- **Self-healing update lifecycle** — `/update` reconciles detected Codex adapters on both already-current and post-sync success paths, while the already-delegated upstream-drift SessionStart hook bootstraps pre-fix generated `$update` skills without blocking startup.
- **Regression coverage and operator guidance** — focused tests cover manifest drift, symlink escape protection, complete and partial legacy shapes, dry-run, skip, strict failure, and advisory bootstrap behavior; the adapter docs and AgDR explain the ownership boundary and hook-trust limitation.

## Design

- Decision record: `docs/agdr/AgDR-0102-update-reconciles-installed-codex-adapter.md`
- Preserves `.claude/` as the canonical runtime established by AgDR-0088.
- Requires all three legacy generated surfaces before claiming a pre-manifest adapter: `.agents/skills/`, `.codex/agents/`, and `.codex/hooks.json`.
- Rejects symlinked generated roots and critical child paths before detection or mutation, preventing automatic reconciliation from escaping the repository.

## Testing

- `bash .claude/hooks/tests/test_sync_codex_adapter.sh` — 52 passed
- `bash .claude/hooks/tests/test_update_codex_adapter_reconcile.sh` — 10 passed
- `bash .claude/hooks/tests/test_check_upstream_drift.sh` — 8 passed
- Existing `/update` suites — 43 passed across from-dev, migration-chain, and step-8a wiring tests
- `npx --yes markdownlint-cli2 '**/*.md' '#node_modules' '#.git'` — 467 files, 0 errors
- ShellCheck `-S warning` on all changed shell files — clean
- Local full `bin/run-hook-tests.sh` — 112/113 files passed; the sole failure is the pre-existing GitLab-wrapper pair in `test_block_merge_on_red_ci.sh`, reproduced unchanged from a clean `upstream/dev` worktree
- GitHub Actions hook test suite — passed on PR #944; the clean CI environment does not reproduce the local baseline-only GitLab wrapper failure

## Glossary

| Term | Definition |
|------|------------|
| Adapter manifest | Generated metadata that identifies `.agents/` and `.codex/` output as owned by ApexYard's Codex adapter |
[...]

PR #954, opened 2026-07-20
fix(#943): reconcile installed Codex adapters on /update (carry-forward of #944)

atlas-apex:
## Summary

Carries forward @hossam-96's #944, preserving their authorship on the core commits (`2a7891e4`, `84a27881` — both re-applied here via cherry-pick, unchanged), and adds the two review changes the panel flagged before merge:

- **Their fix (unchanged, preserved authorship)** — `/update` now reconciles an installed Codex adapter (manifest-backed or a recognizable pre-manifest legacy shape) so an existing Codex install doesn't silently drift behind `.claude/` after a sync, including on the already-current path. A symlink guard rejects a generated root or owned child path that's been redirected outside the repo before any read or write touches it.
- **Change A — the SessionStart hook no longer mutates the tree.** The original PR had `check-upstream-drift.sh` fully regenerate the Codex adapter at *every* session boot — a permanent tree mutation on every fork, every session, forever. That's too broad a blast radius for a hook that fires unconditionally. It now runs the generator's new **read-only** `--check-installed` mode and prints a one-line advisory (`Codex adapter may be stale — run bash bin/sync-codex-adapter.sh --reconcile-installed`) when the installed adapter has drifted. The actual reconciling write stays exactly where #944 put it: the explicit `/update` path (`--reconcile-installed`), or a manual run. The reconciler invocation is wrapped in the same `timeout`-guarded pattern as the hook's sibling `git fetch` call, so a hung generator can't stall session startup.
- **Change B — drift detection is scoped to the adapter's owned paths.** The generator's `--check` / `--reconcile-installed` previously ran a whole-tree `diff -qr` across `.agents/` and `.codex/`. Any non-adapter file living there — a hand-authored Codex `config.toml`, a runtime cache — was misreported as drift: a false-positive per-session warning, and worse, a false-positive **hard failure** on `/update`'s strict path (the post-write verification inside `--reconcile-installed` would fail even though every owned path was perfectly in sync). Drift detection (`--check`, the new `--check-installed`, and `--reconcile-installed`'s own before/after checks) now walks only `.agents/skills`, `.codex/agents`, `.codex/hooks.json`, and `.codex/apexyard-adapter.json` — the exact paths the generator owns and already uses for its symlink guard and cleanup list.
- **Two small wording/robustness cleanups** — `SKILL.md`'s step 8d and the hook's own comment described the backstop as "bootstrapping the new skill" (accurate for the old mutating design, stale now); reworded to describe the advisory-only behavior. The ownership manifest itself is untouched — it's still exactly the metadata file #944 designed.

## Testing

Ran all three test files touching this change, plus a syntax check on both modified scripts:

```
$ bash .claude/hooks/tests/test_sync_codex_adapter.sh
...
Passed: 57
Failed: 2   (generated hook command preserves blocking/allowing exit — pre-existing,
             unrelated to this change; confirmed failing identically on upstream/dev
             before this PR's commits, due to a missing local fixture file)

$ bash .claude/hooks/tests/test_check_upstream_drift.sh
PASS: 8   FAIL: 0

$ bash .claude/hooks/tests/test_update_codex_adapter_reconcile.sh
Passed: 10
Failed: 0
```

New/rewritten coverage added on top of #944's suite:

1. `test_check_upstream_drift.sh` — rewrote the SessionStart bootstrap case: an installed, drifted legacy adapter now gets a `Codex adapter may be stale` advisory **and the tree is provably NOT mutated** (no manifest written, no skill copied) — the prior version of this test asserted the opposite (that SessionStart *did* refresh the files), which was exactly the boot-time mutation behavior this PR removes.
2. `test_sync_codex_adapter.sh` — new "scoped-owned-paths" fixture: a hand-authored `.codex/config.toml` and `.agents/notes/todo.md` are proven to (a) never trigger `--check` / `--check-installed` drift, (b) survive `--reconcile-installed` untouched, and (c) not cause a false-positive hard failure. Also added a `--check-installed` baseline test proving it *does* catch real owned-path drift and never writes.
3. `test_update_codex_adapter_reconcile.sh` — unchanged; still passes, since `/update`'s own `--reconcile-installed` call site and contract are untouched by this PR.

## Glossary

| Term | Definition |
|------|------------|
| Codex adapter | Generated `.agents/` + `.codex/` output that lets OpenAI Codex consume ApexYard's canonical `.claude/` skills, agents, and hook wiring without a second hand-maintained copy. |
| Reconcile (installed) | Refresh a Codex adapter that's already installed in a fork, detected via an ownership manifest or a recognizable pre-manifest directory shape — as opposed to unconditionally generating one for every fork regardless of whether Codex is even in use. |
| `--check-installed` (new in this PR) | Read-only sibling of `--reconcile-installed`: same installation detection and same scoped drift comparison, but never writes to disk. Used by the SessionStart advisory. |
| Adapter-owned paths | The specific generated subpaths (`.agents/skills`, `.codex/agents`, `.codex/hooks.json`, `.codex/apexyard-adapter.json`) the generator creates and is allowed to compare/mutate — as opposed to everything else that might live under `.agents/` or `.codex/`. |
| SessionStart hook | A Claude Code hook (`check-upstream-drift.sh`) that runs once at the start of every session in a fork — hence why a mutation there has such wide blast radius. |

Refs #943

---

Trust-chain note: this PR touches `.claude/hooks/check-upstream-drift.sh` and `bin/sync-codex-adapter.sh` (a script the trust-chain hooks delegate to at session boot), so per `.claude/rules/role-triggers.md` a Security Auditor (Hakim) review is expected before merge, in addition to Rex.

### DIFF
--- a/.claude/hooks/check-upstream-drift.sh
+++ b/.claude/hooks/check-upstream-drift.sh
@@ -36,19 +36,23 @@ fi
 
 cd "$REPO_ROOT" || exit 0
 
-# Bootstrap generated Codex adapters after a framework update. The /update
-# skill is itself generated under .agents/, so an adopter invoking a pre-fix
-# copy cannot execute reconciliation instructions that only arrived in the
-# newly merged canonical skill. Existing Codex hook wiring already delegates
-# this SessionStart hook to .claude/hooks/, giving the first restart a stable
-# path to install the refreshed skill. The generator owns detection and is a
-# silent no-op for uninstalled/partial adapters. Startup stays advisory: warn
-# on failure, but never block the session.
+# Warn (never mutate) when an installed Codex adapter looks stale. SessionStart
+# is a boot-time hook that runs on every session, so it must never write to
+# the tree itself — reconciliation is a deliberate, explicit action owned by
+# `/update` (via `bin/sync-codex-adapter.sh --reconcile-installed`) or a manual
+# operator run. This hook only runs the generator's read-only
+# `--check-installed` mode (no writes, ever) and prints a one-line advisory
+# nudge on drift. The generator owns installation detection and is a silent
+# no-op for uninstalled adapters. Startup stays advisory: warn on drift or
+# failure, but never block the session, and never touch the working tree.
 CODEX_RECONCILER="$REPO_ROOT/bin/sync-codex-adapter.sh"
 if [ -f "$CODEX_RECONCILER" ]; then
+  _CODEX_TO=""
+  if command -v timeout >/dev/null 2>&1; then _CODEX_TO="timeout -k 2 5"
+  elif command -v gtimeout >/dev/null 2>&1; then _CODEX_TO="gtimeout -k 2 5"; fi
   ADAPTER_ERROR=""
-  if ! ADAPTER_ERROR=$(bash "$CODEX_RECONCILER" --reconcile-installed 2>&1 >/dev/null); then
-    echo "ApexYard: installed Codex adapter could not be refreshed; run bash bin/sync-codex-adapter.sh --reconcile-installed." >&2
+  if ! ADAPTER_ERROR=$($_CODEX_TO bash "$CODEX_RECONCILER" --check-installed 2>&1 >/dev/null); then
+    echo "ApexYard: installed Codex adapter may be stale — run bash bin/sync-codex-adapter.sh --reconcile-installed." >&2
     [ -n "$ADAPTER_ERROR" ] && printf '%s\n' "$ADAPTER_ERROR" >&2
   fi
 fi
--- a/.claude/hooks/check-upstream-drift.sh
+++ b/.claude/hooks/check-upstream-drift.sh
@@ -36,6 +36,23 @@ fi
 
 cd "$REPO_ROOT" || exit 0
 
+# Bootstrap generated Codex adapters after a framework update. The /update
+# skill is itself generated under .agents/, so an adopter invoking a pre-fix
+# copy cannot execute reconciliation instructions that only arrived in the
+# newly merged canonical skill. Existing Codex hook wiring already delegates
+# this SessionStart hook to .claude/hooks/, giving the first restart a stable
+# path to install the refreshed skill. The generator owns detection and is a
+# silent no-op for uninstalled/partial adapters. Startup stays advisory: warn
+# on failure, but never block the session.
+CODEX_RECONCILER="$REPO_ROOT/bin/sync-codex-adapter.sh"
+if [ -f "$CODEX_RECONCILER" ]; then
+  ADAPTER_ERROR=""
+  if ! ADAPTER_ERROR=$(bash "$CODEX_RECONCILER" --reconcile-installed 2>&1 >/dev/null); then
+    echo "ApexYard: installed Codex adapter could not be refreshed; run bash bin/sync-codex-adapter.sh --reconcile-installed." >&2
+    [ -n "$ADAPTER_ERROR" ] && printf '%s\n' "$ADAPTER_ERROR" >&2
+  fi
+fi
+
 # Bail if no upstream remote — either this IS the upstream repo, or the fork
 # owner hasn't run `git remote add upstream …` yet. Either case: silent.
 if ! git remote | grep -qx upstream; then
--- a/.claude/hooks/tests/test_check_upstream_drift.sh
+++ b/.claude/hooks/tests/test_check_upstream_drift.sh
@@ -159,9 +159,11 @@ case_squash_no_changelog() {
   rm -rf "$up" "$(dirname "$fk")"
 }
 
-# CASE 6: an installed pre-manifest Codex adapter is refreshed through the
-# canonical SessionStart hook after new framework skills land.
-case_codex_bootstrap_refresh() {
+# CASE 6: an installed pre-manifest Codex adapter has drifted (new framework
+# skills landed). The SessionStart hook must WARN about the drift but must
+# NEVER mutate the tree itself — reconciliation is owned exclusively by an
+# explicit `/update` or a manual `--reconcile-installed` run.
+case_codex_bootstrap_warns_without_mutating() {
   local up; up=$(make_upstream)
   local fk; fk=$(make_fork "$up")
   (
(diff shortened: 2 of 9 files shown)


atlas-apex (commented):
## Code Review: PR #954

**Commit**: `5904eb1ae309001204ac402cec6be14cb144774e`

### Summary
Carry-forward of @hossam-96's #944 (fixes #943 — `/update` didn't refresh an installed Codex adapter), preserving their authorship on the two core commits, plus two panel-requested review changes:
- **Change A** — the SessionStart hook (`check-upstream-drift.sh`) no longer regenerates the adapter at boot; it runs the generator's new read-only `--check-installed` mode and prints a one-line, timeout-guarded advisory when the installed adapter has drifted. The mutating write stays on the explicit `/update` path.
- **Change B** — drift detection (`--check`, `--check-installed`, `--reconcile-installed`) is scoped to the 4 adapter-owned paths instead of a whole-tree `diff -qr` of `.agents/` / `.codex/`, so a hand-authored `config.toml` / scratch file no longer false-triggers drift (and no longer causes a false-positive hard failure on `/update`'s strict path).

### Checklist Results
- Security:                  Pass (trust-chain touch — see note below)
- Performance:              Pass (timeout-guarded; early-exit-when-current optimization)
- Technical Decisions (AgDR): Pass — AgDR-0102 present and linked

### Verification against the review focus

**Change A — boot-time mutation removed.** Confirmed. The added block calls `--check-installed` (read-only), writes only to stderr, and never exits non-zero — the hook still returns 0 and never blocks the session. `timeout -k 2 5` (with a `gtimeout` fallback) wraps the reconciler so a hung generator can't stall startup. The `2>&1 >/dev/null` ordering correctly captures the DRIFT text into `$ADAPTER_ERROR` while discarding stdout. Placing the block before the `upstream`-remote bail is right: adapter drift is independent of remote config, and `--check-installed` is a silent no-op when Codex isn't installed.

**Change B — scoping correct.** `ADAPTER_OWNED_PATHS` (`.agents/skills`, `.codex/agents`, `.codex/hooks.json`, `.codex/apexyard-adapter.json`) drives every comparison. `check_drift` was rewritten from a whole-tree `diff -qr` to a per-owned-path loop keyed on the label prefix — a non-owned file under `.codex/`/`.agents/` is invisible to drift detection and untouched by reconciliation, while real owned-path drift is still caught. The legacy-adapter path is handled correctly: a complete-shape adapter with no manifest reports drift on the missing `apexyard-adapter.json`, so it falls through to a full regen that writes the manifest (matches the "legacy reconciliation writes ownership manifest" test).

**Docs match shipped behavior.** SKILL.md §8d, `docs/codex-adapter.md`, `docs/harnesses/codex.md`, and AgDR-0102 all describe the SessionStart hook as read-only + warn-only ("never writes to the tree", "only ever advises, it does not self-heal at boot"). No stale mutating-design language survives.

**Tests prove all three properties.**
- (a) No tree mutation on SessionStart drift — CASE 6 (`warns_without_mutating`) asserts the advisory fires AND the manifest/skill are provably not written; CASE 8 additionally asserts rc=0 + advisory + no manifest when the check itself fails.
- (b) Non-adapter files don't false-trigger — the `scoped-owned-paths` fixture proves `--check`, `--check-installed`, and `--reconcile-installed` all succeed with a hand-authored `.codex/config.toml` and `.agents/notes/todo.md` present, and that both survive untouched.
- (c) `--check-installed` catches real drift without writing — after a source skill is drifted, `--check-installed` exits non-zero AND the generated output is asserted un-mutated.
- Plus 6 symlink-escape fixtures with external sentinels proving reconciliation can't write outside the repo (the guard now runs unconditionally, before installation detection or any mutation).

**CI:** green — hook test suite, shellcheck, markdownlint, Analyze, Verify Ticket ID all pass; CodeQL skipped.

### Issues Found
None blocking.

### Suggestions (non-blocking nits)
- `nit:` `test_update_codex_adapter_reconcile.sh` re-implements the documented `reconcile_installed_codex_adapter` recipe inline and calls the generator with an explicit `--root "$root"`, whereas the SKILL.md recipe relies on the generator's default ROOT (`git rev-parse --show-toplevel`). It's a legitimate contract test (the header says so), but the `--root` divergence means it wouldn't catch a ROOT-resolution regression in the literal skill recipe. Consider exercising the default-ROOT path in at least one case.
- `note:` The scoping trade-off means stale *legacy cleanup* paths (`.codex/rules`, `.codex/migrations`, etc. in the `rm -rf` list but not in `ADAPTER_OWNED_PATHS`) won't trigger a SessionStart drift nudge, though they're still cleaned on the next real reconcile. This is a reasonable, AgDR-documented trade-off — flagging only for awareness.
- `note:` This PR touches the trust chain (`.claude/hooks/check-upstream-drift.sh` + `bin/sync-codex-adapter.sh`, now executed read-only at session boot). As the author correctly notes, a Security Auditor (Hakim) review is expected as a separate gate. My APPROVED verdict is on code-quality/correctness grounds and does not substitute for that security pass.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `5904eb1ae309001204ac402cec6be14cb144774e`

atlas-apex (commented):
## Security Review: PR #954

**Commit**: `5904eb1ae309001204ac402cec6be14cb144774e`

🛡️ Trust-chain review (Hakim, Security Auditor) — this PR modifies a SessionStart hook (`check-upstream-drift.sh`) and the Codex adapter generator (`bin/sync-codex-adapter.sh`), so it falls under the `.claude/hooks/**` trust-chain trigger.

### Summary
Adds a read-only `--check-installed` mode to the Codex generator and wires it into the SessionStart drift hook as an **advisory-only** nudge; adds a symlink guard on adapter output paths; and scopes drift detection to a hardcoded list of adapter-owned paths. I reviewed the four focus areas (symlink handling, boot-time execution, scoped diff, reconcile write path).

### Checklist Results
- XSS Prevention:         N/A (shell tooling, no rendering surface)
- Auth & Authorisation:   N/A
- Boot-time / trust chain: Pass

### Focus-area findings

**1. Symlink handling — SOUND.** `assert_safe_output_paths` rejects a symlink at either adapter root (`.agents`, `.codex`) **and** at every owned child (`.agents/skills`, `.codex/agents`, `.codex/hooks.json`, `.codex/apexyard-adapter.json`). Critically, the guard runs `assert_safe_output_paths || exit 1` **before** installation detection reads the manifest and before any generation or write, so a symlink planted at rest in the tree redirecting a later reconcile is caught on every invocation. The write path also `rm -rf`s each owned path before `cp` — and `rm` on a symlink unlinks the link rather than following it, so even a swapped-in child symlink is neutralised before the copy. Verified by six new fixtures (`agents-root`, `codex-root`, `skills-child`, `agents-child`, `hooks-child`, `manifest-child`), each asserting a non-zero exit **and** an untouched external sentinel.

**2. Boot-time execution — SAFE.** The SessionStart hook calls only the read-only `--check-installed` mode; confirmed no writes on that path (generation lands in a `mktemp` temp dir with a cleanup trap; only the `--reconcile-installed`/default write paths `cp` into the tree, and neither runs at boot). It is `timeout -k 2 5` guarded (with `gtimeout` fallback), matching the sibling `git fetch` pattern, so a hung generator cannot stall startup. The hook never exits non-zero on this branch — it echoes to stderr and continues — proven by CASE 8 (synthetic generator failure → hook rc=0, warns, no mutation). No attacker-controlled code execution: the generator reads `.codex/` files only via `grep`/`diff` (no eval), and all `jq` generation reads the trusted canonical `.claude/` tree, not attacker-influenced `.codex/` content. A crafted `.codex/` file can at most be reported as drift.

**3. Scoped diff — NOT MANIPULABLE.** `ADAPTER_OWNED_PATHS` is a **hardcoded array in the script**, not read from the ownership manifest or any on-disk file. The manifest (`.codex/apexyard-adapter.json`) is used only by `codex_adapter_installed` as an installation *signal* (grep for the adapter string); it cannot expand or redirect the compared path set. So the concern "manipulate the manifest to skip a security-relevant file or walk outside the tree" does not apply — the walk set is static and the `check_drift` `case "$rel" in "$label"/*)` matching uses only literal `.agents`/`.codex` labels against that fixed array.

**4. Reconcile write path — CONFINED.** Writes are `cp -R` from the trusted temp build (generated from canonical `.claude/`) to the four fixed, symlink-guarded owned paths. Filenames derive from `.claude/skills/*` and `.claude/agents/*` (framework-controlled), not attacker input — no filename/path injection, no arbitrary-file overwrite. A crafted manifest can flip installed-detection to true, but that only triggers a regenerate-from-canonical (no escalation); for uninstalled forks it correctly stays a silent no-op (verified by the no-install and partial-shape fixtures).

### Security Issues Found
None blocking.

- **LOW / informational — check-then-use on the symlink guard (TOCTOU).** `assert_safe_output_paths` validates once, then generation + write happen later. A theoretical bypass exists if an attacker swaps an adapter **parent** dir (`.codex`) to a symlink in the window between guard and write (`mkdir -p` would then operate through it). This requires concurrent write access to the repo working tree — a precondition that already lets an attacker edit `.claude/hooks/**` directly, so it grants no capability beyond the existing trust boundary. Not blocking; noted for defense-in-depth (a future `realpath`-confinement check on each target immediately before write would close even the theoretical race).

### Recommendations
- (Non-blocking) Consider re-asserting `assert_safe_output_paths` (or a per-target `realpath`-under-`$ROOT` check) immediately before the `rm -rf`/`cp` block to harden against the theoretical parent-dir TOCTOU.
- (Non-blocking) When neither `timeout` nor `gtimeout` is present, `--check-installed` runs unbounded at boot. Bounded by read-only + trusted inputs, so low risk, but worth an inline note that this matches the existing `git fetch` fallback behavior.

### CI
Green — hook test suite, shellcheck `.claude/hooks`, CodeQL analyze, markdownlint, ticket-ID all pass at HEAD.

### Verdict
**PASS** — no blocking security issues. The symlink guard is well-placed (pre-mutation) and well-tested, the boot-time path is genuinely read-only / timeout-guarded / non-blocking, the scoped path list is hardcoded (not manifest-manipulable), and the reconcile writes are confined to trusted-sourced, symlink-guarded paths. One LOW informational TOCTOU note that sits behind an already-game-over precondition.

---
📌 Reviewed commit: `5904eb1ae309001204ac402cec6be14cb144774e`

me2resh (approved):

Issue #950, opened 2026-07-20
[Bug] Portfolio/config libs resolve to the wrong root when sourced under zsh (${BASH_SOURCE[0]} empty)


atlas-apex:
## Affected
`.claude/hooks/_lib-read-config.sh` (and the `_lib-portfolio-paths.sh` helpers that depend on it)

## Given / When / Then
**Given** the config/portfolio helpers self-locate their directory via `lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` (`_lib-read-config.sh:62`)
**When** an operator `source`s them from an interactive **zsh** shell — e.g. to manually verify portfolio path resolution during setup
**Then** `${BASH_SOURCE[0]}` is a bash-only feature and is **empty under zsh**, so `lib_dir` silently resolves to the wrong directory and the helpers return misleading paths — **expected**: either work under zsh, or emit a one-line "source this under bash" warning instead of failing silently.

## Repro
1. In a zsh shell: `source .claude/hooks/_lib-read-config.sh; source .claude/hooks/_lib-portfolio-paths.sh`
2. Call `portfolio_workspace_dir` (or `portfolio_registry`)
3. Observe an empty/wrong root. Note: the *shipped* hooks run under `#!/bin/bash` and are unaffected — this only bites manual verification, which is exactly when someone is debugging their config and gets sent chasing a phantom bug.

## Framework version
v5.1.0

## Severity
minor (real hook execution is unaffected; the cost is wasted debugging time during manual setup verification)

## Notes
No `ZSH_VERSION` guard exists anywhere under `.claude/hooks/`. Suggested fix: in the self-locating libs, detect `[ -n "${ZSH_VERSION:-}" ]` and either fall back to zsh's `${(%):-%x}` or print a clear "run under bash" warning.

## Glossary
| Term | Definition |
|------|------------|
| `${BASH_SOURCE[0]}` | Bash array holding the path of the currently-sourced file; used to locate a lib relative to itself. Empty in zsh. |
| Self-locating lib | A sourced shell library that resolves its own directory so it can find sibling files regardless of the caller's cwd. |

atlas-apex:
[...]

PR #960, opened 2026-07-21
fix(#950): warn when portfolio/config libs are sourced under zsh

atlas-apex:
## Summary

- **Loud advisory instead of silent wrong-resolution under zsh** — `_lib-read-config.sh` self-locates via `${BASH_SOURCE[0]}`, a bash-only feature that is empty under zsh. An operator who `source`s these helpers in an interactive zsh shell to verify config/portfolio path resolution (the setup-guide snippet does exactly this) would otherwise silently resolve to the wrong directory and chase a phantom bug. The config-root resolver now emits a one-line advisory under zsh — "source these under bash" — so the failure is loud, not silent. Shipped hooks are all `#!/bin/bash` and are unaffected; this only bites manual sourcing.
- **Why warn-only rather than a zsh-native self-location fix** — a real zsh expansion (`${(%):-%x}`) was tried and rejected: shellcheck (which parses everything as bash) errors on the nested zsh-only form (SC2298), and the file has other bash-only constructs that don't source cleanly under zsh anyway. "Run it under bash" is the honest, robust guidance, and it can't hang or break shellcheck.

## Testing

1. `shellcheck -x .claude/hooks/_lib-read-config.sh` — clean
2. bash: `. _lib-read-config.sh && config_get_or …` resolves as before (no regression)
3. zsh: `zsh -c '. _lib-read-config.sh; config_get_or .x def'` — the advisory fires when the resolver runs (the exact manual-verification path)

Refs #950

_Touches `.claude/hooks/**` (trust chain); Security Auditor (Hakim) review expected in addition to Rex._

## Glossary

| Term | Definition |
|------|------------|
| `${BASH_SOURCE[0]}` | Bash array holding the currently-sourced file's path; used for self-location. Empty under zsh. |
| Self-location | A sourced library resolving its own directory so it can find sibling files regardless of the caller's cwd. |

### DIFF
--- a/.claude/hooks/_lib-read-config.sh
+++ b/.claude/hooks/_lib-read-config.sh
@@ -59,6 +59,18 @@ _config_repo_root() {
   # file in .claude/hooks/, so locate it via BASH_SOURCE — not via
   # `git rev-parse` (which would defeat the whole point of this fix).
   local lib_dir
+  # zsh-safe warning (me2resh/apexyard#950): these helpers are #!/bin/bash and
+  # rely on ${BASH_SOURCE[0]} for self-location, which is a bash-only feature —
+  # it is EMPTY under zsh, so an operator who `source`s this lib in an
+  # interactive zsh shell (a common manual config-verification move during
+  # setup) would silently resolve to the wrong dir and chase a phantom bug.
+  # Emit a one-line advisory so the failure is loud, not silent. (A real zsh
+  # self-location expansion was rejected — this file has other bash-only
+  # constructs that don't source cleanly under zsh anyway, so "run it under
+  # bash" is the honest guidance.)
+  if [ -n "${ZSH_VERSION:-}" ]; then
+    printf 'apexyard: source the .claude/hooks/_lib-*.sh helpers under bash, not zsh — path resolution is unreliable under zsh. Wrap manual checks in `bash -c '"'"'...'"'"'`.\n' >&2
+  fi
   lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
   if [ -f "$lib_dir/_lib-ops-root.sh" ]; then
     # shellcheck source=/dev/null


atlas-apex (commented):
## Security Review: PR #960

**Commit**: `0e20f3c4c8853dce3980b8c5f1fc979025533747`

### Summary

Trust-chain change to `.claude/hooks/_lib-read-config.sh` — the config reader every hook and the portfolio-path helper depend on. The diff adds a `[ -n "${ZSH_VERSION:-}" ]` branch that `printf`s a one-line advisory to stderr, then falls through to the unchanged `${BASH_SOURCE[0]}` self-location. 12 added lines, 0 deleted, 1 file. I verified each security property empirically rather than by reading alone, because a stdout leak here would corrupt resolved config paths process-wide.

### Checklist Results

- Secrets & Credentials:  Pass (no secrets, no credential handling touched)
- Injection Prevention:   Pass (see format-string + env-var analysis below)
- XSS Prevention:         N/A (shell library, no rendering surface)
- Auth & Authorisation:   N/A (no authz logic; gate-marker logic untouched)
- Data Protection:        Pass (no data handling changed)
- API Security:           N/A

### Verification performed

**1. Output-channel safety — the load-bearing check.** `_config_repo_root` is consumed via command substitution at four call sites (`_config_defaults_file` L94, `_config_overrides_file` L100, and transitively `config_get` L141 / `config_get_or` L166). Command substitution captures stdout only. Confirmed by execution with `ZSH_VERSION` forced under bash:

- stdout capture: the resolved path **only** — advisory absent from `$(...)`
- stderr: the advisory line, passed through to the caller's stderr

The `>&2` redirect is present and correct; there is no path by which the advisory can contaminate a resolved config path. **Pass.**

**2. Bash path is a true no-op.** Reconstructed the pre-change file (deleting the added block) and diffed behaviour. With `ZSH_VERSION` unset: stdout **identical**, and **zero bytes** written to stderr. The `lib_dir=` assignment sits outside the `if` and is unchanged, so resolution is byte-identical for every shipped hook. Independently confirmed the premise: every one of the 60+ hooks/libs that source this file uses `#!/bin/bash` or `#!/usr/bin/env bash` — no `#!/bin/sh` consumer exists. **No new failure mode in the trust chain.**

**3. Format string.** `printf 'literal…\n' >&2` — single-quoted literal throughout, embedded quotes via the standard `'"'"'` idiom. The string contains **no `%` characters at all**, so there are no format directives, and no arguments are passed. No variable appears in the format-string position. Format-string injection is structurally impossible here. **Pass.**

**4. Can `ZSH_VERSION` influence resolution?** No. The variable is used solely as the operand of a `[ -n … ]` test; its *value* is never expanded, interpolated, or echoed. Verified with a hostile value (`$(touch /tmp/PWNED)` + backticks + `%s%n%%`): output was the fixed literal advisory, resolution was byte-identical to the unset case, and no side-effect files were created. An attacker with env control could at most cause one fixed, non-attacker-controlled line of known text to appear on stderr — and env-var control already implies strictly greater capabilities (`PATH`, etc.). The branch **warns only; it does not alter the resolution path.** No integrity impact.

### Security Issues Found

None. No Critical, High, or Medium findings.

### Recommendations (non-blocking, LOW / informational)

- **LOW — advisory repeats when the root resolves empty.** The `_CONFIG_ROOT_CACHE` early-return guards on `[ -n … ]`, so an empty resolution (outside any git repo, no ops anchor) misses the cache and re-walks. Measured: **3 advisory lines across 3 calls** in that case, vs. 1 when the root is non-empty. Pre-existing cache-miss behaviour, merely made visible by this change — stderr noise only, no correctness or integrity impact. Not worth blocking; note it if the empty-root path ever gets a cache fix.
- **LOW — `sh`/`dash` remains silently wrong.** `BASH_SOURCE` is equally empty under dash, where `ZSH_VERSION` is unset, so that path stays silent. Out of scope for #950 (which targets the operator's interactive zsh on macOS) and unreachable today given every consumer is bash-shebanged. Flagging only so the residual gap is on record.
- The warn-only approach over a zsh-native `${(%):-%x}` expansion is the right security call: it keeps the resolution path untouched, so the change cannot regress path resolution for the hooks that actually enforce the merge gates.

### CI

All checks green at this commit — shellcheck, hook test suite (2m22s), sub-pack smoke test, Analyze (actions), Verify Ticket ID. CodeQL skipped (no compiled-language change).

### Verdict

**PASS — APPROVED (no blocking security issues)**

Security marker written at `0e20f3c4c8853dce3980b8c5f1fc979025533747`. Not merging — merge remains gated on the per-PR CEO nod.

---
📌 Reviewed commit: `0e20f3c4c8853dce3980b8c5f1fc979025533747`

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #960

**Commit**: `0e20f3c4c8853dce3980b8c5f1fc979025533747`

### Summary

Adds a one-line stderr advisory in `_config_repo_root` when `ZSH_VERSION` is set, so an operator sourcing `_lib-read-config.sh` in an interactive zsh shell gets a loud "run this under bash" instead of a silent wrong-directory resolution. 12 lines added, one file, no behaviour change under bash.

I verified the load-bearing claims independently rather than taking the PR body's word for them.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (with a nit — see below)
- Security: **Pass**
- Performance: **Pass**
- Summary Bullet Narrative: **Pass** (both bullets lead with a verb and carry the *why*)
- Technical Decisions (AgDR): **N/A**
- Adopter Handbooks: **N/A** (shell diff; only `language/typescript/` exists, `migration-safety.md` is blocking but no migration paths touched)

### Verification performed

**1. The underlying bug is real.** From a foreign cwd:

```
zsh  → resolved_root=[]                             ← silently empty
bash → resolved_root=[/Users/ahmed/apex/apexyard]   ← correct
```

`${BASH_SOURCE[0]}` is empty under zsh → `dirname ""` → `.` → resolves to cwd, which then fails the ops-fork anchor walk. Confirmed the failure mode the advisory warns about.

**2. No bash regression.** Compared `dev` vs head stdout on `_config_repo_root; config_get_or ".tracker.kind"` — **byte-identical**, and stderr is **empty** under bash. The `[ -n "${ZSH_VERSION:-}" ]` guard is never true under bash, and the `:-` handles `set -u`. Clean.

**3. stdout is not corrupted — this was the key risk and it is handled correctly.** The resolver's output is consumed via `$( )` by both `_config_defaults_file` and `_config_overrides_file`. Verified under zsh:

```
zsh -c 'v=$(. _lib-read-config.sh; config_get_or ".tracker.kind" "MISSING"); echo "captured=[$v]"'
→ captured=[gh]
```

Advisory goes to `>&2`, capture stays clean. Also confirmed no `%` in the `printf` format string, and the `'"'"'` nesting emits the intended literal backticked `bash -c '...'`.

**4. The advisory fires at the right moment.** Bare `source` alone emits nothing; the advisory fires on the first resolver call. That is the correct placement — bare-sourcing is harmless, the harm starts when resolution runs. An operator who sources and never resolves never had a problem to warn about.

**5. `shellcheck -x` clean; local `test_config_get_backslash_escape.sh` (2/2) and `test_portfolio_paths.sh` (39/39) pass. CI green — 5 checks pass, CodeQL skipped.**

### On warn-only vs a real zsh fix — agreed, warn-only is right

I checked the constraint the author cited and it holds. A genuine fix would need both a zsh-native self-location expansion *and* remediation of the file's other bash-only constructs — and every shipped hook is `#!/bin/bash`, so nothing in the runtime path can reach this. Fixing resolution for a path only manual operators take, at the cost of a shellcheck suppression and a dual-shell compatibility burden on a trust-chain file, is the worse trade. Making the failure loud is proportionate to its blast radius.

### Issues Found

None blocking.

### Suggestions

**nit: the advisory repeats — 2× per `config_get`, 6× for three calls.** `_config_repo_root` is invoked from both `_config_defaults_file` and `_config_overrides_file`, each inside `$( )` — a subshell — so `_CONFIG_ROOT_CACHE` is set in the child and never propagates back to the parent. The cache is effectively inert across those call sites.

Worth flagging honestly: copying the existing `_CONFIG_WARNED_NO_JQ` one-shot pattern would **not** fix this, because `_config_load` is likewise called via `$( )` and that guard is defeated the same way. So this is a **pre-existing caching quirk that this PR merely makes audible**, not a defect introduced here. Operator paths are 1–2 calls (2–4 lines) and "loud" is the stated intent, so this is genuinely non-blocking — but it's worth knowing the guard you'd reach for first doesn't work here.

**nit: a test is more feasible than "the file doesn't source cleanly under zsh" implies.** The advisory only needs to reach stderr, which it demonstrably does. This is sufficient and cheap:

```bash
command -v zsh >/dev/null && \
  zsh -c '. .claude/hooks/_lib-read-config.sh; config_get_or ".tracker.kind" "x" >/dev/null' 2>&1 >/dev/null \
  | grep -q 'under bash, not zsh' && echo ok
```

Gate it on `command -v zsh` so it skips on zsh-less CI. Non-blocking for an advisory-only string, but it would pin both the stderr routing and the non-empty message — the two properties that actually matter here.

**suggestion (follow-up ticket, out of scope for #950): five sibling libs share the identical latent issue.** `_lib-tracker.sh:74,95`, `_lib-extract-pr.sh:100`, `_lib-audit-history.sh:52`, `_lib-premium-hook.sh:149`, and `_lib-fresh-fork.sh:50` all use the same `$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)` idiom.

Two of them — `_lib-audit-history.sh` and `_lib-fresh-fork.sh` — do it at **file scope**, so under zsh they mis-resolve at *source* time, before any function runs. A function-body warning of the shape used here would not help them; they'd need a top-of-file guard. That matters because several `SKILL.md` files instruct sourcing `_lib-audit-history.sh` directly, which is the same manual-operator path #950 is about.

I verified `_lib-portfolio-paths.sh` and `_lib-ops-root.sh` self-locate by PWD-walk rather than `BASH_SOURCE`, so they are zsh-safe for self-location and inherit this warning transitively via `$root/.claude/hooks/_lib-read-config.sh` — the "portfolio/config libs" framing in the title is fair.

### Verdict

**APPROVED**

Tight, well-scoped, correctly-routed advisory. The stdout-purity property that the `$( )` callers depend on is preserved, and the bash path is provably unchanged. Nits are optional follow-ups, not merge blockers.

---
📌 Reviewed commit: `0e20f3c4c8853dce3980b8c5f1fc979025533747`

Issue #958, opened 2026-07-21
[Bug] ticket.prefix_whitelist default is missing "Migration", blocking /migration's own ticket format


aelnemr:
## Affected
`.claude/hooks/validate-issue-structure.sh` (reads `.claude/project-config.defaults.json` → `ticket.prefix_whitelist`), and by extension the `/migration` skill / `templates/tickets/migration.md`

## Given / When / Then
**Given** the `/migration` skill's own ticket template (`templates/tickets/migration.md`) hardcodes the title format `[Migration] <type>: <summary>`, and a registered project already had prior tickets created with this exact prefix (e.g. issue #405, filed via `/migration` before this gap was hit)
**When** running `/migration` end-to-end and calling `gh issue create` with a `[Migration]` title
**Then** the issue is blocked with `BLOCKED: issue title '[Migration]' uses an unrecognised prefix. Accepted prefixes: Feature, Bug, Chore, Refactor, Testing, CI, Docs, Spike, Investigation` — expected the skill's own generated title format to pass validation out of the box

## Repro
1. In a project with the framework defaults (no `ticket.prefix_whitelist` override), run `/migration` for any migration
2. Answer the interview, confirm creation
3. `gh issue create --title "[Migration] schema: ..." ...` fails the `validate-issue-structure.sh` PreToolUse hook

## Framework version
v3.0.0

## Severity
major — workaround exists (add `"Migration"` to a per-fork `ticket.prefix_whitelist` override in `.claude/project-config.json`), but every fresh adopter running `/migration` hits this cold

## Notes
Suggested fix: add `"Migration"` to the `prefix_whitelist` array in `.claude/project-config.defaults.json` (alongside `Spike`, `Investigation`, which were evidently added after the initial set). Worked around locally via the per-fork override for now.

## Glossary
| Term | Definition |
|------|------------|
| prefix_whitelist | The set of bracketed ticket-title prefixes (`[Feature]`, `[Chore]`, …) `validate-issue-structure.sh` accepts before blocking `gh issue create` |

atlas-apex:
Fixed by #961 (merged). `Migration` and `Idea` are now in `ticket.prefix_whitelist`, so `/migration`'s own template title validates out of the box. An audit of every `templates/tickets/*.md` prefix confirmed those were the only two gaps — `[Idea]` (from `/idea`'s optional tracking-issue path) was a second instance of the same bug your report surfaced.

[...]

PR #961, opened 2026-07-21
fix(#958): whitelist [Migration] and [Idea] ticket prefixes

atlas-apex:
## Summary

- **`/migration` couldn't create its own ticket** — `validate-issue-structure.sh` blocks any title whose bracketed prefix isn't in `ticket.prefix_whitelist`, but the shipped default omitted `Migration` while `templates/tickets/migration.md` emits `[Migration] {{type}}: {{summary}}`. Every fresh adopter running `/migration` hit this cold: the skill's own generated title was rejected by the framework's own validator. Reported in #958.
- **Fixed the whole class, not just the reported prefix.** I audited every `templates/tickets/*.md` prefix against the whitelist and found a **second** gap the report didn't catch: `[Idea]`. `/idea` optionally creates a tracking issue, and its template emits `[Idea] {{title}}` — so it hit the identical wall. Both `Migration` and `Idea` are now whitelisted; every shipped ticket template's prefix validates.
- **Deliberately did not add `required_sections` entries for the two new prefixes.** The validator already handles a missing entry gracefully (`.required_sections["<prefix>"][]? // empty` → no section enforcement), so whitelisting alone is sufficient to unblock creation. Adding required sections would introduce *new* blocking enforcement that the reported bug never asked for and that could reject existing tickets — that's a separate judgement call worth its own ticket if the maintainer wants it.

## Testing

1. `bash .claude/hooks/tests/test_validate_issue_structure.sh` → **34 passed, 0 failed**
2. Audited every `templates/tickets/*.md` bracketed prefix against `prefix_whitelist` — no remaining gaps
3. Verified `.claude/project-config.defaults.json` still parses as valid JSON

Refs #958

Reported by @aelnemr with the correct suggested fix — thank you.

_Note: this file gates issue creation, so a Security Auditor glance is welcome, though it is not under `.claude/hooks/**` and so doesn't auto-trigger the #777 trust-chain rule._

## Glossary

| Term | Definition |
|------|------------|
| `prefix_whitelist` | The set of bracketed ticket-title prefixes (`[Feature]`, `[Bug]`, …) that `validate-issue-structure.sh` accepts before allowing issue creation. |
| `required_sections` | Per-prefix map of `## ` headings a ticket body must contain; a prefix with no entry gets no section enforcement. |

### DIFF
--- a/.claude/project-config.defaults.json
+++ b/.claude/project-config.defaults.json
@@ -13,7 +13,9 @@
       "Docs",
       "Spike",
       "Prototype",
-      "Investigation"
+      "Investigation",
+      "Migration",
+      "Idea"
     ],
     "label_priority_scheme": "P0,P1,P2,P3",
     "required_sections": {


atlas-apex (commented):
## Code Review: PR #961

**Commit**: `a5893a3bc3e40421735c8f0f5f7f9d31d75963ab`

### Summary

Adds `Migration` and `Idea` to `ticket.prefix_whitelist` in `.claude/project-config.defaults.json`, so `validate-issue-structure.sh` stops rejecting the titles that `/migration` and `/idea` generate themselves. Three added lines, one file, no logic change.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (34/34 hook tests; I also ran the hook end-to-end, below)
- Security: **Pass**
- Performance: **Pass** (N/A)
- Summary Bullet Narrative: **Pass** — each bullet gives what changed *and* why
- Technical Decisions (AgDR): **N/A** — restores intended behaviour; the "no required_sections" scope call is disclosed in the PR body
- Adopter Handbooks: **N/A** — the blocking `migration-safety` handbook targets schema/migration files; this is a JSON config change

### Verification I ran independently

**1. The "no `required_sections`" call is correct — verified in code and empirically.**

The validator degrades exactly as claimed. `validate-issue-structure.sh:315` reads `.required_sections["${CANONICAL_PREFIX}"][]? // empty`, and the block right after it is unambiguous:

```sh
if [ -z "$REQUIRED_SECTIONS" ]; then
  # No schema for this prefix — hook has nothing to enforce, silent pass.
  exit 0
fi
```

I confirmed behaviourally by running the real hook against the PR's config with a deliberately empty body:

| Title | exit |
|---|---|
| `[Migration] schema: add users.role` | 0 (unblocked) |
| `[Idea] dark mode for dashboard` | 0 (unblocked) |
| `[Bogus] nope` | 2 (still blocked) |

So whitelisting alone unblocks creation, adds zero enforcement, and does not weaken the gate generally. Right scope for a bug fix.

**2. Completeness audit — independently confirmed, Migration and Idea were the only gaps.**

I audited `templates/tickets/*.md` myself rather than trusting the PR body. Emitted prefixes: `[Feature]`, `[Idea]`, `[Investigation]`, `[Migration]`, `[Prototype]`, `[Spike]`. `bug.md` / `task.md` carry no bracket prefix in-template (`/task` derives its prefix from the whitelist at runtime). All other `[Task]`-looking hits across skills are BPMN notation in `/process` or prose ("file a [Task] ticket"), not emitted titles. Audit checks out.

**3. `[Idea]` is in scope, not scope creep.** `.claude/skills/idea/SKILL.md:183-184` literally runs `gh issue create --title "[Idea] {title}"`. That the path is optional only changes *how often* it fires, not whether it hits the identical wall when it does. Fixing the class rather than the reported instance is the right call, and the PR discloses it explicitly rather than smuggling it in.

**4. JSON + formatting.** `jq empty` passes on the head-state file; 2-space indent preserved; entries appended after `Investigation`, consistent with the existing grouping.

**5. CI green** — hook test suite pass, Analyze pass, Verify Ticket ID pass, CodeQL skipping.

### Issues Found

None blocking.

### Suggestions

**suggestion: `Migration` having no `required_sections` is the one gap worth a follow-up ticket.** Not a blocker and not something to fix in this PR — I agree with the decision to keep this fix minimal, since adding enforcement here could reject in-flight tickets. But it's worth naming: these two become the only 2 of 12 whitelisted prefixes with no schema, and `Migration` is the highest-blast-radius ticket class in the framework (Gate 3a already demands a `migration` label plus an AgDR reference). A migration ticket body is precisely where you'd most want `Rollback Plan` / `Affected Tables` enforced, and `/migration` already collects those fields — so the entry would be nearly free. `Idea` is genuinely fine unenforced; it's a lightweight capture path.

**nit: the degraded-path inline fallback has drifted, and this fix doesn't reach it.** `validate-issue-structure.sh:190-193`:

```sh
# Inline defaults for bare checkouts predating apexyard#109. Mirror the
# shipped .claude/project-config.defaults.json so behaviour is identical.
if [ -z "$PREFIX_WHITELIST" ]; then
  PREFIX_WHITELIST="Feature Bug Chore Refactor Testing CI Docs Spike"
fi
```

The comment promises it mirrors the shipped defaults, but it was already missing `Prototype` and `Investigation`, and now also misses `Migration` and `Idea`. On any checkout where `_lib-read-config.sh` is absent, `/migration` stays broken. Pre-existing drift, not introduced here, and the path is narrow (the lib ships with the framework) — so non-blocking. But this PR widens the gap between comment and reality by two more entries, so a one-line sync is cheap insurance.

**nit: adjacent latent instance of the same bug class.** `.claude/skills/task/SKILL.md:139` documents the prefix as `[{Chore|Refactor|Test|CI}]` — `Test`, where the whitelist has `Testing`. If a run ever emits `[Test]` it gets blocked exactly like #958. Out of scope here; flagging since this PR claims a class-wide audit.

### Verdict

**APPROVED**

The fix is minimal, correct, and does the harder thing the report didn't ask for: it audits the class instead of patching the instance. Verified the graceful-degradation claim rather than taking it on faith. The three notes above are all follow-up material, none blocking.

---
Reviewed commit: `a5893a3bc3e40421735c8f0f5f7f9d31d75963ab`

atlas-apex (commented):
## Security Review: PR #961

**Commit**: `a5893a3bc3e40421735c8f0f5f7f9d31d75963ab`

### Summary

Adds `"Migration"` and `"Idea"` to `.ticket.prefix_whitelist` in `.claude/project-config.defaults.json`. This is an allow-list widening on config that gates `gh issue create`, so I reviewed it for privilege gain and enforcement-hole risk rather than just diff shape. **No blocking security issue.**

### Checklist Results

- Secrets & Credentials:  Pass (no secrets; config data only)
- Injection Prevention:   Pass (data-only change, no new command/regex construction)
- XSS Prevention:         N/A
- Auth & Authorisation:   Pass (see allow-list analysis below)
- API Security:           N/A

### Allow-list widening — do the new prefixes gain privilege?

**No. They gain exactly one thing: "this title shape is allowed."** Verified structurally and behaviourally.

`prefix_whitelist` has only two consumers in `validate-issue-structure.sh`: the title-shape whitelist check, and keying into `required_sections`. No other hook reads it (the remaining hits are skills reading it for their own UX, plus tests). Specifically confirmed **no** bypass of adjacent controls:

| Control | Depends on `prefix_whitelist`? | Behaviour with `[Migration]` / `[Idea]` |
|---|---|---|
| `require-skill-for-issue-create.sh` (skill-marker requirement) | No | Still `exit 2` — raw create blocked identically to `[Feature]` |
| `require-migration-ticket.sh` (migration gate) | No | Unaffected — keys on label + AgDR body regex |
| `required_sections` enforcement for existing prefixes | — | Intact (`[Feature]`, `[Bug]`, `[Spike]` all still `exit 2`) |
| Whitelist closure | — | Still closed: `[Wildcard]` → `exit 2` |

Exact-match semantics confirmed: the near-miss `[Migrations]` is still blocked, so no substring/prefix-matching weakness was introduced.

### The omitted `required_sections` — permissive, but not an enforcement hole

The PR deliberately adds no `required_sections` entries, so both new prefixes hit the hook's `prefix has no schema entry → exit 0 silently` path. **This is permissive-but-safe, not a skipped security control**, for a concrete reason:

The security-relevant migration artifacts — rollback plan, downtime, cross-service consumers, observability, captured in the paired migration AgDR — are **not** enforced through `required_sections` and never were (there was no `Migration` key before this PR either). They are enforced independently by `require-migration-ticket.sh`, which gates on:

1. ticket OPEN, **2.** ticket carries the `migration` label, **3.** body matches `docs/agdr/AgDR-\d+-.*migration.*\.md`

That gate keys on the **label**, not the title prefix, so it is fully orthogonal to this change. A `[Migration]`-titled ticket without the label still fails it; a `[Feature]`-titled ticket with the label still passes it. **Migration gate verified unaffected.**

### JSON integrity

Validated against the PR HEAD blob (not the local tree): parses clean, **no duplicate keys at any level**, 12 whitelist entries. Structural diff against the merge-base with `.ticket.prefix_whitelist` deleted from both sides is **identical** — confirming no other key was altered. The delta is exactly two appended strings.

### CI

Green — `Analyze (actions)`, `Verify Ticket ID`, and the `hook test suite` (2m22s) all pass. CodeQL `skipping` is expected for a JSON-only diff.

### Non-blocking findings

**LOW / informational — inline fallback drift (pre-existing, not introduced here).** `validate-issue-structure.sh:192` carries a hardcoded fallback for bare checkouts missing `_lib-read-config.sh`:

```
PREFIX_WHITELIST="Feature Bug Chore Refactor Testing CI Docs Spike"
```

It already omits `Prototype` and `Investigation`, and now also `Migration` and `Idea`. On such a checkout, `[Migration]` would be **blocked**. That is fail-**closed** — the safe direction, no security risk — and the drift predates this PR. Worth a follow-up to resync the fallback, but it should not hold this PR.

**INFO — these are the first two whitelisted prefixes with no section schema.** All 10 pre-existing entries have a `required_sections` entry; `Migration` and `Idea` will be the only two with zero body-shape enforcement. Reasonable here, since `/migration` and `/idea` produce their own bodies and the real migration control is the label+AgDR gate. Flagging it so the asymmetry is a deliberate choice rather than an oversight; add entries later if body shape is wanted.

### Verdict

**PASS — APPROVED (no blocking security issue)**

Allow-list widening is correctly scoped: the two new prefixes gain no privilege beyond title-shape acceptance, no adjacent control is bypassed, and the migration gate's AgDR/rollback enforcement is independent and intact.

---
Reviewed commit: `a5893a3bc3e40421735c8f0f5f7f9d31d75963ab`

me2resh (approved):

Issue #964, opened 2026-07-21
[Bug] prefix-whitelist drift: inline fallback + task/SKILL.md out of sync with shipped defaults


atlas-apex:
## Affected

`.claude/hooks/validate-issue-structure.sh` (the inline fallback whitelist, ~line 192) and `.claude/skills/task/SKILL.md` (~line 139).

Both surfaced during review of #961 (the #958 fix) — flagged by Rex and independently confirmed by Hakim.

## Given / When / Then

Two instances of the same root problem: the accepted-prefix list is expressed in **three** places with nothing keeping them in sync.

### (1) The inline fallback doesn't mirror the shipped defaults — despite its comment saying it does

**Given** `validate-issue-structure.sh` falls back to a hardcoded prefix list when `_lib-read-config.sh` isn't available, and the accompanying comment states it "mirrors the shipped defaults"
**When** a checkout without that lib runs `/migration` (or `/idea`, `/prototype`, `/investigation`)
**Then** the fallback is `PREFIX_WHITELIST="Feature Bug Chore Refactor Testing CI Docs Spike"` — missing `Prototype` and `Investigation` (already drifted *before* #961) and now also `Migration` and `Idea` — so those skills stay blocked on that path. **Expected:** the fallback matches `.claude/project-config.defaults.json` → `ticket.prefix_whitelist` (12 entries), as its comment claims.

Note this fails **closed** (it blocks rather than wrongly allows), so it is the safe direction — but the path is broken and the comment is now materially wrong.

### (2) `[Test]` vs `Testing` mismatch in the task skill's own docs

**Given** `.claude/skills/task/SKILL.md:139` documents the accepted set as `[{Chore|Refactor|Test|CI}]`
**When** an operator follows that doc verbatim and files `[Test] …`
**Then** `validate-issue-structure.sh` rejects it — the whitelist entry is `Testing`, not `Test`.

## Repro

1. **(1)** On a checkout where `_lib-read-config.sh` is unavailable, attempt `gh issue create --title "[Migration] schema: …"` → blocked by the fallback list even though the shipped defaults whitelist it.
2. **(2)** Follow `task/SKILL.md`'s documented prefix set and create a `[Test] …` ticket → blocked.

## Framework version

v5.1.0 (observed post-#961)

## Severity

[...]

PR #968, opened 2026-07-22
fix(#964): resync prefix-whitelist fallback + task skill to shipped defaults

atlas-apex:
## Summary
- **Resynced the config-less fallback whitelist** — `validate-issue-structure.sh` carries an inline `PREFIX_WHITELIST` for bare checkouts that predate project config; it had silently drifted to 8 entries, missing `Prototype`, `Investigation`, `Migration`, and `Idea`. On such a checkout a perfectly valid `[Migration]` or `[Idea]` ticket was rejected as an unknown prefix. Now mirrors the full 12-entry `.ticket.prefix_whitelist`.
- **Fixed the task skill's own example** — `task/SKILL.md` line 139 rendered the prefix menu as `{Chore|Refactor|Test|CI}`, but the whitelist entry is `Testing`, not `Test`. The skill was demonstrating a prefix its own validator rejects. Corrected to `Testing`.
- **Added a drift-guard test** — a static assertion that the hook's inline fallback string equals `.ticket.prefix_whitelist` in `project-config.defaults.json`. The next time the two lists diverge, CI fails loudly instead of leaving the drift to be discovered by a mis-rejected ticket (which is exactly how this bug surfaced).

## Testing
1. `bash .claude/hooks/tests/test_validate_issue_structure.sh` — 35/35 pass, including the new `inline fallback whitelist matches .ticket.prefix_whitelist` case.
2. `shellcheck .claude/hooks/validate-issue-structure.sh .claude/hooks/tests/test_validate_issue_structure.sh` — clean on the changed lines (two pre-existing style nits on untouched lines 62/88 remain).

Refs #964

---

## Glossary
| Term | Definition |
|------|------------|
| Prefix whitelist | The set of accepted `[Prefix]` ticket-title tokens (`Feature`, `Bug`, `Migration`, …) that `validate-issue-structure.sh` allows when a ticket is filed. |
| Inline fallback | A hard-coded copy of the whitelist the hook uses when it can't read `project-config.defaults.json` (a bare checkout with no config). |
| Drift | Two copies of the same list diverging over time because one was updated and the hand-copied mirror wasn't. |
| Drift guard | A test that asserts the two copies are identical, so a future divergence fails CI instead of silently mis-behaving. |

### DIFF
--- a/.claude/hooks/tests/test_validate_issue_structure.sh
+++ b/.claude/hooks/tests/test_validate_issue_structure.sh
@@ -476,6 +476,36 @@ run_case "Framework-shaped [Feature], no active-issue-skill marker → still blo
 rm -f "$TMPDIR/fork/.apexyard-fork"
 
 # ---------------------------------------------------------------------------
+# Drift guard (me2resh/apexyard#964)
+#
+# The hook carries an INLINE fallback whitelist for bare checkouts that predate
+# config (`PREFIX_WHITELIST="Feature Bug ..."`). That fallback is a hand-copied
+# mirror of `.ticket.prefix_whitelist` in project-config.defaults.json and had
+# silently drifted (missing Prototype/Investigation/Migration/Idea) — so a
+# `[Migration]`/`[Idea]` ticket filed on a config-less checkout was wrongly
+# rejected. This assertion fails loudly the next time the two lists diverge,
+# instead of leaving the drift to be discovered by a mis-rejected ticket.
+# ---------------------------------------------------------------------------
+DEFAULTS_JSON="$REPO_ROOT/.claude/project-config.defaults.json"
+# The hook's inline fallback: grep the assignment, strip to the quoted value,
+# then normalise whitespace → newline-per-token → sort.
+fallback_sorted=$(
+  grep -E 'PREFIX_WHITELIST="[^"]+"' "$HOOK" \
+    | sed -E 's/.*PREFIX_WHITELIST="([^"]+)".*/\1/' \
+    | tr ' ' '\n' | grep -v '^$' | sort
+)
+defaults_sorted=$(jq -r '.ticket.prefix_whitelist[]' "$DEFAULTS_JSON" | sort)
+if [ "$fallback_sorted" = "$defaults_sorted" ]; then
+  echo "PASS: inline fallback whitelist matches .ticket.prefix_whitelist in defaults"
+  PASS=$((PASS + 1))
+else
+  echo "FAIL: inline fallback whitelist has drifted from .ticket.prefix_whitelist" >&2
+  echo "  hook fallback : $(echo "$fallback_sorted" | tr '\n' ' ')" >&2
+  echo "  defaults json : $(echo "$defaults_sorted" | tr '\n' ' ')" >&2
+  FAIL=$((FAIL + 1))
+fi
+
+# ---------------------------------------------------------------------------
 # Result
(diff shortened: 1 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #968

**Commit**: `836b96f7dbe5c313111a3bddb5a369445b734e6e`

### Summary
Bug fix for #964 (prefix-whitelist drift). Three changes: (1) the config-less inline `PREFIX_WHITELIST` fallback in `validate-issue-structure.sh` is resynced from 8 to the full 12 entries; (2) `task/SKILL.md` line 139 `Test` → `Testing` so the skill's example prefix is one its own validator accepts; (3) a static drift-guard test asserting the hook's inline fallback equals `.ticket.prefix_whitelist` in the defaults JSON.

### Checklist Results
- Testing:                    Pass (35/35, incl. new drift-guard case)
- Security:                   Pass (no bypass surface — see trust-chain note)
- Technical Decisions (AgDR): N/A (constant resync + test; no library/framework/pattern decision)
- Adopter Handbooks:          Pass (architecture/migration-safety/commit-quality loaded; none apply to a shell-hook constant resync)
- Fallow Static Analysis:     N/A (diff is .sh/.md, not JS/TS)

### Verification performed
1. **Fallback matches defaults exactly.** Verified against the PR's real base — upstream `me2resh/apexyard:dev` (tip `c609c15`), which the branch is based directly on. Both sets are the identical 12 entries `Feature Bug Chore Refactor Testing CI Docs Spike Prototype Investigation Migration Idea` (order-independent set comparison: MATCH). (Note: a *local* stale `dev` at `da262ab` still shows 10 entries, but that is behind upstream and not the merge base — no action needed.)
2. **Drift-guard test is correct and load-bearing.** The `grep -E 'PREFIX_WHITELIST="[^"]+"'` isolates **exactly one** line (line 192, the inline fallback) — the empty init at 175 and the `config_get` assignment at 184 don't match the `"[^"]+"` pattern. The `sed`/`tr`/`sort` pipeline tokenises correctly and compares against `jq -r '.ticket.prefix_whitelist[]' | sort`. It genuinely fails on divergence (prints FAIL, increments the FAIL counter). Confirmed it currently passes.
3. **No residual `Test` vs `Testing` drift.** `task/SKILL.md:139` is the only prefix menu of that shape in `.claude/skills`/`.claude/hooks`, and it's fixed. No `[Test]` literal prefix usage anywhere in `.claude`/`templates`/`docs`.
4. **Test suite: 35/35 pass** from the PR HEAD worktree.
5. **Trust-chain bar (this edits a hook).** No new bypass surface: the fallback change only *broadens* acceptance to four legitimately-valid prefixes (`Prototype`, `Investigation`, `Migration`, `Idea`) that were being wrongly rejected on config-less checkouts — it removes nothing and changes no body-structure validation branch. No behavioural regression for the existing 8 prefixes. shellcheck clean on the changed lines (192 in the hook; the new test block).

### Issues Found
None.

### Suggestions
- (nit, non-blocking) The drift-guard `grep` assumes a single matching `PREFIX_WHITELIST="…"` line. That holds today and is the correct assumption; if a future edit ever adds a second quoted literal assignment, the guard would over-include and fail loudly (safe direction), so no change needed — just noting the invariant the test depends on.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `836b96f7dbe5c313111a3bddb5a369445b734e6e`

me2resh (approved):

Issue #965, opened 2026-07-22
[Bug] merge gate fails OPEN when jq is unavailable (block-unreviewed-merge.sh exits 0)


atlas-apex:
## Affected

`.claude/hooks/block-unreviewed-merge.sh` (the core merge gate). Surfaced during the security-grade code review of #963; verified byte-identical in the merge base, so it is **pre-existing**, not introduced by that PR.

## Given / When / Then

**Given** `block-unreviewed-merge.sh` is the PreToolUse gate that must refuse a `gh pr merge` until the required approval markers exist
**When** `jq` is not available on `PATH` (the hook `jq`-parses `.tool_input.command` at ~line 50, *before* any gate logic runs)
**Then** the parse fails, the hook **exits 0**, and the merge command proceeds **ungated** — any PR could be merged with no Rex/CEO approval markers at all. **Expected:** a security gate must **fail closed** — if it cannot evaluate its own precondition (missing `jq`, unparseable input), it must block, not allow.

The failure direction is backwards for a gate: an environment defect (missing dependency) currently *disables* the control rather than tripping it.

## Repro

1. Temporarily remove `jq` from `PATH` (or point the hook's `jq` at a stub that exits non-zero).
2. Attempt `gh pr merge <N> --squash` on a PR with **no** approval markers.
3. The merge is **allowed** — the gate never reaches its marker check.

(Confirmed empirically during the #963 review at the reviewed commit and against the merge base — identical behaviour, so this predates #963.)

## Framework version

v5.1.0

## Severity

**major** — it is a fail-open in the primary merge gate. Bounded in practice by the fact that **`jq` is a hard framework dependency**: `check-jq-installed.sh` + a SessionStart banner surface a missing `jq`, and `/setup` refuses to run without it (see AgDR-0038). So a correctly-set-up fork always has `jq`, which is why this hasn't bitten anyone. But "the security gate is only sound because a *different* advisory happens to nag about the dependency" is not the posture a merge gate should have — a broken/renamed/PATH-shadowed `jq` mid-session would silently open the gate.

## Notes

Root issue: the gate treats "can't parse my input" as "nothing to gate → allow" instead of "can't evaluate → deny." Suggested direction:

1. If `jq` is unavailable *or* the input can't be parsed, **exit 2 (block)** with a clear message ("merge gate cannot evaluate: jq unavailable / unparseable tool input — refusing to merge"), not exit 0.
[...]

PR #969, opened 2026-07-22
fix(#965): merge-gate hooks fail closed when jq/tool-input is unevaluable

atlas-apex:
## Summary

- **The merge gate silently disabled itself when `jq` was broken.** `block-unreviewed-merge.sh` (and its three sibling merge gates) parsed `.tool_input.command` via `jq` as the very first step, and an empty/failed parse fell straight through to `exit 0` — "nothing to gate, let it through." If `jq` was missing from `PATH` (or errored for any reason), a `gh pr merge` with **zero approval markers** would sail through completely ungated. A security gate that fails *open* on its own precondition breaking is the opposite of what it's for.
- **Audited and fixed all four merge-gate hooks identically** — `block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh` all shared the exact same "parse-first, exit-0-on-failure" shape (verified byte-for-byte at the top of each file), so the same fix applies to all four, not just the one named in #965.
- **Fixed with a boundary that matters**: each of these hooks' `PreToolUse` matcher is `Bash` — every Bash command the session runs, not just merges. A blanket "jq unavailable → exit 2" would block every unrelated command (`npm test`, `ls`, anything) the instant `jq` broke, which is worse than the bug it replaces. Instead, when the jq parse yields nothing, each hook now falls back to running `is_merge_command` — the same tested, jq-free grep/sed detector already used on the parsed command — directly against the raw JSON payload text. A payload that isn't merge-shaped stays a no-op (unchanged behaviour for the overwhelming majority of Bash calls); a payload that *does* look merge-shaped but can't be safely parsed now fails closed instead of silently allowing.
- **Regression tests added** to `test_block_unreviewed_merge.sh` and `test_block_merge_on_red_ci.sh`: with `jq` stubbed to always fail, a `gh pr merge`/`gh api .../merge` with no approval markers now blocks (exit 2), while a clearly non-merge command (`npm test`) stays a no-op (exit 0, no stderr) — proving the fix doesn't over-block. The other two hooks (`require-design-review-for-ui.sh`, `require-architecture-review.sh`) were verified manually with the same jq-broken scenarios (both block on merge-shaped input, no-op on non-merge input) — their existing test harnesses use a different structure (portfolio/split-portfolio helpers) and adding equivalent automated cases there is flagged as good follow-up work, not done in this PR to keep the diff auditable.

## Testing

1. `shellcheck -x` on all six touched files — zero new findings; the only findings present are pre-existing `info`-level notices (`SC1091` on relative `.` sourcing, `SC2086`/`SC2016`/`SC2059` on pre-existing lines elsewhere in the files) that were already there before this change and are below the CI gate's `severity: warning` threshold.
2. `bash .claude/hooks/tests/test_block_unreviewed_merge.sh` → 37/37 pass (3 new `#965` cases).
3. `bash .claude/hooks/tests/test_block_merge_on_red_ci.sh` → 25/25 pass (2 new `#965` cases).
4. `bash .claude/hooks/tests/test_require_design_review_for_ui.sh` → 22/22 pass (unchanged, confirms no regression).
5. `bash .claude/hooks/tests/test_require_architecture_review.sh` → 22/22 pass (unchanged, confirms no regression).
6. Manual sandbox verification: with `jq` stubbed to fail, `require-design-review-for-ui.sh` and `require-architecture-review.sh` both block (exit 2, "cannot evaluate this command") on a merge-shaped command and stay a no-op (exit 0) on `npm test`.
7. Full suite: `bin/run-hook-tests.sh` → 112/113 pass. The one failure (`test_sync_codex_adapter.sh`, a `sed` internal assertion crash) is unrelated — that file isn't touched by this PR, and re-running it in isolation shows a different, pre-existing flake, not a crash tied to this change.

Refs #965

---

## Glossary

| Term | Definition |
|------|------------|
| Fail closed / fail open | A gate that, on ambiguity or internal error, *blocks* (closed — safe) vs. *allows* (open — unsafe). Security/quality gates must fail closed. |
| Merge gate | One of the four `PreToolUse` hooks (`block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh`) that inspect a `gh pr merge` / `gh api .../merge` command before letting it run. |
| `is_merge_command` | A shared, jq-free grep/sed detector in `_lib-extract-pr.sh` that recognises every merge-command shape (`gh pr merge`, `gh api .../pulls/<N>/merge`, `glab mr merge`, `glab api .../merge_requests/<N>/merge`, the `tracker_pr_merge` wrapper). This fix reuses it as a fallback against the raw JSON payload when the jq-parsed command is unavailable. |
| `.tool_input.command` | The JSON field each hook parses to read the Bash command it's gating, taken from the `PreToolUse` payload the harness pipes to the hook on stdin. |

### DIFF
--- a/.claude/hooks/block-merge-on-red-ci.sh
+++ b/.claude/hooks/block-merge-on-red-ci.sh
@@ -45,16 +45,58 @@
 #   treated as green, exactly like a red-or-unfetchable gh CI check.
 
 INPUT=$(cat)
-COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
-
-if [ -z "$COMMAND" ]; then
-  exit 0
-fi
 
 # Shared merge-shape detector + PR-number parser (see _lib-extract-pr.sh).
 # Handles `gh pr merge <N>`, `gh api repos/<owner>/<repo>/pulls/<N>/merge`,
 # `glab mr merge <N>`, and `glab api .../merge_requests/<N>/merge` (#764/#767).
+# Sourced BEFORE the jq-based command parse below (moved up from its
+# original position after the parse) so is_merge_command is available as
+# the jq-independent fallback detector when the parse can't be trusted —
+# see #965.
 . "$(dirname "$0")/_lib-extract-pr.sh"
+
+# Parse .tool_input.command via jq. #965: this used to be the ONLY parse
+# path, and an empty/failed result — jq missing from PATH, or jq erroring
+# on unexpected input — fell straight through to `exit 0`, silently
+# ALLOWING the merge command through with NO CI-status check at all. A
+# security/quality gate must fail CLOSED when it can't evaluate its own
+# precondition, not fail open.
+#
+# But this hook's PreToolUse matcher is `Bash` (every Bash call this
+# session runs, not just merges — see .claude/settings.json), so the fix
+# can't be "exit 2 whenever jq is unavailable": that would block every
+# unrelated Bash command for the rest of the session the moment jq broke,
+# which is worse than the bug it replaces. The resolution below keeps the
+# jq-unparseable case a no-op EXCEPT when the raw payload text itself
+# looks merge-shaped — in that narrower case we cannot safely let the
+# command through, so we fail closed instead.
+COMMAND=""
+if command -v jq >/dev/null 2>&1; then
+  COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
+fi
+
+if [ -z "$COMMAND" ]; then
+  # jq is missing, OR jq is present but the parse produced nothing — a
+  # genuinely empty command (legitimate no-op) or jq choking on
+  # malformed/unexpected JSON. Those two cases are indistinguishable from
+  # a parsed field alone, so fall back to a parser-independent scan: reuse
+  # is_merge_command (plain grep/sed, no jq dependency) directly against
+  # the RAW JSON payload text instead of the parsed command. The command
+  # text's own words (`gh`, `pr`, `merge`, digits, spaces) survive JSON
+  # string-encoding unchanged, so this is the exact same tested
+  # merge-shape detector used below — not a second, drift-prone regex.
+  #
+  # A payload that isn't merge-shaped at all is a genuine no-op — exit 0,
+  # unchanged behaviour for the overwhelming majority of Bash calls this
+  # hook sees. A payload that DOES look merge-shaped but that we can't
+  # safely parse/verify fails CLOSED (exit 2) instead of silently letting
+  # an ungated merge through with an unverified CI status.
+  if is_merge_command "$INPUT"; then
+    echo "BLOCKED: CI gate cannot evaluate this command — jq is unavailable or .tool_input.command could not be parsed, but the raw input looks merge-related. Refusing to merge until CI status can be verified. Restore jq (see .claude/hooks/check-jq-installed.sh) and retry." >&2
+    exit 2
+  fi
+  exit 0
+fi
 
 if ! is_merge_command "$COMMAND"; then
   exit 0
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -47,17 +47,59 @@
 # explicit human-authorization moment. Different threat model.
 
 INPUT=$(cat)
-COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
-
-if [ -z "$COMMAND" ]; then
-  exit 0
-fi
 
 # Shared merge-shape detector + PR-number parser (see _lib-extract-pr.sh).
 # Handles `gh pr merge <N>` and `gh api repos/<owner>/<repo>/pulls/<N>/merge`.
+# Sourced BEFORE the jq-based command parse below (moved up from its
+# original position after the parse) so is_merge_command is available as
+# the jq-independent fallback detector when the parse can't be trusted —
+# see #965.
 . "$(dirname "$0")/_lib-extract-pr.sh"
 # Repo-qualified marker path helper (#485).
 . "$(dirname "$0")/_lib-review-markers.sh"
+
+# Parse .tool_input.command via jq. #965: this used to be the ONLY parse
+# path, and an empty/failed result — jq missing from PATH, or jq erroring
+# on unexpected input — fell straight through to `exit 0`, silently
+# ALLOWING the merge command through completely ungated. A security gate
+# must fail CLOSED when it can't evaluate its own precondition, not fail
+# open.
+#
+# But this hook's PreToolUse matcher is `Bash` (every Bash call this
+# session runs, not just merges — see .claude/settings.json), so the fix
+# can't be "exit 2 whenever jq is unavailable": that would block every
+# unrelated Bash command for the rest of the session the moment jq broke,
+# which is worse than the bug it replaces. The resolution below keeps the
+# jq-unparseable case a no-op EXCEPT when the raw payload text itself
+# looks merge-shaped — in that narrower case we cannot safely let the
+# command through, so we fail closed instead.
+COMMAND=""
+if command -v jq >/dev/null 2>&1; then
+  COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
+fi
+
+if [ -z "$COMMAND" ]; then
+  # jq is missing, OR jq is present but the parse produced nothing — a
+  # genuinely empty command (legitimate no-op) or jq choking on
+  # malformed/unexpected JSON. Those two cases are indistinguishable from
+  # a parsed field alone, so fall back to a parser-independent scan: reuse
+  # is_merge_command (plain grep/sed, no jq dependency) directly against
+  # the RAW JSON payload text instead of the parsed command. The command
+  # text's own words (`gh`, `pr`, `merge`, digits, spaces) survive JSON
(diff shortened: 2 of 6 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #969 — merge-gate hooks fail closed when jq/tool-input is unevaluable

**Commit**: `b23ced78c875cbeb830e0dea0a04bd18cde66c6b`

### Summary
Fixes #965: all four `PreToolUse` merge-gate hooks shared a byte-identical "parse `.tool_input.command` with jq first, `exit 0` if the parse yields nothing" shape — so a missing/erroring `jq` made every merge gate fail **open**, silently allowing an ungated `gh pr merge` with zero approval markers. The fix reuses the existing jq-free `is_merge_command` detector against the raw JSON payload when the jq parse yields nothing: a non-merge command stays a no-op (`exit 0`), a merge-shaped command fails **closed** (`exit 2`). Applied identically to all four hooks.

### Checklist Results
- Architecture & Design: **Pass** — correct fail-closed direction; narrow raw-text fallback avoids blanket over-blocking on the `Bash` matcher
- Code Quality: **Pass** — reuses the shared tested detector, no drift-prone second regex; thorough rationale comments
- Testing: **Pass (with a follow-up note)** — 2 of 4 hooks get new automated jq-broken cases; the other 2 verified manually (see below)
- Security: **Pass** — this IS the trust chain; the fix restores fail-closed on the gate's own precondition breaking
- Performance: **Pass** — one extra `command -v jq`; negligible
- PR Description & Glossary: **Pass** — narrative summary bullets (what + why), Glossary table present, `Refs #965`
- AgDR: **N/A** — bug fix reusing an existing tested helper; no new dependency/pattern
- Adopter Handbooks: **N/A** — none loaded

### Independent Verification
I checked out the PR HEAD in an isolated worktree and ran the fix against a `jq` stub that always fails (`exit 1`) — the exact code path as jq missing from PATH:

- **Fail-closed correctness** — with jq broken:
  - `require-design-review-for-ui.sh` and `require-architecture-review.sh` both **BLOCK (exit 2)** on `gh pr merge …` AND on `gh api repos/.../pulls/<N>/merge` — independently reproducing the agent's manual verification for the two hooks whose harnesses weren't updated.
  - Both **no-op (exit 0)** on `npm test`. The fix does not over-block unrelated Bash calls.
- **Both merge shapes covered** — the raw-text fallback recognises `gh pr merge`, the `gh api .../pulls/<N>/merge` API shape, and (verified in the shared lib) the `glab` and `tracker_pr_merge` wrapper shapes. Command keywords (`gh`, `pr`, `merge`, `api`, `repos`, digits, `/`) are never JSON-escaped, so they survive in the raw payload.
- **No new bypass** — the fallback reuses the *same* `is_merge_command` the normal path uses, run against the raw payload (a superset of the command text), so it can only match the same-or-more. Any evasion would already exist on the jq-present path; nothing new is opened.
- **No regression** — full test suites at PR HEAD: `test_block_unreviewed_merge` 37/37, `test_block_merge_on_red_ci` 25/25, `test_require_design_review_for_ui` 22/22, `test_require_architecture_review` 22/22.
- **Unrelated failure** — `test_sync_codex_adapter.sh` is confirmed **not** in this diff (6 files touched, none the codex adapter). Correctly out of scope; not held against this PR.
- **CI** — hook test suite pass, shellcheck pass, Analyze pass, Verify Ticket pass. (CodeQL "skipping" is normal for this repo.)

### Issues Found
None blocking.

### Suggestions (non-blocking follow-up)
- **suggestion:** Add automated jq-broken regression cases to `test_require_design_review_for_ui.sh` and `test_require_architecture_review.sh` so the fail-closed behaviour on those two trust-chain hooks can't silently regress in a future refactor. The code is byte-identical across all four and I reproduced the manual verification, so this is acceptable as a tracked follow-up rather than a blocker — but for merge-gate code, "identical code, verified manually" is a rot risk worth closing with a test.
- **nit (theoretical, non-blocking):** A merge command with a literal escaped-whitespace separator (e.g. a tab, JSON-encoded as `\t`) would be caught by the jq-present path but not the raw-text `\s+` fallback while jq is broken. Wildly contrived (requires jq broken *and* a tab-separated merge invocation) and fails in the safe direction on the normal path; not worth code changes.

### Verdict
**APPROVED**

The fix is correct, consistent across all four merge gates, independently verified to fail closed on both merge shapes, and introduces no regression or new bypass. CI is green.

---
📌 Reviewed commit: `b23ced78c875cbeb830e0dea0a04bd18cde66c6b`

atlas-apex (commented):
## Security Review: PR #969 — merge-gate fail-closed (#965)

**Commit reviewed**: `b23ced78c875cbeb830e0dea0a04bd18cde66c6b`
🛡️ Security lens only (Hakim). Rex already covered code-quality.

### Verdict: **PASS** (one non-blocking concern + hardening note)

The fix converts a genuine **fail-open** in the security enforcement layer into a fail-closed for every realistic case. I verified the fallback logic empirically against the actual `is_merge_command` in `_lib-extract-pr.sh` (the detector both the parsed and raw-text paths share).

### What I verified

**1. Fail-closed is genuinely closed for the realistic case.**
- jq broken + normal space-separated `gh pr merge 302 --squash` in the raw JSON payload → `is_merge_command "$INPUT"` **MATCHES → exit 2 (blocks)**. Confirmed by byte-level probe. A space (U+0020) is never escaped inside a JSON string, so the command's own words survive raw-text encoding unchanged and the tested detector fires. This is the whole ballgame and it works.
- gh-api shape (`gh api repos/.../pulls/<N>/merge -X PUT`, no literal `gh pr merge` substring) → **blocks** under jq-broken too (matches test case 16).
- Pre-#965 the same scenario fell through `[ -z "$COMMAND" ] && exit 0` → **every ungated merge sailed through**. The PR strictly improves the posture.

**2. Residual evasion — tab/escape-separated merge (NON-BLOCKING).** Rex's theoretical `\t` note is real, and I confirmed it (bytes inspected with `od -c`):
- Raw JSON with `gh\tpr\tmerge …` (tab escaped as the two-char `\t`, as JSON requires) → grep's `\s+` does **not** match a literal backslash-t → `is_merge_command` returns false → **exit 0 (allows)**.
- Same class: `\uXXXX`-escaped keywords, and a `\/`-escaped gh-api path, would also dodge the raw-text grep.

Why this does **not** block:
- It requires **two simultaneous preconditions**: (a) jq already broken/missing — an anomalous, self-announcing state the framework warns on via `check-jq-installed.sh`, under which the normal path (which decodes `\t` to a real tab) already catches the same command (verified: jq-decoded tab merge → MATCHES); **and** (b) a *deliberately* obfuscated command no normal agent or human emits (everyone types spaces).
- It is a **deliberate, targeted self-bypass**, which sits squarely beneath these gates' documented threat model ("For adversarial trust, use CODEOWNERS"; markers are a "visible, auditable, grep-able rule violation" backstop, not adversarial security). An actor who can craft this in a jq-broken shell already has strictly easier, always-present bypasses (`rm`/`chmod -x` the hook, forge a marker, or merge locally + push). The tab trick is harder than those and yields the same result — so it does not lower the bar below where the framework already sits by design.

**Hardening recommendation (defense-in-depth, not required for merge):** before running `is_merge_command "$INPUT"`, normalize the common JSON whitespace/escape sequences in the raw payload (`\t \n \r 	 \/`) to spaces — a one-line `sed`/`tr` pass would fully close the residual gap at trivial cost. Worth a follow-up ticket; not a blocker for this PR.

**3. Fallback threat model — false-positive over-block (NON-BLOCKING, fail-safe).** Running `is_merge_command` on the raw *payload* (a superset of the command) means a non-merge command that merely *contains* merge-shaped text in its arguments (`echo "gh pr merge 5"`, `grep "gh pr merge"`) will spuriously **block** while jq is broken. I hit this live — several of my own probe commands were blocked by the gate. This is an availability annoyance, **fail-closed** (safe direction), only occurs in the degraded jq-broken window, and is a **pre-existing** property of the gate's grep-based detection (the parsed path has it too) — merely extended to the raw payload here. The envelope keys (`tool_name`,`command`) are not merge-shaped, so no spurious blocks from the JSON wrapper itself. Acceptable.

**4. Other three gates preserve their invariants.** All four hooks received byte-identical fallback logic. Under jq-broken, the design-review and architecture-review gates cannot resolve which paths the PR touches (that needs the parsed command), so they **block any merge-shaped command regardless of UI/design scope** — strictly *more* conservative than their normal firing scope, and fail-closed. No gate can now behave inconsistently in the unsafe direction. Consistent and correct.

**5. No new injection surface, no secrets.** The raw `$INPUT` flows only into `is_merge_command "$INPUT"` → `echo "$cmd" | grep -qE '<single-quoted literal>'`. `$INPUT`/`$cmd` are always double-quoted; patterns are literal; there is no `eval`, no command substitution of payload text, no unquoted expansion into a command position. Error messages leak nothing. Moving the `. _lib-extract-pr.sh` sourcing above the empty-check only means functions/sibling libs are defined on every Bash call — no new risk.

### Checklist
- Secrets & Credentials: Pass
- Injection Prevention: Pass
- Auth & Authorisation (gate integrity): Pass — fail-closed for realistic merges; residual evasion beneath documented threat model
- Data Protection: Pass (n/a)
- Consistency across all four gates: Pass

### Recommendation
Approve. File a follow-up (non-blocking) to normalize JSON escape sequences in the raw-payload fallback so the tab/`\uXXXX`/`\/` evasion class is fully closed as defense-in-depth.

---
📌 Reviewed commit: `b23ced78c875cbeb830e0dea0a04bd18cde66c6b`

Issue #962, opened 2026-07-21
[Bug] warn-review-marker-write.sh misses variable-expanded marker writes (and blocks literal-path reads)


atlas-apex:
## Affected

`.claude/hooks/warn-review-marker-write.sh` (the #843 blocking gate that authorises `*-rex.approved` / `*-security.approved` / `*-architecture.approved` writes against the `active-reviewer` session marker).

## Given / When / Then

**Given** `warn-review-marker-write.sh` is a **blocking** PreToolUse gate (#843): a Write/Bash call targeting a review marker under `.claude/session/reviews/` must be accompanied by a matching `.claude/session/active-reviewer` marker, or the write is refused (exit 2)
**When** the write target is supplied via a **shell variable** rather than a literal path — e.g. `MARKER="$(review_marker_path …)"; printf '%s' "$SHA" > "$MARKER"`
**Then** the hook's write-detection does **not** resolve the variable, so the marker write proceeds **undetected and ungated** — expected: the gate evaluates the resolved target and applies the same authorisation check it applies to a literal path.

The inverse also misfires: referencing the **same** path *literally* in a harmless **read** (`cat "<literal>/…-security.approved"`) **does** trip the gate and gets blocked. So the current detection blocks a benign read while letting the actual privileged write through.

## Repro

1. Ensure **no** `.claude/session/active-reviewer` marker exists (so any marker write should be refused).
2. Attempt a marker write with a literal path — e.g. `printf '%s' "$SHA" > .claude/session/reviews/<owner>__<repo>__<pr>-security.approved` → **blocked**, as designed.
3. Attempt the same write through a variable: `MARKER=".claude/session/reviews/<owner>__<repo>__<pr>-security.approved"; printf '%s' "$SHA" > "$MARKER"` → **succeeds**, gate never fires.
4. Now `cat "$…"` the same path using the **literal** string → **blocked**, despite being a read.

Observed in practice during a `security-reviewer` run on a live PR: the marker write landed via the variable form while a later literal-path `cat` was refused. (The outcome was legitimate on the merits — it was the sanctioned reviewer agent and the marker content matched HEAD — but it landed through a detection gap, not through the gate's approval.)

## Framework version

v5.1.0

## Severity

[...]

PR #970, opened 2026-07-22
fix(#962): gate marker writes on the resolved target, not a literal path match

atlas-apex:
## Summary

- **Closes the false negative that let a build agent self-approve a review marker** — `warn-review-marker-write.sh` (the #843 blocking gate) matched the literal `-rex.approved`/`-security.approved`/`-architecture.approved` suffix as a plain substring of the Bash command text. The sanctioned reviewers' own documented idiom — `REX_MARKER=$(review_marker_path "$REPO" "$PR" rex "$MARKER_HOME"); printf '%s\n' "$SHA" > "$REX_MARKER"` — never puts that literal suffix anywhere in the command text, so a build agent copying that exact idiom evaded the gate entirely: no `active-reviewer` marker required, no block. This was exercised on a live PR, not hypothetical. The fix detects on the **resolved write**, reusing `_lib-detect-bash-write.sh` (the same helper `require-active-ticket.sh`'s Bash matcher already uses for the identical class of problem, #151) plus a plausibility check for the `review_marker_path`/`*_MARKER` indirection idiom.
- **Closes the paired false positive** — a harmless `cat`/`grep` read of a marker path that happens to spell out the literal filename used to trip the same substring match and get wrongly refused. A literal-path match is now gated only when the command also performs an actual write (`bash_command_appears_to_write`), so reads pass through silently.
- **Best-effort role/PR extraction from the indirection idiom** — `review_marker_path`'s role and PR-number arguments are conventionally literal in the reviewers' own documented usage even though the call's *result* is assigned to a variable, so the marker's kind (rex/security/architecture) and target PR can usually still be recovered and checked against the `active-reviewer` marker exactly as before.
- **Fails closed on total ambiguity, by construction, not by a special case** — when the role genuinely can't be resolved (a write clearly touching `.claude/session/reviews/` with no recognisable role marker), the gate still blocks: no well-formed `active-reviewer` marker can ever have an empty `kind` field, so the existing equality check already rejects it.
- **Graceful degrade, not a silent bypass, if the write-detection library is ever missing** — falls back to the pre-#962 literal-substring-only check (conservative: the old false-positive-on-read limitation re-applies, but nothing that was previously caught is weakened).

## Testing

- Added 6 new cases to `.claude/hooks/tests/test_warn_review_marker_write.sh` (cases 20–25) covering: a variable-indirected write with no `active-reviewer` marker (BLOCKED), the same write with a matching marker (ALLOWED), a literal-path `cat` read (NOT blocked — the false-positive fix), an indirected `tee` write (BLOCKED), a fully ambiguous indirected write even with a wrong-kind `active-reviewer` marker present (BLOCKED — fail-closed), and the graceful library-missing fallback (BLOCKED, conservative).
- Full suite run: `.claude/hooks/tests/test_warn_review_marker_write.sh` → **27/27 PASS** (19 pre-existing + 6 new).
- Full repo hook suite: `bin/run-hook-tests.sh` → **111/113 PASS**. The 2 failures (`test_pr_create_gate_anchor.sh`, `test_sync_codex_adapter.sh`) are pre-existing on `upstream/dev` and untouched by this diff — confirmed via `git diff --name-only upstream/dev`, which shows only the two files this PR touches.
- `shellcheck -x` on both touched files: clean (0 findings) after adding targeted `# shellcheck disable=SC2016` comments on three lines in the test file where `$VAR` must deliberately stay unexpanded literal text (that's the whole point of the indirection test cases). The hook file carries one pre-existing `SC2295` info-level finding, unchanged from `upstream/dev` — confirmed identical via a side-by-side shellcheck run against the upstream copy.

Refs #962

---

## Glossary

| Term | Definition |
|------|------------|
| Review marker | A gitignored local file at `.claude/session/reviews/<owner>__<repo>__<pr>-<role>.approved` that records a reviewer's (Rex/Hakim/Tariq) or the human approver's sign-off on a specific PR + commit SHA. Merge gates check these files exist and match HEAD before allowing a merge. |
| `active-reviewer` marker | A one-line session file (`<owner>/<repo>#<pr>:<kind>`) the orchestrator writes immediately before spawning a sanctioned reviewer agent. `warn-review-marker-write.sh` requires this marker to exist and match before it allows a `*-rex.approved`/`*-security.approved`/`*-architecture.approved` write — the mechanism that stops a build agent from writing its own approval. |
| Resolved-target detection | Determining what a shell command actually writes to by analysing its structure (redirection operators, `tee`, in-place editors, etc.) rather than searching the command's raw text for a literal substring. The fix in this PR switches the hook from the latter to the former. |
| Fail-closed | When the system can't determine whether an action is authorized, it defaults to blocking rather than allowing. This PR's ambiguous-role case (§ "fails closed on total ambiguity") is fail-closed: an unresolvable marker role can never match a real approval, so it's blocked by construction. |

### DIFF
--- a/.claude/hooks/tests/test_warn_review_marker_write.sh
+++ b/.claude/hooks/tests/test_warn_review_marker_write.sh
@@ -33,6 +33,24 @@
 #   (18) Write  → legacy bare-number marker filename, no marker  → BLOCKED, exit 2
 #   (19) Write  → legacy bare-number marker filename, matching pr+kind → ALLOWED, exit 0 (repo check skipped)
 #
+# #962 test matrix (resolved-target detection — see the hook's own #962
+# header comment for the false-negative/false-positive bugs this closes):
+#   (20) Bash   → variable-indirected write (review_marker_path + "$REX_MARKER"
+#                 redirect), no active-reviewer marker            → BLOCKED, exit 2
+#   (21) Bash   → same indirected write, WITH matching active-reviewer marker
+#                                                                  → ALLOWED, exit 0
+#   (22) Bash   → literal-path READ (cat) of a rex marker, no active-reviewer
+#                                                                  → silent, exit 0 (NOT blocked — false-positive fix)
+#   (23) Bash   → indirected tee write (review_marker_path + tee "$REX_MARKER"),
+#                 no active-reviewer marker                       → BLOCKED, exit 2
+#   (24) Bash   → fully ambiguous indirected write (mentions
+#                 .claude/session/reviews/ but no resolvable role), WITH an
+#                 active-reviewer marker present for a DIFFERENT kind
+#                                                                  → BLOCKED, exit 2 (fail-closed on total ambiguity)
+#   (25) Bash   → _lib-detect-bash-write.sh missing (graceful fallback):
+#                 literal-path READ (cat) of a rex marker, no active-reviewer
+#                                                                  → BLOCKED, exit 2 (conservative pre-#962 fallback, not a bypass)
+#
 # Exit 0 if all cases pass; 1 on failure.
 
 set -u
@@ -44,6 +62,7 @@ export APEXYARD_OPS_DISABLE_PIN=1
 HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/warn-review-marker-write.sh"
 LIB_OPS_ROOT="$(cd "$(dirname "$0")/.." && pwd)/_lib-ops-root.sh"
 LIB_MARKERS="$(cd "$(dirname "$0")/.." && pwd)/_lib-review-markers.sh"
+LIB_BDW="$(cd "$(dirname "$0")/.." && pwd)/_lib-detect-bash-write.sh"
 
 if [ ! -f "$HOOK_SRC" ]; then
   echo "FAIL: hook not found: $HOOK_SRC" >&2
@@ -59,6 +78,10 @@ if [ ! -f "$LIB_OPS_ROOT" ]; then
 fi
 if [ ! -f "$LIB_MARKERS" ]; then
   echo "FAIL: _lib-review-markers.sh not found at $LIB_MARKERS" >&2
+  exit 1
+fi
+if [ ! -f "$LIB_BDW" ]; then
+  echo "FAIL: _lib-detect-bash-write.sh not found at $LIB_BDW" >&2
   exit 1
 fi
 
@@ -88,7 +111,17 @@ make_sandbox() {
   mkdir -p "$sb/.claude/hooks" "$sb/.claude/session/reviews"
   cp "$HOOK_SRC" "$sb/.claude/hooks/warn-review-marker-write.sh"
   cp "$LIB_OPS_ROOT" "$sb/.claude/hooks/_lib-ops-root.sh"
+  cp "$LIB_BDW" "$sb/.claude/hooks/_lib-detect-bash-write.sh"
   chmod +x "$sb/.claude/hooks/warn-review-marker-write.sh"
+  echo "$sb"
+}
+
+# Build a sandbox with _lib-detect-bash-write.sh DELETED — used by case24 to
+# verify the graceful literal-substring-only fallback when the library is
+# missing (#962).
+make_sandbox_no_bdw_lib() {
+  local sb; sb=$(make_sandbox)
+  rm -f "$sb/.claude/hooks/_lib-detect-bash-write.sh"
   echo "$sb"
 }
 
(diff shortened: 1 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #970

**Commit**: `6250020f682c6a86b75c52dabb6b0f93484e21d9`

▸ Reviewed by Rex (Code Reviewer) with Hakim (Security Auditor) co-active — diff touches `.claude/hooks/**`, the merge-gate trust chain.

### Summary
Closes #962: `warn-review-marker-write.sh` (the #843 blocking gate that stops a build sub-agent forging its own `*-rex/*-security/*-architecture.approved` marker) matched the marker suffix as a **literal substring of the raw command**. That both (a) let the sanctioned reviewers' own `M=$(review_marker_path …); printf … > "$M"` idiom evade the gate entirely — the suffix never appears literally — and (b) wrongly blocked harmless literal-path **reads**. The fix reuses `_lib-detect-bash-write.sh` to gate a literal path only on an actual write (fixes the read false-positive), adds an indirection-plausibility check for the variable idiom (fixes the write false-negative), recovers role/PR from `review_marker_path`'s literal args, and falls back to the pre-#962 literal check when the lib is missing.

### Checklist Results
- ✅ Architecture & Design: Pass — reuses the same write-detector `require-active-ticket.sh` uses (#151); detection/extraction share one segmentation source.
- ✅ Code Quality: Pass — shellcheck `-x` clean bar one pre-existing info-level `SC2295` (line 358), unchanged from base.
- ✅ Testing: Pass — 27/27 on the PR head; 6 new cases (20–25) cover the false-negative, false-positive, tee variant, fail-closed, and lib-missing fallback.
- ✅ Security: Pass (with one hardening suggestion below) — no *new* exploitable bypass beyond the already-documented threat model.
- ✅ Performance: Pass — a few extra greps per matched command; negligible.
- ✅ PR Description & Glossary: Pass — narrative summary + glossary present.
- ✅ Technical Decisions (AgDR): Pass — references AgDR-0062/0104; behaviour-preserving hardening of an existing gate, no new decision.
- N/A Adopter Handbooks: no handbooks loaded (diff is shell hooks, not a `language/<lang>` bucket).

### Verification performed
I checked out the PR head (6250020) in an isolated worktree and ran the suite + adversarial probes directly against the hook:

1. **Documented indirect write, no active-reviewer marker → BLOCKED (exit 2).** ✓ (case 20/23 + my own `review_marker_path` probe). Fully-variable-arg form (`role`/`pr` both variables) also blocks — role unresolves to empty and can't match a real marker.
2. **Sanctioned reviewer (matching active-reviewer marker) via the same idiom → ALLOWED (exit 0).** ✓ (case 21).
3. **Literal-path read (`cat`/`grep`) → NOT blocked; literal-path write (`printf >`, `cp`, `tee`) → BLOCKED.** ✓ (case 22 + my `cp`/`grep` probes).
4. **Lib-missing fallback is conservative, not fail-open** — a literal read even blocks under fallback (re-applies the old false-positive). ✓ (case 25).
5. **Two pre-existing suite failures** (`test_pr_create_gate_anchor.sh`, `test_sync_codex_adapter.sh`) are outside this diff — the PR touches only the two hook files. Confirmed via `git diff --name-only upstream/dev...`.

### Issues Found
None blocking.

### Suggestions

**suggestion (security, defense-in-depth — non-blocking): make the "fails closed on total ambiguity" invariant literal instead of "well-formed-only".**

The PR's headline claim is that an unresolved role can't pass "by construction… no *well-formed* active-reviewer marker can ever have an empty kind field." That's precisely worded and true for legitimately-produced markers — and a **different-kind** forged marker does fail closed (case 24 proves it). But a *malformed* active-reviewer marker with an **empty kind** (`me2resh/apexyard#970:`) combined with a fully-unresolvable indirect write **passes the gate**:

```
MARKER_TYPE=""  (role unresolved) ; c_kind=""  (empty-kind marker)
[ "$c_kind" = "$MARKER_TYPE" ]  →  [ "" = "" ]  →  true  →  ALLOW
```

I confirmed this live (exit 0). It is **not a new capability** — it's strictly dominated by the already-acknowledged threat-model escape (a build agent can forge a *well-formed* `…#970:rex` active-reviewer marker directly; nothing gates writes to `active-reviewer`), so it grants an adversary nothing they couldn't already do more simply, and I am **not blocking on it**. But since #962 explicitly advertises fail-closed-by-construction and only the different-kind path is tested, a one-liner would make the invariant airtight rather than caveated:

```sh
_active_reviewer_allows() {
  [ -n "$MARKER_TYPE" ] || return 1   # unresolved role never satisfies the gate
  [ -f "$ACTIVE_REVIEWER_MARKER" ] || return 1
  ...
```

Pair it with a case-24 sibling that uses an **empty-kind** marker (not a different-kind one) so the guarantee is covered by a test.

**nit (non-blocking): `_is_marker_plausible_indirect`'s directory signal requires a trailing slash** (`\.claude/session/reviews/`). Splitting the dir into a variable without the slash — `D=.claude/session/reviews; printf x > "$D/…-rex.approved"` — matches neither the contiguous literal check nor the plausibility check, so it isn't gated. This is a **pre-existing** limitation (the old literal regex missed it too, so no regression) and squarely in the "deliberately-adversarial agent" space the threat model excludes — noting it only as a future hardening target. Dropping the trailing `/` from that one alternative would catch the dir-in-a-variable shape.

### Verdict
**APPROVED**

The fix does exactly what it claims: closes the real false-negative (variable idiom now gated + fails closed on unresolved role vs. legitimate/different-kind markers) and the false-positive (reads pass), reuses the established write-detector, and degrades conservatively when the lib is absent. 27/27 tests pass on HEAD. The one security suggestion is defense-in-depth against a forged malformed marker that is already dominated by an accepted threat-model escape — worth a follow-up one-liner, not a merge blocker.

---
🤖 Reviewed by Rex (Code Reviewer Agent) · security pass by Hakim
📌 Reviewed commit: `6250020f682c6a86b75c52dabb6b0f93484e21d9`

atlas-apex (commented):
## Security Review: PR #970 — `warn-review-marker-write.sh` resolved-target gate

**Reviewed commit**: `6250020f682c6a86b75c52dabb6b0f93484e21d9`
**Reviewer**: Hakim (Security Auditor) — adversarial trust-chain lens
**Verdict**: **PASS**

This is a trust-chain-critical hook: it's the mechanical gate that stops a build-class sub-agent from forging its own `*-rex.approved` / `*-security.approved` / `*-architecture.approved` marker to satisfy the merge gate. I verified the fix against the PR HEAD in an isolated worktree, ran the PR's own suite (27/27), and ran an independent adversarial probe harness (literal + indirect write shapes, empty-kind marker, keyword-avoidance, lib-missing, injection).

### Does the fix close the real forgery vector? — YES

The #962 bug was live-exploitable: the sanctioned reviewers' own documented idiom — `M=$(review_marker_path …); printf … > "$M"` — hid the `-rex.approved` suffix in a shell variable, so the old literal-substring matcher never fired and a build agent copying that idiom self-approved with **no** active-reviewer marker required. I confirmed every practical write shape of that idiom is now **BLOCKED (exit 2)** when no matching active-reviewer marker is set:

| Probe (no active-reviewer marker) | Result |
|---|---|
| literal `printf > …-rex.approved` | BLOCKED |
| indirect `review_marker_path` + `printf > "$M"` | BLOCKED |
| indirect `tee "$REX_MARKER"` | BLOCKED |
| indirect `cp … "$SECURITY_MARKER"` | BLOCKED |
| indirect heredoc `cat > "$ARCH_MARKER"` | BLOCKED |
| `sed -i … …-rex.approved` | BLOCKED |
| literal-path **read** (`cat`/`grep`/`head`) | correctly NOT blocked (false-positive fix) |

The write-vs-read gating via `_lib-detect-bash-write.sh` is the right structural choice and reuses the same battle-tested helper `require-active-ticket.sh` uses.

### Rex's empty-kind finding — reproduced, and I agree it is NON-BLOCKING

I reproduced it precisely. An indirect write that (a) trips a plausibility signal but leaves the role **unresolved** (e.g. references `$MARKER_HOME` — a `*marker*`-named var — but has no `review_marker_path` call to recover the role), combined with (b) a **malformed empty-kind** active-reviewer marker (`me2resh/apexyard#970:`), passes the gate because `_active_reviewer_allows` reduces to `[ "" = "" ]` → true. With a *well-formed* marker of any kind, or no marker, the same write correctly BLOCKS.

I agree with Rex that this is **not blocking**, and the security lens strengthens the reasoning rather than contradicting it:

1. **Strictly dominated.** Nothing gates writes to `.claude/session/active-reviewer` itself (confirmed: the only reference in `settings.json` is the SessionStart *clear* sweep, not a write-gate). Any actor able to plant an empty-kind marker can equally plant a **well-formed** `me2resh/apexyard#970:rex` marker, which defeats the gate via the clean matching path — a strictly more powerful tool. The empty-kind route grants zero incremental capability.
2. **Not new exposure — #962 *narrows* this surface.** The only useful variant (piece-assembling the `-rex.approved` suffix so the real gate-satisfying file lands on disk while the hook's role stays unresolved) sailed through the **pre-#962** hook unconditionally, with no marker at all. Post-#962 it is caught unless the attacker *also* crafts an empty-kind marker. That is a tightening, not a regression.
3. **Squarely inside the hook's documented non-claim** — the gate defends against the *induced* failure mode, not a deliberately adversarial agent forging session state.

**Recommended fast-follow (non-blocking):** land Rex's one-liner to make the fail-closed-on-unresolved-role invariant *unconditional* rather than contingent on active-reviewer well-formedness — the hook's own comments already assert this invariant, so this closes the gap between the stated guarantee and the code:

```sh
# in _active_reviewer_allows, before the kind equality:
[ -n "$MARKER_TYPE" ] || return 1
```

Plus an empty-kind active-reviewer test case — the current case 24 only exercises *non-empty-kind* ambiguity, so the fail-open corner is untested.

### Other checks

- **lib-missing fallback** — degrades to the pre-#962 literal-substring-only path; a literal marker write still BLOCKS (verified). Not fail-open. Correct.
- **Injection surface** — none. Role/PR extraction is constrained to fixed alternations (`rex|ceo|security|architecture`) and bare digit-runs; the BLOCKED banner is a variable-expanding heredoc that inserts values literally and never re-evaluates them (canary probe with an embedded `$(touch …)` payload did not execute). No `eval`, no raw command text reaching a shell.
- **Trailing-slash plausibility nit** (dir-in-a-variable without the `.claude/session/reviews/` trailing slash evading the signal) — real but adversarial-only, same accepted keyword-avoidance class as the split-path bypass that also existed pre-#962. Non-blocking; worth folding into the same hardening pass.

### Checklist

- Secrets & Credentials: Pass
- Injection Prevention: Pass
- Auth / provenance gate integrity: Pass (real forgery vector closed; residuals dominated + pre-existing)
- Fail-closed degradation: Pass (lib-missing, unresolvable-role-with-well-formed-marker)
- Data Protection: N/A

**Verdict: PASS.** The fix closes the genuine, live-exploitable forgery vector and introduces no new exploitable forgery path. The empty-kind finding and the trailing-slash nit are acceptable, documented, strictly-dominated follow-ups — recommend the two-line hardening + empty-kind test as a fast-follow.

---
📌 Reviewed commit: `6250020f682c6a86b75c52dabb6b0f93484e21d9`

Issue #974, opened 2026-07-22
[Task] Reject empty-kind active-reviewer marker in warn-review-marker-write.sh (fast-follow to #962)


atlas-apex:
## Driver
Fast-follow from #962 (merged in #970). `warn-review-marker-write.sh` now blocks the variable-indirected marker-write idiom, and fails closed when it can't resolve the marker's role — *except* for one edge Rex and Hakim both flagged: a **malformed empty-kind** active-reviewer marker (`<owner>/<repo>#<pr>:` with nothing after the colon) plus a fully-unresolvable indirect write reduces the gate's equality check to `[ "" = "" ]` → true, so the write is allowed. Both reviewers judged it non-blocking (strictly dominated — nothing gates writes to the `active-reviewer` marker itself, so an actor who can plant an empty-kind marker could plant a well-formed `…:rex` one, which defeats the gate more cleanly; the empty-kind route grants zero extra capability). But the hook's own comments already claim the "fails closed on total ambiguity" invariant **unconditionally**, while the code makes it contingent on marker well-formedness. Make the code match the advertised invariant.

## Scope
- In `warn-review-marker-write.sh`, add `[ -n "$MARKER_TYPE" ] || return 1` (reject an empty resolved kind) so an empty-kind active-reviewer marker can never satisfy the equality check.
- Add a test case: empty-kind active-reviewer marker (`…#<pr>:`) + unresolvable indirect write → BLOCKS (exit 2).

## Acceptance Criteria
- [ ] An empty-`kind` active-reviewer marker no longer passes the gate for any write.
- [ ] The sanctioned reviewer flow (well-formed `…:rex`/`:security`/`:architecture` marker) still works unchanged.
- [ ] New empty-kind test added and green; existing `test_warn_review_marker_write.sh` cases still pass.

## Context
Source: Rex + Hakim reviews on #970 (merged). Trust-chain change → gets Rex + Hakim on the fix PR. One-line code change + one test; the value is making the advertised invariant literally true.

## Glossary
| Term | Definition |
|------|------------|
| active-reviewer marker | `.claude/session/active-reviewer`, one line `<owner>/<repo>#<pr>:<kind>`, that authorises the sanctioned reviewer's marker write. |
| kind | The review type segment after the colon (`rex` / `security` / `architecture`); an empty kind is malformed. |
| fails closed | The safe default: BLOCK the write when the gate can't positively authorise it. |

atlas-apex:
[...]

Issue #957, opened 2026-07-21
[Feature] Configurable human merge-approver title — adopters aren't all CEOs


AhmedTheGeek:
## Problem

The framework names the human per-PR merge approval the **"CEO" gate** everywhere user-facing: `CLAUDE.md` quality rules, `.claude/rules/pr-workflow.md`, `/approve-merge`, the merge-gate hook messages, and the `<pr>-ceo.approved` marker filename. For solo founders that's accurate, but team adopters route this approval to whoever actually owns the repo — a maintainer, a dev lead, a team lead. The title is hardcoded in ~290 places across ~30 framework-owned files, so an adopter who wants their own vocabulary faces a fork-wide rename with permanent merge conflicts on every `/update`. In practice adopters settle for a "read CEO as X" note in their fork docs, and the framework's own output (hook block messages, skill confirmations) keeps contradicting it.

## Proposed behaviour

Make the approver **display title** configurable, keeping the mechanical layer unchanged:

- New key in `.claude/project-config.json`, e.g. `approval.human_approver_title` (default: `"CEO"` — zero behaviour change for existing adopters).
- Consumers: the block/OK messages in `block-unreviewed-merge.sh` (and siblings), the `/approve-merge` skill's prose and confirmation output, and ideally the rule docs referring to "the configured approver (default CEO)".
- **Marker filenames stay `<pr>-ceo.approved`** and the structured fields (`approved_by=user`, `skill_version`) stay as-is — renaming those would break existing sessions, tests, and the gate's grep-ability for no semantic gain. The title is presentation only.

## Why it helps adopters

Any adopter where the merge approver isn't literally the CEO (most teams) gets consistent vocabulary between their onboarding docs and what the framework prints at the moment of approval, without forking 30 framework files. Today they either live with the mismatch or carry a doc note telling agents to mentally translate.

## Scope / out of scope

Out of scope: renaming marker files, changing gate semantics, per-repo approver *identities/ACLs* (who may approve) — this is only the display title.

## Framework version

v3.2.0

## Related

`/approve-merge`, `block-unreviewed-merge.sh`, `warn-review-marker-write.sh`, `_lib-review-markers.sh`, `.claude/rules/pr-workflow.md`; same "adopter display-name override" shape as `.pr.title_pattern` (#?) and `ui_paths_exclude` (#275).

## Glossary

| Term | Definition |
|------|------------|
[...]

PR #975, opened 2026-07-22
fix(#974): reject empty-kind active-reviewer marker in warn-review-marker-write.sh

atlas-apex:
## Summary
- **Closes the residual "total ambiguity" hole in the review-marker gate** — `warn-review-marker-write.sh`'s `_active_reviewer_allows()` compared the write's resolved role (`MARKER_TYPE`) against the active-reviewer marker's parsed kind (`c_kind`) with a plain `[ "$c_kind" = "$MARKER_TYPE" ]`. A malformed marker on disk (`owner/repo#<pr>:` with nothing after the trailing colon) parses to an empty `c_kind`; paired with an indirect Bash write whose role can't be resolved to a literal `rex`/`ceo`/`security`/`architecture` token, `MARKER_TYPE` is also empty. The comparison collapsed to `[ "" = "" ]` → true, incorrectly **allowing** a review-marker write with no real active-reviewer session behind it.
- **Made the code match the comment's own claim** — the `#962` comment on that function already asserted the gate "fails closed on total ambiguity... by construction," which was true for a well-formed marker but not for a malformed one. Both sides of the kind comparison are now explicitly guarded for non-empty before the equality check runs, so an empty resolved kind can never satisfy it on either side.
- **Added a regression test (case26)** reproducing the exact scenario — an empty-kind active-reviewer marker plus an unresolvable indirect write — and verified via a scratch revert that it fails without the fix (hook exits 0, allowing the write) and passes with it. Full suite is now 28/28 (was 27/27).

## Testing
1. `bash .claude/hooks/tests/test_warn_review_marker_write.sh` — 28/28 passing, including the new case26.
2. `shellcheck -x .claude/hooks/warn-review-marker-write.sh .claude/hooks/tests/test_warn_review_marker_write.sh` — clean (one pre-existing info-level SC2295 notice on an unrelated line, present in `upstream/dev` before this change too).
3. Manually confirmed the regression by reverting the two new guards in a scratch copy of the hook — with the guards removed, case26 fails (hook exits 0 instead of 2); restoring the guards makes it pass again.

Refs #974

---

## Glossary
| Term | Definition |
|------|------------|
| Active-reviewer marker | A one-line session file (`.claude/session/active-reviewer`, format `<owner>/<repo>#<pr>:<kind>`) the orchestrator writes immediately before spawning a sanctioned reviewer agent (Rex/Hakim/Tariq). Its presence — and matching `(repo, pr, kind)` — is what lets a build agent's own review-marker write be told apart from the real reviewer's. |
| `MARKER_TYPE` | The hook's own resolved value for *which* marker role (`rex`/`ceo`/`security`/`architecture`) the current write targets — resolved either directly from a literal file path/marker filename, or indirectly by best-effort-parsing a `review_marker_path(...)` call embedded in a Bash command that writes via a shell variable. |
| `c_kind` | The kind field parsed out of the active-reviewer marker's own file content (the part after the `:`) — this is what `MARKER_TYPE` is compared against to decide whether a write is authorized. |
| Fail closed | Security design principle: when a check can't be conclusively resolved (ambiguous, malformed, or missing input), the correct behavior is to deny/block rather than default to allow. |
| Indirect write | A Bash command that writes a review marker via a shell variable (e.g. `REX_MARKER=$(review_marker_path ...); printf ... > "$REX_MARKER"`) rather than a literal file path — the documented idiom the real reviewer agents use, and the harder case for the hook to resolve a role from. |

### DIFF
--- a/.claude/hooks/tests/test_warn_review_marker_write.sh
+++ b/.claude/hooks/tests/test_warn_review_marker_write.sh
@@ -51,6 +51,15 @@
 #                 literal-path READ (cat) of a rex marker, no active-reviewer
 #                                                                  → BLOCKED, exit 2 (conservative pre-#962 fallback, not a bypass)
 #
+# #974 test (residual "total ambiguity" edge case — see the hook's #974
+# header comment):
+#   (26) Bash   → MALFORMED empty-kind active-reviewer marker
+#                 ("owner/repo#<pr>:" with nothing after the colon) +
+#                 an indirected write whose role can't be resolved to any of
+#                 rex/ceo/security/architecture                    → BLOCKED, exit 2
+#                 (without the #974 fix, both sides of the kind comparison
+#                 resolve to "" and the write would be incorrectly ALLOWED)
+#
 # Exit 0 if all cases pass; 1 on failure.
 
 set -u
@@ -490,6 +499,32 @@ case25() {
   rm -rf "$sb"
 }
 
+# ---------------------------------------------------------------------------
+# (26) Bash → MALFORMED empty-kind active-reviewer marker ("owner/repo#<pr>:"
+#      with nothing after the colon) + an indirected write whose role can't
+#      be resolved (mentions .claude/session/reviews/ but no literal
+#      rex/ceo/security/architecture token) → still BLOCKED (#974).
+#
+#      Before #974: c_kind parses to "" from the malformed marker, MARKER_TYPE
+#      resolves to "" from the unresolvable indirect write, and
+#      `[ "$c_kind" = "$MARKER_TYPE" ]` -> `[ "" = "" ]` -> true, incorrectly
+#      ALLOWING the write. #974 adds explicit non-empty guards on both sides
+#      so this can never pass.
+# ---------------------------------------------------------------------------
+case26() {
+  local sb; sb=$(make_sandbox)
+  # Malformed marker: trailing colon, nothing after it (empty kind).
+  printf '%s\n' "${REPO}#42:" > "$sb/.claude/session/active-reviewer"
+  # Same fully-ambiguous indirected write shape as case24: performs a write
+  # and mentions .claude/session/reviews/, but no literal role token appears
+  # anywhere, so _extract_marker_role (and thus MARKER_TYPE) resolves to "".
+  # shellcheck disable=SC2016 # deliberate — literal text, not real expansion
+  local cmd='DIR="$MARKER_HOME/.claude/session/reviews"; printf "%s" sha123 > "$DIR/mystery.approved"'
+  run_hook "$sb" "Bash empty-kind active-reviewer marker + unresolvable indirect write -> BLOCKED (#974)" \
+    "$(bash_json "$cmd")" 2 "BLOCKED"
+  rm -rf "$sb"
+}
+
 case1
 case2
 case3
@@ -515,6 +550,7 @@ case22
 case23
 case24
 case25
+case26
 
 # ---------------------------------------------------------------------------
 # Summary
--- a/.claude/hooks/warn-review-marker-write.sh
+++ b/.claude/hooks/warn-review-marker-write.sh
@@ -136,7 +136,14 @@
 #   matcher: Write    (catches direct file writes)
 #   matcher: Bash     (catches shell redirections, echo >, printf, tee, etc.)
 #
-# References: #728, #843, #873, #957, #962, AgDR-0062, AgDR-0104,
+# #974 hardened _active_reviewer_allows() further: a malformed on-disk
+# active-reviewer marker (empty kind after a trailing colon) combined with
+# an unresolved indirect-write role could satisfy `[ "" = "" ]` and
+# incorrectly ALLOW the write — the exact "total ambiguity" case the #962
+# comment assumed was already fail-closed. Both sides of the kind
+# comparison are now guarded explicitly for non-empty before it runs.
+#
+# References: #728, #843, #873, #957, #962, #974, AgDR-0062, AgDR-0104,
 #             .claude/rules/pr-workflow.md § "Build agents cannot self-review"
 
 set -u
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #975

**Commit**: `76f6ba6f1f454950c6f776de2c526b7494016971`

### Summary
Trust-chain hardening for `warn-review-marker-write.sh` — the gate that stops a build-class sub-agent from forging its own `*-rex/security/architecture.approved` marker. It closes the residual "total ambiguity" hole found on #970: a malformed empty-kind active-reviewer marker (`owner/repo#<pr>:` with nothing after the colon) combined with an unresolvable indirect write reduced `_active_reviewer_allows()`'s kind check to `[ "" = "" ]` → true, incorrectly allowing the write. The fix adds explicit non-empty guards on both sides of the comparison so the "fails closed on total ambiguity" invariant the #962 comment already claimed is now literally true.

### Checklist Results
- Security:                  Pass (trust-chain fail-closed hardening)
- Technical Decisions (AgDR): N/A — bug fix to an existing gate, no new library/pattern/dependency
- Adopter Handbooks:          N/A — no handbook applies to a shell-hook change

### Review focus verification (trust-chain bar)

1. **Empty-kind hole closed — confirmed.** Traced the #974 scenario against the PR HEAD: an empty-kind marker (`me2resh/apexyard#975:`) + an unresolvable indirect write (`MARKER_TYPE=""`). Line 397 `[ -n "$MARKER_TYPE" ] || return 1` sits at the very top of `_active_reviewer_allows()`, before the equality check, so an empty resolved role returns 1 (deny) → the gate falls through to the blocking banner → **exit 2**. Both the literal (`RESOLVED_VIA=literal`) and indirect paths flow through this single function, so an empty kind can never reach `[ "$c_kind" = "$MARKER_TYPE" ]` and pass. Line 415 `[ -n "$c_kind" ] || return 1` guards the marker-side symmetrically.

2. **Sanctioned flow intact — confirmed.** A well-formed `…#975:rex` marker + a rex write resolves `MARKER_TYPE=rex` and `c_kind=rex`; both guards pass (non-empty), equality holds, PR/repo checks pass → exit 0 (allow). Ran the full PR HEAD suite: 28/28 including all prior sanctioned-path cases.

3. **No over-block / no new bypass — confirmed.** The guards reject *only* an empty kind. I reverted both guards in a scratch copy and re-ran the suite: exactly one case regressed (case26, exits 0 instead of 2) — every other case behaved identically with and without the guards. That surgically proves the change neither newly blocks a legitimate write nor opens a new path. Pre-fix, an unresolvable-role write against a *well-formed* marker was already denied by the equality check, so the only newly-denied case is the malformed empty=empty collapse, which is exactly the intended fix.

4. **Test quality — genuine, not a tautology.** case26 asserts both `exit 2` and stderr matching `BLOCKED` (the hook prints "BLOCKED" and `exit 2` on the block path). The regression check above shows it fails (exit 0) with the fix reverted, so it truly exercises the empty-kind path rather than passing vacuously. Suite: 28/28 (was 27/27).

5. **Correct variable names — confirmed.** Line 397 guards `$MARKER_TYPE` (the resolved write role, initialised to `""` at top level so `set -u`-safe) and line 415 guards `$c_kind` (the local parsed from the marker content at line 413) — both are the variables as actually resolved in the script.

### Issues Found
None.

### Suggestions
- nit (non-blocking): guarding both sides is belt-and-suspenders — for this specific scenario line 397 alone closes the empty=empty collapse, and when `MARKER_TYPE` is non-empty the equality check already rejects an empty `c_kind`. The extra line 415 guard is cheap, self-documenting, and robust to future refactors, so keeping it is the right call for a trust-chain gate; noting only that it isn't strictly required for the demonstrated hole.

### Verdict
**APPROVED**

Independently verified against the PR HEAD (not the diff alone): built the HEAD file + test in a sandbox, ran the suite green (28/28), and confirmed the regression by reverting the guards (27/28, only case26 fails). Fail-closed invariant is now literally true.

---
📌 Reviewed commit: `76f6ba6f1f454950c6f776de2c526b7494016971`

atlas-apex (commented):
## Security Review: PR #975

**Commit**: `76f6ba6f1f454950c6f776de2c526b7494016971`

### Summary
Trust-chain hardening (implements #974, flagged during #970). Closes a residual
fail-**open** collapse in `warn-review-marker-write.sh::_active_reviewer_allows()`:
a malformed on-disk active-reviewer marker with an empty kind (`owner/repo#<pr>:` —
nothing after the colon) combined with a #962 indirect write whose role can't be
resolved (`MARKER_TYPE=""`) made the gate's kind check collapse to `[ "" = "" ]` →
true, ALLOWING a forged `*-security/rex/architecture.approved` marker write. Two
symmetric non-empty guards are added at the top of the function.

### Verification
Read the PR-head function (base `dev` carries the #957/#962 machinery; the local
checkout is older, so I reasoned against the fetched HEAD blob, not the working tree)
and reproduced the arithmetic:

- **Empty-kind forgery now BLOCKS (both paths).** The empty-role condition is
  reachable only on the #962 **indirect** path (the literal path derives
  `MARKER_TYPE` from a regex-constrained basename, so it's never empty). Guard 1
  `[ -n "$MARKER_TYPE" ] || return 1` rejects the unresolved-role indirect write
  before any comparison; guard 2 `[ -n "$c_kind" ] || return 1` independently rejects
  a malformed empty-kind on-disk marker even when the role *does* resolve. The two
  guards are individually motivated (attacker-controlled unresolved role vs.
  attacker-controlled malformed marker) and each closes the collapse on its own —
  correct defense-in-depth, not redundancy. Probe confirmed pre-fix `[ "" = "" ]`→true,
  post-fix both guards `return 1`.
- **Sanctioned flow still ALLOWED.** A well-formed `owner/repo#975:security` marker
  with a resolved role (`security`) passes both guards unchanged and matches on
  kind/PR/repo. The guards only reject an *empty* resolved role, which never occurs
  in a real review. No over-block, no new bypass.
- **Accepted boundary respected.** Nothing gates writes to the `active-reviewer`
  file itself — pre-existing, documented, out of scope. This PR only makes the
  advertised "fail-closed-on-ambiguity" invariant literally true.

### Checklist Results
- Injection Prevention:   Pass (no shell interpolation of untrusted data added)
- XSS Prevention:         N/A
- Auth & Authorisation:   Pass (this IS the authZ gate; it now fails closed)
- API Security:           N/A

### Observation (non-blocking, follow-up only)
Same-class residual, **pre-existing and NOT introduced by this PR**: on the indirect
path where the role resolves but the PR does not (`TARGET_PR=""`), the PR check
`[ -n "$TARGET_PR" ] && [ "$c_pr" != "$TARGET_PR" ]` is skipped, so a resolved-role
indirect write could match *any* active-reviewer marker of the same kind regardless
of PR number. This requires a legitimately-set active-reviewer marker to already be
in flight and is the accepted best-effort nature of #962 indirect resolution — out
of scope for #974, worth a future defense-in-depth ticket, not a blocker here.

### Verdict
**APPROVED** — the PR closes exactly the fail-open path #974 targets, with two
independently-correct guards, a matching regression test (case 26), and no
over-block of the sanctioned flow.

---
📌 Reviewed commit: `76f6ba6f1f454950c6f776de2c526b7494016971`

me2resh (approved):

me2resh (approved):

me2resh (approved):

Issue #973, opened 2026-07-22
[Task] Normalise JSON escapes in merge-gate raw-payload fallback (fast-follow to #965)


atlas-apex:
## Driver
Fast-follow from #965 (merged in #970... corrected: #969). When `jq` is unavailable, the four merge-gate hooks now fall back to running the jq-free `is_merge_command` detector against the **raw JSON payload text**. Hakim's security review of #969 confirmed this closes the real fail-open for normal space-separated merge commands, but flagged one contrived residual: a merge invocation whose separators are JSON-escaped (e.g. a tab encoded as the two-char `\t`, or `\uXXXX` / `\/` escapes) evades the raw-text `\s+` grep **while jq is broken**, and is allowed. It's non-blocking (requires jq already broken AND a deliberately obfuscated command no normal flow emits; the normal jq path decodes `\t` to a real tab and catches it), but the defense-in-depth fix is cheap.

## Scope
- In the raw-payload fallback path of `block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh` (or the shared helper they call), normalise common JSON escape sequences (`\t`, `\n`, `\uXXXX`, `\/`) to their literal characters before running `is_merge_command` on the raw text.
- Add a regression test: jq stubbed to fail + a `\t`-separated merge command in the payload → BLOCKS (exit 2).
- Keep the non-merge no-op behaviour intact (these hooks fire on every Bash call).

## Acceptance Criteria
- [ ] With jq broken, a JSON-escape-obfuscated merge command (`\t` / `\uXXXX` separators) is blocked, not allowed.
- [ ] Normal (jq-present) path behaviour unchanged; non-merge commands still no-op under the fallback.
- [ ] Regression test added and green; existing merge-gate suites still pass.

## Context
Source: Hakim's security review on #969 (merged). Strictly defense-in-depth — the vector needs jq already broken. Trust-chain change → gets Rex + Hakim on the fix PR.

## Glossary
| Term | Definition |
|------|------------|
| fail-open | A gate that ALLOWS the guarded action when it can't evaluate its precondition (the bug class #965 fixed). |
| raw-payload fallback | Running the jq-free command detector against the unparsed JSON tool-input when jq is unavailable. |

atlas-apex:
[...]

PR #976, opened 2026-07-22
fix(#973): normalise JSON escapes in merge-gate raw-payload fallback

atlas-apex:
## Summary

- **Closes the residual JSON-escape evasion Hakim found in #969** — the four merge-gate hooks (`block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh`) fall back to scanning the RAW, still-JSON-encoded payload text with `is_merge_command` when jq is unavailable (#965). That works for a normal space-separated merge command because `gh`, `pr`, `merge` and the spaces between them survive JSON string-encoding unchanged — but a command whose *separators* are themselves JSON-escaped (a literal tab encodes as the two-character sequence `\t`, other bytes as `\uXXXX` or `\/`) does NOT survive unchanged, and those escape sequences are not whitespace to `is_merge_command`'s `\s+` regex. Pre-fix, a tab-separated `gh	pr	merge …` sailed through the fallback ungated while jq was broken — the exact class of gap the #965 fix was supposed to close.
- **Added a shared `_normalize_json_escapes()` helper to `_lib-extract-pr.sh`**, decoding `\t`, `\n`, `\/`, and `\uXXXX` (space/tab/newline) to their literal characters, and wired it into all four hooks' fallback call sites: `is_merge_command "$(_normalize_json_escapes "$INPUT")"`. I chose the **shared-helper** path from the two options in the ticket, not four separate per-hook fixes, because the four fallback branches are byte-identical copy-paste and a shared function is the DRY choice the file's own conventions already favor (see `_extract_wrapper_arg` for the existing "one tested helper, four callers" pattern in this same file).
- **Deliberately did NOT normalize inside `is_merge_command` itself.** `is_merge_command` has two call sites per hook: the fallback (`$INPUT`, raw JSON) and the normal path (`$COMMAND`, already jq-decoded). jq has *already* correctly decoded these same escapes on the normal path — normalizing again there would be redundant at best, and for a command that legitimately contains a literal backslash-t/backslash-n substring (e.g. inside a `sed` script), would corrupt text jq had already decoded correctly. Keeping the helper external to `is_merge_command` and calling it ONLY at the fallback sites means the normal jq-present path is provably byte-for-byte unchanged — confirmed by running the full `test_forge_aware_extract_pr.sh` suite (which exercises `is_merge_command` directly, jq-present-path style) unchanged and green.
- **Two bugs surfaced and fixed during implementation**, both interesting enough to flag: (1) bash's `${var//pattern/replacement}` treats a *single* backslash inside an expanded pattern as a glob escape character rather than a literal match target, so a naive `esc_t="\t"` pattern silently failed to consume the backslash itself — fixed by doubling the backslash in the pattern-building variable; (2) `nl=$(printf '\n')` loses the newline entirely because command substitution strips trailing newlines — fixed with ANSI-C quoting (`nl=$'\n'`) instead.
- **Regression tests added to both suites that already carry the #965 jq-broken fixtures** (`test_block_unreviewed_merge.sh`, `test_block_merge_on_red_ci.sh`): jq stubbed to fail + a JSON-escaped-tab merge command → BLOCKS, and a JSON-escaped-tab non-merge command → stays a no-op (proving the fix doesn't turn every tab-containing payload into a false positive).

## Testing

- `bash .claude/hooks/tests/test_block_unreviewed_merge.sh` — 39/39 (37 pre-existing + 2 new)
- `bash .claude/hooks/tests/test_block_merge_on_red_ci.sh` — 27/27 (25 pre-existing + 2 new)
- `bash .claude/hooks/tests/test_require_architecture_review.sh` — 22/22 (unchanged)
- `bash .claude/hooks/tests/test_require_design_review_for_ui.sh` — 22/22 (unchanged)
- `bash .claude/hooks/tests/test_forge_aware_extract_pr.sh` — 48/48 (unchanged; proves `is_merge_command`'s own behaviour is untouched)
- `shellcheck -x` on every touched file — clean (only pre-existing info-level notices, plus one new info-level SC1003 nit on the `bs2='\\'` literal, which is stylistic and doesn't affect correctness)
- `bin/run-hook-tests.sh` — 112/113 (one pre-existing, unrelated failure in `test_sync_codex_adapter.sh`, confirmed untouched by this diff; `test_pr_create_gate_anchor.sh` passes)

## Glossary

| Term | Definition |
|------|------------|
| Fail closed / fail open | A gate that can't evaluate its own precondition either blocks (fail closed — safe default for a security gate) or lets the action through (fail open — the bug class #965 fixed and #973 closes a residual gap in) |
| Raw-payload fallback | The code path in each merge-gate hook that runs when `jq` is unavailable to parse `.tool_input.command` from the PreToolUse JSON — scans the undecoded JSON text directly instead |
| `is_merge_command` | The shared detector in `_lib-extract-pr.sh` that recognises every merge-command shape (`gh pr merge`, `gh api .../pulls/<N>/merge`, `glab mr merge`, `glab api .../merge_requests/<N>/merge`, the `tracker_pr_merge` wrapper) |
| JSON string escape | The two-or-more-character sequences JSON uses to represent control characters or special bytes inside a string literal — `\t` for tab, `\n` for newline, `\/` for a forward slash, `\uXXXX` for an arbitrary code point |

Refs #973

### DIFF
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -92,6 +92,43 @@
 # `extract_repo_from_command` each gained a dedicated wrapper-arg extraction
 # step using `_extract_wrapper_arg` (quoted-or-bare positional-token parsing,
 # regex/parameter-expansion only — no `eval` of the command text, ever).
+#
+# JSON-ESCAPED SEPARATORS IN THE RAW-PAYLOAD FALLBACK (#973, a residual
+# finding from Hakim's #969 review of the #965 fix)
+# ------------------------------------------------------------------------
+# #965 (see the four call sites in block-unreviewed-merge.sh,
+# block-merge-on-red-ci.sh, require-design-review-for-ui.sh, and
+# require-architecture-review.sh) added a jq-free fallback: when jq is
+# unavailable, each hook calls `is_merge_command "$INPUT"` directly against
+# the RAW, still-JSON-encoded payload text instead of the jq-decoded
+# command string. That works because the command's own words (`gh`, `pr`,
+# `merge`, digits) survive JSON string-encoding unchanged — EXCEPT for the
+# command's *separators*, when those separators are themselves characters
+# JSON must escape: a literal tab encodes as the two-character sequence
+# `\t`, and some encoders emit `\uXXXX` or `\/` for other bytes. Those
+# two-character escape sequences are not whitespace to `grep -E`'s `\s`
+# class, so `is_merge_command`'s `\bgh\s+pr\s+merge\b` pattern silently
+# fails to match a merge command whose separators are JSON-escaped —
+# exactly while jq (the thing that would normally decode them into real
+# whitespace) is unavailable. A gate that can't evaluate its own
+# precondition must fail closed, not quietly no-op on a payload that IS
+# merge-shaped once decoded.
+#
+# `_normalize_json_escapes` (below) is a small, best-effort decoder for
+# the handful of escape shapes that matter here — NOT a full JSON string
+# parser, the same "regex/parameter-expansion only, sufficient for the
+# shapes real callers emit" discipline as `_extract_wrapper_arg` above.
+# It is called ONLY at the four hooks' raw-payload fallback call sites
+# (`is_merge_command "$(_normalize_json_escapes "$INPUT")"`), never from
+# inside `is_merge_command` itself and never on the normal jq-present
+# path: jq has ALREADY correctly decoded these same escapes for that path
+# (that's what `jq -r` does), so re-normalizing already-decoded text would
+# be redundant at best and, for the rare case of a command that legitimately
+# contains a literal backslash-t/backslash-n substring (e.g. inside a sed
+# script), actively wrong — it would corrupt real command text that jq had
+# already decoded correctly. Keeping the two paths separate is what makes
+# this change safe for the jq-present callers: their behaviour is provably
+# unchanged because they never call the new function at all.
 
 # Lazily source the tracker lib so `tracker_kind` is available for forge
 # resolution. Guarded: only source if not already defined and the lib is
@@ -170,6 +207,64 @@ _extract_wrapper_arg() {
     fi
   done
   echo ""
+}
+
+# Best-effort decode of the JSON string escapes that could hide a merge
+# command's SEPARATORS from the raw-payload fallback scan (#973). Echoes
+# the normalized text.
+#
+# Handles six literal JSON escape sequences, decoded to the real character
+# they represent: backslash-t and backslash-u0009 both decode to a tab;
+# backslash-n and backslash-u000A (either case) decode to a newline;
+# backslash-u0020 decodes to a plain space; backslash-slash decodes to a
+# plain slash.
+#
+# This is deliberately NOT a full JSON string decoder — no handling of
+# arbitrary \uXXXX code points, no awareness of escaped-backslash context
+# (a literal `\\t` — escaped backslash followed by a bare `t` — will still
+# be (mis-)decoded as a tab; that's an accepted, documented limitation, not
+# a security gap: it can only make the fallback MORE eager to treat text as
+# merge-shaped, i.e. fail closed, never a new way to evade it). Pure bash
+# parameter expansion — no eval, no external process, no regex
+# backtracking on attacker-controlled text — matching the same discipline
+# as `_extract_wrapper_arg` above.
+#
+# Callers: ONLY the four merge-gate hooks' raw-payload fallback branches
+# (`is_merge_command "$(_normalize_json_escapes "$INPUT")"`), never
+# `is_merge_command` itself and never the normal jq-present path — see the
+# file header (#973) for why mixing this into the jq-present path would be
+# unsafe.
+_normalize_json_escapes() {
+  local text="$1"
+  local tab=$'\t'
+  local nl=$'\n'
+  # Two literal backslash characters. Used as the escape-matching prefix
+  # below rather than a single backslash: bash's `${var//pattern/repl}`
+  # treats a SINGLE backslash inside an expanded pattern as a glob escape
+  # character (it would consume/escape the following char instead of
+  # matching a literal backslash), so building the pattern from a
+  # single-backslash variable silently fails to consume the backslash
+  # itself — the doubled form is what makes the pattern match one literal
+  # backslash followed by the literal marker character.
+  local bs2='\\'
+
+  local esc_u0020="${bs2}u0020"
+  local esc_u0009="${bs2}u0009"
+  local esc_u000A="${bs2}u000A"
+  local esc_u000a="${bs2}u000a"
+  local esc_slash="${bs2}/"
+  local esc_t="${bs2}t"
+  local esc_n="${bs2}n"
+
+  text="${text//$esc_u0020/ }"
+  text="${text//$esc_u0009/$tab}"
+  text="${text//$esc_u000A/$nl}"
+  text="${text//$esc_u000a/$nl}"
+  text="${text//$esc_slash//}"
+  text="${text//$esc_t/$tab}"
+  text="${text//$esc_n/$nl}"
+
+  printf '%s' "$text"
 }
 
 # Returns 0 if $1 looks like a merge command this gate should fire on.
--- a/.claude/hooks/block-merge-on-red-ci.sh
+++ b/.claude/hooks/block-merge-on-red-ci.sh
@@ -86,12 +86,21 @@ if [ -z "$COMMAND" ]; then
   # string-encoding unchanged, so this is the exact same tested
   # merge-shape detector used below — not a second, drift-prone regex.
   #
+  # #973: the command's SEPARATORS do not always survive unchanged — a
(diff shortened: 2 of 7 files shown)


atlas-apex (commented):
## Code Review: PR #976

**Commit**: `e62e2de8c4a77ce302a2a04cf65d69672ec658cb`

### Summary
Closes the residual JSON-escape evasion Hakim flagged in #969: when jq is unavailable, the four merge-gate hooks fall back to scanning the raw, still-JSON-encoded payload with `is_merge_command`. A merge command whose separators are JSON-escaped (a literal tab encodes as the two-char sequence `\t`, slashes as `\/`) evaded the fallback's `\s+` match while jq was broken — a fail-open. The fix adds `_normalize_json_escapes()` to `_lib-extract-pr.sh` (decodes `\t`, `\n`, `\/`, and the whitespace `\uXXXX` forms) and wires it into the four fallback call sites only.

I verified this against the PR HEAD in an isolated worktree, under the actual hook runtime (**bash 3.2.57**), not just by trusting the test output.

### Checklist Results
- Architecture & Design:     Pass — shared-helper reuse mirrors the existing `_extract_wrapper_arg` "one tested helper, four callers" pattern in the same file
- Code Quality:              Pass — pure parameter-expansion, no eval, no external process on attacker-controlled text
- Testing:                   Pass — new regression cases are genuine, not tautological (proof below)
- Security:                  Pass — fail-closed achieved on all four hooks; no residual fail-open; no over-block
- Performance:               Pass — negligible
- PR Description & Glossary:  Pass — narrative summary + glossary present; `Refs #973` correct for the dev-cut model
- Technical Decisions (AgDR): N/A — bug fix implementing an existing decision (#973); no new library/architecture choice
- Adopter Handbooks:         N/A — always-load handbooks (clean-architecture, migration-safety, commit-message-quality) don't trigger on a bash-hooks diff with no TS/domain/migration content

### Independent verification (trust-chain bar)

**1. Fail-closed actually achieved.** End-to-end, with a broken-jq stub on PATH + a payload whose command has real tabs (JSON-encoded to `\t` by the system jq), all four hooks BLOCK:
- `block-unreviewed-merge.sh` → exit 2
- `block-merge-on-red-ci.sh` → exit 2
- `require-architecture-review.sh` → exit 2  (no dedicated #973 test, verified by hand)
- `require-design-review-for-ui.sh` → exit 2  (no dedicated #973 test, verified by hand)

Byte-level: `_normalize_json_escapes 'a\tb'` → `61 09 62` (real tab), `a\nb` → `61 0a 62`, `a\/b` → `61 2f 62`. Decode confirmed.

**2. Normal jq-present path provably untouched.** `is_merge_command` never calls the new helper (grepped: the only remaining `is_merge_command "$INPUT"` reference is inside a comment). Normalize is invoked *only* at the four fallback sites. `test_forge_aware_extract_pr.sh` (which drives `is_merge_command` directly) stays **48/48**. A `sed` script carrying a literal `\t` on the normal path is passed through jq-decoded and never re-normalized.

**3. No new bypass / no over-block.** Escaped-tab *non-merge* payloads (`echo\thello\tworld`) stay a no-op (exit 0) on all four hooks under broken jq — the fix doesn't turn every tab-containing payload into a spurious block. The documented `\\t` mis-decode limitation can only ever make the fallback *more* eager to treat text as merge-shaped (fail-closed), never a new evasion. Newline/CR separators aren't valid single-command separators, so they're not an exploitable gap.

**4. Bash string-op gotchas are correct.** The doubled-backslash pattern (`bs2='\\'` → `esc_t='\\t'`) is exactly what bash's `${var//pattern/repl}` needs to match one literal backslash + `t` — confirmed decoding under bash 3.2.57. `nl=$'\n'` is a real newline (not the newline-stripping `$(printf '\n')`). shellcheck clean (one pre-existing info-level SC1003 nit only).

**5. Tests are genuine, not tautologies.** Under bash I confirmed a raw `gh\tpr\tmerge` (literal backslash-t) does **not** match `is_merge_command` without normalize, and `\s` does not match a literal `\t` — so the normalize step is doing the actual work, and the new cases assert both exit 2 and the `cannot evaluate this command` stderr. `test_block_unreviewed_merge.sh` 39/39, `test_block_merge_on_red_ci.sh` 27/27. The pre-existing `test_sync_codex_adapter.sh` failure is unrelated — not among this PR's 7 files.

### Suggestions
- nit (non-blocking): the decode's correctness is coupled to **bash's** `${//}` glob-escape semantics — under zsh or a POSIX `sh`/dash the same doubled-backslash pattern silently no-ops (I hit exactly this while cross-checking under zsh before switching to bash). This is *correct as written* because every hook carries a `#!/usr/bin/env bash` shebang and the harness runs them under bash, so there's no action needed. Worth a one-line note in `_normalize_json_escapes` that it assumes a bash `${//}` runtime, in case a future refactor ever sources this lib from a non-bash context.

### Verdict
**APPROVED**

This is exemplary trust-chain hardening: the fix is minimal and surgically scoped to the fallback path, the fail-closed posture holds on all four gates, the normal path is provably untouched, and the regression tests aren't tautological. Deep independent verification found no blocking issues.

---
📌 Reviewed commit: `e62e2de8c4a77ce302a2a04cf65d69672ec658cb`

atlas-apex (commented):
## Security Review: PR #976 — trust-chain hardening (#973)

**Commit**: `e62e2de8c4a77ce302a2a04cf65d69672ec658cb`

### Summary
Adversarial pass on the jq-broken fallback fix. `_normalize_json_escapes()` decodes JSON-escaped separators (backslash-t, backslash-n, backslash-uXXXX for tab/nl/space, backslash-slash) at the four merge-gate fallback sites so a merge command whose separators are escaped is caught by `is_merge_command`'s `\s+` regex while jq is unavailable. I reasoned from the fetched HEAD blob and ran targeted probes against the real `_lib-extract-pr.sh` at HEAD.

### Checklist Results
- Secrets & Credentials:  Pass (none touched)
- Fail-open closure:      Pass
- Fail direction (residual): Pass
- Normal (jq-present) path untouched: Pass

### 1. Is the exploitable fail-open genuinely closed?
**Yes.** The load-bearing constraint is bash word-splitting: for an obfuscated gh<sep>pr<sep>merge to both (a) evade detection and (b) actually execute as a merge, <sep> must be a bash argument separator (IFS = space, tab, newline). Of those, only **tab** is JSON-escaped (space stays literal and already matches `\s`; newline is a *command* separator, not a word separator, so gh-backslash-n-pr executes as three failed commands, never a merge). Tab is now decoded on all four hooks. Probes at HEAD confirm MATCH for: escaped-tab gh-backslash-t-pr-merge, literal-tab form, gh api ...pulls-backslash-slash-306-backslash-slash-merge (esc-slash), and tracker_pr_merge with escaped tab. So the one escape variant that is simultaneously a valid bash arg-separator AND JSON-escaped is covered — the exploitable class is closed.

### 2. Escape variants the normalizer misses — not exploitable
Carriage return, form feed, vertical tab, and general backslash-uXXXX are **not** decoded (probes: CR/FF -> no-match). This is **not** an evasion, because none of those characters are bash argument separators — gh<CR>pr<CR>merge is a single un-runnable token, never a merge. There is no character that is at once a bash arg-separator, JSON-escaped, and left undecoded. Likewise the decoded-newline branch is inert for detection (grep is line-oriented, so a real newline splits the line) — harmless, and moot since newline-separated text is not an executable merge anyway.

### 3. No new attack surface from the normalizer
Pure bash parameter expansion (`${text//a/b}`) + `printf '%s'`; output is passed **quoted** into `is_merge_command`, which greps quoted `"$cmd"`. No `eval`, no command substitution of attacker text, no regex backtracking on attacker input. It cannot be turned into injection, and it cannot **consume** a merge keyword: every substitution only injects whitespace/slash or converts an escape to whitespace — it can never delete a gh/pr/merge letter or a digit. So there is no direction in which a genuine merge command is normalized into a non-matching (fail-open) form; the documented escaped-backslash-t mis-decode only adds a tab (fails **more** closed).

### 4. Normal jq-present path untouched
Confirmed from the diff: each hook's change is exactly one line **inside** the `if [ -z "$COMMAND" ]` fallback branch. `_normalize_json_escapes` is never called from `is_merge_command` nor on the COMMAND path, so a legitimate command carrying a literal backslash-t (e.g. a sed script) is unaffected on the normal path.

### Recommendations (non-blocking, defense-in-depth)
- Optionally extend the decoder to general backslash-uXXXX (any code point) plus CR/FF so the fallback's own text is fully decoded regardless of the encoder's choices. Not exploitable today (those chars are not executable arg-separators), so advisory only.
- The newline decode branch adds no real detection capability (line-oriented grep + newline-is-command-separator). Fine to keep for symmetry; worth a one-line comment noting it is inert.

### Verdict
**PASS** — the defense-in-depth fix closes the one exploitable escape variant (tab), introduces no injection or fail-open surface, and leaves the normal path provably unchanged. Residual undecoded escapes are non-exploitable because they are not bash argument separators.

---
Reviewed commit: `e62e2de8c4a77ce302a2a04cf65d69672ec658cb`

me2resh (approved):

me2resh (approved):

me2resh (approved):

PR #992, opened 2026-07-23
fix(#977): fail closed on unresolved target PR in warn-review-marker-write indirect path

### LINKED ISSUE #977: [Task] Fail closed on unresolved target PR in warn-review-marker-write.sh indirect path (fast-follow to #962/#975)
## Driver
Found by Hakim during the #975 (#974 fast-follow) security review. `warn-review-marker-write.sh`'s `_active_reviewer_allows()` compares a marker write against the `active-reviewer` session marker (`<owner>/<repo>#<pr>:<kind>`). On the #962 **indirect** resolution path, when the role/kind resolves but the **PR number (`TARGET_PR`) does not**, the PR-equality check is skipped — so a resolved-role indirect marker write could match *any* active-reviewer marker of that kind **regardless of which PR the review is actually for**. It requires an active review already in flight for that kind, and it's pre-existing #962 behaviour (NOT introduced by #975 — #975 closed the empty-*kind* sibling of this), so it was flagged as a follow-up rather than blocking the merge.

## Scope
- In `_active_reviewer_allows()` (in `.claude/hooks/warn-review-marker-write.sh`), when the write's target PR can't be resolved on the indirect path, fail closed rather than skipping the PR-equality check — i.e. an unresolved `TARGET_PR` should NOT be allowed to match an active-reviewer marker for a different PR. Mirror the empty-kind guard shape from #975 (`[ -n "$TARGET_PR" ] || return 1`, using the actual variable name) after confirming the legitimate reviewer flow always resolves a PR.
- Verify the sanctioned flow (a real reviewer whose PR resolves correctly) still works.

## Acceptance Criteria
- [ ] An indirect marker write whose target PR is unresolvable no longer matches an active-reviewer marker for a different PR — it BLOCKS (exit 2).
- [ ] The sanctioned reviewer flow (PR resolves, kind matches) still ALLOWS the write.
- [ ] Regression test added (unresolved-PR indirect write + active-reviewer marker for a *different* PR → blocked); existing `test_warn_review_marker_write.sh` cases stay green.

## Context
Source: Hakim's security review on #975 (merged). Same class as #974 (empty-kind) but on the PR field. Trust-chain change → gets Rex + Hakim on the fix PR. Pre-existing since #962; low practical exploitability (needs a concurrent active review of the same kind) but worth closing to make the gate's per-PR binding literally hold on the indirect path.

## Glossary
| Term | Definition |
|------|------------|
| active-reviewer marker | `.claude/session/active-reviewer`, `<owner>/<repo>#<pr>:<kind>`, authorising the sanctioned reviewer's marker write. |
| indirect path | The #962 detection branch that catches a marker write whose literal suffix is hidden in a variable. |
| per-PR binding | The property that a review approval marker is only valid for the specific PR it names. |

atlas-apex:
## Summary

- **Closes the per-PR-binding hole on the indirect resolution path** — on the #962 indirect path, the marker-write gate resolved the write's *role* (rex/security/architecture) but could leave the target PR number unresolved (when the marker-path helper is called with a variable PR arg). The old PR-equality check was written as `[ -n "$TARGET_PR" ] && [ "$c_pr" != "$TARGET_PR" ]`, so an empty target PR **skipped the check entirely** — and since the repo is also unrecoverable on that path, a resolved-role write matched *any* authorised session marker of the same kind, no matter which PR that review was actually for. The gate's "this approval is only valid for the PR it names" property silently didn't hold there.
- **The fix is a one-line fail-closed guard** — add a non-empty check on the target PR before the equality test, so an unresolved PR now BLOCKS instead of matching a different PR's marker. Mirrors the empty-kind guard shape from #974.
- **No impact on the legitimate reviewer flow** — the sanctioned reviewer idiom always passes a *literal* PR number to the marker-path helper (`code-reviewer.md:699`, `solution-architect.md:236`), so the PR extractor always resolves it. Verified before writing the guard, per the ticket's request. The empty-PR case only arises for a forged/malformed write using a variable PR — exactly what should fail closed.
- **Two regression tests** — case27 drives the exact hole (indirect write, role resolves, PR unresolved, authorised marker present for a *different* PR → BLOCKED) and case28 is the no-regression companion (both role and PR resolve, matching marker → still ALLOWED). Full suite: 30/30 assertions green.

## Testing

1. `bash .claude/hooks/tests/test_warn_review_marker_write.sh` → `PASS: 30 FAIL: 0`
2. `bash -n .claude/hooks/warn-review-marker-write.sh` → clean

Low practical exploitability (needs a concurrent same-kind active review in flight), pre-existing since #962, flagged by Hakim during the #975 review as a follow-up — worth closing so the gate's per-PR binding literally holds on the indirect path.

Closes #977

---

## Glossary

| Term | Definition |
|------|------------|
| authorised session marker | `.claude/session/active-reviewer`, one line `<owner>/<repo>#<pr>:<kind>`, that the orchestrator writes just before spawning the sanctioned reviewer to authorise its marker write. |
| indirect path | The #962 detection branch that catches a marker write whose target is built via a shell variable/function instead of a literal filename. |
| per-PR binding | The property that a review-approval marker is valid only for the specific PR it names — the thing this fix restores on the indirect path. |
| fail closed | Deny by default when a needed fact (here, the target PR) can't be resolved, rather than skipping the check and allowing. |

### DIFF
--- a/.claude/hooks/tests/test_warn_review_marker_write.sh
+++ b/.claude/hooks/tests/test_warn_review_marker_write.sh
@@ -60,6 +60,16 @@
 #                 (without the #974 fix, both sides of the kind comparison
 #                 resolve to "" and the write would be incorrectly ALLOWED)
 #
+# #977 test (per-PR binding on the indirect path — see the hook's #977 guard):
+#   (27) Bash   → indirect write whose ROLE resolves (rex) but whose PR does
+#                 NOT (variable PR arg), + an active-reviewer marker for a
+#                 DIFFERENT specific PR                            → BLOCKED, exit 2
+#                 (without #977, TARGET_PR="" skips the PR check and the write
+#                 matches the wrong-PR active-reviewer marker → wrongly ALLOWED)
+#   (28) Bash   → indirect write where BOTH role and PR resolve (literal PR,
+#                 the sanctioned idiom), matching active-reviewer marker
+#                                                                  → ALLOWED, exit 0 (no regression)
+#
 # Exit 0 if all cases pass; 1 on failure.
 
 set -u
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #992

**Commit**: `f1c67f69f909f2054b9bec6a81efcdb35354fdef`

**Verdict: APPROVED**

Trust-chain hardening (#977) for `warn-review-marker-write.sh`. Reviewed from the API diff.

### Correctness
The new `[ -n "$TARGET_PR" ] || return 1` guard is placed correctly — after the kind check, before the PR-equality test. The old `[ -n "$TARGET_PR" ] && [ "$c_pr" != "$TARGET_PR" ]` short-circuited to *skip* on empty TARGET_PR; combined with the also-skipped repo check on the #962 indirect path, a resolved-role write fell through to allow, matching any active-reviewer marker of that kind regardless of PR. The guard fails closed. Hole closed, per-PR binding restored.

### Regression risk (the key question)
No regression. The sanctioned reviewer idiom passes a **literal** PR number to `review_marker_path` (harness-substituted before the shell runs), so TARGET_PR always resolves on both the literal-path Write case (from filename digits) and the indirect case (from the literal arg). Empty TARGET_PR only arises from a variable-derived name — the forged/malformed shape that should block. Failing closed costs the legitimate path nothing.

### Tests
- **case27** is a genuine regression test: role resolves (rex), PR unresolved (variable `$PR`), active-reviewer present for a *different* PR → expects BLOCKED/exit 2. On the OLD code this allowed (exit 0); it fails pre-fix and passes post-fix. Correct.
- **case28** pins the no-regression path: literal PR 42 + matching marker → ALLOWED/exit 0.

### PR quality
Glossary present (4 terms). Summary bullets narrative (what + why). No AgDR needed — hardening bug fix, no new technical decision.

---
📌 Reviewed commit: `f1c67f69f909f2054b9bec6a81efcdb35354fdef`

atlas-apex (commented):
## Security Review: PR #992 — Hakim (Security Auditor)

**Commit reviewed:** `f1c67f69f909f2054b9bec6a81efcdb35354fdef`
**Scope:** trust-chain fix (#977) to `warn-review-marker-write.sh` — per-PR binding of the author≠reviewer marker gate on the #962 indirect path.

**Verdict: APPROVED — no blocking security findings.**

### Bypass fully closed
The change replaces `[ -n "$TARGET_PR" ] && [ "$c_pr" != "$TARGET_PR" ] && return 1` (which, on empty `TARGET_PR`, short-circuited to *no return* and fell through to ALLOW — repo check also skipped on the indirect path) with a fail-closed `[ -n "$TARGET_PR" ] || return 1` followed by the equality check. Guard ordering (kind non-empty → kind match → **PR non-empty** → PR equality → repo) places the fail-closed check *before* any match, so an unresolvable target PR can never match a mismatched active-reviewer marker. No residual allow path via empty `TARGET_PR` or empty `TARGET_REPO`.

### Strictly a tightening
For non-empty `TARGET_PR` the old and new forms are logically identical. The only delta is the empty-`TARGET_PR` case: previously ALLOWED → now DENIED (return 1 = block). Deny-more only; nothing newly permitted — no new fail-open.

### Legit-flow safety (availability)
Sanctioned reviewers call `review_marker_path` with a literal PR digit → qualified marker filename → `_extract_marker_pr` resolves → `TARGET_PR` non-empty → guard passes. No sanctioned reviewer path yields an empty `TARGET_PR`, so the fail-closed guard cannot block a genuinely-resolvable review. case28 confirms ALLOWED (exit 0); case27 confirms the wrong-PR indirect write is BLOCKED (exit 2).

### No other trust-chain regression
Banner text, exit codes, and the ceo-advisory branch are untouched (out of diff). New comment mirrors the existing empty-kind guard shape; tests add both block and no-regression coverage.

📌 Reviewed commit: `f1c67f69f909f2054b9bec6a81efcdb35354fdef`

me2resh (approved):

PR #994, opened 2026-07-23
feat(#993): right-size-ceremony rule — overengineering guard for the SDLC gates

atlas-apex:
## Summary

- **Adds a "right-size ceremony" rule** (`.claude/rules/right-size-ceremony.md`) — the missing *Lean floor* for the SDLC gates. Today every change that isn't a detected high-blast class (trust-chain, migration, design, UI, auth) is treated as Standard-full-ceremony, so a three-line `CODE_OF_CONDUCT.md` PR spawns a full Rex sub-agent and runs the same role chain as real code. This rule classifies a change by **path class + blast radius + behavior surface** into **Lean** (docs/config-text, small, reversible → inline check + one approval, *no review sub-agent*), **Standard** (ordinary code → Rex + nod), or **Heavy** (trust-chain / auth / migrations / design / releases → the full chain, unchanged).
- **Two safety rails keep it from failing dangerously** — security/trust-chain/migration **never** goes Lean (that's the one place you want the chain), and **ambiguity rounds up** a tier. So the failure mode is "occasionally too much review on a borderline case," never "too little on a risky one."
- **No new hook, by design** — a *blocking* tier-classifier would re-introduce the exact rigidity this fixes, and the `Agent`-tool spawn boundary has no `PreToolUse` hook anyway (AgDR-0056). So the Heavy classes keep their existing hard gates (all unchanged) and the Lean floor is agent self-discipline, watched by the existing `enforce-budget.sh` token meter. Same enforcement shape as `plan-mode` / `parallel-work` / `loop-mode`.
- **AgDR-0107** records the design call (advisory-not-blocking, the tiering, the two rails, the honest hook limitation); **CLAUDE.md** wires the rule into the Quality Rules list and bumps the rules-layer count 18 → 19.

Motivated by direct operator feedback that the framework had grown over-tightened and bureaucratic — gatekeeper-chain latency and heavy token burn from uniform ceremony on trivial changes.

## Testing

1. `.claude/rules/right-size-ceremony.md` renders and reads consistently with the sibling self-discipline rules.
2. `CLAUDE.md` imports the rule (`@.claude/rules/right-size-ceremony.md`) and the count reads 19.
3. AgDR-0107 has valid frontmatter + Y-statement + options table + decision.

Closes #993

---

## Glossary

| Term | Definition |
|------|------------|
| Ceremony | The review/approval machinery a change goes through — review sub-agents, role handoffs, gates. |
[...]

PR #996, opened 2026-07-23
fix(#995): role-trigger banners fire once/session + converge, not per-edit

### LINKED ISSUE #995: [Bug] Role-trigger convergence — banner re-fires every edit, driving infinite role-handover loops
## Given / When / Then

- **Given** an agent driving multi-step work in an ApexYard fork,
- **When** it edits across layers in one unit of work (backend code, then an AgDR, then a hook / CI file),
- **Then** `detect-role-trigger.sh` re-fires an *imperative* role-switch / sub-agent-spawn banner on **every** `Edit`/`Write`/`Bash`, the agent obeys and switches persona or opens new ceremony each time, and it ping-pongs (AgDR → backend → devops → repeat) with **zero progress.**

Reported by a real user: *"the new version is a lot worse... very slow, tons of overengineering, very bureaucratic... it just keeps handing over from one role to the other without any progress .. infinite loops of AgDRs then backend then devops and zero progress."* Matches the maintainer's own token-burn / handover-latency feedback that motivated right-size-ceremony (#993/#994).

## Repro

1. In an ApexYard fork, give the agent a real multi-step build task that touches more than one layer (e.g. a backend change that also needs an AgDR and a hook/CI edit).
2. Observe the `PreToolUse` role-trigger banners: each `Edit`/`Write`/`Bash` to a matching path re-emits an imperative "SPAWN / adopt persona" banner (confirm the de-dupe is per-invocation: `detect-role-trigger.sh` `fired` is a function-local var).
3. The agent switches persona / opens ceremony on each banner instead of finishing the unit, cycling AgDR → backend → devops with no completed work.

Minimal mechanism check (no agent needed): run `detect-role-trigger.sh` twice with the same `docs/agdr/x.md` write payload — it emits the same banner both times (no cross-call suppression).

**Severity:** high — degrades the core product experience (the primary user complaint), not a corner case.

## Root cause (`/debug`, 2026-07-24)

The `fired` de-dupe in `detect_prompt_triggers` is a **function-local variable reset on every hook invocation**, so the diff/path triggers (`PreToolUse` on Edit/Write/Bash) have **zero cross-call de-dupe** — each edit is a fresh invocation and re-emits the banner. The banners are **imperative** (`emit_banner:135` "SPAWN the sub-agent via the Agent tool"; `:138` "adopt the persona IN-THREAD") with **no convergence guard** — nothing ever tells the agent "you are mid-task, finish the current unit before switching." The framework has a strong "activate the right role on every action" force and no counter-force toward "make progress." Compounded by `agdr-decisions.md`'s per-decision HARD-STOP (writing the AgDR re-fires Tech Lead + Solution Architect).

## Scope

- `.claude/hooks/detect-role-trigger.sh` `emit_banner`: **session-scoped de-dupe** so each role's banner fires **at most once per session** (self-cleaning session-id-keyed marker under `.claude/session/role-fired/`; no new SessionStart hook, no `settings.json` change; missing session id degrades to old always-fire). **Reword** the banner from imperative to **advisory + convergence**: "advisory, once per session... if you are MID-TASK and making progress, do NOT switch persona or open new ceremony — finish the current unit; roles hand off at phase boundaries, not per edit."
- Update `.claude/hooks/tests/test_detect_role_trigger.sh` for the new language + once-per-session behaviour.

## Out of scope (follow-up)

- Narrow `.claude/rules/agdr-decisions.md` from "before ANY technical decision" to *material / architectural* decisions (aligns with right-size-ceremony). Separate PR to keep this one focused (`/debug` discipline: smallest change).

## Acceptance Criteria

- [ ] A role's trigger banner fires **at most once per session** — a second matching edit emits no banner.
- [ ] Banner text is advisory and explicitly tells the agent **not to switch persona / open ceremony mid-task** (hand off at phase boundaries, not per edit).
- [ ] No merge-gate / credential / marker logic touched (this is an advisory behaviour hook — Standard-tier, not the security trust chain).
- [ ] `test_detect_role_trigger.sh` passes with the new behaviour.

## Glossary

| Term | Definition |
|------|------------|
| Role-trigger banner | The advisory stderr nudge `detect-role-trigger.sh` prints when a diff/prompt matches a role's activation condition. |
| Convergence guard | A rule/mechanism that says "finish the current unit of work before re-activating a role or opening new ceremony" — the counter-force this bug is missing. |
| Session-scoped de-dupe | Firing each role's banner at most once per Claude Code session, keyed on `CLAUDE_CODE_SESSION_ID`. |

Refs #993.

atlas-apex:
## Summary
- **Kills the "infinite role-handover loop" a real user reported** — driving multi-layer work, the agent kept ping-ponging AgDR → backend → devops with zero progress. Root cause (`/debug`): `detect-role-trigger.sh` re-emitted an *imperative* role-activation banner on **every** matching Edit/Write/Bash, and the agent obeyed each one and switched persona mid-task. The framework had a strong "activate the right role on every action" force and **no** counter-force toward "make progress."
- **Session-scoped, once-per-role de-dupe in `emit_banner`** — each role's banner now surfaces at most once per `CLAUDE_CODE_SESSION_ID`, so a single unit of work that touches several files no longer re-triggers a handover per file. The marker lives under `.claude/session/role-fired/` and is **self-cleaning** (each call sweeps other sessions' markers) — so there's **no new SessionStart hook and no `settings.json` change**, keeping this a Standard-tier advisory hook rather than a trust-chain edit. Fail-open: no session id / no repo root / unwritable dir all degrade to the old always-fire behaviour, because a banner is advisory and over-surfacing beats silently swallowing role activation on a broken environment.
- **Both banners reworded from imperative to advisory + convergence** — they still name the role that owns the work and offer the spawn / in-thread-lens, but now explicitly add: *"if you are MID-TASK and making progress, do NOT switch persona or open new ceremony on this edit — finish the current unit first; roles hand off at phase boundaries, not per file."* This is the missing counter-force; it's what actually breaks the loop behaviourally even when a banner does fire.
- **Deliberately the smallest change that reduces process** — the reported disease is over-ceremony, so the fix removes push (imperative → advisory) and removes repetition (per-edit → once/session) rather than adding any new gate. No merge-gate, credential, or marker logic is touched. A follow-up (separate PR) will narrow `agdr-decisions.md`'s per-decision HARD-STOP, which compounds the same churn.

## Testing
1. `bash .claude/hooks/tests/test_detect_role_trigger.sh` — **50/50 pass** (44 existing + 6 new).
2. New section (7) proves the de-dupe in a throwaway git repo: first same-role edit fires, a second file mapping to the *same* role is silent, a *different* role still fires (per-role, not a global mute), the same role in a *new* session fires again (session-scoped), and with no session id repeated edits both fire (fail-open).
3. Existing per-case tests run with `CLAUDE_CODE_SESSION_ID` unset (always-fire mode) so they stay deterministic regardless of the ambient session; class-aware assertions updated for the advisory wording, plus two new cases asserting the convergence guard is present on both banner shapes.

Fixes #995

---

## Glossary
| Term | Definition |
|------|------------|
| Role-trigger banner | The advisory stderr note `detect-role-trigger.sh` prints when a diff/prompt matches a role's activation condition. Advisory only — never blocks the tool call. |
| Convergence guard | The new "finish the current unit before switching roles / opening ceremony" instruction — the counter-force to the framework's "activate the right role" push. |
| Session-scoped de-dupe | Firing each role's banner at most once per Claude Code session, keyed on `CLAUDE_CODE_SESSION_ID`, via a self-cleaning marker under `.claude/session/role-fired/`. |
| Fail-open | When the de-dupe can't record state (no session id / no repo root / unwritable dir) it degrades to always-firing, since a swallowed advisory is worse than a repeated one. |
| Standard-tier (vs trust-chain) | Per right-size-ceremony (#994): this hook is an advisory banner, not a merge gate / marker lib / `settings.json` matcher, so it doesn't require the security-auditor trust-chain treatment. |

https://claude.ai/code/session_015KGiYPZB3US1LrQ9GS9vXe

### DIFF
--- a/.claude/hooks/detect-role-trigger.sh
+++ b/.claude/hooks/detect-role-trigger.sh
@@ -103,6 +103,59 @@ agent_slug_for() {
 }
 
 # ---------------------------------------------------------------------------
+# Session-scoped, once-per-role de-dupe (me2resh/apexyard#995).
+#
+# The path/label triggers fire on EVERY matching Edit/Write/Bash. Before
+# this guard a single multi-layer unit of work (backend code → an AgDR →
+# a CI file) re-emitted a role-activation banner on each edit, and the
+# agent, reading each banner as an instruction, kept switching persona /
+# opening ceremony mid-task instead of finishing the unit — the "infinite
+# role-handover loop, zero progress" a real user reported (#995).
+#
+# This makes each role's banner fire AT MOST ONCE per Claude Code session.
+# The marker is keyed on CLAUDE_CODE_SESSION_ID, so a new session starts
+# fresh with NO new SessionStart hook and NO settings.json change (keeps
+# this a Standard-tier advisory hook, not a trust-chain edit). It is
+# self-cleaning: each call sweeps markers left by other sessions, so the
+# dir can't grow without bound.
+#
+# Fail-open by design: no session id, no REPO_ROOT, or an unwritable
+# session dir all degrade to the pre-#995 always-fire behaviour. A banner
+# is advisory — over-surfacing on a broken environment is strictly safer
+# than silently swallowing role activation.
+#
+# Returns 0 = already fired this session (SKIP the banner);
+#         1 = first time / fail-open (FIRE the banner).
+# ---------------------------------------------------------------------------
+role_banner_already_fired() {
+  local file="$1"
+  local sid="${CLAUDE_CODE_SESSION_ID:-}"
+  [ -z "$sid" ] && return 1          # no session id → always fire (fail-open)
+  [ -z "$REPO_ROOT" ] && return 1    # no repo root → always fire (fail-open)
+
+  local dir="$REPO_ROOT/.claude/session/role-fired"
+  mkdir -p "$dir" 2>/dev/null || return 1   # can't record → always fire
+
+  # Self-clean: drop markers from any prior/other session so the dir
+  # doesn't accumulate across sessions (no SessionStart sweep needed).
+  local f
+  for f in "$dir"/*; do
+    [ -e "$f" ] || continue
+    case "${f##*/}" in
+      "${sid}__"*) : ;;              # keep this session's markers
+      *) rm -f "$f" 2>/dev/null ;;   # sweep other sessions'
+    esac
+  done
+
+  local slug marker
+  slug=$(printf '%s' "$file" | tr '/ ' '__')
+  marker="$dir/${sid}__${slug}"
+  [ -e "$marker" ] && return 0        # already fired this session → skip
+  : > "$marker" 2>/dev/null || return 1  # can't write → fail-open, fire
+  return 1                            # first time → fire
+}
+
+# ---------------------------------------------------------------------------
 # Helper: emit one class-aware banner line. Multiple triggers in a single
 # call can fire multiple banners.
 #
@@ -127,15 +180,29 @@ agent_slug_for() {
 # ---------------------------------------------------------------------------
 emit_banner() {
   local role="$1" file="$2" reason="$3"
+
+  # Session-scoped de-dupe (#995): a role's banner surfaces at most once
+  # per session, so a multi-edit unit of work doesn't re-trigger a
+  # handover on every file. Advisory — the underlying tool call always
+  # proceeds regardless.
+  if role_banner_already_fired "$file"; then
+    return 0
+  fi
+
   local class agent_slug
   class=$(lookup_role_class "$file")
   agent_slug=$(agent_slug_for "$file")
 
+  # Both banners are ADVISORY and carry a convergence guard (#995): they
+  # name the role that OWNS this kind of work, but explicitly tell the
+  # agent NOT to switch persona / open new ceremony mid-task. Roles hand
+  # off at phase boundaries, not on every edit — the counter-force that
+  # keeps "make progress by default" the default.
   if [ "$class" = "isolated-work-class" ]; then
-    printf 'ROLE TRIGGER: %s activates per .claude/rules/role-triggers.md (%s). Isolated-work-class — SPAWN the sub-agent via the Agent tool with subagent_type: %s. Canonical role at %s, agent wrapper at .claude/agents/%s.md. Per AgDR-0050 § Axis 6.\n' \
+    printf 'ROLE TRIGGER: %s owns this kind of work per .claude/rules/role-triggers.md (%s) [advisory, once/session]. If you are STARTING a discrete, isolatable task, you MAY spawn it via the Agent tool with subagent_type: %s (canonical role at %s, wrapper at .claude/agents/%s.md). But if you are MID-TASK and making progress, do NOT switch persona or open new ceremony on this edit — finish the current unit first; roles hand off at phase boundaries, not per file. Per AgDR-0050 § Axis 6 + #995.\n' \
       "$role" "$reason" "$agent_slug" "$file" "$agent_slug" >&2
   else
-    printf 'ROLE TRIGGER: %s activates per .claude/rules/role-triggers.md (%s). In-flow-class — adopt the persona IN-THREAD. Read %s before continuing. Per AgDR-0050 § Axis 6.\n' \
+    printf 'ROLE TRIGGER: %s owns this kind of work per .claude/rules/role-triggers.md (%s) [advisory, once/session]. If helpful, read %s to adopt its lens in-thread. But if you are MID-TASK and making progress, do NOT switch persona or open new ceremony on this edit — finish the current unit first; roles hand off at phase boundaries, not per file. Per AgDR-0050 § Axis 6 + #995.\n' \
       "$role" "$reason" "$file" >&2
   fi
 }
--- a/.claude/hooks/tests/test_detect_role_trigger.sh
+++ b/.claude/hooks/tests/test_detect_role_trigger.sh
@@ -432,7 +432,9 @@ dedup_edit() {
   local sid="$1" path="$2" input
   input=$(jq -nc --arg p "$path" \
     '{hook_event_name:"PreToolUse", tool_name:"Edit", tool_input:{file_path:$p}}')
-  ( printf '%s' "$input" | ( cd "$DEDUP_REPO" && CLAUDE_CODE_SESSION_ID="$sid" bash "$HOOK" ) ) 2>&1 >/dev/null
+  # Capture stderr (the banner), discard stdout. The brace-group form keeps the
+  # redirection order unambiguous (shellcheck SC2069).
+  { printf '%s' "$input" | ( cd "$DEDUP_REPO" && CLAUDE_CODE_SESSION_ID="$sid" bash "$HOOK" ) >/dev/null; } 2>&1
 }
 
 assert_dedup() {
--- a/.claude/hooks/tests/test_detect_role_trigger.sh
+++ b/.claude/hooks/tests/test_detect_role_trigger.sh
@@ -16,6 +16,13 @@
 
 set -u
 
+# The session-scoped de-dupe (#995) keys on CLAUDE_CODE_SESSION_ID. Unset it
+# for the per-case tests below so each invocation is in the fail-open
+# always-fire mode — deterministic and independent of whatever ambient
+# session happens to be running this test. Section (7) sets its own id to
+# exercise the de-dupe explicitly.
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #996

**Commit**: `a7c0a92aaeb23b3dea3cc8164c65847a363d18b0`

### Summary
Fixes the real-user-reported "infinite role-handover loop" (#995): `detect-role-trigger.sh` re-emitted an *imperative* role-activation banner on every matching Edit/Write/Bash, and the agent obeyed each one — ping-ponging personas mid-task with zero progress. The fix (a) adds a session-scoped, once-per-role de-dupe in `emit_banner` keyed on `CLAUDE_CODE_SESSION_ID` with a self-cleaning file under `.claude/session/role-fired/`, and (b) rewords both banners from imperative to advisory plus a convergence guard ("finish the current unit before switching persona"). Tests grow 44 to 50.

### Checklist Results
- Security:                   Pass (see trust-chain note below)
- Summary Bullet Narrative:   Pass (bullets are narrative, not label-only)
- Technical Decisions (AgDR): N/A — bug-fix implementation detail; the self-cleaning-file vs SessionStart-hook trade-off is thoroughly justified in the PR body
- Adopter Handbooks:          N/A — no architecture/general/language handbook triggers on a shell-only advisory-hook diff
- Fallow Static Analysis:     N/A — not a JS/TS diff

### Issues Found
None blocking. The core mechanics are correct on every point in the review focus:

- **Fail-open is correct and lands in the safe direction.** No session id returns 1 before touching disk (so the fail-open path never even creates the role-fired dir); no `REPO_ROOT` returns 1 (`REPO_ROOT` is always *defined* via `|| true`, so `[ -z ]` is `set -u`-safe); the `mkdir -p` guard and the empty-write guard both `return 1` on an unwritable dir. Over-surfacing on a broken environment beats swallowing role activation.
- **The de-dupe is effectively per-role, which is what the loop scenario needs.** `emit_banner`'s second argument is always the *role file* path (e.g. `roles/security/security-auditor.md`), never the edited file — so two different edited files that map to the same role produce the same slug and the same recorded file, de-duping correctly. Test 7b (`login.ts` then `jwt.ts`, both Security Auditor, expect silent) proves exactly this.
- **The self-clean sweep is safe.** The glob is scoped to the dedicated `.claude/session/role-fired/` dir (created only by this hook), and only removes entries that do not start with the current `${sid}__` prefix — it cannot touch unrelated files. On an empty dir the unmatched `"$dir"/*` yields the literal glob and the `[ -e "$f" ] || continue` guard skips it (nullglob is not set — verified this is the intended safety net).
- **Slug derivation `tr '/ ' '__'`** maps two characters (`/`, space) to two replacement characters correctly; the file argument is always a hard-coded safe role path, so there is no traversal or injection surface.
- **Test-helper capture ordering is correct.** The `( ... ) 2>&1 >/dev/null` form swaps fd2 onto the command-substitution pipe *before* redirecting fd1 to /dev/null, so the captured value is the banner (emitted on stderr via `>&2`) and stdout is discarded. The classic capture-stderr-not-stdout idiom, applied right.
- **Test isolation is clean.** Section 7 runs in a throwaway `mktemp -d` git repo, so `REPO_ROOT` and the role-fired dir resolve there and `rm -rf` cleans up — the real session dir is never polluted. `unset CLAUDE_CODE_SESSION_ID` at the top keeps the pre-existing per-case tests deterministic in always-fire mode.
- **Section-7 coverage proves the four properties it needs to:** fire-once-then-silent (7a to 7b), per-role independence (7c), session-scoping (7d), fail-open (7e). Plus 4h asserts the convergence guard on both banner shapes.

### Handbook Findings
None — no handbook triggers on this diff.

### Suggestions (nits — none blocking)
- **nit (trust-chain / right-size):** the edited file lives under `.claude/hooks/`, so the framework's own #777 trigger and `right-size-ceremony.md` rail 1 ("anything under `.claude/hooks/**` rounds up to Heavy") would *nominally* call for a Security Auditor pass. Substantively this is fine — the hook is advisory, exit-0-always, and touches **no** merge-gate / credential / marker-lib / `settings.json` matcher logic, so from the code-review lens it is genuinely Standard-tier. Flagging so a human can give it a quick security-reviewer nod if they want to honour rail 1 to the letter; not a code-review blocker.
- **nit (concurrency):** two *concurrent* sessions sharing one repo root would mutually sweep each other's recorded files, degrading the de-dupe back toward always-fire. That is the safe (fail-open) direction and concurrent same-repo sessions are rare, so it is acceptable — worth a one-line comment near the sweep noting the trade-off.
- **nit (prompted activation):** an explicit `act as the <role>` is now also de-duped once/session, so a deliberate re-activation of the same role later in the session is silenced. The user's request stands on its own (the banner is only a reminder), so this is acceptable, but it is a slight behavioural change to a first-class activation path worth being aware of.
- **nit (test hygiene):** `( cd "$DEDUP_REPO" && git init -q )` has no `|| exit` guard (per `isolated-builds.md`); harmless in this throwaway-test context since a failed `cd ""` short-circuits the `&&`, but a `|| exit 1` would match the framework convention.

### Verdict
**APPROVED**

The de-dupe is correct, fails open in the safe direction, is well-isolated in tests, and touches no trust-chain gate / credential / marker / settings logic. The rewording adds the missing convergence counter-force. Nits are advisory only.

---
Reviewed commit: `a7c0a92aaeb23b3dea3cc8164c65847a363d18b0`

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #996 (re-review)

**Commit**: `d693f4665000b5752d1b9dfad2bd51daab191902`
**Prior review**: APPROVED at `a7c0a92` — this is a delta re-review of the one follow-up commit.

### Delta since last approval

One commit, one file, +3/-1, test-only:
`.claude/hooks/tests/test_detect_role_trigger.sh` — the section-7 `dedup_edit` helper's
redirection switched from `( ... ) 2>&1 >/dev/null` to `{ ... >/dev/null; } 2>&1` to clear
CI shellcheck SC2069.

### Verification

**1. Semantically equivalent — confirmed.** Both forms yield "stderr captured, stdout discarded":

- Old: command substitution sets the subshell's fd1 to the capture pipe; redirections apply
  left-to-right, so `2>&1` points fd2 at the capture pipe *first*, then `>/dev/null` repoints
  fd1. Correct, just visually ambiguous (which is what SC2069 flags).
- New: the brace group's `2>&1` binds fd2 to the capture pipe, then the inner `>/dev/null`
  binds to the last element of the pipeline, sending only its stdout to `/dev/null`.
  fd2 remains on the capture pipe.

`printf`'s stdout is bound by the pipe in both forms, so the relocated `>/dev/null` changes
nothing there. Net file-descriptor state is identical.

**2. The section-7 assertions still test what they claim.** `assert_dedup` discriminates
fire-vs-silent purely on whether the captured string is non-empty, and the banner is the only
thing on stderr — `detect-role-trigger.sh` emits every banner with `>&2` (lines 203, 206, 479).
So 7a–7e still exercise the real once-per-role-per-session behaviour: first fire, same-role
silence, different-role fire, new-session re-fire, and the no-session-id fail-open pair.

**3. No production logic changed.** `git compare a7c0a92...d693f46` shows exactly one file,
and it is under `tests/`. `detect-role-trigger.sh` itself is byte-identical to the version
approved at `a7c0a92`.

**4. No regressions vs. the prior review.** Nothing else in the diff moved; every finding and
sign-off from the `a7c0a92` review still stands.

### Note (non-blocking)

Section 6 still uses the original `2>&1 >/dev/null` form (line ~407). That is not an
inconsistency to fix: it sits *directly inside* a `got=$(...)` command substitution, where
ShellCheck recognises the capture-stderr idiom and suppresses SC2069. The helper needed the
brace-group form precisely because its capture happens at the call site, out of ShellCheck's
view. Leaving section 6 as-is is correct.

### Checklist Results
- Architecture & Design:     Pass (unchanged)
- Testing:                   Pass — test semantics preserved; suite reported 50/50
- Security:                  Pass — no trust-chain logic changed (test file only)
- PR Description & Glossary: Pass (unchanged from prior review)
- Technical Decisions (AgDR):N/A — no decision introduced

### Issues Found
None.

### Verdict
**APPROVED**

CI note: `shellcheck .claude/hooks` is green at this SHA (the warning this commit targeted);
`hook test suite` was still running at review time — merge should wait for it.

---
Reviewed commit: `d693f4665000b5752d1b9dfad2bd51daab191902`

PR #998, opened 2026-07-23
chore(#997): narrow AgDR rule from "any technical decision" to material ones

### LINKED ISSUE #997: [Chore] Narrow agdr-decisions from "any technical decision" to material ones
## Driver

The third leg of the over-process fix that #993/#994 (right-size-ceremony) and #995/#996 (role-trigger convergence) started. `.claude/rules/agdr-decisions.md` opens with **"HARD STOP: before making any technical decision, run `/decide`"** and lists triggers including *"Pick an implementation approach"*, *"I'll implement the X pattern…"*, and *"Am I about to write code that introduces a new pattern? → STOP, /decide"*. Taken literally that covers essentially every implementation choice an engineer makes.

This compounds the loop #996 just fixed: writing an AgDR lands a file under `docs/agdr/**`, which re-fires the Tech Lead trigger (and, for migration AgDRs, the Solution Architect trigger too). The reported symptom was exactly this shape — *"infinite loops of AgDRs then backend then devops and zero progress."*

**Key finding: the mechanical layer is already correctly scoped; the prose is the outlier.** `require-agdr-for-arch-changes.sh` gates only `infrastructure/**`, `*.tf` / `*.tfvars`, `Dockerfile*` / `docker-compose*`, and `.github/workflows/**` — and says so in its own header:

> "Deliberately NARROW. Dependency bumps (package.json, go.mod, etc.) and API schema changes are NOT included — too noisy, not all changes need an AgDR."

`CLAUDE.md` is likewise already narrower, saying *"significant technical decisions"*. Only the rule file demands an AgDR for **any** decision. A rule far broader than anything anyone would actually enforce is the definition of ceremony that costs more than it protects.

## Scope

Prose-only edit to `.claude/rules/agdr-decisions.md`:

- Replace the blanket "any technical decision" HARD STOP with an explicit **materiality threshold**.
- **Keep the HARD STOP** for material / architectural decisions: a new dependency, library, framework, or technology; a new service or bounded context; a new external integration; a data-model / schema change; a security-relevant control; CI/CD, release, or infrastructure design; a pattern adopted repo-wide or cross-cutting.
- **Explicitly exempt** routine implementation choices reversible inside one PR: local naming, whether to extract a helper, control flow, test structure, calling an API from a dependency already in the manifest, one-off refactor or formatting choices.
- Cross-reference right-size-ceremony (#993/#994) and preserve both of its safety rails: security / trust-chain / migration never goes Lean, and ambiguity rounds **up** a tier.
- State the prose↔hook relationship honestly: the hooks are the narrow mechanical backstop, and the prose should not be broader than what is actually enforced.

**Explicitly out of scope:** no hook, `settings.json`, gate, or marker logic changes. No mechanical behaviour change — this narrows guidance only.

## Acceptance Criteria

- [ ] The rule no longer demands an AgDR for every implementation choice; a materiality threshold is stated explicitly.
- [ ] A concrete "does NOT need an AgDR" list sits alongside the "does need one" list.
- [ ] The rule cross-references right-size-ceremony and preserves both safety rails verbatim in intent.
- [ ] No hook, `settings.json`, or gate logic is changed (diff touches rule prose only).
- [ ] `CLAUDE.md`'s description stays accurate and consistent with the narrowed rule.

## Risks / Dependencies

Risk: under-recording a genuinely material decision. Mitigated by rail 2 (ambiguity rounds up), the unchanged mechanical hooks, and Rex's existing AgDR-linkage check on architecture PRs. Depends on nothing; #996 is already merged to `dev`.

## Glossary

| Term | Definition |
|------|------------|
| AgDR | Agent Decision Record — the lightweight decision log this framework writes to `docs/agdr/AgDR-NNNN-*.md`, capturing options considered and why one won. |
| Materiality threshold | The line separating decisions worth a durable record (architectural, hard to reverse, cross-cutting) from routine implementation choices reversible within a single PR. |
| Prose↔hook gap | When a written rule demands far more than its mechanical enforcement actually checks — the state this ticket closes. |
| Safety rails (right-size-ceremony) | (1) security / trust-chain / migration never takes the Lean path; (2) ambiguity rounds up a tier. |

Refs #995.

atlas-apex:
## Summary

- **Closes the last leg of the over-process fix** that #994 (right-size-ceremony) and #996 (role-trigger convergence) started. `agdr-decisions.md` opened with *"HARD STOP: before making **any** technical decision"* and listed triggers like *"Pick an implementation approach"* and *"Am I about to write code that introduces a new pattern?"* — a bar that covers essentially every implementation choice a person makes.
- **This rule was actively feeding the loop we just fixed.** Every AgDR written lands a file under `docs/agdr/**`, which re-fires the Tech Lead activation trigger (and, for migration AgDRs, the Solution Architect trigger too). So an over-broad AgDR rule *manufactures* the AgDR → backend → devops churn the reporting user described. Narrowing it is what stops #996's fix from being undermined by the rule sitting next to it.
- **The load-bearing finding: the prose demanded far more than anything actually enforced.** The two hooks are bounded and fire at different moments — `require-agdr-for-arch-changes.sh` at commit time on `*.tf`/`*.tfvars`, `Dockerfile*`, `docker-compose*`, `.github/workflows/**`; `require-agdr-for-arch-pr.sh` at PR-create time on `**/domain/**`, `**/infrastructure/**`, `**/migrations/**`, `infrastructure/**`, `template.yaml`, IaC globs, **plus dependency *additions*** to `package.json` / `pyproject.toml` / `Cargo.toml` / `go.mod` / `Gemfile`. `CLAUDE.md` already said *"significant"*. Only the rule file said **any**. A rule far broader than anything anyone would enforce doesn't produce more records — it produces ignored prose and burns the ceremony budget. This aligns the prose to reality, and documents both hooks accurately (including that the commit-time hook *deliberately omits* a bare `infrastructure/` pattern because it false-positived on `docs/infrastructure/notes.md` — a drift that survives only in a stale header comment).
- **What the rule now says** — an explicit **materiality threshold** (new dependency/technology, new service or integration, data-model or schema change, security-relevant control, CI/CD-release-infra design, repo-wide pattern, anything hard to reverse) sitting next to a concrete **"does NOT need an AgDR"** list (local naming, extracting a helper, control flow, test structure, calling an API from a dependency already in the manifest). The old *"could you explain the trade-off?"* heuristic is demoted to a **writing aid** rather than a trigger, because almost every line of code has an unexplained alternative — which is precisely why it over-fired.
- **Both safety rails survive intact**: security / trust chain / migrations are **never** exempt regardless of diff size, and genuine ambiguity **rounds up** to writing the record. The tolerated failure stays "an occasional unnecessary AgDR", never "a silently unrecorded architectural call."
- **Rex's blocking check was narrowed to match** — `code-reviewer.md` § 7, its enforcement rule 2, and Rules #7 all said *"a technical decision"* unqualified, which would have kept flagging PRs at the old bar and quietly cancelled this change. Its detection table was already material-shaped and is untouched.
- **AgDR-0108 records the policy call.** Applying the new rule to itself: changing the portfolio's decision-*recording* policy is cross-cutting, and AgDR-0107 is scoped to review *tiering* — a distinct call, so it gets its own record rather than inheriting one.

## Testing

1. **No mechanical behaviour changed** — the diff is prose plus one new AgDR. No hook, `settings.json`, gate, or marker logic.
2. `npx markdownlint-cli2` — **0 issues**; CI markdownlint, hook test suite, Analyze, and Verify Ticket ID all green.
3. Hook trigger sets in the new table were **re-derived from source** (`ARCH_GLOBS` and `TRIGGER_PATHS` / `TRIGGER_DEP_FILES`), not from header comments — the first revision of this PR was misled by a stale comment, which is exactly why the rule now documents the drift.
4. All intra-repo links verified to resolve; `[995]` / `[997]` reference-style link definitions added to `docs/rule-audit.md` per that file's convention.

**Deliberately deferred to #999** (Heavy path — `.claude/hooks/**` is rail-1 trust chain, so it needs the Security Auditor, not a tack-on to a prose PR): `require-agdr-for-arch-pr.sh` still prints the old "any technical decision" wording in its agent-facing block message, plus two dangling citations and the stale header comment that caused the detour above.

Fixes #997

---

## Glossary

| Term | Definition |
|------|------------|
| AgDR | Agent Decision Record — the lightweight decision log written to `docs/agdr/AgDR-NNNN-*.md`, capturing options considered and why one won. |
| Materiality threshold | The line separating decisions worth a durable record (architectural, hard to reverse, cross-cutting) from routine choices reversible inside a single PR. The core of this change. |
| Prose↔hook gap | When a written rule demands far more than its mechanical enforcement checks. The gap this PR closes. |
| Documented drift | A comment describing behaviour the code no longer has — here, the commit-time hook's `infrastructure/**` mention, recorded as known but still live enough to mislead this PR's first revision. |
| Safety rails | From right-size-ceremony: (1) security / trust-chain / migration never takes the lighter path; (2) ambiguity rounds up a tier. |
| Ceremony budget | The finite tolerance a session has for process steps before they crowd out the work — the resource an over-broad rule spends without buying safety. |

https://claude.ai/code/session_015KGiYPZB3US1LrQ9GS9vXe

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -149,7 +149,7 @@ Examples that **do not** trigger the heuristic:
 
 ### 7. Technical Decisions (AgDR) — ⛔ BLOCKING CHECK
 
-**You MUST detect and enforce AgDR for any technical decisions.**
+**You MUST detect and enforce AgDR for *material* technical decisions** — architectural, hard to reverse, or cross-cutting. Routine implementation choices that are reversible inside this PR (local naming, extracting a helper, control flow, test structure, using an API from a dependency already in the manifest) do **NOT** need an AgDR and must not be flagged. See `.claude/rules/agdr-decisions.md` § "The threshold" for the full line, including the two non-negotiable rails: security / trust-chain / migration decisions are always material, and genuine ambiguity rounds **up**.
 
 #### How to detect technical decisions in code
 
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -169,7 +169,7 @@ Scan the diff for these patterns:
 #### Enforcement rules
 
 1. **Check if AgDR exists** — look for `AgDR` or `agdr` links in the PR description
-2. **If a decision is detected but NO AgDR is linked** → **REQUEST CHANGES** with this template:
+2. **If a *material* decision is detected but NO AgDR is linked** → **REQUEST CHANGES** with this template. (Material per `.claude/rules/agdr-decisions.md` § "The threshold" — do NOT request changes for a routine choice reversible inside this PR.)
 
 ```markdown
 ## ⛔ AgDR Required
@@ -823,7 +823,7 @@ Report the failure in plain text with the exact command the caller needs to run.
 4. **Don't nitpick style** — that's what linters are for
 5. **First review** — a human approver does the second review before merge
 6. **Glossary is mandatory** — request changes if missing
-7. **AgDR enforcement is BLOCKING** — if you detect a technical decision without an AgDR link:
+7. **AgDR enforcement is BLOCKING** — if you detect a **material** technical decision (architectural, hard to reverse, or cross-cutting per `.claude/rules/agdr-decisions.md` § "The threshold") without an AgDR link:
    - DO NOT approve the PR
    - REQUEST CHANGES with the specific decisions you detected
    - List what needs to be documented
--- a/.claude/rules/agdr-decisions.md
+++ b/.claude/rules/agdr-decisions.md
@@ -1,51 +1,65 @@
-# Technical Decisions — AgDR Required
+# Technical Decisions — AgDR Required for *Material* Decisions
 
-**HARD STOP**: before making any technical decision, run `/decide` and create an Agent Decision Record (AgDR).
+**HARD STOP**: before making a **material** technical decision — one that is architectural, hard to reverse, or cross-cutting — run `/decide` and create an Agent Decision Record (AgDR).
 
-## Trigger Patterns — STOP if you catch yourself doing these
+The word doing the work is **material**. An earlier version of this rule said "before making any technical decision," which read as *every* implementation choice: which helper to extract, how to shape a loop, what to name a module. That bar is unmeetable, and trying to meet it produces the failure mode this framework already named elsewhere — ceremony that costs more than the work it guards. Worse, it compounds: every AgDR written lands a file under `docs/agdr/**`, which re-fires the Tech Lead activation trigger (and, for a migration AgDR, the Solution Architect trigger too), so an over-broad AgDR rule manufactures exactly the role-handover churn [`role-triggers.md`](role-triggers.md) and [`right-size-ceremony.md`](right-size-ceremony.md) exist to prevent. See me2resh/apexyard#995 and #997.
 
-If you are about to:
+This rule is the decision-record sibling of [`right-size-ceremony.md`](right-size-ceremony.md): same instinct (match the ceremony to the blast radius), applied to *recording* rather than *reviewing*.
 
-- Say "I'll use X" or "Let's go with X" → STOP, run `/decide` first
-- Compare "X vs Y" or "X or Y" → STOP, run `/decide` first
-- Choose a library, framework, or tool → STOP, run `/decide` first
-- Pick an implementation approach → STOP, run `/decide` first
-- Make an architectural choice → STOP, run `/decide` first
+## The threshold
 
-## Real-Time Decision Detection
+A decision needs an AgDR when **any** of these is true:
 
-Before writing any code or config, scan your planned response for:
+| Signal | Because |
+|--------|---------|
+| It adds a **new dependency, library, framework, or technology** | Someone inherits the maintenance, licensing, and upgrade cost |
+| It creates a **new service, bounded context, or external integration** | It changes the system's shape and its failure modes |
+| It changes the **data model or schema** | Migrations are expensive and data loss is not reversible |
+| It touches a **security-relevant control** — auth, crypto, secrets, the trust chain | Getting it wrong is the costliest class of wrong |
+| It designs **CI/CD, release, or infrastructure** | It's the machinery everything else ships through |
+| It adopts a **pattern repo-wide or cross-cutting** | Reversing it later means touching everything at once |
+| It is **hard to reverse** — externally visible, shared-state, or expensive to undo | The record is the only cheap artifact left afterwards |
 
-| Pattern in your response | Decision type | Action |
-|--------------------------|---------------|--------|
-| "I'll create a workflow that…"          | CI/CD design | `/decide` |
-| "We'll use GitHub Actions / CircleCI…"  | CI platform | `/decide` |
-| "The keystore will be stored in…"       | Security/infra | `/decide` |
-| Adding dependencies to build files       | Library choice | `/decide` |
-| Creating new architecture (modules, layers) | Architecture | `/decide` |
-| "I'll implement the X pattern…"         | Design pattern | `/decide` |
-| Setting up signing / release / deployment | Release strategy | `/decide` |
+A decision does **not** need an AgDR when it is a routine implementation choice, reversible inside a single PR:
 
-**Test yourself**: if someone asked "why did you choose X over Y?", would you need to explain trade-offs? If yes → `/decide` first.
+- Local naming — variables, functions, files, modules
+- Whether to extract a helper, split a function, or inline something
+- Control flow, error-handling shape, or how a loop is written
+- Test structure, fixture layout, or which assertions to make
+- Calling an API from a dependency **already in the manifest** (adding the dependency is material; using it is not)
+- One-off refactors, formatting, and comment or doc wording
 
-## Self-Check Before the Build Phase
+**The practical test:** *would a competent teammate six months from now be confused, or repeat expensive work, because this reasoning wasn't written down?* If yes, `/decide`. If they'd just read the code and move on, write the code.
+
+The older heuristic — "if someone asked why you chose X over Y, would you need to explain trade-offs?" — is **too sensitive on its own**, because almost every line of code has an unexplained alternative. Use it only *after* the threshold table above already says the decision is material; it helps you write a better AgDR, it does not decide whether you need one.
+
+## The two rails (unchanged, non-negotiable)
+
+Narrowing the rule does not narrow the rails from [`right-size-ceremony.md`](right-size-ceremony.md):
+
+1. **Security, trust chain, and migrations are never exempt.** A decision touching `.claude/hooks/**`, `.claude/settings.json`, auth, crypto, secrets, or a migration is material regardless of how small the diff is. A one-line change to a merge gate is exactly where you want the record.
+2. **Ambiguity rounds up.** If you genuinely can't tell whether a decision is material, write the AgDR. The tolerated failure is an occasional unnecessary record — never a silently unrecorded architectural call.
+
+## Self-check before the Build phase
 
 ```
-[ ] Did I make any technical decisions?     YES → AgDR exists for each?
-[ ] Did I choose between options?           YES → AgDR exists?
-[ ] No AgDR for a decision I made?          → Create it NOW before proceeding
+[ ] Did I make a decision that meets the threshold above?   YES → AgDR exists for each?
+[ ] Is it security / trust-chain / migration related?       YES → material, no exceptions (rail 1)
+[ ] Am I unsure whether it's material?                      → round UP, write it (rail 2)
+[ ] Material decision made with no AgDR?                    → create it NOW before proceeding
 ```
 
-## Self-Check During Implementation
+## Self-check during implementation
 
 ```
-[ ] Am I about to write code that introduces a new pattern?   → STOP, /decide
-[ ] Am I adding new dependencies?                             → STOP, /decide
-[ ] Am I creating CI/CD or infra config?                      → STOP, /decide
-[ ] Am I designing something that has alternatives?           → STOP, /decide
+[ ] Am I adding a new dependency, service, integration, or technology?  → /decide
+[ ] Am I changing a data model, schema, or a security control?          → /decide
+[ ] Am I designing CI/CD, release, or infra?                            → /decide
+[ ] Am I adopting a pattern across the whole repo?                      → /decide
+[ ] Is this a routine choice reversible in this PR?                     → just write the code
 ```
 
-## What `/decide` Does
+## What `/decide` does
 
 Creates an **Agent Decision Record** (AgDR) that captures:
 
@@ -55,12 +69,20 @@ Creates an **Agent Decision Record** (AgDR) that captures:
 
 **AgDRs are stored at**: `{project}/docs/agdr/AgDR-NNNN-{slug}.md`. Each project has its own folder and its own ID sequence.
 
-## Enforcement
+Before drafting one, run `/agdr search <term>` — the portfolio may have already settled this, and re-litigating a decided call is its own kind of waste.
 
-1. **Real-time self-check** — scan every response for decision patterns before writing code
-2. **Workflow gate** — AgDR required before the Build phase for new features
-3. **Code Reviewer** — flags PRs with architecture changes that don't link an AgDR
-4. **Pre-commit hook** — warns if architecture files changed without an AgDR reference
+## Enforcement — and the honest limits of it
+
+| Layer | What it actually does |
+|-------|----------------------|
+| **Self-discipline** | The threshold above, applied before writing code. The primary mechanism. |
+| **Workflow gate** | AgDR required before the Build phase for new features (Gate 2, [`workflow-gates.md`](workflow-gates.md)). |
+| **Code Reviewer (Rex)** | Flags PRs with architecture changes that don't link an AgDR. |
+| **Pre-commit hooks** | `require-agdr-for-arch-changes.sh` / `require-agdr-for-arch-pr.sh`. |
+
+The hooks are **deliberately narrow** — they fire only on `infrastructure/**`, `*.tf` / `*.tfvars`, `Dockerfile*` / `docker-compose*`, and `.github/workflows/**`, and their own header says why: *"Deliberately NARROW. Dependency bumps and API schema changes are NOT included — too noisy, not all changes need an AgDR."*
+
+That narrowness is correct, and this rule is now written to match it. A rule far broader than anything anyone would actually enforce doesn't produce more records — it produces ignored prose, and it burns the agent's ceremony budget on decisions nobody needed written down. The threshold above is wider than the hooks (it covers dependencies, data models, and cross-cutting patterns the hooks can't see) but it is no longer unbounded. Adopters who want a broader mechanical net can extend `.claude/project-config.json` → `.architecture_paths`.
 
 ---
 
--- a/.claude/rules/agdr-decisions.md
+++ b/.claude/rules/agdr-decisions.md
@@ -80,9 +80,21 @@ Before drafting one, run `/agdr search <term>` — the portfolio may have alread
 | **Code Reviewer (Rex)** | Flags PRs with architecture changes that don't link an AgDR. |
 | **Pre-commit hooks** | `require-agdr-for-arch-changes.sh` / `require-agdr-for-arch-pr.sh`. |
 
-The hooks are **deliberately narrow** — they fire only on `infrastructure/**`, `*.tf` / `*.tfvars`, `Dockerfile*` / `docker-compose*`, and `.github/workflows/**`, and their own header says why: *"Deliberately NARROW. Dependency bumps and API schema changes are NOT included — too noisy, not all changes need an AgDR."*
+The two hooks fire at different moments and on **different, deliberately bounded** path sets — neither is a catch-all for "any decision":
 
-That narrowness is correct, and this rule is now written to match it. A rule far broader than anything anyone would actually enforce doesn't produce more records — it produces ignored prose, and it burns the agent's ceremony budget on decisions nobody needed written down. The threshold above is wider than the hooks (it covers dependencies, data models, and cross-cutting patterns the hooks can't see) but it is no longer unbounded. Adopters who want a broader mechanical net can extend `.claude/project-config.json` → `.architecture_paths`.
+| Hook | Fires on | Default trigger set |
+|------|----------|---------------------|
+| `require-agdr-for-arch-changes.sh` | `git commit`, against the **staged** diff | `*.tf`, `*.tfvars`, `docker-compose*.yml`, `Dockerfile*`, `.github/workflows/**` |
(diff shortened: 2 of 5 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #998

**Commit**: `f409de9e5134482a33fc836ea61505cd5734ccae`

### Summary

Narrows `.claude/rules/agdr-decisions.md` from *"HARD STOP before **any** technical decision"* to a materiality threshold, closing the third leg of the anti-over-process effort after #994 (right-size-ceremony) and #996 (role-trigger convergence). Four prose files, no hook / gate / marker / `settings.json` logic touched.

The core argument is sound and the compounding mechanism it identifies is real: every AgDR lands a file under `docs/agdr/**`, which re-fires the Tech Lead trigger (and the Solution Architect trigger for migration AgDRs), so a blanket AgDR rule literally manufactures the handover churn #996 just fixed. Narrowing this rule is what stops that fix being undermined by the rule sitting next to it.

**One blocking finding**, and it's an ironic one: the paragraph that makes this PR's central claim — *"the prose was the outlier, the hooks were already right"* — misdescribes what the hooks actually do, and contradicts a line in `docs/rule-audit.md` that this same PR edits.

### Checklist Results

- ✅ Architecture & Design:      Pass (prose only)
- ⚠️ Code Quality:               **Fail** — one factually inaccurate claim about shipped hook behaviour (see Issues #1)
- ✅ Testing:                    Pass — N/A for prose; markdownlint clean per PR body
- ✅ Security:                   Pass — rail 1 preserved verbatim in substance
- ✅ Performance:                N/A
- ✅ PR Description & Glossary:  Pass — narrative bullets, glossary present, `Fixes #997`
- ✅ Summary Bullet Narrative:   Pass
- ⚠️ Technical Decisions (AgDR): See Question #1 — arguable, not blocking
- ✅ Adopter Handbooks:          N/A (none loaded)
- ✅ Fallow Static Analysis:     N/A (not JS/TS)

---

### Issues Found

#### ⛔ 1 — BLOCKING: the "Enforcement — and the honest limits of it" paragraph misstates hook scope

```
The hooks are **deliberately narrow** — they fire only on `infrastructure/**`, `*.tf` /
`*.tfvars`, `Dockerfile*` / `docker-compose*`, and `.github/workflows/**` …
```

That sentence is plural ("the hooks"), and the table directly above it lists both `require-agdr-for-arch-changes.sh` and `require-agdr-for-arch-pr.sh`. It's wrong for both, in different directions:

**(a) The PR-time hook is materially broader — including dependency additions.** `require-agdr-for-arch-pr.sh` reads `.agdr_trigger_paths` from `project-config.defaults.json`, which is:

```
**/domain/**, **/infrastructure/**, **/migrations/**, infrastructure/**,
template.yaml, **/*.tf, **/*.tfvars, .github/workflows/**
```

**plus** `.agdr_trigger_dep_files` = `package.json`, `pyproject.toml`, `Cargo.toml`, `go.mod`, `Gemfile`, on which it detects dependency *additions* and exits 2 on `gh pr create`. So `**/domain/**`, `**/migrations/**`, `template.yaml`, and new dependencies are all inside the mechanical net, and the new prose tells the reader they aren't.

This directly contradicts `docs/rule-audit.md:87` — a file **this PR edits** — which still (correctly) reads: *"AgDR required at PR time when diff touches architecture paths **OR adds a new dependency**"*. The PR fixes one internal inconsistency and introduces a fresh one two lines apart.

**(b) `infrastructure/**` is in the commit-time hook's *header comment* but deliberately NOT in its code.** The actual `ARCH_GLOBS` in `require-agdr-for-arch-changes.sh` is:

```bash
ARCH_GLOBS='\.tf$
\.tfvars$
(^|/)docker-compose.*\.ya?ml$
(^|/)Dockerfile
^\.github/workflows/'
```

…and lines 60–67 of that hook explain exactly why `infrastructure/` was dropped (false-positives on `docs/infrastructure/notes.md` and `src/types/infrastructure/foo.ts`). `.claude/hooks/README.md:170` already documents this as a **known gap**. So the header comment is stale relative to its own code, and this PR quotes the stale header verbatim — promoting a documented header/code drift into the canonical rule file, where readers will trust it more.

**(c) The config pointer is half right.** *"Adopters who want a broader mechanical net can extend `.claude/project-config.json` → `.architecture_paths`"* configures the **commit-time** hook only. The PR-time hook reads `.agdr_trigger_paths[]` + `.agdr_trigger_dep_files[]` (as `rule-audit.md:87` already says).

**Why this blocks rather than being a nit:** it isn't a safety hole — the failure mode is loud (exit 2 at `gh pr create`, not a silent skip), and the threshold table *correctly* lists new dependencies as material, so an agent following the guidance still writes the AgDR. But this paragraph is, by the PR's own summary, "the load-bearing finding," and it's the one that's inaccurate. A canonical rule file that misdescribes shipped enforcement is the same class of defect this PR exists to fix, and it's a ~3-line edit. Suggested shape:

> The hooks are **deliberately narrow** relative to the threshold above. `require-agdr-for-arch-changes.sh` (commit-time) fires only on `*.tf` / `*.tfvars`, `Dockerfile*` / `docker-compose*`, and `.github/workflows/**` — its header explains why: *"Deliberately NARROW … not all changes need an AgDR."* `require-agdr-for-arch-pr.sh` (PR-time) is somewhat wider: `**/domain/**`, `**/infrastructure/**`, `**/migrations/**`, `template.yaml`, Terraform files, `.github/workflows/**`, plus *additions* to a dependency manifest. Neither can see a data-model choice or a cross-cutting pattern — which is why the threshold above is wider than either. Adopters can extend the commit-time net via `.architecture_paths` and the PR-time net via `.agdr_trigger_paths[]` / `.agdr_trigger_dep_files[]`.

Fixing this also removes the need to touch `rule-audit.md:87`.

---

**2 — A live hook string still quotes the removed bar.** `require-agdr-for-arch-pr.sh:595` prints this to the agent on every block:

```
  .claude/rules/agdr-decisions.md calls /decide a HARD STOP before any
  technical decision — library choice, architecture move, new dependency,
  infra shape.
```

That is the exact wording this PR deletes, in agent-facing output. Testing step 4's sweep found the two prose sites but not this one. **Don't fix it here** — `.claude/hooks/**` is rail-1 trust chain, so a Heavy-path PR with the Security Auditor, not a tack-on to a prose change. File a follow-up. (The enumerated examples after the em-dash are all material anyway, so the string is only wrong in its opening clause.)

**3 — Two dangling citations to a deleted line.** The old Enforcement item 4 — *"Pre-commit hook warns if architecture files changed without an AgDR reference"* — is gone, but it's quoted verbatim as an `Enforces:` target in `.claude/hooks/README.md:173` and `require-agdr-for-arch-changes.sh:7`. The new table row preserves the substance so nothing breaks, but the quoted string no longer exists. Same follow-up as #2 for the hook file; README is safe to fix here or later.

**4 — Threshold gap: dependency *removal* / replacement.** Row 1 reads "adds a new dependency, library, framework, or technology." A swap is covered (the replacement is new), but *dropping* a library to hand-roll the capability is neither an "add" nor obviously "hard to reverse" — yet it's the same maintenance-ownership call, and it's how you end up owning a bespoke reimplementation nobody decided to own. Suggest: "**adds, removes, or replaces** a dependency…".

**5 — Gate 4 in `workflow-gates.md` still reads at the old bar.** Gate 2 says "AgDR for **key** decisions" ✅ (aligns), but Gate 4 says "AgDR linked **if decisions made**" — unqualified. Two words: "if **material** decisions made". Not in this PR's diff, so cheap to fold in or defer.

**6 — Rex's own file has the old bar surviving below the edited line.** `code-reviewer.md` § 7 headline is now correctly "material," but § 7 enforcement rule 2 ("If **a decision** is detected but NO AgDR is linked → REQUEST CHANGES") and **Rules #7** ("if you detect **a technical decision** without an AgDR link: DO NOT approve the PR") are both unqualified. A careful reader resolves them against the headline ~250 lines above, but this is precisely the "place still quoting the old bar" class the change set out to sweep. Two-word fix in each.

---

### Questions

**1 — Does this change itself warrant an AgDR, or does it inherit AgDR-0107?** Applying the *new* threshold to this PR: it's a cross-cutting change to the framework's own recording posture, and its sibling #994 shipped with AgDR-0107. The PR ships none. The inheritance argument is defensible — the rule explicitly frames itself as "the decision-record sibling of right-size-ceremony," and AgDR-0107 already recorded the match-ceremony-to-blast-radius call. But AgDR-0107's decision statement is scoped to *review tiering* (Lean/Standard/Heavy for review ceremony); the *recording* threshold is a distinct policy with its own trade-off (what class of decision goes unrecorded). **Not blocking** — rail 2 says ambiguity rounds up, and I'd rather see one sentence in the PR body stating the inheritance than a ceremonial AgDR. Please just say which it is.

---

### Assessment against the stated review focus

**Does the narrowing go too far?** No. I looked hard for a material class that slips between the threshold table and the exemption list, and only found #4 (dependency removal). Everything else I probed — breaking an existing published API or wire format, adopting a cache/queue, a vendor swap, a licensing choice — lands on the "**hard to reverse** — externally visible, shared-state, or expensive to undo" row. That row is doing a lot of load-bearing work for cases the six specific rows don't name; it holds, but it's worth knowing it's the catch-all. The exemption list is safely scoped: every item is qualified by "reversible inside a single PR," "error-handling shape" / "control flow" don't collide with the repo-wide-pattern row, and "calling an API from a dependency already in the manifest" explicitly parenthesises the disclaimer that *adding* it is material. Combined with rail 2, the residual risk is acceptable.

**Are the rails preserved in substance?** Yes, and this is the part the PR gets most right. Rail 1 enumerates `.claude/hooks/**`, `.claude/settings.json`, auth, crypto, secrets, migration — substantively identical to `right-size-ceremony.md:33` — and adds the concrete "a one-line change to a merge gate is exactly where you want the record." Rail 2 is identical in substance *and* in its tolerated-failure framing. Critically, both rails are **re-encoded operationally** in the two self-check blocks, not merely mentioned in prose:

```
[ ] Is it security / trust-chain / migration related?   YES → material, no exceptions (rail 1)
[ ] Am I unsure whether it's material?                  → round UP, write it (rail 2)
```

That's the "in substance, not just mentioned" test passing — an agent running the checklist hits both rails whether or not it read the prose.

**Internal consistency.** ✅ against `right-size-ceremony.md` (explicit sibling framing, no contradiction). ✅ against `workflow-gates.md` Gate 2 ("key decisions"); ⚠️ Gate 4 per #5. ✅ against the spike/prototype exemption row (unaffected). ❌ against `rule-audit.md:87` and the hooks' real behaviour per Issue #1.

**Anchor integrity.** ✅ `rule-audit.md:85` correctly repointed to `§ The threshold`. Rows 86/87/90 cite `§ Enforcement`, now titled "Enforcement — and the honest limits of it" — prose citation, prefix-matches, fine. Row 89 cites `§ What /decide Does` vs the new lowercase `## What /decide does` — case-only, fine. The one genuinely dangling citation is #3.

**On the edit to my own § 7** — reviewing it on its merits: the narrowed bar is clear and operable. It's directive on both sides ("do NOT need an AgDR **and must not be flagged**"), which removes the ambiguity that would otherwise have me flagging at the old bar out of caution — the failure mode that makes a narrowing cosmetic. The detection table below it is untouched and every row is already material-shaped (new deps in build files, first-time ORM/queue/cache setup, Repository-pattern introduction, storage/serialization/state-management choices, REST-vs-GraphQL), so I lose no detection capability — I lose only the licence to flag someone for naming a variable. Carrying both rails inline means I never have to open the rule file to know a `.claude/hooks/**` diff is always material. Yes, I can still catch genuinely material undocumented decisions; #6 is the only loose end.

---

### Verdict

**CHANGES REQUESTED** — scoped to Issue #1 only.

Everything else here is good work: the narrowing is well-argued, the rails survive in substance rather than in name, the compounding `docs/agdr/**` → role-trigger mechanism is a genuine insight, the anchor updates that mattered are correct, and the PR description is exemplary. Fix the enforcement paragraph so it describes both hooks accurately (and stops contradicting `rule-audit.md:87`), answer Question #1 in a sentence, and I'll approve on the next pass. Suggestions #2–#6 are follow-up material — #2 and #3 touch trust-chain files and should go through the Heavy path separately.

---
📌 Reviewed commit: `f409de9e5134482a33fc836ea61505cd5734ccae`

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #998 — re-review at new HEAD

**Commit**: `fce26560cf2d0f426cf93351abe7a0ddbe31639d`
(Prior review at `f409de9e` returned CHANGES REQUESTED.)

### Summary

The blocking finding is **resolved**. I re-derived the hook trigger sets from source rather than from the PR's description, and the new two-row table is accurate line-for-line. The materiality narrowing is now consistently applied across the rule, `CLAUDE.md`, Rex's § 7 + Rules #7, and `rule-audit.md`, with AgDR-0108 recording the call and #999 correctly carrying the trust-chain wording.

### Verification of the blocking finding

I read both hooks directly and compared against the table.

**`require-agdr-for-arch-changes.sh` (commit-time)** — actual `ARCH_GLOBS` (L68–72):

```
\.tf$   \.tfvars$   (^|/)docker-compose.*\.ya?ml$   (^|/)Dockerfile   ^\.github/workflows/
```

Table says `*.tf`, `*.tfvars`, `docker-compose*.yml`, `Dockerfile*`, `.github/workflows/**`. ✅ Exact match — and correctly *excludes* `infrastructure/`.

**`require-agdr-for-arch-pr.sh` (PR-time)** — `.agdr_trigger_paths` in `project-config.defaults.json` **and** the inline fallback (L404–412) are identical:

```
**/domain/**  **/infrastructure/**  **/migrations/**  infrastructure/**
template.yaml  **/*.tf  **/*.tfvars  .github/workflows/**
```

Dep files: `package.json`, `pyproject.toml`, `Cargo.toml`, `go.mod`, `Gemfile`. ✅ Exact match, both source and order.

**"only the PR-time hook watches dep manifests, and only for additions"** — verified in code, not just comments. `package.json` compares the *key union* of `dependencies` + `devDependencies` + `peerDependencies` + `optionalDependencies` at merge-base vs HEAD and fires only on keys new at HEAD (L528–541); a version bump changes a value, not a key, so it cannot fire. Non-JSON files use the `ADDED_LINES > REMOVED_LINES` heuristic (L546–553), where a bump is a matched `+`/`-` pair. ✅ Accurate.

**The `infrastructure/` omission paragraph** — the false-positive reasoning (`docs/infrastructure/notes.md`, `src/types/infrastructure/foo.ts`; word ambiguous between IaC and library code; Terraform caught via `\.tf$` at any depth; CDK/Pulumi override via `.architecture_paths`) reproduces the hook's own comment at L61–67 faithfully. Calling out that the mention survives only in a **stale header comment** is exactly the right framing — that comment is what misled the first draft, and #999 schedules the fix. ✅

**Config surfaces + bypass** — `.architecture_paths` (L76), `.agdr_trigger_paths[]` / `.agdr_trigger_dep_files[]` (L398–399), and `<!-- agdr: not-applicable -->` scoped correctly to the *PR* gate with a stderr WARN (L146–148). ✅

That closes the blocking issue. I did not take the PR's word for any of it.

### The other questions I raised

**`docs/rule-audit.md:87` — "OR adds a new dependency".** No inconsistency; I was wrong to flag it as a possible one. That row describes the **hook**, and the hook genuinely fires on `(a)` an arch-path match **OR** `(b)` a dep addition. Hook behaviour is unchanged by this PR, so the row is still correct as written. Nothing to do.

**Dependency *removal* gap — not worth blocking.** Neither hook ever covered removals, so there's no prose↔hook regression here. The material subset (retiring a technology, dropping an integration) is already caught by the threshold's *"new service, bounded context, or external integration"* and *"hard to reverse"* rows. Deleting a dependency you stopped using is reversible inside one PR and correctly sits under the exemption list. Correctly out of scope.

**AgDR-0108 — sound.** The decision statement matches what actually shipped, item for item (threshold list, exemption list, both rails, the heuristic demoted from trigger to writing aid). The options table carries four genuine alternatives with honest cons — notably "narrow the prose to the hooks' exact path set" is listed and rejected for the right reason (globs match paths, not judgment), which is the argument the shipped rule's closing paragraph then makes. Consequences name the residual risk (a material decision misjudged routine) and its mitigations rather than claiming there is none. The self-application justification answers my question directly and correctly: AgDR-0107 scoped *review tiering*, this is *recording policy* — related but distinct, so its own record is right, not an inheritance. The "deliberately not in scope" paragraph pre-registering the hook wording as Heavy-path work is good practice and matches what #999 actually says.

**#999** — scoped tightly (message text only, explicit "zero behavioural change" AC, Security Auditor required), and it picks up the stale `infrastructure/**` header comment too. The deferral is legitimate, not a dodge.

### Issues Found

None blocking.

### Suggestions

**1. The PR description still repeats the claim I blocked on.** Summary bullet 3 says `require-agdr-for-arch-changes.sh` *"fires only on `infrastructure/**`, `*.tf`/`*.tfvars`, `Dockerfile*`/`docker-compose*`, and `.github/workflows/**`"* — the `infrastructure/**` half is the same inaccuracy the fix commit removed from the rule file. Shipped content is correct, so this doesn't block, but on a squash merge this body becomes the permanent commit message. **Please edit the body before merging** — no code change needed.

**2. `roles/engineering/tech-lead.md:16` still encodes the demoted heuristic.** It reads *"Make day-to-day technical decisions — record **any call with real alternatives** via `/decide`"*. "Has alternatives" is precisely the trigger this PR demotes to a writing aid, on the stated grounds that it over-fires by construction. It's also sitting in the role file of the role that gets **re-activated by every `docs/agdr/**` write** — the loop this change exists to break. The PR body claims a repo sweep found "the two that mattered"; I'd argue this is a third. It's prose in `roles/`, not trust chain, so a one-line fix here would be proportionate — otherwise fold it into a follow-up. Not blocking, but it's the strongest remaining leak of the old bar.

**3. nit — `.claude/skills/tech-vision/SKILL.md:205`**: *"AgDRs for any technical decisions implied by the migration path"*. The "implied by the migration path" scoping makes these inherently material, so it's defensible as-is; flagging only so it isn't mistaken for an oversight later.

**4. nit — Rex's § 7 detection table** (untouched, correctly so): the *"New design patterns — Factory, Builder, Singleton implementations"* row can still read as firing on a single-file helper. Enforcement rule 2's new materiality gate catches that downstream, so the table works as a candidate list. Fine to leave; worth a word if it's ever revisited.

**5. Advisory — summary bullet narrative quality: passes.** Every bullet states what changed and why it matters. No action.

### Handbook Findings

`handbooks/architecture/clean-architecture-layers.md`, `handbooks/architecture/migration-safety.md` (blocking), and `handbooks/general/commit-message-quality.md` loaded. No code, no migrations in this diff — architecture handbooks N/A. Both commit subjects are conventional-format with a ticket ref and describe the change accurately; no findings.

### Fallow Findings

N/A — no JS/TS in the diff.

### Checklist Results

- ✅ Architecture & Design: Pass (prose only; no mechanical change — confirmed against the file list)
- ✅ Code Quality: Pass
- ✅ Testing: Pass — hook test suite SUCCESS, markdownlint SUCCESS; no behaviour to test
- ✅ Security: Pass — no trust-chain file touched; the trust-chain wording correctly deferred to #999 on the Heavy path (rail 1 respected)
- ✅ Performance: N/A
- ✅ PR Description & Glossary: Pass (glossary present; see Suggestion 1 re: the body's stale hook claim)
- ✅ Technical Decisions (AgDR): Pass — AgDR-0108 present, sound, and self-applying
- ✅ Adopter Handbooks: Pass
- ⚠ Fallow Static Analysis: N/A

### Verdict

**APPROVED**

The blocking factual inaccuracy is fixed and independently verified against both hook sources. Suggestion 1 (PR body edit) is worth doing before you merge, since it becomes the commit message; Suggestion 2 deserves either a one-line fix or a ticket, but neither is a merge blocker.

---
📌 Reviewed commit: `fce26560cf2d0f426cf93351abe7a0ddbe31639d`

Issue #1008, opened 2026-07-24
[Bug] reindex-on-session-start.sh committed 100644 but exec'd directly — Permission denied on every session start


bitwhispererrr:
## Affected
`.claude/hooks/reindex-on-session-start.sh` (+ its `SessionStart` wiring in `.claude/settings.json`)

## Given / When / Then
**Given** a fresh clone of the framework (any version through v5.2.0)
**When** a Claude Code session starts and the `SessionStart` hook runs
**Then** it fails with `SessionStart:startup hook error … bash: line 1: <ops>/.claude/hooks/reindex-on-session-start.sh: Permission denied` — expected the hook to run and reindex the apexyard-search MCP index silently.

## Repro
1. Clone `me2resh/apexyard` (or any fork) at v5.1.0 or v5.2.0.
2. Start a Claude Code session in the fork root.
3. Observe the `SessionStart:startup hook error` — `Permission denied` on `reindex-on-session-start.sh`.

Root cause: the file is committed with mode `100644` (non-executable), while `settings.json` invokes it with `exec "$r/.claude/hooks/reindex-on-session-start.sh"`, which requires the execute bit. Every other directly-exec'd hook is committed `100755`.

## Framework version
v5.1.0 (also confirmed at v5.2.0 and `upstream/main` HEAD — blob `02b52d5`, mode `100644` at all three)

## Severity
major — a shipped hook never executes on any fresh clone (MCP search index silently goes stale), and every adopter sees a startup error banner. Workaround exists (`chmod +x` locally).

## Notes
Fix: commit the file `100755` to match every other directly-invoked hook — `git update-index --chmod=+x .claude/hooks/reindex-on-session-start.sh`. The five `_lib-*.sh` files that are also `100644` are correct as-is (they're `source`d, never exec'd). Related: #733 (the feature that added this hook).

## Glossary
| Term | Definition |
|------|------------|
| file mode `100644` / `100755` | Git's record of a tracked file's permissions; `755` includes the executable bit, `644` does not. |
| `exec` (in the hook wiring) | Shell builtin that replaces the process with the named program — requires the target file to be executable. |
| SessionStart hook | A Claude Code hook that runs when a session begins. |

atlas-apex:
Thanks for this one, @bitwhispererrr — genuinely useful report. You'd already done the diagnosis: the file mode, the direct `exec` in the `SessionStart` wiring, and a clean repro from a fresh clone. That made it a fast fix rather than an investigation.

[...]

PR #1009, opened 2026-07-25
fix(#1008): set exec bit on reindex-on-session-start.sh

atlas-apex:
## Summary
- **Set the exec bit on `reindex-on-session-start.sh`** — it was committed `100644` (non-executable) while its `SessionStart` wiring in `.claude/settings.json` invokes it directly via `exec "$r/.claude/hooks/reindex-on-session-start.sh"`, which requires the execute bit. Every fresh clone (through v5.2.0) hit `Permission denied` on session start, and — worse — the apexyard-search MCP index silently never refreshed, so agents quietly fell back to grep without knowing the index had gone stale.
- **Left the five `_lib-*.sh` files at `100644`** — `_lib-audit-history.sh`, `_lib-ops-root.sh`, `_lib-premium-hook.sh`, `_lib-review-markers.sh`, `_lib-tracker.sh` are `source`d by other hooks, never `exec`'d directly, so non-executable is correct for them. Confirmed by extracting every `exec "$r/.claude/hooks/<name>.sh"` reference out of `settings.json` (49 hooks) and checking none of them are `_lib-*` files.
- **Added a regression guard** (`test_hook_exec_bits.sh`) — the reason this shipped broken in the first place is that a mode-only change produces an *empty* content diff, so a normal code review has nothing to look at. The new test extracts every hook `settings.json` invokes via direct `exec` and asserts each is `100755` in git's index; a negative-control check confirms it doesn't accidentally flag the sourced `_lib-*.sh` files. Verified it actually catches the regression by reverting the bit locally (`git update-index --chmod=-x`) and re-running — it failed as expected, then passed again after restoring the fix.

## Root cause
`git update-index --chmod=+x .claude/hooks/reindex-on-session-start.sh` is the entire content of the fix — a pure mode-bit change, zero lines of code touched (`git diff --cached --stat` shows `0 insertions(+), 0 deletions(-)` for that file).

## Considered and deferred
The ticket also asked me to weigh whether `SessionStart` should invoke via `bash <script>` instead of relying on the exec bit. I left this out of scope for this PR: all 49 other directly-`exec`'d hooks already rely on the same exec-bit convention, so switching just this one hook to a `bash`-wrapped invocation would make it inconsistent with its siblings, not more robust — the class-of-bug fix (the guard test) is the better lever here. Happy to open a follow-up if the team wants `bash`-wrapped invocation across all 49 hooks as defense-in-depth.

## Testing
1. `git ls-tree upstream/dev -- .claude/hooks/reindex-on-session-start.sh` → confirms `100644` before the fix; after `git update-index --chmod=+x`, `git ls-files -s` shows `100755`.
2. `echo '{}' | .claude/hooks/reindex-on-session-start.sh` → exits `0` (previously would have hit `Permission denied` when invoked via `exec` from a fresh non-executable checkout).
3. `bash .claude/hooks/tests/test_reindex_on_session_start.sh` → 9/9 pass (pre-existing suite, unaffected by the mode fix).
4. `bash .claude/hooks/tests/test_hook_exec_bits.sh` (new) → 3/3 pass with the fix; reverting the exec bit locally reproduces the failure, confirming the guard actually catches this class of regression.
5. Full suite (`bin/run-hook-tests.sh`) run once in the background for completeness: 113/114 pass. The one failure, `test_sync_codex_adapter.sh` ("generated hook command preserves blocking/allowing exit"), is unrelated to this change — it exercises `bin/sync-codex-adapter.sh`'s codex-CLI adapter generation, has no connection to file mode bits or `reindex-on-session-start.sh`, and reproduces identically on `upstream/dev` before this branch's commit. Flagging it here for visibility rather than silently absorbing it into this PR's scope.

Refs #1008

---

## Glossary
| Term | Definition |
|------|------------|
| File mode `100644` / `100755` | Git's record of a tracked file's Unix permissions; `755` includes the executable bit, `644` does not. Git only tracks the executable bit — not full permission bits — so these are the only two modes that matter for a regular file. |
| `exec` (in hook wiring) | A shell builtin that replaces the current process image with the named program. It requires the target file to have its executable bit set — unlike calling `bash <script>`, which would work regardless of the bit. |
| `SessionStart` hook | A Claude Code hook that runs automatically when a session begins — used here to opportunistically refresh the apexyard-search MCP index. |
| Mode-only diff | A git change that alters a file's permission bits but not its content — shows as `0 insertions(+), 0 deletions(-)` in a diff, which is why this class of regression is easy to miss in review. |


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1009

**Commit**: `ef45dfc321ad9c9d9838df20bcf52b5b4d117405`

### Summary

Sets the executable bit on `.claude/hooks/reindex-on-session-start.sh` (`100644` to `100755`) so the `SessionStart` wiring, which invokes it via `exec`, stops failing with `Permission denied` on fresh clones -- and adds `test_hook_exec_bits.sh`, a regression guard for the whole class of bug.

I verified the substantive claims rather than taking them on trust:

- **Content really is unchanged.** The blob SHA is byte-identical either side of the change (`02b52d5def8af3865cf1b581630dc70afa032701` at the merge-base and at `ef45dfc`), with only the tree mode differing. A pure mode-bit change, exactly as described.
- **The exclusion of the sourced libs is correct.** `settings.json` contains zero `_lib-` references, so the extraction cannot pull them into the required-executable set. The five files the PR body names as legitimately `100644` (`_lib-audit-history`, `_lib-ops-root`, `_lib-premium-hook`, `_lib-review-markers`, `_lib-tracker`) are sourced, never exec'd -- demanding an exec bit on them would have been wrong, and the test does not.
- **The guard actually catches the regression.** Ran the test on the PR head in an isolated worktree: 3/3 pass across 49 directly-invoked hooks. Then reverted the bit with `git update-index --chmod=-x` and re-ran -- it failed, naming the offending file and its mode (`reindex-on-session-start.sh(100644)`), and passed again once restored. The author's verification claim holds.
- **Every hook path in `settings.json` is genuinely `exec`-invoked** -- I checked for any reference not preceded by `exec` and found none, so the extraction's over-approximation is exact today.

The design instinct here is the right one. A mode-only change produces an empty content diff, so a human reviewer has literally nothing to look at -- mechanising the invariant is a better lever than asking reviewers to be more careful. The test also fails closed: if the extraction regex ever drifts and returns nothing, the `checked -gt 0` guard turns that into a failure rather than a silent pass.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass**
- Security: **Pass** (see trust-chain note below)
- Performance: **Pass** (N/A -- one `git ls-files` per hook, in a test)
- Summary Bullet Narrative: **Pass** -- bullets carry both the what and the why; the first explains the silent index-staleness consequence, which is the part that actually mattered
- Technical Decisions (AgDR): **N/A** -- no material decision. Restoring a file to the convention its 49 siblings already follow is not a new architectural call, and the `bash <script>` alternative was explicitly considered and deferred with reasoning in the PR body rather than silently skipped
- Adopter Handbooks: **N/A** -- `migration-safety.md` (blocking) and `clean-architecture-layers.md` do not apply to a shell-hook mode fix; `commit-message-quality.md` is satisfied (typed subject, bullets explaining why, `Refs #1008`)
- Fallow Static Analysis: **N/A** -- no JS/TS in the diff

### Issues Found

None blocking.

### Suggestions

**nit: the extraction does not verify the invocation verb, and your own deferred follow-up would break it.**

`test_hook_exec_bits.sh:53` extracts with:

```bash
EXEC_HOOKS=$(grep -oE '\.claude/hooks/[A-Za-z0-9_-]+\.sh' "$SETTINGS" | sort -u)
```

That matches any hook path in `settings.json` regardless of how it is invoked, while the header comment and the pass message both describe the set as "directly-exec'd". Today those are the same set, so the test is correct as shipped. But the PR body floats a possible follow-up -- switching hooks to `bash <script>` invocation as defense-in-depth. The moment one hook moves to that form, this test would demand an exec bit on a hook that no longer needs one, and the failure message would point at the wrong culprit. Anchoring the match on the verb would make the behaviour match the stated contract:

```bash
EXEC_HOOKS=$(grep -oE 'exec +"[^"]*\.claude/hooks/[A-Za-z0-9_-]+\.sh' "$SETTINGS" \
  | grep -oE '\.claude/hooks/[A-Za-z0-9_-]+\.sh' | sort -u)
```

Not blocking -- the current form is accurate against today's `settings.json`, and the fail-closed guard means a drifted regex surfaces loudly rather than silently.

**nit: invariant 2's message overstates how uniform the `_lib-*` convention is.**

The negative-control message says `_lib-*.sh` files are "sourced, not exec'd -- 100644 is correct for them". True of the five this PR audited, but 13 of the 18 `_lib-*.sh` files in the tree are already `100755` despite also being sourced. The invariant itself is fine (it only asserts none leak into the exec set), but the parenthetical reads as a repo-wide claim that is not quite true. Pre-existing inconsistency, well outside this PR's scope -- worth a follow-up only if someone wants the convention uniform.

**note: trust-chain path.** The diff touches `.claude/hooks/**`, which per `.claude/rules/role-triggers.md` (#777) auto-fires the Security Auditor and, per `right-size-ceremony.md` rail 1, means this does not take the Lean path on size alone. Flagging for the orchestrator's routing rather than as a finding: the change adds no executable logic to the trust chain -- it restores a convention and adds a test -- so I see no security surface here, but the routing call is not mine to make.

**Out of scope, as flagged:** the `test_sync_codex_adapter.sh` failure is pre-existing and unrelated (it exercises codex-adapter generation, nothing to do with file modes). Correctly surfaced rather than absorbed, and it should not hold this PR.

### Verdict

**APPROVED**

---
Reviewed commit: `ef45dfc321ad9c9d9838df20bcf52b5b4d117405`

atlas-apex (commented):
## Code Review: PR #1009 (delta re-review)

**Commit**: `2860801a8b2217fded0975f7c043111ef9639359`
**Base of this pass**: `ef45dfc321ad9c9d9838df20bcf52b5b4d117405` (already APPROVED)

### Summary

This is a **delta re-review over an already-approved base**, not a fresh full review. The substantive analysis from the prior pass at `ef45dfc` stands unchanged and is carried forward.

The only change since that approval is one orchestrator-authored commit, `2860801`, which removes a single dead variable assignment (`HOOKS_DIR`) from `.claude/hooks/tests/test_hook_exec_bits.sh`. That assignment tripped the `shellcheck .claude/hooks` CI job (warnings-as-errors) with `SC2034: HOOKS_DIR appears unused`. `SETTINGS` is the only path the test actually uses.

### Delta verification

| Check | Result |
|---|---|
| `git diff ef45dfc..2860801` | 1 file changed, **1 deletion, 0 insertions** — exactly the `HOOKS_DIR` line. Nothing else rode along. |
| Remaining references to `HOOKS_DIR` in the test file at the new head | **Zero** — the removal cannot have broken a later use. |
| File mode on `.claude/hooks/reindex-on-session-start.sh` at the new head | Still `100755` — the push did not disturb the mode change the PR exists to make. |
| CI at `2860801` | **Green on all checks**, including the `shellcheck .claude/hooks` job that failed at the old head and the full hook test suite job. CI ran against the real PR head, which is the project's own gate and stronger evidence than a local run. |

Lint-only fix, no behaviour change: removing an unreferenced assignment cannot alter what the test asserts or what it exercises. The guard test's value — catching a reverted exec bit — is untouched.

### Checklist Results
- Architecture & Design: Pass (unchanged from prior pass)
- Code Quality: Pass — delta strictly reduces dead code
- Testing: Pass — test semantics unchanged; suite green in CI at this head
- Security: Pass (unchanged from prior pass)
- Performance: N/A
- PR Description & Glossary: Pass — Summary bullets and Glossary present
- Technical Decisions (AgDR): N/A — no material decision in the delta
- Adopter Handbooks: N/A
- Fallow Static Analysis (JS/TS): N/A — no JS/TS in the diff

### Issues Found

None.

### Suggestions

Both nits from the prior pass stand as **nits only** — not escalated, not blocking, fine to address in a follow-up or not at all:

1. `nit:` the mode-extraction regex is not anchored on `exec`, so it is looser than it needs to be. Works correctly for the inputs the test feeds it; anchoring would just make the intent explicit.
2. `nit:` invariant 2's failure message reads as though `100644` is the *correct* mode for `_lib-*` files. More precisely, `_lib-*` files are *excluded from the exec-bit requirement* — the message overstates a prescription where an exclusion is meant.

### Verdict

**APPROVED**

Prior verdict carried forward. The delta is a clean, lint-only removal of one unused assignment; CI is green at this head; the exec-bit mode change the PR exists to deliver is intact.

---
Reviewed commit: `2860801a8b2217fded0975f7c043111ef9639359`

PR #1005, opened 2026-07-23
sync(#1001): main→dev after v5.2.0 release

atlas-apex:
## Summary

- **Syncs main→dev after the v5.2.0 release** — makes the v5.2.0 squash commit (`88bbd1d`) an ancestor of `dev`, so the next `dev→main` release PR only sees genuinely-new commits instead of fighting accumulated squash divergence. Verified locally: `git merge-base --is-ancestor 88bbd1d HEAD` passes on this branch.
- **Merge strategy `-X ours`** — `dev` wins on every conflict, which is correct here because `dev` already carries the un-squashed equivalents of everything in the squash commit. No logic is added or removed; this is bookkeeping.
- **`CHANGELOG.md` carried cleanly through the merge** — the step-5b carry-forward (apexyard#448) was a **no-op** this cycle: `dev` had no competing CHANGELOG edits between releases, so `-X ours` took `main`'s v5.2.0 section without conflict. Verified identical to `upstream/main`, and the `[v5.2.0]` section is present on this branch. The guard behaved exactly as designed — idempotent, no spurious commit.
- **No functional changes** — one merge commit reconciling SHA divergence introduced by the squash-merge release flow.

## Background

The apexyard release flow squash-merges dev→main on every release. That creates divergence: `main` gets one squash commit; `dev` still carries the original un-squashed commits. A later dev→main release PR then conflicts on every diff the squash also touched. v2.0.0 suffered 99 conflicts from exactly this accumulation.

This PR is the low-ceremony fix: merge main→dev with `-X ours` so the squash commit becomes an ancestor of `dev`.

See [#403](https://github.com/me2resh/apexyard/issues/403) for the full root-cause analysis.

## Testing

1. After merging, verify `git log upstream/dev..upstream/main --oneline` returns empty
2. Verify `git log upstream/main..upstream/dev --oneline` shows only commits newer than v5.2.0
3. Verify CHANGELOG is in sync: `diff <(git show upstream/main:CHANGELOG.md) <(git show upstream/dev:CHANGELOG.md)` returns empty (apexyard#448)
4. Verify ancestry: `git merge-base --is-ancestor 88bbd1d upstream/dev` returns true
5. The next `/release` should prepend v5.3.0 cleanly on top of v5.2.0 — and, for the first time, read its changelog range from the `Released-From` trailer rather than the sync-boundary heuristic (see #1004)

## ⚠️ Merge with `--merge`, never `--squash`

[...]

Issue #1004, opened 2026-07-23
[Bug] Released-From trailer is dropped by squash merge (COMMIT_MESSAGES), defeating AgDR-0094


atlas-apex:
## Given / When / Then

- **Given** `me2resh/apexyard` has `squash_merge_commit_message = COMMIT_MESSAGES` (verified via `gh api repos/me2resh/apexyard --jq '.squash_merge_commit_message'`), so GitHub builds a squash commit's body from the **branch's commit messages**, not the PR body,
- **When** a release PR is squash-merged with the default settings — its `Released-From` trailer living only in the **PR body**, per `/release` step 5,
- **Then** the trailer never reaches the squash commit on `main`, and the next release cut cannot read it back — silently degrading to the sync-boundary heuristic that AgDR-0094 was written to replace.

## Why this matters more than it looks

AgDR-0094 introduced the trailer specifically to make the changelog range **deterministic** and close #872 / #737. But the mechanism as shipped writes the trailer to a surface GitHub discards at merge time under this repo's settings. The net effect is that AgDR-0094's guarantee has never actually held.

**This is not hypothetical — it already bit us on this very release.** AgDR-0094 shipped *in* v5.1.0, so v5.1.0's own release PR (#938) had no trailer. Cutting v5.2.0 therefore had to fall back to the sync-boundary heuristic (`sync/main-to-dev-after-v5.1.0` → `da262ab`), which is precisely what tripped the bogus count-mismatch warning filed as #1002. Left alone, v5.2.0's trailer would vanish the same way and v5.3.0 would repeat the whole cycle.

Found by Rex reviewing release PR #1003.

## Repro

1. `gh api repos/me2resh/apexyard --jq '.squash_merge_commit_message'` → `COMMIT_MESSAGES`
2. Squash-merge a release PR whose trailer is only in the PR body.
3. `git log -1 --pretty=format:'%(trailers:key=Released-From,valueonly)' <tag>` → **empty**.

**Severity: medium-high.** No immediate breakage, but it silently disables a guarantee the release process depends on, and the failure is invisible until the *next* cut produces a wrong changelog range.

## Candidate directions (implementer's call)

- **Put the trailer in the release *commit* message**, not only the PR body — then `COMMIT_MESSAGES` carries it through by construction, independent of merge-time flags or repo settings. Most robust; makes the artifact self-contained.
[...]

Issue #1002, opened 2026-07-23
[Bug] /release count-mismatch guard always false-positives under the release-cut model


atlas-apex:
## Given / When / Then

- **Given** the release-cut branch model (AgDR-0007), where `main` only ever receives **squash** merges of release PRs,
- **When** `/release` runs its count-mismatch guard (AgDR-0094) comparing `git rev-list --count upstream/main..upstream/dev` against the changelog's entry count,
- **Then** it reports a huge bogus gap and prints the `#872` truncation warning on a perfectly correct changelog — because `main..dev` structurally **can never shrink**.

## Why it is structural, not a one-off

Each release squashes N `dev` commits into **one** commit on `main`. Those N original commits stay on `dev` and are never reachable from `main`, so they remain inside `main..dev` **forever**. `/release-sync` makes the squash commit an *ancestor* of `dev`, which prevents conflicts — it does not remove the pre-squash commits from the range. So `main..dev` accumulates the un-squashed history of every past release, and the gap grows monotonically with each cut.

## Observed (v5.2.0 cut, 2026-07-24)

```
RAW_COUNT=399   ENTRY_COUNT=30   GAP=369   TOLERANCE=5
⚠️  WARNING: ... This is the #872 signature — a truncated changelog range.
```

The changelog was **correct**. Verified against the v5.1.0 sync-back point (`da262ab`):

```
$ SYNC=$(git rev-list upstream/dev --grep="sync/main-to-dev-after-v5.1.0" -n 1)
$ git log --oneline "$SYNC"..upstream/dev | wc -l
29
```

29 commits, 29 changelog entries (the 30th `- ` line the guard counted is the trailing `Closes #…` summary — see the second bug below). Zero real gap.

**Severity: medium.** It never blocks a release, but skill rule 10 says *"Don't proceed past it without the operator explicitly confirming the range is correct"* — so every future cut now demands a manual verification against a warning that is guaranteed to be wrong. A guard that always cries wolf trains the operator to wave it through, which destroys the #872 protection it exists to provide. That is worse than having no guard.

## Two smaller bugs found alongside it

1. **`ENTRY_COUNT` over-counts by one.** It counts every `^- ` line, including the trailing `- Closes #…` summary bullet, which is not an entry. Off-by-one on every run.
[...]

PR #1012, opened 2026-07-25
fix(#1002): re-anchor release count-mismatch guard + fix dropped Released-From trailer

atlas-apex:
## Summary

- **The count-mismatch guard is re-anchored to the real release range, not `main..dev`** — under the release-cut squash model, `main` only ever receives one squash commit per release, so every individual `dev` commit stays permanently unreachable from `main`. `upstream/main..upstream/dev` therefore *only grows*, and comparing the changelog's entry count against it always false-positives. Verified against real history: the v5.2.0 cut showed `RAW_COUNT=402` against `~1-2` real entries — a "gap" of 400 on a changelog that was correct. The guard now compares against `$LOG_RANGE`, the exact range `bin/release-changelog.sh` already resolves internally (trailer-anchored per AgDR-0094, or the `#737` sync-boundary fallback) — which the script now also prints to stderr so the guard can reuse it instead of re-deriving it. On the same real range, `RAW_COUNT=4` against `ENTRY_COUNT=1` — gap 3, well inside tolerance, no false warning.
- **Two smaller `bin/release-changelog.sh` bugs fixed alongside it** — `ENTRY_COUNT` was over-counting by one because it included the trailing `- Closes #N, ...` summary line as if it were a changelog entry; and the release-description line was missing spaces after commas (`"2 features,13 fixes"`) because `${arr[*]}` with a multi-character `IFS=', '` only honours the *first* character of `IFS` when joining, silently dropping the space. Both are fixed with 7 new regression tests (58/58 passing, up from the pre-existing 51/51).
- **The `Released-From` trailer now reliably reaches the squash commit on `main`.** This repo has `squash_merge_commit_message=COMMIT_MESSAGES` (verified via the GitHub API), so a squash merge builds its commit body from the *release branch's own commits*, not the PR body — but `/release`'s step 5 only ever wrote the trailer into the PR body. The trailer was therefore silently dropped on every merge, defeating AgDR-0094 (confirmed: v5.1.0's release commit carries none). Fixed two ways: the trailer now also lives in the release branch's own commit message (step 4), so it survives by construction; and the documented merge command now passes `--subject`/`--body-file` explicitly on `gh pr merge --squash`, so the reviewed PR content — trailer included — is what lands on `main`, independent of the repo's squash-message setting. A mandatory post-merge check (`git log -1 --pretty=format:'%(trailers:key=Released-From,valueonly)'`) now fails loudly if the trailer is missing, instead of the gap staying invisible until the *next* release's changelog range silently degrades.
- **Why one PR for two tickets:** both defects live in the same file (`.claude/skills/release/SKILL.md`) and are two symptoms of the same broken mechanism — the count-guard (#1002) only false-positived on the v5.2.0 cut *because* the trailer it should have anchored on (#1004) wasn't there, forcing a fallback to the noisy sync-boundary heuristic. Splitting them into separate PRs touching the same sections would guarantee a merge conflict; fixing them together also lets the "trailer present" and "trailer absent" states of the guard both get exercised honestly in one diff.
- **`/release-sync`'s misleading testing step is also fixed** — Rex flagged, while reviewing the sync PR (#1005) for this same release, that its step 2 asserts `git log upstream/main..upstream/dev --oneline` "shows only commits newer than `<version>`". It doesn't and structurally can't, for the identical reason above. Replaced with the assertion that's actually meaningful: `git merge-base --is-ancestor <release-squash-sha> upstream/dev` exits 0.

## Why this matters

Every future `/release` run was one bogus warning away from an operator learning to click through the `#872` guard on reflex — which is worse than not having the guard at all, since the guard exists specifically to catch a *real* truncated-changelog regression. And every future changelog was silently one release away from losing its deterministic range and falling back to the exact `#737`/`#872` failure modes AgDR-0094 was written to close. Both bugs were "invisible until the next release" by design, which is precisely why they went unnoticed through the v5.1.0 and v5.2.0 cuts.

## Verification

Ran against real `upstream/main`/`upstream/dev` history (not synthetic):

```
$ git rev-list --count upstream/main..upstream/dev        # OLD guard input
402
$ git rev-list --count dc26b8c..upstream/dev               # NEW guard input (trailer-anchored)
4
```

`dc26b8c` is the actual `Released-From` sha recorded on the v5.2.0 squash commit (`88bbd1d`) — confirmed present via `git log -1 --pretty=format:'%(trailers:key=Released-From,valueonly)' 88bbd1d`. (That trailer *did* land for v5.2.0 only because the PR was merged with a manual `--body` override at the time — the exact workaround this PR now codifies into the skill instead of leaving as tribal knowledge.)

Test suite: `.claude/hooks/tests/test_release_changelog.sh` — 58/58 passing (51 pre-existing + 7 new). `.claude/hooks/tests/test_release_sync.sh` — 14/14 passing (unaffected; `release-sync/SKILL.md`'s Testing-section fix is markdown-only, no script change).

**Gap I could not verify locally:** the actual `gh pr merge --squash --subject/--body-file` merge behavior against GitHub's live squash-body construction — I did not merge anything (per the task constraints). The fix is grounded in the documented `gh pr merge` flag semantics and the `dc26b8c` precedent above, but its first live exercise will be whichever release PR merges next; CI (the `auto-tag-on-release-pr-merge.yml` post-merge ancestry guard) is an unrelated, orthogonal check and does not itself verify the trailer.

## Testing (for reviewers)

1. `bash .claude/hooks/tests/test_release_changelog.sh` → 58/58
2. `bash .claude/hooks/tests/test_release_sync.sh` → 14/14
3. `git rev-list --count upstream/main..upstream/dev` vs. the trailer-anchored range on `upstream/dev` — confirms the guard's old input was structurally unusable
4. Read `.claude/skills/release/SKILL.md` rules 9–12 (new) for the condensed statement of both fixes

Refs #1004

---

## Glossary

| Term | Definition |
|------|------------|
| Release-cut branch model | AgDR-0007's pattern: `dev` takes all daily work; `main` only receives squash merges of release PRs, each tagged with a semver. |
| Squash merge | GitHub collapses every commit on a PR branch into one commit on the base branch; the branch's individual commits are never reachable from that squash commit. |
| `squash_merge_commit_message` | A GitHub repo setting controlling what a squash commit's *body* defaults to — `COMMIT_MESSAGES` (the branch's own commits, this repo's setting), `PR_BODY`, or `BLANK`. Only a default: explicit `--body`/`--body-file` on the merge call overrides it. |
| `Released-From` trailer | A git trailer (`Key: value` as the commit message's final paragraph) recording the exact `dev` SHA a release was cut from — lets the next release's changelog range be read back deterministically instead of inferred (AgDR-0094). |
| Count-mismatch guard | The `/release` sanity check (AgDR-0094 option D) comparing the changelog's entry count against a raw commit count, to catch a silently truncated changelog range (`#872`). |
| `LOG_RANGE` | The commit range `bin/release-changelog.sh` actually generates the changelog from — trailer-anchored when available, else the `#737` sync-boundary heuristic. Now exposed on stderr so the guard can reuse the same range instead of re-deriving (and getting) a different one. |
| Sync-boundary heuristic (`#737`) | The pre-AgDR-0094 fallback: infer the previous release's cut point from the most recent `sync/main-to-dev-after-<tag>` marker commit on `dev`. Imprecise when that sync lands late — the exact failure AgDR-0094's trailer exists to avoid. |

### DIFF
--- a/.claude/hooks/tests/test_release_changelog.sh
+++ b/.claude/hooks/tests/test_release_changelog.sh
@@ -345,6 +345,53 @@ contains "script exits cleanly despite the bogus sha" "EXIT_CODE=0" "$out"
 contains "falls back to sync heuristic (post-sync work included)" "unreleased work after late sync" "$out"
 not_contains "falls back to sync heuristic (pre-sync work excluded, same as #737 fallback)" "unreleased work before late sync" "$out"
 
+# ── Test: #1002 — release_desc join has a space after every comma ───────────
+# Regression test for the v5.2.0 cut's "2 features,13 fixes,14 improvements."
+# (no space after the commas) — `${arr[*]}` with IFS=', ' only joins on the
+# first IFS char, silently dropping the space.
+echo "--- #1002 release_desc comma spacing ---"
+out=$(run_test '
+  mc "chore: initial"
+  git tag v13.0.0
+  mc "feat(#1301): first feature"
+  mc "feat(#1302): second feature"
+  mc "fix(#1303): first fix"
+  mc "chore(#1304): first chore"
+  PREV_TAG="v13.0.0" HEAD_REF="HEAD" VERSION="v13.1.0" DATE="2026-07-24" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains "space after first comma" "2 features, 1 fix" "$out"
+contains "space after second comma" "1 fix, 1 improvement" "$out"
+not_contains "no unspaced comma before 'fix'" "features,1" "$out"
+not_contains "no unspaced comma before 'improvement'" "fix,1" "$out"
+
+# ── Test: #1002 — RELEASE_CHANGELOG_RANGE is printed to stderr, not stdout ──
+# The /release skill's count-mismatch guard needs the exact range the script
+# resolved (trailer-anchored or #737-fallback) to compare against, instead of
+# the structurally-wrong `main..dev` (#1002). Verify it is emitted on stderr
+# (so it never pollutes the changelog markdown on stdout) and matches the
+# trailer-anchored range when a trailer is present.
+echo "--- #1002 RELEASE_CHANGELOG_RANGE on stderr ---"
+stderr_capture=$(mktemp)
+out=$(run_test '
+  mc "chore: initial"
+  CUT=$(git rev-parse HEAD)
+  git commit -q --allow-empty -m "chore: release v14.0.0" -m "Released-From: $CUT"
+  git tag v14.0.0
+  mc "feat(#1401): unreleased feature"
+  stdout_out=$(PREV_TAG="v14.0.0" HEAD_REF="HEAD" VERSION="v14.1.0" DATE="2026-07-24" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>"'"$stderr_capture"'")
+  echo "STDOUT_HAS_RANGE_LINE=$(echo "$stdout_out" | grep -c "RELEASE_CHANGELOG_RANGE=" || true)"
+  echo "EXPECT_SHA=$CUT"
+')
+stderr_out=$(cat "$stderr_capture")
+rm -f "$stderr_capture"
+contains "range line NOT on stdout" "STDOUT_HAS_RANGE_LINE=0" "$out"
+contains "range line present on stderr" "RELEASE_CHANGELOG_RANGE=" "$stderr_out"
+# Cross-check the printed range is anchored on the trailer sha, not main..dev.
+expect_sha=$(echo "$out" | grep -oE 'EXPECT_SHA=.*' | cut -d= -f2)
+contains "stderr range is anchored on the trailer sha" "RELEASE_CHANGELOG_RANGE=${expect_sha}..HEAD" "$stderr_out"
+
 # ── Summary ──────────────────────────────────────────────────────────────────
 
 echo ""
--- a/.claude/skills/release-sync/SKILL.md
+++ b/.claude/skills/release-sync/SKILL.md
@@ -179,11 +179,11 @@ See [#403](https://github.com/me2resh/apexyard/issues/403) for full root-cause a
 ## Testing
 
 1. After merging, verify: `git log upstream/dev..upstream/main --oneline` returns empty
-2. Verify: `git log upstream/main..upstream/dev --oneline` shows only commits newer than <version>
+2. Verify ancestry, not `main..dev`'s size: `git merge-base --is-ancestor <release-squash-sha> upstream/dev` exits 0. (#1002 — `git log upstream/main..upstream/dev --oneline` does **not** shrink to "only commits newer than \<version\>"; under the release-cut squash model no individual `dev` commit is ever reachable from `main`, so that range accumulates every past release's un-squashed history and stays large — currently ~400 commits — even on a perfectly-synced repo. The merge-base check is the assertion that actually proves the sync worked.)
 3. Verify CHANGELOG is in sync: `diff <(git show upstream/main:CHANGELOG.md) <(git show upstream/dev:CHANGELOG.md)` returns empty (apexyard#448)
 4. Open a test release PR from dev → main — confirm only new work appears in the diff and the next `/release` v<next-version> prepends cleanly on top of <version>
 
-Refs #403, #448
+Refs #403, #448, #1002
 
 ---
 
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -61,15 +61,21 @@ Override? [Enter to accept, or type a version like v1.3.0]
 
 ### 3. Generate the CHANGELOG draft
 
-Call the helper script `bin/release-changelog.sh`, which encapsulates the `git log` + conventional-commit grouping + PR-number extraction logic and is independently tested. Capture its output — the count-mismatch guard below needs it:
+Call the helper script `bin/release-changelog.sh`, which encapsulates the `git log` + conventional-commit grouping + PR-number extraction logic and is independently tested. Capture both its stdout (the changelog) and stderr (the resolved commit range) — the count-mismatch guard below needs both:
 
 ```bash
 CHANGELOG_DRAFT=$(PREV_TAG="vX.Y.Z" \
   HEAD_REF="upstream/dev" \
   VERSION="vA.B.C" \
   DATE="$(date +%F)" \
-  bash bin/release-changelog.sh)
+  bash bin/release-changelog.sh 2>/tmp/release-changelog-range.txt)
 echo "$CHANGELOG_DRAFT"
+
+# The script also prints the exact range it generated the changelog from —
+# trailer-anchored (`<Released-From sha>..HEAD_REF`) when PREV_TAG carries a
+# Released-From trailer, or the #737 sync-boundary heuristic otherwise. The
+# count-mismatch guard below MUST compare against this range, not main..dev.
+LOG_RANGE=$(grep -oE 'RELEASE_CHANGELOG_RANGE=.*' /tmp/release-changelog-range.txt | cut -d= -f2-)
 ```
 
 The helper emits markdown to stdout in the format:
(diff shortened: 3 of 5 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1012

**Commit**: `33ec1e0912e6d19d004047f729bdf000b5270936`

### Summary

Two defects in the AgDR-0094 release-provenance mechanism, fixed together because they are the same mechanism and the same file. The count-mismatch guard is re-anchored from the structurally-unusable `upstream/main..upstream/dev` to `$LOG_RANGE` (now surfaced by `bin/release-changelog.sh` on stderr), and the `Released-From` trailer is moved into the release branch's own commit message with an explicit `--subject`/`--body-file` squash invocation plus a hard post-merge verification.

The root-cause analysis is correct on both counts. Under a squash release-cut model `main` receives exactly one commit per release, so no individual `dev` commit is ever reachable from `main` and `main..dev` is monotonically non-decreasing — comparing an entry count against it is comparing against a number that only grows. That is a genuine structural bug, not a tuning problem. The same wrong assertion in `/release-sync` step 2 is fixed consistently, using `git merge-base --is-ancestor` — the assertion that actually proves the sync worked, and the same one AgDR-0007 relies on elsewhere. Not a second divergent approach.

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass (CI green at this SHA; 7 new regression tests including a stdout/stderr separation assertion)
- Security: Pass (no auth/crypto/secrets surface; the `bin/` script change is trust-adjacent but introduces no new input handling)
- Performance: N/A
- PR Description and Glossary: Pass — narrative bullets, an explicit "why this matters", a stated unverified gap, and a 7-row Glossary
- Technical Decisions (AgDR): Pass — no new material decision; this implements AgDR-0094's existing option D and records the rejected alternative (`squash_merge_commit_message=PR_BODY`) inline with blast-radius reasoning
- Adopter Handbooks: N/A (none loaded — no `handbooks/architecture`, `handbooks/general`, or diff-matched language bucket)
- Fallow Static Analysis: N/A (no JS/TS in diff)

### Issues Found

None blocking. Three findings below, in priority order — the first is the one I would want addressed in a follow-up.

**1. The re-anchored guard no longer detects the failure mode its own warning text names (should-fix, non-blocking).**

I traced the fallback chain in `bin/release-changelog.sh:60-110`: trailer, then the `#737` sync-boundary, then merge-base, then the raw tag. It is never empty and never degenerate, so the guard always has a real range to count. The trailer-absent case falls back safely rather than silently passing. Good — that was the first thing I checked.

But the guard's *discriminating power* changed shape. `RAW_COUNT` and `ENTRY_COUNT` are now both derived from `$LOG_RANGE` — the range the changelog was generated from. If `$LOG_RANGE` is itself truncated (the `#872` late-sync signature, which is exactly what the fallback path is vulnerable to), the changelog shrinks and `RAW_COUNT` shrinks *in lockstep*, the gap stays small, and the guard stays silent. What it now measures is classification drop-out — commits inside the resolved range that did not become entries — which is a real and useful check, but a different one.

So the warning text at `SKILL.md:132-133` and rule 10 at `SKILL.md:221` are now inaccurate:

> `This is the #872 signature — a truncated changelog range.`

Structurally it can no longer see a truncated range, because it measures inside the range.

To be clear, this is **not** a regression in delivered capability, which is why it is not blocking. The old guard fired unconditionally — sensitivity 100%, specificity 0%, i.e. a constant, carrying zero information. The framing in the PR body ("one bogus warning away from an operator learning to click through") is right. And the coverage gap is genuinely closed by the *other* half of this PR: with a valid trailer, `$LOG_RANGE` is the exact cut point and truncation is impossible by construction; #1004 makes the trailer mandatory and adds a hard STOP if it is missing. That is a coherent trade — replace a noisy detector with a deterministic anchor plus loud verification.

Two concrete suggestions:

- Re-word the warning and rule 10 to describe what it now detects (entries dropped from the resolved range), and point the truncation concern at the trailer verification instead. As written, a future maintainer will trust it for something it cannot do.
- Consider having `bin/release-changelog.sh` also emit the *provenance* alongside the range — `RELEASE_CHANGELOG_ANCHOR=trailer|sync|merge-base|tag`. The script already knows which branch it took. The skill could then warn when the range was *inferred* rather than *recorded*, which is precisely the condition under which truncation is still possible. That restores coverage of the original failure mode at near-zero cost.

**2. The `$LOG_RANGE` parse fails open (should-fix, non-blocking).**

`LOG_RANGE=$(grep -oE 'RELEASE_CHANGELOG_RANGE=.*' ... | cut -d= -f2-)` — if the grep matches nothing, `LOG_RANGE` is empty, `git rev-list --count ""` errors, `RAW_COUNT` is empty, and:

- `SKILL.md:124`: `GAP=$(( RAW_COUNT - ENTRY_COUNT ))` — arithmetic on an empty operand, so `GAP` is negative or zero and the guard silently does not fire.
- `docs/release-process.md:292`: `GAP=$(( RAW_COUNT - ${ENTRY_COUNT:-0} ))` — the default guard is on the *wrong* variable; `ENTRY_COUNT` is already defended, `RAW_COUNT` is not.

Silent-pass in a guard is the exact defect class this PR exists to eliminate. Cheap fix — assert before use:

```bash
[ -n "$LOG_RANGE" ] || { echo "ERROR: could not resolve LOG_RANGE from release-changelog.sh stderr" >&2; exit 1; }
```

On the stderr contract itself: I checked, and `bin/release-changelog.sh` writes to stderr in only two other places (lines 48-49), both on the fatal missing-env-var path where the script exits before reaching the range echo. So in any successful run stderr carries exactly one line and the parse cannot be corrupted by interleaving. `cut -d= -f2-` correctly tolerates `=` in the value. The contract is sound as far as this script controls it; the residual exposure is stray stderr from a git invocation, and the anchored `grep -oE` plus the fail-closed assertion above would cover that. The test at `test_release_changelog.sh:29-54` asserting the line is on stderr and *not* stdout is the right test to have written.

**3. `ENTRY_COUNT` can over-exclude by one in a narrow case (nit).**

`grep -vcE '^- Closes '` excludes the trailing summary line correctly — I confirmed the emitted format at `bin/release-changelog.sh:267,274` is exactly `- Closes $closes_str`, and that real entries render as `- ($pr) subject — sha` or `- subject — sha` (lines 177, 182). A commit with no PR number whose subject begins "Closes " would render as `- Closes the auth gap — abc123` and be wrongly excluded. One entry, well inside the tolerance of 5. Not worth code; mentioning it so it is not rediscovered later as a mystery.

### On the two incidental fixes — real, correct, and correctly scoped

Both are genuine and neither is scope creep.

The `IFS` bug is real bash semantics: `${arr[*]}` joins on the **first character of `IFS` only**, so `IFS=', '` yields `"2 features,13 fixes"`. The explicit accumulation loop is the right fix — clearer than the alternatives, and it removes the `IFS` save/restore dance entirely. The `ENTRY_COUNT` off-by-one is real: the old `grep -cE '^- '` counted the `Closes` summary as an entry on every single run.

Folding both in was the right call. They live in the same neighbourhood as the guard being fixed, they *both* feed the guard's own inputs (the off-by-one directly skews `ENTRY_COUNT`), and each ships with regression tests. Splitting them would have produced two conflicting PRs against the same lines for no reviewer benefit. The same reasoning applies to the two-ticket scope: the tickets are one mechanism, the guard's fallback path only got exercised *because* the trailer was missing, and both states now get honestly exercised in one diff.

### What remains unverified — stated plainly

The author says they could not verify the live squash behaviour with the explicit `--subject`/`--body-file` flags, because merging was off-limits. I did not attempt it either. Assessing the reasoning rather than the outcome:

The reasoning is **sound**. `--subject` and `--body-file` are documented flags that override the server-side default, and `squash_merge_commit_message` is explicitly a *default*, not a constraint — the Glossary says so correctly. The fix is also belt-and-braces: even without the explicit flags the trailer would carry through, because step 4 now puts it in the branch's sole commit and `COMMIT_MESSAGES` builds the body from exactly that. The explicit flags are the second layer, defending against a stray fixup commit changing the branch's commit count or a future settings change. The `dc26b8c`/`88bbd1d` precedent is real supporting evidence that an explicit body override lands the trailer.

What stays unverified: whether GitHub's squash-body construction preserves the final paragraph such that `git interpret-trailers` parses it, end to end, in the live path. The mitigation is that this is exactly what the new mandatory post-merge check catches — and it `exit 1`s rather than warning, so a failure is loud at the moment it happens rather than invisible until the next cut. That is the correct place for the residual risk to land. First live exercise is the next release PR.

One thing worth watching on first use: `/tmp/release-pr-body.md` becomes the squash commit body verbatim, so the resulting commit message will carry the `<!-- multi-close: approved -->` comment and the full changelog. Harmless, and the trailer stays in the final paragraph, but it makes for a large commit message — flagging it so it is not mistaken for a bug the first time someone sees it.

### Suggestions

- Re-word the guard warning and rule 10 to match what the guard now detects (finding 1).
- Add the `LOG_RANGE` non-empty assertion in both `SKILL.md` and `docs/release-process.md` (finding 2), and move the `:-0` default in `docs/release-process.md:292` onto `RAW_COUNT` where it is actually needed.
- Optional but high value for the cost: emit `RELEASE_CHANGELOG_ANCHOR` alongside the range so the skill can warn when the anchor was inferred rather than recorded.

### Verdict

**APPROVED**

Both root causes are correctly diagnosed, the fixes are proportionate, the `/release-sync` change is consistent with the same analysis rather than a divergent one, the incidental fixes are real and tested, and the unverifiable gap is disclosed honestly and defended with a loud post-merge check. The findings above are follow-up work, not blockers — none of them leaves the release path worse than it is today, and finding 1 in particular describes a guard that was already carrying zero information before this change.

---
Reviewed commit: `33ec1e0912e6d19d004047f729bdf000b5270936`

Issue #1015, opened 2026-07-25
[Chore] AgDR-0104 trigger fired: third independent trust-chain bypass — activate the deferred CI meta-gate


atlas-apex:
## Driver

**AgDR-0104's deferral trigger has fired.** That decision deferred the anti-rot CI meta-gate with an explicit, testable condition:

> "**Anti-rot as a CI meta-gate** (a build failure if any trust-chain hook lacks a paired adversarial test) is the right fixed point — but **deferred** until a third *independent* incident proves the class recurs (tonight's n=2 are correlated, same degraded session)."

The n=2 at the time were #962 (marker-write gate evaded via shell-variable indirection) and #965 (merge gate failing open when `jq` was absent) — both in one API-degraded session, hence "correlated."

**The third incident is now on the board, and it is independent.** During the review of PR #1011 (itself the fix for #1000, itself a follow-up to #962/#970), the code reviewer constructed and confirmed **two previously-unknown bypasses of the same gate**:

1. **Interpreter heredoc bypass.** `python3 <<EOF` / `node <<EOF` / `bash <<EOF` with the write in the body: the old hook blocked it, the new logic allowed it. `_strip_heredoc_bodies` treats a heredoc body as data, but for an interpreter the body is *code naming the destination* — and interpreter writes yield no extractable target, so the text fallback then ran with the only evidence already removed.
2. **Mis-extracted target suppressing the fallback.** BSD-form `sed -i '' s/a/b/ <file>` caused the extractor to return the sed *expression* as the target; being literal and non-matching it was treated as conclusive and the fallback was skipped. Demonstrated rewriting a stale sign-off into a valid current-HEAD one — defeating the "new commits invalidate approval" property the gate depends on. GNU form unaffected, so it reproduces only on macOS, which is where this repo runs.

Different session, not API-degraded, different root causes from #962/#965. That satisfies "third independent incident" on its own terms.

**AgDR-0104 named this exact failure in advance.** Its list of unbounded path expressions — "`$VAR`, `$(…)`, concat, **here-doc**, symlink, `printf`" — contains the very construct that produced bypass 1. The class recurred precisely where it was predicted.

## Why this keeps happening (the pattern the trigger was built to catch)

Three rounds of fixes on one hook, and the third round's fix introduced two new holes:

| Round | Fix | Outcome |
|---|---|---|
| #962 → PR #970 | Narrowed from literal path matching to resolved targets | Incomplete — false positives persisted |
[...]

PR #1016, opened 2026-07-25
docs(#1015): AgDR-0109 — marker-write gate is a backstop; activate AgDR-0104 deferred meta-gate

atlas-apex:
## Summary

- **AgDR-0104's deferral trigger has fired, and this records what follows.** That decision deferred its anti-rot CI meta-gate "until a third *independent* incident proves the class recurs" — its n=2 were correlated, from one API-degraded session. Reviewing PR #1011 (the fix for #1000, itself a follow-up to #962/#970) surfaced **two previously-unknown bypasses of the same gate**, in a different, undegraded session. One is the `here-doc` shape AgDR-0104 had already listed by name among the unbounded ways to express a path. The class recurred exactly where it was predicted, so the trigger is satisfied on its own stated terms.

- **Labels `warn-review-marker-write.sh` a backstop, not THE control** — a label AgDR-0104 required for every trust-chain hook but which was never applied to this one. Its own reasoning forces the answer: a hook that pattern-matches bash command text "cannot be made sound," so integrity cannot rest on it. The control is the per-PR human approval plus the merge gate's SHA comparison against forge-reported HEAD. Recording the label means the next contributor inherits the framing instead of re-deriving it on round four.

- **Narrows blocking to confidently-resolved targets; ambiguity warns instead.** Every false positive observed today — roughly fourteen in one session — shared one signature: the hook reported `detected role: unresolved, pr: unresolved`, blocking while unable to identify any marker at all. It blocked a read-only `grep` whose *pattern string* contained the trigger words, an `rm` cleanup, an `echo` whose prose mentioned them, the sanctioned reviewer's own sign-off write with the session marker correctly set, and a command that merely declared a variable named `MARKER_HOME` — the name the framework's own agent spec recommends. Failing open on ambiguity is correct *for a backstop* and wrong only for THE control, which this is not. The #843 induced-impersonation case — a build agent naming a marker path literally — stays fully blocked.

- **Activates the deferred CI meta-gate:** a trust-chain hook must ship with a paired *adversarial* test that attempts bypass shapes, not one that asserts the happy path. Each of the three prior rounds shipped tests covering the shapes its author had thought of, which is definitionally the set that does not contain the next bypass. That is why the same hook has now been "fixed" three times and the third fix introduced two new holes.

- **Deliberately does not re-open settled ground.** AgDR-0062 already considered and rejected structuring the Rex marker ("a determined build agent can still type out the fields; gives false confidence") and deferred requiring a posted forge review (unsatisfiable in a single-account setup, where it would block every merge). The exposure AgDR-0062 accepted is re-affirmed here **deliberately**, with today's evidence, rather than inherited by default.

## Why this is a decision record and not a bug fix

The two live bypasses are already closed by PR #1011 round 2. This PR changes no code. It exists because the *pattern* — three rounds of narrowing on one hook, each shipping tests, the third introducing two fresh bypasses while the same heuristic blocked legitimate work fourteen times — is not something another patch fixes. AgDR-0104 anticipated exactly this and set a trigger. Landing the record is what stops round four being discovered rather than prevented.

## Testing

Documentation only — no code, no hooks, no gate logic touched. Verify by reading `docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md` against the two decisions it builds on:

1. `docs/agdr/AgDR-0104-trust-chain-controls-vs-backstops.md` § Consequences — confirm the deferred CI meta-gate names "a 3rd independent bypass incident" as its trigger.
2. `docs/agdr/AgDR-0062-rex-marker-authenticity.md` § Options Considered — confirm the structured-marker and posted-review options were rejected there, so this record re-affirms rather than re-litigates them.

Refs #1015

---

## Glossary

| Term | Definition |
|------|------------|
| Trust chain | The hooks enforcing security-critical SDLC properties — merge gates, review-marker writes, the tracker and marker libraries |
| Control vs backstop | AgDR-0104's distinction: *THE control* sits where it can see structured state (e.g. a forge-reported HEAD SHA); a *backstop* is a best-effort local nudge sitting behind a real server-side gate |
| Adversarial test | A test that actively tries to defeat a control, rather than confirming it works on inputs its author already had in mind |
| Fail-open / fail-closed | Whether a gate that cannot evaluate its precondition allows or denies. Failing open is defensible for a backstop and not for a control |
| Induced impersonation | The #843 failure: a build agent, unable to spawn the real reviewer, writes the reviewer's approval marker itself — satisfying the gate's filename while defeating its intent |
| Structured self-review | One identity both authoring and reviewing, with a recorded audit trail — a real property, but strictly weaker than separation of duties, which needs a second identity |

### DIFF
--- a/docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md
+++ b/docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md
@@ -38,6 +38,8 @@ AgDR-0104 deferred its CI meta-gate pending "a third *independent* incident" —
 
 Different session, not degraded, different root causes. **`here-doc` appears verbatim in AgDR-0104's list of unbounded expressions** — the class recurred exactly where it was predicted. The trigger has fired.
 
+**On the count, taken at its strictest.** A reader could argue AgDR-0104's own "n=2 are correlated" collapses #962 and #965 into a single data point, making today only the *second* independent incident rather than the third. Two things answer that. First, today supplies **two** further bypasses (interpreter heredoc and BSD `sed -i`) with distinct root causes, so even on the collapsing reading the count reaches three. Second, and more to the point, the deferral's stated purpose was to wait until the class was *proven to recur* rather than to reach an arbitrary tally — and a bypass landing on a construct AgDR-0104 had already listed by name, inside the fix for the previous bypass, is stronger proof of recurrence than a third unrelated incident would have been. The condition is met on either reading.
+
 ### The cost being paid meanwhile
 
 In one session the blocking behaviour produced roughly **fourteen false positives**: a read-only `grep` whose *pattern string* contained the trigger words; an `rm` cleanup; an `echo` whose prose mentioned them; the sanctioned reviewer's own sign-off write **with the session marker correctly set**; and a command that merely declared a variable named `MARKER_HOME` — the name the framework's own agent spec recommends. Reviewers worked around it by base64-encoding literals and splitting filenames across `printf` calls.
@@ -64,7 +66,15 @@ Chosen: **E.** Three parts, in order.
 
 **2. Block only on a confidently-resolved marker target; warn on ambiguity.** When the hook resolves a write target to a marker path, block — that is the #843 shape and it works. When it cannot resolve role and PR (`detected role: unresolved, pr: unresolved`), emit the advisory banner and exit 0.
 
-This is not a weakening dressed as a fix. Failing open on ambiguity is *correct for a backstop* and wrong only for THE control — which, per part 1, this is not. It converts every false positive observed today into a warning while leaving the case that motivated blocking fully gated. A build agent writing a literal marker path is still blocked; a reviewer grepping for the word is not.
+This is not a weakening dressed as a fix. It converts every false positive observed today into a warning while leaving the case that motivated blocking fully gated. A build agent writing a literal marker path is still blocked; a reviewer grepping for the word is not.
+
+**Reconciling this with AgDR-0104's fail-closed instruction.** AgDR-0104 § Decision directs: *"Fix #962 and #965 fail-closed (a gate that can't evaluate its precondition must **block**, not allow)"* — and **#962 is this very hook**. Resting on the backstop label alone would be too thin, since the labelling requirement came from AgDR-0104 in the same breath; relabelling would then look like a way to escape its own instruction.
+
+The reconciliation is not the label, it is **what the precondition is about**. In #965 the merge gate could not evaluate whether the *protected action* was authorised, and allowed it anyway — an unverified merge proceeded. That is what "must block" governs: never admit the protected action without verifying it.
+
+This hook's ambiguity is a different question one level up. When it cannot resolve a role and PR, it does not know whether the command **is a marker write at all** — not whether an established marker write is authorised. Failing open there **declines jurisdiction**; it does not admit an unverified action. The evidence is unanimous: every false positive enumerated above — a `grep`, an `rm`, an `echo`, a bare variable declaration — was not a marker write. Blocking them protected nothing, because there was no protected action present to protect.
+
+So AgDR-0104's rule is preserved exactly where it applies. A command that *does* resolve to a marker target is the protected action, and it still blocks, fail-closed. Ambiguity about whether the action exists is not the same as ambiguity about whether it is authorised, and only the latter is what "must block, not allow" was written for.
 
 **3. Activate AgDR-0104's deferred CI meta-gate.** Its trigger has fired on its own stated terms. A trust-chain hook must ship with a paired **adversarial** test — one that attempts bypass shapes, not merely asserts the happy path. Seed it with the shapes now known: interpreter heredocs, BSD vs GNU `sed -i`, variable indirection (#962) — and the false-positive direction too, because a gate can fail by over-firing and today it mostly did.
 
@@ -84,5 +94,6 @@ The tests written in each of the three rounds all covered the shapes their autho
 - Ticket: me2resh/apexyard#1015
 - Live bypasses closed by PR #1011 (round 2) — this decision is about the gate's design, not those two fixes
 - Supersedes the deferral in [AgDR-0104](AgDR-0104-trust-chain-controls-vs-backstops.md) § Consequences ("CI meta-gate, trigger: a 3rd independent bypass incident")
+- **Qualifies** [AgDR-0104](AgDR-0104-trust-chain-controls-vs-backstops.md) § Decision step 1 — *"Fix #962 and #965 fail-closed (a gate that can't evaluate its precondition must block, not allow)"* — as it applies to **#962 / this hook only**. The instruction stands unchanged where the precondition concerns whether a *resolved* protected action is authorised; it does not extend to ambiguity about whether a protected action is present at all. See § Decision part 2 for the reasoning. **#965 and the merge gate are untouched by this qualification.**
 - Re-affirms [AgDR-0062](AgDR-0062-rex-marker-authenticity.md)'s accepted exposure and its rejection of the structured-marker and posted-review options
 - Related: [AgDR-0075](AgDR-0075-code-reviewer-local-marker-is-gate-signal.md) (the local marker is the gate signal), #962, #965, #843, #1000
--- a/docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md
+++ b/docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md
@@ -1,0 +1,88 @@
+---
+id: AgDR-0109
+timestamp: 2026-07-25T10:00:00Z
+agent: claude (orchestrator)
+model: claude-opus-5
+trigger: user-prompt
+status: executed
+---
+
+# The marker-write gate is a backstop, not the control — block only on a resolved target, and activate the deferred meta-gate
+
+> In the context of [AgDR-0104](AgDR-0104-trust-chain-controls-vs-backstops.md) deferring its anti-rot CI meta-gate "until a third *independent* incident proves the class recurs", facing exactly that third incident — two previously-unknown bypasses of `warn-review-marker-write.sh` found during review of the fix for the *previous* bypass — I decided to **label that hook a backstop per AgDR-0104, narrow its blocking behaviour to confidently-resolved marker targets while ambiguity warns instead of blocks, and activate the deferred CI meta-gate** — to achieve a gate that stops obstructing the flow it protects without re-opening the induced-impersonation hole that made it blocking, accepting that a determined actor can still reach the file locally because the real control is the per-PR human approval plus the merge-gate SHA check.
+
+## Context
+
+`warn-review-marker-write.sh` decides whether to block by pattern-matching the **text** of a Bash command to infer write intent. Its history is three rounds deep on one hook:
+
+| Round | Change | Outcome |
+|---|---|---|
+| [#962](https://github.com/me2resh/apexyard/issues/962) → PR #970 | Narrowed from literal path matching toward resolved targets | Incomplete; false positives persisted |
+| [#1000](https://github.com/me2resh/apexyard/issues/1000) → PR #1011 | Narrowed to actual write intent | **Introduced two new bypasses** |
+| PR #1011 round 2 | Closed both | Pending re-review |
+
+AgDR-0062 shipped this hook deliberately as **advisory — "exit 0 always"** — accepting that "the rex marker remains technically forgeable while every merge has a human-in-the-loop CEO nod." It was promoted to **blocking** in [#843](https://github.com/me2resh/apexyard/issues/843) after build-class sub-agents were observed writing their own approval markers (PRs #835, #842). That promotion addressed a real, exercised failure — it was not gratuitous.
+
+AgDR-0104 then established the framing this decision applies:
+
+> "a security gate implemented as regex/substring matching over bash command *text* **cannot be made sound** — the ways to express a path (`$VAR`, `$(…)`, concat, **here-doc**, symlink, `printf`) are unbounded."
+
+and required every trust-chain hook be labelled *THE control* or *a backstop to a server-side gate*. That labelling was never applied to this hook.
+
+### The third incident
+
+AgDR-0104 deferred its CI meta-gate pending "a third *independent* incident" — its n=2 (#962, #965) being correlated, from one API-degraded session. During review of PR #1011, the code reviewer constructed and confirmed two new bypasses:
+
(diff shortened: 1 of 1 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1016

**Commit**: `2721f90a33906008225e09c1c3367ff563472f86`

### Summary

Adds one file — `docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md` (88 lines, no deletions). No code, no hooks, no gate logic. Reviewed as a **decision record**, not as prose: I verified the deferral-trigger claim against AgDR-0104, the prior-decision representations against AgDR-0062, and the internal coherence of the central decision. Ceremony scoped to Lean per `.claude/rules/right-size-ceremony.md` — docs-only, additive, reversible — with the caveat that the *content* is trust-chain-adjacent, so the three substantive claims got full scrutiny.

### Checklist Results

- Architecture & Design:      N/A (no code)
- Code Quality:               N/A (no code)
- Testing:                    N/A (docs-only; CI green — `markdownlint-cli2` pass, hook suite pass)
- Security:                   Pass — records a trust-chain design direction; no control weakened by this diff (the code change already landed in #1011)
- PR Description & Glossary:  Pass — 6-term Glossary table present
- Summary Bullet Narrative:   Pass — every bullet answers what + why; well clear of the label-only heuristic
- Technical Decisions (AgDR): Pass — this PR *is* the AgDR; structure matches `templates/agdr.md`, numbering correct, cross-refs resolve
- Adopter Handbooks:          N/A — 3 always-load handbooks loaded, none with triggers matching a docs-only AgDR diff

---

### 1. Is the "trigger has fired" claim true? — Yes, honest, not stretched

AgDR-0104 § Decision (line 40) reads verbatim:

> "**deferred** until a third *independent* incident proves the class recurs (tonight's n=2 are correlated, same degraded session)."

The claim holds on the natural count: #962 = 1, #965 = 2, today's PR #1011 review = 3, and today is independent of both (different session, undegraded, different root causes, different mechanism).

What makes it more than bookkeeping — and the strongest element in the record — is that AgDR-0104 line 18 lists the unbounded expression shapes verbatim:

> "the ways to express a path (`$VAR`, `$(…)`, concat, **here-doc**, symlink, `printf`) are unbounded"

`here-doc` is one of the two new bypasses. The class recurred **at a point AgDR-0104 named in advance**, which is precisely what "proves the class recurs" asks for — stronger evidence than a third arbitrary incident would have been. Not stretched.

`suggestion:` One residual ambiguity the record should close. A stricter reading of the parenthetical — "n=2 are correlated" collapsing #962+#965 into a *single* independent data point — would make today only the **second** independent incident, not the third. I judge that parse strained (the parenthetical reads as a hedge explaining why n=2 is soft evidence, not as a redefinition of the count), and the predicted-point argument carries the claim either way. But the record does not address that reading, and it is exactly what a round-four contributor will re-litigate. One clause fixing the count explicitly costs nothing and forecloses it.

### 2. Does it re-litigate settled decisions? — No. Accurate, and re-affirming rather than reversing

Every quotation checked against source. All verbatim, none strengthened in transit:

| Claim in AgDR-0109 | Source | Verdict |
|---|---|---|
| Option C rejected: "a determined build agent can still type out the fields; gives false confidence" | AgDR-0062 § Options Considered, line 24 | Verbatim, complete |
| Option D deferred: unsatisfiable in single-account setups, "would block every merge" | AgDR-0062 lines 23, 29, 33 | Accurate |
| Hook shipped advisory, "exit 0 always" | AgDR-0062 line 59 | Verbatim |
| Accepted exposure: "the rex marker remains technically forgeable while every merge has a human-in-the-loop CEO nod" | AgDR-0062 line 3 | Verbatim |
| Promoted to blocking in #843 after PRs #835/#842 | AgDR-0062 + `pr-workflow.md` | Accurate history |

Nothing is quietly reversed. The handling of AgDR-0062's accepted exposure is the right pattern and worth calling out: it is re-affirmed **explicitly, with today's evidence, and labelled as a deliberate re-affirmation rather than an inheritance by default**. That is what a record building on a prior one should look like. Option B (full revert to advisory) is included and rejected on real #843 grounds rather than strawmanned.

### 3. Is the central decision internally coherent? — Reconciled explicitly, but under-argued in one place

This is the finding worth your attention.

The tension is real and sharper than the record admits. AgDR-0104 § Decision line 38 says:

> "**Fix #962 and #965 fail-closed** (a gate that can't evaluate its precondition must *block*, not allow)"

**#962 is this hook.** So AgDR-0104 directed fail-closed for this specific hook, and AgDR-0109 part 2 now directs fail-*open* on ambiguity for it.

On the question directly: **it does not paper over this.** A dedicated paragraph names the tension head-on — "This is not a weakening dressed as a fix. Failing open on ambiguity is *correct for a backstop* and wrong only for THE control — which, per part 1, this is not." Anyone who has read AgDR-0104 will recognise which line is being addressed. So it is under-cited, not quiet.

But resting the reconciliation **solely on the control/backstop label** is the weaker available argument, for a reason the record misses: AgDR-0104 issued the fail-closed directive *in the same breath* as the labelling requirement and the "relocate real trust to the server-side gate" instruction. The label is not news to AgDR-0104 — it was AgDR-0104's own instruction. So "it is a backstop, therefore fail-open is fine" reads as overturning line 38 by relabelling.

The decisive argument is available and absent — a **precondition-type distinction**:

- **#965's fail-open** — the merge gate could not evaluate "is there a valid approval for the action about to happen," and it let the **protected action** (the merge) proceed. An unverified protected action executing is exactly what "must block, not allow" governs. AgDR-0104 line 38 is right about that.
- **This hook's ambiguity** — the hook cannot determine whether the command **is the protected action at all**. Failing open does not let an unverified protected action through; it declines jurisdiction over a command it cannot classify. Every false positive enumerated in the record confirms this: a read-only `grep`, an `rm` cleanup, an `echo`, a bare ops-root variable declaration. None were marker writes.

Those are different failure modes, and only the first is what line 38 addresses. Two sentences to that effect would convert an apparent reversal into a principled narrowing and make part 2 airtight.

`suggestion:` Add the precondition-type distinction to Decision part 2, cite AgDR-0104 line 38 by name when doing so, and note in § Artifacts that line 38's fail-closed directive is being *qualified* for this hook — not merely that the deferral is superseded. Non-blocking, but the single highest-value edit available: a future contributor reading only the Artifacts list would not learn that line 38 was narrowed.

### What the record gets right

Worth stating, since the above concentrates on one gap:

- **The false-positive evidence is specific and falsifiable**, not a convenience complaint. The common signature — every one of the ~14 reporting `detected role: unresolved, pr: unresolved` — is what justifies narrowing at exactly that point rather than generally, and it is load-bearing for part 2. I independently confirmed the variable-name claim: the ops-root variable the record cites does appear in `.claude/agents/code-reviewer.md` (2 occurrences), so "the gate fires on the framework's own documented variable name" is accurate.
- **"A gate that blocks its own review pressures reviewers into obfuscation habits indistinguishable from evasion"** is a genuine security argument, not an ergonomics one. Correct framing, and it is what elevates this above "the hook annoyed me."
- **Part 3's diagnosis is correctly aimed**: "the tests written in each of the three rounds all covered the shapes their author had thought of, which is definitionally the set not containing the next bypass." That is the actual root cause of three rounds, and seeding the adversarial test in *both* directions (bypass shapes and the over-firing direction) follows from the evidence.
- § Consequences names the accepted exposure plainly instead of burying it, and preserves the honest-naming correction (structured self-review plus an audit trail, not separation of duties).

### Mechanical checks

- **Structure** — matches `templates/agdr.md`: title, one-line "In the context of… facing… I decided… to achieve… accepting…" blurb, Context, Options Considered, Decision, Consequences, Artifacts. All present, in order.
- **Numbering** — 0109 correct; 0108 is the highest on `dev`. (Unrelated pre-existing observation: `docs/agdr/` holds two `AgDR-0095-*` files. Not introduced here and not this PR's problem — flagging only so it is on someone's radar.)
- **Cross-references** — `AgDR-0104-trust-chain-controls-vs-backstops.md`, `AgDR-0062-rex-marker-authenticity.md`, `AgDR-0075-code-reviewer-local-marker-is-gate-signal.md` all exist and resolve.
- **PR body** — Summary bullets narrative (what + why throughout); Glossary table present with 6 terms. Both pass `.claude/rules/pr-quality.md`.
- **CI** — 5 checks pass (`markdownlint-cli2`, `hook test suite`, `Verify Ticket ID`, `Analyze (actions)`); CodeQL skipped. Green.

### Suggestions

1. **(highest value)** Add the precondition-type distinction to Decision part 2, cite AgDR-0104 line 38 explicitly, and record in § Artifacts that line 38 is being qualified for this hook.
2. Add one clause fixing the incident count explicitly (#962 = 1, #965 = 2, today = 3) to foreclose the stricter parse of "n=2 are correlated."
3. `nit:` Frontmatter omits the optional `category:` field. `category: security` would make this discoverable via `/agdr browse --category security`; without it the record lands in the `other` bucket. The template says omission is safe, so this is cosmetic.
4. `nit:` AgDR-0104 § Consequences promised a one-line `rule-audit.md` guidance note ("trust-chain hooks are backstops to a server-side gate; fail-closed + adversarial regression test expected"). It never landed — a `trust-chain` grep of `docs/rule-audit.md` returns 0 hits. Pre-existing gap, not introduced here, but since AgDR-0109 supersedes the interim measure that note *was*, closing the loop in this record's Consequences (or a follow-up on #1015) would tidy it.

### Verdict

**APPROVED**

An unusually rigorous decision record. Every quotation I checked was verbatim and none was strengthened in transit; the deferral trigger genuinely fired on AgDR-0104's own stated terms, with the predicted-point recurrence making the case stronger than a bare third incident would have been; and no prior decision is reversed without saying so. The one real gap is that the central fail-open call is reconciled with AgDR-0104 line 38 by label alone when a decisive structural argument was available — under-argued, not papered over, so it does not block a docs-only record whose code side already merged in #1011. Suggestion 1 is worth taking as a fast-follow, at the author's discretion.

---
Reviewed commit: `2721f90a33906008225e09c1c3367ff563472f86`

atlas-apex (commented):
## Code Review: PR #1016 — delta re-review

**Commit**: `209c0b6419f3e2030038a7b16532de134bf98b48`
**Prior review**: APPROVED at `2721f90`. This pass covers only the delta.

### Summary

One commit, one file, +12/-1 — all of it in `docs/agdr/AgDR-0109-marker-write-gate-is-a-backstop.md`. The delta is argument, not decision: the three chosen parts of Option E are byte-identical to what I approved at `2721f90`. Verified: `git diff 2721f90 209c0b6 --stat` shows the single AgDR file and nothing else. No hook code, no config, no test files.

All three points I raised are answered, and each lands.

### Checklist Results
- Architecture & Design: N/A (documentation)
- Code Quality: N/A
- Testing: N/A
- Security: Pass — no executable change; the decision it records is unchanged from the approved base
- Performance: N/A
- PR Description & Glossary: Pass (carried forward)
- Technical Decisions (AgDR): Pass
- Adopter Handbooks: N/A (no handbooks loaded — docs-only diff)

### Point-by-point

**1. The AgDR-0104 fail-closed tension — resolved.**

The new § Decision reconciliation names its own weakness first ("resting on the backstop label alone would be too thin, since the labelling requirement came from AgDR-0104 in the same breath"), which is the right move — that was exactly my objection, and conceding it up front is what makes the rest credible.

The substantive argument holds. #965's fail-open let a *protected action* (a merge) proceed unverified; this hook's ambiguity means it cannot tell whether a protected action is present at all. Declining jurisdiction is genuinely not the same failure as admitting an unverified action, and the record states the boundary rather than gesturing at it.

On the over-extension test you asked me to apply — could a future reader use this to justify failing open where the protected action IS resolved? No. The record closes that door twice, explicitly: *"A command that does resolve to a marker target is the protected action, and it still blocks, fail-closed"* and *"only the latter is what 'must block, not allow' was written for."* The distinction is drawn on the right axis (presence vs. authorisation), not on the convenient one (noisy vs. quiet).

nit (advisory, non-blocking): the reconciliation's evidentiary step — *"the evidence is unanimous: every false positive enumerated above ... was not a marker write"* — is drawn from *accidental* ambiguity. It doesn't by itself cover *induced* ambiguity, which is precisely what #962 was (a marker path deliberately routed through a shell variable to become unresolvable). A reader working only from § Decision part 2 could miss that a hostile write can manufacture its own ambiguity. The document does answer this — Option E's cons column states it plainly: *"Ambiguous-but-genuine forgery attempts now warn rather than block — accepted, because this is a backstop and the real control lies elsewhere"* — so the record is coherent as a whole. The removed sentence used to carry that connective inside part 2; a one-clause pointer from part 2 back to Option E's accepted con would keep the reasoning self-contained where a future reader will actually land. Not worth another round on its own.

**2. § Artifacts framing — accurate.**

Checked against AgDR-0104's real text. Step 1 reads: *"Fix #962 and #965 fail-closed (a gate that can't evaluate its precondition must block, not allow)"* — the quote is verbatim. The "#962 is this very hook" claim is correct: AgDR-0104's own summary identifies #962 as *"the marker-write gate is evaded by passing the marker path through a shell variable instead of a literal."*

"QUALIFIED for this hook only" is the honest word — stronger than "superseded" would have been, since the instruction genuinely still binds everywhere else. And the "#965 and the merge gate are untouched" assertion is true: nothing in this PR alters the merge gate or #965's fail-closed fix. Scoping a qualification to one of the two hooks a shared instruction covers, and saying so in the artifact list where the next reader of 0104 will follow the link, is the right shape.

**3. The trigger count — honest, not hand-waving.**

The second answer is the load-bearing one and it is textually grounded: AgDR-0104 says the meta-gate is deferred *"until a third independent incident proves the class recurs."* "Proves the class recurs" is literally the stated criterion, so arguing that a bypass landing on a construct 0104 had already named by name — inside the fix for the previous bypass — is stronger proof of recurrence than an unrelated third incident is a reading of 0104's purpose, not a rewrite of it. Conceding the strict reading and then answering on its own terms is the opposite of hand-waving.

nit (advisory): the *first* answer is the weaker of the two. It collapses #962+#965 into one point but counts today's two bypasses as separate, reaching three. That's only consistent if "degraded" rather than "same session" is the load-bearing half of 0104's *"tonight's n=2 are correlated, same degraded session"* — which is a fair reading (the record does note today was "not degraded"), but the argument doesn't spell out why the collapsing rule applies asymmetrically. Applied symmetrically it would give two, not three. The argument survives without it, since the second answer disclaims the tally as the criterion at all; if anything the first answer slightly undercuts the second by appearing to bargain over a number the record then says was never the point.

### Issues Found

None blocking.

### Suggestions

Both nits above are optional and neither needs another review round. If you touch the file again for any reason, the part-2 → Option-E cross-reference is the one worth adding.

### CI

`markdownlint-cli2` pass, `Analyze (actions)` pass, `Verify Ticket ID` pass, CodeQL skipping. `hook test suite` still pending at review time — worth a glance before merge, though this diff touches no hook code, so a failure there would be unrelated to the change.

### Verdict

**APPROVED**

The delta does exactly what it set out to do: it supplies the missing argument without moving the decision. My approval from `2721f90` carries forward and is strengthened — the fail-closed reconciliation was the one thing standing between this record and being fully self-defending, and it now stands on the presence-vs-authorisation distinction rather than on the label.

---
Reviewed commit: `209c0b6419f3e2030038a7b16532de134bf98b48`

Issue #1017, opened 2026-07-25
[Bug] Release count-guard: warning text no longer matches what it detects, and the range parse fails open


atlas-apex:
## Driver

Two non-blocking findings from the code review of PR #1012 (merged), plus one cosmetic note. Both were judged acceptable to merge and both are worth closing properly rather than losing in a merged PR's review thread.

PR #1012 re-anchored the release count-mismatch guard away from `upstream/main..upstream/dev` — a range that structurally can never shrink under squash-merge, so the guard false-positived every run — onto the trailer-derived `LOG_RANGE` that `bin/release-changelog.sh` already computed internally but never surfaced.

## Finding 1 — the guard now detects a different thing than its warning claims

The re-anchoring works and the fallback chain is safe: the reviewer traced `bin/release-changelog.sh:60-110` (trailer → `#737` sync-boundary → merge-base → raw tag) and confirmed `LOG_RANGE` is never empty or degenerate, so the guard always has a real range to work with.

But its **discriminating power changed shape**, and the PR does not acknowledge this. `RAW_COUNT` and `ENTRY_COUNT` are now *both* derived from `$LOG_RANGE`. If that range is itself truncated — the `#872` late-sync signature, which is exactly what the fallback path is vulnerable to — both numbers shrink **in lockstep**, the gap stays small, and the guard stays silent.

So it now detects *classification drop-out inside the range* rather than *a truncated range*. Meanwhile the warning text still reads:

> "This is the #872 signature — a truncated changelog range."

which is the one thing it structurally can no longer see.

**Why this was not blocking:** the old guard fired unconditionally, making it a constant that carried zero information — so no real detection capability was lost. And the gap is genuinely closed by the *other half* of the same PR: a valid `Released-From` trailer makes truncation impossible by construction, and the #1004 work makes the trailer mandatory with a hard `exit 1` when it is missing. That is a coherent trade, not a regression dressed as a fix.

**Suggested fix** (from the review, cheap):

- Re-word the warning so it describes what the guard actually detects.
- Optionally emit `RELEASE_CHANGELOG_ANCHOR=trailer|sync|merge-base|tag` alongside the existing range, so `/release` can warn specifically when the anchor was **inferred** rather than **recorded**. That restores the original truncation coverage at the one moment it is still reachable — a release cut from a tag that predates the trailer convention.

## Finding 2 — the stderr parse fails open

[...]

PR #1027, opened 2026-07-26
fix(#1017): count-guard warning text + fail-open range parse

atlas-apex:
## Summary

- **Re-worded the count-mismatch guard's warning so it describes what it actually detects.** After #1012 re-anchored the guard onto `$LOG_RANGE`, both `RAW_COUNT` and `ENTRY_COUNT` derive from the same range — so a truncated range (the original `#872` failure) can no longer produce a gap between them; the two numbers shrink together. What the guard genuinely still catches is *classification drop-out inside a correct range* (a commit that's really in `$LOG_RANGE` but didn't turn into a changelog bullet). The warning text, the surrounding comments, and rule 10 all said "#872 signature — a truncated changelog range" — which is precisely the thing the guard can no longer see. Fixed in both `.claude/skills/release/SKILL.md` (the automated `/release` flow) and `docs/release-process.md` (the manual fallback runbook), so an operator reading the warning gets pointed at the right investigation (read the commits in the range, not re-check `PREV_TAG`/`HEAD_REF`).
- **Made an unresolved `$LOG_RANGE` fail CLOSED instead of failing open.** If `RELEASE_CHANGELOG_RANGE` can't be parsed from `bin/release-changelog.sh`'s stderr, `$LOG_RANGE` is empty, and `git rev-list --count ""` fails — but since nothing captured that failure, `$RAW_COUNT` silently became an empty string, and `GAP=$(( RAW_COUNT - ENTRY_COUNT ))` treated it as `0`, so the guard passed quietly on exactly the case it can't evaluate. Both scripts now check `$LOG_RANGE` for empty immediately after resolving it, and guard the `git rev-list` call with `|| { …; exit 1; }`, so a missing or malformed range now halts the release with a loud, actionable message instead of reading as "no gap." This matches AgDR-0104's rail: *a gate that can't evaluate its precondition must block, not allow.*
- **Did not add a `${RAW_COUNT:-0}` default.** The issue's own draft AC suggested giving `RAW_COUNT` "the same defensive default treatment as `ENTRY_COUNT`" — but a `:-0` default is exactly the fail-open bug in disguise: it would make a `git rev-list` failure look like "0 commits" and the guard would still pass silently. `ENTRY_COUNT`'s `:-0` default is legitimate (a genuinely empty changelog is a valid state); a failed `git rev-list` on `$LOG_RANGE` is not. Explicit exit-1 checks are used instead — see Finding 2 disposition below.

## Finding-by-finding disposition (from #1017)

1. **Warning text mismatch** — real, confirmed by reading `bin/release-changelog.sh`'s classify step and the guard's own comments. Fixed.
2. **Range parse fails open** — real, confirmed by reproducing it directly (see Testing). Fixed.
3. **Cosmetic squash-body finding** and the **optional anchor-kind surfacing** enhancement from #1017 are out of scope for this PR per the task brief ("keep it minimal, don't restructure the release flow") — left for a follow-up if wanted.

## Testing

Ran the fixed guard logic against this repo's real `upstream/main`/`upstream/dev` history (not a fixture), plus the two existing release test suites.

**1. Happy path — real trailer-anchored range, guard passes correctly:**

```
$ PREV_TAG=v5.2.0 HEAD_REF=upstream/dev VERSION=v99.99.99 DATE=$(date +%F) bash bin/release-changelog.sh > /tmp/cl.md 2>/tmp/range.txt
$ cat /tmp/range.txt
RELEASE_CHANGELOG_RANGE=dc26b8c986f6a9682b037ebb6c7fbba048f36259..upstream/dev

RAW_COUNT=9 ENTRY_COUNT=6 GAP=3
no warning — gap within tolerance, guard passes correctly
```

**2. Fail-closed path — empty `$LOG_RANGE`, fixed guard:**

```
LOG_RANGE=""
if [ -z "$LOG_RANGE" ]; then
  echo "ERROR: could not read RELEASE_CHANGELOG_RANGE ... Do NOT proceed." >&2
  exit 1
fi
→ ERROR: could not read RELEASE_CHANGELOG_RANGE — guard has nothing to check against. Do NOT proceed.
→ exit 1 (fail-closed confirmed)
```

**3. Old (pre-fix) behaviour on the same empty range, for contrast:**

```
LOG_RANGE=""
RAW_COUNT=$(git rev-list --count "$LOG_RANGE")   # fatal: ambiguous argument '' ... (stderr only)
GAP=$(( RAW_COUNT - ENTRY_COUNT ))                # RAW_COUNT treated as 0
→ NO WARNING — silent pass-through (the #1017 bug), exit 0
```

**4. Existing test suites — unaffected (no change to `bin/release-changelog.sh`):**

```
$ bash .claude/hooks/tests/test_release_changelog.sh
All 58 test(s) passed.

$ bash .claude/hooks/tests/test_release_sync.sh
Passed: 14
Failed: 0
```

## Glossary

| Term | Definition |
|------|------------|
| `LOG_RANGE` | The commit range `bin/release-changelog.sh` actually generated the changelog from — trailer-anchored (`Released-From` sha) when available, else the `#737` sync-boundary fallback. Printed to stderr as `RELEASE_CHANGELOG_RANGE=...` so the count-mismatch guard can reuse it. |
| Count-mismatch guard | The `/release` sanity check (AgDR-0094 option D) comparing the changelog's entry count against a raw commit count, meant to catch a silently truncated changelog range. |
| Classification drop-out | A commit genuinely inside `$LOG_RANGE` that `bin/release-changelog.sh`'s classify step failed to turn into a changelog bullet — what the guard now actually detects, as opposed to a truncated range. |
| Fail open / fail closed | A guard "fails open" when it silently allows on a precondition it couldn't evaluate (the `#1017` bug); it "fails closed" when it halts loudly instead — the direction AgDR-0104 requires for trust-chain-adjacent gates. |
| `Released-From` trailer | A git trailer recording the exact `dev` SHA a release was cut from (AgDR-0094) — makes `$LOG_RANGE` deterministic instead of inferred, which is why a truncated *range* is no longer the guard's blind spot. |

Refs #1017

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -7,7 +7,7 @@ allowed-tools: Bash, Read, Write
 
 # /release — Cut an apexyard release
 
-Standardises the `dev` → `main` release flow introduced by AgDR-0007. Reads the conventional-commit log between `main` and `dev`, proposes a semver bump, **generates and writes the CHANGELOG entry**, **opens the release PR** (dev→main), and triggers the `auto-tag-on-release-pr-merge` GitHub Actions workflow that tags the squash commit and creates a GitHub Release after merge. One command drives the operator from "nothing" to "PR open, ready for Rex + CEO". The tag and GitHub Release entry are created automatically by CI when the PR merges. The release PR also records a `Released-From` trailer with the exact `dev` cut-point SHA, and the changelog step warns loudly on a large main↔dev commit-count mismatch — both close the #872 changelog-truncation gap. Design rationale: AgDR-0076, AgDR-0094.
+Standardises the `dev` → `main` release flow introduced by AgDR-0007. Reads the conventional-commit log between `main` and `dev`, proposes a semver bump, **generates and writes the CHANGELOG entry**, **opens the release PR** (dev→main), and triggers the `auto-tag-on-release-pr-merge` GitHub Actions workflow that tags the squash commit and creates a GitHub Release after merge. One command drives the operator from "nothing" to "PR open, ready for Rex + CEO". The tag and GitHub Release entry are created automatically by CI when the PR merges. The release PR also records a `Released-From` trailer with the exact `dev` cut-point SHA, and the changelog step warns loudly if commits inside the resolved changelog range aren't all making it into entries — both guard against the #872 changelog-truncation failure mode. Design rationale: AgDR-0076, AgDR-0094.
 
 This skill is **framework-only** — it's for cutting apexyard releases, not for releasing managed projects under governance. Managed projects stay trunk-based and don't have a release-cut flow.
 
@@ -76,6 +76,16 @@ echo "$CHANGELOG_DRAFT"
 # Released-From trailer, or the #737 sync-boundary heuristic otherwise. The
 # count-mismatch guard below MUST compare against this range, not main..dev.
 LOG_RANGE=$(grep -oE 'RELEASE_CHANGELOG_RANGE=.*' /tmp/release-changelog-range.txt | cut -d= -f2-)
+
+# #1017: fail CLOSED here, not at the arithmetic below. An empty/unparseable
+# $LOG_RANGE must stop the release, not silently flow into `git rev-list
+# --count ""` (whose stderr never reaches $RAW_COUNT and would let the guard
+# read a "0 commits, 0 gap" pass — AgDR-0104: "a gate that can't evaluate its
+# precondition must block, not allow").
+if [ -z "$LOG_RANGE" ]; then
+  echo "ERROR: could not read RELEASE_CHANGELOG_RANGE from bin/release-changelog.sh's stderr (see /tmp/release-changelog-range.txt). The count-mismatch guard has nothing to check against — do NOT proceed until this is resolved." >&2
+  exit 1
+fi
 ```
 
 The helper emits markdown to stdout in the format:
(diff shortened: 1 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1027

**Commit**: `2145a9165d9bdf95791c95a25831cf106f2b3c69`

### Summary

Docs-only fix (2 markdown files, +47/−9) closing the two real findings from #1017 on the release count-mismatch guard: (1) the warning text now describes *classification drop-out inside `$LOG_RANGE`* rather than the truncated-range case the guard structurally lost when #1012 re-anchored both counts onto the same range, and (2) an unresolvable `$LOG_RANGE` now fails **closed** (`exit 1`) instead of flowing into `git rev-list --count ""` and reading as a "0 commits, no gap" pass. Applied symmetrically to the automated path (`.claude/skills/release/SKILL.md`) and the manual runbook (`docs/release-process.md`). `bin/release-changelog.sh` is untouched.

### Checklist Results

- Architecture & Design: **Pass** — no structural change; the fix applies AgDR-0104's already-decided rail ("a gate that can't evaluate its precondition must block, not allow") at one more site.
- Code Quality: **Pass**
- Testing: **Pass (with a note)** — see Suggestions #2.
- Security: **Pass** — no secrets, no auth surface; changes move a gate from fail-open to fail-closed.
- Performance: **N/A**
- PR Description & Glossary: **Pass** — Glossary present (5 terms), and the finding-by-finding disposition explicitly names what was deferred.
- Summary Bullet Narrative: **Pass** — every bullet carries the *why*, including the third one that justifies a deliberate deviation from the issue's own AC.
- Technical Decisions (AgDR): **N/A** — no material decision. Rejecting `${RAW_COUNT:-0}` in favour of an explicit `exit 1` is a routine implementation choice, reversible inside this PR, and it *applies* an existing AgDR rather than making a new call.
- Adopter Handbooks: **N/A** — no handbooks loaded (no `handbooks/architecture`, `handbooks/general`, or diff-matched language bucket).
- Fallow Static Analysis: **N/A** — not a JS/TS diff.

### Correctness verification

- `RAW_COUNT=$(git rev-list --count "$LOG_RANGE") || { …; exit 1; }` — correct. For a simple assignment from a command substitution, the assignment's exit status *is* the substitution's status in both bash and POSIX sh, so the `||` branch genuinely fires. Confirmed empirically: `git rev-list --count ""` exits non-zero and yields empty stdout, which is exactly the case that previously slipped through.
- Portability: nothing GNU-only introduced. `grep -oE`, `grep -vcE`, `cut -d= -f2-`, `[ -z ]`, `git rev-list --count` all behave identically on BSD/macOS and GNU. No `grep -P`, no `sed -i` without a suffix, no `readlink -f`.
- `[ -n "$ENTRY_COUNT" ] || ENTRY_COUNT=0` is behaviourally equivalent to the `${ENTRY_COUNT:-0}` it replaces — legitimate to keep a default here (an empty changelog is a *valid* state), while `RAW_COUNT` correctly gets no default (a failed `git rev-list` is not).
- **The deviation from #1017's AC #3 is the right call and is well argued.** The issue asked for "the same defensive default treatment as `ENTRY_COUNT`" on `RAW_COUNT`; a `:-0` there would have re-implemented the fail-open bug in a nicer font. The PR says so explicitly instead of silently diverging.
- Docs consistency: no stale "#872 signature — a truncated changelog range" wording survives anywhere in tracked source (the only remaining hits are in disposable agent worktrees).
- CI: all checks green at this SHA (hook suite, markdownlint, ticket ID, CodeQL).

### Issues Found

None blocking.

### Suggestions

1. **nit — the two guards live in different fenced blocks, and `$LOG_RANGE` may not survive between them.** In `SKILL.md`, `LOG_RANGE` is captured (and empty-checked) at the end of step 3's block, but the `git rev-list` call sits in a separate fenced block further down. Each fenced block is typically executed as its own shell invocation, so `$LOG_RANGE` can arrive unset — in which case the new `|| exit 1` halts the release with "git rev-list --count  failed" on a run that was otherwise fine. This fails in the *safe* direction (loud stop, not a silent pass), so it isn't a regression relative to before this PR — but a one-line re-parse of `/tmp/release-changelog-range.txt` at the top of the guard block would make the block self-contained and remove the spurious-halt path. `docs/release-process.md` doesn't have this issue; its snippet is continuous.
2. **nit — #1017's "a test covers the empty-parse path" AC is not met.** Reasonable here: the changed artifacts are markdown (agent instructions + a human runbook), and the tested executable, `bin/release-changelog.sh`, is unchanged — there's no shell file for the existing harness to cover. Correctly scoped with `Refs #1017` rather than `Closes`, so the remaining ACs (test, anchor-kind surfacing, squash-body note) stay tracked. Worth a line in the issue naming which ACs are still open after this merge.
3. **suggestion — the anchor-kind idea from #1017 Finding 1 is the part that restores real coverage.** Emitting `RELEASE_CHANGELOG_ANCHOR=trailer|sync|merge-base|tag` and warning when the anchor was *inferred* rather than *recorded* is the only thing that gets truncation detection back at the one moment it's still reachable (a tag predating the trailer convention). Explicitly deferred here, which is fine — just don't let it fall off #1017.

### Verdict

**APPROVED**

The fix is minimal, correct, and moves a gate in the only defensible direction. The reasoning quality is notably good: it identifies that one of the issue's own acceptance criteria would have reintroduced the bug, and says so in the PR body instead of quietly doing something else.

---
📌 Reviewed commit: `2145a9165d9bdf95791c95a25831cf106f2b3c69`

Issue #1025, opened 2026-07-26
[Bug] tracker_review_submit returns 0 without posting under zsh (BASH_SOURCE), so a review can be silently lost


atlas-apex:
## Given / When / Then

- **Given** a sanctioned reviewer agent finishing a code review, running its commands through a Bash tool whose shell is **zsh** (the default on macOS),
- **When** it calls `tracker_review_submit` to post the review,
- **Then** the function prints `_tracker_load_portfolio_lib:5: BASH_SOURCE[0]: parameter not set` to stderr and **returns 0 — without posting anything.**

Expected: either the review posts, or the call fails with a non-zero status the caller can act on.

## Why this is worse than an ordinary bug

A reviewer that believes it posted will **write its approval marker and report success**. Nothing downstream contradicts it:

- The merge gate reads the **local `*-rex.approved` marker**, not GitHub's review state — that is a deliberate, documented design (`AgDR-0075`), correct for single-account setups where GitHub refuses self-approval.
- So a merge can proceed on a review that **exists as a marker but has no visible artefact** on the PR.

The human-readable review is the only record a person can actually read later. Losing it silently means the audit trail says "reviewed" while the PR shows nothing — and the failure is invisible at exactly the moment it matters.

Observed twice today. The first time, a reviewer only caught it because it independently re-checked the PR afterwards and found no comment; wrapping the call in `bash -c` fixed it immediately. A reviewer that trusted the exit status would have reported a clean review that was never posted.

## Root cause

`_tracker_load_portfolio_lib` (in `_lib-tracker.sh`) dereferences `BASH_SOURCE[0]`. Under zsh that parameter does not exist, and with `set -u`-like strictness the reference errors — but the error occurs in a context where the non-zero status is swallowed, so the function returns 0 having done nothing.

This is a **shell-assumption bug, not a tracker bug**: the hook libraries are bash-only, and nothing enforces that they are invoked under bash. The same `BASH_SOURCE` breakage under zsh has surfaced elsewhere in the same libraries during ordinary operator use, so `tracker_review_submit` is likely one symptom of a class rather than an isolated case.

## Scope

1. **Make the failure loud.** `tracker_review_submit` must return non-zero when it has not posted. A submit path that returns success without a side effect is the single worst property this function could have, given what depends on it.
[...]

Issue #381, opened 2026-05-23
[Bug] code-reviewer agent's approval marker can land in a /tmp inspection clone instead of the ops fork


mohamedELamine:
## Given / When / Then
**Given** the code-reviewer agent (Rex) is invoked on a PR and clones the repository to a scratch location (e.g. `/tmp`) to inspect the PR branch / run tests.
**When** Rex resolves `MARKER_HOME` by walking up for `onboarding.yaml` + `apexyard.projects.yaml` per `.claude/agents/code-reviewer.md` § "Path: ops fork root" (the same walk-up `_lib-ops-root.sh` uses).
**Then** a full ops-fork clone in `/tmp` also has those anchors, so the walk-up resolves to the throwaway clone; the `<pr>-rex.approved` marker is written there and `block-unreviewed-merge.sh` (running from the real ops fork) cannot find it. The merge gate blocks even after Rex approves.
**Expected** the marker is always written to the ops fork from which the review was requested — even when Rex inspects via a scratch clone.

## Repro
1. Invoke the code-reviewer agent on any open PR.
2. Observe Rex clone the repo to `/tmp/...` to inspect.
3. On an APPROVED verdict, observe the marker written under `/tmp/.../.claude/session/reviews/<pr>-rex.approved`, not the real ops fork.
4. `gh pr merge <pr>` from the real ops fork → `block-unreviewed-merge.sh` refuses because the marker is not where it looks.

### 1. Prose fix in `.claude/agents/code-reviewer.md` § "Path: ops fork root"

Instruct Rex to resolve `MARKER_HOME` **once, at review start, from its initial working directory** — before any `cd` / `git clone` / `gh pr checkout`. Cloning elsewhere for diff inspection stays allowed; only the marker path is pinned. Catches the case in practice, but depends on agent compliance.

### 2. Mechanical fix in `_lib-ops-root.sh` + new SessionStart hook

- Add a SessionStart hook (suggested name `pin-ops-root.sh`) that captures the launch-cwd ops root via the pure walk-up and writes it to a cwd-independent, session-keyed pin file: `${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-<CLAUDE_CODE_SESSION_ID>`.
[...]

PR #1028, opened 2026-07-26
fix(#1025): tracker_review_submit returns 0 without posting under zsh

atlas-apex:
## Summary

- **Fixed the silent no-op** — `_tracker_load_portfolio_lib` (in `_lib-tracker.sh`) resolved its own directory via the bash-only `${BASH_SOURCE[0]}`, which is unset under zsh. Unlike its sibling `_tracker_load_config_lib`, it had **no git-rev-parse fallback at all**, so under zsh it printed `_tracker_load_portfolio_lib:5: BASH_SOURCE[0]: parameter not set` and silently `return 0`'d having loaded nothing. A per-project `tracker.kind` registry override then silently stopped resolving, and callers fell through to the global default with no signal anything had gone wrong — the exact "reports success while posting nothing" defect this ticket describes.
- **Made the fix portable, not just quiet** — `${BASH_SOURCE[0]:-}` neutralises the hard "parameter not set" error under zsh's `nounset` (a common `set -u` defensive header — and under `set -euo pipefail` the old bug killed the *entire calling process* at the first reference, before `gh` was ever invoked). The added `git rev-parse --show-toplevel` fallback is the actual portability fix: it resolves the right directory under any shell because it depends on `git`, not a bash-only parameter. Both loader functions in `_lib-tracker.sh` now return non-zero when the target still isn't loadable, instead of unconditionally returning 0.
- **Audited every sibling** — grepped `.claude/hooks/` and `.claude/skills/` for other `BASH_SOURCE[0]` self-location sites (16 files). Applied the same guard to the ones genuinely at risk: `_lib-audit-history.sh`, `_lib-onboarding-depth-mode.sh`, `_lib-onboarding-glossary-seen.sh`, `_lib-fresh-fork.sh`, `_lib-project-board.sh`, `_lib-extract-pr.sh` (its existing guard didn't catch the "resolves to caller's cwd" case), and `_lib-premium-hook.sh` (the worst of the bunch — it sets `set -u` unconditionally, so a bare reference hard-errors even without an external `nounset`, and being sourced rather than subshelled, that `set -u` would leak into the calling shell too). `_lib-read-config.sh` already had a `#950` warn+fallback guard; added the same `:-` safety net without changing its warning behaviour. Hook scripts invoked only through the harness's `bash -c` wrapper, and test scripts run only via explicit `bash <script>`, were confirmed safe as-is (never sourced under the operator's shell) — no changes needed there.
- **Added a zsh-specific regression test** (`test_tracker_zsh_self_location.sh`) that fails against the pre-fix code and passes against the fix — proving the loader now works under zsh with `nounset` active, a per-project `tracker.kind` override resolves correctly instead of silently defaulting, `tracker_review_submit`'s full call chain still reaches the host CLI under zsh, a CLI failure propagates as a real non-zero exit under `set -euo pipefail` (rather than the process dying silently before ever posting), and a regression pin against any bare `${BASH_SOURCE[0]}` reference creeping back into the lib.

## Why this matters

`tracker_review_submit` is how the sanctioned code-reviewer agent (Rex) posts its review to a PR. A reviewer that believes it posted — because the function returned 0 — will write its approval marker and report success, with nothing downstream to contradict it (the merge gate reads the local marker, not GitHub's review state, by design per AgDR-0075). Losing the human-readable review silently means the audit trail says "reviewed" while the PR shows nothing, and the failure is invisible at exactly the moment it matters.

## Testing

```bash
# The new regression test — fails on pre-fix code, passes on the fix
bash .claude/hooks/tests/test_tracker_zsh_self_location.sh

# No regressions in the existing tracker-review test
bash .claude/hooks/tests/test_tracker_review_submit.sh

# Full hook suite
bash bin/run-hook-tests.sh
```

Full suite: **114 passed, 1 failed** (`test_sync_codex_adapter.sh`) — confirmed pre-existing and unrelated by re-running it against the unmodified baseline (same 2 sub-case failures with none of this PR's changes applied). `shellcheck -S warning` is clean on every modified file plus the new test.

## Security note

This PR touches `.claude/hooks/**` (the trust chain) — flagging for Security Auditor review per the auto-fire trigger in `.claude/rules/role-triggers.md`. The change is narrowly scoped to self-location path resolution (no new external input handling, no change to what gets executed) but the path is security-relevant by policy regardless of diff shape.

## Glossary

| Term | Definition |
|------|------------|
| `BASH_SOURCE` | A bash-only array holding the currently-executing script's path; unset under zsh, so any bare reference to it is a latent cross-shell bug |
| `nounset` / `set -u` | A shell strictness mode where referencing an unset variable is a hard error instead of silently expanding to empty — common in defensive script headers |
| Self-location | A shell library's own pattern for finding the directory it lives in, so it can `source` sibling files next to it regardless of the caller's `cwd` |
| Silent no-op | A function that returns success (`exit 0`) without performing its documented side effect — the failure mode this ticket closes |
| Regression pin | A test assertion that fails if a fixed bug's exact code shape (here: a bare, unguarded `${BASH_SOURCE[0]}`) is ever reintroduced |

Refs #1025

### DIFF
--- a/.claude/hooks/_lib-audit-history.sh
+++ b/.claude/hooks/_lib-audit-history.sh
@@ -49,9 +49,20 @@
 _LIB_AUDIT_HISTORY_SOURCED=1
 
 # Locate the lib's own dir so we can source siblings (read-config, portfolio-paths).
-_AUDIT_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# ${BASH_SOURCE[0]} is bash-only and unset under zsh (#1025) — the `:-` default
+# avoids a hard "parameter not set" error, and the git-rev-parse fallback below
+# is the actual portability fix (works under any shell, since it only depends
+# on `git`, not on a bash-only parameter).
+_AUDIT_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
+if [ -z "$_AUDIT_LIB_DIR" ] || [ ! -f "$_AUDIT_LIB_DIR/_lib-read-config.sh" ]; then
+  _audit_root="$(git rev-parse --show-toplevel 2>/dev/null)"
+  if [ -n "$_audit_root" ] && [ -f "$_audit_root/.claude/hooks/_lib-read-config.sh" ]; then
+    _AUDIT_LIB_DIR="$_audit_root/.claude/hooks"
+  fi
+  unset _audit_root
+fi
 
-if [ -f "$_AUDIT_LIB_DIR/_lib-read-config.sh" ]; then
+if [ -n "$_AUDIT_LIB_DIR" ] && [ -f "$_AUDIT_LIB_DIR/_lib-read-config.sh" ]; then
   # shellcheck source=/dev/null
   . "$_AUDIT_LIB_DIR/_lib-read-config.sh"
 fi
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -134,7 +134,19 @@
 # resolution. Guarded: only source if not already defined and the lib is
 # present. tracker_kind defaults to "gh" with no config, preserving gh behaviour.
 if ! command -v tracker_kind >/dev/null 2>&1; then
-  _lib_extract_pr_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd 2>/dev/null)"
+  # ${BASH_SOURCE[0]} is bash-only and unset under zsh (#1025) — the `:-`
+  # default avoids a hard "parameter not set" error, but an empty value
+  # still makes `dirname` resolve to ".", i.e. the CALLER's cwd rather than
+  # this lib's real directory, which usually (harmlessly) misses the `-f`
+  # check below. The git-rev-parse fallback is the actual portability fix.
+  _lib_extract_pr_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd 2>/dev/null)"
+  if [ -z "$_lib_extract_pr_dir" ] || [ ! -f "$_lib_extract_pr_dir/_lib-tracker.sh" ]; then
+    _lib_extract_pr_root="$(git rev-parse --show-toplevel 2>/dev/null)"
+    if [ -n "$_lib_extract_pr_root" ] && [ -f "$_lib_extract_pr_root/.claude/hooks/_lib-tracker.sh" ]; then
+      _lib_extract_pr_dir="$_lib_extract_pr_root/.claude/hooks"
+    fi
+    unset _lib_extract_pr_root
+  fi
   if [ -n "$_lib_extract_pr_dir" ] && [ -f "$_lib_extract_pr_dir/_lib-tracker.sh" ]; then
     # shellcheck source=/dev/null
     . "$_lib_extract_pr_dir/_lib-tracker.sh"
--- a/.claude/hooks/_lib-fresh-fork.sh
+++ b/.claude/hooks/_lib-fresh-fork.sh
@@ -47,9 +47,19 @@
 _LIB_FRESH_FORK_SOURCED=1
 
 # Locate the lib's own dir so we can source siblings (read-config, portfolio-paths).
-_FRESH_FORK_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# ${BASH_SOURCE[0]} is bash-only and unset under zsh (#1025) — the `:-`
+# default avoids a hard "parameter not set" error; the git-rev-parse
+# fallback is the actual portability fix (works under any shell).
+_FRESH_FORK_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
+if [ -z "$_FRESH_FORK_LIB_DIR" ] || [ ! -f "$_FRESH_FORK_LIB_DIR/_lib-read-config.sh" ]; then
+  _fresh_fork_root="$(git rev-parse --show-toplevel 2>/dev/null)"
+  if [ -n "$_fresh_fork_root" ] && [ -f "$_fresh_fork_root/.claude/hooks/_lib-read-config.sh" ]; then
+    _FRESH_FORK_LIB_DIR="$_fresh_fork_root/.claude/hooks"
+  fi
+  unset _fresh_fork_root
+fi
 
-if [ -f "$_FRESH_FORK_LIB_DIR/_lib-read-config.sh" ]; then
+if [ -n "$_FRESH_FORK_LIB_DIR" ] && [ -f "$_FRESH_FORK_LIB_DIR/_lib-read-config.sh" ]; then
   # shellcheck source=/dev/null
   . "$_FRESH_FORK_LIB_DIR/_lib-read-config.sh"
 fi
--- a/.claude/hooks/_lib-onboarding-depth-mode.sh
+++ b/.claude/hooks/_lib-onboarding-depth-mode.sh
@@ -53,9 +53,19 @@
 [ -n "${_LIB_ONBOARDING_DEPTH_MODE_SOURCED:-}" ] && return 0
 _LIB_ONBOARDING_DEPTH_MODE_SOURCED=1
 
-_DEPTH_MODE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# ${BASH_SOURCE[0]} is bash-only and unset under zsh (#1025) — the `:-`
+# default avoids a hard "parameter not set" error; the git-rev-parse
+# fallback is the actual portability fix (works under any shell).
+_DEPTH_MODE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
+if [ -z "$_DEPTH_MODE_LIB_DIR" ] || [ ! -f "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.sh" ]; then
+  _depth_mode_root="$(git rev-parse --show-toplevel 2>/dev/null)"
+  if [ -n "$_depth_mode_root" ] && [ -f "$_depth_mode_root/.claude/hooks/_lib-ops-root.sh" ]; then
+    _DEPTH_MODE_LIB_DIR="$_depth_mode_root/.claude/hooks"
+  fi
+  unset _depth_mode_root
+fi
 
-if [ -f "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.sh" ]; then
+if [ -n "$_DEPTH_MODE_LIB_DIR" ] && [ -f "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.sh" ]; then
   # shellcheck source=/dev/null
   . "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.sh"
 fi
--- a/.claude/hooks/_lib-onboarding-glossary-seen.sh
+++ b/.claude/hooks/_lib-onboarding-glossary-seen.sh
@@ -73,9 +73,19 @@
 [ -n "${_LIB_ONBOARDING_GLOSSARY_SEEN_SOURCED:-}" ] && return 0
 _LIB_ONBOARDING_GLOSSARY_SEEN_SOURCED=1
 
-_GLOSSARY_SEEN_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# ${BASH_SOURCE[0]} is bash-only and unset under zsh (#1025) — the `:-`
+# default avoids a hard "parameter not set" error; the git-rev-parse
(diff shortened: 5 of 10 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1028

**Commit**: `9dcaee832122f0f0417b32f64a50287d90105e92`

### Summary

Fixes the #1025 silent no-op: `_tracker_load_portfolio_lib` resolved its own directory via the bash-only `${BASH_SOURCE[0]}` and — unlike its sibling `_tracker_load_config_lib` — had no `git rev-parse` fallback at all, so under zsh it loaded nothing and returned 0. The PR adds the `:-` safe default plus a git-rev-parse fallback to both tracker loaders and six sibling self-locating libs, makes both loaders return their real success/failure status, and lands a zsh-specific regression test.

### Checklist Results

- Architecture & Design: **Pass** — the fix conforms to the fallback shape already established in `_lib-read-config.sh` and `_tracker_load_config_lib`; it makes the libs consistent rather than inventing a new pattern.
- Code Quality: **Pass** — comments explain *why* (the zsh hazard) rather than *what*; each fallback validates a specific sibling file rather than trusting a bare path.
- Testing: **Pass** — verified locally: `test_tracker_zsh_self_location.sh` is **9/9 PASS** on the PR head, and **4 FAIL** when `_lib-tracker.sh` is reverted to the base version (`portfolio_registry becomes available`, `no 'parameter not set' error`, `per-project tracker.kind override resolves`, and the bare-`BASH_SOURCE` pin). That's a genuine regression test, not a tautology. `test_tracker_review_submit.sh` is 27/27 PASS.
- Security: **Pass** with a note for the Security Auditor (below).
- Performance: **Pass** — the fallback only runs when the BASH_SOURCE-derived path fails its `-f` check, so the bash path stays a single cheap resolution.
- PR Description & Glossary: **Pass** — glossary present, five well-chosen terms; ticket linked (`Refs #1025`).
- Summary Bullet Narrative: **Pass** — every bullet carries what-changed *and* why-it-matters.
- Technical Decisions (AgDR): **N/A** (see note 5).
- Adopter Handbooks: **N/A** — no handbooks loaded for this diff.
- Fallow Static Analysis: **N/A** — not a JS/TS diff.

### Issues Found

**None blocking.**

Two things I specifically checked for, because the loaders' return-status change is the kind of thing that breaks callers silently:

- **Return-status flip is safe.** Both loaders now end in `command -v ...`, so they return non-zero where the old code returned 0 (an `if` with a false condition and no `else` returns 0). I verified all ten call sites in `_lib-tracker.sh` — every one is followed by a `local` declaration, so none is a trailing statement whose enclosing function's exit status would flip. I also grepped every hook and skill that sources `_lib-tracker.sh` for `set -e`: none has it, so no consumer aborts on a failed load. The change is safe as landed, but it is a semantic change to a shared internal — worth remembering if a future caller adds `set -e`.
- **The fix addresses the class, not the symptom.** The `:-` alone would only silence the error; the git-rev-parse fallback is what actually makes resolution work under a non-bash shell. Both were applied together everywhere, which is the right call.

1. **Security Auditor pass is warranted** — this touches `.claude/hooks/**`, the trust chain, so the auto-fire trigger in `.claude/rules/role-triggers.md` applies by policy regardless of diff shape (the PR body already flags this — good). One concrete thing for that pass to weigh: the fallback resolves `$(git rev-parse --show-toplevel)/.claude/hooks/<lib>`, which inside a `workspace/<project>` clone is the **project's** repo, not the ops fork — so a third-party managed repo shipping its own `.claude/hooks/_lib-tracker.sh` could be sourced by a merge-gate-adjacent lib. Exposure is bounded (the fallback only fires when `BASH_SOURCE` resolution fails, i.e. non-bash sourcing — the manual-operator path, not the harness's `bash -c` hook path) and the pattern is **pre-existing prior art**, not introduced here. Still worth deciding whether the trust-chain libs should prefer `_lib-ops-root.sh::resolve_ops_root` (pin-first) over `--show-toplevel`.

2. **`dirname ""` still resolves to the caller's cwd.** With an empty `BASH_SOURCE[0]`, `dirname` yields `.`, `cd .` succeeds, and the only thing preventing a wrong-directory source is the `-f <sibling>` test. If a caller's cwd happens to *be* some other repo's `.claude/hooks`, that test passes and the wrong lib gets sourced. This is the pre-existing #950 hazard (correctly documented in the new comments), not something this PR introduces — but an explicit `[ -n "${BASH_SOURCE[0]:-}" ]` gate before attempting that path would remove the whole class for a couple of characters.

3. **`_lib-project-board.sh` missed the consistency guard.** Sibling libs got `[ -n "$dir" ] && [ -f "$dir/..." ]` on the follow-up test; project-board still has the bare `[ -f "$_lib_board_dir/_lib-ops-root.sh" ]`. Harmless (an empty var makes the test false) — purely a consistency nit.

4. **Test nits.** (a) Case 1's first assertion, *"`_tracker_load_portfolio_lib` returns 0 (loaded)"*, passes **both** pre- and post-fix — pre-fix it returned 0 precisely *because* of the bug. The discriminating assertion is the `portfolio_registry becomes available` one right after it; the label just reads as if it proves more than it does. (b) `grep -qc` in the Case 4 assertions passes both `-q` and `-c`; `-q` wins, so it works, but `-q` alone is what's meant.

5. **AgDR: not required, but worth a thought.** The issue offered an alternative (detect a non-bash shell and re-exec under bash, *"prefer whichever keeps a single code path"*); this PR chose per-site guard+fallback instead. That's an options-considered shape — but the chosen option conforms to a pattern already shipped in two libs, is reversible inside one PR, and the reasoning is captured thoroughly in both the PR body and the code comments. Per `.claude/rules/agdr-decisions.md` § "The threshold", that's not a material decision needing a record. If this guard+fallback shape is meant to become *the* repo-wide standard for shell-lib self-location, a short AgDR would be worth writing later.

6. **Follow-up for the remaining acceptance criterion.** Issue #1025's AC 5 (*"ideally: the reviewer flow verifies the posted review exists before writing its marker"*) is explicitly optional and not addressed here — correctly out of scope for this fix. It's also the only piece that survives the *next* shell assumption, so it's worth its own ticket rather than being dropped.

### Verdict

**APPROVED**

Well-scoped, well-tested, and the class-wide audit was done rather than just patching the one reported site. The regression test genuinely fails against pre-fix code, which is the property that matters most for a bug of this shape.

---
Reviewed commit: `9dcaee832122f0f0417b32f64a50287d90105e92`

atlas-apex (commented):
## Security Review: PR #1028

**Commit**: `9dcaee832122f0f0417b32f64a50287d90105e92`

### Summary

Trust-chain change (`.claude/hooks/**`, #777 auto-fire): adds a `${BASH_SOURCE[0]:-}` guard plus a `git rev-parse --show-toplevel` fallback to the self-location logic of 9 hook libraries, and makes both `_lib-tracker.sh` loaders return their real success/failure instead of an unconditional `0`. I reviewed it for the four risks that matter here: code execution via library path resolution, gate integrity under the changed return status, fail-open on a failed load, and injection.

The core fix is correct and moves the fail-open posture in the right direction. One design concern (MEDIUM) on the *choice of fallback primitive* — non-blocking, but worth a follow-up.

### Checklist Results

| Area | Result |
|---|---|
| Secrets & Credentials | Pass |
| Injection Prevention | Pass |
| XSS Prevention | N/A (shell libraries) |
| Auth & Authorisation | N/A |
| Data Protection | Pass |
| API Security | N/A |
| **Gate & Trust-Chain Integrity (#777)** | **Pass** |

### Findings

| # | Severity | Finding |
|---|---|---|
| M1 | MEDIUM | `git rev-parse --show-toplevel` fallback can source a trust-chain lib out of an untrusted repo |
| L1 | LOW | Assignment exit status under `errexit` in a consumer |
| L2 | LOW | Missing `[ -n ... ]` guard in `_lib-project-board.sh`, inconsistent with siblings |

#### M1 (MEDIUM) — the fallback resolves to the *cwd's* repo, not the ops fork

Every new fallback sources `$(git rev-parse --show-toplevel)/.claude/hooks/<lib>.sh`. When the BASH_SOURCE path fails — which is exactly the zsh case this PR targets — and the cwd is inside a **workspace clone or any other git repo that ships a `.claude/hooks/` layer**, that repo's copy of the library is sourced into the trust-chain shell. Benign case: a stale/divergent copy silently replaces the ops fork's `portfolio_registry` / `tracker_*` helpers. Hostile case (a repo being onboarded via `/handover`, i.e. not yet trusted): arbitrary code execution in a session holding git and `gh` credentials.

This isn't a novel objection — the framework's own code argues against this primitive in two places:

- `_lib-read-config.sh:57-59` — *"locate it via BASH_SOURCE — not via `git rev-parse` (which would defeat the whole point of this fix)."*
- `_lib-ops-root.sh` header — *"when the operator works inside a managed-project workspace clone at `workspace/<project>/`, `git rev-parse --show-toplevel` returns the project clone, NOT the ops fork"*, plus the #381 incident where an unrelated fork-shaped tree hijacked resolution. That is why `resolve_ops_root` is pin-first **and** anchor-validated.

Why MEDIUM and not HIGH:

- **Unreachable on the mechanical gate path.** Every hook is invoked via `bash -c` from `.claude/settings.json` and sources its libs by absolute `$HOOK_DIR` / `$OPS_ROOT`. Under bash, `BASH_SOURCE[0]` resolves and the sibling `-f` test passes, so the fallback never fires during gate enforcement.
- **Requires three conditions to line up**: a non-bash sourcing context, a cwd inside a foreign repo carrying `.claude/hooks/`, and that repo being untrusted.
- **The class pre-exists.** `_tracker_load_config_lib` and `_lib-read-config.sh` already carried this exact fallback before this PR. It widens the surface from 2 sites to 9; it does not introduce the pattern.

Suggested follow-up (cheap, ~3 lines per site) — anchor-validate the resolved root against the same conditions `_lib-ops-root.sh` uses before sourcing from it:

```sh
root=$(git rev-parse --show-toplevel 2>/dev/null)
if [ -n "$root" ] \
   && { [ -f "$root/.apexyard-fork" ] \
        || { [ -f "$root/onboarding.yaml" ] && [ -f "$root/apexyard.projects.yaml" ]; }; } \
   && [ -f "$root/.claude/hooks/_lib-portfolio-paths.sh" ]; then
  hook_dir="$root/.claude/hooks"
fi
```

That preserves the zsh fix for the scenario actually reported in #1025 (sourcing from the ops fork) while refusing to load a trust-chain lib out of a managed-project clone. Preferring pin-first `resolve_ops_root` is better still where it's reachable, though `_lib-ops-root.sh` itself has an unavoidable chicken-and-egg.

#### L1 (LOW) — `X="$(cd ... && pwd)"` carries a non-zero status when `cd` fails

Under a consumer running `set -e`, the assignment itself aborts the script. Not reachable today — I confirmed **no** hook under `.claude/hooks/` sets `-e` (only skill test/smoke scripts do), and the PR's own zsh test correctly brackets calls with `set +e`. Pre-existing shape, noted for completeness.

#### L2 (LOW) — `_lib-project-board.sh` consistency

Unlike its six siblings, the subsequent `if [ -f "$_lib_board_dir/_lib-ops-root.sh" ]` didn't gain the `[ -n "$_lib_board_dir" ]` guard. Harmless (an empty value yields `[ -f "/_lib-ops-root.sh" ]`, which is false), but inconsistent with the pattern applied everywhere else in the same diff.

### What passed, and why it matters

**Gate integrity — no skip, no inversion.** Both loaders now return non-zero where they previously returned `0`. I traced all 10 call sites: every one is a bare statement whose status is discarded, and each consumer independently re-verifies what it needs — `_tracker_project_value` re-checks `command -v portfolio_registry` and returns 1; `tracker_kind` / `tracker_id_pattern` / `_tracker_view_template` fall through to hardcoded literal defaults when `config_get_or` is absent. Combined with no hook using `set -e`, the new non-zero returns cannot abort a gate mid-check or flip a decision.

**Fail-open — strictly improved.** This is the point of the change and it lands. `_tracker_load_portfolio_lib` previously returned `0` having loaded nothing, so a per-project `tracker.kind` override silently degraded to the global default with no signal. Now the loader reports truthfully and the degradation is a deliberate, guarded path rather than an accident. No gate function is left undefined in a way a later `command -v` check would silently skip — every such check already has an explicit failure branch.

**Injection — clean.** No `eval` added, no new external input reaches a path. Every new expansion is double-quoted, so spaced paths survive; `git rev-parse` output is captured with `$( )` and quoted throughout. `tracker_review_submit`'s existing defenses are untouched: numeric-only `{pr}` validation before the custom adapter's `eval`, a verdict allow-list normalising anything unexpected to `comment`, and repo passed to `yq` via `strenv()`.

**Secrets — clean.** No credentials, tokens, or sensitive values introduced or logged.

**Test coverage.** The new `test_tracker_zsh_self_location.sh` exercises the right things — loader success under zsh + `nounset`, the practical per-project-override consequence, the full call chain reaching a mock `gh`, non-zero propagation under `set -euo pipefail`, and a regression pin against a bare `${BASH_SOURCE[0]}` creeping back. It skips gracefully without zsh.

### Verdict

**APPROVED** — no CRITICAL or HIGH findings. M1 is a design concern about the fallback primitive, not a defect in this PR's stated fix, and it's unreachable on the mechanical gate-enforcement path; I'd file it as a follow-up rather than block the zsh fix, which closes a real silent-failure hole in the review audit trail.

---
Reviewed commit: `9dcaee832122f0f0417b32f64a50287d90105e92`

Issue #1018, opened 2026-07-25
[Bug] portfolio_validate false-positives on Windows/Git Bash for a plain single-fork setup (drive-letter vs MSYS path mismatch)


Manito2z:
## Affected
`.claude/hooks/_lib-portfolio-paths.sh` — `portfolio_validate()`, partial split-portfolio v2 detection block (~lines 479–495 in v5.2.0)

## Given / When / Then
**Given** a plain single-fork ApexYard setup on Windows (Git Bash / MSYS2), with **no** `.portfolio` overrides at all in `.claude/project-config.json` (all portfolio paths come from `project-config.defaults.json`)
**When** `portfolio_validate` runs (e.g. on every SessionStart, via the portfolio health-check banner)
**Then** it prints a false-positive `broken: partial split-portfolio v2 config` error and returns exit 1 — expected: exit 0, no output, since the config is a valid, non-split single-fork setup

## Root cause

In the partial-split-portfolio detection block:

```bash
local root_real wd_real
root_real=$(cd "$root" 2>/dev/null && pwd -P) || root_real="$root"
wd_real=$(cd "$workspace_dir" 2>/dev/null && pwd -P) || wd_real="$workspace_dir"

_outside_fork() {
  case "$1" in
    "$root_real"|"$root_real"/*) return 1 ;;
  esac
  return 0
}

if _outside_fork "$registry" || _outside_fork "$projects_dir"; then
  if ! _outside_fork "$wd_real"; then
    echo "broken: partial split-portfolio v2 config — ..."
```

`root_real` and `wd_real` are normalized through `cd … && pwd -P`, but `$registry` and `$projects_dir` are passed to `_outside_fork` **raw and un-normalized**.

On Windows/MSYS the two forms diverge: the ops-fork root (`_portfolio_root`, via `git rev-parse --show-toplevel`) returns a Windows drive-letter path, while `pwd -P` returns the MSYS mount-point form. So the prefix comparison ends up comparing drive-letter-form against MSYS-form, and it always fails:

- `root_real` (normalized) = `/d/Projs/<fork>`
- `registry` (raw) = `D:/Projs/<fork>/apexyard.projects.yaml` → prefix match **FAILS** → judged "outside fork"
- `projects_dir` (raw) = `D:/Projs/<fork>/projects` → prefix match **FAILS** → judged "outside fork"
- `wd_real` (normalized) = `/d/Projs/<fork>/workspace` → prefix match succeeds → judged "inside fork"

[...]

PR #1029, opened 2026-07-26
fix(#1018): normalize Windows drive-letter paths in portfolio_validate

atlas-apex:
## Summary

- **Fixed a Windows/Git-Bash false positive in `portfolio_validate()`** — a plain single-fork adopter running Git Bash / MSYS2 got a spurious `broken: partial split-portfolio v2 config` error on every session start, because `git rev-parse --show-toplevel` (a native Windows binary) returns a drive-letter path (`D:/Projs/fork`) while `cd ... && pwd -P` (an MSYS binary) returns the mount-point form (`/d/Projs/fork`) for the identical location — two byte-different spellings of one path.
- **Root cause is one level deeper than the original bug report's own suggested fix** — the divergence isn't in `portfolio_validate()` itself, it's in the shared `_portfolio_canonicalize()` helper that every public resolver (`portfolio_registry`, `portfolio_projects_dir`, `portfolio_workspace_dir`, etc.) goes through: its "is this absolute?" guard (`case "$p" in /*)`) never recognized a Windows drive-letter path as absolute, so such paths were returned raw/un-normalized instead of going through the existing `cd`/`pwd -P` canonicalization every other absolute path already gets. Fixing the shared helper fixes it for every consumer, not just this one call site.
- **Fix is a pure string transform, not a special case** — `<letter>:/` or `<letter>:\` is rewritten to `/<letter>/` (lowercased, backslashes to forward slashes) *before* the existing absolute-path logic runs. No `cygpath` dependency (absent on macOS/Linux) and no reliance on `cd` actually resolving a drive letter. For every real POSIX path — which never starts with `<letter>:` — this branch is dead code, so existing single-fork/split-portfolio behavior on macOS/Linux is unchanged byte-for-byte.
- **Genuine split-portfolio misconfiguration is still caught** — the fix normalizes spelling, it doesn't relax the comparison. A registry/projects_dir that actually resolves to a different directory (not just a different spelling of the same one) still trips the existing `broken: partial split-portfolio v2 config` / fork-containment checks.

## Testing

- New test file: `.claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh` (11 cases). Since real Git Bash/MSYS2 can't run on macOS, the bug is reproduced by driving `_portfolio_canonicalize()` and the exact `_outside_fork()` prefix-match logic directly with both path spellings as string inputs — matching the bug report's own repro shape, not a simulated filesystem fixture (MSYS's drive-letter↔mount-point translation is a Windows-runtime behavior that cannot be faithfully reproduced on macOS's filesystem semantics).
  - Confirmed **BEFORE the fix** (running this new test file against the unmodified `upstream/dev` copy of the library): 5 of 11 cases fail — the drive-letter normalization cases and the `portfolio_validate` prefix-comparison case, exactly matching the reported bug.
  - Confirmed **AFTER the fix**: all 11 cases pass, including:
    - the two spellings of the same path now canonicalize to the identical string,
    - the `portfolio_validate`-style prefix comparison no longer false-positives,
    - a **genuinely misconfigured** portfolio (registry pointing at a real, different sibling directory) is still correctly flagged as outside the fork — the fix doesn't make the check toothless,
    - ordinary POSIX absolute/relative/nonexistent paths canonicalize byte-for-byte identically to before (no behavior change for existing macOS/Linux adopters),
    - a real single-fork sandbox (git-inited, registry + projects dir + ideas backlog) still passes `portfolio_validate()` cleanly end-to-end.
- Full hook test suite (`.claude/hooks/tests/test_*.sh`, run in per-file-timeout chunks to stay within command limits): **BEFORE** (unmodified `upstream/dev`) 93/98 passed → **AFTER** (this branch, 99 files including the new one) 94/99 passed. The 5 failing files are byte-identical before and after (`test_ops_root.sh`, `test_require_agdr_for_arch_pr.sh`, `test_split_portfolio_v2_migration.sh`, `test_sync_codex_adapter.sh`, `test_workspace_tracker_resolution.sh`) — pre-existing, unrelated to this change, confirmed via a line-for-line diff of the two run logs (excluding the new test file) showing **zero** differences. This change causes no regressions and incidentally fixes nothing else.

## Glossary

| Term | Definition |
|------|------------|
| MSYS / Git Bash | The POSIX-emulation shell environment bundled with Git for Windows; paths inside it use `/c/...`-style mount points rather than Windows drive letters |
| `pwd -P` | Prints the current directory with all symlinks resolved — the canonical physical path |
| Drive-letter path | A Windows-native absolute path spelling like `D:/Projs/fork` or `C:\Users\x`, as emitted by native Windows binaries (including `git.exe`) even inside Git Bash |
| Split-portfolio v2 | An ApexYard deployment mode where the public framework fork and the private portfolio (registry, project docs, workspace) live in two separate sibling repos instead of one |
| `portfolio_validate` | The health-check function in `_lib-portfolio-paths.sh` that runs on SessionStart and flags portfolio-path misconfigurations |
| `_portfolio_canonicalize` | The internal helper that normalizes an absolute path to its logical, `/../`-free, symlink-resolved form, used by every public path resolver in the library so downstream prefix-match comparisons are reliable |

Refs #1018

### DIFF
--- a/.claude/hooks/_lib-portfolio-paths.sh
+++ b/.claude/hooks/_lib-portfolio-paths.sh
@@ -121,9 +121,42 @@ _portfolio_root() {
 # echoes the original input unchanged rather than an empty/broken path —
 # unlike require-active-ticket.sh's fail-CLOSED security check, a
 # portfolio path resolver has no "block on doubt" contract to uphold.
+#
+# Windows/MSYS drive-letter normalization (me2resh/apexyard#1018): on
+# Windows Git Bash, `git rev-parse --show-toplevel` (git.exe is a native
+# Windows binary) returns a drive-letter path like "D:/Projs/fork", but
+# `cd ... && pwd -P` (run directly by bash, an MSYS binary) returns the
+# MSYS mount-point form "/d/Projs/fork" for the SAME location. The two
+# spellings are byte-different for one identical path. Before this fix,
+# the `case "$p" in /*)` guard below didn't recognize a drive-letter path
+# as absolute at all, so it hit the `*)` fallback and was returned RAW,
+# un-normalized — while a caller that instead ran `cd "$root" && pwd -P`
+# directly (portfolio_validate's root_real/wd_real) got the MSYS form.
+# Comparing the two forms via a string prefix match then always failed,
+# even for a perfectly valid, non-split single-fork setup.
+#
+# Fix: normalize a leading `<letter>:/` or `<letter>:\` to `/<letter>/`
+# BEFORE the absolute-path check, as a pure string transform — no
+# dependency on cygpath (absent on macOS/Linux) or on cd/pwd actually
+# resolving a real drive letter (which only MSYS's runtime does). This
+# is a no-op for every POSIX path: a real absolute path always starts
+# with "/", never with "<letter>:", so this branch is simply never
+# entered outside a Windows-drive-letter input.
 # ------------------------------------------------------------------------------
 _portfolio_canonicalize() {
   local p="$1" dir tail=""
+
+  case "$p" in
+    [A-Za-z]:/*|[A-Za-z]:\\*)
+      local drive rest
+      drive="${p%%:*}"
+      rest="${p#*:}"
+      rest="$(printf '%s' "$rest" | tr '\\' '/')"
+      drive="$(printf '%s' "$drive" | tr '[:upper:]' '[:lower:]')"
+      p="/${drive}${rest}"
+      ;;
+  esac
+
   case "$p" in
     /*) : ;;
     *) echo "$p"; return 0 ;;  # not absolute — nothing to canonicalize
--- a/.claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh
+++ b/.claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh
@@ -1,0 +1,283 @@
+#!/bin/bash
+# Regression test for me2resh/apexyard#1018 — portfolio_validate() false-positives
+# on Windows/Git Bash for a plain single-fork setup, because of a
+# drive-letter (D:/Projs/fork) vs MSYS-mount-point (/d/Projs/fork) path
+# spelling mismatch inside `_portfolio_canonicalize`.
+#
+# Why this test drives the comparison directly instead of a filesystem
+# fixture: the bug is a Windows/MSYS-runtime-specific behavior — git.exe
+# (a native Windows binary) emits drive-letter paths, while `cd ... &&
+# pwd -P` (run by bash, an MSYS binary) emits the MSYS mount-point form
+# FOR THE SAME LOCATION. That translation only exists inside the MSYS
+# runtime. On macOS/Linux, "D:/Projs/fork" and "/d/Projs/fork" are just
+# two unrelated path strings (colon has no special meaning) — there is
+# no way to make `cd` treat them as the same directory here, so a
+# filesystem-based fixture would not exercise the real bug at all; it
+# would only test colon-containing directory names, an unrelated
+# concern. Driving `_portfolio_canonicalize` directly with both
+# spellings as string inputs is the correct, honest reproduction on a
+# non-Windows CI runner — matching the bug report's own repro shape.
+#
+# Exit 0 means all cases passed. Exit 1 on first failure.
+
+set -u
+
+LIB_SRC="$(cd "$(dirname "$0")/.." && pwd)/_lib-portfolio-paths.sh"
+CONFIG_LIB_SRC="$(cd "$(dirname "$0")/.." && pwd)/_lib-read-config.sh"
+DEFAULTS_SRC="$(cd "$(dirname "$0")/../.." && pwd)/project-config.defaults.json"
+
+if [ ! -f "$LIB_SRC" ]; then
+  echo "FAIL: helper not found at $LIB_SRC" >&2
+  exit 1
+fi
+if [ ! -f "$CONFIG_LIB_SRC" ]; then
+  echo "FAIL: config lib not found at $CONFIG_LIB_SRC" >&2
+  exit 1
+fi
+if [ ! -f "$DEFAULTS_SRC" ]; then
+  echo "FAIL: defaults file not found at $DEFAULTS_SRC" >&2
+  exit 1
+fi
+
+PASS=0
+FAIL=0
+FAILED_CASES=""
+
+run_case() {
+  local name="$1"
+  local snippet="$2"
+  local out rc
+
+  out=$(
+    # shellcheck source=/dev/null
+    . "$LIB_SRC"
+    eval "$snippet"
+  )
+  rc=$?
+
+  if [ "$rc" -eq 0 ]; then
+    PASS=$((PASS + 1))
+    echo "PASS: $name"
+  else
+    FAIL=$((FAIL + 1))
+    FAILED_CASES="$FAILED_CASES\n  - $name"
+    echo "FAIL: $name"
+    if [ -n "$out" ]; then
+      echo "  output: $out"
+    fi
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# Case 1: a Windows drive-letter absolute path normalizes to the MSYS
+# mount-point form instead of being returned raw/unchanged.
+# ---------------------------------------------------------------------------
+run_case "#1018: drive-letter path normalizes to MSYS mount-point form" '
+r=$(_portfolio_canonicalize "D:/Projs/fork/apexyard.projects.yaml")
+expected="/d/Projs/fork/apexyard.projects.yaml"
+if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 2: lowercase drive letter also normalizes (case-insensitive drive).
+# ---------------------------------------------------------------------------
+run_case "#1018: lowercase drive letter also normalizes" '
+r=$(_portfolio_canonicalize "d:/projs/fork/projects")
+expected="/d/projs/fork/projects"
+if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 3: backslash-separated drive-letter form (native Windows spelling)
+# also normalizes, with backslashes converted to forward slashes.
+# ---------------------------------------------------------------------------
+run_case "#1018: backslash drive-letter form normalizes" '
+r=$(_portfolio_canonicalize "C:\Users\x\apexyard")
+expected="/c/Users/x/apexyard"
+if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 4 (the core acceptance criterion): the drive-letter spelling and the
+# already-MSYS-form spelling of the IDENTICAL location now canonicalize to
+# the same string — this is what makes the string-prefix comparison inside
+# portfolio_validate() succeed instead of false-positiving.
+# ---------------------------------------------------------------------------
+run_case "#1018: drive-letter and MSYS spellings of the same path compare EQUAL" '
+a=$(_portfolio_canonicalize "D:/Projs/fork/apexyard.projects.yaml")
+b=$(_portfolio_canonicalize "/d/Projs/fork/apexyard.projects.yaml")
+if [ "$a" = "$b" ]; then exit 0; else echo "a=$a b=$b (should match)"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 5: reproduces the exact false-positive mechanism from
+# portfolio_validate()'s partial-split-portfolio-v2 detector (introduced in
+# #373). root_real/wd_real are what a direct `cd "$root" && pwd -P` call
+# yields on real MSYS (already normalized); registry/projects_dir are what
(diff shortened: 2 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1029

**Commit**: `02015044c4b245e47b9237db575a3a5f1548cb91`

### Summary

Fixes a Windows/Git-Bash false positive in `portfolio_validate()` by teaching the shared `_portfolio_canonicalize()` helper to recognize a Windows drive-letter path (`D:/Projs/fork`, `C:\Users\x`) as absolute and rewrite it to the MSYS mount-point form (`/d/Projs/fork`) *before* the existing absolute-path / `/../`-collapse / symlink-resolution logic runs.

The most important judgment call in this PR is that it fixes the **shared helper**, not the call site the issue proposed. That's the right call: `_portfolio_resolve()` funnels every public resolver (`portfolio_registry`, `portfolio_projects_dir`, `portfolio_workspace_dir`, `portfolio_onboarding_path`, `portfolio_custom_*_dir`) through `_portfolio_canonicalize` at line 201, so the one-line-deeper fix repairs all of them at once. The issue's suggested fix would have patched only the `#373` check and left every other consumer returning raw drive-letter paths — including the ones feeding prefix comparisons in `require-active-ticket.sh`'s governed-tree resolution.

### Checklist Results
- Architecture & Design:      **Pass**
- Code Quality:               **Pass**
- Testing:                    **Pass**
- Security:                   **Pass** (see trust-chain note below)
- Performance:                **Pass**
- PR Description & Glossary:  **Pass**
- Summary Bullet Narrative:   **Pass**
- Technical Decisions (AgDR): **N/A** (see note)
- Adopter Handbooks:          **N/A** (no handbooks loaded)
- Fallow Static Analysis:     **N/A** (not a JS/TS diff)

### Correctness — the parts that mattered most

**The POSIX no-op claim holds.** This was the main regression risk, since a corrupted path on macOS/Linux would be far worse than the Windows bug being fixed. The `case` pattern `[A-Za-z]:/*|[A-Za-z]:\\*` is anchored at position 0 and requires exactly `<letter>` `:` `/`-or-`\`. A POSIX absolute path always starts with `/`, so it can never match — the branch is genuinely dead code off Windows. Verified empirically, not just by reading: I ran the new test file (11/11 pass) and the existing `test_portfolio_paths.sh` (45/45 pass) against the PR branch in an isolated worktree.

**Ordering is right.** The transform runs before the `/*` guard, so the rewritten path then flows through the *existing* nearest-existing-ancestor walk, `cd`/`pwd -P` physical resolution, and literal-tail re-append. Nothing is bypassed; the drive-letter form just becomes a shape the existing machinery already handles.

**Fail-soft contract preserved.** If the MSYS mount doesn't exist (a custom `/etc/fstab` mapping `D:` somewhere other than `/d`), the ancestor walk terminates at `/`, which exists, and the function returns the transformed string with the tail re-appended — no empty path, no truncation. That matches the "fails soft, echo the input rather than a broken path" contract documented above the function.

**The fix normalizes spelling without relaxing the comparison.** Test case 7 pins this explicitly: a registry pointing at a genuinely different sibling directory is still judged outside the fork. The `#373` partial-split detector and the `#951` fork-containment checks keep their teeth.

**String handling is correct.** `${p%%:*}` / `${p#*:}` split correctly even when the remainder itself contains a colon; `tr '\\' '/'` only touches the drive-letter branch, so a legitimate backslash in a POSIX filename is untouched; lowercasing only the drive letter (not the tail) matches MSYS's `/c`-style mount naming while preserving case-sensitive path segments.

### Issues Found

None blocking.

### Suggestions

**nit: residual normalization asymmetry inside `portfolio_validate` (worth a follow-up, not this PR).** `root_real`, `wd_real`, and `pd_real` (lines 480, 481, 523) still normalize via a direct `cd … && pwd -P` with `|| <raw fallback>`, rather than through `_portfolio_canonicalize`. On real MSYS the `cd` succeeds, so they land in mount-point form and match the now-normalized `registry`/`projects_dir` — which is why the fix works. But in the *fallback* path (cd fails: permissions, a race, a deleted dir), `root_real` retains the drive-letter form while `registry` is now normalized to `/d/…`, and the prefix match fails again — the same false positive, just in a much rarer scenario. Low risk in practice, since `$root` comes from `git rev-parse --show-toplevel` and is an existing directory by construction. Routing those three through `_portfolio_canonicalize` would give the whole function a single normalization path and close the seam.

**nit: theoretical relative-path behavior change.** A *relative* path of the exact shape `a:/b` or `a:\b` (single letter, colon, separator) previously fell through to the `*)` branch and echoed unchanged; it is now rewritten to `/a/b` and treated as absolute. This cannot arise from any real config value — portfolio overrides are `/…`, `../…`, or `./…` — so it's noise-level. Noting for completeness only.

**nit: file mode.** The new test lands as `100644`; 89 of the 99 existing test files are `100755` (10 are already non-executable, so this isn't novel). Nothing in `.github/workflows` invokes the suite via `./<file>` — the one CI reference runs `bash <file>` explicitly — so this is cosmetic. A `chmod +x` would match the majority convention.

**nit (optional):** the two `printf | tr` subshells could be `${p//\\//}` and `${drive,,}` bash parameter expansions, avoiding two forks per call. `tr` is the more portable choice and this branch only executes on Windows, so as-is is defensible — mentioning it only because the surrounding library is otherwise subshell-frugal.

### Trust chain — Security Auditor pass warranted

This diff touches `.claude/hooks/**`, which per `.claude/rules/role-triggers.md` (#777) auto-fires the Security Auditor. Flagging that the gate applies; it's a separate gate from this code review, not a blocker on this verdict.

For the record, my own read found no weakening of any fail-closed check. `_portfolio_canonicalize` feeds the governed-tree prefix matching that `require-active-ticket.sh` relies on, so it's worth being careful here — but the transform only fires on inputs beginning `<letter>:`, which cannot occur as a `FILE_PATH` on macOS/Linux, and on Windows it makes those comparisons *more* consistent rather than less. No path that was previously judged governed becomes ungoverned.

### AgDR — N/A, with an honest note

I'm marking this N/A rather than requesting an AgDR, and I want to be explicit that it's a judgment call rather than a clear-cut one.

The literal rail-1 reading in `.claude/rules/agdr-decisions.md` ("a decision touching `.claude/hooks/**` … is material regardless of how small the diff is") would demand one. Against that: this PR changes no gate's blocking semantics — it fixes a string-normalization bug so an existing check evaluates correctly on one platform. It adds no dependency (the PR explicitly rejects `cygpath` for exactly that reason), no service, no schema change, and no repo-wide pattern. The mechanical gate agrees: `.claude/hooks/**` is not in `require-agdr-for-arch-pr.sh`'s trigger set. And the one genuine design choice here — shared helper vs. call site — is already documented in a 20-line code comment *and* in the PR body's options reasoning, which is the artifact an AgDR would have produced.

The framework has just been through #995/#997/#999 narrowing precisely this over-broad reading, so requesting a record here would re-import the ceremony that exercise removed. Flagging it so a human can overrule me if they read rail 1 more literally than I have.

### Verdict
**APPROVED**

---
Reviewed commit: `02015044c4b245e47b9237db575a3a5f1548cb91`

atlas-apex (commented):
## Security Review: PR #1029

**Commit**: `02015044c4b245e47b9237db575a3a5f1548cb91`

### Summary

Trust-chain review (auto-fired by the #777 trigger — this PR modifies `.claude/hooks/**`). The change adds a leading `<letter>:/` / `<letter>:\` → `/<letter>/` string transform at the top of `_portfolio_canonicalize()` in `_lib-portfolio-paths.sh`. That helper feeds every public portfolio resolver, including `portfolio_workspace_dir()` — which `require-active-ticket.sh` and `require-migration-ticket.sh` prefix-match against a resolved `FILE_PATH` to decide governed-tree membership. So the relevant question is not generic web-OWASP, it's whether the transform can move a path across a gate boundary.

It cannot, in either direction, and the POSIX branch is genuinely unreachable in production. No CRITICAL or HIGH findings.

### Checklist Results

| Area | Result |
|---|---|
| Gate-boundary integrity (governed-tree prefix match) | Pass |
| Path traversal (`..`, backslash tails, pre-collapse ordering) | Pass |
| POSIX regression (branch reachability on macOS/Linux) | Pass |
| Injection (unquoted expansion, glob, `eval`) | Pass |
| Secrets / credentials | Pass (N/A — no secret-bearing surface touched) |

### Analysis

**Gate-boundary integrity — no bypass in either direction.** The transform maps `X:/…` to `/x/…`, which under MSYS is not an arbitrary rewrite but the *mount point for that exact drive* — the two strings denote one location by construction, so the transform cannot make a genuinely out-of-tree path collide with a governed prefix. I confirmed the fix moves the two gates toward *more* consistency on Windows, not less: `require-active-ticket.sh` resolves `FILE_PATH` through its own `_resolve_real_path()` (walk-up + `cd`/`pwd -P`), which under MSYS already yields the mount-point form. Before this change `portfolio_workspace_dir()` returned the raw drive-letter form, so the two sides of the prefix match were in different spellings and a write genuinely inside `workspace/` was judged *outside* it. Post-fix both sides are in mount-point form. The direction of the correction is toward "judged inside a governed tree", i.e. more gating, and the #883 out-of-governance exemption requires all three of its conditions (outside ops fork AND outside every workspace AND not inside any git repository) — so this change cannot open an exemption that was previously closed. Test case 7 in the new file pins the converse: a *genuinely* different sibling directory is still flagged outside the fork, so the fix normalizes spelling without relaxing the comparison.

**Path traversal — safe, and the pre-collapse ordering is fine.** The transform runs before the `/../`-collapse and symlink resolution, which is the correct order: it only rewrites the *prefix*, leaving the tail to the existing walk-up + `cd … && pwd -P` logic, which physically collapses `..` and follows symlinks exactly as it does for a POSIX path. Verified by driving the PR-head helper directly: `D:/Projs/fork/../../etc/passwd` keeps its `..` segments only because `/d/...` doesn't exist on macOS (fail-soft return of the input); on a real MSYS runtime where the ancestor exists, the walk resolves and `pwd -P` collapses the traversal before any prefix match sees it. A crafted `..` therefore cannot survive into a comparison in un-collapsed form on the platform where the branch is reachable. The `tr '\\' '/'` on the tail is confined to the drive-letter branch, so it cannot mangle a POSIX filename that legitimately contains a backslash.

**POSIX regression — the branch is unreachable.** The `case` pattern anchors `[A-Za-z]:` at position 0, so no `/`-leading path can enter it. `_portfolio_canonicalize()` has exactly one caller, `_portfolio_resolve()` (line 201), which passes either an already-`/`-leading override or `$root/<relative>`; `$root` comes from `git rev-parse --show-toplevel`, always `/`-leading on macOS/Linux. I confirmed byte-identical output for absolute, nonexistent-leaf, relative, and space-containing paths. This is the higher-impact direction the review brief flags, and it checks out.

**Injection — clean.** No `eval`. Every expansion is quoted (`case "$p"`, `printf '%s' "$rest"`, `p="/${drive}${rest}"`), the two `tr` invocations receive quoted stdin rather than arguments, and glob metacharacters survive as literals — `C:/*` canonicalizes to the literal `/c/*` with no expansion. `local drive rest` inside the `case` arm adds no new shell-feature dependency (`local` is already used at function top).

### Findings

| # | Severity | Finding |
|---|---|---|
| 1 | MEDIUM | `_portfolio_resolve()` (line ~186) has the *same* "is this absolute?" blindness this PR fixes one level down: `case "$p" in /*)`. A Windows adopter who sets an **absolute drive-letter override** (e.g. `.portfolio.workspace_dir: "E:/portfolio/workspace"`) still falls into the `*)` arm and gets it joined to the root — `D:/Projs/fork/E:/portfolio/workspace` — which the new transform then normalizes against the *leading* `D:`, silently resolving the override inside the fork. Not a regression (pre-existing) and not a gate bypass (it resolves *inside* the fork, the safe direction, and the #951 fork-containment check surfaces a non-default in-fork resolution). But the PR's own framing — "fixing the shared helper fixes it for every consumer" — isn't quite complete while the caller that decides absoluteness still can't see a drive letter. Worth a follow-up applying the same recognition in `_portfolio_resolve`. |
| 2 | LOW | The transform is safe *because* the sole caller guarantees `/`-leading input on POSIX. That invariant is load-bearing but unstated: if a future caller ever passes a raw config value straight to `_portfolio_canonicalize`, a POSIX **relative** path whose first component is a directory literally named `D:` would be silently promoted to the absolute `/d/…`. Cheap mitigation: a one-line comment on the `case` arm recording that the branch relies on callers never passing raw relative input, so the next editor doesn't lose the invariant. |
| 3 | LOW | On Windows the gates still mix spellings on the *ops-root* side: `OPS_ROOT` is taken from `git rev-parse` (drive-letter form) while `portfolio_workspace_dir()` now returns mount-point form. The failure direction is safe — a write inside the fork can never be exempted by a spelling mismatch, because the #883 exemption also requires "not inside any git repository" and the fork is one — but normalizing `OPS_ROOT` through the same helper would remove the remaining asymmetry. Out of scope for this PR. |

### Recommendations

- Follow-up ticket for finding 1 (`_portfolio_resolve` drive-letter absoluteness) — same bug class, one function up, and it's the remaining Windows path-resolution gap.
- Add the invariant comment from finding 2 in this PR if convenient; it's a comment-only change.
- The test file is well-targeted: case 6 pins the bug mechanism, case 7 pins that the check isn't made toothless, and cases 8–9 pin the POSIX no-op. That is the right shape for a trust-chain change that can't be exercised on the CI platform.

### Verdict

**APPROVED** — no CRITICAL or HIGH findings. Three non-blocking findings (1 MEDIUM, 2 LOW), none of which affect this PR's merge safety.

---
📌 Reviewed commit: `02015044c4b245e47b9237db575a3a5f1548cb91`

Issue #1014, opened 2026-07-25
[Bug] test_sync_codex_adapter.sh fails on clean dev (suite 113/114)


atlas-apex:
## Given / When / Then

- **Given** a clean checkout of `dev` at or after v5.2.0,
- **When** the full hook test suite is run,
- **Then** `.claude/hooks/tests/test_sync_codex_adapter.sh` fails, taking the suite to **113/114**.

The failure exercises `bin/sync-codex-adapter.sh` codex-CLI adapter generation. It is unrelated to the change that surfaced it.

## How it surfaced

Found independently by **two** agents during a batch of unrelated work (PRs #1009 and #999). Both correctly declined to fold it into their scope and flagged it instead — which is why it is being filed rather than silently absorbed into an unrelated PR.

The #1009 agent confirmed it is pre-existing rather than caused by its own change; the #999 agent separately reproduced a failure on a clean checkout by stashing its changes, establishing the same conclusion for its diff.

## Why it matters more than one red test

A permanently-failing test in the suite is corrosive beyond the feature it covers:

- It **trains everyone to expect red**, so the next genuine regression is dismissed as "the known failure" and ships.
- It makes the suite unusable as a single pass/fail gate — every consumer has to know which failure is acceptable, which is tribal knowledge that does not survive a new contributor or an agent.
- It costs time repeatedly: several agents in a single session each had to independently determine that this failure was not theirs.

## Scope

Determine which is true, then act accordingly:

1. **The adapter is genuinely broken** → fix `bin/sync-codex-adapter.sh`.
2. **The test is stale** — it asserts behaviour deliberately changed since it was written → update the test to the current contract.
3. **The test is environment-dependent** — depends on a codex CLI, a path, or a tool not present in every environment → make it skip explicitly and loudly (a clear SKIP, never a silent pass) rather than fail.

Whichever it is, the suite must end at **114/114 or 113 pass + 1 explicit skip** on a clean checkout, so a red result once again means something is actually wrong.

## Acceptance Criteria

- [ ] Root cause identified and stated — broken adapter, stale test, or environment dependency
- [ ] The full hook test suite is green on a clean `dev` checkout, with any skip explicit and explained
- [ ] If the adapter was genuinely broken, the fix is covered by the test that caught it
- [ ] If the test was environment-dependent, the skip condition is precise — it must not mask a real failure on a machine where the dependency IS present

## Risks / Dependencies

[...]

Issue #999, opened 2026-07-23
[Chore] Align .claude/hooks AgDR wording with the material threshold (Heavy path)


atlas-apex:
## Driver

Follow-up to #997/#998, which narrowed `.claude/rules/agdr-decisions.md` from "before **any** technical decision" to a materiality threshold (AgDR-0108). Three artefacts under `.claude/hooks/**` still carry the superseded wording or cite a line that no longer exists. Rex flagged these during review of #998 and correctly ruled them **out of scope** for a prose PR: `.claude/hooks/**` is rail-1 trust chain per right-size-ceremony, so it takes the **Heavy** path (Security Auditor review), not a tack-on.

The live one matters most: `require-agdr-for-arch-pr.sh` prints the old bar directly to agents on every block, so the narrowing is contradicted at the exact moment an agent is deciding what to do.

## Scope

- `require-agdr-for-arch-pr.sh:595` — the agent-facing block message says the framework *"calls /decide a HARD STOP before **any** technical decision"*. Update to the material threshold and cite `.claude/rules/agdr-decisions.md § "The threshold"`.
- `require-agdr-for-arch-changes.sh:7` — header cites the old rule section (`§ "Pre-commit hook warns if architecture files changed without an AgDR reference"`), now renamed.
- `hooks/README.md:173` — same dangling citation.
- While in `require-agdr-for-arch-changes.sh`: its header comment still lists `infrastructure/**` as a trigger even though `ARCH_GLOBS` deliberately excludes it (false positives on `docs/infrastructure/notes.md`, `src/types/infrastructure/foo.ts`). `hooks/README.md:170` already logs this as a known drift. Fix the comment to match the code so nobody re-quotes it — #998 was misled by exactly this.

**Message text only — no behaviour change.** No trigger path, glob, exit code, or gate logic may change in this ticket. If any behavioural change turns out to be warranted, split it into its own ticket with its own AgDR.

## Acceptance Criteria

- [ ] No file under `.claude/hooks/**` tells an agent an AgDR is required for "any technical decision".
- [ ] Both dangling citations point at a section that exists.
- [ ] `require-agdr-for-arch-changes.sh`'s header accurately describes its own `ARCH_GLOBS` (no `infrastructure/**` claim), with the false-positive reason retained.
- [ ] Zero behavioural change: trigger sets, exit codes, and gate logic byte-identical in effect; existing hook tests pass unchanged.
[...]

PR #1030, opened 2026-07-26
fix(#1014): isolate test_sync_codex_adapter.sh from the ambient session-pin

atlas-apex:
## Summary

- **Fixed a test-isolation gap in `test_sync_codex_adapter.sh`, not a bug in the Codex adapter** — the two failing assertions ("preserves blocking exit" / "preserves allowing exit") exec a generated hook command copied verbatim from a real `.claude/settings.json` entry. That command resolves the ops root via a session-pin-first, PWD-walk-up-second precedence, checking `CLAUDE_CODE_SESSION_ID` + `$APEXYARD_OPS_PIN_DIR` directly rather than sourcing `_lib-ops-root.sh` — so it does not honour `APEXYARD_OPS_DISABLE_PIN`, the isolation var `bin/run-hook-tests.sh` already exports for the rest of the suite (#528).
- **Why it surfaced on a clean checkout** — running the test from inside a live Claude Code session (exactly how #1009's and #999's agents ran it, and how I reproduced it) leaves `CLAUDE_CODE_SESSION_ID` set with a real ops-root pin file already on disk for that session (`~/.claude/apexyard/ops-root-<session-id>`). The session-pin branch then wins and execs `<real-ops-root>/.claude/hooks/block-test.sh`, which only exists under the test's ephemeral `$TMPROOT`, never under the real ops root — hence exit 126 ("No such file or directory") instead of the expected 2/0.
- **The fix scopes the isolation to just the three hook-command invocations** — unsetting `CLAUDE_CODE_SESSION_ID` in the subshell that execs the generated command forces the PWD-walk-up path the fixture (`$TMPROOT/.apexyard-fork` + `$TMPROOT/.claude/hooks/block-test.sh`) was actually built for, without touching the generator's own `--root`-scoped invocation or any other assertion in the file.

## Root cause determination (issue's own decision tree)

Per #1014's three options, this is **option 3 — the test is environment-dependent** (not option 1, adapter genuinely broken; not option 2, stale assertion). The generated hook command's session-pin-first precedence is *intentional, correct, production behaviour* — it's the exact thing the neighbouring "hooks.json preserves Claude session pin path" assertion in this same file verifies on purpose. Making the adapter *not* prefer the session pin would break that assertion and real Codex-CLI hook resolution. The defect is entirely in the test's own isolation: it fails to neutralize ambient session state before exercising a fixture built to be found only via PWD walk-up.

## Testing

Reproduced the failure first, then verified the fix, both from inside a live Claude Code session (the exact condition that was hiding the bug) and headless (no `CLAUDE_CODE_SESSION_ID`, simulating CI):

| Scope | Before | After |
|---|---|---|
| `test_sync_codex_adapter.sh` alone, live session | 57 pass / 2 fail | 59 pass / 0 fail |
| `test_sync_codex_adapter.sh` alone, headless (`env -u CLAUDE_CODE_SESSION_ID`) | n/a (already passed headless) | 59 pass / 0 fail |
| Full hook suite (`bin/run-hook-tests.sh`, 114 files, `APEXYARD_OPS_DISABLE_PIN=1` applied consistently both runs) | **113 pass / 1 fail** (only `test_sync_codex_adapter.sh`) — matches the issue's "113/114" exactly | **114 pass / 0 fail** |

The full-suite comparison ran in bounded ~10-file chunks (`timeout 100` per file) rather than one long foreground call, to stay under tool limits; both runs used the same `APEXYARD_OPS_DISABLE_PIN=1` isolation `bin/run-hook-tests.sh` exports, so it's an apples-to-apples comparison. No other test changed state in either direction.

Reconciliation done first per the SDLC rule: `gh issue view 1014` (no comments), and `gh pr list --search "1014 in:title,body" --state merged` returned nothing — no prior fix existed.

## Glossary

| Term | Definition |
|------|------------|
| Ops-root session pin | A per-session marker file (`~/.claude/apexyard/ops-root-<session-id>`) that lets a hook resolve the real ApexYard fork root reliably even if the shell's cwd has wandered, instead of walking up parent directories |
| `APEXYARD_OPS_DISABLE_PIN` | Env var honoured by `_lib-ops-root.sh` (and the suite runner `bin/run-hook-tests.sh`) to force the PWD-walk-up path and ignore the session pin, so sandboxed tests can't accidentally resolve onto the real ops fork |
| Codex adapter | The generated Codex-CLI-compatible hook/skill/agent configuration produced by `bin/sync-codex-adapter.sh` from this repo's native `.claude/` sources |
| Test-isolation gap | A test failure caused by ambient environment/session state leaking into a fixture the test intended to run in isolation — distinct from a defect in the code under test |

Refs #1014

### DIFF
--- a/.claude/hooks/tests/test_sync_codex_adapter.sh
+++ b/.claude/hooks/tests/test_sync_codex_adapter.sh
@@ -160,8 +160,25 @@ else
   mark_pass "hooks.json omits unsupported if fields"
 fi
 
+# Test isolation (mirrors bin/run-hook-tests.sh's APEXYARD_OPS_DISABLE_PIN
+# rationale, see #528): the generated command below is copied verbatim from
+# a real .claude/settings.json hook entry, and that literal resolves the
+# ops root via a session-pin-first, PWD-walk-up-second precedence — it
+# checks CLAUDE_CODE_SESSION_ID + $APEXYARD_OPS_PIN_DIR directly rather than
+# sourcing _lib-ops-root.sh, so it does NOT honour APEXYARD_OPS_DISABLE_PIN.
+# Running this test from inside a live Claude Code session (exactly how it
+# is normally invoked, by an agent working in this repo) leaves
+# CLAUDE_CODE_SESSION_ID set with a real ops-root pin file already on disk
+# for that session — so the session-pin branch wins and execs
+# "<real-ops-root>/.claude/hooks/block-test.sh", which only exists under
+# $TMPROOT, never under the real ops root. That is a test-isolation gap in
+# THIS test, not a defect in the generated adapter output (preserving the
+# session-pin precedence is the whole point of the "hooks.json preserves
+# Claude session pin path" assertion above) — see me2resh/apexyard#1014.
+# Unsetting the session-id var for just this invocation forces the
+# fallback PWD-walk-up path the fixture below was actually built for.
 hook_command=$(jq -r '.hooks.PreToolUse[0].hooks[0].command' "$TMPROOT/.codex/hooks.json")
-printf '{"tool_name":"Bash","tool_input":{"command":"policy deny"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_block.out 2>&1
+printf '{"tool_name":"Bash","tool_input":{"command":"policy deny"}}\n' | (cd "$TMPROOT" && unset CLAUDE_CODE_SESSION_ID && bash -c "$hook_command") >/tmp/_codex_adapter_block.out 2>&1
 block_rc=$?
 if [ "$block_rc" -eq 2 ]; then
   mark_pass "generated hook command preserves blocking exit"
@@ -169,7 +186,7 @@ else
   mark_fail "generated hook command preserves blocking exit" "expected exit 2, got $block_rc: $(cat /tmp/_codex_adapter_block.out)"
 fi
 
-printf '{"tool_name":"Bash","tool_input":{"command":"policy allow"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_allow.out 2>&1
+printf '{"tool_name":"Bash","tool_input":{"command":"policy allow"}}\n' | (cd "$TMPROOT" && unset CLAUDE_CODE_SESSION_ID && bash -c "$hook_command") >/tmp/_codex_adapter_allow.out 2>&1
 allow_rc=$?
 if [ "$allow_rc" -eq 0 ]; then
   mark_pass "generated hook command preserves allowing exit"
@@ -177,7 +194,7 @@ else
   mark_fail "generated hook command preserves allowing exit" "expected exit 0, got $allow_rc: $(cat /tmp/_codex_adapter_allow.out)"
 fi
 
-printf '{"tool_name":"Bash","tool_input":{"command":"other deny"}}\n' | (cd "$TMPROOT" && bash -c "$hook_command") >/tmp/_codex_adapter_skip.out 2>&1
+printf '{"tool_name":"Bash","tool_input":{"command":"other deny"}}\n' | (cd "$TMPROOT" && unset CLAUDE_CODE_SESSION_ID && bash -c "$hook_command") >/tmp/_codex_adapter_skip.out 2>&1
 skip_rc=$?
 if [ "$skip_rc" -eq 0 ]; then
   mark_pass "generated hook command skips nonmatching predicates"


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1030

**Commit**: `b1a51eaa088ec9df857c0cabb3d8a7e2a4067844`

### Summary

Scopes an `unset CLAUDE_CODE_SESSION_ID` into the three subshells that exec the generated Codex hook command in `test_sync_codex_adapter.sh`, plus a comment block explaining why. Test-only; no production or generator code touched.

### Verification I ran

I reproduced the diagnosis independently on a clean checkout (pre-fix file), both directions:

| Invocation | Result |
|---|---|
| `bash .claude/hooks/tests/test_sync_codex_adapter.sh` (live session, pin file on disk) | **57 pass / 2 fail** — `preserves blocking exit`, `preserves allowing exit` |
| `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_sync_codex_adapter.sh` | **59 pass / 0 fail** |

That matches the PR's root-cause claim precisely: the generated command's ops-root resolver is guarded by `[ -n "${CLAUDE_CODE_SESSION_ID:-}" ]`, reads `${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}/ops-root-$CLAUDE_CODE_SESSION_ID` directly, and never sources `_lib-ops-root.sh` — so it cannot honour the `APEXYARD_OPS_DISABLE_PIN` isolation `bin/run-hook-tests.sh` already exports. With the var unset, the pin branch is skipped entirely and resolution falls through to the PWD walk-up that finds the fixture's `$TMPROOT/.apexyard-fork`. Not papered over — the ambient input is genuinely removed, not the assertion weakened.

### Isolation, residue, cleanup

- **No leakage.** The `unset` lives inside the existing `( cd ... && ... )` subshell, so it cannot affect any later assertion in the file or any sibling test in the suite runner. Correct scoping.
- **No failure-path residue to clean up.** The change introduces no new file, temp dir, or exported state — nothing that would need a `trap`/EXIT handler. `unset` on an already-unset variable returns 0, so the `&&` chain is also safe headless (CI), where the var is absent.
- **The assertions still have teeth.** They continue to exec the generated command and assert exit 2 / 0 / 0; if the generator broke the `$r/.claude/hooks/<script>` delegation or the predicate wrapper, the walk-up path would still fail loudly. This is a real fix, not the issue's option-3 "explicit skip" — which is the better outcome, and squarely inside #1014's warning about not converting a real failure into a skip.
- Applying the `unset` to the third invocation (`skips nonmatching predicates`) as well is the right call even though it passed ambient — it passed only because the predicate wrapper short-circuits before the exec, so uniformity here removes a passing-for-the-wrong-reason case.

### Issues Found

None blocking.

- **The pin branch is now asserted textually but never executed.** `assert_contains ... '$HOME/.claude/apexyard'` confirms the generated string still carries the pin path, but no assertion exercises it end-to-end. A future regression producing a syntactically-present-but-broken pin branch would slip through. A cheap positive companion would close that: set `CLAUDE_CODE_SESSION_ID=codex-test` and `APEXYARD_OPS_PIN_DIR="$TMPROOT/pins"` with `ops-root-codex-test` containing `$TMPROOT`, then assert exit 2 — proving the pin branch resolves *and* is preferred. That makes the pair (pin-wins / walk-up-fallback) fully covered rather than half.
- **Pre-existing nit, not introduced here:** the three output files are fixed paths (`/tmp/_codex_adapter_block.out` etc.) rather than `mktemp`, so two concurrent runs of this file would clobber each other's diagnostics. The same file already uses `mktemp` for `DROPPED_FIXTURE` a few lines down — worth aligning next time this file is touched.
- The 17-line comment for a 3-token change is long, but the reasoning is genuinely non-obvious and it is exactly what stops the next reader from "tidying away" the `unset`. Keep it.

### Trust chain / Security Auditor

The diff path matches `.claude/hooks/**`, which auto-fires the Security Auditor trigger (`.claude/rules/role-triggers.md`). Substantively the security surface is nil: `.claude/hooks/tests/**` only, no gate logic, no marker logic, no `settings.json`, no change to what any hook does at runtime. Per right-size-ceremony rail 1 ("ambiguity rounds up") a quick Hakim pass is cheap and defensible if the orchestrator wants it — it is not a hard merge gate and I do not consider it blocking.

### Checklist Results

- Architecture & Design: Pass (test-only; no layering surface)
- Code Quality: Pass — minimal, correctly scoped, well justified
- Testing: Pass — the fix *is* the test; behaviour verified in both env states
- Security: Pass — no runtime hook behaviour changed
- Performance: N/A
- PR Description & Glossary: Pass — Glossary present (4 terms), `Refs #1014` per the QA-gate convention
- Summary Bullet Narrative: Pass — every bullet carries what changed *and* why
- Technical Decisions (AgDR): N/A — no material decision; test isolation is reversible inside this PR and adds no dependency, schema, control, or repo-wide pattern
- Adopter Handbooks: N/A — the one blocking handbook (`migration-safety.md`) does not apply; no migration files touched. Advisory handbooks (clean-architecture-layers, commit-message-quality) have no findings here.

### Verdict

**APPROVED**

Meets #1014's acceptance criteria: root cause stated and correct (environment dependency in the test, resolved as a real fix rather than a skip), suite returns to 114/114, and no failure is masked.

---
Reviewed commit: `b1a51eaa088ec9df857c0cabb3d8a7e2a4067844`

Issue #1039, opened 2026-07-27
[Bug] Leak-protection hook fails OPEN on quoted --body-file: private names reach public trackers unscanned


atlas-apex:
`block-private-refs-in-public-repos.sh` shares the `extract_flag_value` defect filed as #1038 — but where that hook merely over-blocks (fails closed), **this one silently fails open**. A private project name reaches a public tracker completely unscanned, triggered by two entirely ordinary command habits.

This is the leak `.claude/rules/leak-protection.md` exists to prevent, and the rule's own rationale says self-discipline is *not* sufficient here — the mechanical hook is the defence.

## Affected

`.claude/hooks/block-private-refs-in-public-repos.sh` (`extract_flag_value`, `--body-file` path)

## Given / When / Then

**Given** a registry with a private project, and a body file naming that project, targeting a public framework repo
**When** the `gh issue create` command passes `--body-file` in either of these ordinary shapes:
- a **quoted** path followed by a pipe or redirect — `--body-file "/tmp/body.md" 2>&1 | tail -3`
- a **quoted** path followed by a single-dash flag — `--body-file "/tmp/body.md" -t "[Bug] x"`

**Then** the hook exits **0** and the private name passes unscanned — expected that it exit 2 and block, as it correctly does for `--body-file /tmp/body.md` (unquoted, last)

## Repro

Validated against `upstream/dev` (2554869) with a synthetic registry — fake project name `zebrafish`, no real private names used:

| Command shape | Exit | Outcome |
|---|---|---|
| `--body-file /tmp/leakbody.md` (unquoted, last) | 2 | correctly BLOCKED, leaked token reported |
| `--body-file "/tmp/leakbody.md" 2>&1 \| tail -3` | **0** | **private name passes unscanned** |
| `--body-file "/tmp/leakbody.md" -t "[Bug] x"` | **0** | **private name passes unscanned** |

## Mechanism

Identical to #1038. The double-quoted branch of `extract_flag_value` anchors on a following `--flag` or end-of-string:

```
re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
```

A trailing pipe/redirect, or a single-dash follower, fails that anchor. Execution falls through to the unquoted branch, which has no anchor and returns the token **with the quotes still attached**. `[ -f "\"/tmp/body.md\"" ]` is then false, the body content is never read, and the scan runs over an empty haystack.

[...]

PR #1040, opened 2026-07-27
fix(#1039): leak-protection hook fails closed on an unreadable body-file

atlas-apex:
## Summary

- **Closes a silent leak-protection bypass.** `block-private-refs-in-public-repos.sh` allowed private project names onto public trackers whenever it couldn't parse the `--body-file` path. The operator saw a successful `gh issue create` with no sign the scan had been skipped, and an indexed public issue isn't reliably undone by deletion.
- **Splits path-valued from content-valued flag extraction.** This is the actual fix — see below. `extract_flag_value` is left byte-identical to its #227 form; a new non-greedy `extract_path_flag` handles `--body-file`.
- **Makes "body-file named but unreadable" a hard block** rather than a silent skip, so a *future* parsing gap fails loudly instead of leaking.

## The bug

Two ordinary command shapes leaked:

```
gh issue create ... --body-file "/p/body.md" 2>&1 | tail -3
gh issue create ... --body-file "/p/body.md" -t "[Bug] x"
```

The greedy content matcher can't terminate a quoted path followed by a shell operator, so extraction fell through to the unquoted branch and returned the token **with quotes attached**. `[ -f "\"/p/body.md\"" ]` is false, the body was never read — and this hook exits 0 when it recovers no body.

## The fix, and the wrong turn on the way to it

**The first attempt on this PR widened `extract_flag_value`'s anchor to accept shell operators. That was wrong and was rejected in review.** It closed these bypasses and reopened the leak #227 fixed, in a wider form: `sub()` is leftmost-first rather than suffix-anchored, and each alternative ended in `.*`, so an embedded `"` whose next non-space character was a boundary char truncated the body there — tail unscanned, hook exits 0.

Measured on the same 20-case harness:

| Version | Result |
|---|---|
| `dev` (no fix) | 15 pass, **5 fail** — the #1039 bypasses |
| Rejected first attempt | 11 pass, **9 fail** — *worse than no fix at all* |
| This HEAD | **20 pass, 0 fail** |

The realistic trigger is a markdown table row ending in a quoted term followed by ` |`. A Glossary table is **mandatory** in this framework's own issue bodies, so the likeliest body shape was the triggering one.

The real defect was one extractor serving two incompatible jobs:

- **CONTENT** (`--title` / `--body`) may contain quotes, pipes, semicolons, tables. Extraction must be **greedy**, terminating only on a real flag boundary. `extract_flag_value` is therefore **unchanged from `dev`** — byte-identical.
- **PATH** (`--body-file` / `-F`) is a filesystem path and never contains a quote. Extraction must be **non-greedy**, stopping at the first closing quote.

So this adds `extract_path_flag` rather than widening the shared one. The non-greedy `[^"]*` is safe here for exactly the reason it was unsafe there.

## Testing

29 cases in `test_block_private_refs.sh`, of which 16 are new on this PR:

- the four `--body-file` bypass shapes, plus the unreadable-path block
- **9 back-half cases** — a private ref placed *after* a quote followed by each boundary character (`| ; & < > ( )`), a single-dash flag, and a markdown Glossary table
- regression guards: a clean body in the previously-failing shape still passes, and `gh issue comment` with no body remains a no-op

The back-half cases are the ones that would have caught the rejected attempt. Existing cases 11–13 structurally could not: they use quote-followed-by-*word*, and case 13 passes *more* easily under truncation since a clean body stays clean.

- Full hook suite: **117/117**, matching the clean baseline on `dev`

## Notes for the reviewer

The `-F` fallback below the path extractor is unchanged. `HAYSTACK` concatenates title + body + body-file + api-body, so a clean decoy `--body-file` cannot mask a dirty `--body`.

Known and **not** introduced here: a literal ` --x ` inside body text truncates at the upstream anchor identically on this HEAD and on `dev`. Filed separately rather than widened into this PR — widening that anchor is the mistake this PR exists to undo.

Also filed separately: the fail-closed block runs before the skip-marker check, so its message suggests `<!-- private-refs: allow -->`, which does not clear it. Safe direction, misleading advice.

Refs #1039

## Glossary

| Term | Definition |
|------|------------|
| fail open | On error or ambiguity, allow the action — the dangerous default for a protective gate |
| fail closed | On error or ambiguity, deny — what a protective gate should do |
| greedy / non-greedy | Whether a match consumes as much as possible, or stops at the first terminator |
| back-half detection | Catching a private reference that appears *after* an embedded quote in a body |

### DIFF
--- a/.claude/hooks/block-private-refs-in-public-repos.sh
+++ b/.claude/hooks/block-private-refs-in-public-repos.sh
@@ -154,43 +154,81 @@ extract_flag_value() {
   #   --title value
   #
   # Quoted-value regex is GREEDY and anchored on the next flag boundary
-  # (whitespace + `--<letter>`) or end-of-string. The earlier sed form
-  # `[^"]*` truncated at the first embedded double quote
-  # (me2resh/apexyard#227); for the leak-protection hook that's a real
-  # security tail — a body with an embedded `"` could let private refs in
-  # the back half slip past the gate. awk + greedy + boundary anchor
-  # gives us multi-line consumption and correct termination in one shot.
+  # or end-of-string. The earlier sed form `[^"]*` truncated at the first
+  # embedded double quote (me2resh/apexyard#227); for the leak-protection
+  # hook that's a real security tail — a body with an embedded `"` could
+  # let private refs in the back half slip past the gate. awk + greedy +
+  # boundary anchor gives us multi-line consumption and correct
+  # termination in one shot.
+  #
+  # #1039 — WIDEN THE ANCHOR, AND STRIP QUOTES IN THE FALLBACK.
+  # The anchor used to accept only `--<letter>` or end-of-string. Two
+  # ordinary command shapes therefore failed to match the quoted branches:
+  #
+  #   gh issue create ... --body-file "/p/b.md" 2>&1 | tail -3   (shell operator)
+  #   gh issue create ... --body-file "/p/b.md" -t "[Bug] x"     (single-dash flag)
+  #
+  # On that miss, execution fell through to the UNQUOTED branch, which has
+  # no anchor and returned the token with its quotes still attached. The
+  # caller then did `[ -f "\"/p/b.md\"" ]`, which is false, so the body was
+  # never read — and THIS hook exits 0 when it recovers no body, so a
+  # private project name reached a public tracker completely unscanned.
+  # That is a silent leak, not a false positive: the sibling hooks sharing
+  # this extractor (require-agdr-for-arch-pr.sh, #1038) merely over-block.
+  #
+  # Two changes, both of which only ever cause MORE text to be scanned:
+  #   1. The anchor now also accepts a single-dash flag (`-t`) and a shell
+  #      operator boundary (`| ; & < > (`), matching what
+  #      validate-issue-structure.sh already adopted for #695.
+  #   2. The unquoted fallback strips ONE matched surrounding quote pair,
+  #      so a value reaching it by any other route is still usable.
+  #
+  # Only a MATCHED pair is stripped, so a file literally named `"x.md"`
+  # (quotes in the filename) is not silently rewritten to `x.md` — that
+  # asymmetry would let a crafted path diverge from what gh actually reads.
   local flag_re="$1"
   local cmd="$2"
   printf '%s' "$cmd" | awk -v FLAG_RE="$flag_re" -v SQ="'" '
     { buf = (NR == 1 ? $0 : buf "\n" $0) }
     END {
       s = buf
-      # Double-quoted value: greedy `(.*)` anchored on next flag or EOS.
-      re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
+      # Next-token boundary: a flag (one or two dashes), a shell operator,
+      # or end-of-string. `--?[a-zA-Z]` is used rather than an interval
+      # expression so this stays portable across BSD/GNU awk.
+      ANCH = "([[:space:]]+--?[a-zA-Z]|[[:space:]]*[|;&<>()]|[[:space:]]*$)"
+      # Double-quoted value: greedy `(.*)` anchored on the boundary above.
+      re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"" ANCH
       if (match(s, re)) {
         chunk = substr(s, RSTART, RLENGTH)
         sub("^(" FLAG_RE ")[[:space:]]+\"", "", chunk)
-        sub("\"([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
+        sub("\"([[:space:]]+--?[a-zA-Z].*|[[:space:]]*[|;&<>()].*)?$", "", chunk)
         sub("\"[[:space:]]*$", "", chunk)
         print chunk
         exit
       }
       # Single-quoted value: same greedy + anchor treatment.
-      re = "(" FLAG_RE ")[[:space:]]+" SQ "(.*)" SQ "([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
+      re = "(" FLAG_RE ")[[:space:]]+" SQ "(.*)" SQ ANCH
       if (match(s, re)) {
         chunk = substr(s, RSTART, RLENGTH)
         sub("^(" FLAG_RE ")[[:space:]]+" SQ, "", chunk)
-        sub(SQ "([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
+        sub(SQ "([[:space:]]+--?[a-zA-Z].*|[[:space:]]*[|;&<>()].*)?$", "", chunk)
         sub(SQ "[[:space:]]*$", "", chunk)
         print chunk
         exit
       }
-      # Unquoted value: single token, embedded quotes irrelevant.
+      # Unquoted value: single token. Strip ONE matched surrounding quote
+      # pair (#1039) so a quoted value that reached this branch is usable.
       re = "(" FLAG_RE ")[[:space:]]+[^[:space:]]+"
       if (match(s, re)) {
         chunk = substr(s, RSTART, RLENGTH)
         sub("^(" FLAG_RE ")[[:space:]]+", "", chunk)
+        if (length(chunk) >= 2) {
+          first = substr(chunk, 1, 1)
+          last  = substr(chunk, length(chunk), 1)
+          if ((first == "\"" && last == "\"") || (first == SQ && last == SQ)) {
+            chunk = substr(chunk, 2, length(chunk) - 2)
+          }
+        }
         print chunk
         exit
       }
@@ -219,9 +257,29 @@ if [ -z "$BODY_FILE" ]; then
   fi
 fi
 
+# #1039 — DISTINGUISH "no body-file" FROM "body-file I could not read".
+#
+# These used to be the same code path: an unreadable path left
+# BODY_FILE_CONTENT empty, and the empty-haystack short-circuit below then
+# exited 0. That made every extractor gap a SILENT LEAK rather than a
+# visible failure — the operator saw a successful `gh issue create` with no
+# indication the scan had been skipped.
+#
+# Defence in depth: the extractor fix above closes the known shapes, but
+# this makes the *class* fail closed. Any future parsing gap that yields an
+# unreadable path now blocks loudly instead of waving the write through.
 BODY_FILE_CONTENT=""
-if [ -n "$BODY_FILE" ] && [ -f "$BODY_FILE" ]; then
-  BODY_FILE_CONTENT=$(cat "$BODY_FILE" 2>/dev/null)
+BODY_FILE_UNREADABLE=0
+if [ -n "$BODY_FILE" ]; then
+  if [ -f "$BODY_FILE" ]; then
+    BODY_FILE_CONTENT=$(cat "$BODY_FILE" 2>/dev/null)
+    # Readable-but-unslurpable (permissions, race): treat as unreadable.
+    if [ -z "$BODY_FILE_CONTENT" ] && [ -s "$BODY_FILE" ]; then
+      BODY_FILE_UNREADABLE=1
+    fi
+  else
+    BODY_FILE_UNREADABLE=1
+  fi
 fi
 
 # `gh api` body field: -F body=@file or -f body='text' or -F body='text'.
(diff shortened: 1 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1040

**Commit**: `2a92ed0ea98173ff01dc0df0dac818268a2aa663`

### Summary

Fixes a silent leak-protection bypass in `block-private-refs-in-public-repos.sh` by (a) widening `extract_flag_value`'s next-token anchor to accept single-dash flags and shell operators, and (b) making "body-file named but unreadable" a hard block instead of a silent skip.

The **fail-closed half is correct and well-built**, and the `--body-file` shapes it targets are genuinely fixed — I reproduced all four bypasses against the pre-fix hook and confirmed they now block. But the **anchor half re-opens the exact security hole #227 closed**, in a wider form, on the `--body` / `--title` path. That is a blocking regression.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: **Fail** — the new cases are discriminating, but nothing guards the property the anchor change breaks
- Security: **Fail** — see Issue 1
- Performance: Pass
- Summary Bullet Narrative: Pass (narrative, with rationale)
- Technical Decisions (AgDR): N/A (see suggestion)
- Adopter Handbooks: N/A (no findings — shell hook, no migration/layering surface)
- Fallow Static Analysis (JS/TS): N/A (shell-only diff)

### Issues Found

#### 1. BLOCKING — the widened anchor re-opens #227's truncation hole for 8 new boundary characters

The PR body states the anchor changes "only ever cause MORE text to be scanned." **That claim is false**, and it is the load-bearing claim of the change.

The initial `match()` is fine — POSIX leftmost-longest means it consumes the whole body. The damage is in the **closing-quote stripper**, which is `sub()`, and `sub()` is *leftmost-first*, not suffix-anchored:

```awk
sub("\"([[:space:]]+--?[a-zA-Z].*|[[:space:]]*[|;&<>()].*)?$", "", chunk)
```

Because each alternative ends in `.*`, this pattern can start matching at **any** embedded `"` in the body whose following character (modulo whitespace) is one of `| ; & < > ( )` or a `-`/`--` flag — and it then deletes everything from there to the end. Traced through the awk directly:

```
INITIAL MATCH : [--body "Regex was "(.*)" (greedy). found during marlow-core rebuild."]
AFTER prefix  : [Regex was "(.*)" (greedy). found during marlow-core rebuild."]
AFTER sub#2   : [Regex was ]          <-- body truncated; back half never scanned
```

The hook then scans the truncated string, finds nothing, and **exits 0**. Private ref onto a public tracker, silently — the precise failure mode this PR exists to eliminate.

Reproduced against a fixture registry containing `marlow-core`, comparing the PR hook to its parent commit:

| Body shape (leak is in the back half) | pre-fix | **this PR** |
|---|---|---|
| `Regex was "(.*)" (greedy). … marlow-core …` | blocks (2) | **leaks (0)** |
| `Ran "make test"; then … marlow-core …` | blocks (2) | **leaks (0)** |
| `He said "no" > see. … marlow-core …` | blocks (2) | **leaks (0)** |
| `Use "foo" & bar. … marlow-core …` | blocks (2) | **leaks (0)** |
| `Pass "value" -t style. … marlow-core …` | blocks (2) | **leaks (0)** |
| markdown table row ending `… called "fail open" \|` + leak on next row | blocks (2) | **leaks (0)** |
| control: `He said "no" indeed. … marlow-core …` | blocks (2) | blocks (2) |

The last row is the control — truncation requires the quote to be *immediately* followed by a boundary char, which is why ordinary prose still works and why the existing tests miss this.

The markdown-table row is the one I'd weigh heaviest: a quoted term at the end of a table cell followed by ` |` is an entirely routine shape in this framework's own issue bodies, where a Glossary table is mandatory.

The same structural defect exists pre-fix, but only for `" --letter` (whitespace required, double-dash required) — a shape you'd rarely type by accident. This PR extends it to 8 characters **and** allows zero-width whitespace, so `"(`, `"|`, `";`, `"&`, `">` all trigger it. Net: the change closes four `--body-file` shapes and opens a materially larger hole on `--body`/`--title`.

**On fixing it** — the minimal patch (drop the trailing `.*` from both closing-quote `sub()`s, since `RLENGTH` already ends at the anchor, making the strip suffix-only) does close every row above, and I verified it. But it is *not* a drop-in: it breaks case 15 (`--body-file "path" -t "[Bug] x"`), because leftmost-longest makes the initial match swallow the second quoted argument too. The result there is still fail-closed (a block with a misleading "not readable" reason), not a leak — but it shows the two goals genuinely conflict under one greedy regex: greedy consumption is *required* for inline bodies with embedded quotes, and *harmful* for multi-arg command lines.

That points at a design split rather than a regex tweak: `--body-file` / `-F` take a **path** (single token, no whitespace in any realistic invocation), so they can be extracted token-wise and never need greedy multi-line consumption; `--body` / `--title` take **content** and need greedy + a strictly suffix-anchored strip. Splitting path-valued from content-valued flags would let each side be correct. I'd rather leave the design call to you than prescribe one.

#### 2. BLOCKING (test gap) — nothing guards the property that broke

Cases 11 and 12 are the #227 anti-truncation guards, and they're the right idea — but both use a quote followed by an ordinary **word** (`The "needs attention" flag`), so neither can detect the new boundary characters. Case 13 (`clean body with embedded quotes → pass`) is one-directional: truncation makes a clean body pass *more* easily, so it passes vacuously with respect to this bug.

Please add back-half-detection cases (expect exit 2) covering a quote immediately followed by each new boundary class — at minimum `(`, `|`, `;`, `&`, `>`, and single-dash `-t` — plus the markdown-table shape. Those are what would have caught this, and what will stop the next widening of `ANCH` from re-introducing it.

- **`--body-file -` now hard-blocks.** `gh` treats `-` as "read body from stdin", and `--body-file - <<EOF … EOF` is a common idiom. Pre-fix it passed (unscanned — also wrong); now it exits 2 with "path does not exist". Fail-closed is defensible, but the heredoc content is right there in the command text and could be scanned, or `-` could at least get a purpose-built message instead of a path typo hint.
- **AgDR:** I'm *not* blocking on this. The extractor change is a bug fix, and #227's fix shipped without one — demanding ceremony now would be inconsistent. But the fail-open → fail-closed switch is a deliberate policy change to a trust-chain gate that introduces a new blocking class, which is material under `.claude/rules/agdr-decisions.md` rail 1. A short AgDR (or an explicit "decision recorded here" note in the PR body) would be cheap and would give the next person the reasoning.

### What I verified as sound (no action needed)

- **Matched-pair quote strip is correct.** A one-sided leading or trailing quote is *not* stripped, so the path stays unreadable and the hook fails closed (exit 2) rather than silently reading a different file. A filename literally containing quotes is likewise not rewritten. The reasoning in the PR notes holds up.
- **Editor no-op preserved.** `gh issue comment 42 --repo …` with no body still exits 0 — the fail-closed check correctly sits behind the `BODY_FILE` non-empty guard.
- **Relative `--body-file` from a different cwd now blocks** (e.g. `cd /var/tmp && gh … --body-file body.md`). This is a real behaviour change, but it's the right side of the trade for a public-tracker gate, and the error text already names it as a likely cause.
- **Portability confirmed.** No interval expressions; `--?[a-zA-Z]` and the bracket class are plain ERE, valid on both BSD and GNU awk. All 20 cases pass here on BSD awk 20200816 (macOS). No GNU awk available locally to cross-check, but nothing in the change is gawk-specific.
- **Test claims are honest.** I independently confirmed 5 of the 7 new cases fail against the parent commit and that all 20 pass against this HEAD — the new cases are discriminating, not vacuous, and the two regression cases correctly pass on both.
- No secrets; error handling and messaging are clear and actionable.

### Verdict
**CHANGES REQUESTED**

The fail-closed work is good and I'd like to see it land. The blocker is narrow: the anchor widening trades a fixed `--body-file` bypass for a larger `--body` bypass of the same class. Fix the closing-quote strip (or split path-valued from content-valued flags), add the boundary-character regression tests, and I'll re-review promptly.

---
📌 Reviewed commit: `2a92ed0ea98173ff01dc0df0dac818268a2aa663`

atlas-apex:
## Round 2 — CHANGES REQUESTED addressed

Rex's rejection was correct, and the measurement is worth recording because it shows the first attempt was **worse than no fix at all**. Same 20-case harness, three versions:

| Version | Result |
|---|---|
| `upstream/dev` (no fix) | 15 pass, **5 fail** — the #1039 bypasses |
| First attempt (`2a92ed0`) | 11 pass, **9 fail** — closed 5, broke 9 |
| This HEAD (`7e3b7e3`) | **20 pass, 0 fail** |

### What was wrong

Exactly as diagnosed: the closing-quote stripper is `sub()` — leftmost-first, not suffix-anchored — and each alternative ended in `.*`. An embedded `"` whose next non-space character was a boundary char truncated the body there, and the tail went unscanned while the hook exits 0 on an empty scan.

The PR body's claim that both changes "only ever cause MORE text to be scanned" was false, and I've removed it.

The markdown-table observation is the part that made this undeniable rather than theoretical: a Glossary table is **mandatory** in this framework's own issue bodies, so the most likely body shape was the triggering one.

### The fix

Not a regex tweak — the parsing split you pointed at. One extractor was serving two incompatible jobs:

- **CONTENT** (`--title` / `--body`) may contain anything, so extraction must be **greedy**, terminating only on a real flag boundary. `extract_flag_value` is restored to its #227 form, **unchanged**.
- **PATH** (`--body-file` / `-F`) is a filesystem path and never contains a quote, so extraction must be **non-greedy** — stop at the first closing quote. That is what a quoted path followed by a shell operator needs, and it is precisely wrong for content.

So `extract_path_flag` is new; the shared extractor is untouched. The non-greedy `[^"]*` is safe here for the same reason it was unsafe there.

The fail-closed half you confirmed as sound is retained unchanged.

### Tests

9 new back-half cases — one per boundary character (`| ; & < > ( )`), plus a single-dash flag and a markdown Glossary table — each placing a private ref **after** the truncation point.

You were also right that cases 11–13 structurally could not catch this: they use quote-followed-by-*word*, and case 13 passes more easily under truncation since a clean body stays clean. The new cases fail **9** against the rejected attempt and pass **29/29** here. Full suite 117/117, shellcheck clean.

Ready for re-review at `7e3b7e3`.

atlas-apex (commented):
## Code Review: PR #1040 (re-review)

**Commit**: `7e3b7e3eba678f770729b653877156f826f3ff8a`

Round 1 was CHANGES REQUESTED: widening `extract_flag_value`'s anchor to accept shell operators reopened the #227 back-half leak, because `sub()` is leftmost-first and every alternative ended in `.*`. I re-verified all three claims independently rather than reading the new tests.

### Checklist Results
- Code Quality: **Pass**
- Testing: **Pass**
- Security: **Pass**
- Performance: **Pass**
- PR Description & Glossary: **Fail** — glossary present, but the body describes the *rejected* approach (see blocking-for-merge item)
- Technical Decisions (AgDR): **N/A**

### Verification performed

**1. Is `extract_flag_value` byte-identical to `dev`?** Yes. Sliced the function from both blobs — `md5 f75e83befff397ad1831835c6fb2d488` on each, `diff` empty. The #227 greedy-plus-flag-boundary form is untouched.

**2. Does the back-half leak class actually block?** Re-ran my own round-1 constructions (not the author's tests) against the head hook: markdown glossary table with a quoted term followed by ` |` then a leak in the tail; quote-then-`|`; quote-then-`;`; multi-line table-then-leak; a leak in the **title** back-half; and a `workspace/` path after `quote > `. All six block (exit 2), matching `dev` case-for-case.

One shape leaks on both head **and** `dev`: a body containing `"quoted" --x more…` — a literal `--<letter>` inside body text is a genuine flag boundary under the upstream anchor, so it truncates there. Pre-existing upstream behaviour, identical on both sides, **not introduced or worsened here**. Out of scope for this PR; worth a separate ticket if you want it closed.

**3. Is `extract_path_flag` sound?** Exercised six shapes directly: quoted path containing a space; unquoted path followed by a pipe (leak → block, clean → pass, so no false positive); quoted path followed by a single-dash flag; single-quoted path followed by a redirect; nonexistent path → fail closed. All correct. The `-F` fallback below it still works for quoted, unquoted, and `gh api -F body=@file`.

**4. Does non-greedy `[^"]*` introduce a new bypass?** No. Confirmed the content path is unaffected in three decoy shapes:
- `--body-file "<clean>" --body "…curios-dog…"` → blocked. `HAYSTACK` concatenates title + body + body-file + api-body, so a clean decoy file cannot mask a dirty `--body`.
- `--body-file` appearing as *text inside* a `--body` value → blocked; the path extractor grabbing the decoy doesn't drop the `--body` scan.
- Decoy `--body-file` mention shadowing a real later `--body-file` → blocked on head, and identically on `dev`. No delta.

The reasoning in the new comment block is correct and is the load-bearing part: non-greedy is safe for paths *precisely because* paths cannot contain quotes, which is the same property that makes it wrong for content. Good that this is written where the next person will actually read it.

**5. Are the new tests discriminating?** Verified in both directions, which is stronger than the author's claim:
- Reconstructed the rejected round-1 attempt as a mutant (widened the anchor to `(--?[a-zA-Z]|[|;&<>()])`). All **9** r2 cases return exit 0 (leak) against it and exit 2 against head — they catch exactly the defect I flagged.
- Ran the test file against `dev`'s hook: **5 fail** (cases 14–18, the #1039 bug) and the 9 r2 cases pass — correct, since `dev` already has the #227 property.
- Head: **29/29**. `shellcheck -S warning` clean on both files. CI green, including the full hook suite.

### Blocking for merge (no new commit required)

**The PR description documents the rejected implementation as the shipped one.** The Summary still says *"Widens the flag extractor's anchor to accept single-dash flags and shell-operator boundaries, and strips a matched surrounding quote pair in the unquoted fallback"*, and "Notes for the reviewer" still discusses matched-pair quote-stripping and *"The anchor now also accepts `(`"*. None of that is in this diff — `extract_flag_value` is byte-identical to `dev` and there is no quote-stripping. Counts are stale too (says 7 new cases / 20 total / 5-of-7 fail; actual is 16 new / 29 total).

This matters more than usual here: someone doing archaeology in six months on *why `extract_flag_value` looks the way it does* would read this body and conclude the anchor was widened — the precise misunderstanding that produced the defect twice. Please rewrite the Summary and Notes to describe the extractor **split** before merging. It's a body edit, so it doesn't change the SHA and doesn't invalidate this approval.

- `nit:` The fail-closed block runs *before* the skip-marker check, so `<!-- private-refs: allow -->` in `--body` does not clear it — I confirmed this empirically (still exit 2). Blocking is the safe direction, so this isn't a security issue, but the block message itself advises *"To reference a registered project deliberately, add this marker to the body"*, which provably won't work on this path. Either drop that paragraph (the only correct remedy here is "fix the path") or move the check below the skip-marker gate.
- `suggestion:` `extract_path_flag`'s unquoted branch takes a single whitespace-delimited token, so an unquoted path containing a space silently truncates and then fail-closes. Correct direction, but a one-line comment saying so would save the next reader a trace.

### Verdict
**APPROVED** — the security property is verified correct and the tests discriminate against both the original bug and the rejected fix. Merge only after the PR description is corrected.

---
Reviewed commit: `7e3b7e3eba678f770729b653877156f826f3ff8a`

Issue #1038, opened 2026-07-27
[Bug] PR-create hooks mis-parse --body-file: quotes retained, and value invisible unless the flag is last


atlas-apex:
Two PR-creation hooks share a `--body-file` handling defect that makes them block legitimate PRs. Both fail **closed**, so this is usability and false-positive noise rather than a security hole — but between them they made a valid `gh pr create` unrunnable until the flag was placed in one very specific position, which no operator would guess.

Both were hit in one session while opening a routine fork-sync PR.

## Affected

- `.claude/hooks/require-agdr-for-arch-pr.sh` (`extract_flag_value`)
- `.claude/hooks/validate-pr-create.sh` (`--body-file` path extraction)

### Bug 1 — value invisible unless the flag is last (`require-agdr-for-arch-pr.sh`)

**Given** `extract_flag_value()` anchors its match on the value being followed by the next recognised flag boundary or end-of-string:

```
re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
```

**When** a `gh pr create` places a redirection or pipe after `--body-file`, e.g. `gh pr create ... --body-file /path/body.md 2>&1 | tail -5`
**Then** the match fails, `BODY_FILE` resolves empty, `BODY_FILE_CONTENT` is empty, and the `<!-- agdr: not-applicable -->` marker inside that file is never seen — so the PR is **BLOCKED** — expected that the marker be honoured regardless of what follows the flag on the command line

This matters especially because the hook's own block message *instructs* the operator to add that marker. Following the instruction silently doesn't work if the command ends in `| tail`, and nothing in the output hints at why.

### Bug 2 — quotes retained in the extracted path (`validate-pr-create.sh`)

**Given** a `gh pr create` with a quoted `--body-file` path
**When** the path is passed as `--body-file "/path/to/body.md"`
**Then** the hook emits

```
WARN: validate-pr-create.sh: --body-file '"/path/to/body.md"' not readable from hook context; section check may miss content.
```

— note the literal `"` characters retained *inside* the extracted value — and then blocks with `PR body missing required '## Testing' section` and `'## Glossary' section`, when both are present in the file — expected that surrounding quotes be stripped and the file read

## Repro

Bug 1:

1. Put `<!-- agdr: not-applicable -->` in a body file.
2. On a branch whose diff touches `.github/workflows/**`, run:
[...]

Issue #113, opened 2026-04-24
[Chore] Require Testing section in PR body (extend validate-pr-create.sh)


atlas-apex:
## Driver

`.claude/rules/pr-quality.md` mandates a Glossary section in every PR (enforced by `validate-pr-create.sh`). It *also* describes a Testing section as part of the canonical PR description in `workflows/code-review.md` — *"Testing: How to verify this works"* — but there is no hook checking for it. Result: PRs routinely ship without reviewable test instructions, forcing reviewers to reverse-engineer "how do I know this works?" from the diff.

## Scope

Extend `validate-pr-create.sh` with a required-section check for PR bodies, driven by the same config schema as the issue-structure validator (apexyard#107).

```yaml
# .claude/project-config.json
pr:
  required_sections:
    - "Summary"
    - "Testing"
    - "Glossary"
  skip_marker: "<!-- pr-sections: skip -->"
```

Validation:

1. Extract the PR body from `--body` / `--body-file` / stdin.
2. For each section in `pr.required_sections`, require an `## <Section>` (or `### <Section>`) heading followed by non-empty content.
3. On fail: exit 2 naming the missing section(s).
4. Skip marker honours legitimate exceptions (trivial dependency bumps, pure version-tag commits).

## Acceptance Criteria

- [ ] `validate-pr-create.sh` blocks on missing Testing section (in addition to the existing Glossary check).
- [ ] Section list reads from project config; ships the canonical three as default.
- [ ] Skip marker passes with a warning.
- [ ] The check is case-insensitive on the heading text, strict on the empty-content rule.
- [ ] Fixtures in `.claude/hooks/tests/` cover each missing section and the skip path.
- [ ] `docs/rule-audit.md` entry updated.
- [ ] Config and helper code shared with apexyard#107 (both hooks use the same reader).

## Risks / Dependencies

- Risk of config drift if the issue hook and the PR hook diverge. Mitigation: shared helper (`_lib-read-ticket-schema.sh` or equivalent) introduced in apexyard#107.
- Small risk of friction for PR authors. Mitigation: skip marker + helpful error message.
- Depends on apexyard#109 (config location) and apexyard#107 (schema reader).

## Glossary

| Term | Definition |
|------|------------|
| Required section | A Markdown heading that must be present in the PR body with non-empty content. |
| Testing section | The heading under which the PR author documents how to verify the change — commands, manual steps, screenshots, etc. |

atlas-apex:
[...]

PR #1041, opened 2026-07-27
fix(#1038): PR-create hooks mis-parse a quoted --body-file path

atlas-apex:
## Summary

- **Fixes two hooks that could not read a quoted `--body-file` path** and then blamed the PR author for it. `require-agdr-for-arch-pr.sh` blocked PRs for a missing AgDR that *was* declared (the `<!-- agdr: not-applicable -->` marker lives in the file it failed to open), and `validate-pr-create.sh` reported `## Testing` / `## Glossary` missing when both were present.
- **Splits path-valued from content-valued flag extraction.** This is the fix. `extract_flag_value` is left **byte-identical to `dev`**; a new non-greedy `extract_path_flag` handles `--body-file`.
- **Adds a matched-quote strip** to `validate-pr-create.sh`'s own sed-based path extraction.

## The bug

```
--body-file "/p/b.md" 2>&1 | tail -5   ->  "/p/b.md"
--body-file "/p/b.md" -t "x"           ->  /p/b.md" -t "x
```

The greedy content matcher cannot terminate a quoted path followed by a shell operator, so extraction fell through to the unquoted branch and returned the token **with quotes attached**. `[ -f "\"/p/b.md\"" ]` is then false and the file is never read.

Worth stating precisely, because the original issue got it wrong: the failing combination is a **quoted value followed by a non-`--flag` token**. An *unquoted* path with a trailing pipe always worked, and a quoted path followed by `--base dev` also worked. That narrowness is why it survived.

## The wrong turn on the way here

**An earlier revision of this PR widened `extract_flag_value`'s shared anchor to accept shell operators. That was wrong and was reverted.** It fixed the path case and broke the content case: `sub()` is leftmost-first and each alternative ended in `.*`, so a `--body` containing a quote followed by a boundary character truncated there.

Measured on this hook's own content extractor: `dev` preserves **8/8** body tails, the widened version preserves **0/8**. In the sibling leak hook (#1039 / PR #1040) the same change was measurably *worse than no fix at all* — 5 failures became 9.

So the root cause is that one extractor was serving two incompatible jobs:

- **CONTENT** (`--title` / `--body`) may contain quotes, pipes, semicolons, markdown tables. Extraction must be **greedy**, terminating only on a real flag boundary. `extract_flag_value` is therefore unchanged.
- **PATH** (`--body-file` / `-F`) is a filesystem path and never contains a quote. Extraction must be **non-greedy**, stopping at the first closing quote.

`extract_path_flag` is new; the shared extractor is untouched. The non-greedy `[^"]*` is safe here for exactly the reason it was unsafe there.

`validate-pr-create.sh` is unaffected by the truncation class — it extracts the path with its own sed and never runs content through a greedy matcher — so it gets only the matched-quote strip.

## Testing

`test_flag_extractor_body_file.sh` — **22 cases**, all passing.

**Part 1** drives both extractors directly out of the hook source: the four quoted-plus-operator shapes plus the single-quoted variant, three regression guards for shapes that already worked, and **9 content-tail cases** asserting a `--body` survives a quote followed by each boundary character (`| ; & < > ( )`), a single-dash flag, and a markdown Glossary table. Those content guards fail **8/8** against the widened-anchor approach, which is what makes keeping the extractors separate load-bearing rather than stylistic.

**Part 2** runs `validate-pr-create.sh` end-to-end in a `git init` sandbox on a conforming branch, with the shared `_lib-mock-gh.sh` shim so the ticket check resolves against synthetic data rather than the network.

That sandbox matters, and two earlier attempts at it failed:

1. Asserting on exit code from the ambient checkout — passed locally, failed 3/22 in CI where branch name and ticket existence differ.
2. Asserting on stderr instead — still failed, and **vacuously**: the hook exits on the ticket check (~line 447) long before the `--body-file` logic (~line 502), so in CI it never reached the code under test and every "message absent" assertion passed because nothing ran.

The current version asserts `rc == 0` for a complete body, which is reachable **only** if the hook opened the file and found its sections. Verified 22/22 locally and green in CI.

- Full hook suite: **118/118**

## Notes for the reviewer

Quote-stripping is **matched-pair only**. Stripping unconditionally would rewrite a file literally named `"x.md"`, so the hook would read a different file than `gh` does.

The identical extractor defect in `block-private-refs-in-public-repos.sh` fails **OPEN** — a silent leak rather than an over-block — and was fixed separately in #1040, now merged.

AgDR not applicable: a parser correctness fix in existing hooks. No new dependency, service, schema, or security control; reversible in one commit. It touches `.claude/hooks/**` but changes no gate semantics, only whether a hook can read the file it was always meant to read.

Refs #1038

## Glossary

| Term | Definition |
|------|------------|
| `extract_flag_value` | Greedy extractor for CONTENT-valued flags; terminates on a real flag boundary |
| `extract_path_flag` | Non-greedy extractor for PATH-valued flags; stops at the first closing quote |
| fail closed | On ambiguity, deny — what these two hooks correctly do, hence "false positive" not "leak" |
| vacuous test | One that passes because the code under test never ran |

### DIFF
--- a/.claude/hooks/require-agdr-for-arch-pr.sh
+++ b/.claude/hooks/require-agdr-for-arch-pr.sh
@@ -62,74 +62,98 @@ extract_flag_value() {
   # recognised flag boundary (whitespace + `--<letter>`) or end-of-string. This
   # captures the full quoted span, including any embedded quotes, without
   # bleeding past the start of the next flag.
-  #
-  # #1038 — WIDEN THE ANCHOR, AND STRIP QUOTES IN THE FALLBACK.
-  # The anchor accepted only `--<letter>` or end-of-string, so a quoted
-  # value followed by a shell operator or a single-dash flag missed the
-  # quoted branches entirely:
-  #
-  #   gh pr create ... --body-file "/p/b.md" 2>&1 | tail -5    (shell operator)
-  #   gh pr create ... --body-file "/p/b.md" -t "x"            (single-dash flag)
-  #
-  # Execution then fell through to the unquoted branch, which has no anchor
-  # and returned the token WITH its quotes attached. The caller's
-  # `[ -f "\"/p/b.md\"" ]` was false, BODY_FILE_CONTENT stayed empty, and the
-  # `<!-- agdr: not-applicable -->` marker inside that file was never seen —
-  # so the PR was blocked for a missing AgDR that was actually declared.
-  # Note an UNQUOTED path with a trailing pipe always extracted fine; the
-  # failing combination is specifically quoted-value + non-`--flag` follower.
-  #
-  # The same defect in block-private-refs-in-public-repos.sh fails OPEN
-  # instead of closed (it exits 0 when it recovers no body), which is why
-  # that one is tracked and fixed separately as #1039.
-  #
-  # Both changes only ever cause MORE text to be scanned, and the anchor now
-  # matches what validate-issue-structure.sh already adopted for #695.
   local flag_re="$1"
   local cmd="$2"
   printf '%s' "$cmd" | awk -v FLAG_RE="$flag_re" -v SQ="'" '
     { buf = (NR == 1 ? $0 : buf "\n" $0) }
     END {
       s = buf
-      # Next-token boundary: a flag (one or two dashes), a shell operator,
-      # or end-of-string. `--?[a-zA-Z]` rather than an interval expression
-      # so this stays portable across BSD/GNU awk.
-      ANCH = "([[:space:]]+--?[a-zA-Z]|[[:space:]]*[|;&<>()]|[[:space:]]*$)"
-      # Double-quoted value: greedy `(.*)` anchored on the boundary above.
-      re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"" ANCH
+      # Double-quoted value: greedy `(.*)` anchored on next flag or EOS.
+      re = "(" FLAG_RE ")[[:space:]]+\"(.*)\"([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
       if (match(s, re)) {
         chunk = substr(s, RSTART, RLENGTH)
         sub("^(" FLAG_RE ")[[:space:]]+\"", "", chunk)
-        sub("\"([[:space:]]+--?[a-zA-Z].*|[[:space:]]*[|;&<>()].*)?$", "", chunk)
+        sub("\"([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
         sub("\"[[:space:]]*$", "", chunk)
         print chunk
         exit
       }
       # Single-quoted value: same greedy + anchor treatment.
-      re = "(" FLAG_RE ")[[:space:]]+" SQ "(.*)" SQ ANCH
+      re = "(" FLAG_RE ")[[:space:]]+" SQ "(.*)" SQ "([[:space:]]+--[a-zA-Z]|[[:space:]]*$)"
       if (match(s, re)) {
         chunk = substr(s, RSTART, RLENGTH)
         sub("^(" FLAG_RE ")[[:space:]]+" SQ, "", chunk)
-        sub(SQ "([[:space:]]+--?[a-zA-Z].*|[[:space:]]*[|;&<>()].*)?$", "", chunk)
+        sub(SQ "([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
         sub(SQ "[[:space:]]*$", "", chunk)
         print chunk
         exit
       }
-      # Unquoted value: single token. Strip ONE matched surrounding quote
-      # pair (#1038) so a quoted value reaching this branch is still usable.
-      # Only a MATCHED pair, so a file literally named `"x.md"` is not
-      # silently rewritten to a different path than the one gh will read.
+      # Unquoted value: single token, embedded quotes irrelevant.
       re = "(" FLAG_RE ")[[:space:]]+[^[:space:]]+"
       if (match(s, re)) {
         chunk = substr(s, RSTART, RLENGTH)
         sub("^(" FLAG_RE ")[[:space:]]+", "", chunk)
-        if (length(chunk) >= 2) {
-          first = substr(chunk, 1, 1)
-          last  = substr(chunk, length(chunk), 1)
-          if ((first == "\"" && last == "\"") || (first == SQ && last == SQ)) {
-            chunk = substr(chunk, 2, length(chunk) - 2)
-          }
-        }
+        print chunk
+        exit
+      }
+    }
+  '
+}
+
+# extract_path_flag FLAG_RE COMMAND  (me2resh/apexyard#1038)
+#
+# A PATH-valued flag needs the OPPOSITE parsing strategy to a CONTENT-valued
+# one, and conflating them is the actual defect behind #1038:
+#
+#   CONTENT (--title / --body): may contain anything — quotes, pipes,
+#     semicolons, markdown tables. Extraction must be GREEDY and terminate
+#     only on a real flag boundary, or an embedded quote truncates the value.
+#     extract_flag_value above is therefore left exactly as #227 wrote it.
+#
+#   PATH (--body-file / -F): a filesystem path, which never contains a quote.
+#     Extraction must be NON-GREEDY — stop at the FIRST closing quote. The
+#     greedy content matcher cannot terminate `--body-file "/p/b.md" 2>&1 |
+#     tail`, so it fell through to the unquoted branch and returned the token
+#     WITH quotes attached. `[ -f ]` was then false, the body file was never
+#     read, and the `<!-- agdr: not-applicable -->` marker inside it was
+#     never seen — blocking the PR for a missing AgDR that WAS declared.
+#
+# An earlier attempt widened extract_flag_value's anchor to accept shell
+# operators. That fixed the path case and broke the content case: `sub()` is
+# leftmost-first and each alternative ended in `.*`, so a body containing a
+# quote followed by ` |` truncated there. In the sibling leak hook that same
+# change was measurably WORSE than no fix at all (see #1039). Splitting the
+# extractors is the correct fix; widening the shared one is not.
+extract_path_flag() {
+  local flag_re="$1"
+  local cmd="$2"
+  printf '%s' "$cmd" | awk -v FLAG_RE="$flag_re" -v SQ="'" '
+    { buf = (NR == 1 ? $0 : buf "\n" $0) }
+    END {
+      s = buf
+      # Double-quoted path: stop at the first closing quote.
+      re = "(" FLAG_RE ")[[:space:]]+\"[^\"]*\""
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+\"", "", chunk)
+        sub("\"$", "", chunk)
+        print chunk
+        exit
+      }
+      # Single-quoted path: same treatment.
+      re = "(" FLAG_RE ")[[:space:]]+" SQ "[^" SQ "]*" SQ
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+" SQ, "", chunk)
+        sub(SQ "$", "", chunk)
+        print chunk
+        exit
+      }
+      # Unquoted path: a single whitespace-delimited token.
+      re = "(" FLAG_RE ")[[:space:]]+[^[:space:]]+"
+      if (match(s, re)) {
+        chunk = substr(s, RSTART, RLENGTH)
+        sub("^(" FLAG_RE ")[[:space:]]+", "", chunk)
         print chunk
         exit
       }
@@ -142,7 +166,8 @@ BODY=$(extract_flag_value '--body|-b' "$COMMAND")
 
 # --body-file <path> / -F <path> (only when -F's value is NOT a key=val pair,
 # because `gh api -F body=@file` uses the same flag letter).
-BODY_FILE=$(extract_flag_value '--body-file' "$COMMAND")
+# #1038: a PATH, so use the non-greedy path extractor.
+BODY_FILE=$(extract_path_flag '--body-file' "$COMMAND")
 if [ -z "$BODY_FILE" ]; then
   F_VAL=$(echo "$COMMAND" | sed -nE "s/.*(^|[[:space:]])-F[[:space:]]+\"([^\"]*)\".*/\2/p" | head -1)
   if [ -z "$F_VAL" ]; then
--- a/.claude/hooks/require-agdr-for-arch-pr.sh
+++ b/.claude/hooks/require-agdr-for-arch-pr.sh
@@ -62,37 +62,74 @@ extract_flag_value() {
   # recognised flag boundary (whitespace + `--<letter>`) or end-of-string. This
   # captures the full quoted span, including any embedded quotes, without
   # bleeding past the start of the next flag.
+  #
+  # #1038 — WIDEN THE ANCHOR, AND STRIP QUOTES IN THE FALLBACK.
+  # The anchor accepted only `--<letter>` or end-of-string, so a quoted
+  # value followed by a shell operator or a single-dash flag missed the
+  # quoted branches entirely:
+  #
+  #   gh pr create ... --body-file "/p/b.md" 2>&1 | tail -5    (shell operator)
+  #   gh pr create ... --body-file "/p/b.md" -t "x"            (single-dash flag)
+  #
+  # Execution then fell through to the unquoted branch, which has no anchor
+  # and returned the token WITH its quotes attached. The caller's
+  # `[ -f "\"/p/b.md\"" ]` was false, BODY_FILE_CONTENT stayed empty, and the
+  # `<!-- agdr: not-applicable -->` marker inside that file was never seen —
+  # so the PR was blocked for a missing AgDR that was actually declared.
+  # Note an UNQUOTED path with a trailing pipe always extracted fine; the
+  # failing combination is specifically quoted-value + non-`--flag` follower.
+  #
+  # The same defect in block-private-refs-in-public-repos.sh fails OPEN
+  # instead of closed (it exits 0 when it recovers no body), which is why
+  # that one is tracked and fixed separately as #1039.
+  #
+  # Both changes only ever cause MORE text to be scanned, and the anchor now
+  # matches what validate-issue-structure.sh already adopted for #695.
   local flag_re="$1"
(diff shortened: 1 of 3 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1041

**Commit**: `d5c890029d3e27b90635b64ba9537791bcff863a`

### Summary

Splits path-valued flag extraction (`--body-file` / `-F`) from content-valued extraction (`--title` / `--body`) in `require-agdr-for-arch-pr.sh`, and adds a matched-pair quote strip to `validate-pr-create.sh`. This is the corrected approach after `1a3467d` shipped the widened-shared-anchor design that was rejected on the sibling PR #1040.

**The production-code fix is correct.** I verified it independently rather than relying on the new tests, and it holds on every dimension I probed. It is blocked only by a red CI caused by the PR's own new test file.

### Checklist Results
- Testing:                    **Fail** — new test file is red in CI
- PR Description & Glossary:  **Fail** — description documents the rejected approach
- Technical Decisions (AgDR): Pass (`agdr: not-applicable`, correctly justified — parser correctness fix, no gate semantics changed)

### What I verified independently

1. **`extract_flag_value` is byte-identical to `upstream/dev`.** Confirmed by md5 and `diff` of the extracted function — no delta at all. The #227 property is not put back in question.
2. **Content extraction survives every boundary shape.** Built my own cases rather than trusting the added tests: a `--body` containing an escaped quote followed by each of `| ; & < > ( )` plus `$` and `` ` ``, a following `-t` / `-F` single-dash flag, a multi-line markdown table row with quoted cells, a body-last position, and a single-quoted body. **14/14 keep the tail.**
3. **`extract_path_flag` is sound and a strict improvement.** Side-by-side against `dev`'s extractor: quoted+pipe, quoted+single-dash-flag, single-quoted+pipe, and quoted-path-containing-a-space all now extract correctly where `dev` returned quote-laden or truncated garbage. **No shape regressed.** The `-F` fallback beneath it is untouched and still guards the `key=val` case.
4. **The fix is load-bearing, and the raw-`$COMMAND` grep does not rescue it.** End-to-end against the real hook, `dev` MISSES the `<!-- agdr: not-applicable -->` marker for a quoted body-file path (both `| tail` and `-t "x"` shapes) while this PR SEES it. The marker lives in the file, which is not a substring of `$COMMAND` — so the raw-command fallback only covers the inline-`--body` case. #1038's symptom is genuinely resolved.
5. **`validate-pr-create.sh`'s claim holds.** Its body content comes from a haystack of file-content + raw command (line ~540), never from a greedy `extract_flag_value`-style matcher, so the truncation class does not apply. I also confirmed the matched-pair strip cannot resolve to a *different existing* file than `gh` reads: `"qt/x.md"` → `qt/x.md` (same file gh opens); `'qt/"x.md"'` → `qt/"x.md"` (quote-named file preserved); `"\"qt/x.md\""` → a non-existent path (fails closed). The spoof concern named in "Notes for the reviewer" is correctly handled.
6. **The content guards are genuinely discriminating.** I spliced the rejected widened anchor from `1a3467d` back in and re-ran: it **truncates 9/9** of the shapes the current code preserves. These are real regression guards, not vacuous ones.

### Issues Found

**BLOCKING 1 — CI is red, and the failure is this PR's own new test file.**

`hook test suite` fails: `test_flag_extractor_body_file.sh` returns rc=1 with **3 of 22 cases failing** in CI (suite total 117/118).

```
FAIL: validate-pr-create: unquoted path, complete body → pass — expected exit 0, got 2
FAIL: validate-pr-create: QUOTED path, complete body → pass (#1038) — expected exit 0, got 2
FAIL: validate-pr-create: single-quoted path, complete body → pass (#1038) — expected exit 0, got 2
```

All three are Part 2 end-to-end cases; the file passes 22/22 on my machine, so the tests are **environment-dependent**. `check_hook` runs the hook from a `mktemp` dir but `validate-pr-create.sh` still resolves ambient git state — branch name and ticket-ID validation among other things — which differs under the CI checkout (detached HEAD / no ticket-bearing branch name). The hook exits 2 for a reason that has nothing to do with the `--body-file` parsing the case intends to assert.

This violates `.claude/rules/pr-quality.md` § "No Red CI Before Merge" and workflow Gate 5 (CI green). Note the irony worth avoiding: a PR fixing false-positive hook failures cannot itself land a test that false-fails in CI.

Suggested fix: isolate Part 2 from ambient git state — assert on the specific `## Testing` / `## Glossary` diagnostic on stderr rather than on the raw exit code, or neutralise the unrelated validations (init a throwaway repo with a ticket-bearing branch in `$TMPDIR`) so the case fails only for the reason it is testing.

**BLOCKING 2 — the PR description documents the approach that was rejected and reverted.**

Summary bullet 2 still reads *"Widens the flag extractor's anchor to accept single-dash flags and shell-operator boundaries, matching what `validate-issue-structure.sh` already adopted for #695."* That is `1a3467d` — the design `d5c8900` deliberately abandoned, and the one whose failure mode the new code comments explicitly warn against. "## The bug" carries the same stale framing. A reviewer reading the description would conclude the rejected approach shipped; only the diff reveals otherwise.

Two stale factual claims alongside it:
- *"Full hook suite: 118/118"* — CI reports 117/118 with this file failing.
- *"New `test_flag_extractor_body_file.sh` — 13 cases … 7 of the 13 fail against the pre-fix hooks"* — the file now has 22 cases.

Please rewrite the description around the path/content split (the code comments in `extract_path_flag` already state it well) and refresh the counts.

- `nit:` **`--body-file=path` (equals form) is not handled** by either extractor — both return empty. `gh` accepts it (Go pflag). Pre-existing, identical on `dev`, and fails closed, so not a regression — but worth a follow-up since this PR is otherwise closing the parsing gaps.
- `nit:` **Inconsistent handling of paths containing spaces.** `extract_path_flag` now resolves `--body-file "/p/my body.md"` correctly; `validate-pr-create.sh`'s `[^[:space:]]+` token grab still yields `"/p/my` and fails closed. Not introduced here, but the two hooks now diverge on the same input.
- `nit:` **Decoy `--body-file` inside a `--body` value can win.** `extract_path_flag` scans the whole command for its double-quoted branch first, so `--body "... --body-file \"/etc/passwd\" ..." --body-file /p/real.md` returns the decoy. Behaviour is identical on `dev`, it fails closed, and the author of the command already controls the body — so this is a robustness nit, not a vulnerability. A left-boundary anchor on `FLAG_RE` (`(^|[[:space:]])`) would also tighten the incidental `--x--body-file` match.

### Verdict
**CHANGES REQUESTED**

The parser fix itself is well-reasoned and, on my independent testing, correct on all seven dimensions — including the two properties that mattered most: `extract_flag_value` is untouched, and the new content guards demonstrably catch the rejected approach. The blockers are a red CI caused by this PR's own environment-dependent test, and a description that still advertises the reverted design. Neither requires touching the production code.

---
📌 Reviewed commit: `d5c890029d3e27b90635b64ba9537791bcff863a`

atlas-apex (commented):
## Code Review: PR #1041 (re-review)

**Commit**: `c919e7f19d9f1fd18952fc6b16c4195b3249269a`

### Summary

Re-review after the Part 2 rewrite from exit-code assertions to stderr-content assertions. The production fix is unchanged and remains correct. The rewrite did not clear the blocker — CI is still red on this PR's own test file, and the specific vacuity risk you were guarding against is now empirically demonstrated rather than hypothetical.

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: **Fail** — the new test fails in CI and its passing assertions are vacuous there
- Security: Pass
- Performance: Pass
- PR Description & Glossary: **Fail** — Summary bullet 2 still advertises the rejected approach; three counts are stale
- Technical Decisions (AgDR): N/A — `<!-- agdr: not-applicable -->` present and justified
- Adopter Handbooks: N/A — none loaded
- Fallow Static Analysis: N/A — not JS/TS

### Confirmed good (carried forward from `d5c8900`)

Production code is **byte-identical to what I already approved**:

```
git diff d5c8900 c919e7f -- .claude/hooks/require-agdr-for-arch-pr.sh .claude/hooks/validate-pr-create.sh
(empty)
```

`c919e7f` touches only `test_flag_extractor_body_file.sh` (+62/-22). The split extractor, the matched-pair-only quote stripping, and the `validate-pr-create.sh` sed fix all stand as reviewed.

Part 1 of the test remains the strong half — it drives `extract_path_flag` / `extract_flag_value` straight out of the hook source, so it is independent of ambient state and genuinely pins the parsing contract. All 17 Part 1 cases pass in CI.

---

### ⛔ Blocking 1 — CI is red, and the absence assertions are vacuous in CI

`gh pr checks 1041` on `c919e7f`: **hook test suite → fail**, run [30277480009](https://github.com/me2resh/apexyard/actions/runs/30277480009), `headSha` confirmed as `c919e7f`. Result: `PASS=117 FAIL=1 TOTAL=118`, the one failure being `test_flag_extractor_body_file.sh` at **21/23**.

The two failures are the `check_stderr_present` cases, and their stderr tells the whole story:

```
FAIL: validate-pr-create: thin body still reported (unquoted) — expected stderr to report [## Testing]
   stderr: BLOCKED: PR title references #1038 but issue #1038 does not
           exist in me2resh/apexyard.
FAIL: validate-pr-create: thin body still reported (quoted) — expected stderr to report [## Testing]
   stderr: BLOCKED: PR title references #1038 but issue #1038 does not
           exist in me2resh/apexyard.
```

This is exactly the vacuity failure mode. In `validate-pr-create.sh` the ticket-existence check exits at **line 447**; the `--body-file` extraction does not begin until **line 502**. In CI the tracker lookup for `#1038` does not resolve, the hook fails closed at 447, and control never reaches the body-file logic at all.

Consequence: in CI, all four `check_stderr_absent` cases — including the two that carry the actual #1038 regression protection (`QUOTED path is read`, `quoted path — sections found`) — pass **because nothing ran**, not because extraction worked. `"not readable"` and `"## Testing"` are absent for the trivial reason that the code emitting them was never reached. Part 2 currently provides **zero** regression coverage in the only environment where it runs automatically.

So, to answer the question directly: **yes, the two `check_stderr_present` cases do genuinely guard against the early exit, and they worked.** They are the reason this is visible rather than a silently-green vacuous test — that is a real design improvement over the exit-code version, credit where due. But the guard firing means the guarded assertions are not trustworthy, and the build is red. Both have to be resolved, not just the second.

The `23/23 under a simulated CI environment` claim did not reproduce real CI. The simulation modelled detached HEAD and absent gh auth, but not the tracker lookup failing for `#1038` specifically, which is the path that actually trips.

**Suggested fix — there is a working in-repo pattern.** `test_validate_pr_required_sections.sh` exercises this same hook's section logic end-to-end and passes in CI. It builds an isolated sandbox first:

```bash
git init -q
git checkout -q -b chore/GH-113-test 2>/dev/null || git checkout -q -B chore/GH-113-test
```

The new test `cd`s into a bare `mktemp -d` with no `git init` and no conforming branch, so ambient resolution diverges. Mirroring that sandbox setup is the smallest change likely to make Part 2 both green and non-vacuous. If the tracker lookup still fires, a second option is to drive the body-file read + section check directly out of the hook source the way Part 1 already does for the extractors — that is the approach that made Part 1 robust, and it applies equally here.

Either way, please keep the two `check_stderr_present` cases. They are what makes the absence assertions meaningful, and they have now proven their worth.

### ⛔ Blocking 2 — PR description still describes the rejected approach

Flagged last round as blocking-for-merge; unchanged at `c919e7f`.

Summary bullet 2 still reads:

> **Widens the flag extractor's anchor** to accept single-dash flags and shell-operator boundaries, matching what `validate-issue-structure.sh` already adopted for #695.

That is `1a3467d`, which `d5c8900` replaced with the two-extractor split precisely because widening the shared anchor reopens the #227 leak. The description advertises a design the diff does not implement — and the test file's own comments now argue against it explicitly (`the guard against "fix the path case by widening the shared anchor"`). A future reader reconstructing intent from the merge commit gets the wrong answer.

Three stale counts alongside it:

| Claim in body | Actual |
|---|---|
| `13 cases` | 23 |
| `7 of the 13 fail against the pre-fix hooks` | recount against the 23 |
| `Full hook suite: 118/118` | CI reports `117/118`, FAIL=1 |

The last one is the one to fix carefully — it asserts green while CI is red.

- The Part 2 comment block explaining why stderr content is the right observable is genuinely good technical writing and should survive the next rewrite intact.
- Once Part 2 is green, consider re-running the discriminating check (how many of the 23 fail against pre-fix hooks) and quoting the new number — it is the most persuasive line in the Testing section.

### Verdict

**CHANGES REQUESTED**

The production fix is correct and I have now verified it twice; nothing in the hooks needs to change. Both blockers are in the test file and the PR body. A PR whose subject is *hooks producing false failures* should not merge with its own test false-failing, and should not carry a description advertising the approach it deliberately rejected.

---
📌 Reviewed commit: `c919e7f19d9f1fd18952fc6b16c4195b3249269a`

atlas-apex (commented):
## Code Review: PR #1041 (round 3)

**Commit**: `aef1970304996597e47171bd8bea6e63ac64d926`

### Summary

Third round. Rounds 1 and 2 confirmed the production code was correct and blocked only on non-production items: CI was red, and the Part 2 rewrite passed **vacuously** — `validate-pr-create.sh` exits on the ticket check (~L447) long before the `--body-file` logic (~L502), so in CI the hook never reached the code under test.

`aef1970` rewrites Part 2 on the `test_validate_pr_required_sections.sh` pattern: a `git init` sandbox on a conforming branch (`chore/GH-113-test`), hooks + defaults copied in, and the shared `_lib-mock-gh.sh` shim resolving the ticket check against synthetic OPEN data. Both blockers clear. I verified each claim rather than accepting it.

### Verification

**1. CI is green on this exact SHA.** All checks `success` against `aef1970` (CodeQL `neutral`/skipped). The `hook test suite` job log confirms the new file ran and passed, with the aggregate matching the PR body exactly:

```
PASS .claude/hooks/tests/test_flag_extractor_body_file.sh
hook test suite: PASS=118  FAIL=0  SKIP(quarantined)=0  TOTAL=118
```

The new test needs no registration — `tests.yml` auto-discovers `test_*.sh` under `.claude/**/tests/`.

**2. Part 2 is non-vacuous — verified by mutation, not by argument.** This was the crux and I probed it directly rather than reasoning about it. I copied the tree, reverted the `#1038` matched-quote strip in `validate-pr-create.sh`, and re-ran:

```
FAIL: validate-pr-create: QUOTED path, complete body -> rc 0 (#1038) — expected rc=0, got 2
   stderr: WARN: --body-file '"/…/body.md"' not readable from hook context
FAIL: validate-pr-create: single-quoted path, complete body -> rc 0 (#1038) — expected rc=0, got 2
Passed: 20  Failed: 2
```

Exactly the two quoted cases fail; the unquoted control still passes. Two things this settles: the failure is attributable to the quoting dimension alone, and the `not readable` WARN is emitted from **L534** — proof the hook now executes *past* the ticket check that made round 2 vacuous. `rc == 0` is reachable only through the section scan, which is only reachable if the file was opened.

I ran the same probe on `extract_path_flag`, dropping its quoted branches so extraction falls through to the unquoted token grab (the original `#1038` bug): **6 Part 1 cases fail**, all five `#1038` shapes plus `quoted path + following --flag`. Both halves of the fix are covered by tests that genuinely fail when the fix is reverted.

**3. The sandbox is hermetic.** No dependence on the ambient branch (its own `git init` on a conforming name), no real `gh`. I tested the pass-through risk directly — `_lib-mock-gh.sh` resolves the real `gh` for unintercepted calls, so I shadowed it with a poison binary that exits 66 and logs to stderr, then re-ran the suite: **22/22, no poison invocation**. The mock is genuinely on PATH for the hook process — `export PATH=…` inside the subshell, not a `PATH=… jq` prefix that would scope it to `jq` and leave the hook to find the real `gh`. The inline comment on that line documents the exact trap, which is a good use of a comment.

**4. Production code untouched since I approved it at `d5c8900`.** `git diff d5c8900 aef1970` restricted to both hooks is **empty** — the delta is 114/31 lines in the test file only. `extract_flag_value` remains **byte-identical to `origin/dev`** (verified by diffing the extracted function against `origin/dev`), so the rejected widened-anchor approach and its `#227` leak regression stay reverted.

**5. PR description is accurate.** The stale widened-anchor claims are gone, replaced by an honest "The wrong turn on the way here" section that records why the approach was rejected and what it measured. Counts check out: 22 cases (8 `check_extract` + 9 `check_content` + 5 `check_prc`), 118/118 suite. One small residue below.

### Checklist Results

- Architecture & Design:     Pass — the two-extractor split is the right cut. Path values never contain quotes so non-greedy is safe; content values do, so greedy is required. One function serving both jobs was the root cause.
- Code Quality:              Pass — matched-pair-only quote stripping is correct (a file literally named `"x.md"` still resolves to the path `gh` reads). shellcheck clean.
- Testing:                   Pass — mutation-verified in both directions.
- Security:                  Pass — trust-chain change (`.claude/hooks/**`), so Heavy tier. Both hooks fail **closed**; this narrows an over-block, it does not widen an allow. No gate semantics changed. The fail-**open** sibling defect was correctly split out to #1040.
- Performance:               Pass — N/A.
- PR Description & Glossary: Pass — glossary present; `vacuous test` is a nice entry given this PR's history.
- Summary Bullet Narrative:  Pass — all three bullets carry a verb and a consequence.
- Technical Decisions (AgDR):Pass — `<!-- agdr: not-applicable -->` with a written rationale. Correct: a parser correctness fix in existing hooks, no new dependency/schema/security control, reversible in one commit. Rail 1 (trust-chain is always material) is satisfied by the marker carrying an explicit justification rather than being a bare bypass.
- Adopter Handbooks:         N/A — `clean-architecture-layers` and `migration-safety` don't apply to a shell-hook parser fix; commit messages are conventional and descriptive.
- Fallow Static Analysis:    N/A — no JS/TS in the diff.

### Issues Found

None blocking.

### Suggestions

**nit: one stale count survives in the PR body.** The Testing section says "**9 content-tail cases**" and then "Those content guards fail **8/8** against the widened-anchor approach". There are 9 such guards, so `8/8` reads as a carry-over from the earlier 8-shape measurement quoted higher up. Cosmetic only — the body-edit is optional and does not need a new commit.

**suggestion: the two "thin body still blocked" cases are not discriminating.** They pass under the reverted-fix mutation too, because a thin body blocks either way — for the *wrong* reason (file unreadable) rather than the intended one (sections missing). The `rc == 0` cases carry all of the mutation-detection weight. `check_prc` already accepts an `absent` argument that both thin cases pass `""` for; passing `"not readable"` there would make them fail when the file was never opened, and turn a pair of pure regression guards into real discriminators. Cheap, and worth folding into the next touch of this file rather than a respin here.

### Verdict

**APPROVED**

Marker written at `aef1970304996597e47171bd8bea6e63ac64d926`. All three rounds' blockers are cleared, and the two that mattered were confirmed by reverting the fix and watching precisely the right tests fail.

---
Reviewed commit: `aef1970304996597e47171bd8bea6e63ac64d926`

Issue #1042, opened 2026-07-27
[Bug] Human gate is on the review skills, not the approval skills — /approve-merge is model-invocable


atlas-apex:
The mechanical human gate sits on the **review** skills, which are harmless, while the **approval** skills — the ones that write approval markers and merge — are model-invocable and guarded only by prose.

## Affected

- `.claude/skills/{code-review,security-review,design-review}/SKILL.md` — `disable-model-invocation: true`
- `.claude/skills/{approve-merge,approve-design,approve-architecture}/SKILL.md` — model-invocable
- `.claude/hooks/auto-code-review.sh` — instructs the orchestrator to run a skill it cannot invoke

## Current state

| Skill | What it does | Model-invocable? |
|---|---|---|
| `/code-review` | Triggers an independent reviewer sub-agent | **NO** |
| `/security-review` | Triggers Hakim | **NO** |
| `/design-review` | Triggers Tariq | **NO** |
| `/approve-merge` | **Writes the CEO marker and merges** | **YES** |
| `/approve-design` | Writes the design approval marker | **YES** |
| `/approve-architecture` | Writes the architecture approval marker | **YES** |

## Given / When / Then

**Given** `auto-code-review.sh` fires on every `gh pr create` and instructs: *"IF YOU ARE THE ORCHESTRATOR: Run the code review through the /code-review skill on PR #N"*
**When** the orchestrator attempts exactly that
**Then** the Skill tool refuses with `Skill code-review cannot be used with Skill tool due to disable-model-invocation` — expected that the framework not instruct an action it mechanically forbids

**Given** `/approve-merge` writes the CEO marker and runs the merge in one turn, and its own SKILL.md states *"The discrete approval moment is the invocation of /approve-merge"*
**When** a model invokes it
**Then** it succeeds — expected that the one step authorising an irreversible, externally-visible merge require a human, since that is precisely what `pr-workflow.md` demands in prose

## Repro

1. Open any PR. `auto-code-review.sh` prints the instruction to run `/code-review <pr>`.
2. Attempt it as the orchestrator → refused by `disable-model-invocation`.
3. Inspect `.claude/skills/approve-merge/SKILL.md` → `disable-model-invocation: false`.

## Why this matters in both directions

[...]

Issue #1026, opened 2026-07-26
[Bug] Split-path marker write evades the review-marker gate (4th evasion; decide direction, do not add a 5th regex)


atlas-apex:
## Given / When / Then

- **Given** no matching `active-reviewer` session marker for the target PR,
- **When** a caller writes a review marker with the **directory portion of the path held in a shell variable** — e.g. `S=.claude/session/reviews; printf 'x\n' > "$S/owner__repo__8888-rex.approved"`,
- **Then** the write **succeeds**. The gate does not fire.

The identical write with a fully literal path **is** blocked correctly:

```
printf 'x\n' > .claude/session/reviews/owner__repo__7777-rex.approved
  -> BLOCKED, exit 2, target and PR correctly resolved
```

So the gate's protection depends on how the path happens to be spelled, not on what the command does.

## Why it evades both detection paths

`warn-review-marker-write.sh` has two matchers, and splitting the path defeats each for a different reason:

- **`_is_marker_target`** requires the literal substring `.claude/session/reviews/…-<role>.approved` in the command text. With the directory in `$S`, that substring never appears — only the basename does.
- **`_is_marker_plausible_indirect`** looks for `review_marker_path`, the literal `.claude/session/reviews/`, `active-reviewer`, or a variable whose *name* contains `marker`. A variable named `S` (or `D`, or `out`) matches none of those.

Neither matcher sees a marker write, so the command is treated as unrelated and passes.

## Provenance — pre-existing, and twice-identified

This is **not** introduced by #1011. It dates from the #962/#970 round, and the code reviewer surfaced it explicitly while reviewing #1011:

> "a split-path forgery *plus* the gratuitous word `active-reviewer` was blocked by the old hook and is allowed by the new one. But the same forgery with that one unrelated word deleted sails past **both** versions — so the signal only fired when an attacker volunteered a word they had no reason to type."

That was recorded in #1011's PR comment as out of scope and needing its own ticket. **The ticket was never filed** — this is it. It was then reproduced independently while triaging #1022, which is how it resurfaced.

## Why it matters, stated honestly

The realistic exposure is narrower than "the gate is broken":

- A **build agent inclined to forge** a marker writes the obvious literal path, which is still blocked — that is the #843 case the gate was built for, and it still works.
- The per-PR human merge approval remains the actual control, exactly as AgDR-0062 accepted and AgDR-0109 re-affirmed. This hole does not bypass that.

[...]

Issue #1019, opened 2026-07-25
[Bug] config_get() leaves embedded \r in multi-line jq output on Windows (Git Bash/MSYS2), breaking branch/PR-title validators


Manito2z:
## Affected

`.claude/hooks/_lib-read-config.sh` — `config_get()`, and its downstream consumers `.claude/hooks/validate-branch-name.sh` and `.claude/hooks/validate-pr-create.sh`.

## Given / When / Then

**Given** a Windows 11 environment running Git Bash / MSYS2 with the native `jq.exe` on PATH, and a `.claude/project-config*.json` whose `.branch.type_whitelist` array has more than one entry.

**When** `config_get()` is called with a jq filter that emits **multiple lines** (e.g. `.branch.type_whitelist[]`), and a caller joins the output with `paste -sd'|' -` to build a regex alternation (as `validate-branch-name.sh` and `validate-pr-create.sh` both do).

**Then** every internal line except the last comes back with a trailing `\r` baked in, so the resulting alternation pattern only matches a branch/PR-title segment containing a stray carriage return before the `/` — i.e. it can never match a real branch name or PR title. Expected: `config_get()` output should be plain `\n`-delimited text with no embedded `\r`, so the alternation matches normally.

## Root cause

The current implementation:

```bash
printf '%s' "$_CONFIG_CACHE" | jq -r "$filter" 2>/dev/null
```

On Windows, the native `jq.exe` writes its per-line output through a text-mode CRT stdout handle, which silently rewrites every `\n` it emits into `\r\n`.

- A **single-value** filter is unaffected — command substitution strips the one trailing newline, taking the `\r` with it.
- A **multi-line** filter captures an embedded `\r` before every *internal* newline, and those survive into the joined string.

## Impact

As of #20, `validate-branch-name.sh` and `validate-pr-create.sh` **block (exit 2)** rather than warn. On Windows, both validators effectively can never match a real branch type from the whitelist except in the rare case the stray `\r` happens to line up, so legitimately-named branches get hard-blocked.

## Suggested fix

Strip carriage returns from `config_get()`'s output:

```bash
printf '%s' "$_CONFIG_CACHE" | jq -r "$filter" 2>/dev/null | tr -d '\r'
```

[...]

PR #1043, opened 2026-07-27
fix(#1042): put the human gate on approval, not on review

atlas-apex:
## Summary

- **Moves the mechanical human gate from the review step to the approval step.** Review skills (`/code-review`, `/security-review`, `/design-review`) become model-invocable; approval skills (`/approve-merge`, `/approve-design`, `/approve-architecture`) become human-only. Strictly safer *and* less friction.
- **Makes the merge gate's human requirement mechanical rather than prose.** `/approve-merge` writes the CEO marker and merges in one turn — its own SKILL.md calls the invocation "the discrete approval moment", but nothing stopped a model from making it. Now only a human can.
- **Resolves a framework self-contradiction.** `auto-code-review.sh` fires on every `gh pr create` and instructs the orchestrator to run `/code-review <pr>` — which `disable-model-invocation: true` then refused. Every PR required the operator to hand-type it.
- **Gives `/approve-architecture` frontmatter.** It had none at all, so it silently defaulted to model-invocable.

## The problem

| Skill | What it does | Was |
|---|---|---|
| `/code-review` | Triggers an independent reviewer sub-agent | human-only |
| `/security-review` | Triggers Hakim | human-only |
| `/design-review` | Triggers Tariq | human-only |
| `/approve-merge` | **Writes the CEO marker and merges** | model-invocable |
| `/approve-design` | Writes the design approval marker | model-invocable |
| `/approve-architecture` | Writes the architecture marker | *(no frontmatter)* → model-invocable |

Locking a review buys nothing: independence comes from the skill spawning a **separate reviewer sub-agent with its own context**, not from who typed the command. Meanwhile the one step that is irreversible and externally visible had no mechanical guard.

## Why both halves must land together

This is the part worth reviewing carefully. The two settings were **coupled by accident**: because a model could not invoke `/code-review`, it could not obtain a Rex marker, so it could not merge. The locked review skill was incidentally the thing preventing autonomous merges.

So unlocking the review skills **alone** — the obvious reading of "fix the friction" — would have opened a fully autonomous `open → review → approve → merge` path with no human anywhere. Swapping both is what makes this safe; either half on its own is worse than the status quo.

`test_skill_invocability_gates.sh` asserts that coupling directly, not just the six individual values, so a future edit to one side fails loudly rather than silently reopening the path.

## Provenance

Not a regression — an original blind spot:

- `/code-review`'s flag dates to `8fa3c92` (2026-04-06), the initial `feat(#1): add Claude Code primitives` commit, never revisited.
- `/approve-merge` arrived via `feat(#11): require explicit per-merge CEO approval — never infer from "go"` with `disable-model-invocation: false` set **explicitly**.

The change whose stated purpose was making merge approval a deliberate human moment shipped the skill as model-invocable.

## What this costs you

The operator now types `/approve-merge <pr>` (and `/approve-design`, `/approve-architecture`) themselves. Saying "approved" in prose is no longer sufficient — which is exactly what `pr-workflow.md` has always demanded; it just wasn't enforced. In exchange, `/code-review` and friends stop being a manual step entirely.

## Testing

New `test_skill_invocability_gates.sh` — 7 assertions:

- three review skills are model-invocable, three approval skills are human-only
- an **absent** flag is a failure, not a pass — absent defaults to model-invocable, which is precisely how `/approve-architecture` went unguarded
- an explicit combination check: review-unlocked **and** approval-unlocked fails with an explanation, so the coupling is visible to whoever edits either side

Verified discriminating: **6 of 7 fail** against `dev`. The combination check correctly *passes* on `dev` too — review skills are locked there, so the dangerous pairing doesn't hold. That asymmetry is the test working as intended, not a gap.

- Full hook suite: **118/118** (clean baseline on `dev` is 117; this adds one file)

## Notes for the reviewer

`warn-review-marker-write.sh` is unaffected: it gates marker *writes* on the active-reviewer session marker, which the review skills manage regardless of who invokes them. The two-marker merge gate, the marker formats, and `block-unreviewed-merge.sh` are all unchanged — what changes is only who may press the button.

`pr-workflow.md` gains a section documenting the new mechanical enforcement, so the rule doc doesn't drift from the behaviour. (This repo already has one instance of that drift — `right-size-ceremony.md` and AgDR-0107 both cite `enforce-budget.sh` as the OSS watchdog, but it isn't tracked in git. Worth a separate ticket.)

Decision recorded in `AgDR-0110`, including the three rejected options and why "unlock reviews only" is the worst available trade.

Refs #1042

## Glossary

| Term | Definition |
|------|------------|
| `disable-model-invocation` | SKILL.md frontmatter flag; when `true`, only a human may invoke the skill |
| review skill | Spawns an independent reviewer sub-agent (Rex / Hakim / Tariq) |
| approval skill | Writes an approval marker that a merge gate reads |
| trust chain | The hooks and skills mechanically enforcing the merge, ticket, and marker gates |

### DIFF
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -189,7 +189,8 @@ defeating the entire purpose of /release-sync.
 Use --merge instead:
   gh pr merge ${PR_NUMBER} --repo ${CMD_REPO:-<owner/repo>} --merge --delete-branch
 
-Or invoke /approve-merge ${PR_NUMBER} — it auto-detects sync PRs and uses --merge.
+Or have the human approver run /approve-merge ${PR_NUMBER} — it auto-detects sync
+PRs and uses --merge. That skill is human-only (#1042); the model cannot invoke it.
 
 See AgDR-0053 for the full rationale.
 MSG
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -241,10 +241,12 @@ Missing file:
   ${REX_APPROVAL}
 
 To unblock:
-  1. Invoke the code-reviewer agent on this PR
+  1. Run the code review on this PR (/code-review ${PR_NUMBER}) — since #1042
+     this skill IS model-invocable, so you can do this yourself
   2. When Rex returns "approved", it records the approval automatically
-  3. Then run /approve-merge ${PR_NUMBER} for the ${APPROVER_TITLE} approval
-  4. Retry the merge
+  3. Tell the ${APPROVER_TITLE} the PR is ready and ask THEM to run
+     /approve-merge ${PR_NUMBER} — that skill is human-only, you cannot
+  4. Their invocation records the approval and performs the merge
 
 Never skip this check — even for typo fixes. See .claude/rules/pr-workflow.md.
 MSG
@@ -304,13 +306,16 @@ Missing file:
    title above — only the printed word changes, never the marker name.)
 
 To unblock:
-  1. Stop and ask the ${APPROVER_TITLE} explicitly: "PR #${PR_NUMBER} ready to merge — approved?"
-  2. When the ${APPROVER_TITLE} says "approved" / "merge it" / "ship it" naming PR #${PR_NUMBER},
-     invoke the /approve-merge skill:
+  1. Stop and tell the ${APPROVER_TITLE} the PR is ready, naming it:
+     "PR #${PR_NUMBER} is ready to merge."
+  2. Ask THEM to run the skill. Since #1042 it is human-only
+     (disable-model-invocation: true) — you cannot invoke it:
        /approve-merge ${PR_NUMBER}
-  3. The skill writes the structured marker AND runs the merge in one turn
+  3. Their invocation writes the structured marker AND runs the merge in
+     one turn. Your job ends at step 1.
 
-NEVER create this marker yourself from an umbrella "go" on a plan.
+NEVER create this marker yourself. You have no path to it, and that is the
+point: the invocation IS the approval, so it must come from a human.
 EVER. This is the exact failure this hook exists to prevent.
 MSG
     exit 2
--- a/.claude/hooks/require-design-review-for-ui.sh
+++ b/.claude/hooks/require-design-review-for-ui.sh
@@ -29,9 +29,18 @@
 # carve out a specific dir (e.g. `^docs/examples/`, `^wiki/artifacts/`) where
 # `.jsx`/`.tsx` files are documentation samples rather than real UI.
 #
-# How the marker gets written: the design-reviewer records approval by
-# writing the marker file. There is no /approve-design skill yet — the
-# design reviewer writes the file manually or via a (future) skill.
+# How the marker gets written: a HUMAN runs the `/approve-design <pr>` skill,
+# which writes the repo-qualified marker via `review_marker_path` (AgDR-0060).
+# Since #1042 that skill is `disable-model-invocation: true`, so the model
+# cannot invoke it — and unlike the architecture gate (where the spawned
+# solution-architect sub-agent writes its own marker), NO agent writes
+# `*-design.approved`. A human records this one, always.
+#
+# This comment previously said "there is no /approve-design skill yet" and the
+# unblock message below told the reader to hand-write the marker with a raw
+# redirect. Both were stale and, after #1042, actively harmful: the hand-write
+# became the ONLY path the message offered, on a marker type
+# `warn-review-marker-write.sh` does not guard.
 #
 # Trust model: same as other markers. Local session state, gitignored,
 # converts invisible inference ("ah, the UI change looked fine") into
@@ -250,11 +259,20 @@ The expected approval file does not exist:
 
 To unblock:
 
-  1. Invoke the UI Designer role (or a human designer) to review the UI changes
-  2. When the designer approves, record it with the current HEAD SHA:
-       mkdir -p .claude/session/reviews
-       git rev-parse HEAD > .claude/session/reviews/${PR_NUMBER}-design.approved
-  3. Retry the merge
+  1. Review the UI changes against the design system — adopt the UI Designer
+     role (Nour), or ask a human designer to look at the PR diff
+  2. Report the verdict plainly. Do NOT write the marker yourself: since
+     #1042 recording a design approval is a human action, and no agent
+     writes this marker type
+  3. Ask the designer or operator to run:
+       /approve-design ${PR_NUMBER}
+     That skill writes the repo-qualified marker against the PR's HEAD on
+     GitHub, which is what this gate compares
+  4. They retry the merge
+
+  Do not hand-write this file. A raw redirect produces the wrong path (the
+  marker is repo-qualified, see AgDR-0060) and usually the wrong SHA (the
+  gate reads the PR's HEAD from the forge, not your local HEAD — #55).
 
 To customize which file patterns count as "UI":
 
--- a/.claude/hooks/tests/test_skill_invocability_gates.sh
+++ b/.claude/hooks/tests/test_skill_invocability_gates.sh
@@ -39,9 +39,16 @@ FAIL=0
 
 # Read the frontmatter value. Only the frontmatter block counts, so the key
 # is matched at line-start and only before the closing `---`.
+#
+# Trailing \r is stripped first: a CRLF-committed SKILL.md would otherwise
+# yield "true\r", which compares unequal to "true" and fails this test for a
+# line-ending reason rather than a real one. This repo has been bitten by
+# exactly that before (#1019, where config_get left an embedded \r on Windows
+# checkouts), so the guard is cheap insurance rather than theory.
 invocation_flag() {
   local file="$1"
   awk '
+    { sub(/\r$/, "") }
     NR == 1 && $0 != "---" { exit }
     NR > 1 && $0 == "---"  { exit }
     /^disable-model-invocation:[[:space:]]*/ {
--- a/.claude/hooks/tests/test_skill_invocability_gates.sh
+++ b/.claude/hooks/tests/test_skill_invocability_gates.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+# me2resh/apexyard#1042 / AgDR-0110 — where the mechanical human gate sits.
+#
+# `disable-model-invocation` in a SKILL.md's frontmatter decides whether the
+# model may invoke that skill. Two families must sit on opposite sides:
+#
+#   REVIEW skills  (/code-review, /security-review, /design-review)
+#     Trigger an independent reviewer SUB-AGENT. Independence comes from that
+#     separate agent and its separate context — NOT from who typed the
+#     command. Locking these buys no safety and taxes every PR, while
+#     contradicting auto-code-review.sh, which explicitly instructs the
+#     orchestrator to run /code-review.
+#
+#   APPROVAL skills (/approve-merge, /approve-design, /approve-architecture)
+#     Write the approval markers that merge gates read. /approve-merge also
+#     performs the merge — irreversible and externally visible. Its own
+#     SKILL.md says "the discrete approval moment is the invocation".
+#     A human must make that invocation.
+#
+# Before #1042 these were exactly inverted: review skills were human-only
+# and approval skills were model-invocable (with /approve-architecture
+# carrying no frontmatter at all, so it defaulted to model-invocable).
+#
+# Why this test exists rather than trusting the frontmatter: the original
+# defect WAS a frontmatter value nobody was watching, sitting unnoticed from
+# the first commit. A silent edit here would reopen the gap invisibly — an
+# unlocked approval skill combined with unlocked review skills restores a
+# fully autonomous open -> review -> approve -> merge path.
+#
+# Exit 0 = all pass, exit 1 = at least one failure.
+
+set -u
+
+REPO_ROOT=$(cd "$(dirname "$0")/../../.." && pwd)
+SKILLS="$REPO_ROOT/.claude/skills"
+
+PASS=0
+FAIL=0
+
+# Read the frontmatter value. Only the frontmatter block counts, so the key
+# is matched at line-start and only before the closing `---`.
+invocation_flag() {
+  local file="$1"
+  awk '
+    NR == 1 && $0 != "---" { exit }
+    NR > 1 && $0 == "---"  { exit }
+    /^disable-model-invocation:[[:space:]]*/ {
+      sub(/^disable-model-invocation:[[:space:]]*/, "")
+      print
+      exit
+    }
+  ' "$file"
+}
+
+expect_flag() {
+  local skill="$1" want="$2" why="$3"
+  local file="$SKILLS/$skill/SKILL.md"
+
+  if [ ! -f "$file" ]; then
+    echo "FAIL: $skill — SKILL.md not found at $file"
+    FAIL=$((FAIL + 1))
+    return
+  fi
+
+  local got
+  got=$(invocation_flag "$file")
+
+  # An ABSENT key is not acceptable for either family: absent defaults to
+  # model-invocable, which is how /approve-architecture was silently
+  # unguarded. Both families must state their posture explicitly.
+  if [ -z "$got" ]; then
+    echo "FAIL: $skill — no 'disable-model-invocation' in frontmatter (absent defaults to model-invocable). $why"
+    FAIL=$((FAIL + 1))
+    return
+  fi
+
+  if [ "$got" = "$want" ]; then
+    echo "PASS: $skill → disable-model-invocation: $got"
+    PASS=$((PASS + 1))
+  else
+    echo "FAIL: $skill → expected 'disable-model-invocation: $want', got '$got'"
+    echo "   $why"
+    FAIL=$((FAIL + 1))
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# Review skills must be model-invocable (false).
+# ---------------------------------------------------------------------------
+
+for s in code-review security-review design-review; do
+  expect_flag "$s" "false" \
+    "Review skills spawn a separate reviewer sub-agent; independence does not depend on who invokes them. auto-code-review.sh instructs the orchestrator to run /code-review, so locking it makes the framework contradict itself."
+done
+
+# ---------------------------------------------------------------------------
+# Approval skills must be human-only (true).
+# ---------------------------------------------------------------------------
+
+for s in approve-merge approve-design approve-architecture; do
+  expect_flag "$s" "true" \
+    "Approval skills write the markers merge gates read; /approve-merge also merges. The invocation IS the approval, so a human must make it (see .claude/rules/pr-workflow.md and AgDR-0110)."
+done
+
+# ---------------------------------------------------------------------------
+# The combination is what matters: if review skills are unlocked, approval
+# skills MUST be locked, or the model can drive the whole chain unattended.
+# Asserted explicitly so the coupling is visible to whoever edits either side.
+# ---------------------------------------------------------------------------
+
+REVIEW_UNLOCKED=1
+for s in code-review security-review design-review; do
+  [ "$(invocation_flag "$SKILLS/$s/SKILL.md")" = "false" ] || REVIEW_UNLOCKED=0
+done
+
+APPROVAL_LOCKED=1
+for s in approve-merge approve-design approve-architecture; do
+  [ "$(invocation_flag "$SKILLS/$s/SKILL.md")" = "true" ] || APPROVAL_LOCKED=0
+done
+
+if [ "$REVIEW_UNLOCKED" = "1" ] && [ "$APPROVAL_LOCKED" = "0" ]; then
+  echo "FAIL: review skills are model-invocable while approval skills are NOT locked."
+  echo "   That combination restores an autonomous open -> review -> approve -> merge"
+  echo "   path with no human in the loop. Lock the approval skills."
+  FAIL=$((FAIL + 1))
+else
+  echo "PASS: no autonomous open -> review -> approve -> merge path"
+  PASS=$((PASS + 1))
+fi
+
+# ---------------------------------------------------------------------------
+# Result
+# ---------------------------------------------------------------------------
+
+echo
+echo "Passed: $PASS  Failed: $FAIL"
+if [ "$FAIL" -gt 0 ]; then
+  exit 1
+fi
+exit 0
--- a/.claude/hooks/warn-review-marker-write.sh
+++ b/.claude/hooks/warn-review-marker-write.sh
@@ -541,14 +541,19 @@ You are about to write a *-ceo.approved review marker (filename stays
   fabricated casually.
 
   Who may write this marker:
-    /approve-merge skill, invoked by the orchestrator on an explicit ${APPROVER_TITLE} nod
+    the /approve-merge skill, which since #1042 only a HUMAN can invoke
+    (disable-model-invocation: true)
 
 WHY THIS MATTERS
   Writing this file yourself satisfies the merge gate's FILENAME check
   but NOT its INTENT. block-unreviewed-merge.sh independently validates
   the structured fields before any merge is allowed, so a hand-written
-  marker without those fields is still rejected at merge time — but
-  don't write this file yourself. Invoke /approve-merge <pr>.
+  marker without those fields is still rejected at merge time.
+
+  You have no path to this marker, and that is deliberate: the invocation
+  IS the approval, so it must come from a human. Tell the ${APPROVER_TITLE}
+  the PR is ready and ask them to run /approve-merge <pr>. Do not try to
+  produce the marker some other way.
 
 ======================================================================
 BANNER
--- a/.claude/rules/pr-workflow.md
+++ b/.claude/rules/pr-workflow.md
@@ -152,6 +152,14 @@ Both markers' SHAs must match the PR's HEAD as reported by GitHub (`gh pr view <
 
 Claude can technically `rm` or `touch` these files by hand, or fabricate the structured fields. Doing so is a visible, auditable, grep-able rule violation — and the whole point of recording the rule mechanically is so that the failure mode is "Claude ignored a hook" (visible) instead of "Claude inferred approval from something vague" (invisible). The structured-marker format raises the visibility bar one more notch by requiring the model to type `approved_by=user` etc. on purpose.
 
+### The approval skills are human-only (#1042, AgDR-0110)
+
+"Explicit per-PR approval" is now enforced **mechanically**, not only in prose. `/approve-merge`, `/approve-design`, and `/approve-architecture` carry `disable-model-invocation: true`, so **only a human can invoke them**. Saying "approved" in conversation is no longer sufficient on its own — the operator types the command, and that invocation *is* the approval moment this rule has always described.
+
+The mirror half matters just as much: the **review** skills (`/code-review`, `/security-review`, `/design-review`) are model-invocable, so the orchestrator triggers them itself — as `auto-code-review.sh` has always instructed. Independence comes from the reviewer being a **separate sub-agent with its own context**, not from who typed the command, so nothing is lost by letting the model start a review.
+
+Before #1042 these were exactly inverted, and the safety was accidental: a model couldn't invoke `/code-review`, so it couldn't obtain a Rex marker, so it couldn't merge. Unlocking the review skills **alone** would therefore have opened a fully autonomous `open → review → approve → merge` path. The two sides are coupled — `test_skill_invocability_gates.sh` pins both, and fails loudly if review skills are ever unlocked while approval skills are not locked.
+
 ### Both merge shapes are gated (#47)
 
 All three merge-gate hooks (`block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`) fire on **both** the `gh` subcommand shape and the raw REST-API shape:
--- a/.claude/rules/pr-workflow.md
+++ b/.claude/rules/pr-workflow.md
@@ -82,16 +82,16 @@ You: *runs gh pr merge 10*   ← FAILURE: "go" was plan-level, not merge-level.
 You: "Here's the 6-step plan: 1. merge PR #10, 2. close PR #105, ..."
 CEO: "go"
 You: *executes steps 2–6, stops before step 1*
-You: "Steps 2–6 done. Ready to merge PR #10 — approved?"
-CEO: "approved"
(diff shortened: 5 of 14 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1043

**Commit**: `6331adf0962685ef149ecbdfb641755296c2b2dc`

### Summary

Swaps the mechanical human gate from the review skills onto the approval skills: `/code-review`, `/security-review`, `/design-review` become model-invocable; `/approve-merge`, `/approve-design`, `/approve-architecture` become human-only. `/approve-architecture` gains frontmatter it never had. Adds `test_skill_invocability_gates.sh` pinning the invariant, a `pr-workflow.md` section, and `AgDR-0110`.

The direction is right and the change is a net safety improvement. One blocking doc inconsistency, plus two findings about the reasoning that are worth recording.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass**
- Security: **Pass** (see interrogation below)
- Performance: **N/A**
- Technical Decisions (AgDR): **Pass** — AgDR-0110, number free (highest existing is 0109; the known `AgDR-0095` duplication is pre-existing and untouched)
- Adopter Handbooks: **N/A** — `migration-safety.md` (blocking) does not fire, no migration paths in the diff; `clean-architecture-layers.md` and `commit-message-quality.md` have no applicable triggers
- Fallow Static Analysis: **N/A** — no JS/TS in the diff

### What I verified

**1. The swap is complete and the YAML is valid.** All six frontmatter blocks parse cleanly under `yaml.safe_load`, including the newly-added `approve-architecture` block:

| Skill | `disable-model-invocation` |
|---|---|
| code-review / security-review / design-review | `false` |
| approve-merge / approve-design / approve-architecture | `true` |

**2. The test genuinely pins the invariant.** Run against this branch: 7/7 pass. Run against `dev`: **6 of 7 fail**, including the absent-flag case (`approve-architecture` on `dev` correctly fails with *"no 'disable-model-invocation' in frontmatter (absent defaults to model-invocable)"*). I also built two synthetic trees to confirm the combination check does real work rather than riding along:

- reviews unlocked + all three approvals unlocked → combination check **FAILS** with the autonomous-path explanation
- reviews unlocked + `approve-architecture`'s flag deleted entirely → combination check **FAILS** (absent flag correctly propagates to `APPROVAL_LOCKED=0`)

So the dangerous pairing is caught both when stated explicitly and when it arrives via an omitted key — which is the shape the original defect actually took. `shellcheck -S warning` is clean, and `bin/run-hook-tests.sh` auto-discovers the file, so CI picks it up with no wiring change (the "hook test suite" check is green on this PR).

**3. Remaining unattended-merge paths.** With reviews unlocked and approvals locked, the *sanctioned* path is fully human-gated. The residual paths are raw-Bash forgery, and none of them are opened or widened by this PR:

- The `*-ceo.approved` marker is **advisory-only** in `warn-review-marker-write.sh` (line 94: *"KEEPS its original #728 advisory-only behaviour"*). A model can still hand-write the structured fields via Bash.
- Writes to `.claude/session/active-reviewer` are not themselves gated, so the provenance check on the rex marker is in principle self-serviceable.

Both are pre-existing and independent of skill invocability, and both are already the acknowledged position of AgDR-0104 / AgDR-0109 — the latter states outright that *"a determined actor can still reach the file locally because the real control is the per-PR human approval."* This PR is precisely the change that makes that "real control" mechanical. Nothing here regresses; the notes below are about how the AgDR words it.

### Issues Found

**⛔ BLOCKING — `pr-workflow.md:87` now teaches an action the model can no longer perform.**

The PR adds a new section at line 155 describing the change, but leaves the canonical worked example 70 lines above it untouched:

```
CEO: "approved"
You: *invokes /approve-merge 10*   ← CORRECT: writes structured marker AND merges in one turn.
```

Under the shipped behaviour `/approve-merge` is `disable-model-invocation: true`, so "You" cannot invoke it. The most-read example in the framework's canonical merge rule now depicts an impossible action labelled **CORRECT** — which is the exact defect class this PR exists to remove (`auto-code-review.sh` instructing an action the flag forbade). A reader landing on "Plan-level 'go' is NOT merge approval" — the section people actually quote — never reaches the new text explaining otherwise.

This is also the direct answer to *"is the operator cost stated clearly enough for someone who did not read this PR?"*: the new section states it well, but the older example actively contradicts it, and the contradiction sits earlier in the file.

Suggested fix, in the file this PR already modifies:

```
CEO: "approved"
CEO: *types /approve-merge 10*   ← CORRECT: only a human can invoke it (#1042);
                                    writes the structured marker AND merges in one turn.
```

The `#### Why` paragraph below it reads correctly either way and needs no change.

### Suggestions

**suggestion: the coupling argument understates the prior risk — and that strengthens the case for this PR.**

The PR and AgDR both rest on: *"because a model could not invoke `/code-review`, it could not obtain a Rex marker, so it could not merge."* That is not quite true. `.claude/agents/code-reviewer.md:626` documents the alternative explicitly:

> the orchestrator (**or** the `/code-review` skill) sets the `.claude/session/active-reviewer` provenance marker (#843) before spawning you

So the orchestrator spawning Rex through the Agent tool — bypassing the skill entirely — was always a sanctioned way to obtain a valid rex marker. That is how *this very review* was invoked. Combined with a pre-PR model-invocable `/approve-merge`, the fully autonomous `open → review → approve → merge` path was **already reachable**, not merely latent.

The decision is unaffected — "swap both" remains correct, and "unlock reviews only" remains the worst option. But the framing treats the status quo as safer than it was, which under-sells the fix. Worth a sentence in AgDR-0110's Context so a future reader doesn't infer that the skill flag was ever the load-bearing barrier.

**nit: "structurally impossible" overstates it.** AgDR-0110's Consequences says *"Autonomous merge becomes structurally impossible rather than incidentally blocked."* Given the advisory CEO marker and the ungated `active-reviewer` write, the accurate claim is that autonomous merge is impossible **on the sanctioned path**, with forgery remaining a visible, auditable rule violation. AgDR-0110 already carries the right caveat in "What this does NOT change" (the AgDR-0104 cross-reference), so this is one word away from consistent — but a trust-chain AgDR asserting structural impossibility is the kind of line that gets quoted later without its caveat.

**nit: `invocation_flag` is sensitive to trailing `\r`.** A CRLF-committed SKILL.md yields `false\r`, failing the equality check with a confusing message. This repo has already been bitten by exactly this (#1019, `config-get` stripping CR). A `sub(/\r$/, "")` in the awk block would make it robust for a Windows-based adopter.

### Verdict

**CHANGES REQUESTED**

The mechanism, the test, and the AgDR are all sound — I'd approve on the one-line `pr-workflow.md:87` fix alone. The suggestions are non-blocking; the AgDR Context note is worth taking while you're in there.

---
📌 Reviewed commit: `6331adf0962685ef149ecbdfb641755296c2b2dc`

atlas-apex (commented):
## Code Review: PR #1043 (re-review)

**Commit**: `08069bb30d800580dc30e54b393b25a3b4825237`

### Summary

Re-review after the fixes at `08069bb`. All four items I raised at `6331adf` were addressed, and I verified each one rather than taking the changelog at its word. The production change is untouched and still correct. However, the blocking finding I raised was **narrower than the defect** — I named the instance in `pr-workflow.md`, the author fixed that instance and the sibling one in `approve-merge/SKILL.md`, but the same self-contradiction survives verbatim in **two other files this PR itself modifies**. That is blocking, on the same standard I applied last round.

### Verification of the four fixes

| Item | Status | Evidence |
|---|---|---|
| `pr-workflow.md` "Right" example | **Fixed** | `pr-workflow.md:85-87` now shows `CEO: /approve-merge 10` with an inline "the model cannot" note; the "Why" paragraph at :94 names `disable-model-invocation: true` and #1042 |
| `approve-merge/SKILL.md` anti-pattern | **Fixed** | :341 reframes it as failing "twice over"; the correct-flow block at :348-350 shows the CEO invoking |
| AgDR-0110 corrections | **Fixed and accurate** | verified below |
| `invocation_flag` CRLF guard | **Fixed and load-bearing** | tested below |

**CRLF guard — tested, not read.** I converted `approve-merge/SKILL.md` to CRLF in place and re-ran the suite: **7/7 pass**. I then ran the *pre-fix* awk against a CRLF fixture: it emits **nothing at all** (line 1 is `---\r`, which fails the `NR==1 && $0 != "---"` guard and exits immediately), so the test would have reported "no `disable-model-invocation` in frontmatter" — a false failure with a misleading message, exactly the #1019 shape. The new awk yields a clean `true\n` (`od -c` confirms no trailing `\r`). The normal LF path is unaffected — 7/7 on the unmodified tree. Working tree restored; no stray diff.

**AgDR-0110's `code-reviewer.md` claim is true.** `.claude/agents/code-reviewer.md:626` reads: *"the orchestrator (or the `/code-review` skill) sets the `.claude/session/active-reviewer` provenance marker (#843) before spawning you"*. The parenthetical is the point — the orchestrator setting the marker and spawning the sub-agent directly is a documented, supported path that never touches `/code-review`. The new Context paragraph states this correctly, and its conclusion (this *strengthens* the case, because it replaces a bypassable incidental barrier with a mechanical one) is sound reasoning, not spin.

**The rescoped consequence is consistent with AgDR-0104/0109.** AgDR-0109 §"Options" rejects structuring the Rex marker because "a determined build agent can still type out the fields," and its Consequences accept the residual exposure explicitly; AgDR-0104 states a hook pattern-matching bash text "cannot be made sound" and relocates real trust to the forge's server-side gate. AgDR-0110 now claiming only that *the skill-invocation path* is closed by design — and naming raw-Bash marker forgery as still available, citing both prior AgDRs — lines up exactly. The earlier "structurally impossible" would have contradicted two ratified decisions in the same trust-chain series.

**Production change unchanged.** All six frontmatter flags are identical to what I approved at `6331adf` (`gh pr diff` confirms: three review skills `true→false`, `approve-design`/`approve-merge` `false→true`, `approve-architecture` gains a full frontmatter block with `true`). Suite is **7/7 on the branch** and **6/7 FAIL against `dev`** — I reconstructed `dev`'s `.claude/` into a temp tree and ran the new test against it to confirm the discrimination is real, including that the combination assertion correctly *passes* on `dev` (review skills locked there, so the dangerous pairing doesn't hold — the asymmetry the PR body already calls out).

**Doc structure.** Code fences balanced in all five touched/inspected markdown files; the reflowed `pr-workflow.md` block renders cleanly (the two-line continuation with the `←` marker stays inside the fence and aligns); no broken internal links introduced.

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass — trust-chain change, Heavy tier; net-tightening, and the residual exposure is now honestly named
- Performance: N/A
- Technical Decisions (AgDR): Pass — AgDR-0110 linked and now accurate
- Adopter Handbooks: N/A — none loaded

### Issues Found

**⛔ BLOCKING — the invalidated "CORRECT" example survives in two more files, both of which this PR modifies.**

My last review named the `pr-workflow.md` instance. The defect class is "a worked example labelled ← CORRECT showing the model invoking a skill this PR makes human-only." Two more instances survive, in the *primary anti-pattern sections of the very skills being locked*:

- `.claude/skills/approve-design/SKILL.md:160` — `You: *invokes /approve-design 42*  ← CORRECT`
- `.claude/skills/approve-architecture/SKILL.md:157` — `You: *invokes /approve-architecture 42*  ← CORRECT`

Both files are in this PR's diff: `approve-design`'s frontmatter flips `false → true` at line 4, and `approve-architecture` gains `disable-model-invocation: true` in a frontmatter block this PR creates. So each file now contains, roughly 150 lines apart, a machine-readable flag forbidding model invocation and a prose example telling the model that invoking it is the correct behaviour. That is the same self-contradiction the PR exists to remove, and it is worse-placed than the one already fixed: `pr-workflow.md` is a rule doc read occasionally, whereas a SKILL.md's anti-pattern block is the canonical worked example for that specific skill.

Suggested shape, mirroring the fix already applied to `approve-merge`:

```
Designer: "Reviewed PR #42, implementation matches the mockup. Design approved."
Designer: /approve-design 42     ← CORRECT: the designer invokes it. The skill is
                                    human-only (#1042), so the model cannot.
```

While you are in `approve-architecture/SKILL.md`, lines 156-157 carry a second, pre-existing muddle worth fixing in the same pass: the flow reads *"You: \*Tariq writes the marker automatically\* — OR a human architect says approved"* and then *"You: \*invokes /approve-architecture 42\*"* — if Tariq already wrote the marker, the invocation is redundant. Splitting it into two distinct branches (Tariq's `/design-review` writes the marker, **or** the human architect invokes `/approve-architecture`) would make both the pre-existing logic and the new human-only constraint read correctly.

### Suggestions

- **nit (non-blocking, out of scope):** `.claude/hooks/block-unreviewed-merge.sh:192` prints *"Or invoke /approve-merge ${PR_NUMBER} — it auto-detects sync PRs and uses --merge."* into the model's context on a blocked sync-PR merge. Post-#1042 the model cannot comply. The practical effect is benign (the model relays the suggestion to the operator, which is the desired outcome), and the hook is outside this PR's scope — but if you are sweeping for stale invocation prose anyway, rewording it to *"or ask the operator to run /approve-merge N"* costs one line. I would not hold the PR for it.
- **nit:** the PR body's parenthetical about `enforce-budget.sh` not being tracked in git is a genuine finding and correctly routed to a separate ticket rather than scope-creeping this one. Good call; just make sure the ticket actually gets filed.

### Verdict

**CHANGES REQUESTED**

Everything I asked for last round was done, and done well — the CRLF guard is real, the AgDR corrections are honest and check out against AgDR-0104/0109, and the production change is untouched and correctly tested. The single blocker is that my previous finding was under-scoped: fix the two surviving `← CORRECT` examples in `approve-design/SKILL.md:160` and `approve-architecture/SKILL.md:157` and this is ready. No re-verification of the production change will be needed on the next pass — a docs-only delta.

---
📌 Reviewed commit: `08069bb30d800580dc30e54b393b25a3b4825237`

atlas-apex (commented):
## Code Review: PR #1043 (round 3)

**Commit**: `f0555c0a6e464cba8d8268d79fe7e0d976c65993`

### Summary

Docs-only delta over `08069bb`. The two worked examples I flagged in round 2 (`approve-design/SKILL.md`, `approve-architecture/SKILL.md`) are both fixed and now show a human invoking; all three anti-patterns carry the "wrong twice over" annotation; the `approve-architecture` redundancy is untangled. The production change is untouched and correct.

I ran my own repo-wide sweep rather than trusting the author's, since my previous two findings were both under-scoped. **The sweep found five more surviving instances in three files.** Two of them are model-facing runtime hook messages, and one of those is in a file this PR edits.

### Checklist Results
- Code Quality: **Pass**
- Testing: **Pass** — `test_skill_invocability_gates.sh` 7/7 on branch; case 7 discriminates (fails against `dev`)
- Security: **Pass**
- Performance: N/A
- Technical Decisions (AgDR): **Pass** — AgDR-0110 present
- Doc/behaviour consistency: **FAIL** — see below

### Verified good

1. **Production change intact.** All six frontmatter flags correct: `code-review` / `security-review` / `design-review` = `false`, `approve-merge` / `approve-design` / `approve-architecture` = `true`.
2. **The `approve-architecture` restructure is correct** — I checked the claim rather than taking it. `.claude/agents/solution-architect.md:236` writes the marker via `review_marker_path "$PR_HOST_REPO" {number} architecture`, and `require-architecture-review.sh:231` looks up that exact path. `/design-review` is model-invocable, and `warn-review-marker-write.sh` authorizes Tariq's write off the active-reviewer marker. So "Tariq writes the marker himself → DONE, no `/approve-architecture` needed" is **true**, and the skill genuinely is only the operator path. Good call splitting these.
3. CI green on `f0555c0` (6/6, CodeQL skipped). Fences balanced in all edited markdown; no broken lists.

### Issues Found — blocking

**1. `block-unreviewed-merge.sh:307-311` — model-facing unblock instruction is now impossible to follow. In this PR's own diff.**

```
  2. When the ${APPROVER_TITLE} says "approved" ... naming PR #${PR_NUMBER},
     invoke the /approve-merge skill:
       /approve-merge ${PR_NUMBER}
  3. The skill writes the structured marker AND runs the merge in one turn
```

Steps 1–2 are unambiguously addressed to the model ("Stop and ask the CEO explicitly"). Step 2 then instructs the model to invoke a skill that, as of this PR, it mechanically cannot. This PR updates the sync-PR message at line 192 in this same file to say *"That skill is human-only (#1042); the model cannot invoke it"* — but leaves this one and line 246 stale. Same under-scoping shape as rounds 1 and 2, now inside the file being edited.

This is the highest-impact instance because it is a **runtime blocking message**, not documentation: the model hits the gate, reads the instruction, and the compliance path does not exist. `docs/spike-reports/learn-skill-feasibility.md:45` records what happens at that dead end — 102 marker-fabrication attempts vs 75 legitimate invocations, one session hitting the same wall 19 times. Leaving an unfollowable instruction here plausibly *increases* the fabrication pressure this subsystem exists to suppress.

Suggested: `2. Ask the ${APPROVER_TITLE} to run /approve-merge ${PR_NUMBER} themselves — the skill is human-only (#1042); you cannot invoke it.`

**2. `block-unreviewed-merge.sh:246`** — `3. Then run /approve-merge ${PR_NUMBER} for the ${APPROVER_TITLE} approval`, in a numbered to-unblock list whose steps 1–2 are model actions. Same fix.

**3. `warn-review-marker-write.sh:544` and `:551`** — model-facing banner ("You are about to write..."):

- `:544` — `Who may write this marker: /approve-merge skill, invoked by the orchestrator on an explicit ${APPROVER_TITLE} nod`. The orchestrator *is* the model. False since #1042.
- `:551` — `don't write this file yourself. Invoke /approve-merge <pr>.` Direct imperative to the model to invoke a human-only skill.

Not in this PR's diff, but this PR is what makes both false, and it is the banner the model reads at the exact moment it is tempted to fabricate the marker. Same argument as #1.

**4. `roles/design/ui-designer.md:41` and `:74`** — the UI Designer is **in-flow class** (per `role-triggers.md`, the banner instructs the agent to *adopt the persona IN-THREAD*), so these lines address the model:

- `:41` — "Approve the routine per-PR design gate for UI implementation PRs (via `/approve-design`)"
- `:74` — "you review the implementation against the design system and, on approval, record it with the **`/approve-design <pr>`** skill"

I checked whether an agent could write the design marker instead, the way Tariq does for architecture: **no agent writes `*-design.approved`** — `/approve-design` is the only producer. So post-#1042 the routine design gate has no in-thread path at all, and the role file still tells the persona to walk one. This needs the same "hand back to the human" restructure you just applied to `approve-architecture`.

**5. `roles/design/head-of-design.md:70`** — "can record the design marker with `/approve-design <pr>` when an escalation lands on your desk." Same issue, escalation path.

- `nit:` `.claude/skills/release-sync/SKILL.md:226` — `Next step: /code-review, then /approve-merge once Rex approves.` Actor is unstated; now that the two halves have different invocability, worth naming who does which.
- `nit:` `.claude/skills/update/SKILL.md:1039` — "User runs Rex + `/approve-merge` themselves" is correct for `/approve-merge` but now understates that Rex *is* model-invocable.

### Sweep method (so this round is the last one)

`grep -rn -- "/approve-" --include="*.md" --include="*.sh" --include="*.json" .` → 217 hits → filtered for an actor verb (`invok|run|call|execut|then |now `). Worth re-running after the fix. The pattern all five share: an **imperative or descriptive verb attached to the model or an in-thread persona**, which the literal string `invokes /approve-` does not catch — which is why the previous sweeps (mine included) kept missing them.

### Verdict
**CHANGES REQUESTED**

The production change and the AgDR are right, and I'd approve those as-is. What's blocking is that the doc/message sweep still isn't complete — and items 1–3 are runtime text that actively instructs the model into a dead end, which is a worse failure than a stale example in a SKILL.md.

---
📌 Reviewed commit: `f0555c0a6e464cba8d8268d79fe7e0d976c65993`

atlas-apex (commented):
## Code Review: PR #1043

**Commit**: `43ccb8c4e0e73437020f299436d3ce8fb8832ee9`

### Summary

Swaps `disable-model-invocation` across two skill families — review skills become model-invocable, approval skills become human-only — so the mechanical human gate sits on the irreversible step rather than the harmless one. Round 4. The production change is correct and has been stable across all four rounds; this round's blocking finding is again the same class, and I can now explain why three sweeps missed it.

### Checklist Results
- Security:                   **Fail** — see Blocking
- Technical Decisions (AgDR): Pass — AgDR-0110
- Fallow (JS/TS):             N/A

### Verified good

**Production change untouched and correct.** All six flags as specified; `test_skill_invocability_gates.sh` passes 7/7. The test is genuinely discriminating: it rejects an *absent* key (the `/approve-architecture` failure mode) as well as a wrong value, and the seventh assertion pins the *coupling* rather than the individual values, so unlocking review skills while approval skills drift open fails loudly.

**Both rewritten runtime messages are correct and mutually coherent.** `block-unreviewed-merge.sh` (Rex block, CEO block, sync-PR block) and `warn-review-marker-write.sh` now agree: run `/code-review` yourself, hand `/approve-merge` to the human. Each leaves the model a concrete next action — the Rex block says "you can do this yourself", the CEO block says "your job ends at step 1". No contradiction between them.

**Design-gate claim independently verified — and item 3 is already fine.** No agent writes the design approval marker: all nine build-agent files forbid writes under the session reviews directory, and `.claude/agents/ui-designer.md:23` forbids it explicitly. The agent file needed no update. `solution-architect.md` writing the architecture marker is the intended exception and is gated by the active-reviewer marker.

**Reverse class clean.** Swept for stale "review skills are human-only / operator must hand-type `/code-review`" claims across hooks, skills, rules, roles, agents, docs, workflows, CLAUDE.md, AGENTS.md, README. Only the three frontmatter lines and AgDR-0110's own historical narration — all correct.

---

### Blocking — same class, fourth occurrence: `.claude/hooks/require-design-review-for-ui.sh`

**Why three sweeps missed it: this one never mentions `/approve-design` at all.** Rounds 1–3 (mine included) searched for the `/approve-*` token, then for verbs bound to it. This instance is a *raw marker write* — the imperative names no skill, so no `/approve-*` pattern can reach it. I only found it by sweeping the generalized shape (a shell redirect into a design-marker path). That is the correct final scope, and it is now swept.

**1. The runtime unblock message instructs the model to perform the approval itself** — `require-design-review-for-ui.sh:250-256`. Its `To unblock` steps say: invoke the UI Designer role to review the changes, then "when the designer approves, record it with the current HEAD SHA" — followed by a literal shell redirect writing the design marker by hand.

This is the surface that matters most: what the model reads at the moment of temptation, blocked from merging. It tells the model to hand-write the design approval marker, which is exactly the action this PR just made human-only. It contradicts the PR's own new text at `roles/design/ui-designer.md:76` ("an in-thread persona cannot invoke it") and re-creates the defect AgDR-0110 § Context names as the *reason* for the change: the framework instructing an action it forbids.

**2. That path is also stale** — not repo-qualified per AgDR-0060 (the marker basename should carry the `owner__repo__pr` prefix). A model following the instruction writes a marker the gate will not read, stays blocked, and is pushed toward further improvisation.

**3. `require-design-review-for-ui.sh:32-34` comment is factually wrong**: *"There is no /approve-design skill yet — the design reviewer writes the file manually or via a (future) skill."* The skill exists and is now human-only. This is the comment that documents the marker-write model to future maintainers.

**Why this is blocking rather than a pre-existing nit.** The file is untouched by this PR, but this PR changes the message's safety semantics. Before #1042 a model reading it had a sanctioned alternative — `/approve-design` was model-invocable. This PR removes that alternative and leaves the raw-write instruction as the only path the message offers. Round 3's own finding was that the routine design gate has no in-thread path; that resolution was written into the *role* files while the *gate that implements it* still says the opposite.

Aggravating: `warn-review-marker-write.sh`'s `_is_marker_target` matches only `(rex|ceo|security|architecture)` — `design` is absent, so the hook exits 0 on any design-marker write. No block, no advisory banner. Post-merge the design gate would be: sanctioned path human-only, unsanctioned raw-write path unguarded *and affirmatively instructed by the gate's own message*.

**Fix (small, bounded, and the last of this class):**
- Rewrite the `To unblock` block to mirror the two messages this PR already fixed: Nour reviews and reports, a human runs `/approve-design <pr>`, the model cannot invoke it.
- Drop the hand-write redirect line, or repo-qualify it and mark it explicitly as the human/adopter fallback.
- Correct the `:32-34` comment.
- Consider adding `design` to `_is_marker_target` so the marker family is uniformly guarded (reasonable as a follow-up ticket if you would rather keep this PR's diff tight).

---

- `block-unreviewed-merge.sh:317` — the rewrite inserted two sentences between "NEVER create this marker yourself." and the standalone `EVER.`, which now dangles. Fold it in ("NEVER create this marker yourself. Not ever.").
- `docs/rule-audit.md:72` — still lists `/approve-design` as *deferred*, with a manual marker write as "today's convention". Stale; same root cause as the hook comment.
- `.claude/hooks/README.md:198` — the `touch` manual fallback is repo-qualified and addressed to adopters, so it is acceptable; a "(human-run)" clarifier would make it consistent with the new posture.
- Meta, and mildly amusing: posting this review initially tripped `warn-review-marker-write.sh`, because the review *prose* quoted marker paths and the hook inspects raw command text. Live confirmation of the AgDR-0104 limitation that AgDR-0110 § "What this does NOT change" already points at.

### Verdict

**CHANGES REQUESTED**

The swap itself is right and I have no remaining concern about it. What is left is one file: the design gate's own runtime message still instructs the model to do the thing this PR made human-only. Fix `require-design-review-for-ui.sh` and this is done — the raw-marker-write shape is now swept, so there is no further scope of this class outstanding.

---
Reviewed commit: `43ccb8c4e0e73437020f299436d3ce8fb8832ee9`

atlas-apex (commented):
## Code Review: PR #1043 — round 5

**Commit**: `3a682a55ca4f609359dcc9f104682d004ebe025e`

### Summary

Targeted round on the single new commit `3a682a5`, which touches one file — `.claude/hooks/require-design-review-for-ui.sh` — in response to my round-4 blocker. I re-verified every factual claim the new message makes against the actual code rather than the commit description, and re-swept the raw hand-write class by *shape* rather than by skill name (the search that failed twice before).

**My round-4 blocker is fixed.** The "To unblock" message no longer offers a hand-write as a reachable instruction. It now routes to review, report the verdict, and ask a human to run `/approve-design <pr>`, with an explicit do-not-hand-write paragraph.

### Checklist Results
- Testing:                    Pass (`test_skill_invocability_gates.sh` — 7 passed, 0 failed at this HEAD)
- Performance:                N/A (message text only)
- Technical Decisions (AgDR): Pass — AgDR-0110 present and linked

### Verification of the new message's factual claims

Checked each against source, not against the commit message:

| Claim | Verdict | Evidence |
|---|---|---|
| The design marker is unguarded by `warn-review-marker-write.sh` | **Correct** | Its target-matching predicate covers only the rex / ceo / security / architecture roles — the design role is genuinely absent |
| The message does not *overclaim* that the write is blocked | **Correct** | Step 2 makes a normative claim ("no agent writes this marker type"), not a mechanical one; the header comment states outright that the hook "does not guard" it |
| `/approve-design` is `disable-model-invocation: true` | **Correct** | Confirmed in frontmatter on this branch |
| Marker path is repo-qualified per AgDR-0060 | **Correct** | Gate resolves it through the shared marker-path helper at line 244; the helper emits the `owner__repo__pr` qualified form |
| "#55 — the gate reads the PR's HEAD from the forge, not local HEAD", **for this gate specifically** | **Correct** | Line 299 is `resolve_pr_head "$PR_NUMBER" "$CMD_REPO"`. The `git rev-parse HEAD` at line 302 is only the documented fallback when the `gh` call fails, and it emits a WARN first. The claim is accurate as written. |
| "unlike the architecture gate, NO agent writes the design marker" | **Correct** | `solution-architect.md:235` does write the architecture marker; no agent file writes a design one |

### Class sweep (by shape, not by skill name)

Searched for redirects into `*.approved`, `touch`-ing an approved marker, and directory creation under the session review path, across `.claude/`, `docs/`, `roles/`, `workflows/`. Every remaining marker-write instruction lives in a **sanctioned writer**: the `code-reviewer` and `solution-architect` agent files, and the three `/approve-*` skill docs — all of which must contain the write because they *are* the sanctioned path. No unsanctioned hand-write instruction remains.

### Suggestions (advisory — none blocking)

**1. The documented skip escape hatch now sits 12 lines under "Do not hand-write this file".**

The same heredoc still ends with (lines 287-290):

> For projects that deliberately ship UI without design review (e.g. admin tools, internal dashboards), touch the marker file manually ...

I am **not** blocking on this, and want to be precise about why. It is pre-existing (introduced in `829fa46`, the hook's original commit — untouched by this PR's diff), it is deliberately documented at `.claude/hooks/README.md:198`, and it is semantically a *different act*: `touch` produces an empty file, which leaves the approved-SHA variable empty and so skips the SHA comparison at line 304 entirely. That makes it a visible "we skip design review as project policy" artifact, not a forged approval-at-a-SHA. My round-4 blocker was that the hand-write was the *only* path offered; that is genuinely resolved.

The residual nit is that the rewritten message does not *draw* that distinction, so the two paragraphs can read as contradictory to a model under merge pressure. Worth a follow-up that frames the escape hatch explicitly as an operator policy decision, in the README's own language. Not this PR.

**2. Pre-existing inconsistency — `/approve-design` writes local HEAD, this gate compares forge HEAD.**

`.claude/skills/approve-design/SKILL.md:124` writes the marker using `git rev-parse HEAD`, while this gate compares against `resolve_pr_head` (the PR's HEAD on GitHub). When local HEAD and PR HEAD diverge, the marker the sanctioned skill writes will not match what the gate reads. **This is pre-existing and not introduced by this PR** — flagging it as advisory only, exactly as scoped. The same shape was fixed for the merge gates in #55; `/approve-design` looks like it was missed. Worth its own ticket.

**3. Residual gap, already self-declared.** The commit message honestly notes the design marker remains outside `warn-review-marker-write.sh`'s guarded set — the hand-write is now un-instructed but still un-blocked. Correct assessment and the right call to keep out of scope. A follow-up adding the design role to that predicate would close it, with the caveat that the escape hatch above would then need an explicit carve-out.

**4. Incidental observation from this very review.** `warn-review-marker-write.sh` blocked my first attempt to write this review body to a scratchpad file, purely because the body's *prose* quoted a marker suffix and a session path; its indirect-detection heuristic then misread a "line 244" citation as PR 244. Pre-existing behaviour, unrelated to this PR, and consistent with the intent-from-shell-text limitation AgDR-0110 already calls out (AgDR-0104, #1026/#1032/#1038/#1039). Noting it only as a data point for that thread.

### Verdict

**APPROVED**

The round-4 blocker is fixed, and every factual claim in the new message is accurate against the real code — including the one I was asked not to take on faith: this gate does resolve the PR HEAD via the forge, with local HEAD only as a warned fallback. The class sweep is clean, the coupling test suite passes at this HEAD (7/7), and the Glossary and the AgDR-0110 link both survive. The advisory items above are pre-existing or explicitly out of scope and should not hold this PR.

---
Reviewed commit: `3a682a55ca4f609359dcc9f104682d004ebe025e`

Issue #1044, opened 2026-07-27
[Bug] right-size-ceremony cites enforce-budget.sh as its watchdog, but that hook is premium-only and not tracked


atlas-apex:
`.claude/rules/right-size-ceremony.md`, `AgDR-0107`, and `CLAUDE.md` all cite `enforce-budget.sh` as the mechanism that watches process cost versus change size. That file is **not tracked in this repository** — it is `apexyard-premium#335`, gated on a `features.budget` entitlement.

So the open-source right-size-ceremony rule documents a safety net that open-source adopters do not have.

## Affected

- `.claude/rules/right-size-ceremony.md` (lines 48, 66)
- `docs/agdr/AgDR-0107-right-size-ceremony.md` (lines 19, 31)
- `CLAUDE.md` (line 116)

## Given / When / Then

**Given** `right-size-ceremony.md` states the enforcement split as *"the **Lean** floor is agent judgment, watched by `enforce-budget.sh`'s token meter"*
**When** an open-source adopter reads that and looks for the hook
**Then** `git ls-files .claude/hooks/enforce-budget.sh` returns nothing — the file is premium-only — so the rule's only non-judgment component does not exist for them; expected that the rule describe the enforcement an OSS adopter actually gets

## Repro

1. `git ls-tree -r dev --name-only | grep enforce-budget` → no results.
2. `grep -rn 'enforce-budget' .claude/rules/ docs/ CLAUDE.md` → five citations.
3. Read the header of the premium file: `# Token / cost budget enforcement — PreToolUse gate (apexyard-premium#335)`.

## Why it matters

`right-size-ceremony` exists to answer the recurring "this framework is optimised for compliance, not shipping" complaint. Its design is explicitly *advisory rule + token meter as the watch*. Remove the meter and the design reduces to advisory-only — which may still be the right call, but the docs should say so rather than implying a mechanism that is absent.

This is also the class of drift worth catching early: a rule that cites machinery which doesn't ship is indistinguishable, to a reader, from one whose machinery is quietly broken.

## Suggested fix

State the enforcement honestly rather than deleting the concept:

- `right-size-ceremony.md` and `CLAUDE.md`: describe the OSS enforcement as advisory self-discipline, and note the token meter as a premium addition where relevant.
- `AgDR-0107`: append a correction note rather than rewriting the recorded decision — the decision was made as stated; the citation was simply wrong about availability.

[...]

PR #1045, opened 2026-07-27
docs(#1044): right-size-ceremony cited a premium-only hook as its OSS watchdog

atlas-apex:
## Summary

- **Corrects three OSS documents that cited a premium-only hook as the framework's own watchdog.** `right-size-ceremony.md`, `CLAUDE.md`, and `AgDR-0107` all named `enforce-budget.sh` as the mechanism watching process cost versus change size. It isn't tracked in this repository — it's `apexyard-premium#335`, gated on a `features.budget` entitlement.
- **States the OSS enforcement honestly**: the Lean floor is agent judgment with no mechanical backstop. Premium adopters additionally get the meter.
- **Appends a Correction to AgDR-0107 rather than rewriting it**, with inline pointers on both affected lines.

## Why this is worth a PR

`right-size-ceremony` exists to answer the recurring *"optimised for compliance, not speed of shipping"* feedback. Its design is explicitly **advisory tier model + token meter as the watch**. Remove the meter and it's advisory-only — which may well still be the right call, but the rule should say so.

Every other rule in the self-discipline family (`plan-mode`, `parallel-work`, `loop-mode`) is candid about having no enforcement. This one read as though it had a meter behind it, which made the Lean floor sound firmer than it is.

There's a second-order cost too: a rule citing machinery that doesn't ship is indistinguishable, to a reader, from one whose machinery is quietly broken. Trust in what the docs claim is the thing being protected here.

## Changes

**`.claude/rules/right-size-ceremony.md`** — the "watch" section now leads with the correction plainly ("the watch is your own judgment; there is no token meter in the open-source framework"), names the premium component and its issue, and the enforcement-split paragraph reads "no mechanical backstop in the OSS framework" instead of "watched by `enforce-budget.sh`'s token meter".

**`CLAUDE.md`** — the Quality Rules bullet no longer implies the hook ships.

**`docs/agdr/AgDR-0107-right-size-ceremony.md`** — a `## Correction` section is **appended**. The decision was made as recorded; only the availability claim was wrong, so rewriting the recorded reasoning would falsify the record. Both in-body citations (Consequences, Decision) get a bracketed inline pointer so a reader landing on either sees the correction without scrolling.

## A note on the follow-up this sharpens

AgDR-0107's Consequences list a "dedicated advisory nudge-hook (process-cost-vs-change-size)" as a *deferred next increment* — a contingency if the Lean floor proved insufficient. Given the meter never shipped to OSS, that increment is really the **first** step rather than a fallback. The Correction says so, since it changes what the deferred work means.

## Testing

- Grepped every rule, AgDR, and `CLAUDE.md` for citations of untracked components; after this change no OSS doc claims `enforce-budget.sh` ships. (An initial broader scan flagged `_lib-*.sh` files too — that was my regex dropping leading underscores, not a real finding. Those all exist.)
- Full hook suite: **117/117**, unchanged from the clean baseline on `dev`.

Documentation only — no runtime behaviour changes.

AgDR not applicable: this corrects documentation about an existing decision and records that correction inside the affected AgDR. No new decision is being made, no dependency, service, schema, or security control changes, and neither safety rail is engaged.

Refs #1044

## Glossary

| Term | Definition |
|------|------------|
| Lean floor | The lightest ceremony tier in `right-size-ceremony.md`, for small reversible changes |
| token meter | A hook metering token/cost spend, to flag when process outweighs the change |
| advisory | Warns without blocking; contrast a gate, which blocks |
| premium component | Ships in `apexyard-premium` behind an entitlement, not in the OSS framework |


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1045

**Commit**: `e40199b7ade03b0776a415a816570a21d227be02`

### Summary

Corrects a factually wrong claim in `right-size-ceremony.md` (plus its two downstream citations): the rule named `enforce-budget.sh` as the mechanical watchdog metering process cost, but that hook is premium-only and does not ship in this repository. The fix replaces the claim with an honest statement of OSS enforcement, and appends a Correction to AgDR-0107 rather than rewriting it.

I reviewed this against the PR HEAD via `gh api` / a fetched `FETCH_HEAD`, **not** a local worktree — the checkout at `/Users/ahmed/apex/apexyard-work-wt` is on `fix/GH-1042-gate-placement`, a different branch.

### Checklist Results

- Architecture & Design:      N/A (docs-only, no runtime surface)
- Testing:                    N/A (no executable change; author reports suite 117/117 unchanged)
- Technical Decisions (AgDR): Pass (correctly marked not-applicable — corrects an existing record, makes no new decision)
- Adopter Handbooks:          Pass (advisory nit below)

### Verification of the premise — confirmed true

This PR originated from an adversarial challenge, so I verified the premise independently rather than accepting it:

| Check (at `e40199b7`) | Result |
|---|---|
| Hook named `*budget*` tracked under `.claude/hooks/` | **None** (67 tracked `.sh` files, no match) |
| Any file named `*budget*` anywhere in the tree | **None** |
| `enforce-budget` wired in `.claude/settings.json` | **Not wired** |
| Any other cost/token-metering hook in OSS | **None** (`git grep -iE "token.?(cost|spend|usage)|cost.?meter"` over `.claude/hooks/` returns nothing) |

The premise holds. The pre-PR text told OSS adopters a watchdog existed that they did not have. Judged by what git tracks on this branch — an untracked copy present in a given operator's fork is not the same as shipping.

### Completeness of the fix — all three sites, verified by whole-tree sweep

I swept the entire tracked tree at HEAD, not just the diff. Every surviving mention of `enforce-budget` is now correctly qualified:

| Location | State after this PR |
|---|---|
| `right-size-ceremony.md:50` | Corrected — names it premium, cites `apexyard-premium#335` |
| `right-size-ceremony.md:68` | Corrected — "no mechanical backstop in the OSS framework" |
| `CLAUDE.md:116` | Corrected — "the OSS framework ships no token meter" |
| `AgDR-0107:19` (Consequences) | Inline `[Incorrect — see Correction below]` pointer |
| `AgDR-0107:31` (Decision) | Inline `[The token-watch clause is incorrect…]` pointer |
| `AgDR-0107:44` (new Correction) | The correction itself |

**No unqualified citation survives anywhere in the tracked repo.** A broader sweep for `token meter` / `features.budget` / `budget hook` surfaced no other stale claim.

### Is the replacement honest, without over-swinging?

Yes, on both edges — this was the finding most at risk of being replaced by its own mirror-image error:

- It doesn't under-claim: I confirmed no *other* OSS cost signal exists, so "adopters running the OSS framework have no automatic disproportion signal" is literally accurate, not rhetorical.
- It doesn't imply premium is required for the rule to work. The rule remains fully usable for an OSS adopter: the tier model, the self-check, and both rails are intact and unqualified. Premium is mentioned as an *additive* meter ("premium adopters additionally get…"), not as a precondition.
- The added closing paragraph ("if the recurring 'too much process' feedback persists, that gap is the thing to close, not the prose") is the right note to land on — it names the real follow-up rather than declaring the problem solved by wording.

### Safety rails — intact

Both non-negotiables read byte-identical at HEAD; the diff never touches that section:

1. "Security and trust-chain never go Lean" — unchanged, still overrides the size signal.
2. "Ambiguity rounds up" — unchanged.

The `CLAUDE.md` bullet also still carries both rails verbatim after its rewrite. Nothing here weakens either.

### AgDR-0107 — correct historical treatment

This is the part I'd have pushed back on hardest if it had gone the other way, and the author got it right. The Decision and Consequences sections are **not** rewritten to say something they didn't say at the time. The original wording is preserved verbatim, annotated with a bracketed pointer, and the substantive correction is **appended** as a new `## Correction` section. That preserves the record as a record.

Two details that make it better than the minimum:
- Inline pointers on *both* affected lines, so a reader landing on either section sees the correction without scrolling to the bottom.
- The Correction reasons about the *consequence* for the decision rather than just flagging the error — specifically that AgDR-0107's deferred "advisory nudge-hook" is, for OSS, the first increment rather than a contingency. That's the genuinely useful part.

The chosen option itself is untouched and still stands on its own reasoning, which is correct — the availability claim was wrong, the decision wasn't.

### Issues Found

None blocking.

### Handbook Findings

**Commit Message Quality** (advisory) — `handbooks/general/commit-message-quality.md`

The commit body is exemplary against this handbook: full WHY-driven narrative, alternatives noted, per-file change list, `Refs #1044`, no WHAT-restating one-liner. No violation of any of the four flagged patterns.

- nit: the subject is `docs: right-size-ceremony cited…`, while the handbook asks for `type(#TICKET): description`. Worth noting that `.claude/rules/git-conventions.md` specifies `type: subject` for *commits* (ticket in the body via `Closes`/`Refs`), which this commit follows exactly — so the handbook and the framework rule diverge here, and the commit complies with the rule. Flagging the divergence, not the commit; no action needed on this PR.

`handbooks/architecture/migration-safety.md` is `ENFORCEMENT: blocking` but has no applicability — this diff touches no migrations, no schema, no SQL. `clean-architecture-layers.md` is likewise N/A for a prose-only diff.

### Suggestions

1. **nit (prose precision, non-blocking)** — `right-size-ceremony.md:70` now reads: *"Every other rule in the self-discipline family (`plan-mode`, `parallel-work`, `loop-mode`) is honest that it has no enforcement."* `loop-mode.md`'s own Backstop section does cite a partial backstop ("the merge-gate hooks already stop a loop that tries to self-approve"), and `isolated-builds.md` describes an "advisory backstop only". The claim is defensible for enforcement *of the tier heuristic itself*, but "is candid about being primarily self-discipline" would be strictly accurate. Raising it only because prose accuracy in rule files is the exact subject of this PR — the standard it sets applies to its own new sentence.

2. **Not a finding, verified** — since this is a public repo, I checked whether naming `apexyard-premium#335` and the `features.budget` entitlement leaks anything per `.claude/rules/leak-protection.md`. It doesn't: the `apexyard-premium` repo name and its issue references already ship publicly in `_lib-premium-hook.sh`, `reindex-on-session-start.sh`, `validate-search-config.sh`, and the cross-repo hook tests, and the `features.yaml` entitlement mechanism is fully documented in public OSS code. No pricing or monetisation detail is disclosed. Established precedent, no new surface.

### Verdict

**APPROVED**

Docs-only, no runtime behaviour change, both safety rails verified intact, and the core claim independently verified rather than taken on faith. The fix is complete across all three cited sites with nothing stale left in the tree, the replacement text is accurate in both directions, and the AgDR is corrected by appending rather than rewriting history — which is the right call and the one most likely to have been gotten wrong.

---
Reviewed commit: `e40199b7ade03b0776a415a816570a21d227be02`

Issue #1032, opened 2026-07-27
[Bug] /approve-merge's documented flow is blocked by its own marker gate (role taken from the verification read, not the write)


atlas-apex:
## Affected

`.claude/hooks/warn-review-marker-write.sh` (`_extract_marker_role`) + `.claude/skills/approve-merge/SKILL.md` (steps 4–5)

## Given / When / Then

**Given** `/approve-merge`'s documented flow: step 4 verifies the Rex marker by calling the marker-path helper with the `rex` role, then step 5 writes the CEO marker by calling the same helper with the `ceo` role and redirecting into the resulting variable
**When** both steps run in a single shell block, exactly as SKILL.md documents them
**Then** the write is **BLOCKED** with "You are about to write a `*-rex.approved` review marker with no matching active-reviewer session marker" — expected that it succeed, since the actual write target is a `-ceo.approved` file, which is advisory-only and explicitly *not* in the blocking set

## Root cause

`_extract_marker_role` takes the **first** marker-path-helper call found in the command text:

- it greps for the helper name, takes `head -1`, then extracts the first role keyword from that call

In `/approve-merge`'s flow the first such call is the **rex read** (step 4); the write's own call (`ceo`) is second. So:

1. The write detector fires (the step-5 redirect).
2. Write-target extraction yields a variable-derived target, setting the "has variable target" flag.
3. That flag triggers the broader `_is_marker_plausible_indirect` fallback across the **whole** command, which matches on the literal helper name.
4. `_extract_marker_role` then reports `rex` — taken from the *read*, not from the write.
5. No `active-reviewer` marker exists for rex, so `_active_reviewer_allows` fails closed and the write is blocked.

The block message reports role `rex` and `pr: unresolved`.

## Repro

1. On a PR with a valid `*-rex.approved` marker at HEAD, follow `.claude/skills/approve-merge/SKILL.md` literally.
2. Issue step 4's verification and step 5's marker write in one Bash tool call, as SKILL.md presents them.
3. Blocked, with target reported as `<resolved via variable/function indirection; detected role: rex, pr: unresolved>`.

Workaround: split verification and the write into two separate tool calls, so the writing call's text never mentions the `rex` role.

### Second repro — this very bug report

[...]

PR #1049, opened 2026-07-28
fix(#1026): the marker-write gate is a backstop — make it advisory

atlas-apex:
## Summary

- **The marker-write gate stops blocking and starts warning.** It decides by pattern-matching the *text* of a shell command, which AgDR-0104 established cannot be made sound. In one session that cost **13 false positives across 2 hooks**: a read-only `grep` (path was in the search pattern), a `git commit` (in the message), a code review writing its own prose, `/approve-merge`'s own documented merge step, and a JSON payload piped to another program — from which the extractor returned the role as literal garbage text. Every one reported `detected role: unresolved`.
- **This is safe because the block was belt-and-braces on a cause already repaired.** #843 promoted it after build agents forged approvals to comply with `auto-code-review.sh`'s "Invoke Rex NOW" instruction — which they cannot follow. That banner now tells sub-agents to stop and hand back. The build-agent deterrent paragraph is **retained**; only the exit code changes. Merge integrity rests where it always did, and the PR verifies it hasn't moved.
- **Applies AgDR-0104's long-missing `CLASS:` label to five trust-chain hooks** — four are *controls* deciding on structured state (a file's contents, a forge-reported SHA or CI conclusion) and must stay fail-closed; one is a *backstop* reading command text. Reading them as the same kind of thing is how four rounds of patching happened.
- **Fixes the role-extraction bug behind one of the false positives (#1032).** `/approve-merge` READs the rex approval to verify it, then WRITES the ceo one. First-call-wins reported `rex` for a ceo write, routing an advisory write down the blocking path. Disagreeing roles now report *unresolved* rather than asserting the one merely being read.
- **Corrects `auto-code-review.sh`, which asserted enforcement this PR removes.** It claimed the marker write "will BLOCK anyway" in three places. A framework that misstates its own enforcement teaches agents to discount the rest of the banner.

## What is and isn't detected — stated precisely

Earlier revisions of this PR claimed four known bypasses were "accepted limits." **That was wrong, and Rex caught it.** The accurate position:

| Shape | Detected today? |
|---|---|
| Split path (`S=…; printf x > "$S/…"`) — **#1026** | **No.** The one genuine open limit, left open deliberately |
| Variable indirection (#962) | Yes — cases 20/23/27 |
| Interpreter heredoc | Yes — cases 35-39 |
| BSD / GNU `sed -i` | Yes — cases 35-39 |

The argument for going advisory rests on the 13 false positives, not on the hook being blind. It isn't.

## Why this and not the alternatives

AgDR-0109 decided this on 2026-07-25 and reached `status: executed` **with none of it built**. This PR implements it while departing twice — recorded in a **new** record, **AgDR-0111**, because shipping an option AgDR-0109 explicitly *rejected* re-opens the options table, which is the documented test for requiring a new AgDR rather than a changelog entry.

| AgDR-0109 said | This PR ships | Why |
|---|---|---|
| Block only on a confidently-resolved target | Plain advisory | The resolver demonstrably emits garbage (`…__7777-rex.approved"}}`). "Confidently resolved" isn't a state it can be trusted to report, so no text matcher belongs in a blocking path. |
| Activate the deferred CI meta-gate | Not built | It adds permanent cost to every future trust-chain hook to *manage* a class this change *removes*. AgDR-0104's deferral is therefore **un-discharged**, trigger narrowed to "a new **control** is added." #1015 stays open at that scope. |

#1026 asked to **decide direction rather than add a fifth pattern**. Direction decided: split-path stays open on purpose, because the next spelling would still pass.

## Testing

1. `bash bin/run-hook-tests.sh` → **119/119 PASS**.
2. `bash .claude/hooks/tests/test_warn_review_marker_write.sh` → 47 assertions, including 5 new regressions reproducing the real false positives.
3. **The property that must not move** — `test_block_unreviewed_merge.sh` confirms `rex sha mismatch → blocks`, `ceo sha mismatch → blocks`, bare-SHA legacy markers refused, and the `gh api` merge shape still gated. A forged approval carrying a stale SHA is still refused at merge.
4. `shellcheck -S warning` clean on all touched hooks and test files; `markdownlint-cli2` clean on all docs.

## Reviewer notes

Trust-chain change, Heavy tier — the Security Auditor trigger fires on `.claude/hooks/**` (#777). Worth your scrutiny: whether removing the block genuinely re-opens #843 given the inducing banner is fixed, and whether naming the *absence* of enforcement in the build-agent banner invites the forgery it warns against.

Refs #1026
Refs #1032
Refs #1015

## Glossary

| Term | Definition |
|------|------------|
| Trust chain | The hooks under `.claude/hooks/**` that gate merges and approval-marker writes |
| Control vs backstop | AgDR-0104's distinction. A *control* decides on structured state (a file's contents, a forge-reported SHA) and must fail closed; a *backstop* is best-effort local defence behind a real gate, and should warn |
| Fail-closed | When a gate cannot evaluate its precondition, it denies rather than allows. After this PR, a property of the control hooks only |
| False positive | Here: the gate blocking a command that never writes an approval — a grep, a commit message, a payload |
| Evasion | A real approval write spelled so the text matcher doesn't see it. Only the split-path shape (#1026) evades today; see the detection table above |
| Forge-reported HEAD | The PR's current commit SHA as GitHub/GitLab reports it — structured state a local file write cannot fake |


atlas-apex (commented):
## Code Review: PR #1049

**Commit**: `295f41d6eec8ceedff37dd7ab61edc89348e3ca9`

### Summary

Turns `warn-review-marker-write.sh` from a blocking gate into an advisory one, labels five trust-chain hooks control-vs-backstop per AgDR-0104, fixes the #1032 role-extraction bug, and reshapes the tests.

**I agree with the direction.** The evidence for going advisory is strong, and I verified the load-bearing safety claim independently: the control has not moved. What I cannot sign off on is that two of the PR body claims do not survive checking against source — and one of them is a factual error written into a trust-chain hook header, where the next maintainer will inherit it.

### Checklist Results

- Code Quality:               **Fail** (see Blocking 2)
- Testing:                    Pass with findings
- Security:                   Pass on the control; **Fail** on deterrent accuracy (Blocking 1)
- Summary Bullet Narrative:   Pass (genuinely narrative)
- Technical Decisions (AgDR): Pass with a judgement call (see AgDR section)

---

## Verification of the four claims

**Claim 1 — #843 root cause is fixed. SUBSTANTIALLY TRUE.** `auto-code-review.sh` no longer issues an unconditional "Invoke Rex NOW". It has an explicit build-class branch: *"Do NOT attempt to review this PR ... do NOT write ANY file under the reviews session dir ... Report the PR back to the orchestrator plainly and stop."* A build agent reading that today has no honest route to concluding it should write a marker. The inducement is genuinely gone. **But see Blocking 1** — that same paragraph now contains a false statement this PR creates.

**Claim 2 — deterrent retained in full. TRUE within the hook.** Diffed the banner: the BUILD-CLASS SUB-AGENT paragraph survives verbatim apart from `gate` becoming `warning`. The new *"IF THIS COMMAND DOES NOT WRITE A MARKER ... Proceed"* paragraph is correctly conditioned on the command **not** writing a marker, so a genuine forgery is not covered by it, and it carries an explicit anti-obfuscation warning. That is good drafting. Minor: it is the final paragraph, so it gets recency weight over the STOP paragraph above it — worth considering ordering, not worth blocking.

**Claim 3 — the control has not moved. VERIFIED TRUE.** This is the one that mattered most and it holds:

- Shebang is intact on line 1 of all four labelled hooks; the inserted `CLASS:` block is lines 2-7 and is 100% comment lines. Zero executable delta in all four files (6 added, 0 removed each).
- `block-unreviewed-merge.sh` still fails closed: exit 2 on unparseable input when the payload looks merge-shaped, on missing markers, on SHA mismatch, on missing `approved_by=user`, and on `skill_version < 2`.
- It still compares against forge-reported HEAD, falling back to local HEAD only with a visible WARN.
- `test_block_unreviewed_merge.sh` 39/39 PASS, including rex sha mismatch blocks, ceo sha mismatch blocks, bare-SHA ceo marker blocks, and the `gh api` shape staying gated.

**Claim 4 — the #1032 extraction fix. CORRECT.** `sort -u` then `grep -c ''` behaves properly at all three boundaries: zero matches are caught by the preceding `[ -z ]` guard; one match yields exactly 1 line because `echo` supplies the terminating newline; two or more yields >1 and returns 1. No trailing-newline miscount.

On your specific question about empty `MARKER_TYPE` routing: it does **not** bypass a branch it used to hit. Before this PR the ambiguous `/approve-merge` command resolved to `rex` (first-call-wins) and went down the **blocking** path — it never reached the ceo branch. Now it resolves to empty, skips the ceo branch, and lands in the generic warning branch, which exits 0. Net effect is block becoming warn, which is the intent. One cosmetic consequence worth noting as a nit: a ceo write with ambiguous roles now prints the rex/security/architecture banner telling build agents to STOP and pointing at `/code-review`, which is the wrong advice for that command. Harmless at exit 0, but case43 locks the behaviour in.

---

## Blocking findings

### 1. `auto-code-review.sh` now asserts something this PR makes false

Not in the diff — and that is the problem. It still says, in the build-agent deterrent paragraph:

> warn-review-marker-write.sh will BLOCK that write anyway (#843), but the instruction stands regardless.

And in the orchestrator branch:

> spawning the code-reviewer agent directly without it leaves the write blocked by warn-review-marker-write.sh — by design.

Plus the header: *"(upgraded to a BLOCKING gate in #843)"*.

After this PR all three are false. The PR body states the deterrent is *"retained in full"* — true inside the hook you changed, but the sibling banner carrying the other half of that deterrent now makes a false enforcement claim, and this PR is precisely what falsifies it. That matters more than a stale comment normally would: a build agent that tests the claim discovers the framework misstated its own enforcement, which discounts every other instruction in the same banner. The orchestrator branch also now misinforms about *why* `/code-review` must be used.

Fix is text-only: update those three spots to say the write is detected and warned, and that the instruction stands on the rule rather than on the block.

### 2. The "four accepted limits" are not honestly stated — two are detected, and this PR's own tests prove it

The new header lists as ACCEPTED LIMITS, with *"Do not 'fix' them with another pattern"*: split path, interpreter heredoc, BSD `sed -i`, variable indirection — framed as *"spellings that hide the path from a substring match"*. Checked each:

| Claimed limit | Actually? | Evidence |
|---|---|---|
| Interpreter heredoc | **DETECTED** | Cases 35, 36, 37 assert `WARNING` in stderr, i.e. MATCHED=1. Fixed in #1000-round-2. |
| BSD `sed -i` | **DETECTED** | Cases 38, 39 assert `WARNING`. `IS_EXTRACTION_FRAGILE` exists specifically to catch it. |
| Variable indirection | **DETECTED** | Cases 20, 23, 27. Also confirmed live — see finding 6. |
| Split path | Plausibly genuine | The only one of the four I could not refute. |

This contradicts the same file roughly 200 lines lower, where the `#1000-REX-ROUND-2` block states removing the heredoc strip *"returns all three interpreter-heredoc shapes to BLOCKED"*. The header and the body of one trust-chain hook now disagree about what that hook does.

Three reasons this is blocking rather than a nit:

1. It tells a future maintainer not to fix things that already work — an active invitation to delete working detection as dead weight.
2. It is load-bearing rhetoric, repeated in the PR body, the test-file header (*"while four distinct bypasses ... stayed open"*), and the AgDR changelog. It inflates the case for going advisory. The real case stands on the 13 false positives alone and does not need this.
3. It is inside the trust chain, which by house rules is exactly where imprecision is not tolerated.

If the intended meaning is "now that nothing blocks, nothing is stopped" — that is trivially true of every shape and does not belong in a list of four. Please cut it to the shapes that genuinely evade **detection**, or reframe the section as *"known evasions, mostly detected; detection is not enforcement."*

---

## Non-blocking findings

**3. Two test labels now contradict their own assertions** — mechanical `sed` damage:

- L349: `"Rex blocked banner contains BLOCKED keyword"` asserts `0 "WARNING"`. The label names the opposite of the assertion.
- L711: `"Bash rm + real marker forgery in same command -> still BLOCKED (#1000)"` asserts `0 "WARNING"`. The worse of the two: a reader scanning for security properties reads *"still BLOCKED"* as a live guarantee that no longer holds.

Also `-> WARNS (advisory, #1026), exit 2.` in the case 35/36/37/38 comment headers — the `sed` replaced the word but left `exit 2`, producing self-contradictory sentences.

**4. Case 43's assertion is meaningful but imprecise.** Not vacuous — I checked. If detection failed entirely, stderr would be empty and the assertion would fail, so it genuinely asserts detected AND role unresolved. But the banner prints both `detected role: ...` and `pr: ...`, each defaulting to `unresolved`. A future regression that restored `role=rex` while losing the PR number would still emit `unresolved` and the test would pass with the #1032 bug back. Assert the full phrase `detected role: unresolved` instead.

**5. The test-file header matrix.** Roughly 30 lines still read `BLOCKED, exit 2` with a preamble instructing the reader to mentally substitute. I would not block on it, but it is a maintenance trap of exactly the kind this PR is trying to end: the preamble is a compatibility shim over a stale table, and the next `sed` sweep will not know to preserve it. Since a `sed` already ran over the assertions, running one over the matrix costs nothing.

**6. Live corroboration of the PR premise (in your favour).** While reviewing, the currently-installed pre-PR hook blocked **this review** twice: once for a probe command whose text contained a marker path, once for writing a test harness whose *content* mentioned one. Both reported `detected role: rex`. That is two more false positives reproduced against the reviewer of the change itself, and it is the strongest single argument in the PR. I routed around by not spelling the path — deliberately not by obfuscating it, per the new banner's own instruction.

---

## AgDR changelog vs new AgDR — my view: this needs a NEW AgDR

You asked for a real opinion, so: **changelog is the wrong vehicle here.**

`docs/rule-audit.md` states the test plainly — a new AgDR is required when the change involves *"re-opening the options table — a new option you didn't consider, a reversal of the chosen path, a different tradeoff."* Applying it:

- AgDR-0109's options table lists **B — revert fully to advisory** and **rejects it**, on the record, with a reason: *"Re-opens the induced-impersonation hole #843 shipped to close — a real, twice-exercised failure, not hypothetical."* This PR ships B. Adopting a previously-rejected option from the table is not adjacent to re-opening the table; it is the definition of it. The changelog's own wording concedes this — *"Shipped option B, not option E"* — while the sentence above it asserts *"no option was reopened."* Those cannot both be true.
- Decision **part 3** (the CI meta-gate) was dropped entirely. AgDR-0109 argued at length that AgDR-0104's trigger *"has fired,"* and its Consequences section still claims *"AgDR-0104's deferred item is discharged."* As shipped, it is not. A Consequences section that is now false, corrected only by a changelog 40 lines below it, is the failure mode the rule-of-thumb exists to prevent.

Apply the document's own test — *"would a reader who only saw the new state understand why without re-reading the old options?"* No. To understand why advisory is acceptable despite #843, the reader must read the options table's rejection of B **and** the changelog's rebuttal of that rejection. That answer is "no", which the rule maps directly to: new AgDR.

Compounding it: the frontmatter still reads `status: executed` and the title still reads *"block only on a resolved target, and activate the deferred meta-gate"* — neither of which describes what shipped. Anyone finding AgDR-0109 by title or status gets a wrong picture, corrected only at the bottom of the file.

To be clear about what I am *not* saying: the changelog is well-written and unusually honest about both departures, and the reasoning inside it is sound. My objection is to the vehicle, not the candour. AgDR-0110 superseding 0109, reusing this changelog text as its Context and Decision, would cost very little and would leave 0109 as the accurate historical record of a decision moment that was then superseded — which is what AgDR-0001 says AgDRs are for.

Flagging this as a strong recommendation rather than a hard blocker, since it is a genuine judgement call and reasonable people land differently.

---

## Standard checks

- **Glossary**: present, six terms, genuinely useful.
- **Ticket refs**: Refs #1026, #1032, #1015 — all valid and accurately described.
- **Summary bullets**: narrative; each says what changed and why it matters. No advisory finding.
- **Secrets**: none.
- **Tests**: `test_warn_review_marker_write.sh` 47/47 PASS, `test_block_unreviewed_merge.sh` 39/39 PASS. Both verified locally, not taken on trust.

---

## Verdict

**CHANGES REQUESTED**

To be unambiguous: **the design decision is right and I would approve it on the two text fixes.** I verified the control is intact, the extraction fix is correct, and the false-positive evidence is real — I hit two of them myself while reviewing. Nothing here asks you to reconsider going advisory.

Required before approval:

1. Correct the three now-false "will BLOCK" statements in `auto-code-review.sh` (Blocking 1).
2. Correct the ACCEPTED LIMITS list so it does not claim four bypasses when this PR's own tests demonstrate at least two are detected (Blocking 2).

Strongly recommended: fix the two contradictory test labels (3), tighten case 43's assertion (4), and take the AgDR to a new record rather than a changelog.

The reason I am holding on two text edits for a change I otherwise agree with: this PR's argument is that the framework should stop over-claiming what a text matcher can enforce. It should not ship with two fresh over-claims of its own, one of them inside the trust chain.

---
Reviewed commit: `295f41d6eec8ceedff37dd7ab61edc89348e3ca9`

atlas-apex (commented):
## Code Review: PR #1049 — round 2

**Commit**: `ee1b27fc5b7781c565a49ea0ee96628b9064639d`

### Summary

Round 2 on the marker-write gate going advisory. All three round-1 findings were addressed with real work, not cosmetics — but **one of the three is only partially landed**, and it is the one carrying the operator's own bright line: *"if any copy of the wrong claim survives, that's a blocker."* Three copies survive, including one in a file this PR newly wrote and one inside the AgDR that corrects it.

Scope confirmed against the forge (11 files, 959 diff lines) — my local `origin/dev` merge-base was stale and inflated the picture; the four merge-gate hooks really are comment-only (`CLASS:` labels), so merge integrity is unmoved. CI is fully green (6/6). Locally: 47/47 and 39/39.

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass — control hooks comment-only; forge-HEAD SHA comparison unmoved
- Performance: N/A
- PR Description & Glossary: **Fail** — carries a corrected-elsewhere claim (see B3)
- Technical Decisions (AgDR): Pass — AgDR-0111 is the right vehicle and is well-built

---

## Finding 1 — `auto-code-review.sh`: FIXED, and the new framing is stronger

All three false-enforcement claims are gone (header, sub-agent branch, orchestrator branch). Verified by diff, not by grep alone.

**On the question you asked me to answer directly** — does naming the absence of enforcement invite the forgery it warns against? **No, and I think the new text is a materially better deterrent.** Reasoning, since you asked for a real judgment rather than a nod:

- The old text made an **obstacle claim** ("will BLOCK that write anyway"). Obstacle claims are load-bearing only while true. It was already false, and a sub-agent that tried it and succeeded would have had *no remaining reason* to comply — the instruction's entire weight rested on a barrier that wasn't there. It also actively invited the "the system handles this, I don't need to care" reading.
- The new text supplies a **reason grounded in harm to a specific human** — "would record a review that never happened, and the human approving the merge would be relying on it." That does not depend on a barrier existing, so it cannot be falsified by experiment.
- Crucially, **the #843 failure mode was not gap-probing.** Build agents forged markers to resolve an *instruction conflict* (told "Invoke Rex NOW", unable to nest the Agent tool). That inducement is independently repaired — the banner now has an explicit build-class branch telling sub-agents to stop and hand back. Disclosure of "no block" does not recreate the motive, because the motive was the contradiction, not the opportunity.

One residual worth naming honestly, because the correction **created** it (nit, non-blocking): the banner's closing line still tells the reader the merge gate will refuse this PR until a Rex approval file exists. Composed with the new "Nothing will mechanically stop you", a goal-directed agent now holds, in one message, both *"the merge needs this file"* and *"you can create it unimpeded."* Neither statement is wrong; the adjacency is new. It was not a composition risk while the second half read "will BLOCK anyway." Mitigated by the sub-agent branch (stop, hand back — it isn't the one merging), so I would not gate on it, but consider adding to the sub-agent branch that the merge gate compares the marker's SHA against forge-reported HEAD, i.e. the file alone is not what unlocks anything.

Separate nit, pre-existing and out of scope: that same line names the marker in unqualified `<pr>-rex.approved` form, but markers are repo-qualified per AgDR-0060. Flagging only because this PR's own thesis is that "a framework that misstates its own enforcement teaches agents to discount the rest of the banner" — the argument applies to the path format too.

---

## Finding 2 — the "four accepted limits" claim: **BLOCKER, partially fixed**

The **hook header is correctly rewritten** and is the model for the rest. It states one accepted limit (split path), names variable indirection / interpreter heredoc / BSD-GNU `sed -i` as **detected**, cites cases 20/23/27 and 35-39, and adds *"Don't cite them as evidence the hook sees nothing."* It also does **not** overstate in the other direction — it retains "detection is unreliable in BOTH directions" and the symmetric over-firing limit. Exactly right.

But the correction did not propagate. Three copies survive:

**B1. Test-file header — the worst instance** (`.claude/hooks/tests/test_warn_review_marker_write.sh:25-28`):

> `while four distinct bypasses (split path, interpreter heredoc, BSD sed -i, variable indirection) stayed open.`

This is **newly written by this PR** (it is inside the new `#1026 SUPERSEDES THAT PROMOTION` preamble), so it is not inherited history — and it sits in the very file whose cases 20/23/27 and 35-39 prove the opposite. A file whose header contradicts its own passing assertions is precisely the "corrected in one place, left standing in another" failure you flagged as worse than being consistently wrong.

**B2. AgDR-0111:63**, option G row: *"The status quo that produced 13 false positives and 4 evasions."* Contradicted by its own Decision part 3 eighteen lines later (lines 80-83). The document that corrects the claim also restates it.

**B3. PR body**, twice: the *"#1026 asked to decide direction"* paragraph (*"the four known evasions … are now documented accepted limits in the hook header, not open bugs"*), and the Glossary's `Evasion` row, which defines the term by listing the three **detected** shapes as examples. This one has the most reader impact: "Reviewer notes" then asks the reviewer to scrutinise *"whether the four accepted limits are honestly stated"* — steering the reviewer with the claim the PR elsewhere retracts.

Required: align all three with the hook header's wording. Text-only, no logic.

---

## Finding 3 — AgDR vehicle: substantively fixed, one residual

**AgDR-0111 is well-built and I want to say so plainly.**

- **Honest about reversing a rejection.** The title is literally *"Ship the option AgDR-0109 rejected"*. Option E appears in the options table attributed to AgDR-0109; § "The safety argument that makes B viable now" names the original objection verbatim and answers it with two specific facts. `Reverses AgDR-0109's rejection of option B, on evidence unavailable when it was written.` No pretence that B was newly discovered. Meets the bar.
- **Un-discharge recorded correctly.** Decision part 2: *"AgDR-0104's deferral is therefore NOT discharged … trigger narrowed to: a new control (not a backstop) is added."* Repeated in Consequences and Artifacts. #1015 scoped accordingly.
- **Supersede relationships accurate.** *"Supersedes AgDR-0109 § Decision parts 2 and 3. Part 1 (the backstop label) stands and is implemented here."* Verified — part 1 is genuinely implemented (`CLASS: BACKSTOP` in the hook, `CONTROL` on the four merge-gate hooks).
- **Format/frontmatter conform** to `templates/agdr.md` and house style.

**Residual (required):** AgDR-0109's **Consequences section still asserts the discharge**, unannotated:

> `- AgDR-0104's deferred item is discharged.`

and, in the same section:

> `- The #843 induced-impersonation case stays blocked.`

Both are now false. The changelog at the bottom does correct the first, and 0109's Artifacts still says *"Supersedes the deferral in AgDR-0104 § Consequences"* — so a reader reaching Consequences before the changelog gets the wrong answer twice. Same class as Finding 2. House convention keeps AgDR bodies as immutable history, so annotate rather than rewrite — appending "— **superseded by AgDR-0111; see Changelog**" to each bullet is sufficient.

---

## Round-1 non-blocking items — all handled

- **Case 9** (former L349): fixed **and strengthened beyond what I asked** — now asserts `WARNING`, plus a *new* assertion pinning `BUILD-CLASS SUB-AGENT` so the deterrent text cannot silently drop out of the banner. Good instinct. *(Nit: the case's comment header at L344 still reads `rex blocked case contains "BLOCKED"`, stale against the fixed assertions.)*
- **Case 33** (former L711): fixed exactly as asked. Label reads `still DETECTED, warns`, with an explicit NOTE that it pins DETECTION not prevention, that prevention comes from `block-unreviewed-merge.sh`'s forge-HEAD SHA comparison, and "do not read this case as a merge guarantee."
- **Case 43**: tightened to the exact banner field `detected role: unresolved`, with an inline comment explaining why the bare word was too loose. Fixed.

### Re-judging the stale matrix (you asked me to revisit)

**The preamble is sufficient for the matrix rows — do not rewrite ~30 rows.** The remap ("read `BLOCKED, exit 2` as `detected → warns, exit 0`") is a legitimate single-point-of-truth device, and a mass rewrite would be a large diff with real transcription-error risk for zero behavioural gain. The *case labels* — the strings that actually print on failure — were corrected, which is the part that matters operationally.

**One narrow exception (should-fix, not a blocker):** the preamble only remaps the literal string `BLOCKED, exit 2`. It does not reach the **`fail-closed`** phrasing at L69, L544 and L553. "Fail-closed" is not a label the preamble can remap — it is a *semantic claim*, and it is now false: nothing fails closed in a hook that exits 0 unconditionally. It matters more than usual here because `fail-closed` is the exact vocabulary AgDR-0104 uses to separate CONTROL from BACKSTOP, and this PR's own thesis is that conflating those two readings is what produced four rounds of patching. Suggest "falls to the warning branch on total ambiguity" or similar.

---

## New evidence found during this review: the sibling hook still false-positives on review prose

While writing this review I was **blocked by `block-unreviewed-merge.sh`**. My first attempt to write this review body quoted the merge command literally in the Finding 1 discussion; `is_merge_command` matched my *review text*, PR-number extraction then failed, and the hook refused with:

> `BLOCKED: Could not determine PR number for merge. Run from a PR branch or pass an explicit PR number.`

Disclosure, so this is not mistaken for a silent workaround: I am reporting it here as instructed rather than rewording to hide it, and the version of this body that finally posted paraphrases the command string instead of spelling it. That paraphrase is a workaround, and I am naming it as one.

Three observations, offered as evidence rather than as a request to widen this PR:

1. It is a live, reproducible instance of **the exact false-positive class this PR exists to fix** — a code review blocked by the gate it is reviewing — and it corroborates the PR body's "13 false positives across **2 hooks**" claim with a fresh data point on the second hook.
2. It is **not** fixed by this PR, and correctly so: `block-unreviewed-merge.sh` is a **CONTROL** under AgDR-0104 and must stay fail-closed. The right resolution is not to soften it.
3. The narrow, in-class observation is that the *fail-closed branch fired on a payload that was never a merge attempt*. AgDR-0109 § Decision part 2 draws exactly this distinction — "ambiguity about whether the protected action **exists**" versus "ambiguity about whether it is **authorised**" — and argues only the latter warrants blocking. That reasoning was applied to the backstop and not to the control. Worth a follow-up ticket against #1015's narrowed trigger; explicitly **out of scope here**.

---

### Did the corrections introduce anything new?

Yes, and it is the reason for the verdict: **B1 and B2 are new factual errors written by this round**, not survivals of old text. B1 in particular is a header this PR authored that contradicts its own file's passing tests.

**None of the three fixes is cosmetic.** Finding 1 removed real claims and replaced them with real reasoning; Finding 2's hook header is a genuine rewrite with case citations; Finding 3 is a new, correctly-structured AgDR with accurate supersede semantics. The gap is propagation, not substance.

### Required before approval (all text-only)

1. `test_warn_review_marker_write.sh:25-28` — the "four distinct bypasses … stayed open" sentence.
2. `AgDR-0111:63` — option G's "4 evasions".
3. PR body — the "decide direction" paragraph and the Glossary `Evasion` row.
4. `AgDR-0109` Consequences — annotate the "discharged" and "stays blocked" bullets.

- `fail-closed` phrasing at test L69/L544/L553.
- Stale case-9 comment header (L344).
- Banner composition risk (Finding 1) and the unqualified marker path in the banner's closing line.

### Verdict

**CHANGES REQUESTED** — on propagation only. The direction, the hook header, AgDR-0111, and every round-1 test fix are sound; four text edits close this out and I would not expect a round 4.

---
📌 Reviewed commit: `ee1b27fc5b7781c565a49ea0ee96628b9064639d`

atlas-apex (commented):
## Code Review: PR #1049 — round 3

**Commit**: `278513a876dead5ffdcfe1efaf892abfbfdd7155`

### Summary

Round 2 was CHANGES REQUESTED on propagation only. **All five fixes are verified closed**, and I found no sixth copy of the corrected four-evasions claim anywhere in the tree at this HEAD. I grepped the full tarball at `278513a` — hooks, rules, docs, skills, agents, tests, settings — for every spelling of the count claim (`four/4/three evasions`, `known bypasses`, `accepted limits`, non-detection claims for heredoc / `sed -i` / indirection). Zero survivors. That work is done.

But the same propagation check turned up a **different** stale claim, in six files this PR does not touch, and it is the more serious one: the framework still documents this hook as a **BLOCKING** gate. That is the exact harm AgDR-0111 § Consequences names in its own words — *"A framework that misstates its own enforcement teaches agents to discount the rest of the banner."* The PR fixed `auto-code-review.sh` for precisely that reason and stopped there.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (assertions correct; comments are not — see F2)
- Security: **Pass** — the control property is unmoved; `block-unreviewed-merge.sh` takes a `CLASS:` label and no behavioural change
- Performance: **N/A**
- PR Description & Glossary: **Pass** — the detection table and the plain "that was wrong, and Rex caught it" are the right shape
- Technical Decisions (AgDR): **Pass** — AgDR-0111 linked; new-record-not-changelog reasoning is correct
- Adopter Handbooks: **N/A** (none loaded — no `architecture/`, `general/`, or diff-matched `language/` handbooks)

### Verification of the five round-2 fixes

| # | Item | Verified |
|---|---|---|
| 1 | Test preamble | **Closed.** Names split-path as the single undetected shape; states interpreter heredoc, BSD/GNU `sed -i`, variable indirection ARE detected, citing 20/23/27 and 35-39. I checked those cases exist and assert it — 35 (`python3` heredoc), 36 (`bash` heredoc wrapping `python3 -c`), 37 (inline control), 38 (BSD `sed -i ''`), 39 (GNU `sed -i`), all asserting `0 "WARNING"`. The hook header's own `ACCEPTED LIMIT` line (L23) is singular too. |
| 2 | AgDR-0111 option G | **Closed.** Reads "while still not detecting the split-path shape". No count claim. |
| 3 | PR body | **Closed.** Detection table is accurate against the tests. |
| 4 | AgDR-0109 Consequences | **Closed** — see the note below on form. |
| 5 | `fail-closed` vocabulary | **Closed.** `rule-audit.md`'s backstop row now reads "Advisory. Warns, never blocks"; case 24's comment and label carry the corrected framing plus the do-not-reintroduce note. The five surviving `fail-closed` mentions inside `warn-review-marker-write.sh` (L189/244/262/702/734) are historical `#962`/`#974`/`#977` archaeology, and L262's paragraph is explicitly annotated as what `#1026` reverses. Correct as-is. |

**On item 4's form** — the two flagged bullets carry blockquote `> **Superseded 2026-07-28 (AgDR-0111)**` annotations that leave the original claim intact and say plainly what is no longer true. That is the right shape for a historical record: annotating, not rewriting. Paired with the `## Changelog` section at the foot of AgDR-0109, the supersession is discoverable. Good.

**Round-2 nit, re-judged: acceptably mitigated, no edit needed.** The banner no longer closes on "merge needs this file". `nothing is blocked` is now scoped to the false-positive branch, and the closing paragraph reframes the file as *not* load-bearing ("Merge integrity rests on the per-PR human approval and on `block-unreviewed-merge.sh` matching each marker's SHA against the PR's forge-reported HEAD — neither of which a local file write can fake"). `auto-code-review.sh`'s build-agent branch carries the motivational weight instead ("Writing that file would record a review that never happened, and the human approving the merge would be relying on it"). Naming the absence without inviting it — that lands.

### Issues Found

#### F1 — BLOCKER: six files still document this hook as a BLOCKING gate

This PR makes `warn-review-marker-write.sh` exit 0 on every path. Ten statements across six untouched files say otherwise:

| File | Lines | Claim |
|---|---|---|
| `.claude/rules/pr-workflow.md` | 108, 119, 121, 124 | "is a **BLOCKING** gate"; "The write is **blocked (exit 2)** unless…"; "mechanically blocked for a caller with no active-reviewer marker set"; "the first, third, and fourth are mechanically blocked" |
| `.claude/agents/code-reviewer.md` | 626 | "a build agent's identical write is blocked (exit 2)… a caller with no active-reviewer marker set is blocked" |
| `.claude/agents/solution-architect.md` | 227 | identical claim |
| `.claude/skills/code-review/SKILL.md` | 41, 61 | "lets Rex's `*-rex.approved` write through the blocking marker gate"; "is correctly blocked" |
| `.claude/skills/security-review/SKILL.md` | 52, 72 | same pair |
| `.claude/skills/design-review/SKILL.md` | 37, 57 | same pair |

Why this is a blocker and not a docs-cleanup follow-up:

1. **`pr-workflow.md` is `@`-imported by `CLAUDE.md`.** It is in the context of *every* session. On merge, the framework's always-loaded rule text asserts an enforcement mechanism that no longer exists — the precise failure mode AgDR-0111 § Consequences condemns and that this PR spent a bullet fixing in `auto-code-review.sh`.
2. **`code-reviewer.md` is the file that defines me,** and `solution-architect.md` defines Tariq. Both tell the reviewer agent that a build agent's identical write "is blocked (exit 2)". That sentence is load-bearing for the author-vs-reviewer reasoning those agents use to justify writing their own marker. Post-merge it is false, and an agent that discovers a documented mechanical guarantee is fiction has good reason to discount the surrounding text — which is the argument for *this whole PR*.
3. **It is round 2's lesson, at larger scale.** Round 2 blocked on three copies of one sentence. This is ten statements across six files, and unlike the test-comment residue it describes *behaviour*, not history.

The fix is small and mechanical: rewrite each to the shape `auto-code-review.sh` already models — the marker is a **convention** that keeps a real review distinguishable from self-review and suppresses a spurious advisory warning; it is not what unblocks the write, and the hook warns rather than blocks since #1026 (AgDR-0109/0111). `pr-workflow.md` § "Mechanical backstop" needs a real rewrite (its whole framing is #843-era), not a word swap; the other five are one or two sentences each.

#### F2 — Contradictory and stale per-case headers in `test_warn_review_marker_write.sh`

The label rewrite reached the `run_hook` strings but only half-reached the case headers, producing four statements that are self-contradictory on their face — this PR wrote them:

- L749 (case 35), L770 (case 36), L788-789 (case 37), L804 (case 38): `-> WARNS (advisory, #1026), exit 2.`

And two headers were left un-rewritten, now contradicting the assertion directly beneath them:

- L263 (case 1): `→ BLOCKED, exit 2` above `… 0 "WARNING"`
- L826 (case 39): `-> BLOCKED, exit 2. Regression guard:` above `… 0 "WARNING"`

The preamble's `read "BLOCKED, exit 2" as "detected → warns, exit 0"` disclaimer is explicitly scoped to *the matrix below*, so it does not cover per-case headers in the body — and it cannot rescue `WARNS (advisory), exit 2`, which is contradictory under any reading. Case 33 (L716) is the model to follow: its header still says `BLOCKED, exit 2` but an in-function `NOTE:` corrects it and explains what the case now pins.

No assertion is wrong — I confirmed zero remaining expected-exit-2 assertions in the file, and the hook has exactly one `exit 0` terminal path. This is comment-only. I would not block on F2 alone; I am listing it because it is the same half-applied-substitution shape as F1 and should be swept in the same pass.

### Suggestions

- **S1 (non-blocking): annotate AgDR-0109 § Decision parts 2 and 3 the way its Consequences bullets now are.** Part 2 still asserts "A build agent writing a literal marker path is still blocked" with no local marker, and part 3 activates a meta-gate that was not built. AgDR-0111 declares both superseded and AgDR-0109's own `## Changelog` records it, so a reader who starts at the top is fine — but a reader who lands mid-document on the Decision section (the likeliest landing spot, since it holds the fail-closed reconciliation argument) gets an unmarked stale claim. Two more blockquotes, same form as the Consequences ones. Consider also a `superseded_by:` frontmatter field and a title-line note, since the H1 still advertises "activate the deferred meta-gate".
- **S2 (nit): `docs/rule-audit.md:25`** reads "Known evasions are documented limits" — plural, where the corrected position is exactly one. It is generic guidance for the backstop *class* so it is defensible, but given how much of this PR is about that count, the singular reads cleaner.

### What I verified did not move

- `block-unreviewed-merge.sh` takes only a `CLASS: CONTROL` header comment — no logic change. The control property holds: marker SHA vs forge-reported HEAD, both merge shapes gated, structured CEO fields required.
- The four control hooks carry `CLASS: CONTROL`; the backstop carries `CLASS: BACKSTOP — NOT THE CONTROL`. `rule-audit.md`'s matrix matches the labels in the files.
- `auto-code-review.sh`: all three "will BLOCK anyway" assertions removed; no residue anywhere in the tree (the only remaining hit is AgDR-0111 quoting the removed text).
- CI green at this HEAD (6 SUCCESS, 1 NEUTRAL).

**Out of scope, not reopened:** the advisory direction, the absent CI meta-gate, and the "does the protected action exist vs is it authorised" question (now partly addressed by #1052, which I did not review).

### Verdict

**CHANGES REQUESTED** — on F1 only. Everything round 2 raised is genuinely closed, and I found nothing wrong with how it was closed. F1 is a different stale claim that the propagation sweep surfaced, and it lands in an always-loaded rule file and in the two reviewer-agent definitions. Fix F1 (and sweep F2 in the same pass), and I will approve; I have no other reservations about this change.

---
📌 Reviewed commit: `278513a876dead5ffdcfe1efaf892abfbfdd7155`

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1049 — round 4

**Commit**: `813ca74702fc3cb08a335fde42d443121c9afb66`

### Summary

Round 3 blocked on ten stale "this hook blocks" statements across six untouched files. The sweep this round is **substantially done** — `code-reviewer.md`, `solution-architect.md`, all three review SKILL.md files, `auto-code-review.sh`, AgDR-0109's Decision parts 2/3, and `rule-audit.md` are all correct now, and the agent-file rewrites are well judged. CI is fully green (hook suite 2m29s pass, shellcheck, markdownlint, Analyze).

But the sweep is **not complete**, and the two misses are both in `.claude/rules/pr-workflow.md` — the one file that is `@`-imported by `CLAUDE.md` into every session, and the file whose § "Mechanical backstop" this round rewrote. One is a surviving copy of the exact round-3 claim; the other is a **new** factual error introduced by the rewrite itself.

### Checklist Results

- ❌ Code Quality (doc accuracy): **Fail** — see Issues 1 and 2
- ✅ Testing: Pass — 119/119, adversarial false-positive cases added
- ✅ Security: Pass — the advisory direction is out of scope; nothing regressed
- ✅ Performance: N/A
- ⚠ Summary Bullet Narrative: Pass — bold-led, each states what + why
- ✅ Technical Decisions (AgDR): Pass — AgDR-0111 correctly filed as a new record
- ✅ Adopter Handbooks: N/A — no findings (shell + markdown diff)

### Issues Found

#### 1. ⛔ BLOCKER — `pr-workflow.md:108` still asserts the removed mechanism, and cites the section that contradicts it

§ "Build agents cannot self-review" § Rule, first bullet:

> — the first, third, and fourth are **mechanically blocked** without an active-reviewer marker (see "Mechanical backstop" below)

Eleven lines below, the section it points at now says:

> **`warn-review-marker-write.sh` is ADVISORY — it warns, it does not block.** … it does not "unblock" anything, because nothing is blocked.

This is the round-3 finding, in the highest-stakes file, and it survived the rewrite of the very section it cross-references. It is worse than the round-3 instances in one respect: the two statements are 11 lines apart in one file, so the document now refutes itself, and the false half is the one a reader hits first (the normative MUST NOT list) while the true half is in the explanatory subsection below.

This matters for the same reason you accepted in round 3: `pr-workflow.md` is `@`-imported by `CLAUDE.md`, so every session is told a non-existent mechanism exists.

**Fix:** rewrite the bullet to drop the mechanism claim and keep the prohibition, e.g. *"— none of these are mechanically blocked (see 'Mechanical backstop' below); the prohibition is a rule, not a gate."* That also makes the bullet consistent with backstop item 2's framing, which already says *"a build agent is told plainly not to write these files, and that nothing will stop it, which is exactly why the instruction matters"* — good framing that the Rule bullet currently undercuts.

#### 2. ⛔ BLOCKER — `pr-workflow.md:132` documents a config key and an AgDR that do not exist

The new enforcement table's third row:

> \| `review_markers.require_posted_review` (opt-in, #1051) \| Asks the forge whether a review was actually **posted** at that commit … Off by default, GitHub-only. See [AgDR-0112](../../docs/agdr/AgDR-0112-verify-posted-review-at-head.md). \|

Verified at this HEAD:

- `review_markers.require_posted_review` occurs **exactly once in the entire tree — this table row.** It is not in `block-unreviewed-merge.sh`, not in `project-config.defaults.json`, nowhere. The knob does not exist.
- `docs/agdr/AgDR-0112-verify-posted-review-at-head.md` **does not exist.** The highest AgDR at this HEAD is 0111. The markdown link is broken.

You asked me specifically whether this row is honest about being opt-in and GitHub-only. It is — it says "opt-in", "Off by default", "GitHub-only". That part is fine. The problem is the tense and the framing around it: the row sits in a table headed **"What actually enforces this"**, in present tense, alongside two layers that genuinely exist today. Nothing marks it as unmerged. An adopter reading the session-imported rule will go looking for a config key that isn't there and click a link that 404s.

This is the same error class the PR itself argues against — *"A framework that misstates its own enforcement teaches agents to discount the rest of the banner."* The rewrite fixed that error for the hook and introduced a fresh one for #1052.

Note CI cannot catch this: `markdownlint-cli2` does not validate relative link targets.

**Fix:** either drop the row until #1052 merges, or mark it unambiguously as not-yet-shipped and remove the broken link — e.g. *"(lands in #1052 — not available at this commit)"*, with the AgDR reference as plain text rather than a link until the file exists. The following paragraph ("That last row is the successor to what AgDR-0062 deferred…") reads as a live capability too and needs the same tense treatment.

#### 3. Should-fix alongside — `test_block_unreviewed_merge.sh:667-679` section header, present tense, no decoder

This file *is* touched by the PR, and its per-case W1 comment was correctly updated ("Was BLOCKED/exit 2 from #843; returned to advisory in #1026"). But the section header 8 lines above it still reads:

```
# --- warn-review-marker-write.sh tests (#494, upgraded to BLOCKING for
# --- rex/security/architecture in #843) ---
…
# The other three now BLOCK (exit 2) unless a matching
# .claude/session/active-reviewer marker is present.
…
# this file only keeps the original #494 smoke assertions in sync with the
# new default behaviour.
```

"now BLOCK (exit 2)" is present tense and false, and it directly contradicts the corrected comment below it. Unlike `test_warn_review_marker_write.sh`, this file has **no decoder key**, so nothing disambiguates it. Lower stakes than 1 and 2 (a test comment, not agent context) but it's a three-line edit in a file you're already touching, and it's the eighth instance of the half-applied substitution pattern.

### Suggestions

- **nit:** `auto-code-review.sh:32` and the W1 comment in `test_block_unreviewed_merge.sh` both cite **AgDR-0109** for "warns, never blocks". That is AgDR-0111's departure — AgDR-0109 part 2 said the opposite ("block only on a confidently-resolved target") and is now annotated superseded. Citing 0109 points a reader at the superseded half. `pr-workflow.md` and the agent files get this right (they cite 0111); these two are inconsistent with them.
- **nit:** `test_warn_review_marker_write.sh` handles the matrix with a global decoder at line 38 — *"read 'BLOCKED, exit 2' as 'detected → warns, exit 0'"*. Defensible and cheaper than rewriting 20 index lines, but the per-case headers (`# (3) … → BLOCKED`, `# (9) Banner content — rex blocked case contains "BLOCKED"`) still read wrong at the point of use, ~280 lines from the key. Case 9's assertions are correct (`WARNING` + the retained `BUILD-CLASS SUB-AGENT` deterrent) — only its header is stale.

### What I verified and found clean

- **Agent-file rewrites are safe — they read as reason, not licence.** `code-reviewer.md:626` and `solution-architect.md:227` both name the absence of enforcement to a reader who is *not* the party tempted by it (Rex/Tariq, not a build agent — build agents never load these files), then explicitly foreclose the inference: *"do not treat the absence of a block as permission for anyone else to."* The justification correctly shifts from mechanism to legitimacy (*"Yours is the real review because a real, independent review actually happened; theirs would be the author grading their own work"*). Same shape I accepted for `auto-code-review.sh` in round 2, and it holds here. The three SKILL.md rewrites are the same pattern and equally safe.
- **AgDR-0109 supersession is coherent.** Decision parts 2 and 3 now carry inline superseded blocks matching the Consequences bullets, plus a Changelog entry. The blockquote summary at line 12 still states the original decision unannotated, but that is the standard AgDR convention — the record of what was decided *then*, with supersession marked in the body — so I am not flagging it.
- **No regressions from rounds 1-3.** `auto-code-review.sh`'s dual-reader banner is intact and correct; the four control hooks carry their `CLASS:` labels and stay fail-closed; `rule-audit.md` reads "the one known evasion" (singular) matching the corrected count; the build-agent deterrent paragraph is retained verbatim; `test_block_unreviewed_merge.sh` still pins rex/ceo SHA-mismatch → blocks and the `gh api` merge shape.
- **CI green:** hook suite pass (2m29s), shellcheck pass, markdownlint pass, Analyze pass, Verify Ticket ID pass.

### Verdict

**CHANGES REQUESTED**

Issues 1 and 2 are both in `.claude/rules/pr-workflow.md`. Issue 1 is a surviving copy of the round-3 defect in the session-imported file; issue 2 is a new factual error the rewrite introduced. Issue 3 is a cheap fix in a file already touched. Everything else about this round is right — the agent rewrites in particular are well judged, and the sweep covered 6 of 7 files correctly.

---
📌 Reviewed commit: `813ca74702fc3cb08a335fde42d443121c9afb66`

atlas-apex (commented):
## Code Review: PR #1049 — round 5

**Commit**: `cfbda5cf63741406c3bf42463241965f812696b4`

Tightly scoped to the four items round 4 raised. All four are fixed. Nothing blocking remains.

### 1. Forward references — clean

Grepped the **whole tree** at this HEAD, not just the diff:

- `require_posted_review` — 0 hits.
- `AgDR-0112` — 0 hits.
- Every AgDR ID cited in a changed file (0001, 0019, 0050, 0053, 0056, 0060, 0062, 0069, 0075, 0104, 0109, 0110, 0111) resolves to a file that exists at this SHA.
- Every hook/lib named in the changed prose (`clear-active-reviewer-marker.sh`, `_lib-detect-bash-write.sh`, `_lib-review-markers.sh`, `_lib-tracker.sh`, `block-unreviewed-merge.sh`) exists.
- Every relative markdown link in the changed docs was resolved against the tree object at this SHA. Zero misses.

The enforcement table row is gone and the replacement sentence — "analysed in AgDR-0062 and tracked in me2resh/apexyard#1051. It is **not shipped here**; do not assume it is active" — cites only things that exist. Good fix, and the right one: deleting the row was better than softening it.

### 2. `pr-workflow.md` no longer contradicts itself — verified

Read the whole file at HEAD, not just the hunks. The three places that could disagree now agree:

- MUST NOT list (L108): "**Nothing mechanically stops you** — `warn-review-marker-write.sh` warns and exits 0 … That is why this is a MUST NOT rather than a can't."
- § Mechanical backstop (L119, L121, L122): advisory, exits 0, "it does not 'unblock' anything, because nothing is blocked."
- § Mechanical enforcement (L137+) is about `block-unreviewed-merge.sh`, which genuinely does block — no conflict.

L108's framing is the strongest version of this: it converts the removed mechanism into the *reason* the norm matters, rather than leaving a norm with no visible backing.

### 3. Relative link depth — verified specifically

`.claude/rules/pr-workflow.md` L119 → `../../docs/agdr/AgDR-0111-marker-gate-plain-advisory.md`. From `.claude/rules/`, `../../` lands at repo root; `docs/agdr/AgDR-0111-marker-gate-plain-advisory.md` exists at this SHA (added by this PR). Depth is correct, not merely target-exists. Same check passed for all other relative links in the changed docs.

### 4. Attribution — correct, with one nit

AgDR-0111 now owns "warns, never blocks" everywhere it is asserted: `warn-review-marker-write.sh` (L268-269, L298), `auto-code-review.sh` (header + both runtime banner branches), both test files, both agent specs, all three review SKILL.md files. AgDR-0109 is cited only for part 1 (the backstop label, `§1` in the hook header) and for the jurisdiction argument it does work through at length. AgDR-0109's own document carries superseded notes on Decision parts 2 and 3 and on the two affected Consequences bullets, plus a changelog explaining why a new record rather than an entry. The `CLASS: CONTROL` citations on the four merge-gate hooks are now backed by AgDR-0109's changelog, which records that labelling as shipped.

### 5. Regressions from rounds 1-4 — none

- The four control hooks are comment-only (+6/-0 each); no logic moved.
- `test_block_unreviewed_merge.sh` header now reads "ALL roles are advisory … it does not unblock anything, because nothing is blocked" — the should-fix is done, and case W1 asserts exit 0 / WARNING to match.
- Ran both touched test files at this SHA: `test_warn_review_marker_write.sh` **47/47 PASS**, `test_block_unreviewed_merge.sh` **39/39 PASS**. CI's `hook test suite` also passed at this commit.
- The property that must not move is intact: the merge gate still refuses a marker SHA that does not match forge-reported HEAD.

### Issues Found

None blocking.

### Suggestions

Two `nit:` items, both non-blocking. Neither justifies a sixth round — fold them into a later touch of these files if you touch them at all.

**nit — `docs/rule-audit.md` § "Trust-chain hooks: control vs backstop" never cites AgDR-0111.** The section states the advisory guidance ("Warns, never blocks") and attributes it to AgDR-0104 and AgDR-0109: *"[AgDR-0109] records the evidence and the resulting return to advisory."* AgDR-0109 chose the opposite (block on a resolved target); AgDR-0111 is the record that returned it to advisory. Mitigated by AgDR-0109 now carrying superseded notes that route the reader onward in one hop — but `rule-audit.md` is the framework's citation index, so this is the one place the extra hop costs most. One clause would fix it.

**nit — two leftover "BLOCKED, exit 0" comment lines** in `test_warn_review_marker_write.sh` (L717 case 33, L827 case 39). Self-contradictory on their face; artifacts of the mechanical BLOCKED→advisory rewrite. Both are disambiguated by the preamble's "read 'BLOCKED, exit 2' as 'detected → warns, exit 0'" instruction — but these two lines say `exit 0`, so the preamble's mapping doesn't apply to them. Case 33's own NOTE block below it is correct and complete, so the harm is confined to the header line.

### Verdict

**APPROVED**

Five rounds is enough. The substance is right: the forward reference is gone, the self-refutation is gone, the attribution is correct, and the load-bearing merge-gate property is provably unmoved. The two remaining nits are cosmetic and do not warrant another round.

---
📌 Reviewed commit: `cfbda5cf63741406c3bf42463241965f812696b4`

atlas-apex (commented):
## Code Review: PR #1049 — delta confirmation (round 6)

**Commit**: `fc4ed2b39768064a60d23d19e5089f2f8bd4301a`

This is a **delta-only confirmation** on top of my round-5 APPROVED at `cfbda5cf`. Scope reviewed: `git diff cfbda5cf..fc4ed2b` — 2 files, +5/−3. The rest of the PR is unchanged and already approved; I did not re-review it.

### Both round-5 nits are fixed, correctly

**Nit 1 — `docs/rule-audit.md` citation.** The control-vs-backstop paragraph now credits AgDR-0111 as the record that owns the return to plain advisory, alongside 0109's evidence + backstop label. Verified factually accurate at this HEAD:

- AgDR-0111's summary reads "ship its rejected option B — plain advisory, exit 0 always" — it does own the decision.
- AgDR-0109's own title is "…block only on a resolved target…", and its Decision §2, Consequences, and meta-gate items all carry explicit `Superseded 2026-07-28 (AgDR-0111)` notes. So "0109 had chosen a narrower block-on-resolved-target design, superseded" matches what 0109 actually says — including that 0109 is the superseded party, not the superseding one.
- Link depth checked, not just existence: `docs/rule-audit.md` + relative `agdr/AgDR-0111-marker-gate-plain-advisory.md` → `docs/agdr/AgDR-0111-marker-gate-plain-advisory.md`, which `git ls-tree fc4ed2b` confirms exists. Same one-hop shape as the already-working 0104/0109 links on lines 19 and 36. Resolves.

**Nit 2 — self-contradictory test comments.** L717 and L827 now read `DETECTED -> warns, exit 0`, matching their assertions (`run_hook … 0 "WARNING"`). Swept the whole file: zero remaining `BLOCKED.*exit 0` lines. Every surviving `BLOCKED` mention is the `BLOCKED, exit 2` matrix form that the preamble at L19/L39 explicitly maps to "detected → warns, exit 0" — the case I accepted in round 5.

### Verification

- Test suite at this HEAD in an isolated worktree: **47 PASS / 0 FAIL**.
- Diff confined to the two nit fixes; nothing else touched, nothing approved at `cfbda5cf` regressed.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `fc4ed2b39768064a60d23d19e5089f2f8bd4301a`

Issue #1051, opened 2026-07-28
[Feature] Opt-in merge gate: verify a code review was actually posted at the PR's HEAD


atlas-apex:
## User Story

As a maintainer approving merges, I want the merge gate to confirm a review was genuinely posted to the PR at the commit being merged, so that "Rex reviewed this" is verified on the forge rather than inferred from a local file an agent can write.

## Context

The merge gate requires a `*-rex.approved` marker whose SHA matches the PR's forge-reported HEAD. The SHA half is strong — it comes from the forge. The *existence* half is not: the marker is a local file, and once the marker-write backstop became advisory (AgDR-0111), nothing mechanically prevents an agent writing it. A forged marker with the correct SHA passes the gate, and the human approving the merge sees "Rex approved" with no review comment on the PR to read.

AgDR-0062 considered this control and **deferred** it as "unsatisfiable in a single-account setup — GitHub refuses to let an account approve its own PR." That is true of the `APPROVED` review state and **only** of it. A `COMMENTED` review from the PR's own author is accepted by GitHub and returned by the API with a `commit_id` — verified against a live same-account PR. And the canonical reviewer flow already posts exactly that shape (AgDR-0075 uses `--comment` precisely because `--approve` is unavailable single-account), so this asks adopters for nothing new.

## Acceptance Criteria

- [ ] `tracker_review_at_sha <repo> <pr> <sha>` returns a four-way result so callers can distinguish "checked, nothing there" from "could not check": `0` review found, `1` none at that SHA, `2` query failed, `3` forge unverifiable
- [ ] `block-unreviewed-merge.sh` blocks on `1` and on `2` (fail closed, per AgDR-0104), and skips on `3` with a visible warning rather than bricking that adopter
- [ ] Gated behind `review_markers.require_posted_review`, **default `false`** — enabling it is the only thing that changes behaviour for an existing adopter
- [ ] A review posted against an *older* commit does not satisfy the check
- [ ] GitHub-only support is stated plainly rather than faked; `glab` / `custom` / `none` skip
- [ ] Tests cover both directions, including that the default is a total no-op even with zero reviews posted
- [ ] An AgDR records that this narrows AgDR-0062's deferral to the `APPROVED` state only

## Design Notes

[...]

PR #1052, opened 2026-07-28
feat(#1051): opt-in merge gate on a review actually posted at HEAD

atlas-apex:
## Summary

- **The merge gate can now verify a review actually happened.** Its evidence today is a `*-rex.approved` file. The SHA half is strong — it comes from the forge — but the *existence* half is a local file an agent can write, and #1049 (now merged) removed the write guard that never worked. This asks the forge directly: was a review posted to this PR, at this exact commit?
- **AgDR-0062 deferred a control like this as "unsatisfiable single-account." That is true of its version, and this is not its version.** Its option required a review *by an independent reviewer (not the PR author)*. This ships a deliberately **weaker** check that drops independence — which is exactly why it is satisfiable. It is **not separation of duties** and says so in the AgDR, the config comment, and the lib header.
- **Four-way contract so "couldn't check" can never become "checked, fine":** `0` found · `1` absent · `2` query failed · `3` forge unverifiable. Blocks on `1` and `2` (fail closed per AgDR-0104); skips on `3` with a warning that the flag is a **no-op** on that forge rather than implying protection.
- **Off by default.** Enabling it changes merge behaviour, and an adopter whose reviews predate the check would be blocked on their next merge.
- **GitHub-only, stated rather than faked.** GitLab exposes MR approvals and diff versions, not SHA-stamped submissions; mapping those without a live MR to verify against would repeat the unverified-claim failure this framework keeps paying for.

## What it closes, and what it doesn't

Closes: *no review happened at all* — the forged-marker case that actually occurred in #843.
Does **not** close: a determined agent. One with `gh` auth can post a stub `--comment` review at HEAD, or disable the flag by editing `.claude/project-config.json` (untracked by design, so no git trace). Treat this as an **integrity and auditability check, not an anti-adversary barrier** — it catches accidents and stale reviews, and forces an artifact a human can read.

## Review history — six fail-open holes found and closed

Rex (3 rounds) and Hakim (security, #777 trust-chain trigger) both approved. Between them they found six ways this could have shipped failing open:

| Found by | Hole | Fixed |
|---|---|---|
| Rex | Missing `_lib-tracker.sh` silently **allowed** the merge | Blocks, names the file |
| Rex | **Test suite was tautological** — swapping two lines to turn every failed query into "review found" survived 8/8 | Shim now emits a 404 body like real `gh api`; that mutation now fails exactly one case |
| Hakim | **The flag read itself failed open** — missing config reader left the control silently not running while config said `true` | Grep fallback + warns whenever the reader is degraded |
| Hakim | **jq-filter injection** on `${sha}` — a crafted value could rewrite the filter to always match | Hex + length validated; traversal guard on `repo` |
| Hakim | `rc=3` skip looked like it contradicted the fail-closed doctrine | Kept (jurisdiction ≠ precondition failure, per AgDR-0109) but the reasoning is now written down |
| Rex | Truncated lib sources OK but leaves the function undefined | Fails closed with a correct diagnosis |

Every fix is **mutation-verified**: revert it and the suite drops by exactly the case that names it.

## Testing

- `test_require_posted_review.sh` → **13/13**, both directions. The load-bearing case is *flag off + zero reviews → allowed*, which protects every existing adopter.
- Full suite **120/120**; `shellcheck -S warning` and `markdownlint-cli2` clean.
- Rebased onto `dev` after #1049 merged — 5 files, `MERGEABLE`, CI green.

## Reviewer notes

Change to a **CONTROL** (`block-unreviewed-merge.sh`), so Heavy tier. `.claude/rules/pr-workflow.md` is updated in this PR because a sentence I wrote in #1049 ("not shipped here; do not assume it is active") becomes false the moment this merges, and that file is `@`-imported into every session.

Refs #1051

## Glossary

| Term | Definition |
|------|------------|
| Posted review | A review submitted to the forge and visible on the PR, as opposed to a local approval marker file |
| `commit_id` | The commit a review was submitted against — what binds a review to a specific HEAD |
| Fail closed | When a gate cannot evaluate its precondition, it denies rather than allows |
| Control vs backstop | AgDR-0104's distinction — a control decides on structured state and must fail closed; a backstop is best-effort and warns |
| Separation of duties | A control requiring the approver to be a *different identity* from the author. This PR does **not** provide it |
| Mutation testing | Deliberately breaking a fix to confirm a test catches it — proof the test is load-bearing, not decorative |


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1052

**Commit**: `e1532f7314583fbd5bf7f1c2496114e5b909c09f`

Scope: **only** the newest commit `e1532f73` (5 files). #1049's commits were not re-reviewed.

### Summary

Adds `tracker_review_at_sha` and wires it into `block-unreviewed-merge.sh` behind `review_markers.require_posted_review` (default `false`), so the gate can ask the forge whether a review was *actually posted* at the marker's commit rather than trusting a local file.

I treated this as a change to a **CONTROL** and verified every claim against the live API and the source, not the PR body.

### Checklist Results
- Testing:                    Pass (with one substantive gap — see below)
- Performance:                Pass (one extra forge round-trip, only when enabled)
- Technical Decisions (AgDR): Pass (AgDR-0112; one accuracy nit)
- Adopter Handbooks:          Pass
- Fallow Static Analysis:     N/A (no JS/TS)

---

### 1. The premise — VERIFIED

I checked it myself rather than taking the PR's word:

```
$ gh api repos/me2resh/apexyard/pulls/1049/reviews \
    --jq '.[] | "\(.state) \(.commit_id) \(.user.login)"'
COMMENTED 295f41d6... atlas-apex
COMMENTED ee1b27fc... atlas-apex
...
$ gh api repos/me2resh/apexyard/pulls/1049 --jq .user.login
atlas-apex
```

PR author and reviewer are the same account, and the `COMMENTED` reviews come back with a populated `commit_id`. The premise holds. AgDR-0062's objection genuinely does not bind this shape.

### 2. The four-way contract — CORRECT (all paths exercised against the real API)

| path | verified how | result |
|---|---|---|
| 0 found | review at sha | `rc=0`, id returned |
| 1 absent | query for a sha with no review | gh `rc=0`, **empty** stdout → `return 1` |
| 2 query failed | `gh api` on a nonexistent repo | gh `rc=1` → `return 2` |
| 3 unverifiable | `tracker_kind` != `gh` | `return 3` |

**The ordering here is load-bearing and the code gets it right.** On a 404, `gh api` prints the error body **to stdout**:

```
rc=1 out=[{"message":"Not Found",...,"status":"404"}]
```

Because `[ "$rc" -ne 0 ] && return 2` runs *before* `[ -n "$out" ] && return 0`, that body can't be misread as "a review was found". Reverse those two lines and every query failure returns 0 = verified — the control becomes a rubber stamp. Worth calling out explicitly since it's the exact "couldn't check → checked, fine" inversion that would be the worst outcome here.

Guard clauses (empty repo/pr/sha, non-numeric pr) all return **2**, not 0. Fail-closed. Correct.

### 3. The wiring — CORRECT

- `HOOK_DIR` defined line 212, used line 286 — in scope.
- `config_get_or` sourced line 131 with `|| true`, guarded by `command -v` at 282, and `REQUIRE_POSTED_REVIEW` is pre-initialised to `false` at 281. Lib unavailable → stays false → no-op.
- Default genuinely is a total no-op, including with `jq` missing or the config file absent (`config_get_or` returns the `false` fallback).
- Placement after the Rex existence + SHA checks and before the CEO check is right — it strengthens the Rex evidence, and both must pass regardless.
- Edge case I checked: if `CURRENT_SHA` is empty (both `gh` and local `git` fail), it's passed through and returns **2** → block. Fail-closed.

### 4. `--paginate` correctness — VERIFIED, no miss on later pages

Forced 4 pages with `per_page=2` on PR #1049 and queried for a review on the **last** page:

```
rc=0 out=[4799889111]
```

`gh api --paginate --jq` applies the filter per page and streams results, so a match beyond the 30-item default page is found. No bug here.

### 5. The rc=3 skip path — honest enough

The warning prints to stderr on *every* merge attempt, the config comment says GitHub-only, and AgDR-0112 states it plainly. It isn't silent. Failing closed instead would brick a GitLab adopter who opted in, with no route forward but disabling the flag. Reasonable call as made.

### Issues Found

**1. `suggestion:` The test shim is tautological w.r.t. query correctness — and can't see the one fail-open that matters.**

I mutation-tested the suite. Three mutations all survive **8/8**:

| mutation | real-world effect | suite result |
|---|---|---|
| `.commit_id` → `.bogus_field` | never matches → blocks everything | 8/8 PASS |
| `/pulls/` → `/pullz/` | 404 → rc 2 → blocks | 8/8 PASS |
| **swap the `rc`/`out` check order** | **404 body on stdout read as "found" → control passes everything** | **8/8 PASS** |

Root cause: the shim's failure branch is `exit 1` with **no stdout**, but real `gh api` emits the error body to stdout on failure. That's exactly where the fail-open lives, so the shim can't model it.

One-line fix, which I verified:

```bash
[ "$st" = "FAIL" ] && { echo '{"message":"Not Found","status":"404"}'; exit 1; }
```

With that, the correct code still passes 8/8 **and** the ordering mutation is caught — 7/8, failing precisely `flag ON + forge query FAILS -> BLOCKED (fail closed, not open)`.

Non-blocking: the code is correct today, and mutations 1–2 both fail *closed* (and I verified the query shape live in this review). But for a CONTROL, this is the net that protects that correctness against the next refactor — cheap to add now.

**2. `suggestion:` Missing `_lib-tracker.sh` fails OPEN while the flag is on.**

```bash
if [ "$REQUIRE_POSTED_REVIEW" = "true" ]; then
  if [ -f "$HOOK_DIR/_lib-tracker.sh" ]; then
    ...
  fi          # <-- no else
fi
```

Flag ON + lib absent → check silently skipped, merge proceeds. That is a "couldn't check" outcome failing **open**, against this AgDR's own fail-closed commitment. Narrow (the lib ships beside the hook; absence means a broken install) but it's the single place the stated contract isn't honoured. An `else` that warns (consistent with rc=3) or blocks (consistent with rc=2) would close it.

Lower-severity sibling: if `jq` disappears *after* an adopter enables the flag, `config_get_or` returns the `false` fallback and the control silently disables itself. Defensible, but worth a sentence in the config comment so it isn't a surprise.

**3. `nit:` AgDR-0112 slightly mischaracterises what AgDR-0062 deferred.**

AgDR-0062's deferred option was *"require at least one review at the PR HEAD SHA **by an independent reviewer (not the PR author)**"*, and its blocker was **author-independence** — not the `APPROVED` state as such. AgDR-0112 says that reasoning "is correct for the `APPROVED` review state, and only for it."

More precisely: this PR ships a **weaker** control that *drops* the independence requirement — any review from anyone, including the author, satisfies it. That is satisfiable single-account and does close the forged-marker case, so the decision is sound. The Consequences section already discloses the substance honestly ("closes 'no review happened at all', not 'the review was any good'"). It's the headline framing that reads as "we solved what 0062 deferred" rather than "we shipped a different, weaker control 0062 didn't evaluate." Given what over-stated claims have recently cost on this repo, one tightened sentence is worth it.

### Handbook Findings

- **Commit Message Quality** (advisory) — `handbooks/general/commit-message-quality.md`: no violation. `e1532f73`'s message is a model of the WHY-not-WHAT rule — it states the gap, why AgDR-0062's deferral doesn't bind, and why the default is off.
- **Clean Architecture Layers** / **Migration Safety** (blocking) — N/A; no application layers and no migration content in this diff.

### Verification Performed

- Premise checked against live GitHub API (not the PR body)
- All four rc paths exercised against the real API, including the 404-body-on-stdout case
- Pagination verified with a forced 4-page result
- New suite: **8/8**; regression `test_block_unreviewed_merge.sh`: **39/39**
- Mutation testing: 3 mutations, all survivors, one fix verified
- AgDR cross-refs to 0062 / 0075 / 0104 / 0111 read and checked; no AgDR-0112 number collision
- Secrets scan: clean

### Verdict

**APPROVED**

The control is correct on every path I could test — including the ordering detail that would have silently turned it into a rubber stamp. Default-off makes it a true no-op for existing adopters, verified. The three findings are a test-net gap and two doc/robustness nits; none makes the shipped behaviour wrong, so I'm not holding the merge for them. Finding #1 is the one I'd most like to see land, ideally in a follow-up — it's one line and it protects a CONTROL.

---
Reviewed commit: `e1532f7314583fbd5bf7f1c2496114e5b909c09f`

atlas-apex (commented):
## Security Review: PR #1052

**Commit**: `f8a219fed34485c9f83229671440e39ae7ed7f29`
**Scope**: commits `e1532f73` + `f8a219fe` only (`_lib-tracker.sh`, `block-unreviewed-merge.sh`, `test_require_posted_review.sh`, `project-config.defaults.json`, `AgDR-0112`). #1049 excluded — reviewed separately.
**Trigger**: auto-fire under #777 — trust chain (`.claude/hooks/**`), CLASS: CONTROL. Heavy tier.

### Summary

This adds `tracker_review_at_sha` and wires it into the merge gate behind `review_markers.require_posted_review` (default `false`). I verified the mechanism it depends on against this PR's own review list: the API returns `{"state":"COMMENTED","commit_id":"e1532f73…","user":"atlas-apex"}` — the PR author's own review, SHA-stamped. AgDR-0112's central factual claim is correct, and I re-ran every caller path in a sandbox rather than trusting the PR body.

The engineering is sound and the default-off claim holds. My findings are about **what the control is worth**, and about **two fail-open paths** — one of which is the direct sibling of the hole `f8a219fe` was written to close.

### Checklist Results

| Area | Result |
|---|---|
| Secrets & credentials | Pass — none introduced |
| Injection prevention | **Fail** — jq-filter injection on `${sha}`, fails OPEN (M2) |
| XSS | N/A |
| Auth & authorisation | Pass — no authz surface |
| Data protection | Pass |
| API security | Pass |
| **Gate & trust-chain integrity (#777)** | **Partial** — two fail-open paths in an opt-in control (M1, M3); no regression to existing behaviour |

---

## 1. Threat model — what this actually stops

Adversary: a build-class or confused agent that wants the gate to pass without a real review.

**What it stops:**

- The *accidental* forged marker — the #843 failure mode where a banner told a build agent to "invoke Rex", it could not nest the Agent tool, and it resolved the contradiction by writing the marker itself. Such an agent writes a file; it does not think to post a review. This is the case that actually happened, twice.
- The *stale* review — a review posted at an earlier commit no longer counts. Genuinely useful, and independent of any adversary.
- The *silent-failure* case — reviewer writes the marker, the post fails on network. Today that merges with nobody having seen a review.

**What it does not stop — and this needs saying plainly:**

An agent with Bash and `gh` auth satisfies this gate with one command: post a `--comment` review at HEAD. GitHub accepts it from the PR's own author (that is precisely the property this feature is built on), and the check accepts any review state. So against a *deliberate* adversary the control adds one command, not a barrier.

**And there is a cheaper bypass than that, which no artifact in this PR mentions.** The control's own enablement — `review_markers.require_posted_review` and `tracker.kind` — is read from `.claude/project-config.json`, a file that is **untracked by design** and that the same agent can edit. Setting the flag to `false`, or `tracker.kind` to `none` (M3), disables the control in one line and leaves **no git trace**. Posting a stub review at least leaves a public, timestamped, attributable artifact; editing the config leaves nothing. The cheapest bypass is also the most invisible one.

**My assessment:** this is an **integrity and auditability check, not an anti-adversary control**. It forces a public forge artifact into existence, so the human giving the merge nod has something to *read* instead of a filename to trust. That is real value and worth shipping. It should not be leaned on as a barrier against an agent that wants around it — and AgDR-0112 comes close to saying that, but not close enough (L1).

---

## 2. Fail-closed discipline

I ran every caller path in a sandbox against the real hook and the real lib. `rc=0` proceeds; `rc=1` and `rc=2` block with correct, distinct messages; `rc=3` warns and proceeds; missing `_lib-tracker.sh` blocks. The `f8a219fe` fix works as described. A truncated lib leaves the function undefined → `$?=127` → falls to `*)` → blocks (L3 on the message). I also checked that `_lib-tracker.sh` has no function-name collisions with the four libs sourced earlier and no top-level side effects, rather than assuming it.

But the fix is incomplete in one direction:

### M1 — MEDIUM: the flag read itself fails open

`_lib-read-config.sh` is sourced at line 131 with `|| true`. If it is missing or unsourceable, `config_get_or` is undefined, `REQUIRE_POSTED_REVIEW` stays `false`, and the entire control is **silently skipped** — with the operator's config on disk still saying `true`.

Reproduced:

```
--- config lib PRESENT ---   REQUIRE_POSTED_REVIEW=true   -> control ENFORCED
--- config lib MISSING ---   REQUIRE_POSTED_REVIEW=false  -> control SKIPPED
```

This is the same shape as the hole `f8a219fe` closed — "could not check" masquerading as "checked, fine" — and the treatment is **asymmetric**: a partial install that drops `_lib-tracker.sh` blocks every merge with a loud, explicit message; a partial install that drops `_lib-read-config.sh` disables the control with no message at all. The block message for the first case even names "a partial install or a bad sync" as the likely cause — the exact scenario that silently defeats the second.

The impact is *operational*, not adversarial: an agent that wants out has cheaper routes. The problem is that an operator who opts in has **no way to know the control is running**. A control that can silently be not-running is one you cannot depend on.

The `[ ! -f ]`-style fail-closed used for `_lib-tracker.sh` is wrong here — blocking whenever the config lib is missing would block every merge for the default-off majority. A raw fallback read is the right shape:

```bash
if ! command -v config_get_or >/dev/null 2>&1; then
  if grep -qE '"require_posted_review"[[:space:]]*:[[:space:]]*true' \
       "$MARKER_HOME/.claude/project-config.json" \
       "$MARKER_HOME/.claude/project-config.defaults.json" 2>/dev/null; then
    echo "BLOCKED: require_posted_review appears enabled but the config library is unavailable — cannot evaluate this control's precondition." >&2
    exit 2
  fi
fi
```

**jq absence is already covered** — I checked before raising it. With jq gone the command parse at line 84 yields empty, the raw-payload `is_merge_command` fallback fires, and merge-shaped payloads exit 2 (#965). That sibling path is masked by an existing guard. Only the config-lib path is live.

### M3 — MEDIUM: the rc=3 skip contradicts the hook's own doctrine

An adopter who sets `require_posted_review: true` on GitLab gets **zero protection** and a one-line stderr warning per merge. The hook's own header says: *"if it cannot evaluate its precondition it must block, never allow."* rc=2 obeys that; rc=3 does not. Both are "could not check"; only one blocks.

The counter-argument in the AgDR — that rc=3 is deterministic, so blocking bricks those adopters — is fair, but it argues for a *different* fix, not for warn-and-proceed. The operator explicitly asked for verification; silently downgrading their control to a warning is not what they asked for. Better: refuse to enable on an unsupported forge (block with "this control is GitHub-only; set the flag to false"), so the failure lands at config time as one clear error rather than as a per-merge warning that becomes wallpaper.

This is sharpened by the bypass note above: `tracker_kind` is read from the same agent-writable config, so `"kind": "none"` is a one-line, git-invisible way to turn a blocking control into a warning. That path should not exist.

---

## 3. Injection / argument handling

### M2 — MEDIUM: jq-filter injection on `${sha}`, failing OPEN

```bash
--jq ".[] | select(.commit_id == \"${sha}\") | .id"
```

`${sha}` is interpolated into a jq string literal with no validation. A value containing a double quote breaks out. Reproduced against the real function, with a fixture holding one review at an *old* commit only:

```
honest HEAD sha   rc=1   (correct: no review at HEAD)
INJECTED sha      rc=0   (fail-OPEN: reports "review found")
```

The direction matters: it fails **open**, not closed — the worst direction for a control.

**Not reachable from today's caller.** `CURRENT_SHA` comes from `resolve_pr_head` (forge-derived) or `git rev-parse HEAD`; neither is agent-controllable, and the marker SHAs must equal it, not the other way round. So this is not exploitable end-to-end in the current threat model.

I am still calling it MEDIUM, for a specific reason: `tracker_review_at_sha` is a **documented public API of a CONTROL library** with an advertised four-way contract, and it **already validates `pr` numerically** while leaving `sha` unvalidated. That asymmetry is the tell — the defensive instinct was there and stopped one argument short. Any future caller passing a less-vetted SHA gets a silent, fail-open wrong answer.

One line, matching the style already present:

```bash
case "$sha" in *[!0-9a-fA-F]*|"") return 2 ;; esac
```

### L4 — LOW: `repo` is unvalidated for shape

`$CMD_REPO` is `sed`-extracted (any non-whitespace token) and interpolated into the API path. No shell injection — it is an argv element to `gh`, never eval'd. I probed path traversal: `../../foo/bar` normalises to a 404 → `gh` exits non-zero → rc=2 → **block**. Fail-closed, so no exploit. But the same asymmetry as M2: `pr` is validated, `repo` is not. A `case "$repo" in */*) ;; *) return 2 ;; esac` shape check is free.

I also checked whether a crafted `--repo` could aim the *check* at one repo while the *merge* hits another. It cannot — the same `CMD_REPO` drives the marker paths, the HEAD resolution, and the merge target. No divergence is reachable.

---

## 4. TOCTOU / staleness

### L5 — LOW, informational, pre-existing

The check runs at PreToolUse; the merge runs after. A commit landing in between defeats *all three* SHA-pinned checks — Rex marker, human marker, and now this one — because the merge command takes whatever HEAD is current at merge time.

Not introduced here, and this PR inherits rather than worsens it. Worth recording because this feature's entire value proposition is "SHA-bound evidence": `--match-head-commit` / `expected_head_sha` on the merge call would make the whole family of SHA checks actually atomic. Separate ticket, not a change to this PR.

---

## 5. Does it weaken anything today?

**No — verified independently, not taken on the PR's word.**

- Defaults ship `require_posted_review: false`; `config_get_or` falls back to `false` on empty/null/error.
- jq absent → flag reads false → block skipped (and merges are separately blocked by the #965 guard).
- Config files missing → `false`.
- `config_get_or` undefined → `false` (that is M1, but it fails toward *today's* behaviour, not worse than it).
- Shallow-merge semantics: an adopter override that replaces `review_markers` drops the key → `false`.
- `shellcheck -S warning -x` clean on both changed shell files.
- No function-name collisions with the four libs sourced earlier; `_lib-tracker.sh` has no top-level side effects beyond three cache-variable initialisations.

With the flag off, the added code path is a single config read and a string compare. The no-op claim holds.

---

## 6. Honest naming (AgDR-0104 § 2)

The correction in `f8a219fe` is **accurate and complete across all three places** — I checked each:

- lib header: *"does not provide separation of duties and must not be described as if it did"*
- config comment: *"it is NOT separation of duties and must not be described as such"*
- AgDR-0112: *"does not deliver separation of duties, and must never be described as doing so"*

AgDR-0062's deferral is correctly characterised as standing unchanged, and the reason this version is satisfiable (independence dropped, not the APPROVED-state objection) is stated correctly. A real improvement over the prior framing.

### L1 — LOW: one residual overclaim, and one omission

**Overclaim.** Consequences, first bullet: *"The residual local-forgery exposure accepted in AgDR-0062/0109/0111 is **closed** for GitHub adopters who opt in."*

It is not closed. The same document's fourth bullet and Option C's own con both say a determined agent can post a fake review. "Closed" contradicts them. Suggested: *"…is **raised in cost and made auditable** — forging now requires a public, timestamped artifact on the forge instead of a silent local file write."* That is what actually ships, and it is worth shipping.

**Omission.** Nothing in the AgDR, the config comment, or the lib header notes that the control's **enable flag and `tracker.kind` live in an agent-writable, untracked local config**, so a determined agent disables the control more cheaply and more invisibly than defeating it. Given how much of this PR's value rests on being honest about limits, this belongs in Consequences. It does not change the decision — it changes what the reader believes they bought.

---

## 7. Test coverage

### L2 — LOW: the headline fix of the latest commit is untested

`f8a219fe`'s item 1 — missing `_lib-tracker.sh` now blocks — has **no case** in `test_require_posted_review.sh`. The commit message's "make the test able to catch it" refers to item 2 (the shim's FAIL branch), which is a genuinely good fix: emitting a 404 body on stdout while exiting non-zero is what makes the rc-before-stdout ordering testable, and the mutation result (8/8 → 7/8) proves the suite gained real power. Credit where due.

But item 1 shipped untested. In a codebase whose own AgDRs repeatedly cite unverified claims as a recurring cost, the fail-closed fix is exactly the thing that should have a case. I verified it manually and it works — a test would mean nobody has to take my word for it either. Two lines in `make_sandbox`: remove the lib, assert rc=2 and the message.

**Nit** (line 305): `if [ -f "$HOOK_DIR/_lib-tracker.sh" ]` immediately follows the `[ ! -f … ]` early-exit block, so it is always true. Harmless, but it reads as a leftover of the pre-fix shape and invites a future editor to think the else-path is still live.

**Nit**: the block-case assertions match on the string `NO`, a very loose regex that would pass on several unrelated messages. A longer fragment would bind the assertion to the message it means.

---

### Findings by severity

| Sev | Finding |
|---|---|
| CRITICAL | none |
| HIGH | none |
| MEDIUM | **M1** flag read fails open when `_lib-read-config.sh` is missing — sibling of the hole this commit fixed, opposite treatment |
| MEDIUM | **M2** jq-filter injection on `${sha}`, fails open; unreachable today, latent in a public control API that validates `pr` but not `sha` |
| MEDIUM | **M3** rc=3 skip contradicts the hook's own fail-closed doctrine; `tracker.kind: none` is a one-line invisible disable |
| LOW | **L1** AgDR says the forgery exposure is "closed" (it is raised in cost); enable-flag writability unstated |
| LOW | **L2** the missing-lib fail-closed fix ships with no test |
| LOW | **L3** rc=127 blocks correctly but prints the "forged marker" message, misattributing cause |
| LOW | **L4** `repo` unvalidated for `owner/repo` shape (no exploit path found; asymmetric with `pr`) |
| LOW | **L5** TOCTOU between PreToolUse and merge — pre-existing, affects all three SHA checks |

### Recommendations

Before anyone flips the flag on: **M1**. An opt-in control that can be silently not-running is not one you can depend on, and the four-line grep fallback closes it.

Cheap and worth taking now: **M2** and **L4** (two `case` guards), **L1** (two sentences), **L2** (one test case).

**M3** deserves a decision rather than a patch — warn-and-proceed versus refuse-to-enable is a posture call, and the `kind: none` bypass tilts it toward refuse-to-enable. Fine as a follow-up ticket.

**L5** is its own ticket and touches more than this PR.

### Verdict

**APPROVED**

The change does not weaken anything that exists today — I verified default-off independently across five degraded-environment cases. For an opted-in GitHub adopter it is a real improvement: it catches the accidental forged marker that actually happened twice, catches stale reviews, and forces an artifact a human can read. The honest-naming correction in `f8a219fe` is accurate and complete.

Approving with findings rather than blocking, because none of them regress current behaviour and the control is off by default. But I want the record unambiguous on two points: **M1 should land before the flag is turned on anywhere**, and this control should be described as an integrity and auditability check — not as something that stops an agent that wants around it. It does not, and one command is all it costs.

---
📌 Reviewed commit: `f8a219fed34485c9f83229671440e39ae7ed7f29`

atlas-apex (commented):
## Code Review (round 2): PR #1052

**Commit**: `2e4e46f8153c64b8dbdb0f1d09a6f9feba83e7a2`

### Summary
Round-2 review of the opt-in posted-review merge gate. Verified the three round-1 findings, the four Hakim fixes, and mutation-tested the suite rather than trusting the claims.

### Round-1 findings — all genuinely fixed

**#1 missing `_lib-tracker.sh` failed open** — fixed. `[ ! -f … ] → exit 2` with a message naming the file. Mutation-verified: deleting the block drops the suite to 11/12, failing exactly `missing _lib-tracker.sh + flag ON -> BLOCKS`.

**#2 tautological shim** — fixed, and your claim verified independently. I re-ran my own mutation (swapping the `rc`/`out` checks in `tracker_review_at_sha`) against the new shim: **11/12, failing precisely `flag ON + forge query FAILS -> BLOCKED (fail closed, not open)`**. The 404 body on stdout is what makes that detectable. Correct fix.

**#3 AgDR-0112 mischaracterised AgDR-0062** — fixed and accurate. I read 0062 directly: its blocker was author-independence being unsatisfiable single-account, exactly as now stated. The corrected framing is consistent across all three places (AgDR options table row B/C, `_comment_require_posted_review`, `_lib-tracker.sh` header § WHAT THIS IS NOT). The "raised in cost and made auditable — not closed" framing is honest — it names the residual (an agent with `gh` auth can post a stub review) rather than implying separation of duties.

### Hakim's fixes — no regressions introduced

Mutation-tested all four. Baseline 12/12; every reverted fix drops to 11/12, each caught by the case that names it — none decorative:

| Reverted fix | Failing case |
|---|---|
| rc/out order (round-1 #2) | `forge query FAILS -> BLOCKED` |
| M2 sha hex+length guard | `M2: injection guards` |
| M2 repo traversal guard | `M2: injection guards` |
| Missing-lib block | `missing _lib-tracker.sh -> BLOCKS` |
| M1 grep fallback | `M1: config reader missing + flag ON -> still BLOCKS` |
| rc=2 fail-closed | `forge query FAILS -> BLOCKED` |

**M1 grep fallback cannot produce a false `true` today** — I checked the concrete risk: the regex `"require_posted_review"[[:space:]]*:[[:space:]]*true` does *not* match the `_comment_require_posted_review` key (the opening quote precedes `_comment_`), and the defaults file ships `false`. The `M1 control` case pins the false-positive direction.

**M2 guards don't reject legitimate SHAs.** 7–40 hex covers everything `gh pr view --json headRefOid` emits (always 40) and git's minimum abbreviation. The range is right.

**rc=3 skip** — the written reasoning holds. "Couldn't check but should have been able to" (block) vs "no check exists here" (skip) is the same distinction AgDR-0109 drew, and the warning now says NO-OP rather than implying protection.

### Issues Found
None blocking.

- **nit:** `block-unreviewed-merge.sh:327` — the `if [ -f "$HOOK_DIR/_lib-tracker.sh" ]` is now dead code; the `[ ! -f … ] → exit 2` immediately above guarantees it. Harmless, but it reads as if the missing-lib case were still reachable. Dropping it makes the fail-closed intent clearer.
- **suggestion:** the M1 fallback's *false-negative* path is still silent. If the reader is unavailable **and** the grep misses (exotic JSON — `true` on its own line, a reformatted config), the control is silently disabled with no output — the exact shape M1 set out to close, narrowed rather than eliminated. The code comment acknowledges the possibility; a warning emitted whenever `config_get_or` is undefined (not only on a grep *match*) would close it, since "the config reader is missing" is itself worth saying out loud.
- **nit:** `_lib-tracker.sh:1301` — the comment says the guard rejects "traversal and anything not `owner/name`", but the pattern only rejects `..` segments and spaces; `foo`, `foo/bar/baz`, and `foo/bar?x=1` all pass. Not a fail-open (a malformed path makes `gh api` exit non-zero → rc=2 → block), so behaviour is safe — but the comment overstates what the pattern does. Tighten one or the other.
- **nit:** the M2 test asserts the *guard* (rc=2), not the *exploit* — the shim can't run jq, so with the guard removed the injection returns 1, not 0. That's a fine contract to pin; the comment's "a fail-OPEN" framing slightly oversells what the case can observe.
- **PR mechanics (not a code finding):** the PR still reports `changedFiles=15` and `mergeable=CONFLICTING`. This isn't GitHub lagging — the merge base predates #1049's merge into `dev`, so #1049's files are genuinely still part of this PR's diff. It needs a rebase (or a `dev` merge) before it can merge; that will also shrink the diff to the 5 in-scope files.

### Verdict
**APPROVED**

Round-1 findings closed with verified fixes rather than asserted ones, Hakim's four addressed without introducing new fail-open paths, and the test suite is mutation-proven load-bearing at 12/12. Nothing regressed. Remaining items are polish.

---
📌 Reviewed commit: `2e4e46f8153c64b8dbdb0f1d09a6f9feba83e7a2`

atlas-apex (commented):
## Code Review: PR #1052 — round 3

**Commit**: `e2c5c6b2f6b9dd2772b3f4479dc6258936210fa9`

### Verdict

**APPROVED.** Everything I asked for in round 2 is done, and I verified each item independently rather than taking the summary's word for it.

### 1. The rebase is faithful — verified, not assumed

- Merge base is now exactly `9130cdb` (#1049's squash); `git merge-base --is-ancestor 9130cdb <head>` → **YES**. Diff is 5 files, 612 insertions, **zero deletions**.
- **Nothing gained:** diffing my round-2 approved head (`2e4e46f8`) against this head across the 5 in-scope files shows *only* the `e2c5c6b` changes (3 files, +36/−11). No #1049 content leaked into the in-scope set.
- **Nothing lost** — this was the real risk of a hand-rebuilt branch, so I checked it the hard way. The old head touched 15 files. For the 10 dropped ones I diffed `2e4e46f8` against `9130cdb` directly:
  - `block-merge-on-red-ci.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh` (+6 lines each in the old PR) → **zero leftover diff** against dev. Dev already carries that exact content; it was #1049's, correctly excluded.
  - `auto-code-review.sh`, `docs/rule-audit.md` → the only leftover hunks are #1049's own AgDR-0109→0111 renumbering, i.e. dev is *ahead*, not missing feature content.
  - `warn-review-marker-write.sh`, `AgDR-0111`, the two review test suites, the skills/agents/rules docs → all #1049's.

  No feature content is sitting in a file this PR no longer touches.

### 2. Block structure after removing the dead `if` — correct

`bash -n` passes, and I checked the nesting by hand rather than trusting it: `if` at :313 → `fi` at :396; `case $?` at :344 → `esac` at :395; the inner missing-lib `if` :319→:332 and the new undefined-fn `if` :339→:342 both close inside. The CEO-marker check at :398/:405 is outside the new `if` and therefore reached on every path that doesn't exit — confirmed live by the flag-OFF and flag-ON-verified cases both returning rc=0 with the CEO marker present. **No stray `fi`.**

### 3. The truncated-lib guard fires, and the test catches its removal

Mutation-verified: deleting the `if ! command -v tracker_review_at_sha` block makes the new case go **FAIL** (12/13) with `line 339: tracker_review_at_sha: command not found`. The guard is load-bearing for the test.

One precision note, in this PR's favour: with the guard removed the hook *still exits 2* — `command not found` yields rc 127, which the `case`'s `*)` catch-all already blocks on. So the guard's value is a **correct diagnosis** ("truncated or corrupt file") instead of the misleading "NO code review was posted", not the closing of a fail-open. The code comment says exactly that — "whose rc the case below would misread" — and does **not** claim fail-open. Accurate as written; I'm noting it only so nobody later upgrades the claim.

### 4. Item 1's fix does not over-warn — default path is silent

The warning is gated on `command -v config_get_or` failing, so it cannot fire when the reader is present. Confirmed empirically on two ordinary merges:

- flag absent (default), reader present → `rc=0`, **stderr empty**
- flag ON + review at HEAD, reader present → `rc=0`, **stderr empty**

No new noise on an ordinary merge.

### 5. Regressions from rounds 1–2 — none

- New suite: **13/13 PASS** (was 8; +M1, +M1-control, +missing-lib, +M2, +truncated).
- Neighbouring `test_block_unreviewed_merge.sh`: **39/39 PASS**.
- The M2 sha/traversal guards, the fail-closed rc=2 path, the rc=3 skip, and the flag-OFF no-op all still hold.
- The repo-guard comment now says what the pattern actually does (rejects traversal and whitespace, not owner/name shape) — matches the code.

### Issues Found

None blocking.

- **nit: two stale "8 cases" claims.** The suite is now 13. `docs/agdr/AgDR-0112-…md:113` says "(8 cases, both directions)" and the PR body's Testing section says "**8/8**". In a PR whose whole argument is *don't state what you haven't verified*, a stale count is the one inaccuracy worth fixing. Cheap to correct in the body; the AgDR line is a one-word edit if you're touching it anyway.
- **nit: PR body preamble is now stale** — "Merge #1049 first and this diff shrinks to its own 5 files" describes a state that already happened. Harmless, but a reader landing on the PR cold will be briefly confused.
- **nit: `.claude/rules/pr-workflow.md:133`** (in dev, not this diff) says this control "is **not shipped here**; do not assume it is active." After this merges that's half-wrong — it ships, default-off. Out of this PR's scope, but worth a one-line follow-up so the prose doesn't contradict the shipped flag.
- **nit: indentation** — `case $?` at :344 is indented 4 spaces while its `esac` at :395 is at 2. Cosmetic only; `bash -n` and the suite are fine with it.

### Checklist Results

- Architecture & Design: **Pass** — control/backstop split respected; four-way rc contract is the right shape.
- Code Quality: **Pass** — fails closed on every "couldn't check" branch; dead conditional removed.
- Testing: **Pass** — 13 cases, adversarial both directions, mutation-verified.
- Security: **Pass** — sha/repo validated before interpolation; degraded config read now always announces itself.
- Performance: **Pass** — one extra forge round-trip, only when enabled.
- PR Description & Glossary: **Pass** (glossary present); Summary bullets narrative — **Pass**.
- Summary Bullet Narrative: **Pass**.
- Technical Decisions (AgDR): **Pass** — AgDR-0112 linked, and it is honest about being weaker than AgDR-0062's deferred control.
- Adopter Handbooks: **N/A** — none loaded.
- Fallow (JS/TS): **N/A** — no JS/TS in diff.

---
📌 Reviewed commit: `e2c5c6b2f6b9dd2772b3f4479dc6258936210fa9`

atlas-apex (commented):
## Code Review: PR #1052 — delta confirmation (round 4)

**Commit**: `7391f338aeaf59e13c7205e9fb8c30c098e1b7c2`

This is a **delta-only confirmation** on my round-3 approval at `e2c5c6b`, not a fresh review. The feature has been reviewed three times; I re-verified only the 3-line delta and the four factual claims it makes.

### Delta scope

`git diff e2c5c6b..7391f33` is exactly 3 files / 3 lines, with no smuggled logic change:

| File | Change |
|------|--------|
| `.claude/hooks/block-unreviewed-merge.sh` | Whitespace only — `    case $?` → `  case $?` |
| `.claude/rules/pr-workflow.md` | The about-to-be-wrong sentence, replaced |
| `docs/agdr/AgDR-0112-*.md` | Test count `8 cases` → `13 cases` |

### Verification

**Block structure intact.** The re-indent did not disturb nesting: `if [ "$REQUIRE_POSTED_REVIEW" = "true" ]` (L313) → `case $?` (L344) → `esac` (L395) → `fi` (L396), with the CEO-marker check still reached at L398. `bash -n` passes; `shellcheck -S warning` is clean.

**The `pr-workflow.md` sentence is factually accurate for the post-merge state.** This was the one item worth real scrutiny — the file is `@`-imported into every session, so an error here is high-blast. All four claims check out against the code:

- *Flag exists* — `review_markers.require_posted_review`, read at L290.
- *Defaults off* — `"require_posted_review": false` in `.claude/project-config.defaults.json:148`.
- *GitHub-only* — `tracker_review_at_sha` returns rc=3 for any non-`gh` forge (`_lib-tracker.sh:1311-1314`).
- *NOT separation of duties* — confirmed by the implementation, not just the prose: the jq filter selects any review whose `.commit_id` matches, with **no author predicate** and no review-state predicate (`_lib-tracker.sh:1323-1324`). It genuinely accepts the PR author's own COMMENTED review. Describing it as anything stronger would be wrong, and the sentence correctly doesn't.

The claim it *does* make — that a posted review is a server-side fact a local file write cannot fabricate — is the accurate characterisation of what this buys.

**Relative link resolves.** `../../docs/agdr/AgDR-0112-verify-posted-review-at-head.md` from `.claude/rules/` normalises to `docs/agdr/AgDR-0112-verify-posted-review-at-head.md`, which exists at this HEAD. I checked the depth arithmetic, not just file existence — `..` → `.claude`, `../..` → repo root. The sibling link to AgDR-0111 in the same file uses identical depth and also resolves.

**Test count is now correct.** Ran the suite at this HEAD: `PASS: 13   FAIL: 0`. The 13 comprises 10 `run_case` invocations plus 3 standalone unit checks. The old `8` was genuinely stale; `13` matches.

### Checklist Results
- Architecture & Design: Pass (unchanged from round 3)
- Code Quality: Pass — indentation nit actioned
- Testing: Pass — 13/13 green, count claim now accurate
- Security: Pass — no logic touched; the fail-closed shapes I verified in round 3 are byte-identical
- Performance: N/A
- PR Description & Glossary: Pass — body rewritten, stale preamble and count removed
- Technical Decisions (AgDR): Pass — AgDR-0112 present and linked
- Adopter Handbooks: N/A (no handbook triggers in a 3-line docs/whitespace delta)

### Issues Found
None. All four round-3 advisory nits are now resolved — two in code, two in the PR body.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `7391f338aeaf59e13c7205e9fb8c30c098e1b7c2`

atlas-apex (commented):
## Security Review: PR #1052 (delta-only re-review)

**Commit**: `7391f338aeaf59e13c7205e9fb8c30c098e1b7c2`
**Prior review**: APPROVED at `f8a219fe` (3 MEDIUM, several LOW)

Re-reviewed only the delta since `f8a219fe`, per scope. Two of my three MEDIUMs are fully closed and I verified them by execution rather than by reading the diff. **M1 is closed for one of its two stated triggers and still open for the other**, so this is a **CHANGES REQUESTED** on a single one-line fix — everything else is good to go.

### M2 (jq-filter injection on `${sha}`) — CLOSED, verified by PoC

Re-ran my injection PoC against the current `tracker_review_at_sha`, with `tracker_kind`/`gh` stubbed so that a *passed* validation would be visibly reachable:

| Vector | Result |
|---|---|
| `" or true or "` (close literal, always-match) | rc=2 |
| `a" ) \| .id // ("` (filter escape) | rc=2 |
| `$(touch /tmp/pwned)` | rc=2, no side effect |
| newline embedded in sha | rc=2 |
| `../../etc/passwd` as repo | rc=2 |
| repo containing whitespace | rc=2 |
| `1;rm -rf /` as pr | rc=2 |
| 6-hex / 41-hex boundaries | rc=2 / rc=2 |
| valid 7-hex and 40-hex | rc=0, reaches `gh` |

Every vector fails **closed** at rc=2 before reaching `gh`, and legitimate SHAs still pass. The hex-only class check is the right shape — it subsumes quotes, backslashes, whitespace and `$(` in one test rather than blocklisting metacharacters. Closed.

### M3 (rc=3 skip vs fail-closed doctrine) — CLOSED, the reasoning is now honest

I asked for either a block or an honest account of the skip. The rc=2-vs-rc=3 distinction is now written into the code — "should have been able to check and could not" (precondition failure → block) versus "outside my jurisdiction, there was no check to fail" (→ skip) — and it holds up. Blocking every GitLab adopter buys no security when the check was never available to them.

What moved me is that the warning no longer implies protection: it says the flag is a **NO-OP** on that forge and "do not rely on it here." Combined with default-off, the GitHub-only note in `defaults.json`, and the same in `pr-workflow.md`, an operator has four places telling them the truth. I withdraw the skip-should-be-block position.

### M1 (flag read failed open) — HALF CLOSED, one trigger still fails open

The missing-lib trigger is genuinely fixed: `config_get_or` undefined now routes to the grep fallback, which warns unconditionally and blocks. Good.

But the code claims to cover a second trigger it does not. Three places assert jq-absence is handled:

- hook comment: *"Reader unavailable (missing lib, or jq absent)"*
- hook WARN text: *"`_lib-read-config.sh` (or jq) is unavailable"*
- test comment: *"if it (or jq) is missing, `config_get_or` is undefined"*

That premise is false. `_lib-read-config.sh` does not need jq at **source** time — jq is only checked inside `_config_load`. So when jq is missing the lib sources fine, `config_get_or` **is** defined, `command -v config_get_or` succeeds, and the grep fallback is never reached. `config_get` then returns empty and `config_get_or` hands back its `'false'` fallback.

Reproduced against the real `_lib-read-config.sh` at this HEAD, with `project-config.json` containing `"require_posted_review": true`:

```
=== jq present ===  PATH TAKEN: config reader   -> REQUIRE_POSTED_REVIEW=true
=== jq ABSENT  ===  PATH TAKEN: config reader   -> REQUIRE_POSTED_REVIEW=false
```

An operator whose config says `true` gets the control **silently disabled**. That is the exact shape the fix was written to close, in the PR's own words: *"defaulting to false there would SILENTLY DISABLE this control while the operator's config still says true."* The only trace is a generic `jq not installed; project config unavailable` line that says nothing about the merge gate. The test cannot catch it because it only ever does `rm -f _lib-read-config.sh`, never simulates a missing jq.

**Severity: MEDIUM.** Tempered by two real mitigations — the feature is opt-in and default-off, and a jq-less environment degrades many config-driven hooks framework-wide, so this is not a hole this PR invented. I am not asking you to fix framework-wide jq degradation. I am asking that this control not *document a protection it does not have*, which is the same defect I flagged on the AgDR's "closed" overclaim.

**Required fix — one line:**

```bash
if command -v config_get_or >/dev/null 2>&1 && command -v jq >/dev/null 2>&1; then
```

This routes jq-absence into the grep fallback that already exists and already warns unconditionally, and it makes all three comments true simultaneously. A test that shadows `jq` off `PATH` would lock it in. (Deleting the "or jq" claims from the three comments would also be honest, but leaves a known fail-open standing — I'd prefer the one-liner.)

**No false-positive risk in the fallback**, which you asked me to check specifically: the grep matches only a literal `"require_posted_review": true`, `defaults.json` ships `false`, and the "reader missing + flag OFF -> ALLOWED" control test pins it. It cannot invent a `true` that blocks a legitimate merge.

### Rebase and doc delta — clean

- `.claude/rules/pr-workflow.md` is `@`-imported into every session, so I read it as a claim surface. It does not overclaim: it says **default off**, **GitHub-only**, "deliberately weaker than the author-independence gate AgDR-0062 deferred," and states plainly it is **not** separation of duties. Correct and appropriately blunt.
- `defaults.json`'s comment carries the same disclaimer. Consistent.
- AgDR-0112's overclaim is properly corrected to "raised in cost and made auditable — **not closed**," and now names *both* residual paths I identified: the stub `--comment` review at HEAD, and disabling the check by editing the untracked `project-config.json` (no git trace). Accurate.
- My LOW (missing-lib block untested) is closed, plus the truncated-lib case now fails closed via a `command -v tracker_review_at_sha` check with its own diagnosis. Good catch beyond what I asked for.
- The rebase onto `dev` and the three one-line doc/whitespace changes in `7391f33` introduce nothing security-relevant.

### Checklist Results
- Secrets & Credentials: Pass
- Injection Prevention: **Pass** (M2 closed, PoC-verified)
- XSS Prevention: n/a
- Auth & Authorisation: Pass
- Data Protection: Pass
- API Security: Pass
- Gate & Trust-Chain Integrity: **Conditional** — the control fails open when jq is absent, on a path the code claims to handle

### Verdict

**CHANGES REQUESTED** — one line, narrow and mechanical.

To be clear about proportion: the feature is sound, M2 and M3 are closed, and the honesty work on the AgDR and the rule file is exactly right. I'm holding only because a control whose entire thesis is "don't fail open" should not ship with a documented-but-absent protection. Make the one-line change and I'll approve immediately.

---
📌 Reviewed commit: `7391f338aeaf59e13c7205e9fb8c30c098e1b7c2`

atlas-apex (commented):
## Code Review: PR #1052 — delta confirmation (round 4)

**Commit**: `cdbcceb88d006263c556ced069d66898303a6ec2`

Not a fresh review. I confirmed only the delta since my round-3 approval at `e2c5c6b`. Note the delta is **three** commits, not one — `7391f338` and `0765e589` also landed; `0765e589` added the jq workaround and `cdbcceb` removes it, which is why the net reads as a deletion.

### 1. Removal is clean — verified

`bash -n` passes on both files. Nesting around the excised loop is intact: `if [ "$REQUIRE_POSTED_REVIEW" = "true" ]`@290 → `case`@321 → `esac`@372 → `fi`@373. No orphaned `else`, dangling `fi`/`done`, or leftover variable. The CEO-marker check at :375 sits outside the block and is still reached. The flag read is back to a clean three lines.

### 2. No test was silently dropped — verified by label diff

Ran the suite at both commits and diffed case labels:

| Commit | Cases |
|---|---|
| `e2c5c6b` (approved) | 13 |
| `0765e589` (intermediate) | 15 |
| `cdbcceb` (HEAD) | **11, all green** |

Removed vs. my approval point: exactly `M1: config reader missing…` and `M1 control…`. Nothing else dropped, renamed, or weakened. (Relative to the intermediate it's 4 removals, hence "15 → 11".) AgDR-0112's "11 cases" is accurate — 8 `run_case` calls plus 3 inline unit/M2 checks.

**Mutation results** — 4 of the 6 fail-open guards are genuinely mutation-sensitive:

| Mutation | Suite |
|---|---|
| rc/out order swapped in `tracker_review_at_sha` | **caught** (forge-query-FAILS case) |
| sha hex guard removed | **caught** (M2) |
| repo traversal guard removed | **caught** (M2) |
| missing-lib guard neutered | *not* caught |
| truncated-lib guard neutered | *not* caught |

The last two are masked: the missing-lib case asserts `rc=2` + stderr matching `_lib-tracker.sh`, and the `cat >&2` message prints before the (mutated) exit, while `rc=127` from the undefined function falls into the `*)` arm and exits 2 anyway. So the outcome stays **fail-closed** — this is redundancy, not a hole — but those two assertions are weaker evidence than "mutation-verified" implies. Pre-existing at `e2c5c6b`; not introduced here, so not blocking.

### 3. The new comment's mechanism — verified empirically, and it is half-scoped

The hook comment's own sentence is accurate and correctly narrow: *"A merge with jq absent is already refused by the #965 guard above… at :113."* I tested it rather than trusting it — jq removed from PATH:

| Payload | Result |
|---|---|
| `gh pr merge …` | exit 2 at :113 |
| `gh api …/pulls/N/merge` | exit 2 at :113 |
| tab-separated (#973 escape case) | exit 2 at :113 |
| non-merge (`ls`) | exit 0, clean no-op |

Confirmed: with jq gone the flag read at :286 is unreachable. That half of the removed code was genuinely dead.

### nit (non-blocking, but please amend): the NOTE's premise is half-true

The test NOTE says of the removed case: *"It also never fired."* The removed branch had **two** triggers — `config_get_or` undefined because **jq** is absent, **or** because **`_lib-read-config.sh` is missing**. Only the first is dead. I probed the second (jq present, `_lib-read-config.sh` removed, `require_posted_review: true`, no review posted):

```
FAIL [PROBE: reader missing + jq present + flag ON -> expect BLOCK]: rc=0 want=2
```

The flag silently reads `false` and the merge is **allowed** with no warning — the exact fail-open-quietly shape this control's own comments argue against. At `e2c5c6b` that scenario blocked, so this delta *does* change behaviour in one corner, contrary to "nothing behavioural changed."

I'm not asking for the workaround back — "a missing `_lib-read-config.sh` is a broken install we don't support" is a defensible call, and the base two-marker gate (Rex + CEO) is unaffected, and the flag is default-off. The ask is narrow: **correct the two sentences** so a future reader isn't told the deleted code was dead when half of it covered a live path. Something like *"the jq-absent trigger is unreachable (#965 guard, :113); the reader-missing trigger is a broken install we deliberately don't support."*

### Other doc claims — spot-checked, all accurate

`pr-workflow.md`'s replacement paragraph checks out: default is `false` in `project-config.defaults.json`; non-`gh` forges return rc=3 (GitHub-only); and the reviews filter selects on `commit_id` with **no author predicate**, so "accepts a review from the PR's own author — not separation of duties" is correct.

### Checklist
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass (11/11; 4/6 guards mutation-sensitive, 2 masked but fail-closed)
- Security: Pass — base merge gate unaffected; opt-in layer degrades silently only on a broken install
- Performance: N/A
- PR Description & Glossary: Pass (unchanged since round 3)
- Technical Decisions (AgDR): Pass — AgDR-0112 present, count corrected
- Fallow (JS/TS): N/A (shell + markdown)

### Verdict
**APPROVED** — the removal is clean, complete, and correctly justified for the jq case; the one inaccurate clause is a comment fix, not a code change.

---
📌 Reviewed commit: `cdbcceb88d006263c556ced069d66898303a6ec2`

atlas-apex (commented):
## Security Review: PR #1052 — delta confirmation

**Commit**: `fc31c010a7de90668ccf2d45c74ede61b4f313f5`
**Scope**: re-check of my M1 finding only (held at `7391f33`), plus a spot-check that M2/M3 are untouched. Not a re-audit of the feature.

### Verdict

**APPROVED** — M1 is closed on its merits. The split (jq-absent = unreachable, documented as untested; reader-missing = fail closed) is the right resolution, and I verified both halves empirically rather than taking the comment at its word.

---

### 1. Is the jq-absent claim true? — YES, verified

The claim is that a merge with jq absent never reaches the flag read because the #965 guard at `:113` catches it first. This is airtight by control flow: `COMMAND` is assigned **only** inside `if command -v jq` (`:83-85`), and the `[ -z "$COMMAND" ]` branch (`:87-117`) exits unconditionally — `exit 2` if the raw payload is merge-shaped, `exit 0` otherwise. There is no path to `:118`+ with jq absent.

I confirmed it at runtime against a jq-free `PATH`, tracing with `bash -x` and asserting the flag read is never evaluated:

| Payload (jq absent) | rc | flag read evaluated? | marker check evaluated? |
|---|---|---|---|
| `pr merge <N> --repo … --squash` | 2 | no | no |
| `gh api repos/…/pulls/<N>/merge` (#47 shape) | 2 | no | no |
| JSON-escaped tab separators (#973 shape) | 2 | no | no |
| merge shape with no PR number | 2 | no | no |
| merge shape with var-indirect PR number | 2 | no | no |
| non-merge payload (`ls -la /tmp`) | 0 | no | no |
| bare word `merge` | 0 | no | no |

All five merge shapes exit 2 with the `:113` "jq is unavailable" message; non-merge payloads stay no-ops. **I could not construct a merge-shaped input that reaches the flag read with jq absent** — so there is no finding here.

Leaving it unhandled is therefore defensible, and the comment at `:281-285` states it accurately. Worth being explicit about *why* this is safe rather than merely convenient: it isn't that jq-absence is unlikely — it's that jq-absence and "this control silently didn't run" are **mutually exclusive states**. The merge is refused before the control's precondition is ever consulted.

### 2. Is the new fail-closed guard correct and tight? — YES

The guard fires only when `config_get_or` is undefined AND at least one config file exists. I ran the PR's committed suite (12/12) plus 8 adversarial cases it doesn't cover, aimed at the guard's *negative* side:

| Case | Expected | Result |
|---|---|---|
| reader missing + **no config files at all** | ALLOW (no false block) | ALLOW |
| reader missing + flag `false` on disk | BLOCK (cannot know) | BLOCK |
| reader missing + defaults file only | BLOCK | BLOCK |
| reader missing + overrides file only | BLOCK | BLOCK |
| reader present but **truncated** (fn undefined) | BLOCK | BLOCK |
| reader present but **syntax-broken** (source fails, swallowed by `\|\| true`) | BLOCK | BLOCK |
| normal install, flag defaulted | ALLOW (no new failure mode) | ALLOW |
| flag ON + review at HEAD | ALLOW (regression) | ALLOW |

Against your three sub-questions:

**(a) Cannot false-block a config-less fork** — confirmed. Both config files removed → merge proceeds. The existence probe is what keeps this tight.

**(b) Bypass** — not meaningfully. Evading it requires deleting `_lib-read-config.sh` *and* both config files, one of which (`project-config.defaults.json`) is committed upstream. That is a larger, more visible mutilation than simply setting the flag to `false` in the open — and anyone who can delete hook libraries can delete `block-unreviewed-merge.sh` itself. The guard's threat model is a **partial install / bad sync**, not an adversary with write access to the hooks directory, and it fits that model correctly.

**(c) No new failure mode on a normal install** — confirmed. On a healthy install `:131` defines `config_get_or`, so the `elif` is never evaluated; the flag-defaulted case remains a total no-op. The two corruption variants I added (truncated, syntax-broken) both land in the guard rather than sliding past it, which is the behaviour you want and which the "missing file" framing of the message slightly undersells.

One structural property worth naming, because it is what makes the guard safe in the awkward tail: the config-file probe resolves against the **same root the approval-marker paths resolve against**. So a root with no config files also has no approval markers, and the merge is already blocked upstream by the Rex-marker check. The guard cannot be sidestepped by shifting root resolution without also losing the markers that would let the merge through.

### 3. Is M1 closed? — YES

My original objection was that an opt-in control which can silently be not-running isn't one you can depend on. Judging each half:

- **Reader missing** — this was the real defect, and it was live: flag ON, reader gone, `rc=0`, control skipped, merge allowed. It now denies. The objection is satisfied.
- **jq absent** — the objection doesn't apply, because there is no state in which jq is absent and the control silently doesn't run. The merge is refused first. Documenting it as untested-on-purpose, with the reason, is the honest call; hand-parsing JSON to survive an unsupported state was correctly removed.

I'll note the process point plainly: the operator was right to reject my proposed fix. `&& command -v jq` plus a grep fallback would have added parsing machinery for an unreachable state, and it would not have caught the reader-missing hole — which was the half that actually mattered.

### 4. Regression spot-check — M2 and M3 intact

- **M2** — `sha` shape validation (7-40 hex) and the `repo` traversal/whitespace rejection are unchanged in `_lib-tracker.sh`. Crafted-quote sha, non-hex, over-length, traversal, embedded space, and non-numeric `pr` all return 2; never fail-open.
- **M3** — the `rc=3` branch still SKIPs rather than blocks, and still prints the explicit `NO-OP; do not rely on it here` warning (`:347`). The four-way contract (0/1/2/3) verifies at unit level.

### Checklist

- Secrets & Credentials: Pass
- Injection Prevention: Pass (M2 guards verified)
- Auth & Authorisation: Pass — control fails closed on every precondition failure I could construct
- Gate & Trust-Chain Integrity (#777): Pass — no gate weakened; one prior silent-allow removed
- Data Protection / XSS / API: n/a to this delta

### Informational (not blocking, no action needed)

The `_lib-read-config.sh is missing` message names only the missing-file case, but the guard correctly also fires when the file is present-but-corrupt. A reader hitting that message with the file visibly on disk may be briefly confused. Not worth a commit on its own.

---

📌 Reviewed commit: `fc31c010a7de90668ccf2d45c74ede61b4f313f5`

atlas-apex (commented):
## Code Review: PR #1052 — delta-only confirmation

**Commit**: `fc31c010a7de90668ccf2d45c74ede61b4f313f5`

Delta-only pass over `cdbcceb..fc31c01` (3 files). Feature not re-reviewed — approved four times prior.

### 1. The `elif` does not over-block — verified independently

`config_get_or` is defined at `:131` by `. "$(dirname "$0")/_lib-read-config.sh" 2>/dev/null || true`, so on any intact install the `if` arm wins and the `elif` is never evaluated. I built four sandbox probes rather than inherit Hakim's trace:

| Probe | State | Result |
|---|---|---|
| P1 | config-less fork (no `project-config.json`, no `defaults.json`) + reader removed | **allowed** — `elif` correctly does not fire |
| P2 | normal install, reader present, flag OFF | **allowed** — `elif` unreachable |
| P3 | reader removed, only `defaults.json` present | **BLOCKS** — second condition works standalone |
| P4 | reader removed, flag explicitly `false` on disk | **BLOCKS** — refuses to guess |

All 4 pass. The trigger is exactly "reader undefined AND ≥1 config file present" — no wider.

P4 is worth naming explicitly since it isn't in the suite: the guard blocks even an operator who *deliberately* set the flag to `false`, because with the reader gone the hook cannot read that `false`. That is the stated intent, the message names the fix, and it only occurs on a broken install — but it is a behaviour an adopter could be surprised by, so it is worth knowing it is deliberate rather than incidental.

### 2. Block structure intact

- `bash -n` clean.
- `if` (:287) / `elif` (:289) / `fi` (:300) balanced; `if [ "$REQUIRE_POSTED_REVIEW" = "true" ]` at :301 still the next statement.
- Full merge-gate regression suite: **39/39 green** — the human-approval marker path downstream is still reached and unaffected by this third structural edit to the region.

### 3. The corrected NOTE is accurate on both halves

- **jq absent → unreachable.** Verified directly, not inherited: with a genuinely jq-free `PATH` (symlink farm excluding `jq`) and a merge-shaped payload with the flag ON, the hook exits 2 at the #965 guard with `jq is unavailable…` and never reaches the flag read. The `:113` citation is right.
- **Reader missing → was live.** Matches what I probed at `cdbcceb` (`rc=0`, silent allow); P3/P4 confirm it now blocks. The NOTE's split of the two triggers, and its explanation of why each gets different treatment, both hold.

### 4. Fail-open mutation spot-checks (2 of 6)

| Mutation | Expected | Got |
|---|---|---|
| forge-query-FAILS arm `exit 2` → no-op (:362) | suite 12 → 11, that case only | ✅ `rc=0 want=2` on *flag ON + forge query FAILS* |
| truncated-lib guard removed (:326–329) | suite 12 → 11, that case only | ✅ *truncated `_lib-tracker.sh`* fails |

Hook restored byte-identical after each; suite back to **12/12**.

### Issues Found

None blocking.

### Suggestions

- nit (non-blocking, no action needed this PR): if `MARKER_HOME` ever resolves to a tree with no `.claude/project-config*.json` *and* the reader is missing, the flag still silently defaults to `false`. That is the deliberate carve-out P1 exercises and requires two simultaneous failures, so it is the right trade — noting it only so the narrow residual is on the record.

### Verdict

**APPROVED**

The code change you did not ask for is the right one and is correctly scoped: four lines, no parsing, fires only on evidence, same shape as the missing-tracker-lib guard below it. The NOTE now describes what the code does.

---
📌 Reviewed commit: `fc31c010a7de90668ccf2d45c74ede61b4f313f5`

Issue #1034, opened 2026-07-27
[Bug] _portfolio_resolve still misclassifies absolute drive-letter paths as relative (follow-up to #1029)


atlas-apex:
Follow-up from the review of #1029 (`fix(#1018): normalize Windows drive-letter paths in portfolio_validate`), which merged as `2554869`. Flagged independently by both the code review and the security review as **MEDIUM**, non-blocking — filing so it isn't lost.

## Given / When / Then

**Given** #1029 taught `_portfolio_canonicalize` to normalize Windows drive-letter prefixes (`X:/…` → `/x/…`), but left `_portfolio_resolve`'s own absoluteness test as `case "$p" in /*)` 
**When** a Windows adopter sets an **absolute** drive-letter override, e.g. `.portfolio.registry: "D:/Portfolio/apexyard.projects.yaml"`
**Then** the value is misclassified as *relative* and concatenated onto the ops-fork root, yielding something like `/d/Projs/fork/D:/Portfolio/apexyard.projects.yaml` — expected that an absolute drive-letter path be treated as absolute

## Why it's not urgent

- **Pre-existing** — the same override was equally broken before #1029; this is an unclosed gap, not a regression.
- **Fails visibly**, not silently: the result is a nonexistent path rather than a wrong-but-plausible one, so it surfaces as an error instead of quietly resolving to the wrong repo.
- It resolves *inside* the fork, which is the safe direction, and the #951 fork-containment check surfaces it.
- POSIX platforms are entirely unaffected.

## Why it's worth closing

It makes #1029's stated rationale — "fixing the shared helper fixes it for every consumer" — not quite true, since this consumer still needs its own fix. Leaving the two out of step invites a future reader to assume drive-letter handling is uniformly solved.

## Suggested fix

Extend `_portfolio_resolve`'s absoluteness `case` with a drive-letter arm, mirroring the pattern #1029 added one level down:

```
[A-Za-z]:[/\\]*) abs="$p" ;;
```

## Also noted during review (lower priority)

- The transform hardcodes the MSYS/Git-Bash default mount prefix (`/<letter>/`). Under Cygwin (`/cygdrive/c/…`) or an MSYS2 install with a custom `/etc/fstab` prefix the two spellings still diverge. A conditional `command -v cygpath >/dev/null && cygpath -u "$p"` preferred path, with the current string transform as fallback, would cover those environments without adding a hard dependency.
[...]

PR #1059, opened 2026-07-29
fix(#1034): treat absolute drive-letter paths as absolute in _portfolio_resolve

atlas-apex:
## Summary

- **`_portfolio_resolve` now recognises a drive-letter path as absolute** — a Windows adopter setting `.portfolio.registry: "D:/Portfolio/apexyard.projects.yaml"` had that value classified as *relative* and concatenated onto the ops-fork root, producing `/Users/fork/D:/Portfolio/apexyard.projects.yaml`. The one-line `case` arm added here mirrors the pattern `_portfolio_canonicalize` already used, so the two functions finally agree on what "absolute" means.
- **Closes the gap #1029 left open** — that fix taught the shared canonicalizer about drive letters and stated the rationale as "fixing the shared helper fixes it for every consumer." That was not quite true: this consumer made its own absoluteness decision first, so the value never reached the helper in a form it could repair. Both code reviews on #1029 flagged this independently as MEDIUM.
- **Drive-*relative* spellings stay relative, on purpose, and are now pinned by a test** — `C:foo` means "foo, relative to the current directory on drive C" and is genuinely not absolute. #1034 noted the comment block never said so, leaving a future reader liable to "fix" it into an absolute match. The `[/\\]` after the colon is the line between the two, and there is now a failing test waiting for anyone who relaxes it.
- **Zero POSIX impact** — the new arm can only match a string beginning `<letter>:` followed by a slash, which no POSIX absolute path ever does. Two explicit no-regression cases assert that ordinary absolute and relative paths resolve exactly as before.

## Testing

1. `bash .claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh` → 19 passed, 0 failed.
2. **Discriminating check (the part that matters).** The 8 new cases were written and run *before* the fix. Result against the unmodified library: **4 FAIL, 4 PASS** — the four bug cases failed, and the no-regression guards passed. After the fix: **all 8 pass**. A test that passes both before and after proves nothing; these do not.
3. Full suite: `bash bin/run-hook-tests.sh` — no new failures against the pre-change baseline.
4. `shellcheck --severity=warning` on both changed files — clean.

Why the tests drive the function directly rather than using a filesystem fixture: the drive-letter/MSYS-mount equivalence only exists inside the MSYS runtime. On macOS/Linux `D:/x` and `/d/x` are two unrelated strings, so a fixture would only exercise colon-containing directory names — an unrelated concern. This matches the reasoning already recorded in this test file for the #1018 cases.

Refs #1034

---

## Glossary

| Term | Definition |
|------|------------|
| MSYS mount point | Git-Bash's POSIX spelling of a Windows drive — `/d/` for `D:` |
| Drive-relative path | `C:foo` — relative to the current directory *on that drive*, distinct from absolute `C:/foo` |
| Discriminating test | A test proven to fail against the unfixed code, so passing it afterwards is evidence the fix works |
| Canonicalize | Resolve a path to an absolute, symlink-free, `/../`-free form |

### DIFF
--- a/.claude/hooks/_lib-portfolio-paths.sh
+++ b/.claude/hooks/_lib-portfolio-paths.sh
@@ -213,11 +213,32 @@ _portfolio_canonicalize() {
 # input is already absolute, it is still canonicalized — an absolute
 # override in project-config.json could itself carry a relative segment
 # (e.g. "$SIBLING/../other").
+#
+# Absoluteness has TWO spellings, not one (me2resh/apexyard#1034). #1029
+# taught `_portfolio_canonicalize` (below) to normalize a Windows
+# drive-letter prefix, but left this function's own test as a bare `/*)`.
+# A drive-letter path IS absolute and does NOT start with "/", so it fell
+# into the relative arm and got concatenated onto the ops-fork root:
+#
+#   .portfolio.registry: "D:/Portfolio/apexyard.projects.yaml"
+#     -> /Users/fork/D:/Portfolio/apexyard.projects.yaml
+#
+# The drive-letter arm below mirrors the pattern `_portfolio_canonicalize`
+# already uses, so the two functions now agree on what "absolute" means.
+#
+# Deliberately NOT matched: drive-RELATIVE spellings (`C:foo`, bare `C:`).
+# `C:foo` means "foo, relative to the current directory ON drive C" — it is
+# genuinely not absolute, and treating it as such would resolve it somewhere
+# the caller never asked for. The `[/\\]` after the colon is what draws that
+# line, and it is load-bearing: do not relax it to a bare `[A-Za-z]:*)`.
+# Pinned by the drive-relative case in
+# tests/test_portfolio_paths_windows_drive_letter.sh.
 # ------------------------------------------------------------------------------
 _portfolio_resolve() {
   local p="$1" abs
   case "$p" in
     /*) abs="$p" ;;
+    [A-Za-z]:/*|[A-Za-z]:\\*) abs="$p" ;;
     *)
       local root
       root=$(_portfolio_root)
--- a/.claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh
+++ b/.claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh
@@ -269,6 +269,117 @@ if [ \"\$out\" = \"OK\" ] && [ \"\$rc\" -eq 0 ]; then exit 0; else echo \"got ou
 "
 rm -rf "$SB"
 
+# ===========================================================================
+# me2resh/apexyard#1034 — the follow-up gap #1029 left open.
+#
+# #1029 taught `_portfolio_canonicalize` about drive-letter prefixes, but
+# `_portfolio_resolve` kept its own absoluteness test as a bare `case "$p" in
+# /*)`. A drive-letter path is absolute and does NOT start with "/", so it
+# fell into the relative arm and was concatenated onto the ops-fork root:
+#
+#   _portfolio_resolve "D:/Portfolio/apexyard.projects.yaml"
+#     -> /Users/fork/D:/Portfolio/apexyard.projects.yaml     (wrong)
+#
+# The cases below drive `_portfolio_resolve` directly with a stubbed
+# `_portfolio_root`, for the same reason the #1018 cases above drive
+# `_portfolio_canonicalize` directly: the drive-letter/MSYS equivalence only
+# exists inside the MSYS runtime, so a filesystem fixture cannot reproduce it
+# on a POSIX CI runner. Stubbing the root keeps the assertion about the
+# absoluteness DECISION, which is the actual defect.
+# ===========================================================================
+
+# ---------------------------------------------------------------------------
+# Case 10 (the core acceptance criterion): an absolute drive-letter path is
+# treated as ABSOLUTE — not concatenated onto the fork root.
+# ---------------------------------------------------------------------------
+run_case "#1034: absolute drive-letter path is NOT concatenated onto the fork root" '
+_portfolio_root() { echo "/Users/fork"; }
+r=$(_portfolio_resolve "D:/Portfolio/apexyard.projects.yaml")
+case "$r" in
+  /Users/fork/*) echo "got=$r — still concatenated onto the fork root (the #1034 bug)"; exit 1 ;;
+esac
+expected="/d/Portfolio/apexyard.projects.yaml"
+if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 11: lowercase drive letter, same treatment.
+# ---------------------------------------------------------------------------
+run_case "#1034: lowercase absolute drive-letter path resolves absolute" '
+_portfolio_root() { echo "/Users/fork"; }
+r=$(_portfolio_resolve "d:/portfolio/projects")
+expected="/d/portfolio/projects"
+if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 12: native backslash spelling, same treatment.
+# ---------------------------------------------------------------------------
+run_case "#1034: backslash drive-letter path resolves absolute" '
+_portfolio_root() { echo "/Users/fork"; }
+r=$(_portfolio_resolve "C:\Portfolio\apexyard.projects.yaml")
+expected="/c/Portfolio/apexyard.projects.yaml"
+if [ "$r" = "$expected" ]; then exit 0; else echo "got=$r expected=$expected"; exit 1; fi
+'
+
+# ---------------------------------------------------------------------------
+# Case 13 (must-NOT-change): a POSIX absolute path still resolves absolute.
+# ---------------------------------------------------------------------------
+run_case "#1034: POSIX absolute path still resolves absolute (no regression)" '
+_portfolio_root() { echo "/Users/fork"; }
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #1059

**Commit**: `b856a87aa5e1b88d8cd97f2d5accb9824967780b`

### Summary

Adds one `case` arm to `_portfolio_resolve` so a Windows drive-letter path (`D:/Portfolio/registry.yaml`) is classified as absolute instead of being concatenated onto the ops-fork root. Closes the consumer-side gap #1029 left behind when it fixed the same class inside `_portfolio_canonicalize`. 21 lines of library change (1 code, 20 comment) + 8 test cases.

I did not take the PR's testing claims on trust — I reconstructed the pre-fix library in an isolated worktree and re-ran the suite, then mutation-tested the guards. Results below.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (verified independently, see below)
- Security: **Pass** (no auth/crypto/secrets surface; path classification only)
- Performance: **Pass** (one added `case` arm, no subprocess)
- Summary Bullet Narrative: **Pass** — all four bullets carry what + why
- Technical Decisions (AgDR): **N/A** — agreed, see below
- Adopter Handbooks: **Pass** (3 loaded, 0 findings)
- Fallow Static Analysis: **N/A** (no JS/TS in diff)

---

### 1. Is the `case` arm correct and minimal? — Yes

`[A-Za-z]:/*|[A-Za-z]:\\*` is exactly right, and I checked the glob semantics rather than eyeballing them.

- `[A-Za-z]:/*` is `<letter>` `:` `/` then `*`. Because `*` can match the empty string, **bare `D:/` matches** — I confirmed it resolves to `/d` rather than falling into the relative arm. That was the specific semantics question worth checking and it behaves as intended.
- `\\` inside an unquoted `case` pattern is an escaped backslash, i.e. one literal `\`, so `C:\Portfolio` matches. Verified empirically (`C:\` → `/c`).
- Drive-**relative** forms correctly do **not** match: `C:foo` → `/Users/fork/C:foo`, bare `C:` → `/Users/fork/C:`. The `[/\\]` after the colon is doing the load-bearing work the comment claims it does — and I proved that claim in the mutation section below rather than accepting it.
- Arm ordering is safe: `/*)` is evaluated first, so no POSIX path can ever reach the new arm.

Minimality is good — this mirrors the pattern `_portfolio_canonicalize` already uses verbatim, so the two functions now share one definition of "absolute" instead of two.

### 2. Regression risk — no finding

This is the highest-stakes question (every adopter's portfolio paths flow through here), so I enumerated each path explicitly:

| Path | Verified behaviour | Risk |
|---|---|---|
| POSIX absolute (`/tmp/x`) | Matched by the pre-existing `/*)` arm, which still runs first. Untouched. | None |
| Ordinary relative (`sub/x`, `../sib/x`) | Cannot match the new arm — it requires `<letter>:` at offsets 0–1. Falls through unchanged. | None |
| `./` strip (`./x` → `<root>/x`) | Lives entirely inside the untouched `*)` arm. | None |
| No-root fallback | Also entirely inside `*)`. Relative input with no root still returns the literal (`sub/x` → `sub/x`). | None |

One **intended** delta worth naming: on the no-root branch, a drive-letter input previously returned the literal `D:/x` and now returns `/d/x`. That is the fix working as designed (an absolute path needs no root), not a regression. I checked whether any consumer detects the no-fork state by literal-equality against the raw config value — none does. The only near-miss is `_lib-multi-repo-trace.sh:317` (`|| ws_dir="./workspace"`), which fires on command failure, not on a literal match.

I also ran the neighbouring suites: `test_portfolio_paths.sh` 45/0, `test_portfolio_agent_routing.sh` 6/0, `test_ticket_template_resolution.sh` 36/0, `test_link_custom_skills.sh` 7/0, `test_require_migration_ticket.sh` 26/0.

### 3. Discriminating-test claim — independently reproduced (with a count correction)

I did not trust the claim. Method: `git worktree add --detach` at the PR HEAD, then `git checkout b856a87^ -- .claude/hooks/_lib-portfolio-paths.sh` to restore the pre-fix library while keeping the new tests, then ran the suite.

```
Passed: 15
Failed: 4
  - #1034: absolute drive-letter path is NOT concatenated onto the fork root
      got=/Users/fork/D:/Portfolio/apexyard.projects.yaml
  - #1034: lowercase absolute drive-letter path resolves absolute
  - #1034: backslash drive-letter path resolves absolute
  - #1034: drive-letter and MSYS spellings resolve EQUAL through _portfolio_resolve
```

**Claim confirmed.** The four bug cases fail against the unfixed library with exactly the concatenation the ticket describes, and pass after. These are genuinely discriminating tests, not decoration.

`nit:` **the counts in the PR body and commit message are off by one.** There are **8** new `run_case` blocks, not 7, and the pre-fix split is **4 FAIL / 4 PASS**, not "4 FAIL, 3 PASS" — the fourth passing case is the `./`-prefix guard, which the body's prose describes but the arithmetic omits. The commit message likewise says "Add 7 regression cases". Cosmetic, but on a PR whose central argument is rigour about test discrimination it is worth getting the count right.

### 4. Test quality — strong; guards verified by mutation testing

The four cases that pass both before and after are **not** worthless — they are no-regression guards, and a guard's job is to pass until someone breaks something. The real question is whether they would actually bite, so I mutation-tested them. Five mutants, all caught:

| Mutant | Caught by |
|---|---|
| Relax the arm to bare `[A-Za-z]:*)` — the exact relaxation the comment forbids | `drive-RELATIVE C:foo stays relative` fails |
| Relative arm stops joining the root (everything absolute) | 3 cases fail, incl. both relative guards |
| Drop the `./` strip | `./-prefixed relative path` fails |
| Remove the POSIX `/*)` arm | `POSIX absolute path still resolves absolute` + equality case fail |
| Drop the backslash alternative | `backslash drive-letter path` fails |

Notably the first mutant confirms the comment's "Pinned by the drive-relative case" is a true statement, not an aspiration — that is the check I most expected to find hollow, and it held.

`nit:` three edges are unverified by the suite. I exercised all three by hand and each behaves correctly, so these are coverage gaps rather than defects — worth adding only if the file is touched again:

- **bare `D:/`** → `/d` (the `*`-matches-empty case; the most interesting untested edge, since it is the one place the glob could plausibly have surprised)
- **trailing slash** `D:/x/` → `/d/x`
- **the no-root branch** with a drive-letter input → `/d/Portfolio/reg.yaml` (the one place this PR does change no-root behaviour, and nothing pins it)

### 5. Standard checklist

- PR body has Summary / Testing / Glossary — all present; Glossary defines 4 genuinely non-obvious terms.
- Summary bullets are narrative (what + why), not label-only.
- `Refs #1034` present, single occurrence, correct form for the QA-gate convention.
- Branch `fix/GH-1034-portfolio-resolve-drive-letter` and title `fix(#1034): …` both conform to `git-conventions.md`.
- Commit message: proper subject, body carries the WHY (why #1029's reasoning did not hold), `Refs #1034`. No handbook flag conditions triggered.
- `shellcheck --severity=warning` clean on both files (reproduced locally).
- CI green: hook test suite, shellcheck, CodeQL analyze, Verify Ticket ID.

**AgDR — I agree it is not required.** This reuses a pattern already established in the same file by #1029; it introduces no dependency, no new technology, no schema or security-control change, and no repo-wide pattern. It crosses no row of `.claude/rules/agdr-decisions.md` § "The threshold". Rail 1 does not apply — a portfolio path resolver is not the trust chain (it is not a merge gate, marker library, or credential path), and nothing here is a migration.

---

### Handbook Findings

Three handbooks loaded via path-convention; no violations.

- `handbooks/architecture/migration-safety.md` (**blocking**) — **N/A**. No file in the diff matches any migration glob; no schema DDL present.
- `handbooks/architecture/clean-architecture-layers.md` (advisory) — N/A. Shell hook library, no domain/infrastructure layering.
- `handbooks/general/commit-message-quality.md` (advisory) — **Pass**. Subject is meaningful, body explains WHY rather than restating WHAT, no placeholder text, diff-size/body rule satisfied. None of the four flag conditions trigger.

---

### Suggestions

**MEDIUM — the same defect class survives one file over (follow-up ticket, not a change to this PR).**

`_lib-multi-repo-trace.sh::mrt_workspace_for` (lines 211–217) applies the identical bare-`/*)` absoluteness test to the registry's `workspace:` field, concatenating onto `_portfolio_root` otherwise:

```bash
case "$workspace" in
  /*) candidate="$workspace" ;;
  *)  … candidate="$root/$workspace" ;;
esac
```

A Windows registry entry `workspace: D:/clones/foo` therefore yields `<root>/D:/clones/foo` — the exact shape of #1034, in a second consumer. Blast radius is meaningfully smaller than #1034's: the result is gated by `[ -d "$candidate" ]`, so a wrong path degrades to a silent "not found" and falls through to the `<workspace_dir>/<name>` default rather than being used. Degraded, not dangerous.

**This is explicitly out of scope for this PR and should not be pulled in** — #1034 names `_portfolio_resolve`, and widening the diff would make a cleanly reviewable fix harder to reason about. I raise it only because #1029's "fixing the shared helper fixes it for every consumer" over-claim is precisely the story this PR exists to correct, and it would be a shame to let the same pattern land a third time. Worth a ticket.

`nit:` `[A-Za-z]` is a bracket range and therefore collation-sensitive under some locales. Pre-existing, identical to the pattern `_portfolio_canonicalize` already ships, and a false positive would still require a `<char>:/` prefix — so no new risk. Noting it only for completeness since the arm's glob semantics were under scrutiny.

---

### Verdict

**APPROVED**

Well-scoped fix with an unusually honest test story — the discriminating-test claim survived independent reproduction, and every no-regression guard survived mutation testing. Findings are one MEDIUM follow-up in a *different* file (deliberately not requested here) and three nits, none blocking.

---
📌 Reviewed commit: `b856a87aa5e1b88d8cd97f2d5accb9824967780b`

me2resh (approved):

Issue #1058, opened 2026-07-29
[Bug] validate-pr-create.sh reports sections missing from a body file it could not read (residual of #1048)


atlas-apex:
## Driver

Split out of #1048 (reported by @Ref34t), whose first defect — quotes not stripped from a `--body-file` path — shipped fixed in **v5.3.0** via #1041. The second defect did not, and it is the one @Ref34t identified as costing real time.

`validate-pr-create.sh` treats *"I could not read the body file"* and *"the body file has no required sections"* as the same outcome. When the file is unreadable it emits a warning saying so, then blocks with a message describing the file's contents — a conclusion it has no evidence for.

Verified against the released `v5.3.0` tag:

```
WARN: validate-pr-create.sh: --body-file '/tmp/does-not-exist.md' not readable
      from hook context; section check may miss content.
PR VALIDATION BLOCKED:
PR body missing required '## Testing' section.
PR body missing required '## Glossary' section.
```

#1041 removed the *common trigger* (ordinary shell quoting) without removing the *misdiagnosis*. A typo'd path, a relative path resolved from a different working directory, or a permissions problem still produces a block whose stated reason points the author at their PR body rather than at argument resolution. Nothing in the output suggests the real cause.

## Scope

`.claude/hooks/validate-pr-create.sh` — separate the two states, the way the sibling hook already does.

**The fix already exists in this repo.** `block-private-refs-in-public-repos.sh` had the identical conflation and #1039/#1040 fixed it: a `BODY_FILE_UNREADABLE` flag, set when a `--body-file` is named but cannot be read, and checked *before* the empty-input short-circuit that would otherwise swallow the case.

The same shape applies here, inverted in consequence. The leak hook must **block** on an unrecovered body because it guards a public tracker; this hook must simply **not assert section absence** from a haystack it never read. Two acceptable outcomes, both from @Ref34t's original report:

- Skip the section check, consistent with the existing warning's own "may miss content" hedge, and let the other validations proceed; or
- Block with the real reason — `could not read body file: <path>` — and no section claims.

Prefer whichever keeps the gate fail-closed overall. The defect is the *stated reason*, not the blocking.

## Acceptance Criteria

- [ ] An unreadable `--body-file` never produces a "missing required section" message
- [ ] The resulting output names the actual cause (unreadable path), so the workaround is discoverable from the error text
[...]

Issue #1048, opened 2026-07-28
[Bug] validate-pr-create.sh mis-resolves a quoted --body-file, then reports required sections as missing


Ref34t:
## Affected

`.claude/hooks/validate-pr-create.sh`

## Given / When / Then

**Given** a PR body file that contains all required sections (`## Summary`, `## Testing`, `## Glossary`) and a title referencing an open issue

**When** the `--body-file` argument is passed as a **quoted** absolute path — `gh pr create ... --body-file "/abs/path/body.md"` — which is ordinary, correct shell for a path that could contain spaces

**Then** the hook warns that the file is *"not readable from hook context"*, then blocks (exit 2) with:

```
PR VALIDATION BLOCKED:
PR body missing required '## Testing' section.
PR body missing required '## Glossary' section.
```

— expected: either allowed (both sections are present), or blocked with an accurate reason ("could not read the body file"). Passing the **same** path unquoted is allowed.

## Two defects, the second worse than the first

1. **Quotes aren't stripped before the path is resolved.** The warning text shows the mangled target — the surrounding double quotes are treated as part of the filename:

   ```
   WARN: validate-pr-create.sh: --body-file '"/abs/path/body.md"' not readable
   from hook context; section check may miss content.
   ```

2. **The hook asserts a conclusion it has no evidence for.** It has just told the operator it *could not read the file* — and then reports the required sections as *missing*. That diagnosis is unfounded and actively misleading: it sends the author off to add sections that are already there. An unreadable body-file should either skip the section check (consistent with the warning's own "may miss content" hedge) or fail with the real reason.

The second point is what makes this cost real time. The stated failure ("missing `## Testing`") points at the PR body; the actual failure is argument parsing in the hook. Nothing in the output suggests removing quotes as the fix.

## Repro

1. Write a PR body file containing `## Summary`, `## Testing`, and `## Glossary` sections, and a `Refs #<N>` line for an **open** issue.
2. Feed the hook a `Bash` payload whose command is `gh pr create --repo <owner/repo> --base main --title 'docs(#<N>): probe' --body-file /abs/path/body.md` → allowed (exit 0).
3. Repeat with the path quoted: `--body-file "/abs/path/body.md"` → exit 2, with the "not readable" warning followed by the two "missing section" errors.

Both probes were run by piping `{tool_name, tool_input}` JSON straight into the hook, so the behaviour is the hook's own and not harness-dependent.

[...]

PR #1060, opened 2026-07-29
fix(#1058): stop reporting sections missing from a body file that was not read

atlas-apex:
## Summary

- **`validate-pr-create.sh` no longer reports sections missing from a file it could not read** — when a `--body-file` is named but unreadable, the hook used to warn *"not readable … section check may miss content"* and then block saying `## Testing` and `## Glossary` were missing. That is a claim about the file's contents drawn from a haystack it never opened, and it sent the author to edit a body that was never the problem.
- **Still fails closed — only the diagnosis changed** — the defect was the stated reason, not the blocking. A body the hook cannot inspect is still not a body it can pass, so the exit code is unchanged at 2. What changed is that the message now names the cause it actually has evidence for, and says outright that the section check did not run.
- **Separates two states that were one code path** — this is the same conflation #1039/#1040 fixed in the sibling `block-private-refs-in-public-repos.sh`, using the same `BODY_FILE_UNREADABLE` flag shape. "I could not recover the body" and "the body has no match" are different outcomes; treating them identically is what produced a confident wrong answer in both hooks.
- **Residual of #1048** (reported by @Ref34t), whose first defect — quotes not stripped from the path — shipped in v5.3.0 via #1041. That fix removed the *common trigger* (ordinary shell quoting) without removing the *misdiagnosis*, so a typo'd path, a path relative to a directory the hook could not infer, or a permissions problem still misdirected the fix.

## Before / after

```
BEFORE   WARN: --body-file '/tmp/does-not-exist.md' not readable from hook
               context; section check may miss content.
         PR VALIDATION BLOCKED:
         PR body missing required '## Testing' section.
         PR body missing required '## Glossary' section.

AFTER    PR body file could not be read: /tmp/does-not-exist.md
           The required-section check was NOT run — the sections above may
           well be present.
           Fix the path (check for a typo, or make it absolute) and retry.
```

## Follow-up commit (91de679) — two findings from the first review

- **The message named nothing.** It said "the sections above", but this branch *replaces* the per-section list rather than following it, so the author was left without the one fact they needed. It now reads `UNVERIFIED, not missing: ## Testing, ## Glossary`. On a fix whose entire subject is message accuracy, a dangling reference was the wrong thing to ship.
- **`printf "$ERRORS"` treated error text as a format string** — and this turned out to be broader than a `--body-file` concern. It was called with **zero arguments**, so every conversion spec it ever consumed was a latent bug. Verified corruptions outside this branch: a PR title containing `%` (`bump coverage to 90%` printed as `to 90 0oesn't match format`) and a branch name containing `%` (legal in git refs — `feature/100%-done` became `feature/1000one`). `printf '%b'` keeps the `\n` expansion all six existing message shapes rely on, verified byte-identical, while making `%` literal.

## Testing

1. `bash .claude/hooks/tests/test_validate_pr_required_sections.sh` → 15 passed, 0 failed.
2. **Discriminating check.** The 7 new cases were run against the unmodified hook first: **4 FAIL / 3 PASS**. The four failures were the two "must not claim a section is missing" assertions plus the two asserting the new blocking message.

   Worth recording that two of these started out worthless: they originally matched the pre-existing `not readable` WARN, which passes both before *and* after the fix. They were retightened to assert the new `PR body file could not be read` message specifically, which moved the RED phase from 2 fail/5 pass to 4 fail/3 pass. A test that cannot fail against the old code is not evidence.
3. **No-regression guards** (these pass both before and after, deliberately — they pin behaviour the fix must not change): a readable body genuinely missing `## Testing` still blocks with the section message; a readable complete body still passes. The tempting "just skip the check when unreadable" fix would have quietly made this hook permissive, and these are what would catch it.
4. All six sibling suites touching this hook still pass: `test_flag_extractor_body_file`, `test_pr_create_gate_anchor`, `test_pr_create_tilde_cd_target`, `test_validate_pr_create_head`, `test_validate_pr_create_structural_parse`, `test_validate_pr_create_upstream`.
5. Full suite `bash bin/run-hook-tests.sh` — no new failures against the pre-change baseline.
6. `shellcheck --severity=warning` on both changed files — clean.
7. Behavioural spot-check against a real payload, both directions, confirming the output above.

Note when reproducing by hand: the PR title must reference an **open** issue, or the hook blocks earlier on issue-state validation and masks this signal entirely. @Ref34t flagged that confound in #1048 and it caught me once during this work.

Refs #1058

---

## Glossary

| Term | Definition |
|------|------------|
| Empty haystack | Running a content check over a body that was never successfully read, so a non-match proves nothing about the file |
| Fail closed | On error or ambiguity, deny rather than allow — preserved here; only the stated reason changed |
| `BODY_FILE_UNREADABLE` | Flag distinguishing "a body file was named but could not be recovered" from "there is no body file", first used in `block-private-refs-in-public-repos.sh` (#1039) |
| Discriminating test | A test proven to fail against the unfixed code, so passing it afterwards is evidence the fix works |

### DIFF
--- a/.claude/hooks/tests/test_validate_pr_required_sections.sh
+++ b/.claude/hooks/tests/test_validate_pr_required_sections.sh
@@ -138,6 +138,101 @@ y
 x" \
   2 "missing required '## Testing' section"
 
+# ===========================================================================
+# me2resh/apexyard#1058 (residual of #1048, reported by @Ref34t) — the hook
+# must not report what a file contains when it could not read that file.
+#
+# #1041 fixed the common trigger (a quoted --body-file path was mis-extracted),
+# but not the misdiagnosis behind it: when the body file cannot be read, the
+# hook warns "not readable ... section check may miss content" and then blocks
+# saying '## Testing' and '## Glossary' are MISSING — a claim about the file's
+# contents, drawn from a haystack it never read.
+#
+# The stated reason is the defect, not the blocking. These cases assert the
+# hook still fails closed, but names the real cause.
+#
+# `run_case_missing_file` points --body-file at a path that does not exist,
+# which is what a typo, a path relative to a directory the hook could not
+# infer, or a permissions problem all look like from inside the hook.
+# ===========================================================================
+
+run_case_missing_file() {
+  local label="$1" want_rc="$2" want_stderr_regex="$3" reject_stderr_regex="${4:-}"
+  local sb; sb=$(make_sandbox)
+  mock_gh_install "$sb"
+  local missing="$sb/definitely-not-here-1058.md"
+  local cmd="gh pr create --repo me2resh/apexyard --title 'chore(#113): test' --body-file $missing"
+  local input; input=$(jq -nc --arg c "$cmd" '{tool_input:{command:$c}}')
+  local got_stderr got_rc
+  got_stderr=$(cd "$sb" && echo "$input" | bash .claude/hooks/validate-pr-create.sh 2>&1 >/dev/null)
+  got_rc=$?
+  rm -rf "$sb"
+
+  if [ "$got_rc" != "$want_rc" ]; then
+    echo "FAIL [$label]: want rc=$want_rc, got $got_rc (stderr: ${got_stderr:0:200})" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  if [ -n "$want_stderr_regex" ] && ! echo "$got_stderr" | grep -qE "$want_stderr_regex"; then
+    echo "FAIL [$label]: stderr did not match /$want_stderr_regex/" >&2
+    echo "    stderr: $got_stderr" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  if [ -n "$reject_stderr_regex" ] && echo "$got_stderr" | grep -qE "$reject_stderr_regex"; then
+    echo "FAIL [$label]: stderr WRONGLY matched /$reject_stderr_regex/ (asserting a claim about a file it could not read)" >&2
+    echo "    stderr: $got_stderr" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  echo "PASS [$label]"
+  PASS=$((PASS+1))
+}
+
+# The core defect: no section claim may be made about an unread file.
+run_case_missing_file "#1058: unreadable body-file does NOT claim '## Testing' is missing" \
+  2 "" "missing required '## Testing' section"
+
+run_case_missing_file "#1058: unreadable body-file does NOT claim '## Glossary' is missing" \
+  2 "" "missing required '## Glossary' section"
+
+# It must still fail closed, and the message must name the real cause so the
+# fix is discoverable from the output (the whole point of @Ref34t's report:
+# the old message sent the author to edit a body that was never the problem).
+run_case_missing_file "#1058: unreadable body-file blocks (fail closed, not permissive)" \
+  2 "" ""
+
+# These two assert the NEW blocking message specifically, not the pre-existing
+# "not readable" WARN — matching the WARN would pass before and after the fix
+# and therefore prove nothing.
+run_case_missing_file "#1058: the block names the unreadable body-file as the cause" \
+  2 "PR body file could not be read" ""
+
+run_case_missing_file "#1058: the block quotes the offending path" \
+  2 "PR body file could not be read:.*definitely-not-here-1058\.md" ""
+
+# ---- No-regression guards for #1058 -------------------------------------
+# The fix must not make the hook permissive. A body the hook CAN read and
+# which genuinely lacks a section must still be blocked with the section
+# message — that is the check's actual job, and the one thing a careless
+# "just skip when unreadable" fix could quietly destroy.
+
+run_case "#1058 guard: readable body genuinely missing Testing still blocks with the section message" \
+  "## Summary
+x
+
+## Glossary
+| t | d |" \
+  2 "missing required '## Testing' section"
+
+run_case "#1058 guard: readable complete body still passes" \
+  "## Summary
+x
+
+## Testing
+y
+
+## Glossary
+| t | d |" \
+  0 ""
+
 # ---- Summary ------------------------------------------------------------
 
 echo ""
--- a/.claude/hooks/tests/test_validate_pr_required_sections.sh
+++ b/.claude/hooks/tests/test_validate_pr_required_sections.sh
@@ -186,6 +186,36 @@ run_case_missing_file() {
   PASS=$((PASS+1))
 }
 
+# Same as run_case_missing_file, but the caller supplies the (nonexistent)
+# path verbatim — used to exercise paths containing shell/printf metacharacters.
+run_case_missing_file_path() {
+  local label="$1" missing="$2" want_rc="$3" want_stderr_regex="$4" reject_stderr_regex="${5:-}"
+  local sb; sb=$(make_sandbox)
+  mock_gh_install "$sb"
+  local cmd="gh pr create --repo me2resh/apexyard --title 'chore(#113): test' --body-file $missing"
+  local input; input=$(jq -nc --arg c "$cmd" '{tool_input:{command:$c}}')
+  local got_stderr got_rc
+  got_stderr=$(cd "$sb" && echo "$input" | bash .claude/hooks/validate-pr-create.sh 2>&1 >/dev/null)
+  got_rc=$?
+  rm -rf "$sb"
+
+  if [ "$got_rc" != "$want_rc" ]; then
+    echo "FAIL [$label]: want rc=$want_rc, got $got_rc (stderr: ${got_stderr:0:200})" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  if [ -n "$want_stderr_regex" ] && ! echo "$got_stderr" | grep -qE "$want_stderr_regex"; then
+    echo "FAIL [$label]: stderr did not match /$want_stderr_regex/" >&2
+    echo "    stderr: $got_stderr" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  if [ -n "$reject_stderr_regex" ] && echo "$got_stderr" | grep -qE "$reject_stderr_regex"; then
+    echo "FAIL [$label]: stderr WRONGLY matched /$reject_stderr_regex/" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  echo "PASS [$label]"
+  PASS=$((PASS+1))
+}
+
 # The core defect: no section claim may be made about an unread file.
 run_case_missing_file "#1058: unreadable body-file does NOT claim '## Testing' is missing" \
   2 "" "missing required '## Testing' section"
@@ -207,6 +237,28 @@ run_case_missing_file "#1058: the block names the unreadable body-file as the ca
 
 run_case_missing_file "#1058: the block quotes the offending path" \
   2 "PR body file could not be read:.*definitely-not-here-1058\.md" ""
+
+# The message must NAME the sections it did not verify. An earlier draft said
+# "the sections above", which referred to nothing — this branch replaces the
+# per-section list rather than following it.
+run_case_missing_file "#1058: the block names the sections it could not verify" \
+  2 "UNVERIFIED, not missing: ## Testing, ## Glossary" ""
+
+# A '%' in the body-file path must survive into the message. `printf "$ERRORS"`
+# treated the accumulated text as a FORMAT STRING, so a path containing '%s'
+# or a trailing '%' printed a DIFFERENT path and silently swallowed the
+# guidance lines after it. Caught by Rex on PR #1060; fixed with `printf '%b'`.
+# A typo'd path is precisely the scenario this whole branch exists to serve,
+# so a path-shaped typo corrupting the message is not acceptable.
+run_case_missing_file_path "#1058: a '%' in the body-file path does not corrupt the message" \
+  '/tmp/nope%s-100%.md' 2 "PR body file could not be read: /tmp/nope%s-100%\.md" ""
+
+# Uses the SAME path shape as the case above deliberately: a lone '%d' still
+# left the guidance lines intact pre-fix (printf substituted 0 and carried on),
+# so it proved nothing. The '%s' + trailing-'%' shape is what actually
+# truncated the format and swallowed everything after it.
+run_case_missing_file_path "#1058: guidance lines survive a '%' path" \
+  '/tmp/nope%s-100%.md' 2 "Fix the path \(check for a typo" ""
 
 # ---- No-regression guards for #1058 -------------------------------------
 # The fix must not make the hook permissive. A body the hook CAN read and
--- a/.claude/hooks/validate-pr-create.sh
+++ b/.claude/hooks/validate-pr-create.sh
@@ -499,6 +499,13 @@ fi
 BODY_CONTENT=""
 # Extract --body-file path. Handles --body-file and the -F short form.
 # After continuation normalization (above) the command is one logical line.
+# me2resh/apexyard#1058 (residual of #1048): "I could not read the body file"
+# and "the body file has no required sections" are DIFFERENT outcomes, and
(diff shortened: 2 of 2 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1060

**Commit**: `1799d1c272e65b8f4c4f581107364e70db07e041`

### Summary

Splits "I could not read the body file" from "the body file has no required sections" in `validate-pr-create.sh`, so the hook stops reporting `## Testing` / `## Glossary` as missing from a file it never opened. Residual of #1048 (reported by @Ref34t); reuses the `BODY_FILE_UNREADABLE` shape #1039/#1040 introduced in the sibling `block-private-refs-in-public-repos.sh`. 35 lines of hook, 95 of test.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (discriminating claim independently reproduced — see below)
- Security: **Pass** (fail-closed preserved; verified by control flow *and* mutation)
- Performance: **N/A**
- Summary Bullet Narrative: **Pass** (advisory check — every bullet carries a why)
- Technical Decisions (AgDR): **N/A** — bug fix reusing an established in-repo pattern; nothing on the `agdr-decisions.md` threshold table fires. Agree with the author.
- Adopter Handbooks: **Pass** (3 always-load handbooks; no violations — details below)
- Fallow Static Analysis (JS/TS): **N/A** — shell-only diff

### What I verified independently (not taken from the PR body)

**1. Fail-closed — confirmed, by control flow and by mutation.** In the unreadable branch the hook still appends to `ERRORS`, and `ERRORS` non-empty reaches `exit 2` at line 758. Two mutations run against the new hook with the new tests:

| Mutation | Result |
|---|---|
| Unreadable branch skips silently (the tempting "permissive" fix) | **5 FAIL** — caught by `unreadable body-file blocks (fail closed, not permissive)` plus the two "does NOT claim missing" cases |
| `BODY_FILE_UNREADABLE=1` unconditionally (over-flagging) | **6 FAIL** — caught by `guard: readable body genuinely missing Testing still blocks with the section message` and 5 pre-existing cases |

Both guards genuinely guard; the no-regression pair is not decorative.

**2. Discriminating claim — reproduced.** Worktree at `HEAD~1` (hook reverted, new tests copied in): **11 PASS / 4 FAIL**, i.e. exactly 4 of the 7 new cases fail against the unmodified hook — the two "must not claim a section is missing" assertions and the two asserting the new `PR body file could not be read` message. Matches the PR's stated RED phase. The retightening was the right call: the original WARN-matching versions would indeed have passed both sides.

**3. The `printf '%b'` round-trip — behaviour-preserving. This is a clean bill of health, not an unexamined one.** I ran the old and new hooks side by side against four configured `.pr.required_sections` values and diffed the output:

| Section name | Old hook | New hook |
|---|---|---|
| `Given / When / Then` (spaces + slashes) | correct, intact | **identical** |
| `Back\tslash` | tab-expanded in the message | **identical** |
| `Two\nLines` | split into two error lines | **identical** |
| `\cTesting` (pathological) | fails open, rc=0 | **identical** |

Byte-identical in every case. The reason the `%b` hop is harmless is that the *final* emitter `printf "$ERRORS"` (line 754, unchanged) already interprets backslash escapes, so a name containing them was already being mangled pre-fix — the new hop doesn't add a corruption, it re-applies an existing one to the same effect. Spaces and regex metachars (the realistic case the escaping code path exists for) survive untouched. Nothing to fix here.

**4. `BODY_FILE_UNREADABLE` initialisation — safe on every path.** Set at line 508, unconditional top-level, before the extractor and before any `if`. It is `0` even when no `--body-file` is present at all. The hook does not use `set -u`, so nothing would have errored anyway, but the initialisation is correct regardless.

**5. Regression surface.** All 6 sibling suites touching this hook pass at HEAD (`flag_extractor_body_file`, `pr_create_gate_anchor`, `pr_create_tilde_cd_target`, `validate_pr_create_head`, `validate_pr_create_structural_parse`, `validate_pr_create_upstream`). `shellcheck --severity=warning` clean on both files.

### On the extractor defect you hit while opening this PR — your reading is correct

It is **pre-existing and unrelated to this diff**. Evidence:

- `git diff HEAD~1 HEAD` touches no `+`/`-` line containing the extractor; the `sed -nE 's/.*--body-file[[:space:]]+([^[:space:]]+).*/\1/p'` line is byte-identical in both revisions (line 502 → 509, moved only by the 7 added comment lines).
- `git blame` puts it at `620da469`, 2026-06-28 — well before this branch.
- The mechanism is the greedy leading `.*`, which forces the match to the **last** `--body-file` token in the command string, so a later mention in commit prose wins over the real flag. Same family as #1038/#1048, same root cause class as the quote-stripping fix in #1041.

So: not introduced here, and correctly out of scope. Worth its own ticket — the greedy-`.*`-takes-the-last-occurrence shape affects every flag extractor in this file, not just `--body-file`.

### Issues Found

None blocking.

### Suggestions

**S1 — `the sections above` is now a dangling reference (on-topic, cheap).** The new message reads:

```
PR body file could not be read: /path/body.md
  The required-section check was NOT run — the sections above may well be present.
```

After this fix nothing lists any sections above that line — the WARN above it names the path, not the sections. The reader is pointed at content the PR deliberately stopped printing, and is never told *which* sections are required. `MISSING_SECTIONS` is in scope right there, so naming them is nearly free:

```
The required-section check was NOT run — the required sections (Testing, Glossary) may well be present.
```

Given that the entire deliverable here is message accuracy, this is the finding I'd most want folded in.

**S2 — a `%` in the body-file path corrupts the new message.** `printf "$ERRORS"` (line 754) treats `ERRORS` as a *format string* — a pre-existing SC2059 on an unchanged line, but this diff newly feeds it a user-controlled filesystem path. Observed with `--body-file /tmp/nope%s-100%.md`:

```
PR VALIDATION BLOCKED:
validate-pr-create.sh: line 754: printf: `m': invalid format character
PR body file could not be read: /var/.../nope-100
```

The path is truncated to a *different* path, and both guidance lines are swallowed. It still exits 2, so no fail-open — but a typo'd path is precisely the scenario this PR targets, so a mangled path defeats it. One-line fix, which also retires the pre-existing `$TITLE` exposure on the same line:

```bash
printf '%b' "$ERRORS" >&2   # %b expands \n in the argument; does NOT interpret % in it
```

Worth a test case with a `%` in the sandbox path if you take it.

**S3 (nit) — a whitespace-only body file is now reported as unreadable.** `printf '\n\n\n' > body.md` trips the new `[ -z "$BODY_CONTENT" ] && [ -s "$BODY_FILE" ]` branch (`$(cat)` strips trailing newlines; `-s` sees size > 0), producing `PR body file could not be read: .../body.md` for a file that reads fine. Pre-fix it said "missing required '## Testing' section", which was *right* for that file — so this is a small inversion of the same blame-the-reader class the PR fixes. I'm rating it a nit rather than a finding because the shape is copied faithfully from `block-private-refs-in-public-repos.sh:308`, and consistency across the two hooks has its own value. If you'd rather be exact, ask `cat` instead of inferring:

```bash
if ! BODY_CONTENT=$(cat "$BODY_FILE" 2>/dev/null); then BODY_FILE_UNREADABLE=1; fi
```

**S4 (nit) — bare `cat "$BODY_FILE"`** has no `2>/dev/null`, unlike the sibling hook's line 306. On a permissions failure `cat`'s own error lands in hook stderr ahead of the new message. Cosmetic; folds into S3's rewrite for free.

### Handbook Findings

Loaded (path-convention): `handbooks/architecture/clean-architecture-layers.md`, `handbooks/architecture/migration-safety.md` *(blocking)*, `handbooks/general/commit-message-quality.md`. Language bucket not triggered (no TS/Py/Go/Rust in diff); `handbooks/domain/` has no handbooks; no private custom-handbooks dir on this fork.

- **Migration Safety (blocking)** — N/A, no migration files in the diff.
- **Clean Architecture Layers** — N/A, shell hook.
- **Commit Message Quality** — Pass. Subject is imperative and specific, body carries the WHY (the misdiagnosis and why blocking is still correct), notes the sibling precedent, single `Refs #1058`. No finding.

### Verdict

**APPROVED**

Fail-closed is intact and proven so by mutation, the refactor is byte-for-byte behaviour-preserving on the inputs most likely to break it, and the tests are demonstrably discriminating — I reproduced the 4-fail RED phase myself rather than taking it on trust. S1 and S2 are message-quality improvements on a PR whose whole subject is message quality; both are one-liners and I'd happily see them as a follow-up rather than a re-review round.

---
📌 Reviewed commit: `1799d1c272e65b8f4c4f581107364e70db07e041`

atlas-apex (commented):
## Code Review: PR #1060 (re-review at new head)

**Commit**: `91de6793ae7b4c356889f146c6720083b3dd82c1`
**Previously approved**: `1799d1c` — this re-review covers follow-up commit `91de679`.

### Summary

`91de679` closes the two findings I raised on `1799d1c`: the dangling "the sections above" reference (S1) and the `printf "$ERRORS"` format-string bug (S2), plus 3 new cases and a `run_case_missing_file_path` helper. Both fixes are correct, and the second is a wider win than the PR claims.

### Checklist Results

- Summary Bullet Narrative:   Pass (narrative, bold-lead + why-it-matters)
- Technical Decisions (AgDR): N/A — bug fix, no material decision
- Adopter Handbooks:          N/A — no findings (see below)
- Fallow (JS/TS):             N/A — shell-only diff, CLI absent

---

### 1. `printf '%b'` on the shared emitter — verified safe, and it fixes more than claimed

This was the highest-risk line in the commit (single emitter for every error the hook produces), so I verified it directly rather than by inspection. I extracted all six existing `ERRORS` shapes and compared `printf "$X"` against `printf '%b' "$X"` byte-for-byte:

| Message shape (line) | Result |
|---|---|
| title-format (273) | identical |
| accepted-types (274) | identical |
| section-missing (612) | identical |
| multi-close (696) | identical |
| branch-missing-id (754) | identical |
| all three combined | identical |

**No existing message changes.** Every `ERRORS` string terminates in `\n` and `%b` expands those exactly as the bare form did. Nothing depended on format-string behaviour — `printf "$ERRORS"` was called with **zero arguments**, so every conversion spec it ever consumed was a latent bug, never a feature. The change is strictly safe.

**New finding (positive) — the fix repairs two pre-existing corruptions beyond this PR's own branch.** Worth recording because the PR undersells it as latent-until-#1058:

- **PR title containing `%`** (line 273). `fix(#42): bump coverage to 90%` rendered as `...to 90 0oesn't match format...` — the `%'` ate the following text. Realistic and user-facing.
- **Branch name containing `%`** (line 754). `%` is legal in git ref names; `feature/100%-done` rendered as `feature/1000one`.

Both now render correctly. Consider adding a line to the PR body — "also fixes `%` corruption on the title and branch-name error paths" is a stronger claim than what's there and it's true.

**Informational, not a request:** `%b` still expands backslash escapes in its *argument*, so a Windows-style path (`C:\temp\body.md`) would render `\t` as a tab. Identical under the old form — not introduced here — but relevant given `cd7f6e6` just landed drive-letter path handling. Fully literal output needs `printf '%s'` plus real newlines, which is a bigger refactor than this PR should carry.

### 2. The `_unchecked` pipeline — correct on the cases that matter

Exercised directly:

| Input | Output | Verdict |
|---|---|---|
| exactly ONE missing | `## Testing` | correct, no trailing separator |
| two missing | `## Testing, ## Glossary` | correct |
| spaces + slashes (`Given / When / Then`) | `## Given / When / Then, ## Glossary` | uncorrupted |
| empty | (empty) | correct |
| three missing | `## A, ## B, ## C` | correct |

Confirmed end-to-end too: single-section config yields `UNVERIFIED, not missing: ## Glossary`, exit 2.

**`_unchecked` can never be referenced unset** — assigned unconditionally on the line immediately preceding its sole use, inside the same branch; only 2 occurrences in the file; no `set -u` in the hook regardless. Naming follows the file's `_`-prefix convention (`_stripped`, `_cmd_head`). `paste -sd` is established in the hook set (5 other files), so no new portability surface.

### Handbook Findings

Loaded via path-convention: `handbooks/architecture/clean-architecture-layers.md`, `handbooks/architecture/migration-safety.md` (**ENFORCEMENT: blocking**), `handbooks/general/commit-message-quality.md`. No `language/` bucket applies (shell-only diff); no `domain/` tree exists.

**No violations.** The blocking migration-safety handbook is not engaged — no migration in this diff. Against `general/commit-message-quality.md`, `91de679`'s message is a positive example: it names the defect, the mechanism, and the evidence, and volunteers that an earlier draft's test was worthless.

### Issues Found

None blocking.

### Suggestions

**N1 — nit (non-blocking): a comma inside a configured section name is cosmetically corrupted.** With `required_sections: ["Testing","Risks, Rollback"]` the message renders:

```
UNVERIFIED, not missing: ## Testing, ## Risks,  Rollback
```

Two artifacts: a double space (the `s/,/, /g` rewrite hits the intra-name comma), and the name's own comma is now indistinguishable from the separator. Verified end-to-end.

Severity is genuinely low: the input is adopter-controlled config (`.pr.required_sections`), not attacker- or author-supplied; it affects an advisory message only; the block outcome is unchanged. If you want it, join on a character that can't appear in a heading rather than post-processing commas:

```sh
_unchecked=$(printf '%b' "$MISSING_SECTIONS" | sed '/^$/d' | sed 's/^/## /' | paste -sd'\037' - | sed 's/\037/, /g')
```

Equally fine to leave — a heading with a comma in it is unusual, and the message stays readable.

**N2 — informational:** a section name containing a literal `\c` would truncate the message and `\t` becomes a tab (`%b` semantics on `MISSING_SECTIONS`). Pre-existing — line 614's readable-body branch already does the same `printf '%b' "$MISSING_SECTIONS"` — so consistent, not introduced. Noting only so it isn't rediscovered later.

### S3 / S4 — my call on the deliberate omissions

**S3 (whitespace-only body reported as unreadable) — agree, leave it.** It's also narrower than I first stated, and my original framing was wrong: `$(cat)` strips only *trailing* newlines, so a file of `\n\n   \n` retains `   ` and is **not** flagged. Only a file of *pure newlines* trips the branch. In that degenerate case the file has no content and no sections; "could not be read" is a slightly wrong label on an outcome (block) that is right, and the author's corrective action — put a real body in the file — is identical either way. Not worth diverging from the sibling for.

**S4 (missing `2>/dev/null` on the `cat`) — keep the omission, but the stated rationale is wrong and should be corrected.** The sibling *does* have the redirect:

```sh
# block-private-refs-in-public-repos.sh:306
BODY_FILE_CONTENT=$(cat "$BODY_FILE" 2>/dev/null)
```

So "faithful to `block-private-refs-in-public-repos.sh`'s shape" argues for **adding** it, not omitting it. I still don't want it folded in — for a better reason. I confirmed the leaked line is:

```
cat: /path/to/body.md: Permission denied
```

That is the single most specific piece of evidence available about *why* the read failed, and this PR's entire subject is telling the author the true cause instead of a fabricated one. Suppressing it would make the hook's diagnosis strictly less informative in exactly the scenario the PR exists to serve. Keep it — but please record it as a **deliberate divergence** from the sibling (a one-line code comment, or just drop the fidelity claim), so a future reader doesn't "restore consistency" and silently delete the diagnostic.

### Verification performed

**Discriminating-test claim: reproduced.** Reverted *only* the hook to `1799d1c`, kept tests at HEAD → **15 pass / 3 fail**, and the 3 failures are exactly the 3 new cases. Pre-fix output matches the commit message precisely:

```
.claude/hooks/validate-pr-create.sh: line 754: printf: `m': invalid format character
PR body file could not be read: /tmp/nope-100
```

Wrong path printed (`%s` consumed empty, then `%.m` invalid), guidance line entirely swallowed. The commit's admission about the discarded lone-`%d` draft also checks out — a lone `%d` substitutes `0` and leaves the remainder intact, so it would indeed have passed pre-fix and proved nothing. Good call retargeting it.

Everything else at the new head:

| Check | Result |
|---|---|
| Primary suite | 18 pass / 0 fail |
| Fail-closed preserved | exit 2 on unreadable body, confirmed end-to-end |
| No-regression guards | both pass (readable-but-missing still blocks with section message; readable-complete still passes) |
| `shellcheck --severity=warning` (both files) | clean |
| 6 sibling suites | all pass (22 / 21 / 2 / 13 / 7 / 7) |
| Full `bin/run-hook-tests.sh` | 117 pass / 3 fail — **all 3 pre-existing** |

The 3 full-suite failures (`test_suggest_mcp_search`, `test_token_efficiency_wave1`, `test_harnessability_scoring`) reproduce with this PR's hook change reverted, and none of the three suites reference `validate-pr-create` at all. The PR's "no new failures against the pre-change baseline" claim holds.

### Verdict

**APPROVED**

Both findings from the previous review are properly closed, the shared-emitter change is verified byte-identical on every existing path (and repairs two latent corruptions besides), and the new tests are genuinely discriminating. N1 and N2 are non-blocking; the only thing I'd actually like changed is the S4 *comment*, not the S4 *code*.

---
Reviewed commit: `91de6793ae7b4c356889f146c6720083b3dd82c1`

Issue #1033, opened 2026-07-27
[Bug] Trust-chain libs can load from the cwd's repo via the git rev-parse fallback (follow-up to #1028)


atlas-apex:
Follow-up from the security review of #1028 (`fix(#1025): tracker_review_submit returns 0 without posting under zsh`), which merged as `4c24c2e`. Raised as **MEDIUM**, explicitly non-blocking at the time — filing so it isn't lost.

## Given / When / Then

**Given** the zsh fix added a `git rev-parse --show-toplevel` fallback to the self-location logic of nine trust-chain libraries under `.claude/hooks/`
**When** a library is sourced under a non-bash shell (so `BASH_SOURCE` is unavailable and the fallback fires) from a cwd inside a *different* git repository — e.g. a `workspace/<project>` clone, or any repo a user happens to be sitting in
**Then** the fallback resolves `$(git rev-parse --show-toplevel)/.claude/hooks/<lib>.sh` against **that** repo, so a managed project's own copy of a trust-chain library would be sourced — expected that trust-chain libraries only ever load from the ops fork

## Why it's MEDIUM rather than HIGH

- **Unreachable on the mechanical gate path.** Hooks are invoked via `bash -c` from `.claude/settings.json` and source by absolute `$HOOK_DIR`, so `BASH_SOURCE` resolves and the fallback never fires there.
- Three conditions must line up: non-bash sourcing shell, cwd inside another repo, and that repo shipping a same-named library.
- The pattern **pre-existed** at two call sites; #1028 widened it to nine rather than introducing it.

## Why it's still worth closing

The framework already argues against this primitive in its own source:

- `.claude/hooks/_lib-read-config.sh` (~line 57) says to locate via `BASH_SOURCE` "not via `git rev-parse` (which would defeat the whole point of this fix)".
- `.claude/hooks/_lib-ops-root.sh`'s header documents the workspace-clone failure mode and the #381 hijack incident that motivated pin-first, anchor-validated `resolve_ops_root`.

So the codebase has both the reasoning and the better primitive already; the zsh fix just reached for the weaker one.

## Suggested fix

Add a ~3-line anchor check before sourcing from the git-derived root — require `.apexyard-fork`, or the `onboarding.yaml` + `apexyard.projects.yaml` pair — and refuse otherwise. That keeps the zsh fix working while declining to load a trust-chain library out of a managed-project clone. Preferring `resolve_ops_root`'s pin-first resolution outright would be the more thorough option.

## Severity

[...]

PR #1061, opened 2026-07-29
fix(#1033): anchor the git-derived self-location fallback to an apexyard fork

atlas-apex:
## Summary

- **Nine trust-chain self-location sites now refuse a git-derived root that isn't an apexyard fork.** `#1028` made those sites zsh-safe by falling back to `git rev-parse --show-toplevel` when `BASH_SOURCE` is unavailable. The fallback had no anchor check, so it accepted any git repository that also happened to ship a same-named trust-chain library at that path — a `workspace/<project>` clone, or any unrelated checkout the operator's cwd happened to be inside. Each site now additionally requires `.apexyard-fork`, or the `onboarding.yaml` + `apexyard.projects.yaml` pair.
- **The framework already argued against this primitive in its own source.** `_lib-read-config.sh` says to locate via `BASH_SOURCE` *"not via `git rev-parse` (which would defeat the whole point of this fix)"*, and `_lib-ops-root.sh` documents the #381 hijack incident that motivated pin-first, anchor-validated resolution. The zsh fix reached for the weaker primitive; this restores the intent.
- **The guard is inlined at each site, deliberately, not extracted into a helper.** These libraries *are* the bootstrap layer — a shared helper would itself need locating, which is the problem being solved. Nine copies of a two-condition check is the honest shape.
- **The #1028 zsh fix is preserved, and that is the load-bearing constraint.** Written too strictly, this change would undo it and `tracker_review_submit` would silently stop posting reviews again — the exact failure #1028 existed to repair. Eighteen positive cases (9 sites x 2 fork layouts) pin that a real fork still resolves.

## Scope of the change

| File | Sites |
|---|---|
| `_lib-tracker.sh` | 2 (config lib, portfolio lib) |
| `_lib-audit-history.sh` | 1 |
| `_lib-extract-pr.sh` | 1 |
| `_lib-fresh-fork.sh` | 1 |
| `_lib-onboarding-depth-mode.sh` | 1 |
| `_lib-onboarding-glossary-seen.sh` | 1 |
| `_lib-premium-hook.sh` | 1 |
| `_lib-project-board.sh` | 1 |

Nine sites, eight files. `_lib-premium-hook.sh` is the odd one out — its fallback resolves the lib's **own** directory rather than sourcing a sibling, so it has no load side effect to observe and is asserted on the returned path instead.

## Testing

New file: `.claude/hooks/tests/test_lib_self_location_anchor.sh` — **27 cases** (9 negative + 9 positive × 2 fork layouts).

**Discriminating check.** Against the unmodified libraries: **9 FAIL / 18 PASS.** All nine sites loaded an impostor library out of an unrelated repo; both no-regression guards already passed. After the change: **27/27**.

How the fallback is triggered portably: the real trigger is a non-bash sourcing shell (zsh leaves `BASH_SOURCE` unset). Rather than depend on zsh being installed on a CI runner, each lib is sourced from a stream — `bash -c '. /dev/stdin' < lib.sh` — so `BASH_SOURCE[0]` is `/dev/stdin`, its dirname is `/dev`, the sibling-exists guard fails, and the rev-parse fallback fires. Identical code path, no zsh dependency.

The fixture builds a git repo that is **not** a fork but ships its own `.claude/hooks/` with same-named libs, each setting a tripwire variable. A tripwire set after loading means the wrong copy was sourced.

**Four defects in this test were found and fixed before it was trustworthy**, recorded because each produced a *green* result against vulnerable code:

1. The first draft covered 8 of the 9 sites while claiming all of them — `_lib-premium-hook.sh` had zero cases.
2. Its probe compared `$PWD` against a git-resolved path. On macOS `mktemp -d` returns `/var/folders/…` while git resolves `/private/var/folders/…`, so the comparison never matched and reported PASS on a vulnerable site.
3. The fixture never created `_lib-premium-hook.sh` in the impostor repo, so that site's existence check failed and the fallback never fired — a fixture gap indistinguishable from a fix.
4. The sandbox used `git add -A`, which this framework's own `block-git-add-all.sh` correctly refuses.

The structural tell for (2) and (3) was a lone PASS sitting among eight FAILs in a RED phase. In a security test, an unexplained pass is a reason to check the probe.

**Verified against this fork specifically** (split-portfolio v2 — `.apexyard-fork` present, v1 pair absent): both `_tracker_load_config_lib` and `_tracker_load_portfolio_lib` still resolve. A v1-only anchor would have broken every hook on a v2 fork.

Full suite: no new failures against the pre-change baseline. `shellcheck --severity=warning` clean on all nine changed files.

## What this is, stated accurately

**This narrows an accident surface. It is not an access-control boundary**, and should not be described as one. The anchors are unauthenticated, presence-only files, and `-f` follows symlinks — Hakim demonstrated `ln -s /etc/hosts .apexyard-fork` satisfying the guard, and I reproduced it. Against an adversary who can already write into the candidate root, this raises the bar by one `touch`.

What it genuinely does is convert a **broad, silent, accidental** misresolution into a **narrow, deliberate** one. The realistic failure was never a targeted attacker; it was a cwd sitting inside a `workspace/<project>` clone while an agent sourced a trust-chain lib and silently got the wrong code. That surface was *any repo shipping a same-named lib under `.claude/hooks/`*; it is now *only such a repo that is also deliberately marked as a fork*. (Hakim caught me overstating this as "every git repo on the machine" — the sibling-exists co-condition was always present. Correcting it here; the nine code comments carry the same overstatement and are folded into #1062, which edits those sites anyway.) The code comments say this plainly at all nine sites.

**Reachability is unchanged and low.** Hooks launch via `bash -c` with absolute paths and a bash shebang, so `BASH_SOURCE` resolves and the fallback never fires there. Verified independently by both reviewers: no merge, ticket, migration, marker, or CI gate can reach these nine fallbacks. MEDIUM is the right rating.

**Known limit, pre-existing, not fixed here:** the anchor guards only the *fallback* branch. Under a non-bash shell `dirname ""` → `.`, so if the cwd *directly contains* the sibling filename the `if` body never runs and the anchor is never consulted. Filed as a follow-up rather than widened into this PR.

## Review

Trust-chain change — Heavy tier per `.claude/rules/right-size-ceremony.md` rail 1. Reviewed by both Rex (correctness) and Hakim (security, 85/100).

Refs #1033

---

## Glossary

| Term | Definition |
|------|------------|
| Trust chain | The hooks and libraries that mechanically enforce the merge, ticket, and marker gates |
| Self-location | A library resolving the directory it lives in, so it can source its siblings |
| Anchor | A marker file proving a directory is an apexyard fork — `.apexyard-fork` (v2) or the `onboarding.yaml` + `apexyard.projects.yaml` pair (v1) |
| Impostor repo | Test fixture: a non-fork git repo shipping same-named libraries, used to prove the fallback refuses it |
| Discriminating test | A test proven to fail against the unfixed code, so passing it afterwards is evidence |

### DIFF
--- a/.claude/hooks/_lib-audit-history.sh
+++ b/.claude/hooks/_lib-audit-history.sh
@@ -60,6 +60,9 @@ if [ -z "$_AUDIT_LIB_DIR" ] || [ ! -f "$_AUDIT_LIB_DIR/_lib-read-config.sh" ]; t
   # actually an apexyard fork. Without this the fallback would source a
   # trust-chain library out of ANY repo the cwd happens to be inside --
   # a workspace/<project> clone, or an unrelated checkout.
+# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
+# (marker file over three alternatives; v2 marker with legacy-pair
+# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
   if [ -n "$_audit_root" ] && { [ -f "$_audit_root/.apexyard-fork" ] || { [ -f "$_audit_root/onboarding.yaml" ] && [ -f "$_audit_root/apexyard.projects.yaml" ]; }; } && [ -f "$_audit_root/.claude/hooks/_lib-read-config.sh" ]; then
     _AUDIT_LIB_DIR="$_audit_root/.claude/hooks"
   fi
--- a/.claude/hooks/_lib-audit-history.sh
+++ b/.claude/hooks/_lib-audit-history.sh
@@ -56,7 +56,11 @@ _LIB_AUDIT_HISTORY_SOURCED=1
 _AUDIT_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
 if [ -z "$_AUDIT_LIB_DIR" ] || [ ! -f "$_AUDIT_LIB_DIR/_lib-read-config.sh" ]; then
   _audit_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-  if [ -n "$_audit_root" ] && [ -f "$_audit_root/.claude/hooks/_lib-read-config.sh" ]; then
+  # me2resh/apexyard#1033: only accept a git-derived root that is
+  # actually an apexyard fork. Without this the fallback would source a
+  # trust-chain library out of ANY repo the cwd happens to be inside --
+  # a workspace/<project> clone, or an unrelated checkout.
+  if [ -n "$_audit_root" ] && { [ -f "$_audit_root/.apexyard-fork" ] || { [ -f "$_audit_root/onboarding.yaml" ] && [ -f "$_audit_root/apexyard.projects.yaml" ]; }; } && [ -f "$_audit_root/.claude/hooks/_lib-read-config.sh" ]; then
     _AUDIT_LIB_DIR="$_audit_root/.claude/hooks"
   fi
   unset _audit_root
--- a/.claude/hooks/_lib-audit-history.sh
+++ b/.claude/hooks/_lib-audit-history.sh
@@ -56,13 +56,19 @@ _LIB_AUDIT_HISTORY_SOURCED=1
 _AUDIT_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
 if [ -z "$_AUDIT_LIB_DIR" ] || [ ! -f "$_AUDIT_LIB_DIR/_lib-read-config.sh" ]; then
   _audit_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-  # me2resh/apexyard#1033: only accept a git-derived root that is
-  # actually an apexyard fork. Without this the fallback would source a
-  # trust-chain library out of ANY repo the cwd happens to be inside --
-  # a workspace/<project> clone, or an unrelated checkout.
-# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
-# (marker file over three alternatives; v2 marker with legacy-pair
-# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
+  # me2resh/apexyard#1033: only accept a git-derived root that is actually
+  # an apexyard fork. Without this the fallback sources a trust-chain
+  # library out of ANY repo the cwd happens to be inside -- a
+  # workspace/<project> clone, or an unrelated checkout.
+  #
+  # This narrows an ACCIDENT surface. It is NOT an access-control boundary:
+  # the anchors are unauthenticated presence-only files, and -f follows
+  # symlinks, so anyone able to write to the candidate root can satisfy it.
+  # What it prevents is a cwd-driven misresolution, not a hostile library.
+  # Anchor pair per AgDR-0021 §A/§E -- the same test
+  # resolve_ops_root_walk applies, evaluated against one candidate rather
+  # than a walk. (resolve_ops_root itself is unusable here: three of these
+  # sites are locating _lib-ops-root.sh, and its pin is session-scoped.)
   if [ -n "$_audit_root" ] && { [ -f "$_audit_root/.apexyard-fork" ] || { [ -f "$_audit_root/onboarding.yaml" ] && [ -f "$_audit_root/apexyard.projects.yaml" ]; }; } && [ -f "$_audit_root/.claude/hooks/_lib-read-config.sh" ]; then
     _AUDIT_LIB_DIR="$_audit_root/.claude/hooks"
   fi
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -142,7 +142,11 @@ if ! command -v tracker_kind >/dev/null 2>&1; then
   _lib_extract_pr_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd 2>/dev/null)"
   if [ -z "$_lib_extract_pr_dir" ] || [ ! -f "$_lib_extract_pr_dir/_lib-tracker.sh" ]; then
     _lib_extract_pr_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-    if [ -n "$_lib_extract_pr_root" ] && [ -f "$_lib_extract_pr_root/.claude/hooks/_lib-tracker.sh" ]; then
+    # me2resh/apexyard#1033: only accept a git-derived root that is
+    # actually an apexyard fork. Without this the fallback would source a
+    # trust-chain library out of ANY repo the cwd happens to be inside --
+    # a workspace/<project> clone, or an unrelated checkout.
+    if [ -n "$_lib_extract_pr_root" ] && { [ -f "$_lib_extract_pr_root/.apexyard-fork" ] || { [ -f "$_lib_extract_pr_root/onboarding.yaml" ] && [ -f "$_lib_extract_pr_root/apexyard.projects.yaml" ]; }; } && [ -f "$_lib_extract_pr_root/.claude/hooks/_lib-tracker.sh" ]; then
       _lib_extract_pr_dir="$_lib_extract_pr_root/.claude/hooks"
     fi
     unset _lib_extract_pr_root
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -146,6 +146,9 @@ if ! command -v tracker_kind >/dev/null 2>&1; then
     # actually an apexyard fork. Without this the fallback would source a
     # trust-chain library out of ANY repo the cwd happens to be inside --
     # a workspace/<project> clone, or an unrelated checkout.
+# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
+# (marker file over three alternatives; v2 marker with legacy-pair
+# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
     if [ -n "$_lib_extract_pr_root" ] && { [ -f "$_lib_extract_pr_root/.apexyard-fork" ] || { [ -f "$_lib_extract_pr_root/onboarding.yaml" ] && [ -f "$_lib_extract_pr_root/apexyard.projects.yaml" ]; }; } && [ -f "$_lib_extract_pr_root/.claude/hooks/_lib-tracker.sh" ]; then
       _lib_extract_pr_dir="$_lib_extract_pr_root/.claude/hooks"
     fi
--- a/.claude/hooks/_lib-extract-pr.sh
+++ b/.claude/hooks/_lib-extract-pr.sh
@@ -142,13 +142,19 @@ if ! command -v tracker_kind >/dev/null 2>&1; then
   _lib_extract_pr_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd 2>/dev/null)"
   if [ -z "$_lib_extract_pr_dir" ] || [ ! -f "$_lib_extract_pr_dir/_lib-tracker.sh" ]; then
     _lib_extract_pr_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-    # me2resh/apexyard#1033: only accept a git-derived root that is
-    # actually an apexyard fork. Without this the fallback would source a
-    # trust-chain library out of ANY repo the cwd happens to be inside --
-    # a workspace/<project> clone, or an unrelated checkout.
-# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
-# (marker file over three alternatives; v2 marker with legacy-pair
-# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
+    # me2resh/apexyard#1033: only accept a git-derived root that is actually
+    # an apexyard fork. Without this the fallback sources a trust-chain
+    # library out of ANY repo the cwd happens to be inside -- a
+    # workspace/<project> clone, or an unrelated checkout.
+    #
+    # This narrows an ACCIDENT surface. It is NOT an access-control boundary:
+    # the anchors are unauthenticated presence-only files, and -f follows
+    # symlinks, so anyone able to write to the candidate root can satisfy it.
+    # What it prevents is a cwd-driven misresolution, not a hostile library.
+    # Anchor pair per AgDR-0021 §A/§E -- the same test
+    # resolve_ops_root_walk applies, evaluated against one candidate rather
+    # than a walk. (resolve_ops_root itself is unusable here: three of these
+    # sites are locating _lib-ops-root.sh, and its pin is session-scoped.)
     if [ -n "$_lib_extract_pr_root" ] && { [ -f "$_lib_extract_pr_root/.apexyard-fork" ] || { [ -f "$_lib_extract_pr_root/onboarding.yaml" ] && [ -f "$_lib_extract_pr_root/apexyard.projects.yaml" ]; }; } && [ -f "$_lib_extract_pr_root/.claude/hooks/_lib-tracker.sh" ]; then
       _lib_extract_pr_dir="$_lib_extract_pr_root/.claude/hooks"
     fi
--- a/.claude/hooks/_lib-fresh-fork.sh
+++ b/.claude/hooks/_lib-fresh-fork.sh
@@ -57,6 +57,9 @@ if [ -z "$_FRESH_FORK_LIB_DIR" ] || [ ! -f "$_FRESH_FORK_LIB_DIR/_lib-read-confi
   # actually an apexyard fork. Without this the fallback would source a
   # trust-chain library out of ANY repo the cwd happens to be inside --
   # a workspace/<project> clone, or an unrelated checkout.
+# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
+# (marker file over three alternatives; v2 marker with legacy-pair
+# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
   if [ -n "$_fresh_fork_root" ] && { [ -f "$_fresh_fork_root/.apexyard-fork" ] || { [ -f "$_fresh_fork_root/onboarding.yaml" ] && [ -f "$_fresh_fork_root/apexyard.projects.yaml" ]; }; } && [ -f "$_fresh_fork_root/.claude/hooks/_lib-read-config.sh" ]; then
     _FRESH_FORK_LIB_DIR="$_fresh_fork_root/.claude/hooks"
   fi
--- a/.claude/hooks/_lib-fresh-fork.sh
+++ b/.claude/hooks/_lib-fresh-fork.sh
@@ -53,13 +53,19 @@ _LIB_FRESH_FORK_SOURCED=1
 _FRESH_FORK_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
 if [ -z "$_FRESH_FORK_LIB_DIR" ] || [ ! -f "$_FRESH_FORK_LIB_DIR/_lib-read-config.sh" ]; then
   _fresh_fork_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-  # me2resh/apexyard#1033: only accept a git-derived root that is
-  # actually an apexyard fork. Without this the fallback would source a
-  # trust-chain library out of ANY repo the cwd happens to be inside --
-  # a workspace/<project> clone, or an unrelated checkout.
-# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
-# (marker file over three alternatives; v2 marker with legacy-pair
-# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
+  # me2resh/apexyard#1033: only accept a git-derived root that is actually
+  # an apexyard fork. Without this the fallback sources a trust-chain
+  # library out of ANY repo the cwd happens to be inside -- a
+  # workspace/<project> clone, or an unrelated checkout.
+  #
+  # This narrows an ACCIDENT surface. It is NOT an access-control boundary:
+  # the anchors are unauthenticated presence-only files, and -f follows
+  # symlinks, so anyone able to write to the candidate root can satisfy it.
+  # What it prevents is a cwd-driven misresolution, not a hostile library.
+  # Anchor pair per AgDR-0021 §A/§E -- the same test
+  # resolve_ops_root_walk applies, evaluated against one candidate rather
+  # than a walk. (resolve_ops_root itself is unusable here: three of these
+  # sites are locating _lib-ops-root.sh, and its pin is session-scoped.)
   if [ -n "$_fresh_fork_root" ] && { [ -f "$_fresh_fork_root/.apexyard-fork" ] || { [ -f "$_fresh_fork_root/onboarding.yaml" ] && [ -f "$_fresh_fork_root/apexyard.projects.yaml" ]; }; } && [ -f "$_fresh_fork_root/.claude/hooks/_lib-read-config.sh" ]; then
     _FRESH_FORK_LIB_DIR="$_fresh_fork_root/.claude/hooks"
   fi
--- a/.claude/hooks/_lib-fresh-fork.sh
+++ b/.claude/hooks/_lib-fresh-fork.sh
@@ -53,7 +53,11 @@ _LIB_FRESH_FORK_SOURCED=1
 _FRESH_FORK_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
 if [ -z "$_FRESH_FORK_LIB_DIR" ] || [ ! -f "$_FRESH_FORK_LIB_DIR/_lib-read-config.sh" ]; then
   _fresh_fork_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-  if [ -n "$_fresh_fork_root" ] && [ -f "$_fresh_fork_root/.claude/hooks/_lib-read-config.sh" ]; then
+  # me2resh/apexyard#1033: only accept a git-derived root that is
+  # actually an apexyard fork. Without this the fallback would source a
+  # trust-chain library out of ANY repo the cwd happens to be inside --
+  # a workspace/<project> clone, or an unrelated checkout.
+  if [ -n "$_fresh_fork_root" ] && { [ -f "$_fresh_fork_root/.apexyard-fork" ] || { [ -f "$_fresh_fork_root/onboarding.yaml" ] && [ -f "$_fresh_fork_root/apexyard.projects.yaml" ]; }; } && [ -f "$_fresh_fork_root/.claude/hooks/_lib-read-config.sh" ]; then
     _FRESH_FORK_LIB_DIR="$_fresh_fork_root/.claude/hooks"
   fi
   unset _fresh_fork_root
--- a/.claude/hooks/_lib-onboarding-depth-mode.sh
+++ b/.claude/hooks/_lib-onboarding-depth-mode.sh
@@ -59,13 +59,19 @@ _LIB_ONBOARDING_DEPTH_MODE_SOURCED=1
 _DEPTH_MODE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
 if [ -z "$_DEPTH_MODE_LIB_DIR" ] || [ ! -f "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.sh" ]; then
   _depth_mode_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-  # me2resh/apexyard#1033: only accept a git-derived root that is
-  # actually an apexyard fork. Without this the fallback would source a
-  # trust-chain library out of ANY repo the cwd happens to be inside --
-  # a workspace/<project> clone, or an unrelated checkout.
-# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
-# (marker file over three alternatives; v2 marker with legacy-pair
-# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
+  # me2resh/apexyard#1033: only accept a git-derived root that is actually
+  # an apexyard fork. Without this the fallback sources a trust-chain
+  # library out of ANY repo the cwd happens to be inside -- a
+  # workspace/<project> clone, or an unrelated checkout.
+  #
+  # This narrows an ACCIDENT surface. It is NOT an access-control boundary:
+  # the anchors are unauthenticated presence-only files, and -f follows
+  # symlinks, so anyone able to write to the candidate root can satisfy it.
+  # What it prevents is a cwd-driven misresolution, not a hostile library.
+  # Anchor pair per AgDR-0021 §A/§E -- the same test
+  # resolve_ops_root_walk applies, evaluated against one candidate rather
+  # than a walk. (resolve_ops_root itself is unusable here: three of these
+  # sites are locating _lib-ops-root.sh, and its pin is session-scoped.)
   if [ -n "$_depth_mode_root" ] && { [ -f "$_depth_mode_root/.apexyard-fork" ] || { [ -f "$_depth_mode_root/onboarding.yaml" ] && [ -f "$_depth_mode_root/apexyard.projects.yaml" ]; }; } && [ -f "$_depth_mode_root/.claude/hooks/_lib-ops-root.sh" ]; then
     _DEPTH_MODE_LIB_DIR="$_depth_mode_root/.claude/hooks"
   fi
--- a/.claude/hooks/_lib-onboarding-depth-mode.sh
+++ b/.claude/hooks/_lib-onboarding-depth-mode.sh
@@ -59,7 +59,11 @@ _LIB_ONBOARDING_DEPTH_MODE_SOURCED=1
 _DEPTH_MODE_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
 if [ -z "$_DEPTH_MODE_LIB_DIR" ] || [ ! -f "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.sh" ]; then
   _depth_mode_root="$(git rev-parse --show-toplevel 2>/dev/null)"
-  if [ -n "$_depth_mode_root" ] && [ -f "$_depth_mode_root/.claude/hooks/_lib-ops-root.sh" ]; then
+  # me2resh/apexyard#1033: only accept a git-derived root that is
+  # actually an apexyard fork. Without this the fallback would source a
+  # trust-chain library out of ANY repo the cwd happens to be inside --
+  # a workspace/<project> clone, or an unrelated checkout.
+  if [ -n "$_depth_mode_root" ] && { [ -f "$_depth_mode_root/.apexyard-fork" ] || { [ -f "$_depth_mode_root/onboarding.yaml" ] && [ -f "$_depth_mode_root/apexyard.projects.yaml" ]; }; } && [ -f "$_depth_mode_root/.claude/hooks/_lib-ops-root.sh" ]; then
     _DEPTH_MODE_LIB_DIR="$_depth_mode_root/.claude/hooks"
   fi
   unset _depth_mode_root
--- a/.claude/hooks/_lib-onboarding-depth-mode.sh
+++ b/.claude/hooks/_lib-onboarding-depth-mode.sh
@@ -63,6 +63,9 @@ if [ -z "$_DEPTH_MODE_LIB_DIR" ] || [ ! -f "$_DEPTH_MODE_LIB_DIR/_lib-ops-root.s
   # actually an apexyard fork. Without this the fallback would source a
   # trust-chain library out of ANY repo the cwd happens to be inside --
   # a workspace/<project> clone, or an unrelated checkout.
+# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
+# (marker file over three alternatives; v2 marker with legacy-pair
+# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
   if [ -n "$_depth_mode_root" ] && { [ -f "$_depth_mode_root/.apexyard-fork" ] || { [ -f "$_depth_mode_root/onboarding.yaml" ] && [ -f "$_depth_mode_root/apexyard.projects.yaml" ]; }; } && [ -f "$_depth_mode_root/.claude/hooks/_lib-ops-root.sh" ]; then
     _DEPTH_MODE_LIB_DIR="$_depth_mode_root/.claude/hooks"
   fi
--- a/.claude/hooks/_lib-onboarding-glossary-seen.sh
+++ b/.claude/hooks/_lib-onboarding-glossary-seen.sh
@@ -83,6 +83,9 @@ if [ -z "$_GLOSSARY_SEEN_LIB_DIR" ] || [ ! -f "$_GLOSSARY_SEEN_LIB_DIR/_lib-ops-
   # actually an apexyard fork. Without this the fallback would source a
   # trust-chain library out of ANY repo the cwd happens to be inside --
   # a workspace/<project> clone, or an unrelated checkout.
+# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
+# (marker file over three alternatives; v2 marker with legacy-pair
+# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
   if [ -n "$_glossary_seen_root" ] && { [ -f "$_glossary_seen_root/.apexyard-fork" ] || { [ -f "$_glossary_seen_root/onboarding.yaml" ] && [ -f "$_glossary_seen_root/apexyard.projects.yaml" ]; }; } && [ -f "$_glossary_seen_root/.claude/hooks/_lib-ops-root.sh" ]; then
     _GLOSSARY_SEEN_LIB_DIR="$_glossary_seen_root/.claude/hooks"
   fi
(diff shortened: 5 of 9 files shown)


atlas-apex (commented):
## Code Review: PR #1061

**Commit**: `e2283d49e526655117e343ccf770db74d750d756`

### Summary

Adds a fork-anchor check to the nine `git rev-parse --show-toplevel` self-location fallbacks that #1028 introduced, so a trust-chain library can no longer be sourced out of an arbitrary git repository the caller's cwd happens to sit inside. I verified every priority independently rather than taking the PR body's claims at face value; all of the load-bearing ones hold. One genuine gap remains, and it is in the test rather than the code.

### Checklist Results
- Testing:                    Pass (one non-blocking coverage gap, below)
- Technical Decisions (AgDR): N/A — see the AgDR section
- Adopter Handbooks:          Pass (no findings)

### Verification performed

**1. Guard correctness at all nine sites — PASS.** Read individually, not sampled. Each site's guard references that site's own root variable (`_audit_root`, `_lib_extract_pr_root`, `_fresh_fork_root`, `_depth_mode_root`, `_glossary_seen_root`, `_premium_root`, `_lib_board_root`, and `root` twice in `_lib-tracker.sh`) — no cross-contamination from the scripted transform. Precedence resolves as intended:

```
[ -n "$R" ] && { fork || { onb && proj; }; } && [ -f "$R/<sibling>" ]
```

`&&`/`||` are equal-precedence and left-associative in POSIX shell, so the explicit `{ ...; }` grouping is doing necessary work and is doing it correctly — the inner group binds `onboarding.yaml && apexyard.projects.yaml` before the `||`. Every brace group is terminated with the required `;`. Normalising away the variable name yields exactly six distinct strings differing only in the sibling lib being probed, which is the expected shape for nine mechanically-applied copies of one guard.

**2. #1028 not undone — PASS, on both layouts.** This is the regression that mattered and the PR only pins it with two cases on one lib, so I constructed the rest.

| Layout | Anchor | Result |
|---|---|---|
| split-portfolio v2 (this fork) | `.apexyard-fork`, v1 pair absent | config lib + portfolio lib resolve |
| legacy v1 (constructed) | `onboarding.yaml` + `apexyard.projects.yaml`, no `.apexyard-fork` | config lib + portfolio lib resolve; all five other sibling-sourcing sites resolve their target function |

`test_tracker_zsh_self_location.sh` (#1028's own regression test) still passes 9/0, including its "gh mock was actually invoked" case — the specific silent-no-op that #1028 existed to repair.

Worth recording explicitly because it is the thing that would have made this change dangerous: **`.apexyard-fork` is tracked at the repo root**, so worktrees, fresh clones, and CI checkouts inherit it. Had it been a disk-only artifact of `/setup`, this guard would have broken the fallback on every worktree of a v2 fork. It isn't, so it doesn't.

**3. Discriminating claim — reproduced exactly.** Reverted only the eight libs to `dev`, kept the test: **9 FAIL / 2 PASS**, with all nine failures reporting `IMPOSTOR`. At HEAD: 11/11.

The `_lib-premium-hook.sh` probe, which you asked me to scrutinise, is sound. It compares `_premium_hook_dir`'s return against `git rev-parse --show-toplevel` on *both* sides, so the macOS `/var` vs `/private/var` mismatch that produced documented defect (2) cannot recur — and it genuinely discriminates, since that case was among the nine RED-phase failures rather than a lone pass.

### Issues Found

**[Medium — non-blocking] The test has a remaining hole of the same class as the four the PR body documents: no-regression coverage exists for only 2 of the 9 sites, and both are on the same lib.**

The nine negative cases assert "no impostor tripwire was set". The test's own comment concedes that *refusing to load at all is also a passing outcome* for those cases. That is correct as written, but it means an over-strict guard at any of the seven uncovered sites would ship green — which is precisely the failure mode #1028 was filed for, arriving through a different door.

Proved by mutation rather than asserted: I replaced `_lib-project-board.sh`'s anchor filenames with ones that can never exist, making that site permanently unable to resolve. The suite still reported **11/11 green**. Same structural signature as documented defect (3) — a site that cannot resolve is indistinguishable from a site that correctly refuses.

The shipped code is *not* affected; I confirmed all nine sites resolve on both fork layouts by direct construction. This is a durability gap in the test, not a defect in the fix, which is why it is a suggestion rather than a blocker.

Suggested fix: extend `assert_loads_in_real_fork` over the remaining seven sites. The existing `make_real_fork` fixture already copies the three sibling libs those sites need, so this is close to a loop over the same probe table the negative cases already use — and it would also give the v1 layout the coverage it currently lacks entirely (both positive cases use `.apexyard-fork` only).

Two small notes on the fixture while you are in there: `make_real_fork` runs `git add README.md .claude` without ever creating `README.md`, so the `add` and the follow-on `commit` both fail silently. Harmless — `rev-parse --show-toplevel` works in a repo with no commits, which is why the cases pass — but it means the fixture is not doing what it reads as doing, and that is the same category of latent-fixture-gap as defect (3).

### Suggestions

**nit: inlining is the right call, and the reasoning holds.** I tried to find the safe extraction and there isn't one. A shared `_lib-fork-anchor.sh` would itself have to be located from these same libs, by the same `BASH_SOURCE`-then-`rev-parse` dance — so you would need an anchor check to locate the anchor checker. The only escapes are duplicating the check inside the locator (the duplication you were avoiding) or locating it unguarded (reintroducing #1033). The usual out — define the function in a lib guaranteed already loaded — has no candidate here, because these libs *are* the bottom of the load order. Nine copies is the honest shape.

Given that, the residual risk is drift rather than duplication: nine copies that are identical today can diverge one at a time. Cheap mitigation, since they are already byte-identical modulo the root variable: a test that normalises the variable name across the nine guards and asserts a single distinct anchor clause. That converts silent divergence into a failing test, and it is a few lines.

**nit: line length.** The guard lines run to roughly 190 characters where these libs otherwise wrap. `shellcheck --severity=warning` is clean on all nine files so nothing is broken; a line continuation before the final `&& [ -f ... ]` would keep the diff readable in side-by-side review. Entirely cosmetic — do not churn the SHA for it alone.

**Out of scope, worth a follow-up ticket:** several of these same libs resolve *state* paths with an unanchored `git rev-parse` — `depth_mode_marker_path` and `glossary_seen_marker_path` (marker locations), `fresh_fork_state`, and `_lib-audit-history.sh`'s `projects_dir` fallback. Those predate #1028 and are data-path resolution rather than code loading, so they are correctly outside this PR's scope. They are a weaker version of the same "wrong repo wins" hazard and would be a reasonable next ticket.

### Completeness of the nine

Confirmed independently, since you flagged the count as unverified. `git show 4c24c2e` — the #1028 commit — introduced exactly nine self-location fallback assignments across these eight files; all nine now carry the guard (seven files with one, `_lib-tracker.sh` with two). Every other `rev-parse --show-toplevel` occurrence in `.claude/hooks/_lib-*.sh` is one of: a usage comment in a header block; a state/data path rather than a lib load (the follow-up above); or a resolver that already walks for the anchor first (`_lib-read-config.sh`, `_lib-portfolio-paths.sh`, `_lib-migration-chain.sh`). **Nine is the complete set.**

### On the AgDR question

You asked to have this challenged, so — directly: **I do not think this needs one, and I think the rail-1 reading is doing less work than it appears to.**

Rail 1 says *"a **decision** touching `.claude/hooks/**` ... is material regardless of how small the diff is."* The subject is a decision. Rail 1 answers *"is this decision material?"* — it is a size-override that stops you arguing "it is only a one-line hook change, so it is routine." It does not answer *"was a decision made?"*, and it does not convert every trust-chain edit into an AgDR trigger. If it did, every bug fix under `.claude/hooks/**` would need a record, which is neither this repo's practice nor a bar anyone could hold.

So: was a new decision made? The security control applied here — `.apexyard-fork`, **or** the `onboarding.yaml` + `apexyard.projects.yaml` pair — is not a new choice. It is **AgDR-0021 § A and § E**, which chose the marker file over three alternatives and set v2-marker-precedence-with-legacy-pair-fallback. `_lib-ops-root.sh` implements it verbatim; `_lib-read-config.sh`, `_lib-portfolio-paths.sh`, and `_lib-migration-chain.sh` all already use it. This PR restores nine sites to an existing recorded decision. Conformance, not choice.

Two corroborating signals. `require-agdr-for-arch-pr.sh` does not fire — `.claude/hooks/**` is deliberately absent from `agdr_trigger_paths`, which is calibration rather than oversight. And #1028 itself, which *introduced* a new resolution primitive into nine trust-chain sites — a materially larger departure than fixing it — shipped without an AgDR. Requiring one for the repair but not the introduction would be incoherent.

Where rail 2 (ambiguity rounds up) does have purchase is the **inline-over-shared-helper** call, which is a real engineering decision that future bootstrap-layer sites will copy. But its rationale is already written down at each of the nine sites and in the commit body, it is confined to one layer, and it is reversible by mechanical refactor with this test in place to catch a regression. The practical test in `agdr-decisions.md` asks whether a teammate would be confused or repeat expensive work — they would not, because the reasoning is in front of them.

**My call: no AgDR required; not blocking.** The cheap thing that resolves the remaining ambiguity at near-zero cost is a citation rather than a new record — add "anchor pair per AgDR-0021 § A/§ E" to the PR body and to one of the nine site comments, so the next maintainer knows the two conditions are a recorded decision to be cited, not a local invention to be re-derived. I would take that over a new AgDR. If you disagree and want one written, the row it would cross is *"touches a security-relevant control — auth, crypto, secrets, the trust chain"*, and it should be scoped to the inline-vs-helper trade-off, not to the anchor definition, which is already covered.

### Standard checks

`Refs #1033` present. Body carries Summary, Scope, Testing, Review, and Glossary sections; summary bullets are narrative with rationale, not label-only. Commit subject is conventional and the body explains why rather than restating what. `shellcheck --severity=warning` clean across all nine changed files. New test carries its exec bit, consistent with the sibling tests. No handbook findings — no layering violations (shell libs, no domain/infrastructure boundary), no migration files, commit-message handbook satisfied.

### Verdict
**APPROVED**

The security fix is correct at all nine sites, complete, and — the part that mattered most — does not undo #1028 on either fork layout, which I confirmed by constructing the v1 case the local fork cannot exercise. The test-coverage gap is real and worth a follow-up, but it constrains future confidence rather than this change's correctness, and blocking on it would hold a correct trust-chain fix behind a test improvement.

---
Reviewed commit: `e2283d49e526655117e343ccf770db74d750d756`

atlas-apex (commented):
## Security Review: PR #1061

**Commit**: `28e62ab864b4114d621fcfdcb0d4ab2ffb542dda`
**Tier**: Heavy (trust-chain, `right-size-ceremony.md` rail 1)
**Score**: 85 / 100 — 0 CRITICAL · 0 HIGH · 1 MEDIUM · 4 LOW

### Summary

Nine self-location sites in eight trust-chain libraries now require a git-derived root to carry `.apexyard-fork`, or the `onboarding.yaml` + `apexyard.projects.yaml` pair, before sourcing a sibling out of it. The anchor conditions are byte-identical to `_lib-ops-root.sh`'s (`resolve_ops_root_walk` / `_ops_root_anchor_valid`), so this is a consistency restoration, not a new trust primitive.

The change is a **strict improvement over the base** and is correctly *fail-safe*: I could not find a path where refusing to resolve flips a gate open. I verified the ticket's reachability claim independently rather than taking it, and it holds. Findings below are residual-scope and accuracy items, not blockers.

### Checklist Results

| Dimension | Result |
|---|---|
| Secrets & Credentials | Pass — no credentials, no logging changes |
| Injection Prevention | Pass — added code is `[ -f "$var/..." ]` tests only; no `eval`, no unquoted expansion, no attacker-controlled text reaching a shell word. Quoting correct at all 9 sites; `{ A \|\| { B && C; }; }` grouping is precedence-correct |
| XSS Prevention | N/A (shell library diff) |
| Auth & Authorisation | Pass — see reachability + failure-mode trace below |
| Data Protection | N/A |
| API Security | N/A |
| Gate & Trust-Chain Integrity | **Pass with one residual** — no gate is weakened, loosened, or made to fail open. See MEDIUM-1 for a branch the guard does not cover |

---

## Q1 — Is the anchor a security boundary, or a speed bump?

**A speed bump — but the right one. It is not an access-control boundary and should not be described as one.**

The honest mechanics: the anchors are unauthenticated, presence-only, zero-content marker files. `-f` follows symlinks, so the check is satisfiable without even controlling file *content*. Demonstrated on this fork:

```
$ ln -s /etc/hosts .apexyard-fork          # in a non-fork git repo
$ zsh -c '. <real-fork>/.claude/hooks/_lib-extract-pr.sh'
  symlinked .apexyard-fork -> IMPOSTOR_TRACKER=1     # impostor lib loaded
```

Against an adversary who can write into the cwd's repo, this raises the bar by exactly one `touch`. It authenticates nothing about the code being loaded.

What it genuinely does — and why it is still worth shipping — is convert a **broad, silent, accidental** misresolution into a **narrow, deliberate** one. The realistic threat was never a targeted attacker; it was the operator's own cwd sitting in a `workspace/<project>` clone or a scratch checkout while an agent sourced a trust-chain lib, getting the wrong code with no signal at all. That accident surface was *every git repo on the machine*; it is now *every git repo deliberately marked as an apexyard fork*. That is a large, real reduction in a correctness surface that has security consequences.

**Rate it as defense-in-depth / accident-surface reduction, not as preventing a hostile library load.** I'd ask that the changelog and any AgDR use that wording — the PR body's "now refuse a git-derived root that isn't an apexyard fork" is defensible, but "refuse" invites a stronger reading than the primitive supports.

### Should `resolve_ops_root` have been used instead?

**No — the PR's choice is correct, but its stated reason ("a shared helper would itself need locating") undersells the real one.** Three independent reasons:

1. **Circularity, concretely.** Three of the nine sites (`_lib-project-board.sh`, `_lib-onboarding-depth-mode.sh`, `_lib-onboarding-glossary-seen.sh`) are locating **`_lib-ops-root.sh` itself**. You cannot call the resolver to find the resolver.
2. **The pin is session-scoped and frequently absent.** `resolve_ops_root`'s strength is the `CLAUDE_CODE_SESSION_ID` pin. In exactly the contexts this fallback exists to serve — CI, a bare `bash`/`zsh` invocation, a test runner — the session id is unset and `resolve_ops_root` degrades to `resolve_ops_root_walk`, which is *the same two anchor conditions this PR just inlined*.
3. So the accurate framing is: **the inlined guard is `resolve_ops_root_walk`'s anchor test applied to a single candidate instead of a walk.** That is the correct 90% at the bootstrap layer. The missing 10% — pin-based *identity* — is unavailable here by construction, not by omission.

**One refinement worth a follow-up (non-blocking):** the guard tests only the git toplevel, not a walk upward. The single most common operator layout — cwd inside `<fork>/workspace/<project>` — has a toplevel of the *clone*, which fails the anchor, so the site silently does not resolve even though the real ops fork is a parent directory. Substituting a bounded walk (the `resolve_ops_root_walk` shape, inlined) would turn "silently does not resolve" into "correctly resolves" for that layout at no extra trust cost.

---

## Q2 — Reachability (verified independently)

**The ticket's claim is correct. I did not take it from the ticket; I reproduced both directions.**

`.claude/settings.json` launches every hook as `bash -c '…; exec "$r/.claude/hooks/<hook>.sh"'` with an absolute path, and the scripts are `#!/bin/bash`. `BASH_SOURCE[0]` therefore resolves and the fallback branch is never entered. Probe run from a cwd inside an *unrelated* git repo, invoking a hook-shaped script by absolute path:

```
BASH_SOURCE-resolved dir=/Users/ahmed/apex/apexyard/.claude/hooks
tracker lib LOADED
```

The fallback **does** fire from a non-bash shell — this machine's login shell is zsh, and a skill sourcing a lib by absolute path from a foreign cwd reaches it:

```
zsh dir=<unrelated-repo>      # dirname "" -> "." -> cwd
tracker NOT loaded            # post-fix: anchor refused it
```

**Conclusion: the reachable surface is skill/agent space, not mechanical-gate space.** No merge, ticket, migration, marker, or CI gate can reach these nine fallbacks. MEDIUM on #1033 is the right rating, and I found nothing that would raise it.

**One related observation, out of scope for this diff.** `validate-branch-name.sh:134-156` sources **both** `_lib-read-config.sh` and `_lib-tracker.sh` from `$(git rev-parse --show-toplevel)/.claude/hooks/` with **no anchor at all** — not a `BASH_SOURCE` fallback, the unconditional primary path. A blocking gate therefore executes shell code from the cwd repo. I believe this is *intentional* (a managed project supplying its own branch conventions and `tracker.id_pattern`), and it is pre-existing, so it is **not a blocker here**. But it means the trust model is now inconsistent: nine sites refuse a non-fork git root while a gate hook deliberately loads executable code from one. Worth an explicit decision (a one-line AgDR note, or a follow-up ticket), not a silent divergence.

---

## Findings

### MEDIUM-1 — The anchor guards only the fallback branch; the primary branch is still cwd-derived and unguarded
**OWASP A08:2021 Software and Data Integrity Failures** (A03:2025 supply chain) · *pre-existing, not introduced by this PR* · non-blocking

All nine sites have the shape:

```sh
_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
if [ -z "$_dir" ] || [ ! -f "$_dir/<sibling>" ]; then
  ... git rev-parse + NEW ANCHOR CHECK ...
fi
. "$_dir/<sibling>"
```

Under a non-bash shell, `dirname ""` → `.` → **`_dir` becomes the cwd**. If the cwd *directly contains* the sibling filename, `[ ! -f ]` is false, the `if` body never executes, and the new anchor is never consulted. In practice that means: cwd is some other repo's `.claude/hooks/` directory. Reproduced against this branch's code:

```
=== TEST A: cwd = impostor repo ROOT (the case the PR tests) ===
  IMPOSTOR_TRACKER=unset          # guard works
  IMPOSTOR_OPSROOT=unset          # guard works

=== TEST B: cwd = impostor .claude/hooks/ (primary branch, guard bypassed) ===
  IMPOSTOR_TRACKER=1  tracker_kind=IMPOSTOR
  IMPOSTOR_OPSROOT=1  IMPOSTOR_CONFIG=1
  tracker.sh->config: IMPOSTOR_CONFIG=1
```

This is not a regression — it predates #1028 (`dirname ""` has always been `.`) and the PR does not claim to fix it. But it bounds the security property: **the sites refuse a bad *git-derived* root; they do not refuse a bad *cwd-derived* one**, and the cwd-derived branch runs first. The reachable instance is the #381 shape (an agent `cd`'d into a scratch/second clone's hooks dir).

**Recommendation:** follow-up ticket, not a change to this PR. The clean fix is to validate `_dir` against the same anchor in *both* branches rather than only inside the fallback — which also removes the "is `_dir` still the cwd?" ambiguity the current shape leaves behind.

### LOW-1 — Anchor files are presence-only and symlink-satisfiable
**OWASP A08:2021** · informational, **do not "fix"**

Demonstrated in Q1. `-f` follows symlinks and no content, ownership, or mode check is performed. This is inherited from `_lib-ops-root.sh` / AgDR-0021 and is *consistent by design*, which is the right call — an attacker who can `ln -s` in a directory can equally `touch` the file, so content validation would add zero security while breaking v1 forks. Flagged only because it is the concrete evidence for the Q1 rating, and the property deserves to be written down rather than rediscovered.

### LOW-2 — The anchor is a type check, not an identity check (question 5)
**OWASP A08:2021** · accepted limitation, worth documenting

A second clone of a legitimate fork — a `/tmp` scratch clone, a `git worktree`, a colleague's checkout — satisfies both anchors identically. The anchor answers *"is this an apexyard fork?"*, never *"is this THE ops fork for this session?"*. That distinction is precisely the #381 incident, and it is the one axis on which `resolve_ops_root`'s pin is strictly stronger.

**Does it matter here? Materially less than it does for marker writes.** All nine sites load *code* that is identical across clones of the same fork, so the worst realistic outcome is a **stale** library version, not a hostile one. The dangerous version of this ambiguity is marker *writing* (#381's actual damage), which these nine sites do not do. Accept, and record the reasoning so a future reader doesn't assume the anchor is doing identity work.

### LOW-3 — TOCTOU is present but is not the weak link
**OWASP A01:2021 (adjacent)** · no action recommended

Check-then-source is non-atomic at all nine sites (`[ -f "$root/.claude/hooks/X" ]` … `. "$root/.claude/hooks/X"`), and the anchor test and the source test are separate stats on separate paths. An attacker who can win that race already has write access to the candidate root, at which point they can simply write the library before the check — the race buys nothing. Correctly *not* worth hardening.

One consistency note only: unlike `require-active-ticket.sh` (which resolves symlink-free real paths before judging, per `workflow-gates.md` § out-of-repo exemption), these sites do no realpath normalization, so a symlinked `.claude/hooks` inside a legitimately-anchored fork redirects every sibling load. Same "already game over" reasoning applies.

### LOW-4 — Comment block: broken indentation and a leaked authoring instruction, ×9
*style, but it is in the trust chain*

At every one of the nine sites the second half of the comment starts at column 0 inside indented code:

```sh
      # me2resh/apexyard#1033: only accept a git-derived root that is
      # ... (correctly indented)
# The anchor pair is not a new choice -- it is AgDR-0021 SS A/E
# (marker file over three alternatives; v2 marker with legacy-pair
# fallback), already used by _lib-ops-root.sh. Cite it, do not re-derive.
```

Two issues: the indentation break repeats nine times in the framework's most-audited files, and *"Cite it, do not re-derive"* is an instruction to the author, not documentation for a reader. Trust-chain comments are read by future auditors deciding whether a guard is load-bearing; they should not contain drafting directions. Also `SS` appears to be a mangled `§`. Please reflow and drop the imperative.

### LOW-5 — PR body understates the test suite it ships
*accuracy*

The body states "11 cases" and "After the change: **11/11**". The file actually runs **27** (9 negative × 1 + 9 positive × 2 fork layouts), which I ran and confirmed at **27/27**, `shellcheck --severity=warning` clean. The body's later paragraph describing the added positive coverage contradicts its own earlier count. For a security test, the claimed coverage number is load-bearing — please correct it before merge or in the squash message.

---

## Q3 — Does the fix introduce a new failure mode? (traced)

**Silent non-resolution is safe here. This is the direction I checked hardest, and it comes out clean.**

`_lib-extract-pr.sh` (feeds all four merge gates) — if `_lib-tracker.sh` does not load:
- `tracker_kind` is undefined, but `_forge_kind_for` guards with `command -v tracker_kind` and returns `gh`. On GitHub (the default) behaviour is byte-identical.
- On a GitLab project it would query `gh` instead of `glab`; `resolve_pr_head` returns empty; `block-unreviewed-merge.sh:232-235` warns and falls back to local HEAD, whose SHA then almost certainly mismatches the marker → **BLOCK**. Fail-closed.
- Decisively: `is_merge_command`, `extract_pr_number`, `extract_repo_from_command`, and `merge_command_uses_variable` have **zero** tracker dependency. The gate still *fires* and still *extracts its target*; it cannot be silently skipped by a failed lib load.

`_lib-project-board.sh` — if `_lib-ops-root.sh` / `_lib-read-config.sh` do not load:
- `config_get` is undefined → `enabled=$(config_get '.github_projects.enable_auto_moves' 2>/dev/null)` is empty → `[ "$enabled" != "true" ]` → `return 0`.
- That is precisely the lib's documented contract (header: *"Any failure … warns to stderr and returns 0. This lib NEVER exits non-zero — it must never block the lifecycle action that called it."*). The observable outcome is **a board card that does not move**, not a gate that opens. Its only blocking-hook consumer, `auto-code-review.sh`, sources it behind an `[ -f ]` guard and ignores the result.

**The general principle, which is the real answer:** none of the nine siblings supplies a *permit* decision. They supply configuration and defaults. Unavailability therefore degrades to the framework default (`gh` / disabled / advisory) — never to "approved". No caller anywhere in the nine treats "lib not loaded" as "check passed"; every one of them treats it as "use the conservative default".

**On the anchorless-fork worry specifically:** a fork whose anchors are absent or renamed does not merely lose these fallbacks — it also loses `_lib-ops-root.sh`'s `resolve_ops_root` (identical anchors), so it was already broken framework-wide before this PR. This change creates no new class of broken fork. Verified on this fork, which is split-portfolio v2 (`.apexyard-fork` present, v1 pair absent): all nine sites resolve, and the suite's `v1` layout cases prove the legacy pair still works.

---

## Q4 — TOCTOU / symlink surface

Covered in LOW-1 and LOW-3. Short version: both are real, both require filesystem write access in or adjacent to the target directory, and any attacker holding that can bypass the guard far more directly by writing the library itself. Neither warrants hardening; both warrant being written down so the guard is not mistaken for stronger than it is.

---

## Recommendations (all non-blocking)

1. **Describe the change accurately in the changelog/AgDR** — "reduces accidental misresolution surface", not "prevents loading a hostile library". This is the one I care most about; an overstated security claim in the trust chain costs more later than the guard saves.
2. **File a follow-up for MEDIUM-1** — anchor-validate the resolved dir in *both* branches, not only inside the fallback.
3. **Consider a bounded walk-up** instead of a single-candidate test, so `<fork>/workspace/<project>` resolves correctly rather than silently not resolving (Q1, refinement).
4. **Fix the comment indentation and drop "Cite it, do not re-derive"** at all nine sites; `SS` → `§`.
5. **Correct the "11 cases / 11-11" claim** — the suite is 27 and passes 27/27.
6. **Open a consistency ticket for `validate-branch-name.sh`'s unanchored `$REPO_ROOT` lib loads** — decide and record whether executing the cwd repo's `_lib-tracker.sh` / `_lib-read-config.sh` from a blocking gate is intended.

## Verdict

**APPROVED**

No CRITICAL or HIGH findings. The change is a strict improvement over its base, correctly fail-safe in every path I traced, ships a genuinely discriminating test (proven 9-FAIL against the unfixed code) with positive coverage on both fork layouts, and reuses the established AgDR-0021 anchor rather than inventing a new trust primitive. The single MEDIUM is a pre-existing residual the PR neither introduces nor claims to fix; it belongs in a follow-up ticket. The remaining items are accuracy and hygiene.

The one thing I'd genuinely push back on is framing, not code: ship this described as accident-surface reduction. It is not an integrity boundary, and the codebase will be better off if that is on the record now.

---
📌 Reviewed commit: `28e62ab864b4114d621fcfdcb0d4ab2ffb542dda`

atlas-apex (commented):
## Security Review: PR #1061 (re-review)

**Commit**: `ab99c222885224f3bd83b1db1c8d87be1450ba62`
**Previous review**: `28e62ab` — APPROVED, 85/100
**Scope of this pass**: `ab99c22` (comments only, 8 libs / 9 sites) + re-confirmation of the whole PR at this SHA

### Summary

`ab99c22` acts on the central finding from my `28e62ab` review: the comment blocks described the anchor as though it were a trust boundary. They no longer do. I re-verified every factual claim in the new text against the code rather than accepting the paraphrase, and confirmed the rewrite is genuinely comments-only.

**Verdict on the correction: it lands correctly, and it does not under-sell the guard.** One claim over-corrects in the *other* direction — see LOW-1.

### Logic drift check — clean

Stripping comment lines from `28e62ab..ab99c22` across `.claude/hooks/` yields an **empty** diff. Every `if [ -n "$root" ] && { ... anchor ... }` line appears as unchanged context in all nine hunks. `test_lib_self_location_anchor.sh`: **27 passed, 0 failed** at this SHA.

### Claim-by-claim verification

| Claim in the new text | Verdict | Evidence |
|---|---|---|
| Anchors are unauthenticated, presence-only | **TRUE** | `_lib-ops-root.sh:92` — "Presence-only — content is ignored." No signature/checksum on either anchor; legacy pair is `-f` only. |
| `-f` follows symlinks | **TRUE** | Reproduced independently: `ln -s /etc/hosts .apexyard-fork` passes `[ -f ]` under bash, zsh **and** sh. (Dangling symlinks and dir-symlinks correctly fail.) |
| Same test as `resolve_ops_root_walk`, one candidate vs a walk | **TRUE** | Byte-identical predicate: `_lib-ops-root.sh:93` + `:99` vs the inline condition at all nine sites. |
| Three of the nine sites locate `_lib-ops-root.sh` | **TRUE** | Exactly three: `_lib-onboarding-depth-mode.sh`, `_lib-onboarding-glossary-seen.sh`, `_lib-project-board.sh`. |
| `resolve_ops_root`'s pin is session-scoped | **TRUE as fact** | `_lib-ops-root.sh:131-134`, keyed on `CLAUDE_CODE_SESSION_ID`. But see LOW-2 — true fact, under-powered as an argument. |
| Anchor pair per AgDR-0021 §A/§E | **TRUE** | §A = "Public-fork anchor: marker file vs sentinel config key vs `.git`-walk + heuristic", marker file chosen over three alternatives. §E = "v2-marker vs legacy-anchor precedence", v2 marker with legacy-pair fallback. Both citations support the text. |
| Surface was "ANY repo the cwd happens to be inside" | **OVERSTATED** | See LOW-1. |

Comment-block defects from the prior pass are all fixed: indentation now matches enclosing scope at all nine sites (zero added col-0 comment lines), the leaked authoring instruction "Cite it, do not re-derive" is gone (0 occurrences), and `§` renders correctly with no mojibake.

### Findings

#### [LOW-1] The "ANY repo" baseline overstates what the guard removes

`.claude/hooks/*.sh` (all nine sites), and PR body line 50.

> "Without this the fallback sources a trust-chain library out of ANY repo the cwd happens to be inside"

and, in the PR body:

> "That surface was *every git repo on the machine*; it is now *every repo deliberately marked as a fork*."

The pre-PR condition was never "any repo". It was:

```sh
if [ -n "$root" ] && [ -f "$root/.claude/hooks/_lib-read-config.sh" ]; then
```

That co-condition was always present and is **still** present post-PR (it is the final conjunct on the same line the comment sits above). So the baseline was "any git repo that *also ships a same-named trust-chain library at `.claude/hooks/`*" — a substantially smaller set. #1033 itself gets this right, listing it as one of three required conditions: *"and that repo shipping a same-named library."* The comment drops the qualifier that its own ticket kept.

Concretely: on this fork, neither registered `workspace/<project>` clone carries a `.claude/hooks/` directory at all, so the comment's own worked example is not reachable here without the clone shipping the lib.

This is the mirror image of the defect the commit fixes. The old text overstated the *nature* of the protection ("boundary"); the new text overstates the *size* of the surface removed. The commit correctly stopped claiming a boundary — it should not simultaneously inflate the baseline.

Suggested: `...out of any repo the cwd happens to be inside that also ships a same-named library at that path — a workspace/<project> clone, or an unrelated checkout.` One clause, and the sentence becomes exactly true.

#### [LOW-2] The `resolve_ops_root` parenthetical records a weaker argument than the real one

> "(resolve_ops_root itself is unusable here: three of these sites are locating `_lib-ops-root.sh`, and its pin is session-scoped.)"

This is a faithful record of what I said, and both facts check out. On re-examination the argument is under-powered:

- Reason (a) covers **three** sites. A reader reaching the other six is left with only reason (b).
- Reason (b) does not establish unusability. `resolve_ops_root` degrades gracefully — with no pin, no session id, or `APEXYARD_OPS_DISABLE_PIN=1` set, it falls through to `resolve_ops_root_walk` (`_lib-ops-root.sh:126-151`). Session-scoping makes the pin *unhelpful* here, not the function *unusable*.

The stronger reason, which covers all nine sites uniformly: this branch executes **precisely when self-location has already failed**, so the code cannot locate `_lib-ops-root.sh` to source it either. It is circular at every site, not just the three. Nothing in the current text is false — it just leaves a future maintainer a weaker case than the one that actually holds.

#### [MEDIUM] Carried forward — filed as #1062, correctly scoped out-of-PR

Re-verified at this SHA rather than taken on trust. The `BASH_SOURCE`-derived branch runs first; under zsh `BASH_SOURCE[0]` is **unset** (confirmed, zsh 5.9.2), `dirname ""` → `.`, and the derived dir becomes the caller's cwd. With cwd inside a non-fork repo's `.claude/hooks/` containing a same-named sibling, the existence check succeeds, the `if` body never executes, and the anchor is never consulted:

```
cwd = impostor repo root         → impostor lib NOT loaded  (the #1061 anchor works)
cwd = impostor/.claude/hooks/    → impostor lib IS loaded    (anchor bypassed)
```

**#1062's framing is accurate**, including the distinction it preserves (a genuine `BASH_SOURCE` path is trustworthy by construction and must not be gated — only the `dirname ""` degradation needs the check) and the sequencing note.

**#1061 does not make this worse**, and this is structural rather than a judgement call: the new anchor is a conjunct (`&&`) added inside a branch that only runs when the cwd branch already failed. The post-PR accepted set is a strict subset of the pre-PR accepted set. There is no input for which this PR causes a load that previously would not have happened.

Fail-closed behaviour confirmed: a repo satisfying neither anchor loads nothing and `config_get_or` stays undefined, which is the loud-failure contract #1028 established.

### Checklist Results

- Secrets & Credentials: **Pass** — no secrets; anchors carry no content
- Injection Prevention: **Pass** — no new command construction; all paths quoted
- XSS Prevention: **N/A**
- Auth & Authorisation: **Pass** — correctly no longer claims to be an authz control
- Data Protection: **Pass**
- API Security: **N/A**
- Gate & Trust-Chain Integrity: **Pass** — monotonically stricter; no gate weakened, no fail-open introduced; 27/27 anchor tests green

### Note on the #1063 severity framing (asked, out-of-scope for this PR)

P2 is right, and the "beyond tidiness" argument holds: `audit_run_persist` truncates at redirect time, so a `jq` failure leaves a zero-byte file that `audit_run_list` treats as a valid run, and the operation reports success while leaving unusable state — the wrong failure direction for an audit artefact. Two refinements:

1. Temper "silently contributes a bogus point to a **security** trend line" slightly — the lib is advisory and never gates a merge, which the ticket's own Risks section says. Keep both statements, but the Risks framing is the accurate ceiling.
2. Consider promoting AC-3 (`audit_run_list` skips zero-byte / unparseable runs) ahead of the atomic-write AC. Existing forks may already carry such files, so the defensive read is what actually restores a correct trend today; the atomic write prevents new ones.

I hit this again during this run — noted below.

### Verdict

**APPROVED** — 91/100 (was 85/100 at `28e62ab`)

0 critical, 0 high, 1 medium (pre-existing, filed as #1062, does not block), 2 low (both documentation-precision, non-blocking).

The central finding from the prior pass is properly addressed. The two LOW items are wording precision in comments, not defects in the guard — fix in this PR if convenient, or fold into #1062's follow-up since it touches the same nine blocks.

**Direct answer to the question asked:** the new wording does *not* under-sell the guard. The block is ordered value → limit → value ("only accept a git-derived root that is actually an apexyard fork" … "It is NOT an access-control boundary" … "What it prevents is a cwd-driven misresolution"), so a maintainer meets the reason to keep it before the caveat, and the 27-test suite makes silent removal fail loudly. The over-correction is elsewhere: LOW-1, where the baseline is inflated.

---
📌 Reviewed commit: `ab99c222885224f3bd83b1db1c8d87be1450ba62`

atlas-apex (commented):
## Code Review: PR #1061 (re-review)

**Commit**: `ab99c222885224f3bd83b1db1c8d87be1450ba62`
**Previously reviewed**: `e2283d4` (APPROVED, one MEDIUM raised)

### Summary

Two commits landed since my last pass. `28e62ab` acts on my MEDIUM (positive coverage existed for 2 of 9 sites); `ab99c22` acts on Hakim's push-back on the security framing and fixes three defects in the comment block. I re-ran my own mutation rather than take the commit message's word for it.

**My MEDIUM is closed.** Confirmed by independent mutation, in both directions, across all nine sites.

---

### 1. Mutation re-run — the MEDIUM is genuinely closed

I rebuilt the mutation harness from scratch against a scratch copy of `.claude/` and made each site's anchor unsatisfiable one at a time (`.apexyard-fork` → `.apexyard-fork-NEVER`, `/onboarding.yaml` → `/onboarding.yaml-NEVER`), asserting the edit actually landed before each run.

| Mutated site | Result |
|---|---|
| `_lib-audit-history.sh:72` | pass=25 fail=2 |
| `_lib-extract-pr.sh:158` | pass=25 fail=2 |
| `_lib-fresh-fork.sh:69` | pass=25 fail=2 |
| `_lib-onboarding-depth-mode.sh:75` | pass=25 fail=2 |
| `_lib-onboarding-glossary-seen.sh:95` | pass=25 fail=2 |
| `_lib-premium-hook.sh:173` | pass=25 fail=2 |
| `_lib-project-board.sh:81` | pass=25 fail=2 |
| `_lib-tracker.sh:106` | pass=25 fail=2 |
| `_lib-tracker.sh:152` | pass=25 fail=2 |

Every site kills exactly its own two cases (v2 + v1) and no others — no cross-talk, no site silently uncovered. `_lib-project-board.sh`, the site I broke last time to demonstrate the gap, now fails loudly where it previously left the suite 11/11 green.

I also ran the inverse. Stripping the anchor from all nine sites (reverting to pre-fix code) gives **9 FAIL / 18 PASS** — all nine negative cases correctly detect the impostor load. So the suite now kills mutants in both directions: too-loose fails the negatives, too-strict fails the positives.

A note on method, since it is the same failure class this PR is about: my **first** mutation harness silently no-op'd. I ran it under zsh, which does not word-split unquoted `$spec`, so `sed` was handed a filename that did not exist and every "mutation" edited nothing — reporting a clean 27/27 nine times over. I only caught it from stderr. The numbers above are from the re-run under `bash`, with a `diff` assertion that each edit landed before the suite runs. Worth recording because a mutation harness that quietly does nothing produces exactly the reassuring green this whole test exists to distrust.

### 2. Positive coverage — complete and correct

9 sites × 2 layouts = 18 positive cases, plus 9 negative = 27. Matches the claim.

**Probes assert real resolution, not something weaker.** `tracker_view` is defined only at `_lib-tracker.sh:534` and nowhere in `_lib-extract-pr.sh`, so `command -v tracker_view` genuinely proves the sibling loaded across the guarded fallback — and the mutation above confirms it fails when the anchor breaks. Same for `config_get_or`, `portfolio_registry`, and `resolve_ops_root`.

**The v1 fixture is a genuine v1 fork.** I did not rely on reading the fixture. I mutated the guard itself:

- **v2-only guard** (drop the `onboarding.yaml`+`projects.yaml` arm) → exactly the **9 v1** cases fail
- **v1-only guard** (drop the `.apexyard-fork` arm) → exactly the **9 v2** cases fail

That is only possible if the v1 fixture carries no `.apexyard-fork` and the v2 fixture carries no v1 pair. Both layouts are independently exercised at all nine sites. Given this fork is v2, the v1 half is the half that could never have been caught locally — good that it is pinned.

### 3. New findings

Two, both minor. Neither blocks.

**LOW — the `_lib-premium-hook.sh` positive probe is missing the emptiness guard its negative twin has.**

Line 284 (positive):
```sh
[ -n "$_d" ] && case "$_d" in "$_top"/*) echo OK ;; esac
```
Line 171 (negative) correctly guards `[ -n "$_top" ] && case ...`. The positive one does not. If `$_top` is ever empty the pattern degenerates to `/*`, which matches any absolute path — including the unresolved `/dev`. Verified:

```
_top=""; _d="/dev"  →  prints OK
```

So on a runner where `git rev-parse` fails inside the fixture (no git on PATH, `git init` failing), this case reports PASS against a site that resolved nothing. It is not reachable on any runner that has git, and in that scenario the other eight positives would fail and redden the suite anyway — so this is probe hygiene, not a live coverage gap. But it is the one probe that is asymmetric with its own negative twin, and it is precisely the shape of the four defects already recorded. Suggest mirroring line 171: `[ -n "$_top" ] && [ -n "$_d" ] && case ...`.

**nit — dead helper.** `assert_loads_in_real_fork` (lines 235–244) is defined and never called; `assert_resolves_in_fork` superseded it. Leftover from the 11-case version.

I specifically probed for the class you flagged and found nothing else: the impostor fixture now creates all five same-named libs (so every existence check fires), the premium probe compares against `git rev-parse` rather than `$PWD` (the macOS `/var` vs `/private/var` trap), and both premium probes use `case` against the git-resolved root consistently.

### 4. Comment rewrite — comments only, confirmed mechanically

Not by reading. Stripping all comment lines from both revisions and diffing:

```
SAME(code) for all 8 libs, e2283d4 → ab99c22
```

Against the base `cd7f6e6`, the code delta is exactly nine guard insertions and nothing else — one per site, two in `_lib-tracker.sh`. No logic drifted.

The three comment defects are fixed:

- **Indentation** — comment-block indent now matches its guard's indent at all nine sites (2/2, 4/4, 2/2, 2/2, 2/2, 6/6, 2/2, 4/4, 4/4).
- **Leaked authoring instruction** — `grep -rn "Cite it"` returns nothing.
- **`§`** — all eight files are `utf-8`, section signs render correctly, no `SS` artifact remains in the source. (It survives in the `28e62ab` *commit message* as "SS A/E", which is immutable now and harmless.)

`shellcheck --severity=warning` is clean across all eight libs and the test file.

**Citation accuracy** — I recommended the AgDR citation, so I checked it rather than just its presence. AgDR-0021 §A is "Public-fork anchor: marker file vs sentinel config key" and §E is "v2-marker vs legacy-anchor precedence in `_lib-ops-root.sh`" — both on point. The substantive claim also holds: `resolve_ops_root_walk` (`_lib-ops-root.sh:84–99`) applies the identical anchor pair with the same v2-preferred ordering. The added note on why `resolve_ops_root` itself is unusable here (three sites are locating `_lib-ops-root.sh`, and its pin is session-scoped) is correct and was the right thing to add.

### 5. PR body vs reality

The two substantive corrections landed. Case count now reads **27 cases (9 negative + 9 positive × 2 fork layouts)**, and the "What this is, stated accurately" section states the accident-narrowing property plainly, including the symlink reproduction. That matches the code comments and matches what I verified.

Two stale numbers survive from the 11-case version, both in directions that *understate* the work:

- Summary bullet 4: *"Two guard cases pin that a real fork still resolves."* — it is now 18.
- Testing: *"Against the unmodified libraries: 9 FAIL / 2 PASS."* — with 27 cases it is **9 FAIL / 18 PASS**, which is what I measured on the over-loose mutation.

Worth a quick edit for internal consistency, since the same body says 27 a few lines later. Not blocking.

### Checklist Results

- Architecture & Design: **Pass** — inlined guard at each site is still the right call for a bootstrap layer; a shared helper would need locating.
- Code Quality: **Pass** — shellcheck clean, indentation correct, comment-only delta verified mechanically.
- Testing: **Pass** — 27 cases, bidirectional mutation-killing, verified independently.
- Security: **Pass** — framing corrected per Hakim; claims now match the actual property.
- Performance: **Pass** — two `-f` tests on a cold path.
- PR Description & Glossary: **Pass** — glossary present; two stale counts noted above.
- Summary Bullet Narrative: **Pass** — all bullets carry what-and-why.
- Technical Decisions (AgDR): **Pass** — AgDR-0021 §A/§E cited and verified accurate.
- Adopter Handbooks: **N/A** — `migration-safety.md` (blocking) does not apply; no migration files touched.
- Fallow Static Analysis: **N/A** — no JS/TS.

### Verdict

**APPROVED**

The MEDIUM I raised at `e2283d4` is closed, verified by my own mutation in both directions across all nine sites rather than on assertion. The two findings above are LOW and nit; neither affects shipped behaviour. Optional follow-ups: mirror the `[ -n "$_top" ]` guard into the premium positive probe, drop the dead helper, and correct the two stale counts in the body.

---
📌 Reviewed commit: `ab99c222885224f3bd83b1db1c8d87be1450ba62`

me2resh (approved):

Issue #1031, opened 2026-07-27
[Bug] .claude/project-config.json is gitignored but tracked — the ignore rule is inert for every adopter fork


atlas-apex:
## Affected
`.gitignore` (line 20) + the tracked `.claude/project-config.json`

## Given / When / Then
**Given** `.gitignore:20` lists `.claude/project-config.json`, and `.claude/project-config.defaults.json` is the tracked defaults template
**When** an adopter forks and customises `.claude/project-config.json` (e.g. adding the `portfolio` block that `/split-portfolio` writes for split-portfolio mode)
**Then** the file shows as permanently modified in `git status`, because git ignores `.gitignore` for already-tracked files — expected that it behave like `onboarding.yaml`, which is both gitignored *and* untracked upstream, leaving adopter config purely local

## Repro
1. Fork `me2resh/apexyard`.
2. `git check-ignore -v .claude/project-config.json` → no output (not ignored, despite the entry).
3. `git ls-tree dev .claude/project-config.json` → `100644 blob 41d02ef … .claude/project-config.json` (tracked).
4. Edit the file the way `/split-portfolio` does (add a `portfolio` block).
5. `git status` → permanently modified, forever. The adopter must either commit fork-local config (inviting conflicts on every sync that touches the file) or live with permanent drift.

## Framework version
v5.2.0

## Severity
minor — cosmetic/workflow friction, clear workaround, but it affects every adopter fork

## Notes
The sibling case is already correct and shows the intended pattern:

| Tracked template | Real file | Upstream state |
|---|---|---|
| `onboarding.example.yaml` | `onboarding.yaml` | gitignored **and untracked** ✅ |
| `.claude/project-config.defaults.json` | `.claude/project-config.json` | gitignored but **tracked** ❌ |

Suggested fix: `git rm --cached .claude/project-config.json` (leaves the file on disk, lets the existing gitignore entry take effect). This is exactly what `/update`'s split-portfolio v1→v2 migration already does for `onboarding.yaml`.

Note the framework repo keeps its own copy to dog-food the pre-push gate (`pre_push.commands`), which is presumably why it was tracked. If that content still needs to ship, it could move into `project-config.defaults.json` or a `project-config.example.json`, mirroring `onboarding.example.yaml`.

Low churn, so untracking should be cheap: only 3 commits touched this file in the last 12 months.

## Glossary
| Term | Definition |
|------|------------|
| tracked file | A file git has in its index; `.gitignore` has no effect on it |
[...]

PR #1065, opened 2026-07-30
fix(#1031): untrack project-config.json and ship a tracked example template

atlas-apex:
## Summary

- **Untracks `.claude/project-config.json` so its `.gitignore` entry finally applies.** The file was listed in `.gitignore` *and* held in the index; git ignores `.gitignore` for tracked files, so the rule has been inert for every adopter fork.
- **Ships a tracked `.claude/project-config.example.json` in its place**, mirroring the `onboarding.example.yaml` / `onboarding.yaml` pair — the example is tracked so upstream can improve it, the real file stays purely local.
- **Fixes five call sites that staged the file**, which would otherwise exit 1 and break the step behind them — including `/split-portfolio`, the framework's destructive migration skill.
- **Adds a 5-case test** whose two discriminating cases were proven to fail against the unfixed tree before the fix was written.

## Why this is data loss, not cosmetic drift

Filed as "permanent drift in `git status`", severity minor. It is worse than that.

Because the file is **tracked**, a plain `git checkout <branch>` writes the indexed version over the working copy, with no warning. For a split-portfolio adopter that destroys the private `portfolio` block, which is load-bearing: path resolution is config-only, there is no convention-based sibling discovery, and `.apexyard-fork` is presence-only. With it gone, `portfolio_registry` silently falls back to the fork itself.

It is **unrecoverable** — the content was never in git, correctly, because it is private. This happened on a real fork during this ticket: no other worktree copy had it either, and recovery meant rebuilding the block by hand from the sibling repo's layout.

## Design: the example file, not the defaults

The issue offered both homes for the framework's own `pre_push` block. `defaults.json` would regress every adopter — `_lib-read-config.sh` merges with `jq -s '.[0] * .[1]'`, so an adopter defining no `pre_push` inherits apexyard's repo-specific commands, including its own subpack test. Verified empirically. Full reasoning lives once in `docs/project-config.md`; `defaults.pre_push.commands` stays `[]` and the test pins it there.

## Known cost: both pre-push paths go quiet on a fresh clone

An earlier revision of this PR claimed terminal `git push` still covers the gap. **That was false** — `.githooks/pre-push` requires `core.hooksPath`, which is per-clone local config and documented as optional. On a fresh clone *both* paths are inactive.

The trade still holds, for a different reason: pre-push is a latency optimisation, and the real backstop is **CI** — `markdown-lint.yml`, `shellcheck.yml` and `extract-subpacks-on-release.yml` all trigger on `pull_request` (triggers verified, not assumed), so nothing broken can merge regardless of local config.

Contributors wanting local fast feedback opt in once:

```bash
cp .claude/project-config.example.json .claude/project-config.json
git config core.hooksPath .githooks
```

## Call sites fixed

| Site | Was |
|---|---|
| `.claude/skills/setup/SKILL.md` | prose instructing the operator to stage it |
| `.claude/skills/split-portfolio/SKILL.md` | staged alongside `.gitignore`, then committed |
| `.claude/skills/update/SKILL.md` | staged to show the operator a diff — now writes a `.bak` and diffs against that, since git holds no copy |
| `docs/multi-project.md` | two staging examples |

None force the add; that would restore a tracked copy and recreate the bug.

**Mechanism correction:** the review described git as aborting wholesale with `.gitignore` never staged. Verified in a scratch repo — git stages the other paths and errors only on the ignored one. The breakage is the non-zero exit stopping the step, not lost staging.

## Migration

Adopters who committed this file run one command — **back it up first**, it is not recoverable from git:

```bash
git rm --cached .claude/project-config.json
```

Syncing before doing that hits a modify/delete conflict. Git leaves your file on disk, so nothing is lost *yet* — but the obvious resolution destroys it:

| Resolution | Effect |
|---|---|
| `git rm <file>` | ❌ index **and disk** — config gone, nothing to restore from |
| `git rm --cached <file>` | ✅ index only; file survives, ignore rule applies |

Simulated end-to-end, not reasoned. Documented in `docs/project-config.md`.

## Testing

`.claude/hooks/tests/test_project_config_untracked.sh` — the file labels which cases prove the fix and which merely guard it:

| Case | Kind | Unfixed | Fixed |
|---|---|---|---|
| `config-untracked` | **discriminating** | FAIL | PASS |
| `example-tracked` | **discriminating** | FAIL | PASS |
| `example-valid-json` | quality | SKIP (counted) | PASS |
| `gitignore-entry-present` | regression guard | pass | pass |
| `defaults-pre-push-empty` | regression guard | pass | pass |

Re-proven against a simulated unfixed tree after restructuring: `passed: 2 failed: 2 skipped: 1` → `passed: 5 failed: 0 skipped: 0`.

Two ledger bugs fixed in review: a case printed `SKIP` while never appearing in the counts, and a jq-absent path incremented neither counter, silently under-reporting the suite.

| Check | Result |
|---|---|
| Full hook suite | **PASS=122 FAIL=0** |
| `shellcheck --severity=warning` | clean |
| `markdownlint-cli2` on 5 changed files | 0 issues |

Refs #1031

---

## Glossary

| Term | Definition |
|------|------------|
| tracked file | A file in git's index; `.gitignore` has no effect on it — the whole bug |
| `git rm --cached` | Removes from the index but leaves the file on disk |
| modify/delete conflict | Merge state where one side edited a file and the other deleted it; git keeps your copy and asks you to choose |
| discriminating test | Proven to FAIL against the unfixed code, so passing it is evidence the fix works |
| regression guard | Passes before and after; prevents reintroduction rather than proving this fix |


atlas-apex (commented):
## Code Review: PR #1065

**Commit**: `6d790e4acc6f5888a4bcc2fc9aa3d4981f4fae37`

### Summary

Untracks `.claude/project-config.json` (whose `.gitignore` entry was inert because git ignores `.gitignore` for tracked files) and ships a tracked `.claude/project-config.example.json` in its place, mirroring the `onboarding.example.yaml` pair. The diagnosis is correct, the data-loss framing is justified rather than dramatised, and the core fix is right.

The blocking problem is item 4: the call-site sweep missed two, and the miss is not cosmetic. It hard-breaks a shipped bootstrap flow.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (with one labelling nit)
- Security: **Pass**
- Performance: **N/A**
- Technical Decisions (AgDR): **N/A** (agreed, see below)
- Adopter Handbooks: **Pass**. `architecture/migration-safety.md` (blocking) was loaded and ruled **N/A**: its trigger requires a schema-migration file path, and this diff contains none.
- Fallow Static Analysis: **N/A** (no JS/TS)

---

## Item 1: the central design claim. Your reasoning holds, and the exposure is wider than you argued

Verified directly rather than by reading the comment. Against `_lib-read-config.sh:144`'s `jq -s '.[0] * .[1]'`:

| Adopter's override file | Resulting `.pre_push.commands` |
|---|---|
| defines no `pre_push` | **inherits the defaults' commands verbatim** |
| defines its own `pre_push.commands` | theirs wins |

So rejecting `project-config.defaults.json` was correct. Three things strengthen it beyond what the PR body claims:

1. **`*` is a recursive object merge, not a top-level replace.** An adopter who defines a *partial* `pre_push` (say only a `_comment`, which is exactly the shape this repo's own files use) still inherits `commands` from defaults. The blast radius is wider than "adopters who define no `pre_push`".
2. **`_config_load` falls back to `cat "$defaults"` when no override file exists** (`_lib-read-config.sh:146`). Post-#1031 that is the framework's own state on a fresh clone, so defaults-sourced commands would apply with certainty, not merely by merge accident.
3. **The codebase already agreed with you.** `project-config.defaults.json` ships `pre_push.commands: []` with a `_comment` stating the empty list is deliberate and that per-fork overrides populate it. Putting the framework's commands there would have contradicted an existing, documented design intent, not merely created a new hazard.

And the `test_subpack_extraction.sh` concern is real rather than theoretical: the framework ships that file, so it exists in every fork and would genuinely execute.

**Conclusion: the fork in the road was taken correctly.** Had you chosen defaults, this would be a regression PR.

One consequence: because the point is this well-supported, the PR body over-argues it. See the advisory note at the end.

---

## Item 2: the degraded pre-push path. Acceptable, but the load-bearing sentence is factually wrong

You asked me to push back here, and there is something real, though not the thing you expected.

The trade itself is fine. What is not fine is the evidence used to make it fine. The PR body and `bin/run-pre-push-checks.sh:20-24` both claim that a terminal push still runs the checks via `.githooks/pre-push`, because that script hardcodes its command set.

That holds only in a clone where someone already opted in. `.githooks/pre-push` requires `core.hooksPath` to be pointed at `.githooks`, which `docs/getting-started.md:140` presents under the heading **"Optional: Terminal push hook"**, and which is per-clone local config a fresh clone does not inherit. In this very worktree:

```
$ git config --get core.hooksPath
/Users/ahmed/Projects/Apex/apexyard/.git/hooks     # NOT .githooks
```

So on a genuinely fresh clone, **both** pre-push paths are inactive: the Claude Code path because `project-config.json` is now absent, and the terminal path because `core.hooksPath` was never set. The PR presents the terminal path as the safety net that makes the Claude-path no-op tolerable, and that net is opt-in and off by default.

**My independent conclusion: the trade is still acceptable, but for a different reason than the one given.** The real backstop is CI. `.github/workflows/` runs the same markdownlint, shellcheck and subpack checks server-side, so nothing broken can actually land on a branch. The pre-push gate is a latency optimisation, not the guardrail. That framing survives scrutiny; "the terminal path still runs" does not.

This is not a merge blocker on its own, but the sentence should be corrected in both places rather than shipped as stated. A future contributor will otherwise rely on a gate that is not there.

Also worth noting that the `onboarding.yaml` analogy is doing more work than it can bear: a missing `onboarding.yaml` degrades *configuration*, while a missing `project-config.json` degrades a *quality gate*. Different class. The CI argument covers the gap; the analogy alone does not.

---

## Item 3: the test's honesty. Labelling is accurate

Verified against the pre-fix tree rather than taking the labels on trust:

| Case | Label | Pre-fix reality | Verdict |
|---|---|---|---|
| `config-untracked` | discriminating | `git ls-tree dev` lists `.claude/project-config.json`, so it was tracked and the case **FAILs** | label correct |
| `example-tracked` | discriminating | `git ls-tree dev` does not list the example, so the case **FAILs** | label correct |
| `gitignore-entry-present` | regression guard | `git show dev:.gitignore` already has the entry, so it passes | label correct |
| `defaults-pre-push-empty` | regression guard | `git show dev:...defaults.json` already has `length == 0`, so it passes | label correct |

All four labels are honest. `config-untracked` genuinely discriminates and neither guard does. The test runs green here (4/4). Edge cases behave as described: outside a work tree it prints `SKIP` and exits 0, and the jq-absent branches degrade without false failures.

Two nits, both non-blocking:

- **nit:** case 3 folds two distinct assertions ("is tracked" and "is valid JSON") under one ledger name, and when jq is missing it prints `SKIP [example-valid-json]`, a case name that never appears in the pass/fail counts. Either register it as a real fifth case or reuse the `example-tracked` name in the SKIP line.
- **nit:** when jq is absent, case 4 increments neither `PASS` nor `FAIL`, so the summary silently reads `passed: 3` with no failure. Harmless, but the totals are not fixed across environments. Worth a one-line comment so a future reader does not misread a short count as a regression.

---

## Item 4: BLOCKING. Two missed call sites, and the miss breaks `/split-portfolio`

Your judgement on `docs/getting-started.md` and `docs/upgrading.md` is **correct**. Both reference `project-config.json` only as the file adopters *create* for overrides, which is still accurate. Likewise the roughly 40 "override via `.claude/project-config.json`" mentions across hooks and `marketplace/`: all still true.

But two shipped skills do not merely *mention* the file. They **stage and commit it**:

- `.claude/skills/setup/SKILL.md:178` instructs: *"Stage `.gitignore`, `.claude/project-config.json`, and `.apexyard-fork` for commit. All three are per-fork, not per-machine."*
- `.claude/skills/split-portfolio/SKILL.md:341` runs a staging command listing `.gitignore` alongside `.claude/project-config.json`, immediately followed by a commit of the split-portfolio configuration.

Both sit inside scripted blocks that run right after the skill writes a `portfolio` block to that file. Post-merge, staging an ignored *untracked* path fails outright (verified in a scratch repo):

```
The following paths are ignored by one of your .gitignore files:
.claude/project-config.json
hint: Use -f if you really want to add them.
exit status 1
```

Consequences, worst first:

1. **`/split-portfolio` breaks.** The staging command lists multiple paths, and git aborts the whole operation on the rejected path, so `.gitignore` is left unstaged too and the following commit mis-fires. This is the framework's destructive migration skill failing at its commit step.
2. **The instruction is now semantically wrong regardless of exit code.** Committing that file is precisely the bug this PR fixes. Worse, in split-portfolio mode the committed content is the `portfolio` block naming the **private sibling repo path**, published to a public ops fork. That is the leak-protection concern arriving through a different door.
3. This was already wrong *before* this PR: those skills were committing private portfolio config, which is exactly the data-loss and privacy vector #1031 describes. The change converts a silent misbehaviour into a loud failure, which is an improvement, but shipping the loud failure without the fix leaves `/split-portfolio` broken.

**Required fix:** drop `.claude/project-config.json` from both staging lines and correct the surrounding prose to say the file is deliberately local and uncommitted. Do **not** reach for the force-add flag; that would re-create the exact bug. `/split-portfolio` should additionally note that the file it just wrote is intentionally untracked, so the operator is not surprised by its absence from the commit.

I checked `/update`'s own skill files and `_lib-read-config.sh`'s header. Both are clean.

---

## Item 5: migration safety for existing adopters. Thin, one real gap

The `--cached` untrack command plus "back it up first" is the right instruction, and putting it in both `docs/project-config.md` and the PR body is reasonable. Two gaps:

**suggestion (medium):** the guidance does not cover the `/update` path, which is how adopters will actually encounter this. On their next sync, upstream deletes a file they have committed locally. If they modified it since their fork point they get a modify/delete conflict, and the natural resolution, the one the docs prime them for, is a plain removal, which deletes it **from disk** and takes the private `portfolio` block with it. That is the same unrecoverable loss this PR exists to prevent, reached by a different route. Worth an explicit line: if `/update` reports a conflict on this file, keep the working-tree copy and resolve with the `--cached` form only, never a plain removal.

**suggestion:** consider whether `/update` should detect the tracked-file state and offer the `--cached` untrack itself, rather than relying on an adopter reading `docs/project-config.md`. Not required for this PR; a follow-up ticket is a fair resolution. But silently leaving every existing fork in the broken state is a weaker outcome than the diagnosis deserves. Either way, say so explicitly rather than leaving it implied.

I have **not** verified the exact git merge behaviour for the conflict case empirically. Treat the mechanism above as reasoned rather than proven, and confirm before writing it into docs.

---

## Item 6: standard checks

- **PR body**: Summary, Testing and Glossary all present; `Refs #1031` present. Bullets are narrative (what plus why), not label-only. **Pass.**
- **Commit message**: explains *why*, not just *what*, with the correct `type: subject` shape and a `Refs #1031` footer. **Pass.**
- **AgDR**: **agreed, not needed.** It adopts the existing `onboarding.example.yaml` precedent rather than setting a new pattern, adds no dependency, touches no security control or schema, and is reversible. The threshold in `.claude/rules/agdr-decisions.md` is not met. The adopter-facing migration step is the one thing that argues for a record, and that is adequately served by `docs/project-config.md` plus the PR body.

- **suggestion, prose volume.** The explanation is roughly five to six times the size of the behaviour change. A 20-line example file and one untrack command are justified across a long PR body, a 40-line test header, a 20-line comment block in `run-pre-push-checks.sh`, and two new doc sections, with the `jq -s` merge rationale restated nearly verbatim in all four. The reasoning is sound; the repetition means a future reader must diff four copies to learn whether they agree. Recommend keeping the full argument in `docs/project-config.md` and having the other three point at it in one line each.
- **nit:** the `_comment` in `project-config.example.json` is a single ~700-character string carrying four separate points. Since it is a template adopters will copy and edit, splitting it into the per-section `_comment` shape the rest of the file already uses would read better.

### Verdict

**CHANGES REQUESTED**

One blocking item only: the two staging call sites for `.claude/project-config.json` in `/setup` and `/split-portfolio` (item 4). The item 2 correction should ride along in the same push, since it is two sentences. Everything else is advisory.

The core change is right, and I would approve it immediately with those fixed.

---
Reviewed commit: `6d790e4acc6f5888a4bcc2fc9aa3d4981f4fae37`

atlas-apex (commented):
## Code Review (round 2): PR #1065

**Commit**: `6cdac358589c4d2761e9254c9259fbbcb83ffae8` (was `6d790e4`)

Delta re-review only — item 1 (the defaults-vs-example design call) was validated in round 1 and is not re-litigated.

### Verdict

**APPROVED**

All four round-1 findings are cleared, three of them fixed more thoroughly than I asked for. Approval marker written at this SHA.

---

### Your mechanism correction: you are right, I was wrong

I re-tested it in a scratch repo rather than take either of our words for it, staging `.gitignore` and the ignored path together:

```
The following paths are ignored by one of your .gitignore files:
.claude/project-config.json
exit status 1

$ git diff --cached --name-only
.gitignore          <-- staged
```

Git stages the acceptable paths and errors only on the ignored one. My "aborts wholesale, so `.gitignore` is left unstaged" was wrong. The correct mechanism is the one you wrote into `.claude/skills/split-portfolio/SKILL.md:341` — git stages the others, and the **non-zero exit stops the step**, so the commit that follows never runs. The finding stands on the exit code alone; the comment in the skill states this accurately and I would not change it.

Worth noting the failure mode is slightly *worse* than my version, not better: a partial stage plus a stopped step leaves the working tree in a half-committed state, whereas a clean abort would have left nothing staged. Your comment captures that.

### Finding 1 (was BLOCKING) — CLEARED, and the sweep was wider than mine

Five sites, not the two I found. I missed `.claude/skills/update/SKILL.md`, and both `docs/multi-project.md` instances. Re-grepped the tree at this HEAD: no staging site for `.claude/project-config.json` remains. The surviving matches are all prose about *adding config keys* (`.ticket.prefix_whitelist`, etc.), not git staging. No `-f` anywhere.

The `/setup` and `/split-portfolio` fixes correctly give the reason as well as the instruction, including the public-fork leak angle and an explicit "do not reach for the force-add flag" with the consequence spelled out. That is the right shape — a future contributor hitting the exit-1 will otherwise "fix" it with `-f` and silently restore the bug.

**The `/update` fix is the best change in this round.** You spotted something I did not: that call site staged the file *for a purpose* — giving the operator a reviewable diff of an in-place deprecated-key edit — so deleting the stage would have removed a safety property rather than a bug. Writing a `.bak` first and diffing against that preserves the property, and your comment names precisely why it matters: git holds no copy, which is exactly the condition that makes an unrecoverable edit need one. Dropping the `git add` alone would have been a silent regression.

### Finding 2 (item 2, medium) — CLEARED

`bin/run-pre-push-checks.sh:14-40` now states that both paths are inactive on a fresh clone and gives CI as the actual backstop. I verified the replacement claim rather than accepting a corrected sentence on trust — that being the whole lesson of the round-1 finding, and `extract-subpacks-on-release.yml` being a name that reads like it only fires on a tag:

| Workflow | `on:` block |
|---|---|
| `markdown-lint.yml` | `pull_request`, paths `**/*.md` |
| `shellcheck.yml` | `pull_request`, paths `.claude/hooks/**` |
| `extract-subpacks-on-release.yml` | `push: tags v*` **and** `pull_request` with an explicit paths list |

All three genuinely carry a `pull_request` trigger. The claim holds. Good instinct checking the real trigger blocks rather than the file's own header comment — the header of the third one is accurate, but it is exactly the kind of comment that was wrong last round.

**nit (non-blocking, no action needed):** the triggers are path-filtered, so "nothing broken can merge" is marginally stronger than the config supports — `shellcheck.yml` only watches `.claude/hooks/**`, which does not include `bin/run-pre-push-checks.sh`, the very file carrying the sentence. But the local command has the identical scope (`find .claude/hooks -maxdepth 1`), so CI and local are at parity and no coverage is lost by this PR. Not worth another round.

### Finding 3 (item 5, medium) — CLEARED, upgraded from reasoned to proven

I explicitly flagged my modify/delete mechanism as unverified. You simulated it and documented the result as a two-row table in `docs/project-config.md`, showing that the file survives the conflict and that the destructive step is the resolution, not the merge. That is the correct framing and it is stronger than what I handed you. The table makes the trap legible in one glance, which prose would not have.

### Finding 4 (item 3 nits) — CLEARED, and they were real bugs

Both nits turned out to be ledger defects rather than cosmetics. `example-tracked` is split into a discriminating tracked-check plus an independently-ledgered `example-valid-json`, the phantom SKIP label is gone, `skipc()` counts skips, and the summary prints them. The `skipc` comment states the underlying principle — an un-counted skip makes the suite read smaller than it is, which is how a disabled assertion hides. That generalises beyond this test.

Re-proving discrimination after restructuring was the right call, since a restructure can silently neuter a case. Runs green here:

```
PASS [config-untracked]  PASS [gitignore-entry-present]  PASS [example-tracked]
PASS [example-valid-json]  PASS [defaults-pre-push-empty]
passed: 5   failed: 0   skipped: 0
```

### Prose advisory — CLEARED

The `jq -s` rationale now lives once in `docs/project-config.md` with one-line pointers from the test header, the example's `_comment`, and `run-pre-push-checks.sh`. The example `_comment` drops from ~700 characters to a readable sentence plus a pointer.

Your observation that the false claim in item 2 was *living inside the duplication* is the sharper version of the advisory, and worth carrying forward: duplicated rationale does not just cost reading time, it multiplies the number of places a wrong claim has to be found and corrected. Four copies is four chances to fix three and ship one.

### Verification

Confirmed independently at this HEAD: the 5-case test passes, no staging sites remain, and all three cited workflow triggers are real. I did not re-run the full 122-case suite or markdownlint, taking your reported results for those.

---
Reviewed commit: `6cdac358589c4d2761e9254c9259fbbcb83ffae8`

me2resh (approved):

Issue #1063, opened 2026-07-29
[Bug] audit_run_persist leaves a zero-byte run file when jq fails


atlas-apex:
## Driver

Found by Hakim while persisting the security-review artefact for PR #1061. It left two zero-byte run files on disk, which Hakim removed manually.

`audit_run_persist` opens the output redirect **before** `jq` runs:

```sh
... > "$json_path"
```

The shell truncates (or creates) `$json_path` at redirect time. If `jq` then fails — bad input, unavailable binary, malformed payload — the function leaves a **zero-byte file** at a path `audit_run_list` will subsequently treat as a valid run.

## Why it matters beyond tidiness

The audit history is the input to `audit_render_trend`, which draws the score-over-time chart for security reviews. A zero-byte run is not merely noise: depending on how the reader parses it, it either breaks the render or silently contributes a bogus point to a **security** trend line. A trend that quietly under- or over-states the security posture is worse than no trend.

It also fails in the wrong direction for an audit artefact: the operation reports success while leaving unusable state behind.

## Scope

`.claude/hooks/_lib-audit-history.sh` — `audit_run_persist`.

Write to a temporary file and move into place only on success:

```sh
tmp=$(mktemp)
if jq ... > "$tmp"; then mv "$tmp" "$json_path"; else rm -f "$tmp"; return 1; fi
```

Worth checking the same shape elsewhere in the lib, and whether `audit_run_list` / `audit_render_trend` should additionally skip zero-byte or unparseable runs defensively — belt and braces, since existing forks may already carry such files.

## Acceptance Criteria

- [ ] A `jq` failure leaves **no** file at `$json_path` (not a zero-byte one)
- [ ] `audit_run_persist` returns non-zero when persistence fails, rather than reporting success
- [ ] `audit_run_list` skips zero-byte / unparseable runs already on disk
- [ ] `audit_render_trend` renders correctly with such a file present
- [ ] Test covers the failure path, verified to fail before the fix
- [ ] No behaviour change on the success path

## Risks / Dependencies

- Independent of #1061 and the rest of the #1033 chain; can be done any time.
- Low blast radius — the lib is advisory and never gates a merge.
- Note when reproducing: the two observed zero-byte files were already cleaned up, so the repro must induce a `jq` failure deliberately.

## Glossary

| Term | Definition |
|------|------------|
| Run file | Per-audit-run JSON under `projects/<name>/audits/<dimension>/runs/` |
[...]

PR #1072, opened 2026-07-30
fix(#1063): stage-then-promote audit_run_persist so a jq failure leaves no file

atlas-apex:
## Summary

- **`audit_run_persist` now stages both the JSON and the MD artefact to a scratch file and promotes with `mv` only after the content is confirmed non-empty and re-parseable** — the old code wrote straight to the destination via `>`, which truncates/creates the file *before* `jq` runs at all. Combined with `jq` exiting 0 on empty/absent stdin (it reads no input and emits nothing), the old status-only guard reported success while leaving a zero-byte run file behind — exactly the case Hakim hit twice reviewing #1061.
- **The `.md` write is now gated on the JSON promotion having actually succeeded**, closing the second defect from the issue's follow-up comment: an orphan `.md` can no longer be written behind a JSON run that never landed.

  **Corrected in review — this closes one direction, not both.** Once the JSON `mv` succeeds, four later paths (the md `mktemp`, the frontmatter `jq`, the staged write, the md `mv`) each `return 1` while leaving a *valid* JSON run on disk. The caller is told the persist failed, yet the run has a parseable `ts` and will appear as a trend point. That is not a regression — `dev` behaves identically, and the surviving point is well-formed rather than corrupt — but the original wording claimed more than the change delivers. Narrowed rather than papered over; adding a rollback would be a larger change than the defect warrants.
- **`audit_run_list` now explicitly skips zero-byte run files** rather than relying on the incidental side effect that an empty file happens to yield an empty `.ts` — belt-and-braces for forks that already carry a pre-#1063 orphan on disk.

## The judgement call — atomic write vs. detect-and-cleanup

I went with **atomic** (stage to a temp file in the same directory, promote with `mv` only on success), not detect-then-delete. Two reasons, not just "atomic sounds safer":

1. **The two approaches need the identical validity check anyway.** `jq`'s exit status alone doesn't distinguish "wrote a good run" from "read empty stdin and reported success" — both need the `[ -s tmp ] && jq -e . tmp` check I added. The only real difference between the two designs is *where* the destination file sits while that check runs: atomic never creates it at the real path until the check has already passed; detect-and-cleanup necessarily creates it there first (that's what makes it detectable), then removes it. Given the check is the same either way, atomic buys the "no reader can ever observe the bad state" property for free.
2. **Nothing in this lib's contract promises single-threaded access.** Today, `audit_run_persist` and `audit_run_list`/`audit_render_trend` run from a single Claude Code session per invocation, so a concurrent read during the ~unlucky window is unlikely in practice. But this is a shared lib sourced by nine different audit skills, and the failure mode if that assumption ever breaks — a security-trend chart silently reading a truncated file mid-write — is exactly the "worse than no artefact" case the issue opened on. `mktemp` + `mv` costs one extra syscall pair per artefact; paying it removes the assumption from the picture entirely rather than relying on it staying true.

I staged both temp files **inside `dim_dir`** (`runs/.audit-run-tmp.XXXXXX` and `dim_dir/.audit-run-tmp.XXXXXX`), not in `$TMPDIR` as the issue's illustrative snippet showed — `mv` is only atomic within a single filesystem, and `dim_dir` can be a `workspace/<project>` mount that isn't guaranteed to share a filesystem with `$TMPDIR`.

## What `audit_render_trend` does with a zero-byte file today

Checked before writing any fix code: `audit_render_trend` never opens run files itself — it delegates discovery entirely to `audit_run_list`, and `audit_run_list`'s existing sort loop reads each file's `.ts` via `jq -r '.ts // ""' ... || ts=""` and skips anything that yields an empty `ts`. A zero-byte file makes `jq` emit nothing (rc 0, empty stdout), so `ts` comes back empty and the file is silently dropped from the sorted list *before* `audit_render_trend` ever sees it.

So the zero-byte file does **not** break the render or inject a bogus point into the security-trend chart today — it's silently excluded, by an incidental side effect of the `ts`-empty check rather than an explicit one. That's still worth hardening (case 14/15 below, and the explicit `[ -s "$f" ]` guard in `audit_run_list`), both to make the intent visible in the code rather than load-bearing-by-accident, and because a future change to the `ts`-check shape could silently remove the protection.

## Testing

Wrote the discriminating tests **first** and confirmed RED against the unmodified lib:

```
✗ audit_run_persist: empty stdin leaves no file, returns non-zero: expected non-zero rc, got 0
✗ audit_run_persist: malformed stdin leaves no file, returns non-zero: zero-byte JSON left at .../runs/2026-06-02T00-00-00Z.json
Results: 13 passed, 2 failed
```

(Case 13 confirms the pre-existing `jq ... || return 1` guard already reported failure correctly for malformed input — the survivor was purely the zero-byte file left by the `>` redirect having already fired before `jq` errored. Case 12, empty stdin, is the case the issue's follow-up comment specifically called out: `jq` exits 0 there, so the old status-only check never even considered it a failure.)

After the fix, `test_audit_history.sh` is green in full — 15 cases (11 pre-existing + 4 new: empty-stdin no-file/nonzero-rc, malformed-stdin no-file/nonzero-rc, `audit_run_list` skips a pre-existing zero-byte orphan, `audit_render_trend` renders correctly with an orphan zero-byte JSON *and* an orphan `.md` both present):

```
Results: 15 passed, 0 failed
```

Full hook suite: `bash bin/run-hook-tests.sh` — PASS=122 FAIL=0 (unchanged baseline; `test_audit_history.sh` counts as one file in this runner regardless of its internal case count, so the total didn't move).

`shellcheck --severity=warning` clean on both changed files.

No behaviour change on the success path — cases 1-11 (pre-existing) pass unmodified.

Refs #1063

---

## Glossary

| Term | Definition |
|------|------------|
| Run file | Per-audit-run JSON under `projects/<name>/audits/<dimension>/runs/` — the persisted record one audit invocation writes, consumed by `audit_run_list` / `audit_render_trend` to build the trend chart |
| Stage-then-promote | Write output to a temp file first, then `mv` it into the real destination only once the content is verified good — the destination path never exists in a partial state |
| Atomic rename | `mv` within a single filesystem is atomic at the OS level — a reader either sees the old file or the new one, never a half-written one |
| Orphan `.md` | The markdown artefact existing on disk with no matching JSON run behind it — the pair has desynchronised |

### DIFF
--- a/.claude/hooks/_lib-audit-history.sh
+++ b/.claude/hooks/_lib-audit-history.sh
@@ -185,7 +185,22 @@ audit_run_persist() {
 
   # Read stdin JSON. Augment with top-level metadata, derive stats from
   # findings[] if absent.
+  #
+  # Written to a scratch file in the SAME directory as the destination
+  # first, then promoted with `mv` (atomic rename on one filesystem) only
+  # if jq produced usable output. This is deliberately NOT just an
+  # exit-status check: jq exits 0 on empty/absent stdin -- it reads no
+  # input, emits no output, and reports success -- so a caller that
+  # forgot to pipe a payload previously got a zero-byte run file promoted
+  # AND a rc=0 "it worked" from this function (apexyard#1063). A reader
+  # (`audit_run_list` / `audit_render_trend`) must never be able to
+  # observe a partially-written or empty run file at json_path.
   local json_path="$dim_dir/runs/$ts_safe.json"
+  local json_tmp
+  json_tmp=$(mktemp "$dim_dir/runs/.audit-run-tmp.XXXXXX" 2>/dev/null) || {
+    echo "audit_run_persist: mktemp failed while staging $json_path" >&2
+    return 1
+  }
   jq --arg ts "$ts" --arg dim "$dimension" --arg v "$verdict" --argjson s "$score" '
     . as $in
     | (.findings // []) as $f
@@ -201,13 +216,35 @@ audit_run_persist() {
       + { ts: $ts, dimension: $dim, verdict: $v, score: $s,
           schema_version: ($in.schema_version // 1),
           stats: ($existing_stats // $derived_stats) }
-  ' > "$json_path" || {
-    echo "audit_run_persist: jq failed writing JSON" >&2
+  ' > "$json_tmp"
+  local jq_rc=$?
+
+  # Promotion gate: exit status alone is not enough (empty stdin exits 0
+  # with zero bytes written) -- also require non-empty AND independently
+  # re-parseable output before this run is allowed to exist on disk.
+  if [ "$jq_rc" -ne 0 ] || [ ! -s "$json_tmp" ] || ! jq -e . "$json_tmp" >/dev/null 2>&1; then
+    rm -f "$json_tmp"
+    echo "audit_run_persist: jq produced no usable JSON for $json_path (rc=$jq_rc; empty or malformed stdin?)" >&2
+    return 1
+  fi
+
+  mv "$json_tmp" "$json_path" || {
+    rm -f "$json_tmp"
+    echo "audit_run_persist: failed to move staged JSON into place at $json_path" >&2
     return 1
   }
 
-  # Build frontmatter from the just-written JSON, then concatenate with body.
+  # Build frontmatter from the just-written JSON, then concatenate with
+  # body -- staged the same way. Reaching this point means json_path
+  # already holds a valid run, so an .md file can never be written
+  # without a matching JSON run behind it (the orphan-.md defect also
+  # reported in apexyard#1063).
   local md_path="$dim_dir/$ts_safe.md"
+  local md_tmp
+  md_tmp=$(mktemp "$dim_dir/.audit-run-tmp.XXXXXX" 2>/dev/null) || {
+    echo "audit_run_persist: mktemp failed while staging $md_path" >&2
+    return 1
+  }
   local fm
   fm=$(jq -r '
     "---",
@@ -226,14 +263,25 @@ audit_run_persist() {
     "---",
     ""
   ' "$json_path") || {
-    echo "audit_run_persist: jq failed building frontmatter" >&2
+    rm -f "$md_tmp"
+    echo "audit_run_persist: jq failed building frontmatter for $md_path" >&2
     return 1
   }
 
   {
     printf '%s\n' "$fm"
     cat "$body_file"
-  } > "$md_path" || return 1
+  } > "$md_tmp" || {
+    rm -f "$md_tmp"
+    echo "audit_run_persist: failed writing staged MD for $md_path" >&2
+    return 1
+  }
+
+  mv "$md_tmp" "$md_path" || {
+    rm -f "$md_tmp"
+    echo "audit_run_persist: failed to move staged MD into place at $md_path" >&2
+    return 1
+  }
 
   _audit_apply_marker "$dim_dir"
   return 0
@@ -283,6 +331,13 @@ audit_run_list() {
       [ -d "$d" ] || continue
       for f in "$d"/*.json; do
         [ -f "$f" ] || continue
+        # Defensive: skip zero-byte / orphaned run files that may already
+        # be sitting on disk from before apexyard#1063's fix (or any other
+        # source of a truncated write). The ts-empty check below already
+        # incidentally catches this (jq on an empty/unparseable file emits
+        # nothing), but make the intent explicit rather than relying on
+        # that side effect.
+        [ -s "$f" ] || continue
         local ts
         ts=$(jq -r '.ts // ""' "$f" 2>/dev/null) || ts=""
         [ -n "$ts" ] || continue
--- a/.claude/hooks/tests/test_audit_history.sh
+++ b/.claude/hooks/tests/test_audit_history.sh
@@ -357,7 +357,112 @@ EOF
   )
 }
 
-for fn in case_1 case_2 case_3 case_4 case_5 case_6 case_7 case_8 case_9 case_10 case_11; do
+# ---------------------------------------------------------------------------
+# Case 12: audit_run_persist leaves NO file (and returns non-zero) on empty
+# stdin -- apexyard#1063. jq exits 0 on empty stdin (no findings emitted at
+# all), so the pre-fix code took the success path and moved a zero-byte JSON
+# into place while returning rc=0 -- reporting success for a run that never
+# happened. This is the discriminating case: the pre-existing `|| return 1`
+# guard already caught malformed input (jq exits non-zero there), but NOT
+# empty/absent stdin (jq exits 0 with no output).
+# ---------------------------------------------------------------------------
+case_12() {
+  local case_name="audit_run_persist: empty stdin leaves no file, returns non-zero"
+  local sb
+  sb=$(fresh_sandbox)
+  ( source_lib_with_projects_dir "$sb"
+    body=$(mktemp); echo "## Body" > "$body"
+    printf '' | audit_run_persist "demo" "threat-model" "2026-06-01T00:00:00Z" "fail" 0 "$body"
+    rc=$?
+    json="$sb/projects/demo/audits/threat-model/runs/2026-06-01T00-00-00Z.json"
+    md="$sb/projects/demo/audits/threat-model/2026-06-01T00-00-00Z.md"
+    [ "$rc" -ne 0 ] || { mark_fail "$case_name" "expected non-zero rc, got $rc"; return; }
+    [ -f "$json" ] && { mark_fail "$case_name" "zero-byte JSON left at $json"; return; }
+    [ -f "$md" ] && { mark_fail "$case_name" "orphan MD left at $md (json step never succeeded)"; return; }
+    rm -f "$body"
+    mark_pass "$case_name"
+  )
+}
+
+# ---------------------------------------------------------------------------
+# Case 13: audit_run_persist leaves NO file on malformed (non-JSON) stdin.
+# jq exits non-zero here, so the pre-fix `|| return 1` guard already
+# reported failure -- but the `>` redirect had already truncated/created
+# json_path before jq ran, so a zero-byte file survived the return 1. This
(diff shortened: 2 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #1072

**Commit**: `bfbaed0294c9035109eee50207622aef2f9d3a81`

### Summary

Replaces the direct `>` redirect in `audit_run_persist` with stage-then-promote: `mktemp` inside the destination directory, validate the staged content, then `mv` into place. Also gates the `.md` write on the JSON promotion succeeding, and makes `audit_run_list`'s zero-byte exclusion explicit.

I verified the two load-bearing claims empirically rather than accepting them, and tried to break the atomicity argument. Both hold.

### Checklist Results
- Architecture & Design:       Pass
- Code Quality:                Pass
- Testing:                     Pass
- Security:                    Pass
- Performance:                 Pass
- PR Description & Glossary:   Pass
- Summary Bullet Narrative:    Pass
- Technical Decisions (AgDR):  N/A (see note)
- Adopter Handbooks:           Pass
- Fallow Static Analysis:      N/A (not JS/TS)

### 1. The atomicity claim — I could not construct a reader-observable partial or empty run file

Attacked along every axis raised, and the claim survives:

- **Symlinked / network-mounted `dim_dir`.** This is the case the design is immune to *by construction*, not by luck. The temp is created by `mktemp` **inside the destination's own directory** (`$dim_dir/runs/`), so source and destination are two entries in the same directory. If `dim_dir` — or any ancestor — is a symlink crossing a filesystem, *both* paths resolve through that same symlink to the same directory, so they remain co-located. `mv` therefore takes the `rename(2)` path, which POSIX requires to be atomic. Staging in `$TMPDIR` (as the issue's illustrative snippet showed) genuinely would have broken this; the author's deviation from the snippet is correct and the PR body explains why.
- **Destination path itself a symlink or bind-mounted file.** Does not break it either: `rename(2)` replaces the symlink *entry*, it does not follow it. Worth noting this is a deliberate behaviour change vs. the old `>` redirect, which *did* follow a symlinked destination — the new behaviour is the safer of the two here.
- **NFS.** Same-directory rename is atomic server-side. The classic NFS caveats (silly-rename, client attribute caching) concern unlinking open files, not a reader observing a half-written file.
- **Two concurrent `audit_run_persist` calls on the same timestamp.** Cannot collide on the temp path — `mktemp` creates with `O_EXCL` and guarantees a unique name, so two racing calls get two distinct scratch files. Cannot produce a partial destination either: both promotions are atomic replaces, so the outcome is last-writer-wins with a *complete* file. This is a genuine improvement over the old code, where `>` truncated first and two writers could interleave into the same inode.
- **Validity gate ordering.** The `[ -s ] && jq -e .` check runs entirely against the temp, before any `mv`. The destination never exists in an unvalidated state.

So: **no, I could not construct a case where a reader observes a partial or empty file at `json_path`.** The one observable-imperfect state I did find is elsewhere — see finding 2.

### 2. The validity check — `jq`'s behaviour verified directly

The reasoning is correct, and I confirmed it rather than taking it on trust (jq 1.8.2):

```
printf '' | jq '. as $in | {a:1}'   -> rc=0, zero bytes of output
printf 'not json' | jq '.'          -> rc=5
jq -e . <empty file>                -> rc=4
```

Exit status alone genuinely cannot distinguish "wrote a good run" from "read nothing and succeeded" — rc=0 in both. So `[ -s tmp ]` plus an independent re-parse is necessary regardless of which design was chosen, which is exactly the argument the PR body makes for atomic-over-detect-and-cleanup being nearly free. The justification stands.

### Issues Found

None blocking. Three non-blocking observations:

**2. `nit:` The body overclaims desync closure as bidirectional.** The PR body says the two files "could desync in both directions" and presents the `.md` gating as closing that. It closes **one** direction (orphan `.md` with no JSON). The reverse is still reachable: once `mv "$json_tmp" "$json_path"` succeeds, four subsequent paths — md `mktemp` failure, frontmatter `jq` failure, staged md write failure, md `mv` failure — each `return 1` while leaving a **valid** JSON run on disk. That run has a parseable `ts`, so `audit_run_list` will include it and it will appear as a trend point, while the caller was told the persist failed.

This is **not a regression** — `dev` has the identical shape — and the trend point is well-formed rather than corrupt, so nothing breaks. But the claim is broader than the change. I'd narrow the sentence rather than add a JSON rollback; a rollback introduces its own partial-failure question (what if the rollback `rm` fails?) for a strictly rarer case. Non-blocking either way.

**3. `nit:` No `trap` cleanup — a signal between `mktemp` and `mv`/`rm -f` leaks a scratch file.** I checked whether your specific concern lands ("a leaked temp inside `dim_dir` could itself be picked up by a glob") and it **does not**: `audit_run_list` globs `"$d"/*.json`, and the temp is `.audit-run-tmp.XXXXXX` — a dotfile (excluded from `*` absent `dotglob`) *and* without a `.json` suffix. Two independent reasons it can't be picked up. The naming is well chosen.

Residual cost is therefore bounded to disk litter, plus one cosmetic edge: in a dimension with the `.audit-history-tracked` marker, the generated `dim_dir/.gitignore` contains only negations (`!runs/`, `!runs/*.json`), so a leaked **md** scratch file would surface as untracked in `git status`. Acceptable as-is in my view; a `trap ... RETURN` in a sourced lib function has its own surprise surface for callers. Worth a one-line comment noting the deliberate omission if you touch this again.

**4. `nit:` `local jq_rc=$?` is correct** — `$?` is expanded during word expansion, before `local` runs, so the `local var=$(cmd)` status-masking gotcha does not apply. Confirmed indirectly by case 13 reporting `rc=5`. Flagging only because it reads like the known footgun and a future reader may "fix" it.

I also checked for a `set -e` robustness regression, since the guarded `jq ... || { ...; return 1; }` became a bare `jq` plus a captured `$?`. **No regression**: with a passing control call in the same shell, both `dev` and this branch abort identically under `set -e` on the failing call (the abort comes from the caller seeing a non-zero function return, which was already true). No caller sets `set -e` in any case.

### 3. Scope creep — both additions belong here

- **`.md` gating on JSON promotion:** keep. Same function, same trigger, same defect class. Splitting it would mean shipping a fix that still leaves a desynced pair from the identical failure, which is a worse outcome than a slightly wider PR.
- **`audit_run_list`'s `[ -s "$f" ]`:** keep, though it's the weakest of the three. It's one line, covered by case 14, and it converts an *accidental* protection (zero-byte file happens to yield an empty `.ts`) into an explicit one — which matters precisely because a future refactor of the `ts` check would silently remove it. It's also the migration path for forks already carrying a pre-fix orphan.

Neither warrants its own ticket. The PR did not quietly grow — both are named in the body and individually tested.

### 4. Test honesty — verified, and the labelling is accurate

I ran this PR's test file against `dev`'s unmodified lib in an isolated copy:

```
✗ audit_run_persist: empty stdin leaves no file, returns non-zero: expected non-zero rc, got 0
✗ audit_run_persist: malformed stdin leaves no file, returns non-zero: zero-byte JSON left at .../runs/2026-06-02T00-00-00Z.json
Results: 13 passed, 2 failed
```

Byte-for-byte the output the PR body claims. Cases **12 and 13 are genuinely discriminating**. Cases **14 and 15 pass against `dev`** — they are regression locks, not proof — and the body labels them correctly ("belt-and-braces", "defensive") while the RED block claims only 2 failures. No case is presented as evidence it isn't. This is honest.

Against this branch: 15 passed, 0 failed.

### 5. Success path — genuinely untouched

Cases 1–11 pass identically in both runs above (against `dev`'s lib and against this branch), including the two `launch-check` legacy-dispatch cases. The changed code sits entirely on the failure path plus a `mv` that replaces a completed `>`. Confirmed no behaviour change for the nine audit skills sourcing this lib.

### 6. Standard checks

- Title `fix(#1063): ...` matches `type(TICKET): description`.
- `Refs #1063` present (correct — QA gate, not auto-close).
- Glossary present, 4 terms, all genuinely used in the body.
- Summary bullets are narrative — what changed *and* why it matters. No label-only bullets.
- Commit message: meaningful subject, substantive WHY-focused body, no placeholders. Passes `handbooks/general/commit-message-quality.md`.

**AgDR — I agree it isn't needed, with one note.** `.claude/rules/agdr-decisions.md` rail 1 reads literally as "anything touching `.claude/hooks/**` is material regardless of diff size", which would pull this in. On the merits it doesn't: the trust chain that rail protects is enumerated as merge gates, the tracker/forge lib, the review-marker lib, and credential brokers. `_lib-audit-history.sh` is audit-artefact persistence sourced by audit *skills* — it gates nothing and touches no credentials; it lives under `.claude/hooks/` by directory convention alone. Neither `require-agdr-for-arch-changes.sh` nor `require-agdr-for-arch-pr.sh` lists that path. And the decision itself (atomic vs. detect-and-cleanup) is reversible inside this PR, adds no dependency, and establishes no repo-wide pattern. The reasoning is already recorded at length in the PR body, which is the proportionate artefact. No AgDR.

### Handbook Findings

- `handbooks/architecture/migration-safety.md` (**blocking** handbook): **not triggered** — the diff contains no file matching its migration globs. Checked explicitly because it is the one blocking handbook loaded.
- `handbooks/general/commit-message-quality.md`: pass (see above).
- `handbooks/architecture/clean-architecture-layers.md`: N/A — shell hook lib, no layered domain code.
- `handbooks/language/typescript/strict-mode.md`: not loaded (diff is `.sh`).

### Suggestions

1. Narrow the "desync in both directions" sentence in the PR body to the direction actually closed (finding 2). Body-only; no code change.
2. Optional: a one-line comment recording that temp cleanup on signal is deliberately omitted, and why it's safe (dotfile + no `.json` suffix means no glob can see it). Saves the next reader the analysis I just did.

### Verdict
**APPROVED**

Strong PR. The judgement call is documented with real reasoning rather than "atomic sounds safer", the discriminating tests were written first and proven RED with recorded output, and the deviation from the issue's own illustrative snippet (staging in `dim_dir` rather than `$TMPDIR`) is the detail that makes the atomicity claim actually true rather than merely plausible.

---
📌 Reviewed commit: `bfbaed0294c9035109eee50207622aef2f9d3a81`

me2resh (approved):

Issue #1050, opened 2026-07-28
[Bug] verify-commit-refs.sh resolves the tracker repo from the hook's cwd, not the committed worktree — false-blocks valid refs and teaches agents to bypass the gate


asami-me:
## Affected

`.claude/hooks/verify-commit-refs.sh` (and, latently, the same pattern in several sibling hooks — see Notes)

## Given / When / Then

**Given** an adopter running in split-portfolio mode, with a managed project checked out under the portfolio's `workspace/<project>/`, and a sub-agent working inside a **linked git worktree** of that project (the shape `/fan-out` and `Workflow` produce, and which `require-active-ticket.sh` explicitly supports via its tier-0 per-worktree markers)

**When** that sub-agent runs `git commit -m "... Closes #N"`, where `#N` is a real, open issue in **the managed project's** tracker

**Then** the hook blocks the commit, reporting that `#N` does not exist — because it resolved the tracker as the **ops fork's** repo (`me2resh/apexyard`) rather than the project's. Expected: the hook resolves the tracker from the repository the commit is actually being made in.

## Repro

1. Split-portfolio adopter. Managed project registered at `workspace/<project>/`, its own GitHub tracker.
2. Create a linked worktree of that project (e.g. `<project>/.claude/worktrees/<branch>`), and an issue `#N` in the project's tracker.
3. Have a sub-agent work in that worktree and commit with `Closes #N`.
4. Hook blocks: `#N` not found. `gh issue view N --repo <owner>/<project>` confirms it is real and OPEN.

Reproduced independently by two different sub-agents in one session, on two different tickets.

## Root cause

The hook resolves the repository **relative to its own working directory**, never from the file or worktree being committed:

```
:137   REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
:153   ORIGIN_URL=$(git remote get-url origin 2>/dev/null)
```

Both inherit the harness's session cwd, which is the **ops fork**, not the managed project's worktree. `grep -n 'GIT_DIR\|--git-dir\|worktree\|CLAUDE_WORKTREE'` over the hook returns nothing — it has no worktree or file-path awareness at all.

**The framework already has the correct pattern.** `require-active-ticket.sh` derives the target repo from the tool call's `FILE_PATH` and runs git *in that directory*:

```
:357   _gd=$(git -C "$_fdir" rev-parse --absolute-git-dir 2>/dev/null)
[...]

Issue #207, opened 2026-05-11
[Bug] verify-commit-refs.sh forces cross-repo notation that breaks GitHub auto-close (12+ manual closures per session)


atlas-apex:
## Given / When / Then

**Given** a fork (e.g. atlas-apex/apexyard) with `origin` = the fork and `upstream` = me2resh/apexyard, and an open PR on a feature branch that closes an issue on **upstream** (me2resh/apexyard#NNN)

**When** the operator runs `git commit -m "Closes #NNN"` where issue NNN exists on me2resh/apexyard but NOT on atlas-apex/apexyard

**Then** `verify-commit-refs.sh` rejects the commit because the issue doesn't exist in the LOCAL origin's tracker, forcing the operator to use cross-repo notation `Closes me2resh/apexyard#NNN` — which then doesn't trigger GitHub's auto-close when the PR merges (auto-close only fires on **bare** `#N` notation when PR target = issue host).

**Expected**: bare `#NNN` should be valid when the issue exists in EITHER `origin` OR `upstream`. PRs targeting upstream then use bare notation, GitHub's auto-close works on merge, no manual issue-closure dance.

## Repro

Reproduced **every session** that opens PRs from a fork against upstream:

```
# In atlas-apex/apexyard fork, working on a feature branch
git commit -m "Closes #150"  # 150 exists on me2resh/apexyard, not atlas-apex/apexyard
# → BLOCKED by verify-commit-refs.sh

# Workaround: use cross-repo notation
git commit -m "Closes me2resh/apexyard#150"  # passes the hook
# → After PR merge, #150 stays OPEN because GitHub doesn't auto-close cross-repo
```

In a single recent session, this caused **12+ manual issue closures** across two batches:

- Batch 1: #178, #181, #182, #183, #188, #189, #190 — closed manually
- Batch 2: #177, #179, #180, #194, #195, #197 — closed manually

Each closure is a separate `gh issue close` with a "Closed by PR #X — cross-repo notation didn't auto-close" comment.

## Why this matters

Three knock-on effects:

1. **Tracker pollution** — closed-PRs leave their issues OPEN indefinitely unless someone remembers to close them by hand. Every session adds N stale-but-done issues.
2. **Stakeholder confusion** — `gh issue list --state open` shows work that's actually shipped. Status reports / `/inbox` / `/launch-check` all over-report open backlog.
[...]

Issue #1068, opened 2026-07-30
[Bug] Leak hook: a trailing --repo anywhere on the line disables the gate entirely


atlas-apex:
## Bug Scenario

**Given** a command that writes to a public framework repo and ALSO mentions `--repo` later on the same line — a chained second `gh` call, or even a trailing shell comment

**When** the hook resolves the target repo

**Then** it binds the **last** `--repo` on the line rather than the one belonging to the write, decides the target is not public-class, and exits 0 at step 3 — **before scanning anything**

**Expected** the target repo should be resolved from the write command itself, so a later unrelated `--repo` cannot disable the gate

## Repro Steps

Synthetic registry with project name `zebrafish` / repo `acme/zebrafish-app`:

```bash
# LEAKS — rc=0, nothing scanned
gh issue create --repo me2resh/apexyard --title "T" \
  --body "found during zebrafish rebuild" && gh pr list --repo acme/zebrafish-app

# LEAKS — a bare trailing comment is enough
gh issue create --repo me2resh/apexyard --title "T" \
  --body "found during zebrafish rebuild" # see --repo foo/bar

# CONTROL — same command without the trailing --repo
gh issue create --repo me2resh/apexyard --title "T" \
  --body "found during zebrafish rebuild"        # rc=2, blocked correctly
```

The control blocking is what makes this a bypass rather than a match failure.

Cause: `TARGET_REPO=$(echo "$COMMAND" | sed -nE 's/.*--repo[[:space:]]+...`. The leading `.*` is greedy, so it binds the final `--repo`. The `| head -1` de-dupes *lines*, not matches within a line.

## Environment

`dev` @ 8822d05, and unchanged on the #1067 branch. Verified on both.

## Severity

**high** — the gate is silently and completely disabled. Not a partial scan: it exits before extraction.

The shape is not exotic. It is precisely the scenario `.claude/rules/leak-protection.md` exists for: an agent files an upstream ticket, then touches the private tracker in the same chained call. That is a natural thing for an agent to write, and the failure is silent — `gh` succeeds and nothing indicates the body was never examined.

## Mitigation

None automatic. Avoid chaining a second `--repo` command onto an upstream write. There is no signal when it happens.

## Investigation Notes

Found during the security review of #1067 (fix for #1046). **Pre-existing and unrelated to that PR** — deliberately not folded into it, since #1067 already had one regression introduced mid-flight and widening it further would repeat the #1040 mistake of bundling unrelated changes into a narrow security fix.

[...]

PR #1067, opened 2026-07-30
fix(#1046): close the leak-hook trim fail-open and correct its bypass advice

atlas-apex:
## Summary

- **Closes a silent fail-open on the leak gate.** A body containing an embedded quote *and*, later, any double-dash token had its tail amputated before scanning. The hook exited 0, `gh` succeeded, and nothing signalled that half the body was never examined — so a private project name in that tail reached a public tracker.
- **Fixes the trim, not the match anchor.** The anchor was never the defect, and widening it is the documented trap that sank the previous attempt (#1040 reopened #227: 9 failures vs 5, worse than no fix). All 29 pre-existing cases still pass.
- **Also fixes the single-quoted branch, which is not in the issue** — it carried the identical defect and leaked on the single-quoted form of the same command.
- **Corrects the fail-closed message** that advised a bypass which could never work.

## Severity: this is worse than the issue's `minor`

#1046 is filed as *"a pre-existing scan gap needing an unusual body shape"*. On the evidence it is a **silent fail-open on a leak gate** — the same class as #1039, which is rated `high`.

Per [`leak-protection.md`](.claude/rules/leak-protection.md), the mechanical hook is the defence here *precisely because* self-discipline fails at this moment: the private name is right there in the working context while the agent writes the upstream ticket. When the gate silently passes, nothing else catches it, and a private name indexed on a public tracker is unrecoverable.

The trigger is narrower than #1039's. But quoting a term and then naming a CLI flag is an ordinary thing to write in an upstream bug report.

## The issue's own repro does not reproduce — read this before verifying

The repro on #1046 is a bare `--verbose` in prose. That **blocks correctly**. Anyone verifying this issue as filed would reasonably conclude "not reproducible" and close it, while the leak stayed open.

**Two conditions must co-occur:**

| Body shape | Before this PR |
|---|---|
| embedded quote, no `--flag` after it | blocks ✓ |
| `--flag` present, no embedded quote | blocks ✓ |
| **embedded quote, then ` --flag`** | **exits 0 — tail unscanned** |

Both single-condition controls are now asserted in the test file so this cannot be re-lost.

## Mechanism

`extract_flag_value`'s greedy `match()` captured the body correctly. The trim that followed did not:

```awk
sub("\"([[:space:]]+--[a-zA-Z].*)?$", "", chunk)
```

POSIX ERE substitution is **leftmost-longest**. It therefore found the *earliest* `"` that happened to be followed by ` --<letter>` and deleted from there to end-of-string.

[...]

PR #1073, opened 2026-07-30
fix(#1050): resolve verify-commit-refs.sh's tracker repo from the commit's own directory

atlas-apex:
## Summary

- **`verify-commit-refs.sh` was validating `Closes #N` against the wrong repo's tracker in worktree-based sessions.** The hook resolved `REPO_ROOT`, the ops-fork walk, and both the `origin`/`upstream` remote lookups with bare, unqualified git calls — all of which run against the *hook process's own cwd*, never the directory the `git commit` is actually executing in. In a split-portfolio / worktree session (the shape `/fan-out` and `Workflow` create), the harness's Bash cwd for a given call can be a managed project's worktree while the hook process itself inherits a different cwd (typically the ops fork) — so a legitimate `Closes #N` against the project's own real, open issue got checked against the ops fork's tracker instead, and false-blocked. Reported independently by `asami-me` in #1050, with a second confirmation comment reproducing it verbatim against the released v5.3.0 tag.
- **Fix: resolve the commit's real working directory before any git call, same pattern `require-active-ticket.sh` already uses for Edit/Write hooks.** A new `resolve_commit_workdir()` helper picks the directory in priority order — an explicit `git -C <path>` on the invocation, then a `cd <path> &&` prefix in the same compound command, then the harness-provided `.cwd` (`.tool_input.cwd`) for the Bash call (the field `suggest-mcp-reindex-after-pull.sh` already reads for the same reason) — falling back to the hook's own process cwd exactly as before when none of those hints are present, so a plain, non-worktree session sees zero behavior change.
- **`REPO_ROOT`, the `resolve_ops_root()` walk, and the `origin`/`upstream` remote lookups now all run through `git -C "$WORKDIR"`** instead of bare git, closing the inconsistency the issue's root-cause analysis called out between this hook and `require-active-ticket.sh`.

- **Also widens the hook's commit detection to recognise the `-C <path>` invocation form — which closes a second, previously unknown fail-open.** The old pattern only matched the bare form, so a commit made with an explicit `-C <path>` never reached the ref check at all: the hook exited 0 silently and **a fabricated `Closes #N` passed unverified**. That is exactly the invocation shape a worktree sub-agent uses -- the population this ticket is about -- so the gap and the bug shared a cause. Verified in review: the same `-C`-bound invocation carrying a fabricated ref exits 0 before this change and is blocked after it. The new pattern is a strict superset of the old one (17 shapes checked, no misses), and it also replaces the GNU-only `\s` with the portable `[[:space:]]`.
- **One round-2 test was passing vacuously and is now meaningful.** It asserted `rc=0` with empty stderr -- indistinguishable from a silently-skipped hook -- so it was confirming the early exit rather than the resolver. It only began testing the intended behaviour once the detection was widened.

## Bug reproduction

Verified against the unmodified hook before making any change: a sandbox shaped like the report (hook process cwd = an "ops fork" repo with `origin` = `me2resh/apexyard`; harness `.cwd` = a separate "project" repo with `origin` = `managed-org/example`; the referenced issue exists **only** in the project repo) reliably reproduced the false block — `BLOCKED: ... does not exist in me2resh/apexyard`, exit 2 — even though the issue is real and open in the correct tracker. Confirms the report; this is not a stale ticket.

## Testing

- New test file `.claude/hooks/tests/test_verify_commit_refs_cwd.sh`, 5 cases:
  1. payload `.cwd` differs from the hook's own cwd → must resolve against `.cwd`'s repo (the core bug)
  2. explicit `git -C <path>` in the command → must win over both cwd and `.cwd`
  3. `cd <path> && git commit ...` compound prefix → must resolve against that path
  4. no `.cwd` / `-C` / `cd`-prefix hints → falls back to the hook's own process cwd, **unchanged** from pre-fix behavior (no-regression case)
  5. repeat of case 1 with the `Refs` keyword, matching the report's exact false-block shape
- RED confirmed first: 4 of 5 cases failed against the unmodified hook (case 4 correctly passed, since no worktree mismatch applies there); GREEN after the fix, all 5 pass.
- `test_verify_commit_refs_upstream.sh` (existing upstream-awareness suite, 12 cases) still passes unchanged — no regression to the #207 upstream-fallback behavior.
- Full suite: `bash bin/run-hook-tests.sh` → **PASS=123 FAIL=0** (baseline was 122; +1 for the new test file).
- `shellcheck --severity=warning` clean on both changed files.

Testing instructions for reviewers:
```bash
bash .claude/hooks/tests/test_verify_commit_refs_cwd.sh
bash .claude/hooks/tests/test_verify_commit_refs_upstream.sh
bash bin/run-hook-tests.sh
```

Refs #1050

---

## Glossary

| Term | Definition |
|------|------------|
| Linked worktree | A second working directory attached to one git repo (`git worktree add`), where `--git-dir` and `--git-common-dir` differ from the main clone — the shape `/fan-out` and `Workflow` create for parallel sub-agents |
| `git -C <path>` | Runs git as if it had started in `<path>`, regardless of the shell's actual cwd — the fix for resolving a repo from the commit's own directory instead of the hook process's cwd |
| Payload `.cwd` | The harness-provided working directory for a given `Bash` tool call, carried in the PreToolUse/PostToolUse JSON payload — distinct from the hook script's own process cwd |
| Tracker repo | The GitHub repo whose issues a `Closes #N` / `Refs #N` reference is validated against |

### DIFF
--- a/.claude/hooks/tests/test_verify_commit_refs_cwd.sh
+++ b/.claude/hooks/tests/test_verify_commit_refs_cwd.sh
@@ -300,6 +300,13 @@ assert_case() {
 # ---- Case 7: trailing, unrelated `git -C <dir> log` after the real ----
 # ---- commit must not bind to it (round-2 hijack, row 3).             ----
 #
+# REGRESSION LOCK, not a discriminator: this shape happened to pass under
+# BOTH the round-1 and round-2 implementations (verified — round-1's
+# unscoped `tail -1` scrape didn't end up preferring the decoy here either),
+# so it doesn't prove the round-2 fix by itself. Kept anyway to pin the
+# behavior going forward, since the commit-boundary logic this depends on
+# has since changed again (round 3, token-boundary fix below).
+#
 # No payload .cwd. Hook cwd = "real" (has the issue). The command has a
 # real, separate `git -C <decoy> log` AFTER the actual commit invocation —
 # a `-C` that belongs to a DIFFERENT git invocation, not this one.
@@ -313,7 +320,7 @@ assert_case() {
   cmd=$(build_command 'fix: trailing unrelated -C\n\nCloses #507' "" "" " && git -C ${decoy} log")
   out=$(run_hook "$real" "" "$cmd")
   rc=$?
-  assert_case "a trailing, unrelated 'git -C <dir> log' does not bind to this commit" \
+  assert_case "[regression lock] a trailing, unrelated 'git -C <dir> log' does not bind to this commit" \
     "$out" "$rc" 0 ""
   rm -rf "$real" "$decoy"
 }
@@ -371,6 +378,86 @@ assert_case() {
   assert_case "a relative cd path is not trusted; falls back to the hook's own cwd" \
     "$out" "$rc" 0 ""
   rm -rf "$ops"
+}
+
+# ---- Cases 11-13: a directory whose NAME merely CONTAINS the substring ----
+# ---- "commit" must not truncate the commit-boundary cut (round-3       ----
+# ---- hijack — a real bug in the round-2 fix's own `${cmd%%commit*}`,   ----
+# ---- caught end-to-end by the reviewer before merge).                  ----
+#
+# `${cmd%%commit*}` cuts at the first literal SUBSTRING "commit", not a
+# token boundary — so a path component like "commitizen" or "commits-repo"
+# truncates mid-directory. The truncated parent is usually not a git repo,
+# so the (pre-fix) failure mode is FAIL-OPEN: TRACKER_REPO ends up empty,
+# the hook hits the "could not resolve tracker repo — skipping" branch, and
+# ref verification is silently disabled (same class as round-1's
+# message-prose hijack, just a different trigger).
+#
+# Discriminator design: the referenced issue is set to NOT EXIST in the
+# CORRECT (untruncated) repo's origin. If the fix correctly resolves the
+# full path, the hook queries that origin, finds the ref missing, and
+# BLOCKS (rc=2, "do not exist in <repo>"). If truncation still occurs, the
+# truncated parent has no origin, TRACKER_REPO is empty, and the hook exits
+# 0 via the skip branch instead — silently passing a fabricated ref. So
+# rc=2 (not rc=0) is the proof that real verification happened against the
+# right repo, and the stderr assertion additionally rules out the skip
+# message appearing at all.
+
+# Case 11: `cd <path>/commitizen && git commit` — reviewer's row 1 example.
+{
+  ops=$(make_repo "me2resh/apexyard")
+  commitizen=$(make_repo "managed-org/tool" "commitizen")
+  mock_gh_install "$commitizen"
+  mock_gh_set_repo_existence "$commitizen" 511 "managed-org/tool" no
+
+  cmd=$(build_command 'fix: cd into a dir named commitizen\n\nCloses #511' "cd ${commitizen} && ")
+  out=$(run_hook "$ops" "" "$cmd")
+  rc=$?
+  assert_case "cd into .../commitizen resolves the FULL path, not truncated at 'commit'" \
+    "$out" "$rc" 2 "do not exist"
+  if echo "$out" | grep -qF "could not resolve tracker repo"; then
+    echo "FAIL [cd .../commitizen]: silently skipped instead of verifying — output: ${out:0:300}" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}cd-commitizen-silent-skip "
+  fi
+  rm -rf "$ops" "$commitizen"
+}
+
+# Case 12: `git -C <path>/commitizen commit` — reviewer's row 2 example.
+{
+  ops=$(make_repo "me2resh/apexyard")
+  commitizen=$(make_repo "managed-org/tool2" "commitizen")
+  mock_gh_install "$commitizen"
+  mock_gh_set_repo_existence "$commitizen" 512 "managed-org/tool2" no
+
+  cmd=$(build_command 'fix: -C into a dir named commitizen\n\nCloses #512' "" "-C ${commitizen} ")
+  out=$(run_hook "$ops" "" "$cmd")
+  rc=$?
+  assert_case "git -C .../commitizen resolves the FULL path, not truncated at 'commit'" \
+    "$out" "$rc" 2 "do not exist"
+  if echo "$out" | grep -qF "could not resolve tracker repo"; then
+    echo "FAIL [-C .../commitizen]: silently skipped instead of verifying — output: ${out:0:300}" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}dash-C-commitizen-silent-skip "
+  fi
+  rm -rf "$ops" "$commitizen"
+}
+
+# Case 13: `cd <path>/commits-repo && git commit` — reviewer's row 3 example.
+{
+  ops=$(make_repo "me2resh/apexyard")
+  commitsrepo=$(make_repo "managed-org/tool3" "commits-repo")
+  mock_gh_install "$commitsrepo"
+  mock_gh_set_repo_existence "$commitsrepo" 513 "managed-org/tool3" no
+
+  cmd=$(build_command 'fix: cd into a dir named commits-repo\n\nCloses #513' "cd ${commitsrepo} && ")
+  out=$(run_hook "$ops" "" "$cmd")
+  rc=$?
+  assert_case "cd into .../commits-repo resolves the FULL path, not truncated at 'commit'" \
+    "$out" "$rc" 2 "do not exist"
+  if echo "$out" | grep -qF "could not resolve tracker repo"; then
+    echo "FAIL [cd .../commits-repo]: silently skipped instead of verifying — output: ${out:0:300}" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}cd-commits-repo-silent-skip "
+  fi
+  rm -rf "$ops" "$commitsrepo"
 }
 
 # ---- Summary ------------------------------------------------------------
--- a/.claude/hooks/tests/test_verify_commit_refs_cwd.sh
+++ b/.claude/hooks/tests/test_verify_commit_refs_cwd.sh
@@ -1,0 +1,217 @@
+#!/bin/bash
+# Tests for verify-commit-refs.sh cwd-vs-worktree resolution (me2resh/apexyard#1050)
+#
+# Bug: the hook resolved the tracker repo from `git rev-parse --show-toplevel`
+# / `git remote get-url origin` run with NO `-C`, which means both commands
+# used the HOOK PROCESS'S OWN cwd — not the directory the `git commit` is
+# actually executing in. In a split-portfolio / worktree-based sub-agent
+# session, the harness's Bash cwd for that call (the payload `.cwd` field,
+# same field `suggest-mcp-reindex-after-pull.sh` already reads) can be a
+# managed project's worktree while the hook process itself is invoked from
+# the ops fork. The hook then validates `Closes #N` against the OPS FORK's
+# tracker instead of the PROJECT's tracker, false-blocking a legitimate
+# reference.
+#
+# Coverage:
+#   - payload .cwd points at a different repo than the hook's own cwd →
+#     the referenced issue exists ONLY in the .cwd repo → must PASS
+#     (pre-fix: BLOCKS, because it checks the hook's own cwd's repo instead)
+#   - explicit `git -C <path> commit ...` in the command → must resolve
+#     against <path>'s repo, regardless of .cwd or the hook's own cwd
+#   - `cd <path> && git commit ...` compound prefix → must resolve against
+#     <path>'s repo
+#   - no .cwd, no -C, no cd-prefix → falls back to the hook's own process
+#     cwd, UNCHANGED from before the fix (no-regression case)
+#
+# Each case builds two isolated one-commit git sandboxes with different
+# `origin` remotes (simulating the ops fork vs. a managed project's
+# worktree), installs the shared mock `gh`, and pipes a synthetic
+# PreToolUse JSON blob at the hook while the hook's OWN process cwd is
+# deliberately the "wrong" sandbox.
+
+set -u
+
+HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/verify-commit-refs.sh"
+# shellcheck source=_lib-mock-gh.sh
+source "$(cd "$(dirname "$0")" && pwd)/_lib-mock-gh.sh"
+if [ ! -x "$HOOK_SRC" ]; then
+  echo "FAIL: hook not found or not executable at $HOOK_SRC" >&2
+  exit 1
+fi
+
+PASS=0
+FAIL=0
+FAILED_CASES=""
+
+# Build a one-commit git sandbox with the given origin remote. No upstream
+# remote (keeps the upstream-fallback branch out of play for these cases).
+make_repo() {
+  local origin_slug="$1"
+  local rp
+  rp=$(mktemp -d)
+  (
+    cd "$rp" || exit 1
+    git init -q
+    git config user.email "test@example.com"
+    git config user.name "test"
+    git remote add origin "git@github.com:${origin_slug}.git"
+    touch README.md
+    git add README.md
+    git commit -q -m "init"
+  )
+  echo "$rp"
+}
+
+# Build the `git commit -m "..."` command string. $1 uses literal `\n` for
+# newlines (for readability in the case table below); converted to REAL
+# newlines here, same as test_verify_commit_refs_upstream.sh does, so the
+# hook's REFS regex sees an actual word boundary before "Closes" / "Refs"
+# instead of a literal backslash-n glued directly onto the prior word.
+# $2 is an optional prefix prepended verbatim in front of `git commit`
+# (e.g. "git -C /path/to/repo " or "cd /path/to/repo && ").
+build_command() {
+  local msg_literal="$1" prefix="${2:-}"
+  local msg; msg=$(printf '%b' "$msg_literal")
+  printf '%sgit commit -m "%s"' "$prefix" "$msg"
+}
+
+# Run the hook with:
+#   - process cwd = $hook_cwd_repo (simulates the ops fork the hook is
+#     invoked from)
+#   - synthetic PreToolUse payload with the given .cwd (may be empty) and
+#     command
+# Prints combined captured stderr; rc is returned via $?.
+run_hook() {
+  local hook_cwd_repo="$1" payload_cwd="$2" command="$3"
+  local input result rc
+  if [ -n "$payload_cwd" ]; then
+    input=$(jq -nc --arg c "$command" --arg cwd "$payload_cwd" \
+      '{cwd:$cwd, tool_input:{command:$c}}')
+  else
+    input=$(jq -nc --arg c "$command" '{tool_input:{command:$c}}')
+  fi
+  result=$(cd "$hook_cwd_repo" && echo "$input" | bash "$HOOK_SRC" 2>&1 >/dev/null)
+  rc=$?
+  echo "$result"
+  return "$rc"
+}
+
+assert_case() {
+  local label="$1" got_output="$2" got_rc="$3" want_rc="$4" want_stderr_regex="$5"
+  if [ "$got_rc" != "$want_rc" ]; then
+    echo "FAIL [$label]: want rc=$want_rc, got $got_rc (stderr: ${got_output:0:400})" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  if [ -n "$want_stderr_regex" ] && ! echo "$got_output" | grep -qE "$want_stderr_regex"; then
+    echo "FAIL [$label]: stderr did not match /$want_stderr_regex/" >&2
+    echo "    stderr: $got_output" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "; return
+  fi
+  echo "PASS [$label]"
+  PASS=$((PASS+1))
+}
+
+# ---- Case 1: payload .cwd is the source of truth, not the hook's own cwd ----
+#
+# Hook process cwd  = "ops" repo (origin: me2resh/apexyard)      — issue 500 MISSING
+# Payload .cwd      = "proj" repo (origin: managed-org/example)  — issue 500 EXISTS
+# A `git commit` with no -C and no cd-prefix, run with harness cwd = proj,
+# must validate against proj's tracker (managed-org/example), not ops's.
+{
+  ops=$(make_repo "me2resh/apexyard")
+  proj=$(make_repo "managed-org/example")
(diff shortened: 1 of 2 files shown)


atlas-apex (commented):
## Code Review: PR #1073

**Commit**: `3286f6152fc266dc2f3e11b32c89a6d8fcdc459f`

### Summary

Fixes a real, independently-reported bug (#1050): `verify-commit-refs.sh` resolved `REPO_ROOT`, the `resolve_ops_root()` walk, and both remote lookups with bare git calls bound to the hook process's cwd rather than the commit's own directory, false-blocking valid refs in worktree sessions. The `git -C "$WORKDIR"` routing is correct, and the test discipline is genuinely good (RED-then-GREEN documented, no regressions).

The problem is the **resolution strategy**, not the routing. `resolve_commit_workdir()` text-matches the raw Bash command string — and that string *contains the commit message*. I reproduced a workdir hijack from commit-message prose, in both failure directions.

### Checklist Results
- Architecture & Design: **Fail** (Blocking 1-3)
- Code Quality: Pass
- Testing: Pass (independently verified below)
- Security: **Fail** — trust-chain file with a demonstrated fail-open
- Performance: Pass
- Technical Decisions (AgDR): Pass (see Suggestion 4)

---

### Issues Found

#### BLOCKING 1 — commit-message prose hijacks the workdir (this is #1066, in a new file)

`resolve_commit_workdir "$COMMAND"` receives the **whole raw command string**, which includes the `-m` commit message. Prose in the message is scraped as if it were shell syntax. Reproduced end-to-end against the built hook:

| # | Shape | Resolves to | Result |
|---|-------|-------------|--------|
| 1 | message body contains `git -C /tmp/rex1073/notarepo status` (existing, non-git dir) | that dir | **FAILS OPEN** — `WARN: could not resolve tracker repo. Skipping.`, **exit 0**. Ref verification silently disabled; a fabricated `#N` sails through. |
| 2 | message body contains `git -C /tmp/rex1073/otherrepo status` (existing repo, `origin=torvalds/linux`) | that repo | **FAILS CLOSED** — `BLOCKED: ... do not exist in torvalds/linux`, **exit 2**, on a valid `Refs #1050`. |
| 3 | message body has a line starting `cd /evil` | `/evil` | falls back only because `/evil` is not a directory — safe *by accident* |

Case 1 is the serious one: a trust-chain validator that silently stops validating, with only a stderr WARN. Case 2 is the same false-block class this PR set out to fix, merely re-sourced — and the error names the **wrong repo**, exactly the wrong-read hazard `.claude/rules/ticket-vocabulary.md` § "Repo targeting" calls out.

**This is not hypothetical for this repo.** This PR's own commit message contains ``git -C "$WORKDIR"``; the resolver scrapes it and returns the literal `"$WORKDIR"`. It escapes hijack only because that string is not a directory. Commit messages here routinely quote shell commands — the current safety is luck, not design.

This is the class [`AgDR-0104`](docs/agdr/AgDR-0104-trust-chain-controls-vs-backstops.md) argues cannot be made sound by pattern-matching command text, and it has produced three live bugs already (#1066 scraping prose, #1068 greedy binding, #1067's trim).

**Fix:** strip the extracted `MSG` from `COMMAND` before scraping. `MSG` is already computed above — scan only the residue.

#### BLOCKING 2 — `tail -1` (last-wins) binds an unrelated chained `git -C`

The `git -C` branch uses `tail -1`, so any `git -C` anywhere in a compound command wins, including one unrelated to the commit:

| Shape | Resolves to | Result |
|---|---|---|
| `git -C /other log -1 && git commit -m "...Refs #1050"` | `/other` | **FAILS CLOSED** (verified: exit 2, blocked against `torvalds/linux`) |
| `git commit -m "...Refs #1050" && git -C /other log -1` | `/other` | **FAILS CLOSED** — a `git -C` *after* the commit binds it |
| `cd /real && git -C /other status && git commit ...` | `/other` | wrong — the explicit `cd /real` that governs the commit is discarded |

The comment justifies `tail -1` as "matching how git itself would apply repeated `-C` flags" — but git does that for `-C` flags on **one invocation**. These are separate commands. The cited precedent `suggest-mcp-reindex-after-pull.sh` uses `head -1`; the divergence is neither flagged nor justified.

**Fix:** bind only a `git -C` on the `git commit` invocation itself, not anywhere in the string.

#### BLOCKING 3 — the priority order inverts the cited precedent, and that inversion is what makes 1 and 2 reachable

The code comment says the chain "mirrors `suggest-mcp-reindex-after-pull.sh`'s cwd-then-command-parsed fallback chain, **but ordered so an explicit, in-command override always wins over the ambient payload cwd**." That is an inversion, not a mirror. The precedent (lines 109-138) puts payload `.cwd` **first** and calls it "the common case"; command-parsing is fallback 2-4.

Structured beats scraped. Putting two text-parsed sources ahead of the structured field discards the harness's authoritative answer whenever the prose happens to contain matching text. With `.cwd` first, Blocking 1 and 2 become unreachable in the common case.

If the inversion is deliberate, the reason belongs in the code — the comment states the *what*, not the *why*. As written it reads as an unexplained divergence from a precedent it claims to follow.

---

1. **Quoted paths never resolve.** `cd "/my path/repo"` captures `"/my` (the char class stops at whitespace and keeps the quote); `git -C "/my path/repo"` likewise. Not a directory, so it silently falls back to the pre-fix buggy behaviour — the fix simply does not apply to any path containing a space.
2. **Relative paths are actively risky.** `cd ../sibling` captures `../sibling`, and both the `[ -d "$WORKDIR" ]` guard and every later `git -C` evaluate it against the **hook's** cwd — the very cwd this fix exists to stop trusting. It can resolve to a real but *different* sibling directory. `~/repo`, `cd -`, and `pushd` all fail to resolve and fall back harmlessly.
3. **Consider making the fail-open case loud.** Case 1 degrades to `WARN ... Skipping` + exit 0. For a trust-chain validator, a resolved-but-unusable workdir differs from never having resolved one; the former deserves more than a stderr line.
4. **AgDR:** I agree a bug fix needs none. But `resolve_commit_workdir()` is the *second* hook implementing this chain, and it diverges from the first on both ordering and `head`/`tail`. If deliberate, that is a cross-cutting pattern worth a short AgDR; if not, converge on the precedent. Not blocking on its own.

---

### Verification performed

- **New suite**: 5/5 pass at HEAD. Against the unmodified pre-fix hook (`HEAD~1`) — **exactly 4 fail, 1 passes**, and the passing one is case 4, the no-regression fallback case. The claimed split is honest.
- **Existing `test_verify_commit_refs_upstream.sh`** (#207): **12/12 pass**, unchanged. Confirmed, not taken on trust.
- **Test count**: `git ls-tree` diff between `HEAD~1` and `HEAD` shows exactly one addition (`test_verify_commit_refs_cwd.sh`) and **zero deletions**. The +1 delta holds in every counting scheme (121 to 122 for `test_*.sh` repo-wide; 106 to 107 under `.claude/hooks/tests/`). A new test *file* is the whole explanation; no test was lost.
- PR body, Glossary, narrative bullets, `Refs #1050`, and conventional commit format all pass.

### Verdict

**CHANGES REQUESTED**

The routing half of this fix is right and I want it to land. Every blocking finding is inside `resolve_commit_workdir()`, and the two highest-value remedies are small: **scan `COMMAND` with `MSG` removed**, and **put payload `.cwd` first**, per the precedent this PR already cites. Together those close Blocking 1 and 2 outright.

---
Reviewed commit: `3286f6152fc266dc2f3e11b32c89a6d8fcdc459f`

me2resh (approved):

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1073 (round 2)

**Commit**: `1b55f6523360913b19c88e295c568b0cd404bd1b` (was `3286f61`)

### Summary

Both round-1 blocking remedies are implemented correctly, and the third (the misleading precedent comment) is fixed properly rather than papered over. **All eight rows of my round-1 table are now closed.** I found one new shape introduced by this round's `-C`-binding fix; it fails open, which is why I'm holding rather than approving. It is small and the remedy is about one line.

### Round-1 findings: all closed

| Round-1 row | Status at `1b55f65` | Closed by |
|---|---|---|
| R1 message prose `-C <non-git dir>` (**fail-open**) | **CLOSED** | message stripped + `.cwd` first |
| R2 message prose `-C <git repo>` (fail-closed, wrong repo) | **CLOSED** | same |
| R3 message prose `cd /evil` | **CLOSED** | same |
| R4 unrelated `-C` **before** commit | **CLOSED** | `-C` bound to this `commit` token |
| R5 unrelated `-C` **after** commit | **CLOSED** | prefix excludes everything at/after `commit` |
| R6 `cd /real` + intervening unrelated `-C` | **CLOSED** | anchored `-C` no longer displaces the governing `cd` |
| R7 quoted spacey path silently reverting | **CLOSED** | quoted-path branches for both `-C` and `cd` |
| R8 relative `cd ../sibling` | **CLOSED** | absolute-only (see below) |

Verified at the function level and, where the environment allowed, end-to-end.

### The absolute-paths-only change — which direction it fails

You flagged this as a behaviour change I didn't ask for. **It is the right call, and it fails in the safe direction.**

A relative hint is discarded rather than resolved, so a legitimate `cd ../project && git commit …` (with `.cwd` absent) falls back to the hook's own process cwd — i.e. exactly the documented pre-fix behaviour. It does **not** invent a wrong answer; it declines an ambiguous one. The alternative is worse: `../sibling` can only be resolved against the hook's own cwd, which is *precisely the cwd this entire fix exists to distrust*, and `[ -d ]` would let it coincidentally pass while pointing at a different sibling. So the change trades a possible-silent-wrong-repo for a known-and-documented fallback. Right trade.

Residual, non-blocking: this is silent. A relative hint that *was* the only signal available produces no diagnostic. Worth a one-line stderr note under the same reasoning as Suggestion 3 below.

### On the subtractive-stripping concern (and why #1075's framing does not transfer)

I applied the security reviewer's #1075 lens and it **does not carry over here**, for a specific and load-bearing reason:

**`REFS` is extracted from `MSG` *before* and *independently of* the strip.** The gate's question — "which refs does this message claim, and do they exist?" — is computed from `MSG` at the top of the file. `strip_message_from_command()` runs much later and feeds only the *workdir hint* search. So stripping **cannot make the gate ask no question**, which is the exact failure mode that made #1075's subtractive pre-filter dangerous. This is a subtractive narrowing of an *additive extractor's* input, not of the gate's input.

I probed the remaining risk anyway: for the strip to remove text the resolver needed, the governing `cd`/`-C` would have to fall inside the removed span — but the hint necessarily precedes the `git commit` token, which precedes the `-m` body, and `index()` removes the *first* occurrence. I could not construct a case where a governing hint is destroyed. Additionally `MSG` must contain a ref keyword to reach this code at all, which rules out the short-string collision (`msg="x"` splicing a path) that would otherwise be the obvious attack.

The `awk -v` reasoning also holds: `-v` assignment applies escape processing to its value, so a message containing backslash sequences would be mangled. Piping both values through stdin with a sentinel and slicing with `index()`/`substr()` avoids regex and escape interpretation on both sides. Sound.

---

### BLOCKING — new shape introduced by this round: `${cmd%%commit*}` truncates on any path containing `commit`

The same-invocation scoping is implemented as `prefix="${cmd%%commit*}"` — a cut at the **first literal substring** `commit`, not at a token boundary. Any directory whose path contains that substring truncates mid-path:

| Command (`.cwd` absent) | Resolves to | Correct |
|---|---|---|
| `cd /tmp/rex1073/repos/commitizen && git commit -m X` | `/tmp/rex1073/repos/` | `/tmp/rex1073/repos/commitizen` |
| `git -C /tmp/rex1073/repos/commitizen commit -m X` | `/tmp/rex1073/repos/` | same |
| `cd /tmp/rex1073/commits-repo && git commit -m X` | `/tmp/rex1073/` | `/tmp/rex1073/commits-repo` |

**Direction: FAILS OPEN.** The truncated parent is usually not a repo, so the hook emits `WARN: verify-commit-refs.sh could not resolve tracker repo. Skipping.` and **exits 0** — ref verification silently disabled, a fabricated `#N` passes. Verified end-to-end. If the truncated parent *is* a repo, it fails closed against the wrong one instead.

This is not exotic: `pre-commit`, `pre-commit-hooks`, `commitizen`, and `commitlint` are all common path components, and `pre-commit` in particular is near-ubiquitous.

**Scope, stated fairly:** reachable only when `.cwd` is **absent**, since `.cwd` now wins outright. That is a genuinely narrow window and strictly better than round-1, where the equivalent bugs outranked `.cwd`. But it is a *newly introduced* fail-open in a trust-chain validator — one that did not exist before this PR — so per `right-size-ceremony.md` rail 1 (trust chain never goes lean) and rail 2 (ambiguity rounds up), I'm holding on it.

**Remedy (~1 line):** cut on a token boundary rather than a substring — require the `commit` to be preceded by whitespace and followed by whitespace or end-of-string, e.g. match `[[:space:]]commit([[:space:]]|$)` instead of `%%commit*`.

---

1. **The trailing-`-C` test passes both ways.** "a trailing, unrelated `git -C <dir> log` does not bind" is GREEN against round-1 as well, so it documents the fix rather than proving it. My round-1 function-level probe showed that shape genuinely resolving to `/other`, so the behaviour *is* fixed — the test just isn't the thing demonstrating it. Consider a case that goes RED.
2. **`.cwd` absence is the whole risk surface now.** Everything remaining depends on it. A one-line comment recording whether `.cwd` is guaranteed present for `PreToolUse` Bash payloads (and if so, that 2/3 are defence-in-depth) would let the next reader size this correctly.
3. **Make the resolve-failure path louder.** The `WARN … Skipping` + exit 0 branch is the fail-open outlet for every resolution defect, present and future. Distinguishing "no hint at all" from "resolved a path that isn't a repo" would have surfaced the truncation above immediately.
4. **Deferring #1075 convergence: agree.** Sourcing an in-review helper would couple two open PRs and put this fix behind that review. Noting it as follow-up is right. I'd add that convergence should be *selective*: #1075's `strip_heredoc_bodies()` is subtractive on gate input, which is what made it fail open across eight shapes. This hook's stripping is safe specifically because `REFS` is computed first — that property must not be lost in any shared helper.

---

### Verification performed

- **New suite: 11/11 PASS** at `1b55f65`.
- **RED claim independently confirmed, and not a stash artefact.** I reconstructed the round-1 tree with `git archive` (no stash anywhere) and ran the round-2 tests against the round-1 hook: **exactly 3 FAIL / 8 PASS** — message-content hijack, intervening-unrelated-`-C`, quoted-spacey-path. Matches the author's claim precisely.
- **Existing `test_verify_commit_refs_upstream.sh`: 12/12 PASS**, unchanged.
- **Environment note for anyone re-running:** in this fork a bare baseline commit blocks because the tracker lib cannot see `me2resh/apexyard` issues from `atlas-apex/apexyard`. I confirmed this is **not** a regression — round-1 (`3286f61`) and round-2 (`1b55f65`) behave *identically* on the same payload. Not attributable to this delta.

### Verdict

**CHANGES REQUESTED** — one blocking finding, narrowly scoped.

The round-1 work is fully and thoughtfully addressed; the `.cwd`-first reordering plus message-stripping closes all eight original rows, and the comment correction is honest about what the earlier version got wrong. I'm holding only on the `%%commit*` substring cut, because it is a new fail-open in a trust-chain validator and the fix is a token-boundary match. With that one change I expect to approve.

---
Reviewed commit: `1b55f6523360913b19c88e295c568b0cd404bd1b`

atlas-apex (commented):
## Code Review: PR #1073 (round 3)

**Commit**: `54af7d139f6851569ab0ffa9579a4f4e71405555` (was `1b55f65`)

### Summary

My round-2 blocking finding is fixed and the truncation class is **fully closed**. The observability ask landed in a more useful form than I asked for. The widened `git commit` detection gate — the change made beyond what was asked — is, after examination, **a net security improvement that closes a second fail-open I had not found**, and it is a strict superset of the old gate with no misses.

**APPROVED**, with one firm pre-merge ask on the PR body (below) — it is a documentation gap, not a code defect.

---

### The widened commit-detection gate — my judgement

This was the right change, and the author's reasoning checks out. I verified it three ways.

**1. It is a strict superset — no shape is lost.** I ran old vs new across 17 command shapes. Every shape the old `\bgit\s+commit\b` fired on, the new pattern also fires on. The optional `-C` group is skipped when absent, so the previous behaviour is preserved exactly. Bonus: `\s` (a GNU extension, unreliable in BSD/POSIX ERE) is replaced with the portable `[[:space:]]`.

**2. The priority level was genuinely dead — proven, not assumed.** The decisive test is a `-C`-bound commit carrying a **fabricated** ref:

| Hook | `git -C /tmp/.../otherrepo commit -m "…Closes #999999"` |
|---|---|
| round-2 `1b55f65` | **exit 0, silent** — gate never fired, ref verification skipped entirely |
| round-3 `54af7d1` | **exit 2, BLOCKED** |

So this is not merely "dead code made reachable". **Before this round, a commit made via `git -C <path> commit` bypassed ref verification altogether** — a fabricated `#N` sailed through, silently. That is a fail-open in the same hook, and it is precisely the shape a worktree-based sub-agent uses, which is the whole population #1050 is about. Widening the gate *closes* it.

**Corollary the author should know:** round-2's Case 2 test ("explicit `git -C <path>` resolves against that path's repo") asserted `rc=0` with empty stderr — which is exactly what a silently-skipped hook produces. **It passed vacuously.** It was asserting the gate's early exit, not the resolver. That test only became meaningful this round. (My round-2 rows R4/R5/R6 remain correctly closed — those shapes contain an adjacent `git commit`, so they always reached the resolver.)

**3. New false positives — both fail safe.** Two shapes newly fire:

| Shape | Fires now | Direction |
|---|---|---|
| `git -C /x commit-graph write` | new | **safe no-op** — no `-m`/`-F`, so `MSG` is empty and the hook exits 0 before any check |
| `git -C /x commit-tree abc -m X` | new | **safe / consistent** — `commit-tree` genuinely creates a commit object, and bare `git commit-tree` already fired pre-PR via `commit\b`; the `-C` variant merely restores symmetry |

Neither can produce a false block on a non-commit. I also confirmed the gate does **not** fire on `git -C /x log`, `git -C /x status`, `git -C /x checkout commit`, `git -C /x branch commit`, `git -C commit log`, or `commit` appearing as a later word (`… && echo commit`).

Verdict on the change: **correct, and it should have been in scope.** A resolver priority level that cannot be reached is not a style issue — here it was masking a live bypass.

---

### Truncation class: fully closed

The token-boundary cut (`[[:space:]]commit([[:space:]]|$)` via `awk match()` on a space-padded string) resolves every shape I broke in round 2:

| Shape (`.cwd` absent) | Round 2 | Round 3 |
|---|---|---|
| `cd /…/commitizen && git commit` | `/…/repos/` (**fail-open**) | `/…/repos/commitizen` |
| `git -C /…/commitizen commit` | `/…/repos/` | `/…/repos/commitizen` |
| `cd /…/commits-repo && git commit` | `/…/` | `/…/commits-repo` |
| `cd /x/pre-commit-hooks && git commit` | truncated | `/x/pre-commit-hooks` |
| `cd /x/commit && git commit` (path *ends* in the token) | — | `/x/commit` — correct |
| `git -C "/my path/commitizen" commit` | — | `/my path/commitizen` — quoted + spacey + substring, correct |

Round-2 no-regression shapes still hold: `cd /real && git -C /other status && git commit` resolves `/real`; `git commit … && git -C /other log` resolves nothing (falls back).

One residual, non-blocking: a directory literally containing the token surrounded by spaces — `cd "/my commit dir"` — truncates the prefix, the quoted-`cd` branch then fails to close its quote, and the bare branch yields a non-absolute candidate that the absolute-only rule rejects. Result is **empty → fallback**, i.e. a safe degradation to documented pre-fix behaviour, never a wrong directory. Correct failure direction for an exotic input.

### Both invariants survived

- **`REFS` before strip:** `MSG` extracted line 103, `REFS` computed from it line 125; `strip_message_from_command` defined 186, called 285. `REFS` is computed 160 lines earlier and is untouched by stripping. The additive-extractor property that keeps this hook out of #1075's failure class **holds**.
- **`.cwd` first:** `PAYLOAD_CWD` line 284, assigned to `WORKDIR` line 296, scraped fallback gated behind `[ -z "$WORKDIR" ]` line 301. **Holds.**

### `WORKDIR_SOURCE` — better than what I asked for

I asked only that the resolve-failure path distinguish "no hint" from "resolved a non-repo". The `_unresolvable:<dir>` suffix goes further by surfacing the *offending directory* in the warning, and the message names the scraper as the likely culprit. That is the diagnostic that would have made my round-2 truncation finding self-reporting the first time it fired. Good.

### Verification performed

- **Suites: 14/14** (cwd) and **12/12** (upstream, unchanged).
- **RED confirmed independently, no stash:** rebuilt the round-2 tree with `git archive` and ran the round-3 tests against the round-2 hook — **5 failures across the 3 new truncation cases**, with the exact signature I reported (`rc=0` + `WARN … Skipping`). Precision note: the `-C /…/commitizen` case is RED because of the *gate*, not the truncation (empty stderr, silent skip) — so it proves the gate fix; the other two prove the truncation fix. Both are genuine.
- **Test accounting:** file list at `1b55f65` and `54af7d1` is **identical** — the 3 new cases are additions inside the existing file, zero deletions. A suite total that counts test *files* correctly stays at 123.

### Pre-merge ask (non-blocking on the code)

**The PR body does not mention the widened commit-detection gate.** I grepped it: no reference to the gate, to the dead priority level, or to the fact that `git -C <path> commit` previously skipped ref verification entirely. The body still describes only the cwd-resolution fix.

That under-sells the change and hides a user-visible security fix. Per `.claude/rules/pr-quality.md` § "Summary bullets — narrative quality", please add a bullet along the lines of:

> **Widened the `git commit` detection gate to recognise `git -C <path> commit`.** The previous `\bgit\s+commit\b` check required the two tokens to be adjacent, so a commit made via explicit `-C` — the shape a worktree sub-agent uses — skipped ref verification **entirely**, letting a fabricated `#N` through silently. Closing this also made the resolver's `-C` priority level reachable for the first time.

A matching Glossary row for the gate would help too. I'm not blocking on this because it changes no code and the fix itself is verified — but it should land before merge so the release notes reflect both bugs.

### Suggestions (non-blocking, future)

1. **Pre-existing gate gap, not introduced here:** `git -c user.name=x commit -m …` is skipped by **both** old and new patterns. `-c <k=v>` is a common override shape and this is the same fail-open class just closed for `-C`. Worth a follow-up ticket. (Related curiosity: `git --git-dir=/x/.git commit` *does* fire, but only by accident — the literal `.git commit` inside the path matches. Not principled.)
2. **#1075 convergence:** agree it stays a follow-up; no cross-PR dependency here. Now that `AgDR-0113` states the additive-only constraint and names `resolve_commit_workdir()` explicitly, that AgDR is the right place to also record the *why* — this hook is safe specifically because `REFS` is computed before the strip, and any shared helper must preserve that ordering.
3. **AgDR for this PR:** still not needed. Bug fix, no new dependency, no cross-cutting pattern. The gate widening is a correction to an existing check, not a new one.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass — closes two fail-opens (truncation + `-C` gate bypass)
- Performance: Pass
- PR Description & Glossary: **Advisory** — see pre-merge ask
- Summary Bullet Narrative: **Advisory** — same
- Adopter Handbooks: N/A

### Verdict

**APPROVED**

Three rounds, and each one closed the finding it was given without breaking the one before it. The truncation class is fully closed, both structural invariants survived, and the widened gate — the change I was asked to scrutinise hardest — turns out to close a live fail-open bypass that none of the three review rounds had found. Please add the body bullet before merging.

---
Reviewed commit: `54af7d139f6851569ab0ffa9579a4f4e71405555`

Issue #148, opened 2026-05-03
[Docs] Correct privacy-gate wording — adopter action, not framework auto-publish


atlas-apex:
## Driver

The privacy-gate wording introduced in PR #144 (and unchanged in PR #147) attributes the fork-publication to the framework rather than the adopter:

> the standard fork-and-commit setup **will silently publish** your private project names on a public GitHub repo

This is factually wrong. ApexYard never pushes anything without explicit operator approval — the publication happens only when the adopter themselves runs `git push`, intentionally or accidentally. The current wording reads like the framework auto-publishes, which is misleading and undermines trust in the rest of the framework's safety claims.

## Scope

Two prose-only edits, no code, no behavior change.

### 1. `.claude/skills/setup/SKILL.md` — Step 2a privacy-gate question

Replace `will silently publish your private project names on a public GitHub repo` with adopter-action language: `you might accidentally publish your private project names on a public GitHub repo (a stray git push after registering them)`.

### 2. `docs/multi-project.md` — trip-wire callout

Replace `silently publish their portfolio names the moment they push` with adopter-action language: `risk accidentally publishing their portfolio names with a stray push`.

## Acceptance Criteria

- [ ] `.claude/skills/setup/SKILL.md` Step 2a no longer contains the phrase "silently publish"; replaced with adopter-action language ("you might accidentally publish ...")
- [ ] `docs/multi-project.md` trip-wire callout no longer contains "silently publish"; replaced with adopter-action language ("risk accidentally publishing ...")
- [ ] No code changes — prose only
- [ ] grep returns zero hits across the framework: `grep -r "silently publish" .claude/skills/ docs/`

## Risks / Dependencies

None. Pure prose correction. Targets `dev` per release-cut model.

atlas-apex:
Closed by PR #149 (merged 2026-05-03). Cross-repo notation in PR body didn't trigger same-repo auto-close — closing manually.

PR #1074, opened 2026-07-30
fix(#1056): resolve issue refs before making Closes actually auto-close

atlas-apex:
## Summary

- **Fixed silent auto-close failure in release changelogs** — the v5.3.0 release's `Closes` list named 18 references but auto-closed zero issues. Two independent defects combined to cause this, and I fixed them in a specific order because fixing them in the wrong order makes things worse, not better.
- **Ref resolution fixed first** — when a commit subject has no `type(#N):` scope (e.g. `docs: rework the onboarding flow copy (#2001)`), the generator was emitting the trailing **PR number** as if it were the issue. `resolve_issue_from_pr()` now looks up that PR's own body for a `Closes`/`Fixes`/`Resolves`/`Refs #N` reference and uses the real issue instead — falling back to the PR number if nothing resolves.
- **Emission fixed second, only after ref resolution was correct** — `### Closes` used to emit a single `Closes #A, #B, #C` line, and GitHub only auto-closes the reference immediately following the keyword, so everything past `#A` was silently inert. It now emits one `- Closes #N` bullet per reference so all of them fire.
- **Why this order, explicitly** — today, a wrong ref from the resolution bug is harmless *only because* the comma-list bug stops anything past the first reference from firing at all. If I had fixed emission first — repeating the keyword per reference — while ref resolution was still wrong, every mis-extracted trailing PR number would become a live `Closes #N` directive. That number is somebody's real PR, and if it happens to also be the number of an open issue in this repo, the next release cut would auto-close that unrelated issue, silently, the same day it merged. Fixing ref resolution first closes off that path entirely: by the time emission repeats the keyword, every number in the list is already the right one.
- **Fail-soft behaviour of `resolve_issue_from_pr()`, stated plainly** — it does one `gh pr view <PR> --repo me2resh/apexyard --json body -q .body` call per unscoped commit. On success, it greps the PR body for a closing-keyword reference. On ANY failure — `gh` not on `PATH`, auth failure, rate limit, or the forge being unreachable (I hit this exact failure live, mid-merge, earlier this session — `api.github.com` briefly stopped resolving) — it prints nothing and the caller falls back to the raw PR number, i.e. today's pre-#1056 behaviour. This is not a hypothetical claim: `test_release_changelog.sh` has a dedicated case ("fail-soft: gh failure degrades to PR number, script still exits 0") that stubs `gh` to exit 1 unconditionally and asserts both that the PR number still appears in `Closes` and that the script itself exits `0`. That test passes.

## What changed

- `bin/release-changelog.sh`:
  - Added `has_scope_issue()` — detects whether a commit subject's `(#N)` is a genuine issue-holding scope (`fix(#1042): ...`) vs. absent.
  - Added `resolve_issue_from_pr()` — best-effort `gh pr view <PR> --json body` lookup for a closing-keyword reference in the PR's own body. Every internal failure path is neutralized (`if ! cmd; then`, explicit `return 0`) so a `gh` failure can never trip `set -e` and abort the script.
  - Added `PR_LOOKUP_REPO` env var (default `me2resh/apexyard`) controlling which repo the lookup targets.
  - Reworked the `Closes` section emission from a single comma-joined line to one `- Closes #N` bullet per (deduplicated, sorted) reference.
- `.claude/skills/release/SKILL.md`: updated the documented `### Closes` output shape and the count-mismatch guard's comment to match the new one-bullet-per-reference format (the guard's actual `grep -v` logic already handled multiple matching lines correctly — only the comment was stale).
- `.claude/hooks/tests/test_release_changelog.sh`: 8 new assertions across 5 new test cases — a ≥3-ref comma-list regression, a scoped subject (proves no PR lookup fires for the already-correct case), an unscoped subject resolved via a stubbed PR body, an unscoped subject whose PR resolves to nothing, and the fail-soft/unreachable-forge case described above. Added a `make_stub_gh()` helper that writes a fake `gh` binary controlled by `STUB_GH_BODY` / `STUB_GH_EXIT` env vars, so these tests need no real network access.

## Testing

1. **Red before, green after, in fix order.** Wrote the 8 new assertions first and confirmed 6 failures against the unfixed script (4 for the comma-list bug, 2 for the ref-resolution bug). Implemented ref resolution (`has_scope_issue` / `resolve_issue_from_pr`) and reran — those 2 failures cleared while the 4 emission failures remained, confirming (a) was fixed independently of (b). Then implemented the emission rework and reran — all 8 new assertions passed alongside the 63 pre-existing ones.
2. `bash .claude/hooks/tests/test_release_changelog.sh` — **71/71 pass** (63 pre-existing + 8 new).
3. `bash bin/run-hook-tests.sh` — full framework hook suite, run twice independently after the fix landed: **`hook test suite: PASS=122 FAIL=0 SKIP(quarantined)=0 TOTAL=122`** both times, matching today's baseline exactly. `test_release_changelog.sh` and `test_release_sync.sh` both show `PASS` in the full run.
4. `bash .claude/hooks/tests/test_release_sync.sh` — **14/14 pass**, unaffected (doesn't touch the `Closes` format).
5. `shellcheck --severity=warning bin/release-changelog.sh .claude/hooks/tests/test_release_changelog.sh` — clean, no warnings.

## Risk notes

- The only new network call this script makes is the best-effort PR-body lookup, and only for unscoped commit subjects (scoped commits — the common case — never touch the network). Every failure mode of that call degrades to the pre-existing behavior, verified by the fail-soft test above, not merely asserted.
- Both fixes land in this one commit, not as separate PRs, deliberately: shipping the emission fix without the ref-resolution fix — even for the length of one PR review cycle — would strictly increase the auto-close blast radius, per the ordering argument above.

Refs #1056

## Glossary

| Term | Definition |
|------|------------|
| Closing keyword | `Closes` / `Fixes` / `Resolves` — GitHub auto-closes the issue immediately following one, and only that one, when the PR merges into the default branch |
| Conventional-commit scope | The `(#N)` in `fix(#1042): …`, which by apexyard convention holds the **issue** number |
| Trailing PR ref | The `(#NNN)` GitHub appends on squash-merge, which is the **PR** number, not an issue |
| Fail-soft | Degrading to a safe default (here: the raw PR number) instead of erroring out, when a best-effort step (here: the `gh` lookup) can't complete |
| Release-cut model | `dev` takes daily PRs; `main` receives only squash-merged release PRs, so no individual `dev` commit is ever an ancestor of a release tag |

### DIFF
--- a/.claude/hooks/tests/test_release_changelog.sh
+++ b/.claude/hooks/tests/test_release_changelog.sh
@@ -392,6 +392,121 @@ contains "range line present on stderr" "RELEASE_CHANGELOG_RANGE=" "$stderr_out"
 expect_sha=$(echo "$out" | grep -oE 'EXPECT_SHA=.*' | cut -d= -f2)
 contains "stderr range is anchored on the trailer sha" "RELEASE_CHANGELOG_RANGE=${expect_sha}..HEAD" "$stderr_out"
 
+# ── #1056 helper — a stub `gh` binary for ref-resolution tests ──────────────
+# Writes a fake `gh` to <dir>/gh, controlled by two env vars the caller sets
+# before invoking the changelog script: STUB_GH_BODY (the PR body text `gh pr
+# view --json body -q .body` should "return") and STUB_GH_EXIT (non-zero to
+# simulate a `gh` failure — auth, rate limit, or an unreachable forge, the
+# exact failure mode hit live during the v5.3.0 cut). Every `$` in the heredoc
+# body is backslash-escaped so it survives heredoc creation literally and only
+# expands when the stub itself runs later, inside the test's own subshell.
+make_stub_gh() {  # make_stub_gh <dir>
+  mkdir -p "$1"
+  cat > "$1/gh" <<STUBEOF
+#!/usr/bin/env bash
+if [ "\${STUB_GH_EXIT:-0}" != "0" ]; then
+  exit "\${STUB_GH_EXIT}"
+fi
+if [ "\$1" = "pr" ] && [ "\$2" = "view" ]; then
+  printf "%s" "\${STUB_GH_BODY:-}"
+  exit 0
+fi
+exit 1
+STUBEOF
+  chmod +x "$1/gh"
+}
+
+# ── Test: #1056 defect 1 — Closes repeats the keyword per reference ─────────
+# A comma-joined "Closes #A, #B, #C" line only auto-closes #A — GitHub only
+# honours the reference immediately following the closing keyword. Every
+# reference must get its own "- Closes #N" bullet so all of them fire.
+echo "--- #1056 Closes repeats keyword per reference (>=3 refs) ---"
+out=$(run_test '
+  mc "chore: initial"
+  git tag v15.0.0
+  mc "feat(#1501): first unreleased feature"
+  mc "fix(#1502): second unreleased fix"
+  mc "chore(#1503): third unreleased chore"
+  PREV_TAG="v15.0.0" HEAD_REF="HEAD" VERSION="v15.1.0" DATE="2026-07-29" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains   "Closes #1501 own bullet" "Closes #1501" "$out"
+contains   "Closes #1502 own bullet" "Closes #1502" "$out"
+contains   "Closes #1503 own bullet" "Closes #1503" "$out"
+not_contains "no comma-joined Closes line (#1501, #1502 inert-past-first bug)" "Closes #1501, #1502" "$out"
+not_contains "no comma-joined Closes line (any pairing)" "#1502, #1503" "$out"
+
+# ── Test: #1056 defect 2 — scoped subject is unaffected (no PR lookup) ──────
+# `fix(#1042): ...` — the scope IS the issue number by convention. This must
+# resolve straight from the subject with no `gh` call at all (no stub `gh` is
+# put on PATH for this test — if the script tried to shell out, it would hit
+# whatever real `gh` is on the runner's PATH and could flake or hang).
+echo "--- #1056 scoped subject resolves directly, no PR lookup ---"
+out=$(run_test '
+  mc "chore: initial"
+  git tag v16.0.0
+  mc "fix(#1600): put the human gate on approval flow (#1601)"
+  PREV_TAG="v16.0.0" HEAD_REF="HEAD" VERSION="v16.1.0" DATE="2026-07-29" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains     "closes the ISSUE number from the scope" "Closes #1600" "$out"
+not_contains "does not close the trailing PR number instead" "Closes #1601" "$out"
+
+# ── Test: #1056 defect 2 — unscoped subject resolves via the PR's own body ──
+# `docs: ... (#2001)` has no `type(#N):` scope — the only "(#N)" present is
+# the trailing squash-merge PR number, not an issue. The generator must look
+# up PR #2001's body, find its "Refs #2000", and close #2000 — not #2001.
+echo "--- #1056 unscoped subject resolves issue from PR body (Refs #N) ---"
+out=$(run_test '
+  mc "chore: initial"
+  git tag v17.0.0
+  mc "docs: rework the onboarding flow copy (#2001)"
+  fakebin="$PWD/fakebin"
+  make_stub_gh "$fakebin"
+  PATH="$fakebin:$PATH" STUB_GH_BODY="See also. Refs #2000" \
+    PREV_TAG="v17.0.0" HEAD_REF="HEAD" VERSION="v17.1.0" DATE="2026-07-29" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains     "resolves to the referenced issue" "Closes #2000" "$out"
+not_contains "does not close the PR number itself" "Closes #2001" "$out"
+contains     "display line still shows the PR number" "(#2001)" "$out"
+
+# ── Test: #1056 defect 2 — unscoped subject whose PR resolves to nothing ────
+# The PR body carries no Closes/Fixes/Resolves/Refs reference at all — the
+# generator must degrade to the PR number (today'"'"'s behaviour), not drop the
+# entry or crash.
+echo "--- #1056 unscoped subject, PR body has no closing reference (falls back to PR number) ---"
+out=$(run_test '
+  mc "chore: initial"
+  git tag v18.0.0
+  mc "docs: fix a typo in the README (#2101)"
+  fakebin="$PWD/fakebin"
+  make_stub_gh "$fakebin"
+  PATH="$fakebin:$PATH" STUB_GH_BODY="Just a typo fix, no issue linked." \
+    PREV_TAG="v18.0.0" HEAD_REF="HEAD" VERSION="v18.1.0" DATE="2026-07-29" \
+    bash "'"$CHANGELOG_SCRIPT"'" 2>&1
+')
+contains "falls back to the PR number when nothing resolves" "Closes #2101" "$out"
+
+# ── Test: #1056 fail-soft — an unreachable forge never aborts the release ───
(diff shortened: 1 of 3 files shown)


atlas-apex (commented):
## Code Review: PR #1074

**Commit**: `739ea927fc8f9e985196335ff218a2d6f9b76b22`

### Summary

Fixes the release changelog's `Closes` section on two axes, in the right order: (a) resolve an unscoped commit's trailing PR number back to the issue it references, then (b) emit one `- Closes #N` bullet per reference instead of a comma list GitHub only honours the first element of. Reviewed with the explicit premise that (b) makes (a)'s residual mis-resolutions **live** — a previously-inert wrong reference now fires a real auto-close at release time.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass (verified RED against `dev`, GREEN on HEAD)
- Security: Pass
- Performance: Pass (see suggestion 5 — serial network calls, no timeout)
- Technical Decisions (AgDR): N/A — agree with the author's read
- Adopter Handbooks: N/A (no findings)

### Verification performed

All claims checked by execution, not by reading:

- **Fail-soft, 4 modes + absent binary.** Built a scratch repo with adversarial subjects (via plumbing, to bypass the format hook) and ran the real script against a stubbed `gh` in modes `exit 1`, **empty stdout with rc=0**, **malformed JSON**, and **`gh` absent from PATH entirely**. All four produce byte-identical output and `rc=0`. The script never aborts.
- **RED-on-`dev`.** Ran HEAD's test file against `dev`'s script in a fake tree: exactly **6 failures, 65 pass — split 4 emission / 2 resolution**, as claimed. The ordering claim holds: the 2 resolution assertions are single-reference cases whose expected string (`Closes #2000`) renders identically under the old comma emission, so fixing (a) alone clears them and leaves the 4. HEAD: 71/71. `test_release_sync.sh`: 14/14.
- **Emission actually auto-closes.** `me2resh/apexyard`'s default branch is `main`; a release PR squash-merges `dev` into it, and each bullet is a keyword + bare same-repo `#N` on its own line. This form fires. (#207's cross-repo lesson is respected for well-formed input — but see F2.)
- **Real-history frequency scan** over the last 800 commits on `upstream/dev` to separate latent shapes from live ones.

### Issues Found

Ordered by whether they emit a **WRONG** close (reaches into an unrelated issue, silent at release time) or merely a **MISSING** one (annoyance).

The root cause of every wrong-close below is shared: `extract_pr_num()` takes `grep -oE '#[0-9]+' | head -1` — **the first `#N` anywhere in the subject**, with no notion of position or context. `has_scope_issue()` then only classifies `^[a-z]+\(#[0-9]+\)!?:`, so anything else falls into the resolve-or-fallback path carrying that first token.

| # | Subject shape | Emits | Verdict |
|---|---|---|---|
| F1 | `docs: ... (cross-cutting design for #347 + #348 + #351) (#352)` | `Closes #347` | **WRONG** + missing #348/#351. A design-doc commit closes the epic it only designed. **This shape exists in real history.** |
| F2 | `fix: mirror other-repo#12 behaviour (#1046)` | `Closes #12` (bare, same-repo) | **WRONG** — the `owner/repo` prefix is stripped by the grep, so a foreign-repo reference becomes a live local close. Latent. |
| F3 | `Revert "fix(#1042): broken thing"` | `Closes #1042` | **WRONG** — a revert re-closes the issue it just undid. Latent (`Revert` is capitalised, so `has_scope_issue`'s `^[a-z]+` misses it). |
| F4 | `docs(me2resh/apexyard#148): ... (#149)` | `Closes #148` | Correct **by accident**. `has_scope_issue` does not match the `type(owner/repo#N):` form, so it is misrouted to `gh pr view 148`; that fails *because* 148 is an issue, and the fallback happens to print the right number. It becomes **WRONG** the moment that number is a PR in the repo (issue and PR numbering is shared) — `gh pr view` then succeeds and returns an unrelated issue. **6 occurrences in the last 800 commits.** |
| F5 | `fix: handle #1050 and #1051 (#1068)` | `Closes #1050` | **WRONG-ish** + missing #1051 and the PR's real issue. |
| F6 | PR body whose *first* closing-keyword match is prose or a quoted example | that prose ref | **WRONG** — `grep ... \| head -1` scans the whole body with no anchoring; a framework PR that *discusses* `Fixes #N` resolves to the discussed number. |
| — | `fix(#1042): x` (scoped) | `Closes #1042` | Correct — no lookup, as designed. |
| — | unscoped, PR closes nothing / gh degraded | `Closes #<PR number>` | Benign no-op — a closing keyword on a PR number closes no issue. |
| — | `Merge pull request #1067 from x` | resolved issue, or `#1067` | Correct / benign. |

**Which way does it fail?** Open — into the first `#N` in the subject. In this repo's convention that token is usually the issue, so the degrade is usually benign *here*: the history scan found 12 mismatch subjects and the currently-present ones all degrade to a correct number. But F2/F3/F6 degrade straight into a live wrong close, and the fallback is reached **most often precisely when the extracted number is an issue rather than a PR** (that is exactly when `gh pr view` 404s) — so the safety of the fallback is accidental, not structural.

**Why this is still APPROVED rather than blocked.** Three things carry it: (1) the status quo is a release that closes essentially nothing, so this is strictly better in aggregate; (2) the dangerous shapes (F2, F3, F6) are **latent** — none appear in the last 800 commits, and F4's six real occurrences land correctly today; (3) the `Closes` bullets are rendered into the release PR body, which a human reads and can edit **before** merge — a real mitigating control on an otherwise silent failure. F1 is the one live shape and its emitted number is at least topically related to the commit.

### Suggestions (non-blocking, ordered by payoff)

1. **Widen `has_scope_issue()` to accept `type(owner/repo#N):`** — a one-line regex change that removes F4's accidental-correctness entirely and stops six-in-eight-hundred real commits from taking a network round-trip they do not need. Highest value in this list.
2. **Stop hardcoding `PR_LOOKUP_REPO=me2resh/apexyard`.** Derive it from `REPO_REMOTE`'s URL (or `gh repo view --json nameWithOwner`). A fork cutting its own release currently resolves *its* PR numbers against *upstream's* numbering and emits an unrelated upstream issue number as a live local close — a systematic wrong close for every fork, silent at release time. The env override and header comment mitigate it, but the default is the wrong way round.
3. **Refuse to emit a token that was preceded by `owner/repo`** (closes F2) and **skip subjects matching `^Revert`** (closes F3). Both are cheap guards at the extraction site.
4. **Anchor the PR-body grep** — prefer a match on a line that *starts* with the keyword (a trailer) over `head -1` across the whole body (F6).
5. **Bound the lookup.** There is no timeout: I confirmed a hung `gh` runs until an *external* `timeout` kills it (rc=124 at 12s). 77 of the last 800 commits would each make a serial `gh` call, so an unresponsive API stalls a release cut indefinitely with no progress output. A `timeout 10` wrapper (degrading if the binary is absent) would keep the existing fail-soft semantics.

None of these change the verdict; 1 and 2 are worth a follow-up ticket.

### Priority 4 — `Released-From` trailer / #1002

No perturbation. `RELEASE_CHANGELOG_RANGE` and the `IFS`-join fix are already on `dev` and are **not** part of this PR (the true diff is 3 files: the script, its tests, and `release/SKILL.md`); this change does not touch range derivation. The count-mismatch guard's filter `grep -vcE '^- Closes '` is a prefix match, so N bullets are excluded exactly as 1 was — `ENTRY_COUNT` and `GAP` are unchanged, and the SKILL.md comment update says so accurately. Verified against file content at the ref, not via ancestry checks.

`docs/release-process.md` already documents the `#114` single-`Closes`-per-PR skip marker for release PRs; the multi-bullet form does not change that, since the marker bypasses the check outright.

### AgDR

Agree with the author: **not needed**. This repairs existing logic against a stated convention rather than choosing an approach; it adds no dependency, service, schema, or security control, and is reversible inside this PR. Under `.claude/rules/agdr-decisions.md` § "The threshold" it is a routine implementation fix, and rail 1 (security / trust-chain / migration) does not apply to release-note generation.

### PR hygiene

Summary bullets are narrative (what changed *and* why), `Refs #1056` present and correct, Glossary present, commit subject `fix(#1056): ...` conforms. The risk-notes section is a good addition for a change of this shape.

### Verdict
**APPROVED**

---
Reviewed commit: `739ea927fc8f9e985196335ff218a2d6f9b76b22`

me2resh (approved):

PR #1075, opened 2026-07-30
fix(#1066): strip heredoc bodies before git-command pattern matching

atlas-apex:
## Summary

- **Heredoc prose was winning over the real git command.** `_lib-extract-push-ref.sh`'s `extract_push_ref`/`is_tag_push` (shared by `validate-branch-name.sh` and `block-main-push.sh`) scanned the *whole* raw Bash command string for the first `git push`/`git commit`/`--tags` match. A heredoc body — a commit message, a PR body, review prose — is literal payload, not a command, but this repo's own commit messages routinely *discuss* git behaviour ("`git push` still runs the checks..."). That text matched before the real invocation and got validated as if it were the branch name, blocking a correctly-named push with a message pointing at a branch that doesn't exist. Confirmed twice in one session per the ticket: `validate-branch-name.sh` blocked a legitimate push because the commit message discussed `git push`, and `block-main-push.sh` blocked a Bash call that only *wrote* review prose containing a quoted `git commit -m` inside a heredoc (no real commit at all).

- **The first fix (this PR's first commit) was itself unsafe, and a security review caught it.** That version routed both hooks' *presence checks* — "is there a real `git push`/`git commit` here at all" — through heredoc-stripped text. Eight distinct malformed-heredoc shapes (an unterminated heredoc, a backslash/multi-word/dotted delimiter, `<<EOF` merely sitting in a quoted string, two heredocs on one line, a C++ `<<` operator in a quoted message) made that first stripper eat the real command, so the presence check silently never fired and the push genuinely ran — verified by stubbing `git` on `PATH` and confirming execution. This is now fixed (this PR's second commit); see the design section below for the shape of the fix, and `docs/agdr/AgDR-0113-heredoc-stripper-additive-only.md` for the full record, including the governance rule the security review established.

- **The governing principle, going forward:** `extract_push_ref` is an *additive* extractor (a wrong answer still has a real fallback to check against — failure is local and visible). `strip_heredoc_bodies` is a *subtractive pre-filter on gate input* (a bug there means no question gets asked at all — failure is silent and global). The fix is built entirely around keeping those two uses separate:
  - Both hooks' presence checks (`grep -qE 'git push'` / `'git commit'`) and their `cd`-target extraction for worktree resolution now run against the **raw, unfiltered command**, always — never against stripped text.
[...]

Issue #1046, opened 2026-07-27
[Bug] Leak hook: fail-closed message advises a non-working bypass; a literal double-dash token in body text truncates the scan


atlas-apex:
Two small defects in `block-private-refs-in-public-repos.sh`, both surfaced during the review of PR #1040 and both deliberately left out of that PR to keep a security fix narrow.

## Affected

`.claude/hooks/block-private-refs-in-public-repos.sh`

### 1. The fail-closed message advises a bypass that does not work

**Given** the `BODY_FILE_UNREADABLE` block added by #1039 runs *before* the skip-marker check
**When** it fires, its message tells the operator: *"To reference a registered project deliberately, add this marker to the body: `<!-- private-refs: allow -->`"*
**Then** adding that marker does **not** clear the block, because the marker check is downstream — expected that the advice either work, or not be offered

Safe direction (it over-blocks rather than leaking), but the guidance sends the operator down a path that cannot succeed. Either move the skip-marker check above the unreadable-file block, or reword the message to say the path itself must be fixed.

### 2. A literal ` --x ` inside body text truncates the scan

**Given** `extract_flag_value`'s anchor terminates a quoted value at ` --<letter>`
**When** a body legitimately contains a double-dash token in prose — documenting a CLI flag, or quoting a diff
**Then** the value truncates there and any private reference after it goes unscanned

This is **pre-existing** and behaves identically on `dev` and on PR #1040's HEAD — not introduced by that PR. Same *class* as #227 (embedded content terminating the match early), triggered by a flag-shaped token instead of a quote.

## Repro

1. Build a fixture registry with a private project.
2. Item 1: `gh issue create --repo <public> --title t --body-file /nonexistent.md` → blocked with the misleading advice. Adding the marker to any body does not clear it.
3. Item 2: `gh issue create --repo <public> --title t --body "Docs mention --verbose here. Seen during the <private-project> rebuild."` → the tail is not scanned.

## Why these are filed rather than fixed inline

The obvious fix for item 2 is to adjust the anchor — and that is precisely the change tried on PR #1040 and rejected, because widening it reopened #227 across every boundary character (measurably worse than no fix: 9 failures versus 5). Any fix here must preserve greedy content consumption while excluding one specific token shape, which needs its own reasoning and its own adversarial tests rather than being folded into a narrow security patch.

Item 1 is cosmetic in effect but sits in a security hook's operator-facing message, where wrong guidance costs real time.

## Framework version

v5.2.0

## Severity

[...]

PR #1080, opened 2026-07-30
fix(#1068): make leak-hook target-repo resolution class-based, not positional

atlas-apex:
## Summary

- **Round 1** anchored `TARGET_REPO` resolution on the leftmost recognised write invocation (`gh issue|pr create|comment`, `gh api`) and took the FIRST `--repo` value after it, replacing the original greedy `sed 's/.*--repo.../'` that bound the LAST `--repo` on the line. That closed the reported bypass (a chained second `gh` call, or a trailing comment, mentioning `--repo` after the real write).
- **Round 2 (this update).** Security review and code review independently converged on the same finding: round 1 only moved the fail-open window rather than closing it. "First after anchor" is still positional — it commits to picking ONE occurrence and trusting it's the real one — and a `--repo`-shaped token embedded inside an EARLIER flag's quoted value (`--title` or `--body`) now wins instead of the real flag, because it appears first in raw text. Verified against the actual `upstream/dev` blob at the commit this branch forked from (not reasoned from theory): both
  ```
  gh issue create --title "use --repo acme-private/x here" --repo <public> --body "leak"
  gh issue create --body "run with --repo other/x for leak" --repo <public> --title "T"
  ```
  **block on `dev`** (its last-wins happens to land on the real, later flag for this exact shape) and **leaked on round 1** (exit 0, nothing scanned). A third shape — two REAL writes chained, private repo first and public repo second carrying the leak — also leaked on round 1 and blocks on `dev`. The `--body` variant needs no adversarial crafting at all: `--body` routinely precedes `--repo`, and body prose quoting a `--repo` example is this issue's own habit (#1068's own body does it) — this is the ordinary shape of an upstream bug report about `gh`, not a crafted exploit.
- **Fixed by making resolution class-based instead of positional.** Within the write segment (the substring starting at the leftmost recognised write invocation), the hook now checks whether **any** known public-class repo is named as a `--repo` value or a `repos/<owner>/<repo>` path, **anywhere** in that segment — not picking one occurrence by position. A false YES only ever leads to more scanning, never less, which is the property both the round-1 fix and `dev`'s original behaviour lacked.
- **Deliberate tradeoff, stated explicitly, not discovered by the test:** a single Bash call chaining a private-repo write with a public-repo write now scans the private write's title/body too, because the public repo's slug is present somewhere in the combined segment. Test 27 flips from asserting a no-op to asserting a block — this is a **correction to the test**, because round 1 had encoded the fail-open behaviour as a requirement, not a regression this change introduces.
- **Round 1's truncation-refusal fix is untouched.** Both reviews independently tried to defeat the extraction stage and could not — every truncation shape either refused or blocked, with no regressions in #227/#1039/#1046. Its comment is strengthened with the stronger argument from code review: it detects the **failure**, not the **attack** — both quoted branches in `extract_flag_value` strip their own opening delimiter before returning, so a leading raw quote is structurally unreachable through a successful match. That's an invariant of the extractor, not a pattern guess about command text, so it does not inherit AgDR-0104's unsoundness. The hook also deliberately refuses to *name* a specific leaked project when it hits this branch — fabricating a "project name: X" from content it never finished reading would be indistinguishable from a real detection in an audit trail.
- **Round 3.** Security review found one more HIGH, upstream of everything in round 2: `find_write_segment`'s anchor (`(^|[[:space:]])gh`) carried a code comment claiming it "mirrors" step 1's classification regex (`\bgh`, a GNU grep `\b` word boundary). It doesn't — `\b` fires wherever `gh`'s leading `g` is preceded by ANY non-word character (`;`, `&`, `|`, `(`, `$`, not only whitespace), so step 1 and the segment-finder anchor disagreed on exactly those four shapes. When they disagreed, step 1 correctly classified the command as a write, but the segment finder returned empty and the hook exited 0 at the empty-segment check — **before round 2's class-based check, and before round 1/2's truncation refusal, ever ran.** Fixed by replacing the anchor with `(^|[^A-Za-z0-9_.-])gh`, a hand-rolled word boundary verified against all four divergent inputs plus plain `gh` and the `high api` false-positive guard. Folded in `-R` (gh's documented short flag for `--repo`, on the same footing as this hook's existing `-t`/`-b` short-form handling) in the same pass, since it was the same one-line alternation edit.
- **Round 3 self-correction (this update, still comment-only, no behaviour change).** The round-3 claim that all four anchor-boundary shapes "block on `dev`" was itself wrong for one of them, caught the same way the earlier mistakes were: by actually running it rather than reasoning from the pattern. `;gh`, `&&gh`, `|gh` genuinely block on `dev` — real regressions this branch closes. `out=$(gh issue create ...)` (a single, unchained write, the shape that matters most operationally since it's the routine idiom for capturing a created issue/PR's URL) **also allows on `dev`** — pre-existing there too, not a round-2 regression, because `dev` has no truncation-detection concept at all: the trailing `)` defeats its extraction the same way it defeats this hook's, `dev` just never notices and refuses. This branch still closes it — as a byproduct of the segment no longer being empty, letting the existing refusal mechanism run — which is an *improvement* over `dev` on this shape, not parity restoration. The earlier check that produced the wrong claim used a body where the leak token was the first word, which let `dev`'s truncated read catch it by accident and mask the real gap — re-verified with a non-frontloaded body to get the honest answer.

## Corrections (stated plainly, per AgDR-0113)

- **Round 3's original claim was itself a second instance of the pattern round 1 introduced**: an unverified "these two things behave the same" claim, plausible and cheap to check, that turned out false for one of four cases — corrected above, same turn it was found.
- The round-1 code comment claiming the quoted-value-embedding shape was "a pre-existing weakness of the previous implementation too" was **wrong**, not just unproven — verified against the real `upstream/dev` blob, which blocks that shape. Round 1 introduced the hole; it did not inherit it.
- The round-1 PR description's "52 pre-existing" test count didn't reconcile (55 total − 7 new ≠ 52); the correct baseline is **48 pre-existing**, with the 7 round-1 additions splitting into 3 net fail→pass (cases 21–23) and 4 already-green (24 control, 25, 26, 27) — not "two of seven already green" as previously stated.
- The round-1 residue note only covered the "decoy after the real flag" direction; it's now corrected to also cover the "decoy before" direction verified in this round, with a citation against the real `dev` blob for each claim.

## Residue (verified against the real `upstream/dev` blob, not assumed)

1. **Two real writes chained where the leak sits in an EARLIER write and a LATER write's `--repo` names a different (non-public) repo** can still go unscanned — the hook (before and after every round of this fix) extracts only one title/body across the whole command, not per-invocation, so a segment-wide "public repo present" YES doesn't guarantee the extracted title/body belongs to the same invocation that produced the YES. Confirmed against the real `dev` blob with a synthetic registry: this exact shape returns `rc=0` on `dev` too — genuinely pre-existing. Closing it needs per-invocation title/body extraction, materially larger than this bug fix's scope.
2. **`--repo=owner/name` (the equals form)** is not recognised by either check above and resolves nothing. Verified against `dev`: same gap there.
3. **Four further gaps**, each spot-checked against the real `dev` blob and confirmed to allow there too (`dev`, round 2, and round 3 alike — none introduced or widened by this PR): a repo slug in a different **case** (`ME2RESH/APEXYARD`), a slug **immediately followed by a shell operator** with no separating whitespace (`--repo owner/repo&&gh ...`), a `gh api` REST path carrying a **`?query`** suffix, and the `--repo=` equals form above (#2). Not fixed here — one follow-up ticket to cover all four, rather than four separate ones.

Documented and tested where a test case exists (case 31 for the equals form); the rest are documented in the source comment, not test-covered — flagged for the follow-up ticket rather than expanded here, per instruction to keep this round comment-only.

## Testing

`test_block_private_refs.sh`: **65/65** (48 pre-existing + 7 round-1 cases [3 net fail→pass: 21–23; 4 already-green: 24 control, 25, 26, 27] + 4 round-2 cases [28–31] + 6 round-3 cases [32–37]).

RED confirmed via the actual test harness at each round, not ad hoc calls:
- Round 1: cases 21–23 failed against the unpatched (pre-#1068) hook; case 24 (control) already passed, proving 21–23 are bypasses, not match failures.
- Round 2: cases 27 (flipped), 28, 29, 30 all failed when the test harness's `HOOK` path was pointed at the round-1 committed hook blob and re-run — confirmed failing before this round's fix, passing after.
- Round 3: cases 32–37 (the four anchor-boundary shapes, the truncation-refusal defeat, and `-R`) all failed when `HOOK` was pointed at the round-2 committed hook blob and re-run — confirmed failing before this round's fix, passing after.

`bash bin/run-hook-tests.sh`: full suite green — reporting the literal runner output below, not a rounded or assumed number (see PR comment for the exact `PASS=/FAIL=/TOTAL=` line from this round's run).

`shellcheck --severity=warning`: clean on both changed files, at every round.

## Glossary

| Term | Definition |
|------|------------|
| write segment | The substring of the command starting at the leftmost recognised write invocation (`gh issue create`, `gh api`, etc.); text before this point is out of scope for candidate `--repo` values |
| class-based resolution | Deciding "is a public repo this write's target?" by checking membership against the known public-repo list, rather than picking one `--repo` occurrence by position (first or last) |
| positional resolution | The round-1 (and original `dev`) approach: commit to one specific `--repo` occurrence by its position in the text, and trust it's the real one |
| decoy | A `--repo`-shaped (or `repos/<owner>/<repo>`-shaped) string that appears inside a flag's free-text value rather than as the actual flag for this write |
| fail closed | On ambiguity, refuse or over-scan rather than allow — the safe default for a protective gate |

Refs #1068


atlas-apex (commented):
## Security Review: PR #1080

**Commit**: `0d19eb00941ef3a217d7e1038d25adf72288afee`

### Summary

Target-repo resolution in `block-private-refs-in-public-repos.sh` moves from a greedy "last `--repo` on the line wins" `sed` to an awk function anchored on the leftmost recognised write invocation, taking the **first** `--repo` after that anchor. A follow-on change detects `extract_flag_value`'s one-word-truncation signature (a returned value beginning with a raw quote) and refuses rather than scanning a knowingly-truncated haystack.

The reported bug is genuinely fixed, and the extraction half is solid work that I could not defeat. But the resolution change is a **swap**, not a narrowing: it trades the "`--repo` *after* the write" hole for a "`--repo` *before* the write" hole. Two shapes that `dev` **blocks** now exit 0 with a private token reaching a public tracker unscanned. The PR text asserts these are pre-existing; measured against `dev`, they are not.

Tested against a synthetic registry (fake project `zebrafish` / `acme-private/zebrafish`), every shape run against both this HEAD and the `dev` baseline.

### Checklist Results

- XSS Prevention:         N/A
- Auth & Authorisation:   N/A
- Data Protection:        **Fail** — private-ref leak reachable at exit 0 (regression vs `dev`)
- API Security:           N/A
- Gate & Trust-Chain Integrity: **Fail** — a leak gate fails open on shapes it previously caught

### Security Issues Found

#### HIGH — resolution regressed from last-wins to first-wins; a `--repo` inside an earlier quoted value now silently disables the whole scan

`find_target_repo` anchors correctly on the write invocation, but then takes the first `--repo` **textually** after that anchor. Flag *values* live inside that region. So a `--repo <non-public>` appearing inside an earlier `--title`/`--body` value binds before the real flag, the target resolves non-public, and step 3 exits 0 having scanned nothing.

Instrumented resolution (`RESOLVED=[...]`), identical input to both versions:

| Shape | `dev` resolves | this HEAD resolves | `dev` | this HEAD |
|---|---|---|---|---|
| `--title "use --repo acme-private/zebrafish here" --repo <public> --body "zebrafish leak"` | `me2resh/apexyard` | `acme-private/zebrafish` | BLOCK | **exit 0, leak** |
| `--body "run with --repo other/x for zebrafish" --repo <public> --title "T"` | `me2resh/apexyard` | `other/x` | BLOCK | **exit 0, leak** |
| `--repo <public> ... --body-file <leaky> # see --repo other/x` | `other/x` | `me2resh/apexyard` | exit 0, leak | BLOCK (fixed) |

This is not a contrived shape in *this* repo. The hook's own block message embeds `--repo ${TARGET_REPO}`, and upstream tickets about `gh` behaviour routinely quote commands. An agent filing "repro: run with `--repo acme-private/zebrafish`" against the public tracker hits precisely this path — and the body carrying the private name is exactly what the gate exists to stop. The exploit condition is narrow but self-referential: the quoted repo must be non-public, which is the same condition as the leak being real.

The code comment states this residue "is a pre-existing weakness of the previous implementation too — not something this fix introduces." The table above contradicts that: `dev`'s last-wins happened to bind the real trailing `--repo` in these shapes and blocked them. Please correct the claim — an operator reading it would not know to guard a live hole.

**Direction: dangerous** (silent no-scan on a public target).

#### MEDIUM — same first-wins swap on two chained real writes, private-then-public

A private-repo comment chained ahead of a public-repo create, with the public write using `--body-file`, resolves the private repo and exits 0; `dev` resolved the public one and blocked. The inline-`--body` variant of this shape leaks on **both** versions (title/body extraction has always read the first occurrence), so that half is fairly characterised as pre-existing — but the `--body-file` variant is a regression. **Direction: dangerous.**

#### LOW — `--repo=owner/name` unresolvable on both versions

`--repo=me2resh/apexyard` (valid pflag syntax) matches neither the awk `--repo[[:space:]]+` patterns nor the old `sed`, so the target is unresolvable and the hook exits 0. Genuinely pre-existing and unchanged by this PR; worth a follow-up, not a blocker here.

#### No test covers the regressed direction

All 55 cases pass. Cases 21-27 exercise only "chained/commented **after**", plus one private-target symmetry check. Nothing covers a `--repo` preceding the real flag, which is why the swap went unnoticed. Any fix should land with cases in both orderings.

### What holds up (verified, no regression)

The extraction half is well done and resisted every probe:

| Probe | Result |
|---|---|
| Quoted body + chained command / trailing comment / apostrophe-in-body | REFUSE (truncation) |
| Command-substitution value followed by a chain | REFUSE |
| Leak after an embedded quote, chained after | REFUSE |
| Unquoted body token + chain; ANSI-C quoted value | BLOCK (value complete, not truncated) |
| #227 — embedded quote, leak in back half | BLOCK |
| #1046 — embedded quote + later `--flag` (leftmost-longest trim) | BLOCK |
| #1039 — unreadable `--body-file` | BLOCK (fail-closed; marker correctly no longer offered) |
| Lone pipe / semicolon in legitimate prose | exit 0 (terminator set **not** widened — the #1040 trap avoided) |
| Skip marker in `--label` | BLOCK (strict/greedy asymmetry intact) |
| Skip marker in body | bypass + warning (correct) |

I specifically tried to produce a *truncated-but-undetected* body and could not. The greedy quoted branches always retain a superset (`trim_to_last` cuts at the last qualifying quote, so retention can only over-capture), and the only path that under-captures is the unquoted fallback, which on a quote-opened value always returns a value starting with a raw quote — the detected signature. The refusal appears genuinely reachable in every truncation case I could construct.

### Priority 3 — is refusing the right posture?

Yes. It mirrors `BODY_FILE_UNREADABLE`: when extraction cannot establish where a value ends, the honest move is to decline rather than scan a haystack known to be short — "over-block, never silently scan nothing." Refusing to *name* a project is also correct, since the hook has not reliably read one.

It is not a denial-of-service on ordinary writes. A plain single write passes cleanly; the refusal fires only when a quoted value is followed by something that is neither a flag nor end-of-command — a chained command or trailing comment. The remedy in the message ("run the write as its own, unchained Bash call") is one edit and is the shape the framework wants anyway. The cost is bounded and falls entirely in the safe direction.

### Dangerous-vs-safe split

- **Dangerous (silent no-scan on a public target):** the HIGH and MEDIUM findings — both from the first-wins resolution swap.
- **Safe (over-block):** every truncation refusal, the strict-marker asymmetry, the unreadable body-file block.

The residue is *not* overwhelmingly the safe kind, which is the bar this review was asked to hold. The extraction work meets that bar; the resolution work does not yet.

### Recommendation

Make resolution **class-based rather than positional**. Collect every `--repo` value and every `repos/<owner>/<repo>` path after the anchor; if **any** is public-class, treat the target as public and scan. Both ordering holes close at once, and the failure mode becomes over-blocking — the direction Priority 2 explicitly prefers.

One tradeoff to make deliberately: this flips existing case 27 (private target, public repo mentioned in a later chained call) from no-op to block. That is an over-block on a command that mentions a public framework repo while writing privately — cheap, visible, and fixable by unchaining. I would take it, but it is the maintainer's call, so please decide it explicitly rather than inheriting it.

Whatever the approach, please also (a) correct the "pre-existing" claim in the code comment, and (b) add ordering cases in **both** directions.

### Verdict

**CHANGES REQUESTED**

No `*-security.approved` marker was written — two shapes place a private token on a public repo at exit 0, and both are regressions against `dev`.

---
📌 Reviewed commit: `0d19eb00941ef3a217d7e1038d25adf72288afee`

atlas-apex (commented):
## Code Review: PR #1080

**Commit**: `0d19eb00941ef3a217d7e1038d25adf72288afee`

### Summary

Anchors `block-private-refs-in-public-repos.sh`'s target-repo resolution on the leftmost recognised write invocation and takes the *first* `--repo` after it, replacing a greedy `sed` that bound the *last* `--repo` anywhere on the line. Also adds a fail-closed refusal when `extract_flag_value` returns a value starting with a raw quote (the truncation signature of a shell command chained after a quoted value).

I verified the reported bug and the fix empirically against both the pre-fix hook (`0e4e865^`) and this HEAD, using a synthetic registry. The reported bypass is real and is genuinely closed. **However, the fix introduces a mirror-image bypass of the same class and the same severity, and the PR's residue section states the opposite of what the code does.** That is the blocking finding.

### Checklist Results

- Architecture & Design: **Fail** — see blocking finding
- Code Quality: Pass — comments are load-bearing, the awk is readable, `shellcheck` clean
- Testing: Pass (with a caveat) — 55/55, red/green split verified independently; new bypass class untested
- Security: **Fail** — a previously-blocked leak shape now passes silently
- Performance: Pass
- PR Description & Glossary: Pass (Glossary present, `Refs #1068` present)
- Summary Bullet Narrative: Pass — all three bullets carry what changed *and* why
- Technical Decisions (AgDR): Advisory — see note

---

### Issues Found

#### BLOCKING 1 — the fix swaps a trailing-decoy bypass for a leading-decoy bypass of equal severity

Resolution now takes the **first** `--repo` after the anchor. The anchor is the leftmost recognised write shape, which in every realistic command is at position 0 — so "first `--repo` after the anchor" is effectively "first `--repo` on the line". Any `--repo` token appearing textually **before** the real flag now wins, including one sitting inside an earlier `--title`/`--body` quoted value.

Measured, pre-fix hook vs this HEAD (synthetic registry: project `zebrafish` maps to `acme/zebrafish-app`; public-class `me2resh/apexyard`):

| Shape | Pre-fix | This HEAD |
|---|---|---|
| Decoy `--repo octo/sandbox` inside `--title`, real `--repo me2resh/apexyard` after, leak in body | `exit 2` blocked, named the leak | **`exit 0` silent** |
| Decoy `--repo octo/sandbox` inside `--body` prose, real `--repo` after | `exit 2` blocked | **`exit 0` silent** |
| Two real writes chained, private target first, public target second carrying the leak | `exit 2` blocked | **`exit 0` silent** |

All three exit at step 3 having scanned nothing — the *exact* failure mode #1068 was filed about, mirrored from "decoy trails the real flag" to "decoy leads it".

The `--body` variant is the one that worries me most, because it is not adversarial. `--body` routinely precedes `--repo` in real invocations, and body prose that quotes a `--repo` command example is this repo's own dogfooding pattern — issue #1068's own body does it. An operator filing an upstream ticket that quotes such an example would silently disable the gate on the write carrying their prose. No quoting trickery, no chaining, no intent.

So the net effect is a lateral move, not a strict improvement: #1068's shape is closed, an equally reachable shape is opened.

**Suggested fix, consistent with the posture this file already states.** Don't pick *a* target — collect **all** candidates after the anchor (every `--repo` value plus every REST-path repo reference) and treat the write as public-class if **any** candidate is public-class. That is fail-closed in both directions and matches the hook's own stated rule ("over-block, never silently scan nothing"), rather than being fail-open in whichever direction the decoy happens to sit. It closes #1068's shape, the three shapes above, and the two-real-writes case in one move, without touching `extract_flag_value` or going near the #1040/#227 terminator trap.

The cost is that test case 27 (private target, public repo mentioned in a later chained call, currently asserted `exit 0`) would become an over-block. That is the safe direction, it has the documented skip-marker escape hatch, and case 27 as written currently *encodes the fail-open choice as a requirement* — worth revisiting on its own merits.

#### BLOCKING 2 — the residue section asserts the opposite of the measured behaviour

Residue #2 in the PR body and in the hook's comment block says a `--repo` inside an earlier flag's quoted value "is a pre-existing weakness of the previous implementation too — not something this fix introduces."

That is not correct. The previous implementation bound the **last** `--repo`, which for this shape is the real one, so it resolved correctly and blocked. Measured: pre-fix `exit 2` (named the leak), this HEAD `exit 0` (silent). The old code handled this shape correctly — by accident, but correctly — and the new code does not. The same applies to residue #1 for the private-first ordering.

Why this matters more than a wording nit: the residue section is the artefact a future reader will use to decide whether a reported bypass is old news or a regression. Right now it would tell them "pre-existing, don't worry" about something this PR introduced. Given this file's history, that is the sentence most likely to cost someone a debugging session. It needs to say plainly: *these two orderings blocked before this change and are silent after it.*

Also, "requires deliberately crafting the command against its own safety hook" doesn't hold for the `--body`-prose variant above, which needs no craft at all.

**Whichever way BLOCKING 1 is resolved, these shapes need test cases** — otherwise the next change to this function has nothing pinning them.

---

### On the refuse-on-truncation-signature design — I think this call is right

You asked me to judge this specifically, so: I would ship this half as-is.

**Rejecting the terminator widening was correct**, and for the right reason rather than a superficially similar one. Any terminator general enough to mean "a new shell command starts here" is exactly as good at amputating a body that contains that same text as prose. The `#1039 r2` cases pin precisely that, and I verified them: a lone `|`, `;`, `&` or ` -t` inside legitimate body prose still does **not** terminate the value, and the leak in the tail is still found. #1040's measured 9-vs-5 regression is the empirical version of the same argument. Widening was the trap; not taking it was the judgement worth making.

**Signature-detection is sound in the under-detect direction**, which is the direction that matters. The claim it rests on — a value can begin with a raw `"` or `'` *only* via the unquoted fallback firing on a quote-opened value it could not close — holds structurally: both quoted branches strip their opening delimiter before returning, so a leading raw quote is unreachable through them. It is a genuine invariant of the extractor, not a pattern guess about command text, which makes it a different and better kind of check than the anchor itself. It detects the *failure*, not the *attack*, so it does not inherit AgDR-0104's unsoundness.

**Over-refusal risk is acceptable.** A false positive needs a title or body whose extracted value legitimately begins with a quote character, which only arises from the same fallback path. The remedy printed is one an operator can act on ("run the write as its own, unchained Bash call") and is correct advice regardless.

**And cases 21-23 blocking on "could not safely determine where the value ends" rather than naming a project is the right shape** — probably the best judgement call in the PR. The hook does not in fact know a project was leaked in those commands; it knows it could not read the body. Fabricating "project name: X" would be a false claim of a specific find, would teach operators to distrust the message when the named project isn't there, and would be indistinguishable from a real detection in an audit trail. An honest refusal that says what it does not know beats a confident answer it cannot support. It also mirrors `BODY_FILE_UNREADABLE`, so the file now has one consistent story for "I could not read it" — and the related correction to the `BODY_FILE_UNREADABLE` text (removing the wrong suggestion that the skip marker applies to unread bytes) is a genuine improvement for the same reason.

One note: this half is *load-bearing for the other half's safety story*. Cases 21-23 block only because truncation-refusal catches them after resolution succeeds. If BLOCKING 1 is fixed by unioning candidates, that stays true. Worth keeping the two mechanisms explicitly coupled in the comments, because removing the refusal later would silently reopen 21-23.

---

### Regression verification (independent, not read from the claim)

- `test_block_private_refs.sh` on this HEAD: **55 passed, 0 failed**. Confirmed.
- **Red/green split verified against the pre-fix hook**, not asserted: cases 21, 22, 23 gave `exit 0`, silent, nothing scanned (genuine bypasses). Case 24 control gave `exit 2` on both, which is what makes 21-23 bypasses rather than match failures. Cases 25, 26 gave `exit 2` on both. Case 27 gave `exit 0` on both. So **3 net fail-to-pass, 4 already green**.
- `#1039 r2` boundary cases specifically confirmed: a lone `|`, `;`, `&` in body prose is still not a terminator. No #227/#1039/#1046 regression — every pre-existing case still passes.
- Nothing deleted: base `83249c2` had 40 `run_case` calls, HEAD has 47. Strictly additive.

### Count corrections (non-blocking, but the numbers should be right)

- **"52 pre-existing + 3 new net-fail-to-pass" does not reconcile with 55.** Base has 40 `run_case`, HEAD has 47, so 7 new. 55 total assertions minus 7 new is 48 pre-existing, not 52.
- **"two of the seven new cases were already green" undercounts — it is four** (24 control, 25, 26, 27). The test file's own block comment is more accurate than the PR body here.
- **"full suite green" is not what I measured**: `PASS=121 FAIL=2 TOTAL=123`. The two failures are `test_pr_create_gate_anchor.sh` and `test_validate_pr_create_structural_parse.sh` — both **pass standalone** (21/0), both cover `validate-pr-create.sh` which this PR does not touch, and both were last modified by an unrelated commit. So **not a regression from this PR**, an aggregate-runner interaction. But the total is 123 with 2 red, not green, and the claim should say so.

- `nit:` **The `--repo=owner/name` equals form resolves to nothing.** `extract_repo_after` requires `--repo` followed by whitespace, so the equals spelling yields an empty `TARGET_REPO` and the hook exits 0 as "nothing to evaluate". `gh` accepts that spelling. Pre-existing (the old `sed` had the same gap) and **not** introduced here, but undisclosed — and unlike terminator widening it is safe to fix, since this is the flag/value *separator*, not a value *terminator*, so it carries none of the #227 risk.
- `nit:` A leading quoted string containing a REST-path repo reference for a private repo, with no `--repo` anywhere, resolves to that private path and exits 0. Pre-existing and unchanged by this PR (the path fallback was already first-occurrence); noting it only so the residue list is complete.
- `suggestion:` The comment blocks are excellent and I would keep every load-bearing line, but the anchor rationale is restated at near-full length in three places (hook comment, test file comment, PR body). One canonical home with pointers would age better.

### AgDR

I agree an AgDR is not warranted for the fix itself — a bug fix to existing logic, no new pattern, and the security rail in `agdr-decisions.md` is satisfied by the review chain rather than by a record.

One pointer: sibling PR #1075 landed **AgDR-0113** governing command-text parsing helpers. It is not in this worktree (this branch predates it). `find_target_repo` is a new command-text parsing helper, so on rebase please check whether AgDR-0113 imposes anything here and reference it if so — I could not evaluate it from this branch.

---

### Verdict

**CHANGES REQUESTED**

To be clear about what I am asking for: the truncation-refusal half is good work and I would keep it unchanged. The blocker is that target-repo resolution moved the fail-open window rather than closing it, and the residue section currently says the opposite. Minimum to clear: make resolution fail-closed across decoy position (the union-of-candidates shape, or another approach that over-blocks rather than skipping), pin the three measured shapes with tests, and correct the residue text. If you would rather ship the resolution fix as-is and handle the mirror bypass in a follow-up, I could live with that — but then the residue section must say plainly that these shapes regressed, and the follow-up needs to exist before this lands, because "documented as pre-existing" is exactly how this one would get lost.

---
Reviewed commit: `0d19eb00941ef3a217d7e1038d25adf72288afee`

atlas-apex:
Literal `bash bin/run-hook-tests.sh` output at HEAD (`63177da`), this run, no rounding:

```
============================================================
  hook test suite: PASS=123  FAIL=0  SKIP(quarantined)=0  TOTAL=123
============================================================
```

`test_block_private_refs.sh` standalone: `Passed: 59  Failed: 0`.

`shellcheck --severity=warning` clean on both changed files.

atlas-apex (commented):
## Security Review: PR #1080 (round 2)

**Commit**: `63177daa15d4b0fcbf555ed43307b731b40c167a`

### Summary

The class-based resolution I recommended was implemented as specified, and it works: all four round-1 bypass shapes now block, with resolution measured identically to `dev`. The false "pre-existing" claim was retracted and replaced with a verified citation. The extraction/truncation half is untouched. That is the round-1 finding fully closed.

But round 2 does not yet clear the bar, because of a defect that is **upstream of all of it**. The new `find_write_segment` anchor requires a literal whitespace character before `gh`, while step-1 classification uses `\b`. Those two are documented in the code as equivalent. They are not. When they disagree, step 1 says "this is a write", the segment finder says "I found no write", and the hook exits 0 having scanned nothing.

**`out=$(gh pr create --repo me2resh/apexyard --body "...zebrafish...")` is not blocked.** Capturing the created URL in a variable is one of the most common ways an agent issues this exact command.

### Checklist Results

- Data Protection:        **Fail** — private-ref leak reachable at exit 0 via the segment anchor
- Gate & Trust-Chain Integrity: **Fail** — leak gate fails open on shapes `dev` catches
- Extraction / truncation: Pass (untouched, re-proven)

### 1. The four round-1 shapes, re-measured

Instrumented `TARGET_REPO`, not exit codes alone:

| Shape | `dev` resolves | round-1 | this HEAD resolves | this HEAD |
|---|---|---|---|---|
| decoy in `--title` | `me2resh/apexyard` → BLOCK | exit 0, leak | `me2resh/apexyard` | **BLOCK** |
| decoy in `--body` | `me2resh/apexyard` → BLOCK | exit 0, leak | `me2resh/apexyard` | **BLOCK** |
| private-then-public, two real writes | `me2resh/apexyard` → BLOCK | exit 0, leak | `me2resh/apexyard` | **BLOCK** |
| decoy-after (the original #1068 bug) | `other/x` → exit 0, leak | BLOCK | `me2resh/apexyard` | **BLOCK** |

All four fixed, and the original bug stays fixed. The class-based membership test is the right shape and I could not turn "any public repo in the segment" into a *narrower* net anywhere — every over-detection I produced only caused more scanning.

### 2. HIGH — public-class target present but undetected: the segment anchor rejects shell operators that abut `gh`

`find_write_segment` anchors on `(^|[[:space:]])gh...`; step 1 classifies with `\bgh\s+issue\s+create\b`. `\b` matches after `;`, `&`, `|`, and `(`. `[[:space:]]` does not. Measured divergence:

| Command | step 1 `\bgh` | segment `[[:space:]]gh` |
|---|---|---|
| `echo x ;gh <write> --repo <public>` | YES | **no** |
| `out=$(gh <write> --repo <public>)` | YES | **no** |
| `high api --repo <public>` | no | no |

Result — all leak a private token to a public repo at exit 0, and all four are **blocked by `dev`**:

| Shape | `dev` | this HEAD |
|---|---|---|
| `echo hi ;gh <write> --repo <public> --body "zebrafish"` | BLOCK | **exit 0, leak** |
| `echo hi &&gh <write> ...` | BLOCK | **exit 0, leak** |
| `out=$(gh <write> ...)` | BLOCK | **exit 0, leak** |
| `echo hi \|gh <write> ...` | BLOCK | **exit 0, leak** |

This also **defeats the truncation refusal**, which is what makes it structural rather than cosmetic. `--body "zebrafish leak" && gh pr list` correctly REFUSES; wrap the identical command in `out=$(...)` and it exits 0 silently. The segment finder runs before extraction, so when it returns empty, every downstream protection — leak scan and refusal alike — is skipped.

The code comment asserts the anchor is "mirroring the `\b` word-boundary the step-1 classification regex already relies on". It does not mirror it, and that gap is precisely the fail-open. **Direction: dangerous.**

**Fix, validated:** change the anchor's leading alternation to `(^|[^A-Za-z0-9_.-])gh`. Measured against the same inputs — matches `;gh`, `&&gh`, `|gh`, `$(gh`, and plain `gh`; still correctly rejects `high api`. This genuinely mirrors `\b` and keeps the "gh as a substring" protection the comment describes. Please add cases for all four operators, since the round-2 suite has none.

### 3. Other undetected public-target forms (all verified pre-existing on `dev`, none regressions)

| Form | `dev` | this HEAD |
|---|---|---|
| `--repo=<slug>` (equals form) | exit 0 | exit 0 |
| `-R <slug>` (documented gh short flag) | exit 0 | exit 0 |
| `--repo github.com/<slug>` (documented `HOST/OWNER/REPO`) | exit 0 | exit 0 |
| `--repo https://github.com/<slug>` | exit 0 | exit 0 |
| case-variant slug (`ME2RESH/ApexYard`) | exit 0 | exit 0 |
| `--repo "$VAR"` (shell indirection) | exit 0 | exit 0 |

All REST-path forms **are** detected, including leading-slash, quoted, and full `api.github.com` URL, plus mixed `gh api` private + subcommand public. That half is solid.

Of these, **`-R` is worth fixing in this PR**: it is gh's documented short form, the hook already handles the `-t`/`-b` short forms elsewhere, and adding it to the `--repo` alternation is the same one-line shape as the anchor fix. The rest (URL/host forms, case, variable indirection) are genuine AgDR-0104 territory — text matching cannot close them — and documenting rather than fixing is right.

### 4. `--repo=owner/name` — pre-existing, documenting is correct

Confirmed pre-existing: `dev` exits 0 on it too, because the old `sed` required `--repo` followed by whitespace exactly as the new `grep -E` does. Not a regression. Documenting rather than fixing is defensible on its own, though if you take the `-R` fix above, extending the same alternation to accept `=` is nearly free and closes it.

### 5. Test 27's deliberate flip — the call is right

I agree with the author's framing. The round-1 bar ("private target with a later public mention stays a no-op") encoded the *positional* model as a requirement, and that model is what produced the ordering hole. Under class-based resolution the honest answer is: a command that names a public framework repo anywhere in its write segment gets scanned. The failure that produces is an over-block on a chained private+public write — visible, cheap, and fixable by unchaining, which is the shape the framework wants anyway. That is the safe direction, and the test comment documents the flip and its reason plainly rather than quietly re-baselining. Correct call.

### 6. Truncation half — confirmed untouched

The only change to that region between rounds is an added comment; no logic changed. I re-proved it empirically at this HEAD rather than relying on the diff:

Chained-command / trailing-comment / apostrophe-body / command-substitution values all REFUSE. #227, #1046 still BLOCK. #1039 unreadable body-file still fails closed. A lone pipe in legitimate prose still passes (terminator set not widened). Marker in `--label` blocked, marker in body bypasses. Clean single write passes — no DoS.

The added comment's invariant argument is correct and worth keeping: both quoted branches strip their own opening delimiter before returning, so a leading raw quote is structurally unreachable via the success path. It detects the *failure*, not the *attack*, so it does not inherit AgDR-0104's unsoundness. That is a genuinely stronger justification than round 1 carried.

### 7. Counts

Verified independently rather than relayed: my own run of the hook suite at this HEAD gives **`Passed: 59  Failed: 0`**, matching the reported 59/59. I did not observe the aggregate/standalone discrepancy flagged in the calibration note.

### Dangerous-vs-safe split

- **Dangerous (silent no-scan on a public target):** the segment-anchor finding — four operator forms, one of which (`$(...)` capture) is a routine agent idiom. Regression against `dev`.
- **Safe (over-block):** every truncation refusal, the test-27 flip, the class-based widening, the strict-marker asymmetry.
- **Pre-existing, unchanged:** the value-form table in §3.

Apart from the anchor, the residue is now overwhelmingly the safe kind — which is the bar. The anchor is the one thing standing between this PR and that being true.

### Verdict

**CHANGES REQUESTED**

One defect, one validated one-line fix. No `*-security.approved` marker was written: `out=$(gh <write> --repo <public> --body "<private>")` places a private token on a public repo at exit 0, and `dev` blocks it.

With the anchor changed to `(^|[^A-Za-z0-9_.-])gh` plus cases for `;` `&&` `|` `$(`, and ideally `-R` added to the `--repo` alternation, I expect to approve on the next pass.

---
📌 Reviewed commit: `63177daa15d4b0fcbf555ed43307b731b40c167a`

atlas-apex (commented):
## Code Review: PR #1080 (round 2)

**Commit**: `63177daa15d4b0fcbf555ed43307b731b40c167a`

### Summary

Round 1's positional resolution ("first `--repo` after the anchor") is replaced by a **class-based membership test**: within the write segment, does *any* known public-class repo appear as a `--repo` value or a `repos/<owner>/<repo>` path? Both round-1 blockers are fixed. I verified by measurement, three-way (dev blob / round-1 / round-2), not by reading the claims.

### Checklist Results

- Architecture & Design: Pass — the positional-to-class reframing is the right shape
- Code Quality: Pass — simpler than round 1; `find_write_segment` now does one job
- Testing: Pass — 59/59, round-2 cases independently RED-proven against round 1
- Security: Pass — no shape found where round 2 under-detects relative to dev or round 1
- Performance: Pass
- Technical Decisions (AgDR): N/A — bug fix; AgDR-0113's citation discipline is now applied in-comment
- Adopter Handbooks: N/A

---

### Blocker verification (measured, three-way)

| Shape | dev | round 1 | round 2 |
|---|---|---|---|
| Decoy `--repo` inside `--title`, real public repo after, leak in body | BLOCK | **ALLOW** | **BLOCK** |
| Decoy `--repo` inside `--body` prose, real public repo after | BLOCK | **ALLOW** | **BLOCK** |
| Two real writes chained, private first, public second carrying leak | BLOCK | **ALLOW** | **BLOCK** |
| Control: plain public write with leak | BLOCK | BLOCK | BLOCK |
| Private-only write, clean, no public mention | ALLOW | ALLOW | ALLOW |
| Skip marker in body on a public write | ALLOW | ALLOW | ALLOW |

Both blockers are genuinely closed, and the escape hatch and private-target no-op both still work.

**Blocker 2 is properly retracted.** The corrected comment cites the check against the real `upstream/dev` blob and states plainly that round 1 introduced the hole. That matches what I measured independently. The added rule — every "pre-existing" claim in this file must cite the check that proved it against the actual dev blob rather than be reasoned from the code's shape — is the right generalisation, and is worth more than the specific fix.

---

### 1. Is class-based safe in the direction claimed?

For the property as stated — *a false YES only causes more scanning* — yes. `TARGET_REPO` is now used only for message interpolation and the target's-own-name exemption; the security decision is the YES/NO membership test, and a spurious YES can only widen scanning.

**The segment boundary specifically: I could not make it drop a real target.** The segment starts at the leftmost recognised write shape and runs to end of command, so it is over-inclusive by construction — anything that shifts the anchor *earlier* only enlarges it. I tried decoy anchor text in a leading quoted string, mixed `api`/`create` shapes in one line, and heredoc bodies; in each the segment still contained the real target.

The one divergence worth a comment: step-1 classification uses `\bgh`, while the anchor uses `(^|[[:space:]])gh`. Step 1 is therefore marginally *looser*, so a string that classifies as a write but produces an empty segment exits 0. The only strings that differ are shapes like `foo-gh issue create`, which is not an invocation, so I don't consider it reachable — but the two regexes should carry a comment noting they must stay aligned, since a future loosening of the anchor's boundary is the thing that would turn this into a real hole.

#### Shapes where a public-class target is present but NOT detected

I found five. **All five are ALLOW on the dev blob too — none is introduced by round 1 or round 2.** Round 2 is a strict improvement over both on everything I could construct. Listing them because you asked for the dangerous direction, and because only one is currently documented:

| # | Shape | dev / r1 / r2 | Status |
|---|---|---|---|
| U1 | `-R owner/repo` — gh's documented short alias for `--repo` | ALLOW / ALLOW / ALLOW | **undocumented** |
| U2 | Case variation (`me2resh/ApexYard`) — gh slugs are case-insensitive, the grep is not | ALLOW / ALLOW / ALLOW | **undocumented** |
| U3 | Slug immediately followed by a shell operator with no space (`...apexyard&&true`) — the trailing boundary class is `["'[:space:]]|$` and excludes `&`, `;`, `|`, `)` | ALLOW / ALLOW / ALLOW | **undocumented** |
| U4 | REST path with a `?query` suffix — the path boundary class is `[/[:space:]"'>]|$` and excludes `?` | ALLOW / ALLOW / ALLOW | **undocumented** |
| U5 | `--repo=owner/name` equals form | ALLOW / ALLOW / ALLOW | documented, claim verified |

**U1 is the one I would prioritise.** `-R` is gh's own documented shorthand, it needs no craft at all, and there is an internal asymmetry that makes it easy to miss: this hook already recognises short flags for *content* (`-t`, `-b`, `-F`) but not for the *repo*. Anyone reading the flag handling would reasonably assume `-R` was covered.

None of these blocks this PR — they are pre-existing gaps in a hook that is measurably better after this change than before it, and widening the boundary classes is exactly the kind of change this file's history says should be done deliberately with its own tests rather than bolted onto a fix. But U1–U4 should be written into the residue list with the same dev-blob citation discipline the PR just adopted, and U1 is worth a follow-up ticket.

### 2. The test-27 flip

**Right resolution, honestly stated.** The old expectation asserted `exit 0` for "private target, public repo named in a later chained call" — which, under a class-based rule, is precisely the fail-open case. Keeping it would have pinned the hole as a requirement, which was my round-1 objection to that case.

The new comment says it plainly: *"DELIBERATELY FLIPPED in round 2 (was 'stays a no-op' in round 1 — see the block comment above case 21 for why that was the wrong bar)"*, and the tradeoff is stated in the hook comment too. It is labelled a deliberate over-block, not dressed up as a bug fix. That is the honest framing.

Measured: dev ALLOW, round 1 ALLOW, round 2 REFUSE. Note the *mechanism* — it lands on the truncation refusal, not a named-leak block, so the operator gets "could not safely determine where the value ends" and the unchain instruction rather than a false claim that a project leaked. The test asserts that exact string, so the expectation matches the real behaviour rather than approximating it.

### 3. The `--repo=` equals-form residue

**Claim verified.** Measured on the actual dev blob: `exit 0`, no scan — identical to round 1 and round 2. dev's `sed` required `--repo` followed by whitespace, so the equals spelling never resolved a target there either. "Pre-existing gap, not a regression" is accurate, and pinning it as a test case rather than leaving it undiscovered is the right call.

### 4. The truncation half

**Confirmed untouched.** The only delta in that region is the added comment. `TITLE_TRUNCATED` / `BODY_TRUNCATED`, the refusal branch, the message, and both `extract_flag_value` quoted branches are byte-identical to round 1. The `#1039 r2` boundary behaviour is unchanged (a lone `|`, `;`, `&` in body prose is still not a terminator), and 59/59 includes every #227 / #1039 / #1046 case.

The recorded invariant argument is stated accurately — it points at the actual `sub(/^--title.../)` calls as the evidence that a leading raw quote is unreachable through the success path, rather than just asserting it. That is the difference between an invariant and a pattern guess, and it now reads as one.

Restating the known cost, unchanged since round 1 and still the right trade: a **clean** body plus any chained command or trailing comment now refuses where dev allowed (measured: dev ALLOW, r1/r2 REFUSE). Fail-closed, actionable message, documented.

### 5. Counts

- `test_block_private_refs.sh` on this HEAD: **59 passed, 0 failed** — run, not read.
- **48 pre-existing + 7 round-1 + 4 round-2 = 59 reconciles exactly.** `run_case` count went 40 (base) → 47 (r1) → 51 (r2); 59 total assertions minus 51 `run_case` = 8 non-`run_case` assertions, consistent across all three revisions. Strictly additive, nothing deleted.
- **Round-2 cases independently RED-proven against the round-1 commit**, by direct probe rather than by reading the claim: all three regression shapes ALLOW on round 1 and BLOCK on round 2; test 27's shape ALLOW on round 1 and REFUSE on round 2.
- The round-1 split is now stated correctly as 3 fail→pass and 4 already green.

**On the full suite — my round-1 number was wrong and I'm retracting it.** I reported `PASS=121 FAIL=2`. Both files I named (`test_pr_create_gate_anchor.sh`, `test_validate_pr_create_structural_parse.sh`) **PASS** in this tree — I saw the latter pass in a full run against these files, and confirmed the former standalone at 21/0. I could not complete a clean end-to-end run to print the tally myself (the run exceeded my time budget on a loaded machine), so I'm accepting 123/0 on the coordinator's and author's runs rather than restating a number I can't reproduce. My conclusion that it was never this PR's regression stands; the figure I attached to it did not, and the author should not have spent any time on it.

---

### Verdict

**APPROVED**

The reframing from "which repo is the target?" to "is a public repo plausibly in scope?" is the right move, and it is the rare fix that makes the code simpler while closing the hole — `find_write_segment` now does one job instead of three. Both blockers are closed by measurement, the retraction is properly evidenced, and the AgDR-0113 citation rule the author added is a better outcome than the bug fix itself.

Follow-ups, none blocking: document U1–U4 in the residue list with the dev-blob citation discipline, and file U1 (`-R` short form) separately — it is the most reachable undetected-target shape left in this hook.

---
Reviewed commit: `63177daa15d4b0fcbf555ed43307b731b40c167a`

atlas-apex (commented):
## Code Review: PR #1080 (round 3)

**Commit**: `596a95b43782d57635cc8abf1da631fc79446081`

### Summary

I had posted a round-2 approval when HEAD moved under me — `596a95b` landed while I was writing. I pulled the marker rather than approve a commit I hadn't read, and reviewed the new commit. It fixes the write-segment anchor's word boundary and adds `-R`. Both changes are correct and both close things I had underweighted.

**I got one of these wrong in round 2 and want to be direct about it.** I flagged the anchor/step-1 divergence (`(^|[[:space:]])gh` vs `\bgh`) and then dismissed it: *"the only strings that differ are shapes like `foo-gh issue create`, which is not an invocation, so I don't consider it reachable — but the two regexes should carry a comment noting they must stay aligned."* That was wrong. `\b` fires after **any** non-word character, which includes `(` — so `out=$(gh issue create …)`, a routine agent idiom for capturing a created issue URL, diverges too. I reasoned about which *literal* strings differed instead of enumerating the character class, recommended a comment where a fix was needed, and left a live bypass on the table. The security pass caught it. Credit there, and the lesson is mine: when two expressions are claimed equivalent, diff the character classes, don't sample strings.

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass — 65/65; round-3 cases pin all six shapes
- Security: Pass — anchor now matches step 1's classification; no under-detection found
- Performance: Pass
- Adopter Handbooks: N/A

---

### Verification (measured three-way: dev blob / round 2 / round 3)

**The anchor fix — segment-empty bypass.** Step 1 classified these as writes; the round-2 anchor found no match, returned an empty segment, and the hook exited 0 having scanned nothing.

| Shape | dev | round 2 | round 3 |
|---|---|---|---|
| `out=$(gh …)` command substitution | ALLOW | **ALLOW** | **REFUSE** |
| backtick capture | ALLOW | **ALLOW** | **REFUSE** |
| `;gh …` | BLOCK | **ALLOW** | **BLOCK** |
| `&&gh …` (no space) | BLOCK | **ALLOW** | **BLOCK** |
| `\|gh …` | BLOCK | **ALLOW** | **BLOCK** |
| `$(…)` wrapping the truncation-refusal shape | ALLOW | **ALLOW** | **REFUSE** |
| `high api …` must NOT anchor (false-anchor guard) | ALLOW | ALLOW | ALLOW |

The sixth row is the one I would highlight: wrapping in `$(…)` silently defeated the *truncation refusal* as well, because the segment check runs upstream of content extraction. A shape that correctly refuses unwrapped exited 0 wrapped. Catching that the fix has a blast radius beyond its own block — and saying so in the comment — is good work.

**`-R` short form** (the gap I raised in round 2): dev ALLOW, round 2 ALLOW, round 3 **BLOCK**, quoted and unquoted. Framing it as closing an inconsistency rather than adding scope is right — the hook already recognised `-t`/`-b`/`-F`, so `-R` was the odd one out.

**No regressions.** All eight control shapes hold: the three round-1 regression shapes still BLOCK, the plain-leak control still BLOCKs, and private-only writes, clean public writes, the skip marker, and non-`gh` commands all still ALLOW.

**Boundary class.** `[^A-Za-z0-9_.-]` is very slightly narrower than `\b` — it also excludes `-` and `.` predecessors, so `foo-gh` / `x.gh` won't anchor where `\b` would. Neither is a real invocation, and `/usr/local/bin/gh`, `` ` ``, `(`, `{`, `$`, newline and whitespace all anchor correctly. Narrower in the safe direction here, since the residual non-matches aren't invocations. Fine as written.

### One claim to correct (non-blocking)

The new comment states that all four divergent shapes *"BLOCK on the real `upstream/dev` blob and LEAKED (exit 0, nothing scanned) on the round-2 HEAD this replaces."*

Three of the four hold — `;gh`, `&&gh`, `|gh` all BLOCK on dev. But **`out=$(gh …)` ALLOWs on dev too**, which I measured directly. dev leaks it by a *different* route: the `)` after the body's closing quote defeats dev's quoted-branch terminator, extraction falls to the unquoted fallback, and dev has no truncation detection to catch it. So for the shape the comment singles out as mattering most, this is **pre-existing on dev, not a round-2 regression**.

Two reasons to fix the wording rather than let it ride. First, it points the wrong way from round 2's correction: that one wrongly labelled a new hole "pre-existing", this one wrongly labels a pre-existing hole a round-2 regression. Second, the claim is an application of the citation rule the author added one commit earlier — the rule is right, and it should hold on the commit that invokes it. The accurate version is stronger anyway: *`$(…)` leaked on dev through the extraction path and on round 2 through the anchor path; round 3 is the first revision that catches it.*

### Residual gaps (unchanged, correctly out of scope)

Verified still ALLOW on dev, round 2 and round 3 alike — all pre-existing, none touched by this PR:

| Shape | Note |
|---|---|
| Case variation (`me2resh/ApexYard`) | gh slugs are case-insensitive; the grep is not |
| Slug immediately followed by a shell operator (`…apexyard&&true`) | trailing boundary class excludes `&`, `;`, `\|`, `)` |
| REST path with `?query` suffix | path boundary class excludes `?` |
| `--repo=owner/name` | documented and tested |

Right call not to widen the boundary classes inside a bug fix. Worth a follow-up ticket together, with the dev-blob citation discipline applied.

### Counts

- `test_block_private_refs.sh` on this HEAD: **65 passed, 0 failed** — run, not read.
- **48 pre-existing + 7 round-1 + 4 round-2 + 6 round-3 = 65**, reconciles exactly. `run_case` 40 → 47 → 51 → 57; 65 total minus 57 `run_case` = 8 non-`run_case` assertions, stable across all four revisions. Strictly additive.
- Round-3 cases independently RED-proven against round 2 by direct probe: all six shapes ALLOW on round 2, and BLOCK or REFUSE on round 3.
- Full suite: I could not complete an end-to-end run within my time budget on a loaded machine, so I'm not restating a figure. My round-1 `121/2` remains retracted — both files I named pass in this tree.

---

### Verdict

**APPROVED**

Three rounds in, the hook is meaningfully better than the dev blob it replaces on every shape I could construct, and the process improvement is the more durable output: the citation rule now in the file is what turns "I believe this is pre-existing" into a checkable claim. I'd ask for two follow-ups, neither blocking: correct the `$(…)`-on-dev claim, and file the four residual boundary gaps as one ticket.

---
Reviewed commit: `596a95b43782d57635cc8abf1da631fc79446081`

atlas-apex:
Round 3 pushed (`596a95b`), then a comment-only correction (`5253442`) after re-checking my own round-3 claim and finding it wrong for one of four cases — same mistake class as round 1's, caught this time before merge rather than by review.

**What changed in `596a95b`:** `find_write_segment`'s anchor was `(^|[[:space:]])gh`, with a comment claiming it mirrors step 1's `\bgh` (GNU grep word boundary). It doesn't — `\b` also fires after `;`, `&`, `|`, `(`, `$`, not only whitespace. When step 1 and the anchor disagreed, the segment came back empty and the hook exited 0 having scanned nothing — upstream of round 2's class-based check and round 1/2's truncation refusal alike. Fixed with `(^|[^A-Za-z0-9_.-])gh`. Added `-R` (gh's documented short flag for `--repo`) in the same pass.

**What changed in `5253442` (comment-only, no behaviour change):** the claim that all four divergent shapes (`;gh`, `&&gh`, `|gh`, `$(gh`) block on `dev` was right for three and wrong for the fourth. `out=$(gh issue create ...)` (a single, unchained write — the operationally important shape) also allows on `dev`, because `dev` has no truncation-detection at all and the trailing `)` defeats its extraction silently. Pre-existing on `dev`, not a round-2 regression; this branch still closes it as a byproduct, which is an improvement over `dev`, not parity restoration. Caught by re-running the claim with a body where the leak token isn't the first word — the earlier check used a frontloaded body that let `dev`'s truncated read catch the leak by accident and mask the gap.

Also added the consolidated residue line for four further pre-existing gaps found in review (case variation, slug+operator with no space, REST `?query`, `--repo=` equals form) — spot-verified against the real `dev` blob, left alone per instruction, one follow-up ticket to cover all four.

**Verification, literal output:**

`test_block_private_refs.sh`: `Passed: 65  Failed: 0` (unchanged by the comment-only fix).

`bash bin/run-hook-tests.sh` at `5253442`:
```
============================================================
  hook test suite: PASS=123  FAIL=0  SKIP(quarantined)=0  TOTAL=123
============================================================
```
(literal output from the clean re-run at the prior commit `596a95b`, which the comment-only diff in `5253442` cannot affect — no code changed, only comments)

`shellcheck --severity=warning`: clean on both files.

Halting here — no further changes without another review pass.

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1080 (round 4)

**Commit**: `52534425e23c9db82a4e75f19ccad44f58cb58e0`

### Summary

Comment-only delta correcting the `$(…)`-on-dev claim I flagged in round 3, plus a consolidated residue line for the four pre-existing gaps. **Confirmed executably inert; the corrected comment is accurate and does not over-correct.** Re-approving at this HEAD.

### The delta is executably inert — two independent checks

1. **Content hash.** Comment-stripped, blank-normalised: **370 lines, identical md5** across `596a95b` and `52534425` for the hook, and identical for the test file. Matches your result (different absolute hash, different strip method — what matters is A == B under one consistent method, which holds).
2. **Hunk read.** A hash of stripped content would hide a change to a `#` *inside* a string literal, so I read every hunk. All four are pure comment blocks; no code line, string, or regex is touched. `bash -n` clean.

Behaviourally re-ran the full battery against `596a95b` and `52534425` side by side — `-R`, decoy-in-title, clean public write, private-only write, skip marker, false-anchor guard, and all four boundary shapes: **identical on every one**. Suite unchanged at **65/65**, `run_case` 57.

### Is the corrected comment accurate?

Yes, on every claim I could check, and it is precise about the distinction rather than blurring it:

| Claim | Measured |
|---|---|
| `;gh`, `&&gh`, `|gh` BLOCK on dev, leaked on round 2 — genuine regressions | dev BLOCK / r2 ALLOW / r3–r4 BLOCK ✓ |
| `out=$(gh …)` ALSO leaks on dev (rc=0), `"found during zebrafish rebuild"` truncating to `"found` | dev ALLOW ✓, and that is the exact mechanism |
| Pre-existing on dev, **not** a round-2 regression | ✓ |
| This fix still closes it — as a byproduct of the segment being non-empty so round 1/2's refusal can run | r3–r4 REFUSE, not a named-leak BLOCK ✓ |
| "improvement over dev on this shape, not parity restoration" | ✓ — dev allows it, so there is no parity to restore |
| Case 36 is not a dev-comparison case; dev never had a refusal to defeat | ✓ dev ALLOW |
| Case 37 (`-R`) is new capability, not a regression fix | dev ALLOW / r2 ALLOW / r3–r4 BLOCK ✓ |
| Four residual gaps ALLOW on dev, r2 and r3 alike — none introduced or widened here | ✓ all four |

**On over-correction specifically** — the failure mode would be sweeping the three real regressions into the same "pre-existing" bucket, which is what round 2's error did in the opposite direction. It doesn't. The three genuine regressions are still named as regressions and I re-verified each with a **non-frontloaded** body (dev BLOCK, r2 ALLOW). The one pre-existing shape is separated out with its mechanism. And it resists the easier overstatement in the other direction too: it says this round *exceeds* dev on `$(…)` rather than claiming parity. That is the harder sentence to write and it is the correct one.

### The fixture that passed for the wrong reason

This is the part worth flagging beyond this PR. I verified the root cause rather than accepting it, and it reproduces exactly — same command shape, only the body's word order changed:

| Body | dev | r2 | r3 | r4 |
|---|---|---|---|---|
| `"zebrafish leak"` (leak frontloaded) | **BLOCK** | ALLOW | REFUSE | REFUSE |
| `"found during zebrafish rebuild"` (leak not frontloaded) | **ALLOW** | ALLOW | REFUSE | REFUSE |

dev's truncated read keeps roughly the first word. Put the leak token there and dev catches it **by accident**, through the very truncation that is the bug. The fixture reports GREEN and certifies the opposite of the truth.

That is a different and more dangerous class than "didn't verify" — the check ran, the evidence looked real, and the artefact was a green result pointing the wrong way. Worth crediting: catching it required going back to a check that had already passed and asking why. The generalisable rule, and I'd suggest it belongs alongside the AgDR-0113 citation discipline this file already carries: **when the mechanism under test is truncation, the fixture must place the signal outside the retained fragment** — otherwise the test exercises the accident, not the mechanism. My own round-3 measurement got the right answer only because I happened to use a non-frontloaded body throughout; that was luck, not method.

### Residue line

Accurate. All four spot-checks match mine (case variation, slug followed immediately by a shell operator, REST `?query` suffix, `--repo=`) — ALLOW on dev, r2, r3 and r4 alike. One consolidated follow-up ticket rather than four is the right call; they share a root cause (boundary character classes that don't cover shell-adjacent punctuation), and splitting them would invite four separate narrow widenings of exactly the kind this file's history warns about.

---

### Verdict

**APPROVED**

Comment-only, executably inert, and the correction is accurate in both directions. The residue list now reflects what I measured, and the file carries an honest account of a green test that was lying. Nothing further from me.

---
Reviewed commit: `52534425e23c9db82a4e75f19ccad44f58cb58e0`

atlas-apex (commented):
## Security Review: PR #1080 (round 3)

**Commit**: `52534425e23c9db82a4e75f19ccad44f58cb58e0`

### Summary

The anchor fix landed as specified and the HIGH finding is closed. All eight bypass shapes now block, `-R` was folded in, and the truncation refusal is intact — including the interaction that the anchor bug had been defeating. I found no shape where a real `gh` write to a public-class repo goes unscanned.

I also want to flag the round-2 correction explicitly, because it corrects **me**. My round-2 review asserted all four divergent shapes block on `dev`. Three do. `out=$(gh …)` does not — and the reason is that my test body was `"zebrafish"`, leak token first. `dev` has no truncation detection, so its truncated one-word read happened to contain the token and caught it by accident. Re-measured with a realistic body (`"found during zebrafish rebuild"`), `dev` exits 0 on that shape. So it was pre-existing, not a round-2 regression, and this PR *improves* on `dev` there rather than restoring parity. The author caught a methodological error in my own testing and documented it precisely. That is the right instinct and the comment now says it accurately.

### Checklist Results

- API Security:           Pass
- Gate & Trust-Chain Integrity: Pass
- Extraction / truncation: Pass

### 1. Full bypass table, re-measured (instrumented `TARGET_REPO`)

| # | Shape | `dev` | this HEAD | resolved |
|---|---|---|---|---|
| S1 | decoy `--repo` in `--title` | BLOCK | **BLOCK** | `me2resh/apexyard` |
| S2 | decoy `--repo` in `--body` | BLOCK | **BLOCK** | `me2resh/apexyard` |
| S3 | private-then-public, two real writes | BLOCK | **BLOCK** | `me2resh/apexyard` |
| S4 | decoy-after (the original #1068 bug) | exit 0, leak | **BLOCK** | `me2resh/apexyard` |
| B1 | `;gh` (no space) | BLOCK | **BLOCK** | `me2resh/apexyard` |
| B2 | `&&gh` | BLOCK | **BLOCK** | `me2resh/apexyard` |
| B3 | `out=$(gh …)` | exit 0, leak (pre-existing) | **REFUSE** | `me2resh/apexyard` |
| B4 | `\|gh` | BLOCK | **BLOCK** | `me2resh/apexyard` |
| V6 | `-R <slug>` short flag | exit 0 | **BLOCK** | `me2resh/apexyard` |
| V6b | `-R` with a decoy `-R` in `--title` | exit 0 | **BLOCK** | `me2resh/apexyard` |

`-R` is a genuine improvement over `dev`, and it resists the same decoy-ordering attack as `--repo` because it went into the class-based membership test rather than a positional one.

### 2. Attacking the new anchor

I compared step-1 classification (`\bgh`) against the new segment anchor (`(^|[^A-Za-z0-9_.-])gh`) across twelve prefix forms. Three still diverge — step 1 says "write", the segment finder says "none":

| Form | step 1 | segment | Is it a real `gh` write? |
|---|---|---|---|
| `x.gh <write>` | YES | no | **No** — `x.gh` is a different binary |
| `some-gh <write>` | YES | no | **No** — different binary |
| `"gh issue create" -r …` | YES | no | **No** — not a valid invocation |

In all three the segment finder is **right** and step 1 is over-classifying. Exiting 0 is the correct answer because no `gh` write executes. `dev` blocks the first two, which is `dev` over-blocking a non-`gh` command, not a protection this PR gives up. The divergence set is now exactly the set of non-invocations, which is what you want.

Everything that *is* a real invocation anchors correctly: `;gh`, `&&gh`, `|gh`, `$(gh`, backtick, `sudo gh`, `/usr/local/bin/gh`, tab-separated, and plain `gh`. The `high api` false-anchor guard still holds.

**Inverse direction (segment starting too early):** cannot cause a miss. The segment runs from the anchor to end-of-string, so an earlier start is strictly a superset — more text scanned, never less. Confirmed empirically: prose containing `gh api` ahead of the real write still blocks; a private-target write that merely mentions the public repo in prose still exits 0 correctly.

**One residue not in the documented list:** a line continuation *between* `gh` and its subcommand — `gh \`⏎`issue create --repo <public> …` — exits 0 on **both** `dev` and this HEAD. It is step-1's line-based `grep` that misses it, not the anchor, so it is pre-existing and untouched by this PR. Breaking the line after the subcommand (`gh issue create \`⏎`--repo …`) blocks correctly, which is the far more common style. Worth adding to the same follow-up ticket; not a blocker.

### 3. Truncation refusal — intact, and the interaction is now correct

| Probe | `dev` | this HEAD |
|---|---|---|
| Quoted body + chained command | exit 0 | **REFUSE** |
| Same, wrapped in `$( )` — the shape the anchor bug defeated | exit 0 | **REFUSE** |
| #227 leak after embedded quote | BLOCK | BLOCK |
| #1046 embedded quote + later `--flag` | BLOCK | BLOCK |
| #1039 unreadable body-file | BLOCK | BLOCK |
| Lone pipe in legitimate prose | exit 0 | exit 0 (terminator set not widened) |
| Marker in `--label` | BLOCK | BLOCK |
| Marker in body | — | bypass + warning |
| Clean single write | exit 0 | exit 0 (no DoS) |

The second row is the one that matters: in round 2 the empty segment short-circuited the refusal, so wrapping a chained command in `$( )` turned a REFUSE into a silent exit 0. It now refuses in both spellings. The anchor sits upstream of extraction, and with it fixed the refusal gets to run on every shape that reaches it.

### 4. The four documented residues — pre-existing, and documenting is right

Verified against the `dev` blob:

| Residue | `dev` | this HEAD | Pre-existing? |
|---|---|---|---|
| Case variation (`ME2RESH/APEXYARD`) | exit 0 | exit 0 | Yes |
| Slug immediately abutting a shell operator | exit 0 | exit 0 | Yes |
| REST `?query` directly after the slug | exit 0 | exit 0 | Yes |
| `--repo=` equals form | exit 0 | exit 0 | Yes |

All four confirmed unchanged. Note the `?query` case is narrower than it reads: the ordinary form `repos/<slug>/issues?state=open` **blocks** on both, because the `/` before `issues` terminates the match. Only a query abutting the slug itself slips, which additionally needs `issues`/`pulls` elsewhere in the command to classify at all.

Documenting rather than fixing is the right call for this PR. These are text-matching residues in AgDR-0104 territory — case-insensitivity and operator-abutment are each a boundary-regex change with their own false-positive surface, and bundling them into a bug fix that already spans three rounds and two mechanisms would widen the blast radius of the thing being fixed. One consolidated follow-up ticket over four separate ones is also the right granularity: they share a root cause (slug matching is textual and boundary-sensitive) and will want a single coherent fix.

### Direction check

Every imperfection I found is in the safe direction — over-blocking (`dev`-style blocks on non-`gh` binaries that this PR correctly stops issuing), or refusal when extraction cannot be trusted. The dangerous direction — a real write to a public-class repo scanned partially or not at all — I could not reach at this HEAD.

### Counts

Verified independently rather than relayed: my own run of the hook suite at this HEAD gives **`Passed: 65  Failed: 0`**, matching the reported 65/65. I also independently confirmed the comment-only claim for `5253442`: comment-stripped, blank-line-normalised executable content is byte-identical to `596a95b` (370 lines, same md5).

### Verdict

**APPROVED**

The `*-security.approved` marker is written at this HEAD. Three rounds, and the shape of each was right: round 1 swapped one ordering hole for another, round 2 fixed the ordering but left an anchor that silently disabled the whole gate on a routine idiom, and round 3 closes it with the boundary that actually mirrors `\b`. The extraction half was never disturbed. Please carry the follow-up ticket for the four documented residues plus the `gh \`⏎`subcommand line-continuation gap.

---
📌 Reviewed commit: `52534425e23c9db82a4e75f19ccad44f58cb58e0`

Issue #1078, opened 2026-07-30
[Bug] Release changelog: no timeout on the gh lookup, so a hung forge stalls a release cut


atlas-apex:
## Bug Scenario

**Given** `resolve_issue_from_pr()` in `bin/release-changelog.sh` calls `gh` once per commit needing resolution

**When** the forge is reachable but unresponsive — hanging rather than failing

**Then** the call never returns and the release cut stalls indefinitely, producing no output and no error

**Expected** a hung forge should time out and fall through the existing fail-soft path

## Repro Steps

Observed in review: a hung `gh` ran until an external `timeout` killed it at 12s (rc=124). Nothing in the script bounds it.

77 of the last 800 commits would each make a serial call, so an unresponsive API compounds — the cut hangs rather than degrading.

## Why this is separate from the fail-soft work already done

#1074 explicitly handles and tests four failure modes: `gh` exiting non-zero, empty stdout with rc=0, malformed JSON, and `gh` absent from PATH. All degrade cleanly and the release never aborts. That was the stated requirement and it was met.

**A hang is a fifth mode that was not considered.** "Never abort" and "never block" are different requirements. This is not a gap in the testing so much as a gap in the requirement it was written against.

Not hypothetical: `api.github.com` stopped resolving mid-merge during the session that produced #1074, so forge unavailability is a live condition in this workflow, not a thought experiment.

## Severity

**low-medium** — no wrong output, and the operator sees a stalled command rather than a bad release. But it stalls a release cut with no diagnostic, and the natural operator response (Ctrl-C, re-run) is fine only if they realise what happened.

## Suggested fix

Bound the call — `timeout <n>s gh ...` or the equivalent — and treat expiry as another fail-soft path, degrading exactly as the four handled modes do. Add it to the same test that stubs `gh`, so all five modes are covered in one place.

## Related

Found in the code review of #1074, filed rather than fixed there because that PR shipped deliberately with known residue.

## Glossary

| Term | Definition |
|------|------------|
| fail-soft | Degrading to a usable result on error rather than aborting |
| release cut | Running `/release` to generate the changelog and open the release PR |

atlas-apex:
Closing as not worth tracking separately — folding into #1077.

[...]

Issue #1077, opened 2026-07-30
[Bug] Release changelog: PR_LOOKUP_REPO hardcodes me2resh/apexyard, so forks emit unrelated live closes


atlas-apex:
## Bug Scenario

**Given** a fork of apexyard cutting its own release with `bin/release-changelog.sh`

**When** `resolve_issue_from_pr()` looks a PR number up to find the issue it closed

**Then** it queries `PR_LOOKUP_REPO`, hardcoded to `me2resh/apexyard` — so the fork resolves **its own** PR numbers against **upstream's** numbering, and emits whatever unrelated issue upstream's PR of that number happened to close

**Expected** the lookup should target the repo the release is actually being cut for

## Why this matters more than it looks

apexyard is a framework **built to be forked** — that is the documented adoption model. Every adopter cutting a release hits this, not an edge case.

Before #1074 it was harmless: `Closes #A, #B, #C` was a comma list and GitHub honoured only the first, so mis-resolved refs were inert text. **#1074 made every emitted ref live.** So on a fork, a release cut can now auto-close an arbitrary local issue whose number happens to match an upstream PR's closed issue.

Failure is silent and only visible to whoever reads that fork's release PR body closely enough to notice a number they do not recognise.

## Repro Steps

1. Fork `me2resh/apexyard`.
2. Merge a PR in the fork — say it becomes the fork's PR #12.
3. Cut a release; the changelog generator resolves `#12` against `me2resh/apexyard`, not the fork.
4. Whatever issue upstream's PR #12 closed is emitted as a live `Closes` line in the fork's release PR.

## Severity

**medium-high for adopters, low for this repo.** In `me2resh/apexyard` itself the hardcoded value is correct, so it never misbehaves here — which is exactly why it could sit unnoticed. It is wrong for every fork, systematically.

## Suggested fix

Derive the repo from the actual remote rather than hardcoding it. If it cannot be derived confidently, **emit no `Closes` line** rather than guessing — the same missing-over-wrong principle as the sibling ticket.

Worth checking the same class elsewhere in the release path while fixing: any other hardcoded `me2resh/apexyard` is wrong for the same reason.

## Related

Found in the code review of #1074. Filed rather than fixed there because #1074 was shipped deliberately with known residue; see the sibling ticket for the ref-extraction shapes.

## Glossary

| Term | Definition |
|------|------------|
[...]

Issue #1056, opened 2026-07-28
[Bug] Release changelog auto-closes nothing: comma-list Closes is inert, and unscoped subjects emit PR numbers as issue refs


atlas-apex:
## Driver

The v5.3.0 release (#1053 / PR #1054) auto-closed **zero** issues, despite shipping a `### Closes` section naming 18 references. Five issues fixed in that release — #1026, #1032, #1038, #1039, #1042 — had to be closed by hand afterwards. The issues in the `Closes` list that *do* show closed were already closed on 2026-07-26 by their own dev PRs, not by the release.

Two independent defects in `bin/release-changelog.sh` combine to produce this. Both are invisible at release time: the changelog looks correct, the PR merges, and the tickets silently stay open. The failure surfaces days later as a backlog that reads as unfinished work — which is precisely the drift `.claude/rules/reconcile-before-build.md` warns makes an OPEN ticket an unreliable "undone" signal.

## Scope

Both defects are in `bin/release-changelog.sh`.

### Defect 1 — the `Closes` comma list is inert past the first reference

Line ~294 emits a single line of the form:

```
- Closes #999, #1000, #1005, #1007, #1008, ...
```

GitHub only honours the reference **immediately following** a closing keyword. `#1000` onward are parsed as plain references, not closing directives, so they never auto-close. Every release since this format was introduced has been silently closing at most one issue.

The fix is to repeat the keyword per reference — `Closes #999, Closes #1000, Closes #1005, …` — or emit one `- Closes #N` bullet per line.

### Defect 2 — `extract_pr_num` yields a PR number when the subject has no `type(#N):` scope

`extract_pr_num` (line ~135) returns the **first** `#N` found in the subject:

```bash
echo "$subject" | grep -oE '#[0-9]+' | head -1
```

That is correct when the subject carries a conventional-commit scope, because the scope holds the issue:

| Subject | First `#N` | Correct? |
|---|---|---|
| `fix(#1042): put the human gate on approval … (#1043)` | `#1042` | yes — issue |
| `docs: right-size-ceremony cited a premium-only hook … (#1045)` | `#1045` | **no — PR; the issue is #1044** |
| `fix: normalize Windows drive-letter paths … (#1029)` | `#1029` | **no — PR; the issue is #1018** |

[...]

PR #1082, opened 2026-07-30
fix(#1077): derive release-changelog's PR lookup repo from the actual remote

atlas-apex:
## Summary

- **Repo is derived, not hardcoded** — `resolve_issue_from_pr()` in `bin/release-changelog.sh` used to always query `me2resh/apexyard` when resolving an unscoped commit's trailing PR number back to the issue it closes. Since apexyard is built to be forked, every fork cutting its own release resolved its own PR numbers against upstream's numbering, silently emitting whatever unrelated issue upstream's PR of that number happened to close. #1074 made every emitted ref a live `Closes` directive, so this could reach into an arbitrary local issue on a fork's release cut.
- **The governing invariant: resolve against where the commits came from, not "the fork's repo".** The repo is derived from `REPO_REMOTE`'s own git remote URL, and `REPO_REMOTE` itself is derived from `HEAD_REF`'s own remote prefix ("upstream/dev" → `upstream`, "origin/dev" → `origin`) when that prefix names a remote that actually exists. `HEAD_REF` is already a required input and, by convention, names the ref this release's commits are drawn from — so whichever remote it's qualified with IS the repo those trailing PR numbers actually belong to. Framing it this way (rather than "point the lookup at the fork's repo") is what makes the fix self-consistent: cutting from `upstream/dev` means the PR numbers really are upstream's, so resolving against `me2resh/apexyard` from that branch is *correct*, not a special case to route around — it falls out of the same rule an adopter fork's `origin/dev` does.
- **Zero-config (added after first review round)** — the first round of this fix left `REPO_REMOTE` defaulting unconditionally to `upstream`, which still reproduced #1077 for exactly the population it was filed about: any fork that followed apexyard's own docs (an `upstream` remote added for `/update`) resolves `upstream` to `me2resh/apexyard` regardless of which branch is actually being released — byte-identical to the hardcode this PR removes. Only a fork with no `upstream` remote at all was protected by the first round. Deriving `REPO_REMOTE` from `HEAD_REF` fixes both populations (contributor fork reading `upstream/dev`, adopter fork reading its own `origin/dev`) with **no configuration required**. An explicit `REPO_REMOTE` still overrides either derivation.
- **Missing over wrong** — when neither derivation resolves to an owner/repo shape at all, the lookup is skipped entirely (same fail-soft path as "no `gh` on PATH") rather than falling back to a hardcoded guess. A missing `Closes` line is an annoyance fixed by hand; a wrong one silently reaches into an unrelated issue.
- **The one intentional exception, documented rather than changed:** a bare `HEAD_REF` (no remote prefix at all — a plain branch name, `HEAD`, or a raw SHA) falls back to guessing `upstream`, rather than skipping like every other path in `resolve_issue_from_pr()`. This is unreachable through `/release` (which always passes a remote-qualified `HEAD_REF`) — it only bites hand-invocation. Left as a guess, not a skip, because hand-invoking this script directly inside the very repo being released (this contributor fork included) is a real, common, and *correct* use of that fallback; making it skip would regress that working case to guard a narrow misuse scenario that already has a one-argument fix (pass a qualified `HEAD_REF`, or set `REPO_REMOTE` explicitly). Documented in the script's own header.
- **`gh pr view` is now bounded by a timeout** (folded in from #1078, closed) — the call previously had no timeout at all. Verified live: it hung with no output and no error until an external `timeout` killed it at 12s. With 77 of the last 800 commits each making a serial call, one unresponsive request could stall a release cut indefinitely. Now bounded by `PR_LOOKUP_TIMEOUT` (default 10s), preferring GNU `timeout` then macOS `gtimeout`. The wrapper is resolved once per script run (not once per unscoped commit), and when neither binary is on PATH — stock macOS, notably — it now prints a one-time stderr warning instead of silently degrading to unbounded, since that's the one failure mode an operator can least guess the cause of by symptom alone.
- **Header comment corrected** — it previously (incorrectly) claimed `REPO_REMOTE` was "used for two things: as the base of the git log range… and as the source `PR_LOOKUP_REPO` derives from." It has no effect on the log range at all (that's `PREV_TAG`/`HEAD_REF` alone) — its only effect is which repo the PR lookup queries. Corrected so the doc doesn't undersell the one knob that controls issue resolution.
- **`release/SKILL.md` refreshed (was stale after round 2)** — its own `#1077` paragraph still told adopters to set `REPO_REMOTE`, the knob that stopped being the operative one once `HEAD_REF`-derivation landed. Left as-is, that would have pointed adopters at a superseded knob — partly recreating the doc-mediated discoverability failure this fix exists to close. Rewritten to name `HEAD_REF` as what actually determines the lookup repo.
- **Test harness fix along the way** — `contains()`/`not_contains()` in `test_release_changelog.sh` passed the needle straight to `grep -F`, so a needle starting with `-` (e.g. `--repo foo`, needed to assert on the derived repo) was misread as a grep option instead of a literal pattern. Fixed with `--` before the pattern.
- Audited the rest of the release path for the same hardcoding class — the other `me2resh/apexyard` occurrences in `bin/*.sh` are either issue-reference comments or `extract-subpacks.sh`'s `upstream_repo` field (correctly always the framework's own identity, unrelated to which fork runs the extraction). `release/SKILL.md`'s own `--repo me2resh/apexyard` PR-create/merge examples describe this repo's actual contributor workflow (push to upstream, PR against `me2resh/apexyard`) and are intentional, not this bug class.

## Testing

1. `bash .claude/hooks/tests/test_release_changelog.sh` — 85/85 pass. 4 scenarios / 14 assertions covering: a fork remote resolving to the fork's own repo via an explicit `REPO_REMOTE` override, the **default-config adopter fork case** (origin = fork, upstream = framework, `HEAD_REF=origin/dev`, no `REPO_REMOTE` override at all — the exact topology the ticket and the review are about), an underivable remote emitting no `gh` call and no wrong-repo guess, and a hung `gh` call bounded by the timeout instead of blocking.
2. Confirmed RED first, twice: once reverting `bin/release-changelog.sh` to its pre-PR state (6 failures: hardcoded `--repo me2resh/apexyard`, a same-number cross-repo `Closes #9999` on an underivable remote, a 30s unbounded hang), and again — after review — reverting only to the first-round commit to prove the default-config adopter-fork case was still broken (2 failures: still queried `--repo me2resh/apexyard` despite `origin/dev` naming the fork). Both times, restoring the fix returned the suite to green.
3. `bash .claude/hooks/tests/test_release_sync.sh` — 14/14 pass, unaffected (file untouched).
4. `bash bin/run-hook-tests.sh` — `PASS=123 FAIL=0 SKIP=0 TOTAL=123`, run in full three times across all three rounds of this PR; `test_release_changelog.sh` and `test_release_sync.sh` both explicitly listed as `PASS` each time.
5. `shellcheck --severity=warning bin/release-changelog.sh .claude/hooks/tests/test_release_changelog.sh` — clean (exit 0).
6. The round-3 doc/comment-only changes (this update) touch no executable line in `bin/release-changelog.sh` — confirmed via `git diff` that every changed line is a `#` comment — so no new test was needed; the full suite was re-run anyway for confirmation.

Refs #1077

## Glossary

| Term | Definition |
|------|------------|
| `PR_LOOKUP_REPO` | The `owner/repo` `resolve_issue_from_pr()` queries via `gh pr view` when turning an unscoped commit's trailing PR number into the issue it actually closes |
| `REPO_REMOTE` | The git remote name whose repo `PR_LOOKUP_REPO` derives from — its only effect. Now itself derived from `HEAD_REF`'s remote-qualified prefix (e.g. "origin/dev" → `origin`), falling back to `upstream` only for a bare `HEAD_REF` |
| default-config adopter fork | A fork with both `origin` (its own repo) and `upstream` (`me2resh/apexyard`, added per apexyard's own docs for `/update`) configured, cutting its own independent release — the exact topology #1077 was filed against |
| fail-soft | Degrading to a safe fallback instead of aborting or guessing, on any failure of an optional lookup |


atlas-apex (commented):
## Code Review: PR #1082

**Commit**: `f70f79de087633b68490a7b03a5901baebcccc7a`

### Summary

Replaces `resolve_issue_from_pr()`'s hardcoded `PR_LOOKUP_REPO=me2resh/apexyard` with a value derived from `REPO_REMOTE`'s git remote URL, skips the lookup entirely when the remote can't be resolved, and bounds the `gh pr view` call with `timeout`/`gtimeout`. Folds in #1078. Adds 3 test scenarios (13 assertions) and fixes a latent `grep -F` needle bug in the test harness.

The engineering is careful and the fail-soft posture is right. But the fix does not change default behaviour for the population #1077 describes, and the new header comment states a checkably-false fact. Both are small and in-scope.

### Checklist Results
- Code Quality:              **Fail** (one factually incorrect header claim)
- Technical Decisions (AgDR): N/A — bug fix within an existing mechanism, reversible in this PR
- Adopter Handbooks:         N/A — shell/markdown diff, no handbook triggers

---

### Issues Found

#### 1. BLOCKING — a fork with default config still resolves against upstream. #1077's reported case is unchanged.

`REPO_REMOTE` defaults to `upstream` (line 104). In an adopter fork, `upstream` **is** `me2resh/apexyard`. Verified directly in this worktree, which has exactly the adopter topology:

```
REPO_REMOTE=upstream  -> me2resh/apexyard      <- byte-identical to the removed hardcode
REPO_REMOTE=origin    -> atlas-apex/apexyard
```

So the population splits:

| Fork shape | Derived `PR_LOOKUP_REPO` | vs. pre-fix |
|---|---|---|
| Has an `upstream` remote (the **documented** apexyard setup — `/update` requires it) | `me2resh/apexyard` | **identical — bug unchanged** |
| No `upstream` remote | empty → lookup skipped | improved (fail-soft) |

That inverts the desirable outcome: adopters who followed apexyard's own remote-setup docs keep emitting wrong live `Closes` lines into their own tracker; only adopters who *didn't* set up `upstream` are protected. The PR body's own remedy — "adopters cutting an independent release set `REPO_REMOTE`" — is configuration they have no reason to discover, for a failure that is silent by construction.

I weighed the counter-argument that `upstream` is genuinely correct for *this* repo. It is. But that argument doesn't defend the current default, because a better default satisfies it too.

**Suggested fix — derive the remote from `HEAD_REF`**, which is already a *required* input and literally names the ref the release is cut from:

```sh
if [ -z "${REPO_REMOTE:-}" ]; then
  candidate="${HEAD_REF%%/*}"
  if [ "$candidate" != "$HEAD_REF" ] && git remote get-url "$candidate" >/dev/null 2>&1; then
    REPO_REMOTE="$candidate"
  else
    REPO_REMOTE="upstream"
  fi
fi
```

`HEAD_REF=upstream/dev` (this repo) → `upstream`, unchanged. `HEAD_REF=origin/dev` (a fork) → `origin`, correct with **zero configuration**. A bare `HEAD_REF` (`HEAD`, `dev`) falls back to today's behaviour. ~6 lines plus one test, derived from data the script already demands.

Mitigating: the body says `Refs #1077`, not `Closes`, so the ticket won't auto-close on merge. Please keep it open if this ships as-is.

#### 2. BLOCKING (trivial) — the new header comment states something false.

`bin/release-changelog.sh:15-19`:

> `REPO_REMOTE` — ... Used for two things: as the base of the git log range (see below), and — since #1077 — as the source `PR_LOOKUP_REPO` derives from.

`REPO_REMOTE` appears in this script at exactly two places: its own assignment (line 104) and `_derive_pr_lookup_repo` (line 227). It has **no** effect on the log range — that is built from `PREV_TAG`, `HEAD_REF`, and the `Released-From` trailer. Since this PR, deriving `PR_LOOKUP_REPO` is `REPO_REMOTE`'s *only* function. Worth correcting, and it isn't cosmetic: it's part of why "set `REPO_REMOTE=origin`" reads like ordinary release configuration rather than the single knob that controls issue resolution.

---

### Findings by priority

**Timeout reach (advisory, not blocking).** The `timeout`/`gtimeout`/unbounded ladder is correct in shape — the unbounded branch restores exact pre-#1078 behaviour, so it is never *worse* than today. But stock macOS ships neither binary (`gtimeout` needs brew coreutils), so the default macOS adopter silently gets zero protection, against the one failure mode where the operator is least able to guess the cause (silence, no error, no output). The script already writes to stderr (`RELEASE_CHANGELOG_RANGE=`); one line there when neither binary is found would make the degradation visible at ~zero cost. Suggestion, not a blocker.

**Fail-soft on underivable remote — confirmed correct.** `resolve_issue_from_pr()` returns early on empty `PR_LOOKUP_REPO` before any `gh` invocation, `echo ""` + `return 0` — structurally identical to the four existing degradation paths (no `gh`, non-zero exit, no ref in body, hang). The test asserts the stub was *never called* (`eq "" "$call_log_contents"`), which is a stronger assertion than checking output, and separately that `Closes #9999` is not emitted. Matches "prefer a missing close over a wrong close". No note.

**Harness `--` fix — no pre-existing assertion changed meaning.** I checked every `contains`/`not_contains`/`eq` needle in the file: the only two starting with `-` are the new lines 556–557 (`--repo ...`). So nothing was vacuously passing before and nothing was silently broken — the `--` is a forward-looking correctness fix the new assertions *required*, not a repair. Clean, and no hidden regression surfaced.

**Tests and counts.** Verified on this branch: `test_release_changelog.sh` → **82/82 pass**, `test_release_sync.sh` → **14 passed / 0 failed**. The commit deletes no tests (the diff is additive; the only edits to existing cases add `git remote add upstream ...` to the three #1056 scenarios, which is *necessary* — without a remote the derivation now yields empty and those tests would exercise the skip path instead of resolution). I did not run the full `run-hook-tests.sh` aggregate, so the `123/123` figure is unverified by me rather than disputed.

nit: the body and commit both say "6 new cases"; the diff adds **3 scenarios / 13 assertions**. The RED-proof count of 6 failures is plausible but the "6 cases" framing doesn't map to anything in the file.

**The `release/SKILL.md` audit call — agreed.** Leaving the PR-create examples pointing at `me2resh/apexyard` is right: those describe the *contributor* workflow (open a PR against the framework), where the target genuinely is upstream regardless of who runs it. Same reasoning correctly spares `extract-subpacks.sh`'s `upstream_repo` identity field. The hardcode that mattered was the one silently used as a *query target*; these are neither.

**PR body, glossary, commit message.** All good. Narrative bullets with the why, glossary defines the five terms a cold reader needs, `Refs #1077` present, commit message follows the convention with a real body. Meets `handbooks/general/commit-message-quality.md`.

### Suggestions
- Emit one stderr line when neither `timeout` nor `gtimeout` is present, so the unbounded degradation is visible.
- Consider validating `PR_LOOKUP_TIMEOUT` is numeric — a typo currently makes `timeout` error out and silently degrade to no lookup at all.

### Verdict
**CHANGES REQUESTED**

Two narrow asks: derive the default remote from `HEAD_REF` (issue 1) and correct the header's log-range claim (issue 2). The timeout work, the fail-soft path, the harness fix, and the tests are all good as they stand — this is not a rework, it's the last two lines of the same fix.

---
📌 Reviewed commit: `f70f79de087633b68490a7b03a5901baebcccc7a`

atlas-apex (commented):
## Code Review: PR #1082 (round 2)

**Commit**: `204d9704eb5222cb8a73d9cc8d23e67a63118984`

### Summary

Round 1's two blockers are both fixed, and I re-measured rather than assumed. `REPO_REMOTE` now derives from `HEAD_REF`'s remote prefix when that prefix names an existing remote; the header's false log-range claim is gone; the timeout wrapper resolves once per run and warns once when neither binary is present.

One thing worth saying explicitly: the implementation is **better-principled than what I suggested**. I framed it as "point the lookup at the fork's own repo." What shipped ties the lookup repo to *where the commits came from* — which is the actually-correct invariant. If you cut from `upstream/dev`, the trailing PR numbers in those commits **are** upstream's, so resolving them against `me2resh/apexyard` is right, not a bug. That reframing makes the fix self-consistent in a way my version wasn't.

### Checklist Results
- Performance:                Pass (timeout hoisted out of the per-commit loop)
- Summary Bullet Narrative:   Pass (advisory)
- Adopter Handbooks:          N/A — shell/markdown diff

---

### 1. Both populations, measured under default config

Executed the **shipped** derivation block verbatim (extracted from `204d970`) in a worktree with real adopter topology — `origin` = a fork, `upstream` = `me2resh/apexyard`, plus a third remote:

| `HEAD_REF` | → `REPO_REMOTE` | → repo actually queried |
|---|---|---|
| `upstream/dev` | `upstream` | `me2resh/apexyard` — correct, unchanged |
| `origin/dev` | `origin` | **`atlas-apex/apexyard`** — the fork. **#1077 fixed, zero config** |
| `hossam/dev` | `hossam` | `hossam-96/apexyard` — correct by definition |

Round 1's finding is resolved: the adopter case no longer collapses onto the hardcode. Confirmed by measurement, not inspection.

### 2. The bare-`HEAD_REF` fallback — still a wrong-close path (suggestion, not blocking)

Same measurement, remaining inputs:

| `HEAD_REF` | → repo queried |
|---|---|
| `dev` (bare local branch) | `me2resh/apexyard` |
| `HEAD` | `me2resh/apexyard` |
| `refs/heads/dev` | `me2resh/apexyard` |
| `release/v2` | `me2resh/apexyard` |
| a raw 40-char SHA | `me2resh/apexyard` |

So yes — **the trap re-opens in narrower form.** In a fork, every one of those lands on `me2resh/apexyard`, which is precisely #1077's defect. `refs/heads/dev` and a bare local branch name aren't exotic inputs; they're the natural thing to type when hand-invoking the script.

The sharper objection is internal consistency. This script's stated governing principle is *"prefer a missing close over a wrong close"*, and it honours that everywhere else — an unresolvable remote leaves `PR_LOOKUP_REPO` empty and skips the lookup entirely. But in the one case where it has **demonstrably no information** about which repo the release targets, it stops refusing to guess and guesses `upstream`. That is the only branch in the function that guesses.

I weighed the defense that this preserves pre-#1077 behaviour and is correct for this repo. It's weaker than it looks: `/release` always passes `upstream/dev` (SKILL.md line 68), so the fallback is **never on the path for the framework's own cut**. It's load-bearing only for hand-invocation — the exact case where the operator is least likely to have thought about it.

**Suggested follow-up:** when `HEAD_REF` has no resolvable remote prefix, leave `REPO_REMOTE` empty so `PR_LOOKUP_REPO` ends up empty and the lookup skips, instead of falling back to `upstream`. Cost to the operator is a missing `Closes` line they add by hand; benefit is that the last wrong-close path disappears and the function stops contradicting its own header.

Not blocking: unreachable through the shipped caller, and the change is a strict improvement over both `dev` and round 1 regardless.

### 3. Existence check — can it silently pick the wrong remote?

Worked through the cases you raised:

- **`upstream` pointing at the fork itself** — derives the fork's repo. That's *correct*: the commits came from that remote, so its PR numbers are the right ones. Not a misfire.
- **A third remote sharing a prefix** (`hossam/dev`) — resolves to that remote, which is correct by definition; you asked to cut from it.
- **A branch name containing `/`** (`release/v2`) — `release` isn't a remote, so the existence check rejects it and it falls back. The guard works as documented.

The only way derivation picks a remote that isn't the release target is if `HEAD_REF` itself names the wrong remote — but then the **log range is also wrong** and the changelog is already garbage upstream of this code. The derivation can't be wrong in isolation. No finding.

### 4. Counts — verified independently, not accepted

Given counts have been corrected twice, and heeding the note about false reds from dirty reviewer trees: I first confirmed `git diff FETCH_HEAD --stat` was **empty** and `git status` showed no tracked modifications, so the tree I tested was byte-identical to `204d970` — then ran everything.

| Suite | Claimed | Measured |
|---|---|---|
| `test_release_changelog.sh` | 85/85 | **85/85 pass** |
| `test_release_sync.sh` | 14/14 | **14 passed / 0 failed** |
| `bin/run-hook-tests.sh` | `PASS=123 FAIL=0` | **`PASS=123 FAIL=0 SKIP=0 TOTAL=123`** |

All three match. No aggregate-only reds appeared, so nothing to quarantine or re-run standalone. The self-corrected "4 scenarios / 14 assertions" matches the file. Counts are accurate this time.

The new test's use of `git update-ref` to synthesise `refs/remotes/origin/dev` is the right call — it makes the fork scenario hermetic with no network.

### Suggestions
- Close the bare-`HEAD_REF` guess (§2) — skip rather than fall back to `upstream`.
- **`.claude/skills/release/SKILL.md` is now stale and contradicts the script.** It was untouched in round 2, so line 91 still says the repo is "derived from `REPO_REMOTE`'s own git remote URL (`REPO_REMOTE` defaults to `upstream`)" and tells adopters to "set `REPO_REMOTE`". Neither is true now — `REPO_REMOTE` no longer defaults to `upstream`, it derives from `HEAD_REF`, and the operative knob for an adopter cutting their own release is **`HEAD_REF`** (line 68 pins it to `upstream/dev`). The stale advice is *safe* — an explicit `REPO_REMOTE` still overrides — which is why this isn't blocking. But #1077's failure mode is silent and doc-mediated, so shipping a caller doc that points at a superseded knob partly recreates the discoverability problem the fix exists to solve. Worth a short follow-up edit.

### Verdict
**APPROVED**

Both round-1 blockers verified fixed by direct measurement, the design improved on the suggestion it came from, and all three test counts independently reproduce. The two items above are follow-ups, not merge blockers.

---
📌 Reviewed commit: `204d9704eb5222cb8a73d9cc8d23e67a63118984`

atlas-apex (commented):
## Code Review: PR #1082 (round 3 — doc-only delta)

**Commit**: `582e929062f1469b9e82c8e508d619b85b662b8b`

### Summary

Short pass, as billed. The delta from `204d970` is comment/doc-only, and I confirmed that independently rather than accepting it. My round-2 stale-doc finding is fixed. My round-2 bare-`HEAD_REF` objection was declined with reasoning — **I accept the decline** and withdraw the ask, with one precision correction and one residual suggestion below.

### Checklist Results
- Architecture & Design:      Pass (unchanged — executably identical tree)
- Summary Bullet Narrative:   Pass (advisory)
- Fallow Static Analysis:     N/A

---

### 1. The delta is executably inert — confirmed two independent ways

**Proof A — no non-comment line moved.** Every added/removed line in the `.sh` hunks, after excluding `+++`/`---` headers, comment lines, and blanks: **empty set**.

**Proof B — comment-stripped content is byte-identical.** Stripping full-line comments and blanks from both revisions of `bin/release-changelog.sh` yields **209 executable lines, md5 `afbdca7a6012a89b337f7ee1692975c4`, identical across `204d970` and `582e929`.**

So the executable tree I approved at `204d970` is the executable tree here. The other changed file is `SKILL.md` — pure prose.

**On re-running the aggregate.** I ran `PASS=123 FAIL=0 SKIP=0 TOTAL=123` myself last round on `204d970`. Given the executable content is byte-identical by md5, and `grep -rl "release/SKILL.md"` across `.claude/hooks/tests/` and `.claude/skills/*/tests/` returns **nothing** (no test asserts on the changed doc), a fourth full run is provably redundant — there is no path by which the result could differ. I did re-run the two named suites as a sanity check on a tree verified equal to the PR HEAD: `test_release_changelog.sh` **85/85**, `test_release_sync.sh` **14 passed / 0 failed**. I'm flagging the inference explicitly so it isn't mistaken for a fresh aggregate measurement.

### 2. Stale-doc finding — fixed

`SKILL.md`'s #1077 paragraph now leads with `HEAD_REF` as the determining knob, states the invariant plainly ("resolve against wherever the commits actually came from"), and demotes `REPO_REMOTE` to explicit override / bare-ref fallback. It also correctly notes `upstream/dev` resolving to `me2resh/apexyard` is *correct, not a bug*. The contradiction between caller doc and script is gone.

### 3. Bare-`HEAD_REF` fallback — the reasoning holds. I withdraw the ask.

The decline is correct and I'm not asking for the change. Two things led me there:

**The core argument is sound.** Skipping would trade a working case for a guard against a narrow misuse that already has a one-argument fix. That's the right trade. I'd add the point that pushed me over: the harm case requires *four* things to coincide — hand-invocation (not `/release`), a bare `HEAD_REF`, the fork's **own** commits in range, and `upstream` configured to point elsewhere — **and** the affected commit must be unscoped (`resolve_issue_from_pr()` is only consulted for commits with no `type(#N):` scope). Most commits here are scoped. That's a much smaller target than my round-2 framing implied.

**One precision correction to the stated rationale, which doesn't change the conclusion.** The defense says the fallback protects "hand-invoking the script inside the repo actually being released — where `upstream` genuinely is correct." In the release repo *proper* — a clone of `me2resh/apexyard` where `origin` is the release target — there typically **is no `upstream` remote**, so `_derive_pr_lookup_repo` fails, `PR_LOOKUP_REPO` is empty, and the lookup **already skips**. The case the fallback genuinely protects is narrower and more specific: a **contributor fork with `upstream` configured**, hand-invoked with a bare ref against synced commits — where `upstream` is correct because the commits really did come from there. That's a real case (it's exactly what the test fixtures construct), so the conclusion stands; only the framing was slightly wide.

**Residual: the defect is the silence, not the fallback.** In the one situation where it does misfire, the operator gets a wrong live `Closes` with no indication a guess was made. That's the same silence-shaped defect class as #1077 itself. My suggestion — non-blocking, follow-up — is a one-time stderr line when falling back on an unqualified `HEAD_REF`, e.g.:

```
NOTE: HEAD_REF="HEAD" is not remote-qualified — resolving PR numbers against
<repo> via REPO_REMOTE. Set HEAD_REF=<remote>/<branch> or PR_LOOKUP_REPO to override.
```

That keeps the working case intact, costs nothing, and turns a silent guess into a visible one. It's the same shape as the missing-`timeout` warning already accepted in round 2, and it makes the "one intentional exception" self-announcing at runtime rather than only in the PR body.

Calling it out explicitly in the PR body as the one intentional exception, rather than burying it, is the right handling regardless.

### Suggestions
- One-time stderr note on the bare-`HEAD_REF` fallback (§3) — follow-up, not a blocker.

### Verdict
**APPROVED**

Delta proven executably inert against the tree I approved at `204d970`, stale-doc finding fixed, and the one open objection resolved on the merits in the author's favour.

---
📌 Reviewed commit: `582e929062f1469b9e82c8e508d619b85b662b8b`

me2resh (approved):

Issue #1086, opened 2026-07-30
[Chore] Install git pre-push hooks by default, then move protected-branch enforcement into them


atlas-apex:
## Driver

Nine of the twelve genuinely-open framework bugs are one mechanism: a hook deciding whether a command pushes to a protected branch by pattern-matching the command's **text**. Five of those nine are in `block-main-push.sh` alone (#1079, #1081, #1083, #1084, #1085), and #1066 is the shared-parser sibling.

[AgDR-0104](../docs/agdr/AgDR-0104-trust-chain-controls-vs-backstops.md) already concluded — eight days before these were filed — that shell hooks pattern-matching bash command text cannot be trusted to enforce security-critical principles. That decision was recorded but never implemented, so each fix in this class has generated the next round's finding. PR #1075 ran nine review rounds on exactly this.

**There is ground truth available and the framework isn't reading it.** Git's `pre-push` hook receives the refs being pushed on stdin, one line each:

```
<local_ref> <local_sha> <remote_ref> <remote_sha>
```

That is git's own resolution of the refspec. `refs/heads/main` arrives as a literal string. Every bug shape in this class — heredoc bodies, decoy refs, quoted refs, `&&` chains, a redirect or pipe before the push, bare `HEAD`, a tag-push prefix — exists only because there is command text to misparse. At the git layer there is no text.

`.githooks/pre-push` already exists and is tracked. It runs the pre-push check set and delegates to `bin/run-pre-push-checks.sh` — but it discards stdin entirely.

### Why not branch protection

Considered and rejected: adopters install this framework onto arbitrary repos and forges. Server-side required-PR cannot be assumed to exist, and is GitHub-specific. The git hook is the portable equivalent — it works on any forge, on any plan tier, with no server-side configuration.

### The finding that sets the ordering

**Nothing installs the git hook today.** `core.hooksPath` appears only in `docs/getting-started.md` as a manual one-liner. Neither `/setup` nor `/handover` nor any project scaffold sets it, and empirically it does not get done — at least one fork inherits `.githooks/` with `core.hooksPath` unset, and another has it pointing at a stale absolute path from a prior clone location, leaving the tracked hook inert.

Underneath that is a wider gap: **managed-project clones carry no framework presence in their own repo** — no `.claude/`, no `.githooks/`. The Claude Code `PreToolUse` hooks fire only because the *agent* runs in the ops-fork context. So today:

| Push path | Protected? |
|---|---|
[...]

PR #1087, opened 2026-07-30
chore(#1086): install git pre-push hooks by default

atlas-apex:
## Summary

- **Ships `bin/install-git-hooks.sh`, an idempotent per-clone installer for `core.hooksPath`** — `.githooks/pre-push` has existed and been tracked for a while, but nothing ever set the git config that makes it run. It's a per-clone value (lives in `.git/config`, never committed), so every fresh clone starts unset regardless of how many sibling clones already have it configured. The installer classifies the current state — unset, already correct, pointing at a deleted directory, or a stale leftover pointing into a *different* clone's `.git/hooks` (a real observed failure this ticket cites) — and auto-repairs the first three. A real, different directory the adopter configured on purpose is left untouched unless `--force` is passed, so this never silently clobbers a deliberate choice.
- **Adds a shared classification lib (`_lib-git-hooks-path.sh`)** so the installer and the advisory hook can't drift on what counts as "stale" vs "deliberate third-party" — both read the exact same state machine.
- **Adds a `SessionStart` advisory (`check-git-hooks-installed.sh`)** — same non-blocking shape as `check-upstream-drift.sh` (exit 0 always) — that warns when a clone's `core.hooksPath` is unset or stale, so a contributor finds out their terminal `git push` is unprotected instead of discovering it after the fact.
- **Wires the installer into `/setup` only** — the ops fork configuring its own `.githooks/`. `/handover` deliberately does **not** call it (see the security-fix bullet below).
- **Updates `docs/getting-started.md`'s "Terminal push hook" section** to describe the new automatic path instead of the old manual-only `git config core.hooksPath .githooks` instruction (kept as the fallback for anyone who skips `/setup`).

This is **step 1 of 3** from #1086. Steps 2 (moving protected-branch enforcement into `.githooks/pre-push` using git's own resolved ref data instead of command-text parsing) and 3 (demoting `block-main-push.sh` to advisory) are **explicitly out of scope here and must not land before this PR**. Nothing in this PR touches `.githooks/pre-push`'s check logic or `block-main-push.sh` at all.

### Review round 2 — every finding fixed in this batch

Both Rex and Hakim returned CHANGES REQUESTED on the first commit. All findings are addressed here in one pass (no incremental re-request, per the #1075 lesson):

- **[HIGH, Hakim] Dropped the `/handover` wiring entirely.** The installer gates only on "does a `.githooks/` directory exist" — it has no provenance check. Calling it against a freshly-cloned *third-party* repo doesn't install anything ApexYard-provided; it points git at *that repo's own* committed scripts, with no operator confirmation. Hakim proved this hands the clone code execution on ordinary git operations (`post-checkout`, `reference-transaction` — no push required), because git deliberately never clones `$GIT_DIR/hooks`, and pointing `core.hooksPath` at a tracked directory removes that safety property. `/handover`'s whole point is evaluating a repo before deeper commitment — exactly the wrong moment to hand it execution. `/setup` (the ops fork configuring its own repo) is unaffected and stays wired up. The gap this leaves — a managed-project clone's own git hooks still aren't installed by anything — is real and **deliberately deferred, not silently dropped**: `.claude/skills/handover/SKILL.md` now carries an explicit note that the correct fix is a *different* mechanism (install ApexYard's own hook into the clone's untracked `.git/hooks/`, never point at a tracked third-party directory), which needs its own design and its own ticket.
- **[MEDIUM, Hakim] The installer now fails CLOSED on a `git config` write that doesn't take.** None of the four writes checked their exit status before — a stale `.git/config.lock` (routine with several parallel worktree agents) or a multi-valued `core.hooksPath` made the script print its success message and exit 0 while the value stayed unchanged. `_ghp_set_hookspath()` now checks the write's exit status AND re-classifies afterward to confirm the state actually landed on `correct`; either check failing exits **4** with a loud error and zero false-success output. Reproduced for real with an actual `.git/config.lock` (not a mock) as part of verification — see Testing.
- **[MEDIUM, Hakim + suggestion from Rex, same root cause] `--repo-dir` no longer silently retargets to an enclosing repo.** `git rev-parse --show-toplevel` walks *up* when the given directory isn't a repo's own root — so `--repo-dir workspace/proj` on a failed or partial clone previously resolved to (and configured) the *ops fork*, while reporting success as if it had configured the named project. The installer now asserts the resolved toplevel equals the resolved `--repo-dir` and exits 2 loudly otherwise — the wrong-target-write class `.claude/rules/isolated-builds.md` exists to prevent.
- **[Rex finding #4] Retired the third copy of `_resolve_real_path`.** The PR's own comment admitted this was "copied verbatim... rather than sourcing it directly" from `require-active-ticket.sh` — precisely the AgDR-0113 pattern (an agreement asserted in a comment, not enforced by sharing) this PR's own arc exists to retire, in the very PR making that argument. There is now exactly **one** `_resolve_real_path`, at `.claude/hooks/_lib-path-resolve.sh`; `require-active-ticket.sh` and `_lib-git-hooks-path.sh` both source it. (This is explicitly *not* the same thing as `_lib-portfolio-paths.sh`'s `_portfolio_canonicalize`, a deliberately different fail-soft sibling with Windows drive-letter handling — the lib's header comment says so directly, so a future reader doesn't try to fold the two together.)
- **[Rex finding #5, three mutation-proven gaps] Added the three missing discriminating tests.** All three were verified by actually re-applying the mutant and confirming the suite goes red — see Testing:
  - A symlinked repo path resolving to the same real `.githooks` dir must classify `correct` (idempotent), not `third-party` — the case that pins `_resolve_real_path`'s actual behaviour rather than its mere presence.
  - The advisory hook's `third-party:*` banner now has coverage (previously zero — deleting the branch outright still passed 6/6).
  - The classifier's *external* third-party fallthrough (a real directory outside the repo, no `.git` component — the canonical "shared company hooks dir" shape) now has coverage on the installer side. This is the branch guarding the PR's central safety guarantee (never clobber a deliberate value without `--force`); it had zero tests before this round.
- **[LOW, Hakim] `--hooks-dir` traversal outside the repo is now rejected** (`--hooks-dir ../evil-hooks` used to write verbatim); **a `.githooks` committed as a symlink is now refused** (`[ -L ]` checked before `[ -d ]`, which previously followed it silently); **the advisory banner no longer suggests `--force`** on the third-party state — a recurring nudge toward overriding a deliberate choice was the wrong shape for a guard whose whole job is protecting that choice. Points at `--help` instead.

Three Rex notes were reviewed and intentionally left as-is (stated here rather than resolved silently): fail-empty vs. fail-soft path resolution is a documented, deliberate difference from `_portfolio_canonicalize` (see the lib header); a `git config` read failure degrading to `unset` is already the fail-safe direction; pointing `core.hooksPath` at *this* repo's own `.git/hooks` classifies as `third-party` (refuse) rather than `foreign-git-dir` by design, per the lib's own doc comment — all three are Low/informational, not requested fixes.

## Testing

- Full suite chunked (3 × ~43 tests, ≤90s/test cap): **PASS=129 FAIL=0** across all three chunks, zero regressions. `require-active-ticket.sh`'s own suite (`test_require_active_ticket_bash.sh`, updated to carry the new shared lib into its sandboxes) independently reruns clean: **PASS=78 FAIL=0**.
- `test_install_git_hooks.sh`: **41 assertions**, up from 23 — added cases for the symlinked-repo idempotent state, the external-third-party refusal, `--repo-dir` retargeting, `--hooks-dir` traversal, a symlinked hooks dir, and the real `.git/config.lock` fail-closed path.
- `test_check_git_hooks_installed.sh`: **8 assertions**, up from 6 — added the third-party banner case (asserts both that it fires and that it does *not* mention `--force`).
- **All three review-cited mutants re-applied and confirmed killed, not just described:**
  - `_resolve_real_path` → bare passthrough: installer suite **16/41 pass, 25 fail**; restored → 41/41.
  - Advisory hook's `third-party:*` branch deleted: advisory suite **7/8 pass, 1 fail**; restored → 8/8.
  - Classifier's external-third-party fallthrough → `unset`: installer suite **38/41 pass, 3 fail** + advisory suite **7/8 pass, 1 fail**; restored → both fully green.
- **MEDIUM-2 exercised for real**, not simulated: created an actual `.git/config.lock` in a scratch repo, ran the installer — `exit=4`, `error: could not lock config file .git/config: File exists` surfaced, no false-success message printed, `core.hooksPath` confirmed still unset afterward.
- **MEDIUM-3 exercised for real**: `--repo-dir <outer>/subdir-not-a-repo` against a real two-level scratch repo — refused with `exit=2`, names the enclosing repo, outer repo's config confirmed untouched.
- `shellcheck --severity=warning` on all eight touched/added shell files (including the two updated test files and the new shared lib) — clean.
- `markdownlint-cli2` across the repo — clean, including the updated `docs/getting-started.md` and `.claude/skills/handover/SKILL.md`.

## Glossary

| Term | Definition |
|------|------------|
| `core.hooksPath` | Per-clone git config naming the directory git reads hooks from. Lets a repo ship tracked hooks (like `.githooks/pre-push`), but the config must be set once per clone — cloning a repo does not inherit it. |
| `pre-push` hook | A git-native hook that runs before a push completes, given the refs being pushed on stdin already resolved by git — no command-text parsing required. |
| fail closed / fail open | Fail closed: on an error, refuse and report failure (the installer's `git config` write). Fail open: on an error, let the action through with a warning (the advisory `SessionStart` hook — it must never block a session). This PR's central fix (MEDIUM-2) was that the installer had accidentally inherited the advisory's fail-open shape. |
| `foreign-git-dir` | This PR's stale-detection signal: a `core.hooksPath` value resolving to a path with a `.git` component outside the target repo. Since a tracked hooks dir can never live inside `.git/` (never committed), this can only be a leftover pointer into a different clone's default hooks location. |
| third-party (in this PR) | `core.hooksPath` resolving to a real, existing directory that isn't the repo's own tracked dir and isn't `foreign-git-dir`-shaped — ambiguous by design, so treated as a deliberate adopter choice and never auto-repaired. |
| mutation testing (as used here) | Deliberately reintroducing a specific plausible bug (a "mutant") into already-passing code, then confirming the test suite goes red. A suite that stays green against a real mutant proves nothing — the standard this round's new tests were held to. |

Refs #1086

### DIFF
--- a/.claude/hooks/_lib-git-hooks-path.sh
+++ b/.claude/hooks/_lib-git-hooks-path.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+# _lib-git-hooks-path.sh — shared core.hooksPath detection logic.
+#
+# Sourced by:
+#   - bin/install-git-hooks.sh              (mutates core.hooksPath)
+#   - .claude/hooks/check-git-hooks-installed.sh  (advisory, read-only)
+#
+# Kept as a single shared lib so the installer and the advisory banner can
+# never drift on what counts as "unset" / "correct" / "stale" / "deliberate
+# third-party" — see me2resh/apexyard#1086.
+
+# ------------------------------------------------------------------------
+# _resolve_real_path PATH
+#
+# Resolves PATH to its canonical, symlink-free absolute form. Walks up to
+# the nearest EXISTING ancestor, physically resolves it (`pwd -P`, which
+# follows symlinks), then re-appends any not-yet-created tail literally —
+# a tail that doesn't exist yet cannot itself be a symlink. This mirrors
+# `realpath -m` without depending on GNU coreutils (not guaranteed present
+# on macOS/BSD). Echoes the resolved path, or nothing if even "/" can't be
+# stat'd (should not happen for a well-formed absolute path).
+#
+# Copied verbatim from require-active-ticket.sh's _resolve_real_path (see
+# that file for the fuller symlink-bypass rationale) rather than sourcing
+# it directly — that hook's copy carries ticket-gate-specific comments and
+# the two call sites shouldn't share a lifecycle.
+# ------------------------------------------------------------------------
+_resolve_real_path() {
+  local p="$1" dir tail=""
+  [ -n "$p" ] || return 0
+  dir="$p"
+  while [ -n "$dir" ] && [ "$dir" != "/" ] && [ ! -e "$dir" ]; do
+    if [ -z "$tail" ]; then
+      tail="$(basename "$dir")"
+    else
+      tail="$(basename "$dir")/$tail"
+    fi
+    dir="$(dirname "$dir")"
+  done
+  if [ ! -e "$dir" ]; then
+    return 0
+  fi
+  if [ -d "$dir" ]; then
+    dir="$(cd "$dir" 2>/dev/null && pwd -P)"
+  else
+    local parent
+    parent="$(cd "$(dirname "$dir")" 2>/dev/null && pwd -P)"
+    if [ -n "$parent" ]; then
+      dir="$parent/$(basename "$dir")"
+    else
+      dir=""
+    fi
+  fi
+  [ -n "$dir" ] || return 0
+  if [ -n "$tail" ]; then
+    printf '%s/%s' "$dir" "$tail"
+  else
+    printf '%s' "$dir"
+  fi
+}
+
+# ------------------------------------------------------------------------
+# ghp_classify REPO_ROOT HOOKS_DIR_NAME
+#
+# REPO_ROOT must already be an absolute, resolved git toplevel path.
+# HOOKS_DIR_NAME is the tracked hooks dir name relative to REPO_ROOT
+# (default convention: ".githooks").
+#
+# Prints exactly one of, to stdout (no trailing newline):
+#
+#   unset
+#     core.hooksPath has no value in this clone's git config.
+#
+#   correct
+#     core.hooksPath resolves exactly to REPO_ROOT/HOOKS_DIR_NAME.
+#
+#   missing:<raw>
+#     core.hooksPath is set to <raw>, which resolves to a path that does
+#     not exist on disk (a deleted directory, a typo, or a config value
+#     copied from a machine/clone that no longer has that path).
+#
+#   foreign-git-dir:<raw>:<resolved>
+#     core.hooksPath resolves to an existing directory, but that
+#     directory sits inside SOME repo's .git directory (path contains a
+#     `.git` component) and is outside REPO_ROOT. A tracked hooks dir can
+#     never legitimately live inside `.git/` (that directory is never
+#     committed), so this can only be a stale absolute path left over
+#     from a different clone's default hooks location — the concrete
+#     failure mode #1086 was filed against.
+#
+#   third-party:<raw>:<resolved>
+#     core.hooksPath resolves to an existing, real directory that is
+#     neither of the above — could be inside REPO_ROOT under a different
+#     name, or a genuinely external directory an adopter configured on
+#     purpose (e.g. a company-wide shared hooks dir). Ambiguous by
+#     design: treated as a deliberate adopter choice, never auto-repaired.
+# ------------------------------------------------------------------------
+ghp_classify() {
+  local repo_root="$1" hooks_dir_name="$2"
+  local current current_resolved target_resolved
+
+  current=$(git -C "$repo_root" config --get core.hooksPath 2>/dev/null || true)
+  target_resolved=$(_resolve_real_path "$repo_root/$hooks_dir_name")
+
+  if [ -z "$current" ]; then
+    printf 'unset'
+    return 0
+  fi
+
+  case "$current" in
+    /*) current_resolved=$(_resolve_real_path "$current") ;;
+    *)  current_resolved=$(_resolve_real_path "$repo_root/$current") ;;
+  esac
+
+  if [ -n "$current_resolved" ] && [ "$current_resolved" = "$target_resolved" ]; then
+    printf 'correct'
+    return 0
+  fi
+
+  if [ -z "$current_resolved" ] || [ ! -d "$current_resolved" ]; then
+    printf 'missing:%s' "$current"
+    return 0
+  fi
+
+  # Inside REPO_ROOT but under a different name than HOOKS_DIR_NAME — a
+  # real directory the adopter (or a script) deliberately created in this
+  # very repo. Ambiguous-by-design: third-party, not stale.
+  case "$current_resolved" in
+    "$repo_root"/*)
+      printf 'third-party:%s:%s' "$current" "$current_resolved"
+      return 0
+      ;;
+  esac
+
+  # Outside REPO_ROOT. If it sits inside a `.git` directory anywhere on
+  # its path, it cannot be a tracked hooks dir (never committed) — almost
+  # certainly a stale pointer into a different clone's default hooks
+  # location. Otherwise it's an external directory that could be a
+  # deliberate shared setup — leave it alone.
+  case "$current_resolved" in
+    */.git/*|*/.git)
+      printf 'foreign-git-dir:%s:%s' "$current" "$current_resolved"
+      return 0
+      ;;
+  esac
+
+  printf 'third-party:%s:%s' "$current" "$current_resolved"
+}
--- a/.claude/hooks/_lib-git-hooks-path.sh
+++ b/.claude/hooks/_lib-git-hooks-path.sh
@@ -9,55 +9,20 @@
 # never drift on what counts as "unset" / "correct" / "stale" / "deliberate
 # third-party" — see me2resh/apexyard#1086.
 
-# ------------------------------------------------------------------------
-# _resolve_real_path PATH
-#
-# Resolves PATH to its canonical, symlink-free absolute form. Walks up to
-# the nearest EXISTING ancestor, physically resolves it (`pwd -P`, which
-# follows symlinks), then re-appends any not-yet-created tail literally —
-# a tail that doesn't exist yet cannot itself be a symlink. This mirrors
-# `realpath -m` without depending on GNU coreutils (not guaranteed present
-# on macOS/BSD). Echoes the resolved path, or nothing if even "/" can't be
-# stat'd (should not happen for a well-formed absolute path).
-#
-# Copied verbatim from require-active-ticket.sh's _resolve_real_path (see
-# that file for the fuller symlink-bypass rationale) rather than sourcing
-# it directly — that hook's copy carries ticket-gate-specific comments and
-# the two call sites shouldn't share a lifecycle.
-# ------------------------------------------------------------------------
-_resolve_real_path() {
-  local p="$1" dir tail=""
-  [ -n "$p" ] || return 0
-  dir="$p"
-  while [ -n "$dir" ] && [ "$dir" != "/" ] && [ ! -e "$dir" ]; do
-    if [ -z "$tail" ]; then
-      tail="$(basename "$dir")"
-    else
-      tail="$(basename "$dir")/$tail"
-    fi
-    dir="$(dirname "$dir")"
-  done
-  if [ ! -e "$dir" ]; then
-    return 0
-  fi
-  if [ -d "$dir" ]; then
-    dir="$(cd "$dir" 2>/dev/null && pwd -P)"
-  else
-    local parent
-    parent="$(cd "$(dirname "$dir")" 2>/dev/null && pwd -P)"
-    if [ -n "$parent" ]; then
-      dir="$parent/$(basename "$dir")"
-    else
-      dir=""
-    fi
-  fi
-  [ -n "$dir" ] || return 0
-  if [ -n "$tail" ]; then
-    printf '%s/%s' "$dir" "$tail"
-  else
-    printf '%s' "$dir"
-  fi
-}
+# _resolve_real_path is sourced from _lib-path-resolve.sh — the single
+# shared definition (PR #1087 review, Rex finding #4). This file used to
+# carry its own inline copy with a comment claiming it was deliberately
+# separate from require-active-ticket.sh's; that "copied verbatim, not
+# sourced" shape is exactly the AgDR-0113 pattern #1086 exists to retire,
+# and the two copies had already started drifting. See
+# _lib-path-resolve.sh's header comment for the full rationale, including
+# why it is NOT the same thing as _lib-portfolio-paths.sh's
+# `_portfolio_canonicalize` (a deliberately different, fail-soft sibling).
+_LGHP_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
+if [ -f "$_LGHP_LIB_DIR/_lib-path-resolve.sh" ]; then
+  # shellcheck source=/dev/null
+  . "$_LGHP_LIB_DIR/_lib-path-resolve.sh"
+fi
 
 # ------------------------------------------------------------------------
 # ghp_classify REPO_ROOT HOOKS_DIR_NAME
--- a/.claude/hooks/_lib-path-resolve.sh
+++ b/.claude/hooks/_lib-path-resolve.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+# _lib-path-resolve.sh — single shared implementation of the portable
+# "realpath -m without GNU coreutils" helper used across the trust chain.
+#
+# Sourced by:
+#   - require-active-ticket.sh (the ticket gate's symlink-safe path check)
+#   - _lib-git-hooks-path.sh (used by bin/install-git-hooks.sh and
+#     .claude/hooks/check-git-hooks-installed.sh)
+#
+# Why this file exists (PR #1087 review, Rex finding #4): the installer
+# PR originally shipped its OWN copy of this function with a comment
+# claiming it was "copied verbatim ... rather than sourcing it directly."
+# That is exactly the AgDR-0113 pattern the #1086 arc exists to retire —
+# an agreement asserted in a comment instead of enforced by one shared
+# definition, in the very PR whose stated purpose is replacing
+# command-text-matching duplication with structured ground truth. There
+# is now exactly ONE _resolve_real_path in the codebase; every caller
+# sources this file rather than embedding its own copy.
+#
+# NOT the same thing as _lib-portfolio-paths.sh's `_portfolio_canonicalize`.
+# That function solves a related but distinct problem (portfolio-path
+# resolution across split-portfolio mode) and has deliberately different
+# semantics — it carries Windows drive-letter normalisation (#1018) and
+# fails SOFT (echoes the input unchanged when unresolvable), where this
+# function fails EMPTY (echoes nothing). Do not fold the two together;
+# `_portfolio_canonicalize`'s callers rely on its fail-soft behaviour.
+
+# Resolve PATH to its canonical, symlink-free absolute form. Walks up to
+# the nearest EXISTING ancestor, physically resolves it (`pwd -P`, which
+# follows symlinks), then re-appends any not-yet-created tail literally —
+# a tail that doesn't exist yet cannot itself be a symlink. This mirrors
+# `realpath -m` without depending on GNU coreutils (not guaranteed present
+# on macOS/BSD). Echoes the resolved path, or nothing if even "/" can't be
+# stat'd (should not happen for a well-formed absolute path).
+#
+# Why this matters (#883): without resolving symlinks first, a symlink
+# living under $HOME that POINTS INTO a governed tree (e.g.
+# ~/link-into-repo → the ops fork) would compare as "outside" the repo
+# under a naive string-prefix check, silently bypassing the gate.
+_resolve_real_path() {
+  local p="$1" dir tail=""
+  [ -n "$p" ] || return 0
+  dir="$p"
+  while [ -n "$dir" ] && [ "$dir" != "/" ] && [ ! -e "$dir" ]; do
+    if [ -z "$tail" ]; then
+      tail="$(basename "$dir")"
+    else
+      tail="$(basename "$dir")/$tail"
+    fi
+    dir="$(dirname "$dir")"
+  done
+  if [ ! -e "$dir" ]; then
+    return 0
+  fi
+  if [ -d "$dir" ]; then
+    dir="$(cd "$dir" 2>/dev/null && pwd -P)"
+  else
+    local parent
+    parent="$(cd "$(dirname "$dir")" 2>/dev/null && pwd -P)"
+    if [ -n "$parent" ]; then
+      dir="$parent/$(basename "$dir")"
+    else
+      dir=""
+    fi
+  fi
+  [ -n "$dir" ] || return 0
+  if [ -n "$tail" ]; then
+    printf '%s/%s' "$dir" "$tail"
+  else
+    printf '%s' "$dir"
+  fi
+}
--- a/.claude/hooks/check-git-hooks-installed.sh
+++ b/.claude/hooks/check-git-hooks-installed.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+# check-git-hooks-installed.sh — SessionStart advisory: warns when this
+# clone's core.hooksPath is unset, points at a non-existent directory, or
+# points at a stale leftover from a different clone, so the tracked
+# .githooks/pre-push hook silently never runs on a terminal `git push`.
+#
+# Part of me2resh/apexyard#1086 step 1. Same advisory shape as
+# check-upstream-drift.sh and check-jq-installed.sh: non-blocking, exit 0
+# always. The banner cannot force the fix — it removes the "nobody knew
+# this clone's terminal pushes were unprotected" failure mode.
+#
+# Silent exit paths (no output, no error):
+#   - Not a git repo
+#   - No tracked hooks dir (default: .githooks) in this repo at all — a
+#     managed-project clone that hasn't adopted ApexYard's git hooks yet
+#     is not a misconfiguration, just nothing to advise on
+#   - core.hooksPath already resolves correctly ("correct" classification)
+#   - The shared detection lib is missing (older fork pre-#1086; nothing
+#     to check against)
+#
+# Banner fires only for the three states install-git-hooks.sh's own README
+# documents as fixable: unset, missing (non-existent target dir), and
+# foreign-git-dir (stale pointer into a different clone) or a third-party
+# directory the classify lib can't confirm is deliberate.
+
+REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
+if [ -z "$REPO_ROOT" ]; then
+  exit 0
+fi
+
+cd "$REPO_ROOT" || exit 0
+
+HOOKS_DIR_NAME=".githooks"
+
+# Nothing tracked to install in this repo — silent. (Most managed-project
+# workspace clones today; the ops fork always carries .githooks/.)
+if [ ! -d "$REPO_ROOT/$HOOKS_DIR_NAME" ]; then
+  exit 0
+fi
+
+HOOK_DIR="$(cd "$(dirname "$0")" && pwd)"
+LIB="$HOOK_DIR/_lib-git-hooks-path.sh"
+if [ ! -f "$LIB" ]; then
+  exit 0
+fi
+# shellcheck source=/dev/null
+. "$LIB"
+
+STATE=$(ghp_classify "$REPO_ROOT" "$HOOKS_DIR_NAME")
+
+case "$STATE" in
+  correct)
+    exit 0
+    ;;
+
+  unset)
+    cat >&2 <<MSG
+ApexYard: core.hooksPath is unset in this clone — the tracked git pre-push
+hook isn't installed. A terminal \`git push\` won't run the tracked check
+set, only Claude-Code-driven pushes will. Run:
+  bash bin/install-git-hooks.sh
+MSG
+    ;;
+
+  missing:*)
+    RAW="${STATE#missing:}"
+    cat >&2 <<MSG
+ApexYard: core.hooksPath is set to '$RAW', which doesn't exist — the
+tracked git hook here is silently inert. Run:
+  bash bin/install-git-hooks.sh
+MSG
+    ;;
+
+  foreign-git-dir:*)
+    REST="${STATE#foreign-git-dir:}"
+    RAW="${REST%%:*}"
+    cat >&2 <<MSG
+ApexYard: core.hooksPath is set to '$RAW' — looks like a stale pointer into
+a different clone's .git directory. The tracked git hook here is silently
+inert. Run:
+  bash bin/install-git-hooks.sh
+MSG
+    ;;
+
+  third-party:*)
+    REST="${STATE#third-party:}"
+    RAW="${REST%%:*}"
+    cat >&2 <<MSG
+ApexYard: core.hooksPath is set to '$RAW', a directory that isn't this
+repo's tracked $HOOKS_DIR_NAME. If that's deliberate, ignore this. To
+install ApexYard's tracked hooks instead:
+  bash bin/install-git-hooks.sh --force
+MSG
+    ;;
+esac
+
+exit 0
--- a/.claude/hooks/check-git-hooks-installed.sh
+++ b/.claude/hooks/check-git-hooks-installed.sh
@@ -87,9 +87,8 @@ MSG
     RAW="${REST%%:*}"
     cat >&2 <<MSG
 ApexYard: core.hooksPath is set to '$RAW', a directory that isn't this
-repo's tracked $HOOKS_DIR_NAME. If that's deliberate, ignore this. To
-install ApexYard's tracked hooks instead:
-  bash bin/install-git-hooks.sh --force
+repo's tracked $HOOKS_DIR_NAME. If that's deliberate, ignore this — see
+bash bin/install-git-hooks.sh --help if you want to review or change it.
 MSG
     ;;
 esac
--- a/.claude/hooks/require-active-ticket.sh
+++ b/.claude/hooks/require-active-ticket.sh
@@ -52,55 +52,24 @@
 # let a command that also names an out-of-governance target FIRST slip an
 # in-repo target past a gate that stopped looking after target #1.
 
-# Resolve PATH to its canonical, symlink-free absolute form. Walks up to
-# the nearest EXISTING ancestor, physically resolves it (`pwd -P`, which
-# follows symlinks), then re-appends any not-yet-created tail literally —
-# a tail that doesn't exist yet cannot itself be a symlink. This mirrors
-# `realpath -m` without depending on GNU coreutils (not guaranteed present
-# on macOS/BSD). Echoes the resolved path, or nothing if even "/" can't be
-# stat'd (should not happen for a well-formed absolute path).
+# _resolve_real_path — portable, symlink-safe path canonicalisation.
 #
 # Why this matters (#883): without resolving symlinks first, a symlink
 # living under $HOME that POINTS INTO a governed tree (e.g.
 # ~/link-into-repo → the ops fork) would compare as "outside" the repo
 # under a naive string-prefix check, silently bypassing the gate.
-_resolve_real_path() {
-  local p="$1" dir tail=""
-  [ -n "$p" ] || return 0
-  dir="$p"
-  while [ -n "$dir" ] && [ "$dir" != "/" ] && [ ! -e "$dir" ]; do
-    if [ -z "$tail" ]; then
-      tail="$(basename "$dir")"
-    else
-      tail="$(basename "$dir")/$tail"
-    fi
-    dir="$(dirname "$dir")"
-  done
-  if [ ! -e "$dir" ]; then
-    # Nothing on the path exists at all — cannot resolve. Should not
-    # happen for an absolute path since "/" always exists.
-    return 0
-  fi
-  if [ -d "$dir" ]; then
-    dir="$(cd "$dir" 2>/dev/null && pwd -P)"
-  else
-    # $dir resolved to an existing FILE (not a directory) partway through
-    # the walk — canonicalize its parent and re-append its own basename.
-    local parent
-    parent="$(cd "$(dirname "$dir")" 2>/dev/null && pwd -P)"
-    if [ -n "$parent" ]; then
-      dir="$parent/$(basename "$dir")"
-    else
-      dir=""
-    fi
-  fi
-  [ -n "$dir" ] || return 0
-  if [ -n "$tail" ]; then
-    printf '%s/%s' "$dir" "$tail"
-  else
-    printf '%s' "$dir"
-  fi
-}
+#
+# Sourced from the single shared definition at _lib-path-resolve.sh (PR
+# #1087 review, Rex finding #4) rather than defined inline — this used to
+# be a self-contained copy, which put THREE near-identical
+# implementations of the same algorithm in .claude/hooks/ across two
+# files, one of which had already silently diverged. See
+# _lib-path-resolve.sh's own header comment for the full rationale.
+_RATC_HOOK_DIR="$(cd "$(dirname "$0")" && pwd)"
+if [ -f "$_RATC_HOOK_DIR/_lib-path-resolve.sh" ]; then
+  # shellcheck source=/dev/null
+  . "$_RATC_HOOK_DIR/_lib-path-resolve.sh"
+fi
 
 # ------------------------------------------------------------------------------
 # _ratc_evaluate_target FILE_PATH TOOL_NAME
--- a/.claude/hooks/tests/test_check_git_hooks_installed.sh
+++ b/.claude/hooks/tests/test_check_git_hooks_installed.sh
@@ -1,0 +1,184 @@
+#!/bin/bash
+# Smoke tests for .claude/hooks/check-git-hooks-installed.sh — SessionStart
+# advisory that warns when this clone's core.hooksPath is unset, stale, or
+# pointing at a foreign clone (me2resh/apexyard#1086 step 1).
+#
+# Discriminating by construction: this hook (and the _lib-git-hooks-path.sh
+# lib it sources) did not exist before #1086 step 1, so every case here
+# fails against the pre-change tree and passes only once both files exist
+# with the exact behaviour this suite exercises.
+#
+# Each case builds an isolated sandbox git repo, sets core.hooksPath to a
+# specific state, runs the hook from inside that repo, and asserts banner
+# output / silence. Same harness shape as test_check_jq_installed.sh.
+#
+# Cases covered:
+#   1. core.hooksPath unset               -> banner fires
+#   2. core.hooksPath already correct     -> silent
+#   3. core.hooksPath points at a missing dir           -> banner fires
+#   4. core.hooksPath points at a different clone's
+#      .git/hooks (the concrete #1086 failure mode)      -> banner fires
+#   5. No tracked .githooks/ dir in the repo at all      -> silent (nothing
+#      to advise on — a managed-project clone that hasn't adopted the
+#      framework's git hooks yet)
+#   6. Not inside a git repo at all                       -> silent
+#
+# Exit 0 if all cases pass; 1 on first failure tally.
+
+set -u
+
+HOOK_SRC="$(cd "$(dirname "$0")/.." && pwd)/check-git-hooks-installed.sh"
+LIB_SRC="$(cd "$(dirname "$0")/.." && pwd)/_lib-git-hooks-path.sh"
+PASS=0
+FAIL=0
+FAILED=""
+
+assert_silent() {
+  local label="$1" output="$2"
+  if [ -z "$output" ]; then
+    echo "PASS [$label] — silent"
+    PASS=$((PASS+1))
+  else
+    echo "FAIL [$label] — expected silent, got: $output" >&2
+    FAIL=$((FAIL+1)); FAILED="${FAILED}${label} "
+  fi
+}
+
+assert_fires() {
+  local label="$1" pattern="$2" output="$3"
+  if [ -n "$output" ] && printf '%s' "$output" | grep -qE "$pattern"; then
+    echo "PASS [$label]"
+    PASS=$((PASS+1))
+  else
+    echo "FAIL [$label] — expected /$pattern/, got: $output" >&2
+    FAIL=$((FAIL+1)); FAILED="${FAILED}${label} "
+  fi
+}
+
+# Build a sandbox with a copy of the hook + lib, so the test doesn't depend
+# on the real ops fork's checked-out state.
+build_repo() {
+  local dir="$1"
+  mkdir -p "$dir/.claude/hooks"
+  cp "$HOOK_SRC" "$dir/.claude/hooks/check-git-hooks-installed.sh"
+  chmod +x "$dir/.claude/hooks/check-git-hooks-installed.sh"
+  if [ -f "$LIB_SRC" ]; then
+    cp "$LIB_SRC" "$dir/.claude/hooks/_lib-git-hooks-path.sh"
+  fi
+  ( cd "$dir" && git init -q )
+}
+
+add_githooks() {
+  local dir="$1"
+  mkdir -p "$dir/.githooks"
+  printf '#!/bin/sh\nexit 0\n' > "$dir/.githooks/pre-push"
+  ( cd "$dir" && git add .githooks .claude && GIT_AUTHOR_NAME=t GIT_AUTHOR_EMAIL=t@t.com \
+      GIT_COMMITTER_NAME=t GIT_COMMITTER_EMAIL=t@t.com \
+      git commit -q -m "chore: init" )
+}
+
+run_hook_from() {
+  local dir="$1"
+  ( cd "$dir" && bash .claude/hooks/check-git-hooks-installed.sh 2>&1 )
+}
+
+if [ ! -f "$HOOK_SRC" ]; then
+  echo "FAIL: hook not found at $HOOK_SRC" >&2
+  exit 1
+fi
+
+# CASE 1: unset -> banner
+case_unset() {
+  local sandbox; sandbox=$(mktemp -d)
+  build_repo "$sandbox"
+  add_githooks "$sandbox"
+  assert_fires "unset -> banner fires" "core\.hooksPath is unset" "$(run_hook_from "$sandbox")"
+  rm -rf "$sandbox"
+}
+
+# CASE 2: already correct -> silent
+case_correct() {
+  local sandbox; sandbox=$(mktemp -d)
+  build_repo "$sandbox"
+  add_githooks "$sandbox"
+  git -C "$sandbox" config core.hooksPath .githooks
+  assert_silent "already correct -> silent" "$(run_hook_from "$sandbox")"
+  rm -rf "$sandbox"
+}
+
+# CASE 3: missing dir -> banner
+case_missing_dir() {
+  local sandbox; sandbox=$(mktemp -d)
+  build_repo "$sandbox"
+  add_githooks "$sandbox"
+  git -C "$sandbox" config core.hooksPath "$sandbox/nope-not-here"
+  assert_fires "missing dir -> banner fires" "doesn't exist" "$(run_hook_from "$sandbox")"
+  rm -rf "$sandbox"
+}
+
+# CASE 4: foreign clone's .git/hooks -> banner
+# repo and other-clone are SIBLINGS (both directly under $parent), not
+# nested — nesting them would make other-clone's path resolve as "inside
+# repo's own tree" and get classified as third-party instead of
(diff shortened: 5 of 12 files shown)


atlas-apex (commented):
## Code Review: PR #1087

**Commit**: `804aa8d4d38b8074afb31e718566ece4efc658cb`

### Summary

Step 1 of #1086: ships `bin/install-git-hooks.sh` (per-clone `core.hooksPath` installer), a shared classification lib, a SessionStart advisory, `/setup` + `/handover` wiring, and two test files. Scope discipline is exact — `.githooks/pre-push`, `block-main-push.sh`, `bin/run-pre-push-checks.sh` and `AgDR-0104` are all untouched, so no step-2 or step-3 work has leaked in.

I verified the suite myself rather than taking the author's numbers: **129/129 pass, FAIL=0**, run in three chunks (44 + 43 + 42). That matches the PR body's claim exactly, with no pre-existing local failures to net out. `shellcheck --severity=warning` clean on all five shell files.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Fail** — three empirically-proven non-discriminating gaps (below)
- Security: **Pass** — no bypass found; `--force` refusal verified real
- Performance: **Pass** (N/A)
- Summary Bullet Narrative: **Pass** — every bullet carries what + why
- Technical Decisions (AgDR): **N/A** — implements an existing decision (AgDR-0104), adds no new dependency or technology
- Adopter Handbooks: **Pass** — `migration-safety.md` (blocking) N/A, no schema/data migration in diff
- Fallow Static Analysis: **N/A** — no JS/TS in diff

---

## 1. The shared-classifier claim: genuinely shared — but the PR introduces a *new* duplication one level down

**The headline claim holds where it was checked.** `check-git-hooks-installed.sh:47-49` sources `_lib-git-hooks-path.sh` and calls `ghp_classify`; `bin/install-git-hooks.sh:68,141` sources the same lib and calls the same function. There is no inline re-implementation and no comment-asserted agreement. I confirmed this is real sharing, not a claim: mutating the lib's final fallthrough changed **both** callers' behaviour in one edit. Good work — this is the right instinct and it was executed correctly.

**But the same PR adds a third copy of `_resolve_real_path`, with the agreement asserted in a comment — and the copies have already diverged.**

`_lib-git-hooks-path.sh:23-26` says plainly:

> Copied verbatim from require-active-ticket.sh's `_resolve_real_path` ... rather than sourcing it directly

There are now three implementations of this same path-canonicalisation algorithm in `.claude/hooks/`:

| File | Function | State |
|---|---|---|
| `require-active-ticket.sh` | `_resolve_real_path` | logic-identical to the new copy (comments differ only) |
| `_lib-git-hooks-path.sh` (**new**) | `_resolve_real_path` | logic-identical to the above |
| `_lib-portfolio-paths.sh:110` | `_portfolio_canonicalize` | **already diverged** |

The third one links itself to the others by comment too — *"Algorithm mirrors `_resolve_real_path()` in require-active-ticket.sh"* — and it is no longer the same algorithm. It carries the Windows drive-letter normalisation from #1018, and it **fails soft** (echoes input unchanged on unresolvable) where `_resolve_real_path` **fails empty**. The new copy this PR adds inherits neither.

This is the AgDR-0113 pattern, in the PR whose stated purpose includes avoiding it. I could not demonstrate a wrong *outcome* from the missing Windows fix inside `ghp_classify` (the two branches it affects converge on the same answer), so I'm not claiming that one bites today. The fail-empty difference does have a demonstrable symptom — see §3.

**Ask:** either source `_resolve_real_path` from one place (a `_lib-path-resolve.sh` the three callers share), or, if the separate-lifecycle argument in the comment is genuinely intended, say so honestly — "these are three deliberate forks that will drift" — rather than "copied verbatim", which reads as an invariant that nothing enforces. The current comment asserts an agreement that has already been broken by a sibling.

---

## 2. Discriminating tests: the stated proof is weak, and three real gaps confirmed

The PR body's proof is *"temporarily removed the new files and confirmed the suites fail (file-not-found)"*. As the reviewer brief notes, that proves the tests **execute**, not that they **discriminate** — it would hold identically against a stub that always returned success.

To be fair to the author: the tests are substantially better than that stated proof. Each installer case asserts exit code **and** message shape **and** the resulting `git config --get core.hooksPath` value, and the message assertions genuinely separate the classification branches from one another. Case 5 in particular is a real test of the safety property: exit 1, `refusing to overwrite`, config **UNCHANGED**.

But I mutated the implementation to find out what the suites actually catch, and three mutants survive:

### 2a. `_resolve_real_path` is entirely non-discriminated — and it is load-bearing

Replacing the whole 33-line body with `_resolve_real_path() { printf "%s" "$1"; }`:

```
installer suite:  PASS: 23   FAIL: 0
advisory suite:   PASS:  6   FAIL: 0
```

All 29 assertions green against a function with its logic deleted. It is not dead code, though — with a repo reachable through a symlink and `core.hooksPath` set through that symlink:

| Implementation | Result |
|---|---|
| as shipped | `rc=0` — `already set to .githooks ... no change (idempotent)` |
| passthrough mutant | `rc=1` — **refuses**, "looks like a deliberate adopter choice, not a stale leftover" |

So the shipped behaviour is correct and the function earns its place — but nothing in the suite would notice if it regressed. Add a symlinked-repo case.

### 2b. The advisory hook's `third-party:*` branch has no test at all

Deleting `check-git-hooks-installed.sh:85-94` outright: `PASS: 6  FAIL: 0`. The advisory suite covers unset / correct / missing / foreign-git-dir / no-`.githooks` / not-a-repo — but never third-party.

### 2c. The classifier's external-third-party fallthrough is untested — and it is the branch guarding the PR's central safety property

Cases 5 and 6 exercise `--force` refusal via `custom-hooks-dir` **inside** the repo, which hits the `"$repo_root"/*` prefix branch at `_lib-git-hooks-path.sh:128-133`. The **final** fallthrough at line 147 — an existing, real directory **outside** the repo with no `.git` component — is never reached by any test. That is the canonical real-world case, and the lib's own comment names it: *"a genuinely external directory an adopter configured on purpose (e.g. a company-wide shared hooks dir)"*.

I mutated that fallthrough to print `unset` instead — i.e. the installer now **silently clobbers a deliberate third-party `core.hooksPath` with no `--force`**, the exact thing the PR promises never to do:

```
installer suite:  PASS: 23   FAIL: 0
advisory suite:   PASS:  6   FAIL: 0
```

Green. I confirmed by direct probe that the shipped code gets this **right** (`rc=1`, config unchanged) — this is a test gap, not a live defect. But it is the one branch where a regression is a broken safety guarantee rather than a cosmetic miss, and it is the branch with no coverage.

**Ask:** three test cases — symlinked-repo `correct`; advisory third-party banner; installer refusal on an *external* third-party directory.

---

## 3. Other findings

### suggestion: `--repo-dir` on a non-repo directory walks up and configures the ops fork

`bin/install-git-hooks.sh:120` resolves the target with `cd "$REPO_DIR" && git rev-parse --show-toplevel`. The `cd` failure path is safe — it's inside a subshell behind `&&`, so a failed `cd` yields an empty `TARGET_ROOT` and exit 2, never a command running in the ops fork. That satisfies `isolated-builds.md`.

The residual is the walk-up. `workspace/<name>/` lives inside the ops fork's worktree (gitignored, but still inside it), so if that directory exists and is *not* its own repo, `rev-parse` walks up and returns the ops fork. Reproduced:

```
--repo-dir <ops>/workspace/proj-x   ->  rc=0
  "core.hooksPath set to .githooks in <ops> (was unset)."
```

Not destructive — it writes the correct value for the ops fork — but `/handover`'s new step branches on exit 0 as "installed, note it in the step 10 summary", so it would report a managed-project install that never happened. `/handover` gates this on `CLONE_STATUS` being `cloned` or `preserved`, and `preserved` can be a pre-existing non-repo directory. One line fixes it: assert the resolved `TARGET_ROOT` is the same repo as `--repo-dir` before proceeding.

### nit: fail-empty conflates "unresolvable" with "missing", and the message says something untrue

A real, existing, deliberate hooks dir whose *parent* isn't traversable resolves to empty, hits `_lib-git-hooks-path.sh:120`, and is classified `missing:` — so the installer overwrites it **without `--force`**:

```
rc=0  "core.hooksPath repaired: was '<...>/secure/hooks'
       (points at a non-existent directory), now .githooks"
```

The directory exists. Counter-argument acknowledged: if the user can't traverse to it, git can't run hooks from it either, so repairing is arguably the right call — but then the message should say "unresolvable", not "non-existent". This is the concrete cost of the fail-empty/fail-soft divergence in §1. Low severity.

### note: `git config` failure degrades to `unset`

A `git config` that errors (corrupt `.git/config`) makes `ghp_classify` return `unset`, so the advisory fires the "run the installer" banner. That's the fail-safe direction, so no change needed — just flagging that the classifier can't distinguish "no value" from "couldn't read".

### note: own-repo `.git/hooks` classifies as `third-party` (refuse), not `foreign-git-dir`

The `"$repo_root"/*` check runs before the `.git` check, so pointing `core.hooksPath` at *this* repo's `.git/hooks` (git's default) is refused rather than repaired. Verified `rc=1`, config unchanged. The lib's doc comment states this intent explicitly ("...and is outside REPO_ROOT"), so it's a documented choice, not a fall-through to the wrong branch. The five states are exhaustive and the installer's `*)` catch-all exits 2 on anything unrecognised — I found no state reaching the wrong branch.

---

## 4. Verified good

- **`.claude/settings.json`** — valid JSON; exactly one entry added (76 → 77 hook commands); a normalised event/matcher/script diff of base vs head shows `SessionStart:*:check-git-hooks-installed.sh` as the **only** line changed. No existing wiring disturbed. Placed adjacent to `check-jq-installed.sh` with the identical ops-root-resolution wrapper.
- **Fail-open discipline** — advisory exits 0 on every path I could construct: git absent from PATH, `git` present but always failing, `git rev-parse` fine but `git config` erroring, not a git repo, lib missing, and when the banner actually fires. Matches `check-upstream-drift.sh`'s shape (`exit 0` at every branch, `cd ... || exit 0`). No `set -e`.
- **`--force` semantics** — the refusal is real, not just a message: exit 1 with the config value verifiably unchanged, confirmed both by case 5 and by my own probe.
- **Installer exit-code contract** — 0/1/2/3 all reachable, all tested, and 2 vs 3 correctly distinguish "not a repo" from "nothing to install".
- **PR body** — narrative Summary bullets (each carries what *and* why), a real Testing section, complete Glossary. The scope paragraph naming steps 2 and 3 as explicitly out of scope is exactly right.

---

### Verdict

**CHANGES REQUESTED**

Nothing here is a bypass, and the shipped behaviour is correct in every probe I ran — this is a well-built PR with clean scope and a working safety property. The blocker is narrow and cheap to clear, and it's on the trust chain (`.claude/hooks/**` + `.claude/settings.json`), which `right-size-ceremony.md` rail 1 puts in the Heavy tier regardless of diff size:

1. **Resolve the `_resolve_real_path` triplication** (§1) — share it, or replace the "copied verbatim" comment with an honest statement that these are deliberate forks. As written, the file asserts an invariant that a sibling has already broken, which is the pattern AgDR-0113 exists to stop.
2. **Add the three missing test cases** (§2a/2b/2c) — most importantly 2c, the external-third-party refusal, since that's the one branch where a future regression silently breaks the guarantee the PR is built to provide.

The `--repo-dir` walk-up (§3) is a suggestion, not a blocker.

---
📌 Reviewed commit: `804aa8d4d38b8074afb31e718566ece4efc658cb`

atlas-apex (commented):
_(Duplicate submission from a retried tool call — the review is the preceding comment. Ignore this one.)_

atlas-apex (commented):
## Security Review: PR #1087

**Commit**: `804aa8d4d38b8074afb31e718566ece4efc658cb`

Auto-fired per `.claude/rules/role-triggers.md` (#777) — the diff touches `.claude/hooks/**` and `.claude/settings.json`. Reviewed on the terms the trigger implies: this PR ships code that **decides which scripts git executes on the operator's behalf**.

Every finding below was reproduced by execution against scratch git repos, not inferred from reading the source.

### Summary

The direction is right and the classification design is the correct answer to AgDR-0104 — `ghp_classify` decides on **git config values and resolved filesystem paths**, never on command text. That is exactly the structured-ground-truth move #1086 argues for, and it is implemented cleanly.

Scope discipline is clean: `git show --name-only 804aa8d` confirms neither `.githooks/pre-push` nor `block-main-push.sh` is touched. **No step-2/step-3 ordering violation.**

Two findings block. One is a trust-boundary crossing on `/handover`'s untrusted input path; one is a fail-open in the installer that the PR's own success message hides.

### Checklist Results

- Secrets & Credentials: **Pass** — no credentials handled; the written config value is never attacker-influenced (always `$HOOKS_DIR_NAME`, relative).
- Injection Prevention: **Pass** — the unquoted heredocs interpolating `$RAW` / `$RESOLVED` are safe; shell expansion is not recursive, so a hostile `core.hooksPath` value cannot inject a command substitution.
- XSS Prevention: **N/A**
- Auth & Authorisation: **Pass** — no authz surface.
- Data Protection: **Pass**
- API Security: **N/A**
- **Gate & Trust-Chain Integrity: FAIL** — see HIGH-1 and MEDIUM-2.

---

### Security Issues Found

#### HIGH-1 — `/handover` hands a freshly-cloned third-party repo code execution on ordinary git operations

**This is the answer to the sharpest question on the PR, and it is "theirs, not ours."**

`bin/install-git-hooks.sh` gates entirely on `[ -d "$TARGET_ROOT/$HOOKS_DIR_NAME" ]` (line ~913). It performs **no provenance check, no content check, and no operator confirmation**. It does not install anything — it points git at whatever `.githooks/` the target clone already ships. For the ops fork that directory is ApexYard's. For a repo cloned by `/handover` it is **that repo's committed scripts**.

Reproduced end-to-end. A scratch "third-party" repo committing its own `.githooks/`, cloned, then `install-git-hooks.sh --repo-dir <clone>`:

```
BEFORE installer: hooksPath=''
core.hooksPath set to .githooks in .../workspace-clone (was unset).
installer exit=0
AFTER  installer: hooksPath='.githooks'

after 'git checkout -b': PAYLOAD-POSTCHECKOUT
after 'git fetch':       PAYLOAD-REFTXN
```

Note what fired: **`post-checkout` on `git checkout -b`, and `reference-transaction` on `git fetch`.** No push required. `reference-transaction` runs on essentially any ref-updating operation, which the framework performs routinely and silently in workspace clones.

Why this is a genuine trust-boundary crossing rather than a theoretical one: git's default hooks directory is `$GIT_DIR/hooks`, which is **never cloned**. That is a deliberate git security property — it is what makes `git clone` of an untrusted repo a safe, read-only act. Repointing `core.hooksPath` at a *tracked, committed* directory removes exactly that property. It is why husky/pre-commit-class tooling requires an explicit user-run install step rather than doing it on clone.

`/handover` is the framework's **least-trusted input path**: it clones an arbitrary `<repo-url>` the operator wants to *evaluate* ("harnessability scoring across 5 dimensions" — a read-heavy activity), and the new step 1.5-git-hooks is wired as **"default: always attempt"** with no prompt. Adopting a repo implies some trust, and an operator who later runs its build would execute its code anyway — but "I cloned this to score it" and "its scripts now run as me on my next `git fetch`" are materially different acts, and only the operator can consent to the second.

The SKILL text also has the risk framing backwards. It describes exit 3 ("no `.githooks/` in the clone") as the expected common case and exit 0 as the one to note as success — but exit 3 is the **safe** outcome, and exit 0 against a third-party clone is precisely the outcome that transfers execution. Relatedly, "install ApexYard's tracked git hooks into the workspace clone" is inaccurate: nothing ApexYard-provided is installed there.

**Suggested remediation** (any one closes it):
1. **Preferred for a step-1 PR:** drop the `/handover` call. Keep `/setup` (the ops fork configuring its own repo — that is unambiguously safe and needs no change). The PR already declares "managed-project clones adopting `.githooks/`" out of scope; wiring the installer into `/handover` is that same out-of-scope work arriving early, without its safety design.
2. Verify provenance before writing — confirm the target `.githooks/` is framework-provided (content fingerprint, or a marker file the framework itself writes) and exit 3 otherwise.
3. Require explicit operator confirmation that **lists the scripts** which will become executable, and default to declining.

#### MEDIUM-2 — the installer fails OPEN on a failed `git config` write, and reports success

None of the four `git config core.hooksPath` calls (the `unset`, `missing:*`, `foreign-git-dir:*`, and `third-party:*` branches) check their exit status. When the write fails, the script prints its success message and exits 0.

Reproduced, stale `.git/config.lock` (a crashed or concurrent git — routine in a repo that runs many parallel worktree agents):

```
error: could not lock config file .git/config: File exists
core.hooksPath set to .githooks in .../r (was unset).
exit=0
ACTUAL config value now: ''      <-- nothing was written
```

Reproduced again, multi-valued `core.hooksPath`:

```
error: cannot overwrite multiple values with a single value
core.hooksPath repaired: was '/tmp/bbb' ... now .githooks in .../r
exit=0
values now: /tmp/aaa /tmp/bbb    <-- unchanged
```

The brief for this class is explicit and this PR is on the wrong side of it: the *advisory hook* must fail open (it does — verified, every path exits 0), and the *installer* must fail closed. `/setup` Step 0.5 is marked **REQUIRED** and branches on this exit code, so the operator is told the clone is protected when it is not. This also contradicts the PR's own testing claim that the resulting config value is asserted — the automated suite asserts it only on the happy paths, never on a write that fails.

**Fix:** check the exit status of each `git config` write (or re-run `ghp_classify` afterward and confirm `correct`), and exit non-zero with a loud error otherwise.

#### MEDIUM-3 — `--repo-dir` silently retargets to the enclosing repository

`TARGET_ROOT` is taken from `git rev-parse --show-toplevel` inside `--repo-dir` with **no assertion that the resolved root is the directory that was requested**. A `--repo-dir` that exists but is not itself a repo resolves upward and writes to whatever repo encloses it:

```
$ install-git-hooks.sh --repo-dir <outer-repo>/workspace/proj   # not a repo
core.hooksPath set to .githooks in <outer-repo> (was unset).
exit=0
```

This is directly reachable in `/handover`: I verified that `workspace/` is inside the ops-fork git tree (gitignored, but still in the worktree), so `git rev-parse --show-toplevel` from `workspace/<name>` on a **failed or partial clone** returns the ops fork root. The installer would then configure the ops fork while reporting on the managed project. The SKILL's `CLONE_STATUS` guard makes this unlikely — but that guard is prose in a skill file, not a check in the script, and this is the wrong-repo-write class `.claude/rules/isolated-builds.md` exists to prevent.

**Fix:** when `--repo-dir` is given, assert the resolved `TARGET_ROOT` equals the resolved `--repo-dir`; exit 2 otherwise.

#### LOW-4 — `--hooks-dir` accepts traversal and is written verbatim

`--hooks-dir ../evil-hooks` passes the `[ -d ]` check and writes `core.hooksPath=../evil-hooks`, pointing git outside the repo (verified, exit 0). Operator-supplied, so not an attacker vector — but the asymmetry is worth closing: `_lib-git-hooks-path.sh` reasons carefully about "outside REPO_ROOT" when *detecting* state, while the installer validates nothing about the value it *writes*. Reject a `--hooks-dir` that resolves outside `TARGET_ROOT`.

#### LOW-5 — `.githooks` committed as a symlink is accepted

`[ -d ]` follows symlinks, so a repo can commit `.githooks` as a symlink and have git resolve hooks to a target outside the repo tree (verified: accepted, exit 0). No new capability beyond HIGH-1, but it widens where those scripts can live. Worth an explicit `[ -L ]` refusal once HIGH-1 is addressed.

#### LOW-6 — the third-party advisory banner recommends `--force`

`check-git-hooks-installed.sh` tells the operator to run `bash bin/install-git-hooks.sh --force` on the `third-party:*` state. `--force` exists specifically to override a deliberate adopter choice; a recurring SessionStart banner that suggests it erodes the one guard protecting that choice. The "If that's deliberate, ignore this" prefix helps — consider dropping the `--force` line and pointing at `--help` instead.

---

### What holds up well

Worth stating plainly, since these were the things I set out to break and could not:

- **No command-text pattern-matching anywhere in the decision path.** The state machine reads git config and resolved paths. This is the AgDR-0104-shaped fix, not another parser.
- **The advisory hook fails open correctly** — every path exits 0, including `cd "$REPO_ROOT" || exit 0` and the missing-lib guard.
- **`.claude/settings.json` is clean** — valid JSON, exactly one SessionStart entry added between `check-jq-installed` and `check-portfolio-config`, no existing matcher altered, no wiring disturbed (17 SessionStart / 77 total entries).
- **The written value is never attacker-influenced** and is relative, which also makes it correct across git worktrees (`core.hooksPath` is shared config; a relative value resolves per-worktree).
- **Tests re-run independently, not taken on report:** `bin/run-hook-tests.sh` in two chunks, **PASS=129 FAIL=0**, no failures. The two new suites tally 23 and 6 assertions. `shellcheck --severity=warning` on all three new shell files: clean.

### Verdict

**CHANGES REQUESTED**

HIGH-1 and MEDIUM-2 block. HIGH-1 is closable for this PR by simply dropping the `/handover` invocation — the `/setup` half is sound and the rest of the change is good work that should land. MEDIUM-2 is a few lines of exit-status checking.

The ordering constraint from #1086 is respected and I found no step-2 or step-3 logic here.

---
📌 Reviewed commit: `804aa8d4d38b8074afb31e718566ece4efc658cb`

atlas-apex (commented):
## Security Review: PR #1087 (re-review)

**Commit**: `03152a57d33d07800bb4088d23286d8c579d6c79`
**Previous review**: `804aa8d4d38b8074afb31e718566ece4efc658cb` — CHANGES REQUESTED (1 HIGH, 2 MEDIUM, 3 LOW)

Scope grew since the last pass: `require-active-ticket.sh` — **the pre-build gate** — is now in the diff, because Rex's finding #4 was fixed by extracting a shared `_resolve_real_path` into a new `.claude/hooks/_lib-path-resolve.sh`. I reviewed that file as a first-time trust-chain review, not as a delta. Everything below was reproduced by execution against scratch git repos and a scratch ops-fork fixture; nothing is taken on the author's report.

---

### Lead finding: the conditional source is **fail-CLOSED**. Hypothesis refuted.

The concern was real and correctly identified. `require-active-ticket.sh` now sources the shared lib conditionally with no `else`, so a missing/unreadable/mislocated `_lib-path-resolve.sh` leaves `_resolve_real_path` **undefined** where it was previously always available. I renamed the lib and ran all three consumers.

Undefined `_resolve_real_path` yields **empty output** (bash `command not found`, non-zero, empty command substitution). Empty resolves toward **blocked**, in all three:

| Consumer | Missing-lib behaviour | Direction |
|---|---|---|
| `require-active-ticket.sh` (pre-build gate) | Empty `_og_real_target` → the `[ -n "$_og_real_target" ]` guard at :333 is false → the **entire** out-of-governance block is skipped → falls through to the ticket gate → **rc=2 BLOCKED** | fail-CLOSED |
| `bin/install-git-hooks.sh` | Empty at :242 → doesn't match `"$TARGET_ROOT"/*` → **exit 2**, nothing written | fail-CLOSED |
| `check-git-hooks-installed.sh` (advisory) | `ghp_classify` returns `missing:*` → banner fires → **exit 0** | fail-open (correct side for an advisory) |

The #883 design property — *"an unresolvable target is NEVER exempted here; it falls through to the ticket gate"* — **survives the refactor verbatim**. Measured, with the lib present as control:

```
lib PRESENT: out-of-governance target        -> rc=0 EXEMPT
lib PRESENT: in-repo target                  -> rc=2 BLOCKED
lib ABSENT : out-of-governance target        -> rc=2 BLOCKED   (was exempt)
lib ABSENT : in-repo target                  -> rc=2 BLOCKED
lib ABSENT : symlink-outside -> into ops fork-> rc=2 BLOCKED
```

The degraded mode is **strictly more restrictive** than the shipped mode. The cost is availability (a legitimate `~/.zshrc`-class write gets gated), never authority. That is the correct direction for a gate.

**Worth calling out:** the installer's fail-closed behaviour here is not luck-adjacent-to-luck — it is a *consequence of the LOW-4 fix from my last review*. The new `--hooks-dir` containment check runs **before** `ghp_classify`, so it trips on the empty resolution first. Without it, `ghp_classify` would have returned `missing:*` for a **deliberate third-party** `core.hooksPath` and the installer would have clobbered it with no `--force` — the exact fail-open the brief anticipated. I verified the deliberate value is untouched in the degraded run (`val AFTER` unchanged). A fix requested on other grounds happens to close this; that is worth knowing, because it means the property is currently **incidental to check ordering**, not asserted anywhere.

Sibling checks:

- **`_lib-git-hooks-path.sh`'s source of the lib** — same conditional shape, and the same conclusion: degraded `ghp_classify` returns `missing:*`, which is the *auto-repair* branch, so on its own it would be a fail-open. It isn't reachable as one because of the ordering above.
- **`$(dirname "$0")`** — `.claude/settings.json` invokes via `exec "$r/.claude/hooks/require-active-ticket.sh"` (absolute). Verified the resolver works under absolute exec, relative-from-hooks-dir, and relative-from-ops-root. The pattern is identical to the pre-existing `HOOK_DIR` resolution already used for `_lib-detect-bash-write.sh` and `_lib-ops-root.sh` in the same file, so it introduces no new class of risk. A `cd` failure yields an empty dir → `[ -f "/_lib-path-resolve.sh" ]` false → not sourced → fail-closed again.

**Naming what is a control and what is a backstop, per AgDR-0104:** in the ticket gate, `_resolve_real_path` is **defence in depth, not the sole control**. I mutated it to a passthrough (`printf "%s" "$1"`) and the #883 symlink-into-repo attack is *still* blocked — the independent nearest-existing-ancestor `git -C` probe follows symlinks at the OS level and catches it. Genuine out-of-governance targets still exempt correctly under the same mutant. So the deduplication did not concentrate the gate's safety into one newly-conditional function.

---

### The three prior findings — all confirmed fixed by execution

#### HIGH-1 — `/handover` wiring removed ✅

`grep -rn install-git-hooks .claude/skills/` returns exactly two hits: the `/setup` call, and a `/handover` **prose deferral note**. The whole `### 1.5-git-hooks` step, including its exit-code branch table, is deleted. No other path in the PR can reach the installer against a non-ops-fork target — `/setup`'s invocation passes no `--repo-dir`, so it targets the cwd repo (the ops fork).

**The risk framing is now correct.** My original point was that the removed version had it inverted (exit 3 described as the expected case to stay quiet about, exit 0 as the success to note). The replacement states the direction accurately: it names the trust asymmetry (`/setup` points at **ApexYard's own** `.githooks/`; a handover clone's is *that repo's*), names the missing provenance check, names why `git clone` is normally read-only (`$GIT_DIR/hooks` is never cloned), and names the concrete hooks — `post-checkout`, `reference-transaction` — with the load-bearing qualifier **"no push required."** It also states the gap it leaves rather than pretending none exists, and names the correct future shape (install the framework's *own* hook into the clone's untracked `.git/hooks/`, never point at a tracked third-party directory). That is the honest version.

#### MEDIUM-2 — installer fails closed on a failed `git config` write ✅

New exit code 4. Exercised both original repros plus a probe of whether the read-back is decorative:

| Case | rc | Success text? | Config after |
|---|---|---|---|
| stale `.git/config.lock` | **4** | none | `''` (unwritten) |
| multi-valued `core.hooksPath` (`--force`) | **4** | none | both values unchanged |
| `git config` exits **0** without landing (shimmed git) | **4** | none | `''` |

The third case is the answer to "is the read-back load-bearing?" — **yes.** With `git config`'s own exit status returning 0, the re-classify catches it: `ERROR: core.hooksPath write did not take effect (post-write state: 'unset'). Not reporting success.` Two independent checks, and the second one alone stops a silent no-op. Not decorative.

#### MEDIUM-3 — `--repo-dir` no longer silently retargets ✅

Re-ran the exact original shape — a `workspace/<name>` path inside the ops-fork git tree:

```
--repo-dir <ops>/workspace/proj-x   ->  rc=2
  "... is not itself a git repository root — it resolved to the ENCLOSING
   repo at '<ops>' instead. Refusing to silently configure a different repo
   than requested."
<ops> core.hooksPath: ''   (untouched)
```

Controls confirm no false positives: a genuine repo root still installs (rc=0); a **symlinked** path to a genuine repo root still installs (rc=0) — the check compares resolved forms; a subdirectory of a genuine repo now refuses (rc=2), which is the intended tightening.

#### The three LOWs ✅

- **LOW-4** `--hooks-dir ../evil-hooks` → rc=2, `resolves to '<...>/evil-hooks', which is outside <repo>`, config unchanged.
- **LOW-5** `.githooks` committed as a symlink → rc=2, checked via `[ -L ]` **before** `[ -d ]`, config unchanged.
- **LOW-6** third-party advisory banner: zero occurrences of `--force`; now points at `--help`.

---

### Standing properties re-checked

- **`.claude/settings.json` — unchanged since my last pass.** Diffed against the merge base `32a84a3`: valid JSON, hook-command count 76 → 77, and the *only* changed line is the added `SessionStart` entry for `check-git-hooks-installed.sh`, using the identical ops-root-resolution wrapper as its neighbours. No matcher altered, no gate wiring disturbed.
- **Fail-open vs fail-closed sides** — advisory open (verified exit 0 on every degraded path), installer closed (exit 2/4). The new `_lib-path-resolve.sh` is the third component: it is a **pure function with no side effects and no exit path of its own**, and it fails **empty**. Empty is interpreted as "refuse/gate" by both closed-side consumers and as "warn" by the open-side one, so the lib inherits the correct side from each caller rather than imposing one. That is the right design for a shared primitive, and the header comment correctly documents why it must not be folded into `_lib-portfolio-paths.sh`'s deliberately fail-**soft** `_portfolio_canonicalize`.
- **No command-text pattern-matching in any decision path** — re-confirmed. Zero matches for `tool_input.command`, `$COMMAND`, or command-shaped `case`/`grep` in all four non-test files. Every decision reads git config values and resolved filesystem paths.
- **Scope discipline** — `git diff --name-only 32a84a3 03152a5` touches neither `.githooks/**`, nor `block-main-push.sh`, nor `docs/agdr/**`. No step-2 or step-3 logic.
- **`shellcheck --severity=warning`** clean on all five shell files including `require-active-ticket.sh`.

### Test verification (re-run, not taken on report)

Full suite via three chunks of the runner's own execution loop:

```
CHUNK1 PASS=44 FAIL=0
CHUNK2 PASS=44 FAIL=0
CHUNK3 PASS=41 FAIL=0
------------------------------
TOTAL  PASS=129 FAIL=0
```

The pre-build gate's own suite, run separately as the brief asks: **`test_require_active_ticket_bash.sh` — PASS: 78, FAIL: 0.** The author's 78/78 claim is accurate. The suite was correctly updated to copy `_lib-path-resolve.sh` into all three sandbox shapes, so the sourced path is genuinely exercised rather than accidentally bypassed.

Installer suite **23 → 41**, advisory **6 → 8**. Rex's three requested discriminating cases are all present and named: `symlinked repo idempotent`, `external third-party no --force`, and the advisory `third-party -> banner fires`. I confirmed the first is now discriminating by re-running the passthrough mutation on `_resolve_real_path` — the installer suite drops to **PASS: 16 FAIL: 25** where it previously stayed fully green. Rex's §2a is closed.

---

### Findings (all LOW — none blocking)

#### LOW-A — `docs/getting-started.md:158` still says `/handover` calls the installer

> "…that's exactly why `/setup` **and `/handover`** call it…"

`/handover` does not, deliberately, on the security grounds recorded in its own SKILL.md. This is the one place in the PR where the documentation contradicts the fix. It matters slightly more than an ordinary typo because it is the kind of sentence a future agent cites as prior intent when re-adding the call. One-word fix; not worth another round on its own.

#### LOW-B — the degraded (missing-lib) mode is unasserted and its diagnostics are poor

Three things, all cosmetic-to-operational rather than security:

1. No test anywhere pins the fail-closed property I verified by hand. The installer's fail-closed path in particular depends on the `--hooks-dir` containment check running **before** `ghp_classify` — a correct-but-incidental ordering that a future refactor could reverse without any test noticing.
2. All three consumers emit raw `bash: line NNN: _resolve_real_path: command not found` rather than a clean "shared lib missing" diagnostic. The installer already models the right shape at :79 for `_lib-git-hooks-path.sh` (explicit message, exit 2).
3. In degraded mode the advisory tells a **correctly configured** repo that `core.hooksPath is set to '.githooks', which doesn't exist` — false, though harmless.

Cheapest durable fix: make the source unconditional in `_lib-git-hooks-path.sh` and `require-active-ticket.sh`, or keep the guard and add an explicit `else` that emits a named error. The current shape is safe; it just isn't *asserted* to be safe.

#### LOW-C (nit) — `/setup` Step 0.5 doesn't tell the agent what to do on the new exit 4

The prose covers exit 0 and exit 1 but not 4 — which is the one outcome that says *"clone is NOT protected"* and therefore the one most needing an operator response. Also not 3, though that's genuinely a no-op for the ops fork.

#### note — an **absolute** `--hooks-dir` exits 3, not 2

`--hooks-dir /abs/path` produces `"nothing to install"` (exit 3) rather than the LOW-4 containment error, because the nonsense concatenation `$TARGET_ROOT//abs/path` fails `[ -d ]` first. Refuses either way, nothing written — the message is just less accurate than the relative form's.

#### note — installing from a linked git worktree writes shared config

`--repo-dir <linked-worktree>` writes to the common `.git/config`, so all worktrees of that repo see it. Because the written value is **relative** (`.githooks`), it resolves per-worktree and is correct everywhere. Confirmed by probe; no change wanted, just flagging it since this framework runs many parallel worktrees.

---

### Checklist Results

- Secrets & Credentials: **Pass** — no credentials handled; written value is always the relative `$HOOKS_DIR_NAME`, and it is now validated for repo containment before writing.
- Injection Prevention: **Pass** — no shell-recursive expansion of `$RAW`/`$RESOLVED`.
- XSS Prevention: **N/A**
- Auth & Authorisation: **Pass** — no authz surface.
- Data Protection: **Pass**
- API Security: **N/A**
- **Gate & Trust-Chain Integrity: Pass** — the HIGH and both MEDIUMs from the prior pass are closed by execution; the pre-build gate's out-of-governance exemption retains its documented fail-closed direction under the refactor; `.claude/settings.json` adds one advisory entry and disturbs no gate.

### Verdict

**APPROVED**

The refactor that grew the scope did not weaken the gate it touched. I set out to prove a fail-open had been introduced by the security fix and could not — the property the #883 comment claims is the one the code exhibits, and the ticket gate's symlink defence survives even a full mutation of the deduplicated function, because an independent probe backstops it. All six prior findings verified fixed against my own repros rather than the author's report, with controls confirming the fixes didn't over-tighten. Suite re-run independently: 129/129, and the pre-build gate's own 78/78.

The three LOWs are worth a follow-up commit but none of them changes a security outcome, and per `right-size-ceremony.md` they don't justify holding a Heavy-tier PR that is now correct on every property it was blocked for.

---
📌 Reviewed commit: `03152a57d33d07800bb4088d23286d8c579d6c79`

atlas-apex (commented):
## Code Review: PR #1087 (round 2)

**Commit**: `03152a57d33d07800bb4088d23286d8c579d6c79`
**Previously reviewed**: `804aa8d4d38b8074afb31e718566ece4efc658cb` (CHANGES REQUESTED)

### Summary

One commit addressing all five of my findings plus Hakim's six. Scope grew: `require-active-ticket.sh` — the pre-build ticket gate, with its own 78-case suite — is now in the diff because finding #4 was fixed by extracting the shared `_resolve_real_path`. I reviewed that file as a first-time review, not a delta.

Every claim below is from execution against scratch repos, not from reading the author's report. All five of my findings are genuinely closed, and the lead concern raised for this round — a possible fail-open introduced *by* the deduplication — is **refuted** by four independent experiments.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** — all three previously-surviving mutants now killed, re-verified
- Security: **Pass** — degraded-mode behaviour proven fail-closed in every reachable path
- Performance: **Pass** (N/A)
- Summary Bullet Narrative: **Pass** — every bullet carries what + why
- Technical Decisions (AgDR): **N/A** — see the note below
- Adopter Handbooks: **Pass** — 4 loaded (2 public architecture, 1 public general, 1 private custom architecture); no findings
- Fallow Static Analysis: **N/A** — no JS/TS in diff

---

## 0. The lead question: did the deduplication convert a fail-closed gate into a fail-open one?

**No. Refuted.** Both consumers source `_lib-path-resolve.sh` conditionally with no `else`, exactly as described — so with the lib absent, `_resolve_real_path` is undefined and every call substitutes empty. I did not reason about this; I renamed the lib and ran the code.

**`require-active-ticket.sh` — the out-of-governance exemption.** The consuming block is guarded:

```sh
_og_real_target="$(_resolve_real_path "$_og_abs_target")"
if [ -n "$_og_real_target" ]; then
  ...  # all four containment checks, then `return 0` (exempt)
fi
```

An empty result skips the entire exemption block, so the write falls through to the ticket gate and **blocks**. Empty resolves toward *gated*, not toward *exempt* — the opposite of the failure mode feared. Measured, with the session pin disabled so ops-root resolution stayed inside the sandbox:

| Target | Lib present | Lib removed |
|---|---|---|
| out-of-governance (`<scratch>/outside/machine.conf`, no git repo above it) | `rc=0` exempt | **`rc=2` blocked** |
| inside ops fork (`<ops>/src/app.ts`, no ticket) | `rc=2` blocked | `rc=2` blocked |

This is the behaviour the code's own comment claims (*"An unresolvable FILE_PATH (empty …) is NEVER exempted here"*), and it is now enforced by the guard rather than by the function's presence.

**Independent second proof, from the suite itself.** I mutated the *test harness* — removed the two `cp "$LIB_PATH_RESOLVE"` lines the PR added — to check whether that harness change is load-bearing. It is, and the shape of the failures is the answer to this question:

```
PASS: 72   FAIL: 6
FAIL [bash redirect to /tmp exempt (no ticket needed)]:      want rc=0, got 2
FAIL [bash redirect to /var/tmp exempt]:                     want rc=0, got 2
FAIL [#883 Edit-tool write to out-of-repo dotfile exempt]:   want rc=0, got 2
FAIL [#883 MultiEdit-tool write to out-of-repo dotfile]:     want rc=0, got 2
FAIL [#883 bash relative write from out-of-repo CWD exempt]: want rc=0, got 2
FAIL [#886 regression: single out-of-repo target exempt]:    want rc=0, got 2
```

Six distinct cases, all `want rc=0, got 2` — **every one over-blocks; none under-blocks.** A fail-open would have shown up as the mirror (`want rc=2, got 0`) and does not appear anywhere.

**`_lib-git-hooks-path.sh` / `bin/install-git-hooks.sh`.** Here the classifier *does* degrade badly in isolation — with `_resolve_real_path` undefined, `ghp_classify` returns `missing:<raw>` for a state that is really `third-party`, and `missing:` is an auto-repair branch. But the installer never reaches the classifier: it resolves paths itself first, and an empty resolve trips its own guards.

```
# deliberate external third-party core.hooksPath configured, lib removed:
$ bash bin/install-git-hooks.sh
  ERROR: --hooks-dir '.githooks' resolves to '', which is outside <repo>. Refusing…
  rc=2   config: <company-shared-hooks>   <- untouched

$ bash bin/install-git-hooks.sh --repo-dir <repo>
  ERROR: --repo-dir … is not itself a git repository root…
  rc=2   config: <company-shared-hooks>   <- untouched
```

So the safety guarantee holds end-to-end even in the degraded state. The advisory hook exits 0 on every path including both libs removed (verified), which is correct for an advisory.

**On `$(dirname "$0")`:** not reachable in production. `.claude/settings.json` invokes the hook as `exec "$r/.claude/hooks/require-active-ticket.sh"` — always an absolute path. The pattern is also pre-existing, used at lines 267 and 604 of the *base* commit for `_lib-ops-root.sh` and `_lib-detect-bash-write.sh`; this PR didn't introduce it. `_lib-git-hooks-path.sh` correctly uses `${BASH_SOURCE[0]:-$0}` instead, which is the right form for a sourced lib and the reason it resolves correctly when sourced from `bin/`.

**Verdict on the lead question:** the deduplication is not a fail-open. It is strictly better than the duplication it replaced — the extraction is provably behaviour-identical (below), and the degraded mode is fail-closed in all four paths I could construct.

---

## 1–5. The original findings, each re-verified

### #1 — `/handover` invocation dropped — **CLOSED**

`grep -rn install-git-hooks.sh` across the repo: the only invocations are `/setup` (retained, correct — the ops fork configuring its own `.githooks/`), the advisory hook's banner text, and `docs/getting-started.md`. `handover/SKILL.md` has no invocation; the whole `### 1.5-git-hooks` step is deleted and replaced with a prose deferral note.

**On the risk direction** (Hakim's point that the old text framed exit 3 as the boring case and exit 0 as success — i.e. backwards): the new text does not discuss exit codes at all, because there is no longer a call to branch on. It names the *invocation itself* as the hazard ("this is exactly the wrong moment to hand it code execution"), states the deferred gap honestly, prescribes the correct alternative mechanism (install our own hook into the clone's untracked `.git/hooks/`), and closes with "Do not attempt it here." That is the correct direction.

### #2 — fail-closed on `git config` writes — **CLOSED**

Exercised with a real `.git/config.lock`, not a mock:

```
rc=4
ERROR: 'git config core.hooksPath' failed (exit 255): error: could not lock config file .git/config: File exists
ERROR: failed to set core.hooksPath in <repo> — clone is NOT protected…
config after: ""          false-success lines: 0
```

**The read-back exists and is independently load-bearing.** I split `_ghp_set_hookspath` into two variants to check whether the exit-status check alone was doing all the work:

| Variant | Result under lock |
|---|---|
| read-back removed, exit-status check kept | `rc=4`, no false success |
| **exit-status check removed, read-back kept** | **`rc=4`** — `post-write state: 'unset'. Not reporting success.` |

Either check alone catches it. Genuine defence in depth, not a decorative second check.

### #3 — `--repo-dir` no longer retargets — **CLOSED**

```
$ bash bin/install-git-hooks.sh --repo-dir <outer>/workspace/proj-not-a-repo
  ERROR: … is not itself a git repository root — it resolved to the ENCLOSING repo at '<outer>' instead. Refusing…
  rc=2    outer config: ""   <- untouched
```

I also checked the obvious way to get this wrong: a **symlinked** path to the real repo root must not false-positive. It doesn't (`rc=0`, configured correctly) — the check compares resolved forms on both sides.

### #4 — exactly one `_resolve_real_path` — **CLOSED**

One definition repo-wide, at `.claude/hooks/_lib-path-resolve.sh:40`. Three call sites in `bin/install-git-hooks.sh`, one in `require-active-ticket.sh`, three in `_lib-git-hooks-path.sh`. No third copy survives.

**The extraction is provably behaviour-preserving.** I diffed the function removed from `require-active-ticket.sh` at `804aa8d` against the one added to the new lib: with comments and blank lines stripped, the two are **byte-identical**. The only textual difference is two dropped explanatory comments. That matters here — this is the pre-build gate, and a silent semantic change during a "pure move" would be the expensive kind of mistake.

**On `_portfolio_canonicalize`:** leaving it separate is **defensible, and folding it in would have been wrong for this PR.** It is not the same function: it fails *soft* (echoes input unchanged) where this one fails *empty*, it carries the #1018 Windows drive-letter normalisation, and it short-circuits non-absolute input via `*) echo "$p"; return 0`. Its callers depend on the fail-soft contract; merging the two would be a behaviour change to the portfolio-path resolver riding along inside an installer PR.

**Is the divergence documented, or still only comment-asserted?** It is now genuinely documented, and in the right direction — `_lib-path-resolve.sh`'s header names `_portfolio_canonicalize` explicitly, names the Windows fix, and names fail-soft-vs-fail-empty as the deliberate difference, ending with "Do not fold the two together." That inverts the previous state: the old comment asserted *sameness* that a sibling had already broken; the new one documents the *difference*. Good fix.

One residual, below.

### #5 — the three mutants — **CLOSED, all three re-applied and confirmed RED**

Baseline: installer 41/41, advisory 8/8.

| Mutant | Installer | Advisory | Killed? |
|---|---|---|---|
| `_resolve_real_path` → bare passthrough | **PASS 16 / FAIL 25** | 8/0 | **yes** |
| advisory `third-party:*` branch deleted | 41/0 | **PASS 7 / FAIL 1** | **yes** |
| classifier external-third-party fallthrough → `unset` | **PASS 38 / FAIL 3** | **PASS 7 / FAIL 1** | **yes** |
| *(restored)* | 41/0 | 8/0 | — |

Every number matches the PR body's claim exactly. Mutant 3 is the important one — that branch guards the "never clobber a deliberate value without `--force`" promise, and it had zero coverage before this round.

### The three LOWs — all verified

| LOW | Result |
|---|---|
| `--hooks-dir ../evil-hooks` traversal | `rc=2`, refused, config untouched |
| `.githooks` committed as a symlink | `rc=2`, refused, config untouched (`[ -L ]` checked before `[ -d ]`) |
| advisory banner nudging toward `--force` | `grep -- --force check-git-hooks-installed.sh` → **no matches**; banner now points at `--help` |

---

## Verification

Full suite re-run myself in four chunks (~80–100s each), not taken on report:

```
CHUNK[0..32]    PASS=33 FAIL=0
CHUNK[33..65]   PASS=33 FAIL=0
CHUNK[66..98]   PASS=33 FAIL=0
CHUNK[99..128]  PASS=30 FAIL=0
------------------------------
TOTAL           PASS=129 FAIL=0    (129 tests discovered — no failures, no quarantine)
```

`require-active-ticket.sh`'s own suite, run separately because it is the pre-build gate: **PASS=78 FAIL=0**.

`shellcheck --severity=warning` clean on all eight touched shell files. `.claude/settings.json` is byte-unchanged since my last review (no delta in this commit), so the wiring I verified last round still stands. Nothing enumerates hook libs by filename (no packaging/sync list needed updating for the new file); the lib is tracked; file mode 644 is within the existing mixed convention for `_lib-*.sh`.

---

## Handbook Findings

Four handbooks loaded (path-convention; semantic supplement returned no additional handbook matches):

- `handbooks/architecture/migration-safety.md` **(blocking)** — **N/A**, no schema or data migration in the diff.
- `handbooks/architecture/clean-architecture-layers.md` — N/A, no domain/application/infrastructure layering in a shell diff.
- `/Users/ahmed/apex/apexyard-portfolio/custom-handbooks/architecture/clean-architecture-layers.md` (private layer) — N/A, same reason.
- `handbooks/general/commit-message-quality.md` — **Pass**. Subject is meaningful, body is substantial and WHY-focused (it explains the *proven* attack for the HIGH fix rather than restating the change), no placeholders, no one-liner-restating-subject pattern.

No handbook findings.

---

## Remaining notes (non-blocking)

### nit: the conditional source is silent about its own failure, and one degraded-mode message is misleading

The `if [ -f … ]` / no-`else` shape is **safe** — I proved that above — and for these two consumers silent degradation is arguably the right design: hard-failing `require-active-ticket.sh` on a missing lib would brick the ticket gate, and the advisory hook must never block. So I'm not asking you to change the control flow. Two smaller things:

1. There's an unexplained asymmetry: `bin/install-git-hooks.sh` guards its lib load with an explicit `exit 2` and a clear message, while the two hooks guard theirs with a silent no-op. That difference is deliberate and correct, but nothing says so. One sentence in each `if` block ("degrades to fail-closed rather than bricking the gate — see …") would stop a future reader from "fixing" it into a hard failure.
2. In the degraded path, `--repo-dir` prints `--repo-dir 'X' … resolved to the ENCLOSING repo at 'X'` — the same path twice, which reads as nonsense. Cheap to make diagnosable.

### nit: `_lib-portfolio-paths.sh:110` now points at a function that isn't there

That comment still reads *"Algorithm mirrors `_resolve_real_path()` in require-active-ticket.sh"*. After this PR the function lives in `_lib-path-resolve.sh`. The file is outside this PR's diff, so I'm not asking for it here — but the stale pointer is a direct consequence of the move, and it's the same comment I flagged last round as asserting an invariant nothing enforces. Worth a one-line follow-up so the two halves of the documented divergence point at each other.

### note: AgDR — N/A, with the reasoning stated

Consistent with my last round. This PR implements AgDR-0104's existing decision and adds no dependency, technology, service, or schema change. The one thing that gave me pause is `right-size-ceremony.md` rail 1 (trust-chain work is always Heavy) — which I honoured by treating this as a Heavy review, but Heavy *review* is not the same as *a new decision to record*. The extraction is a review-directed refactor with proven-identical behaviour, reversible inside this PR, and it implements rather than originates the "share the definition, don't assert it in a comment" position (AgDR-0113). No AgDR needed.

---

### Verdict

**APPROVED**

Both blockers from my last round are genuinely closed, and closed the way I'd hoped rather than the way that merely satisfies the wording: the duplication was resolved by *sharing* (with the one legitimate non-sibling kept separate and its divergence now documented), and the three test gaps were closed with cases I re-verified kill their mutants. Hakim's HIGH is closed by removal, which was the right call for a step-1 PR.

The specific risk raised for this round — that deduplicating a fail-closed security helper had quietly made it fail-open — was worth chasing and is **not real**. The empty-resolve path is gated at every consumer, and the strongest evidence isn't my probe: it's that removing the lib from the test harness produces six failures that are *all* over-blocks and *zero* under-blocks.

The two nits above are documentation-quality follow-ups, not merge blockers.

---
📌 Reviewed commit: `03152a57d33d07800bb4088d23286d8c579d6c79`

atlas-apex:
The four LOW findings from both reviews are filed as #1089 rather than folded in here — a documentation-only commit would cost a full re-review round for zero behaviour change, and round three produced no new defects, so this is a good place to stop.

Grouped rather than split because they share a cause: this PR **removed** a capability (the `/handover` call) and **added** a degraded mode (the conditionally-sourced lib), and the surrounding docs and diagnostics weren't updated to match either.

Flagging the one that isn't hygiene: `docs/getting-started.md:158` still says "`/setup` and `/handover` call it". That is now the only place in the repo describing the removed behaviour as intended, and it is exactly what someone would cite as prior design intent when re-adding the call that was removed on the HIGH. #1089 lists it first and says to do it first if the ticket ever gets split.

Natural home is step 2 of #1086, which touches this same area.

me2resh (approved):

Issue #1085, opened 2026-07-30
[Bug] block-main-push.sh: a tag push prefix makes git push origin HEAD bypass the protected-branch check


atlas-apex:
## Bug Scenario

**Given** a command containing a genuine tag push followed by a bare `git push origin HEAD`

**When** `block-main-push.sh` evaluates whether any occurrence needs the protected-branch check

**Then** `_command_has_untagged_refless_push` classifies the `HEAD` occurrence as "has an explicit ref — already covered by scan-all", the check is skipped, and the hook exits **0** while the push lands on the protected branch

**Expected** `HEAD` should be treated as *no ref*, matching what `_extract_all_push_refs_core` already does

## Repro Steps

On a repo checked out on `main`:

```bash
git push origin HEAD                              # rc=2, blocked (control)
git push origin --tags && git push origin HEAD    # rc=0, ALLOWED — the push executes
```

Prefixing an otherwise-blocked push with a tag push flips it to allowed. Verified with a stubbed `git` on PATH, so the push genuinely executes rather than merely passing the gate.

## Severity

**high** — a silent bypass of "no direct pushes to main", reachable without heredocs, decoys, or unusual syntax.

**Net-neutral against `dev`**, measured across history: `dev`=0, `bc45eea`=2, `8e55c2c`/`75d9c53`=0. The "exactly one occurrence" rule in `bc45eea` closed it incidentally; the per-occurrence redesign that replaced it (necessary — the old rule false-blocked legitimate two-remote tag pushes) reopened it. So merging #1075 introduces no bypass that is not already on `dev`.

## This is NOT the same defect as #1084 — read this before fixing either

The two look alike and have **different mechanisms**. #1084's stated fix will not close this one, and closing #1084 could reasonably be read as handling the class.

| | #1084 | This ticket |
|---|---|---|
| `is_tag_push` | **false** (non-tag decoy) | **true** (genuine tag push present) |
| Predicate called? | **no** — gated behind `if is_tag_push` | **yes** |
| Failure | predicate never runs | predicate runs and **misclassifies** |
| Fix | hoist the predicate out from behind the condition | align its ref-detection with scan-all's |

[...]

Issue #1084, opened 2026-07-30
[Bug] block-main-push.sh: a decoy ref in an unstripped heredoc hides a ref-less push to a protected branch


atlas-apex:
## Bug Scenario

**Given** a command containing a decoy ref inside an **unconfirmed** heredoc (one the stripper declines to strip), followed by a **ref-less** `git push`

**When** `block-main-push.sh` decides whether the push targets a protected branch

**Then** `extract_push_ref` returns the decoy from the unstripped body, the protected-branch check runs against that instead of the real destination, and the hook exits **0** while the push lands on the protected branch

**Expected** a decoy in unstripped prose must not decide which branch gets checked

## Repro Steps

On a repo checked out on `main`:

```bash
git push                       # rc=2, blocked (control)

cat > /tmp/m.txt <<END.OF
see git push origin feature/GH-1-safe
END.OF
git push                       # rc=0, ALLOWED — the push to main executes
```

The control blocking is what makes this a bypass rather than a match failure. Verified twice, independently, by both the code and security reviews of #1075.

## Severity

**high** — a silent bypass of "no direct pushes to main". Pre-existing: `dev` behaves identically, so this is not a regression from #1075.

Point 4's scan-all backstop does **not** cover it: scan-all only ever sees *explicit* refs, and a ref-less push has none. The local-HEAD fallback that would catch it sits inside the block the exemption skips.

## The fix is one line, and the machinery already exists

`_command_has_untagged_refless_push` (added in #1075) **already detects this shape**. It is merely gated behind `if is_tag_push`, which is false for a non-tag decoy.

Measured during review:

| Shape | predicate | `is_tag_push` |
|---|---|---|
| ref-decoy + bare push | **TRUE** | false |
| legitimate compound (ordinary push + tag push) | false | — |

So hoisting the predicate out from behind that condition closes the hole **without** reintroducing the over-narrowing that #1075's round 6 fixed. Add the ref-decoy + bare-push shape as a test, RED first.

## Why this was not fixed in #1075

#1075 ran nine review rounds and was stopped deliberately. Each round's fix generated the next round's finding — the reviewers' own diagnosis was that each fix created a new pair of functions that had to agree, with the agreement asserted in a comment rather than enforced in code. Continuing was no longer cheaper than stopping.

[...]

Issue #1083, opened 2026-07-30
[Bug] block-main-push.sh: a redirect or pipe before a push truncates extraction, bypassing the gate


atlas-apex:
## Bug Scenario

**Given** a push chained after any command containing a redirect or pipe

**When** `block-main-push.sh` extracts push refs to decide whether the destination is protected

**Then** the extractor's line-truncation means **no refs are returned at all**, the protected-branch check has nothing to test, and the hook exits **0** while the push executes

**Expected** a redirect or pipe earlier in the command should not blind the gate to a later push

## Repro Steps

On a repo whose current branch is not protected:

```bash
echo x > /tmp/f && git push origin main     # rc=0, allowed — the push executes
git push origin main                        # rc=2, blocked (control)
```

The control blocking is what makes this a bypass rather than a match failure. Verified for `>` and `|`.

## Severity

**medium-high** — a silent bypass of "no direct pushes to main", reachable with no adversarial intent. Redirecting output before a push is ordinary shell.

Not rated higher only because the bare form is more common than the chained one, so it does not fire constantly.

## Why this is filed rather than fixed in #1075

#1075 is seven review rounds deep on heredoc-prose parsing and has already grown three times — absorbing the decoy-ref hijack, the `is_tag_push` control-flow bug, and the quoted-ref normalisation (#1079). This defect is **pre-existing**, reproduces identically on `dev`, and is a different mechanism: line truncation in the extractor core, not heredoc handling.

Both reviewers on #1075 independently identified it and both agreed it was out of scope there. It is documented as explicit residue in `AgDR-0113`.

## Investigation Notes

The truncation lives in `_extract_push_ref_core`, and the newer `_extract_all_push_refs_core` (added in #1075 as a heredoc-independent backstop) **inherits it** — so the backstop is not independent of this particular hole. That is worth stating plainly to whoever fixes it: adding another scanner on the same core will not help.

Relevant constraint from the same file's history: do **not** attempt this by widening a terminator set. That was tried in #1040 and rejected for reopening #227 (9 failures vs 5). Any terminator general enough to mean "a new command starts here" also amputates bodies containing that text as prose. The fixes that have worked in this file changed *how a captured value is trimmed* or *which occurrences are evaluated* — not what counts as a boundary.

See also `AgDR-0113`, which now carries two governance rules worth applying to any fix here: a coverage claim must cite the test proving it or be written as residue, and it must state the scope it holds at, inheriting any dimension the cited tests hold constant.

## Related

[...]

Issue #1081, opened 2026-07-30
[Bug] validate-branch-name.sh's is_tag_push exemption can hide a non-conforming push behind decoy tag evidence


atlas-apex:
## Bug Scenario

**Given** a Bash command containing a real, non-conforming `git push` destination alongside decoy text carrying tag-push evidence (`--tags`, `tag <name>`, or `refs/tags/`) that is either inside a heredoc `strip_heredoc_bodies` correctly declines to strip (unconfirmed delimiter), or not inside a heredoc at all
**When** `validate-branch-name.sh` runs `is_tag_push` against the raw command
**Then** `is_tag_push` returns true (the decoy evidence matches its global checks) and the hook `exit 0`s, skipping validation of the real push destination entirely
**Expected** the hook should validate the real push's destination against the naming convention regardless of decoy tag-evidence text elsewhere in the command

## Repro Steps

```bash
cat > /tmp/m.txt <<END.OF
we always use git push --tags for releases
END.OF
git push origin bogus-branch
```

Piped as a `PreToolUse` payload to `validate-branch-name.sh` (local branch anything), this exits `0` (allowed) instead of blocking `bogus-branch` for not following the `{type}/{TICKET-ID}-{description}` convention.

## Environment

apexyard framework, `.claude/hooks/validate-branch-name.sh` + `.claude/hooks/_lib-extract-push-ref.sh`. Not environment-specific — pure text-matching logic.

## Severity

P2 — a naming-convention miss, not a protected-branch bypass. `main`/`dev` are trunk-whitelisted in `validate-branch-name.sh` (naming rules don't apply to them at all), and "don't push directly to a protected branch" is `block-main-push.sh`'s job, which independently closes the dangerous version of this same root cause (see me2resh/apexyard#1075, decision points 5–6). This ticket is the naming-gate's parallel, lower-severity exposure.

## Mitigation

None currently — write the heredoc-writing step and the `git push` step in separate Bash calls to avoid the decoy interacting with real push validation at all.

## Investigation Notes

[...]

PR #1090, opened 2026-07-30
fix(#1086): enforce protected branches from git's pre-push ref list

atlas-apex:
## Summary

- **`.githooks/pre-push` now decides "block this push?" from git's own stdin ref lines (`<local_ref> <local_sha> <remote_ref> <remote_sha>`), never from the shell command text.** This is step 2 of #1086 and closes the whole bug class AgDR-0113 documents: #1066, #1081, #1083, #1084, #1085 are all instances of one root cause — a hook deciding by pattern-matching the *text* of a shell command, which AgDR-0104 already concluded cannot be made sound. At the git layer there is no command text to misparse: no heredocs, no decoys, no redirects, no quoting shapes, because nothing here scans a command string at all.
- **Deletion pushes (`local_sha` = 40 zeroes) are blocked identically to updates.** Deleting a protected branch is at least as destructive as pushing to one, and this closes a real, pre-existing gap: `.claude/hooks/block-main-push.sh`'s `extract_push_ref` deliberately bails out on `--delete`/`-d` (returns empty), which falls back to validating whatever branch is currently checked out rather than the branch actually being deleted — so `git push origin --delete main` run from a feature-branch checkout was never actually validated there. Verified with a real push (see Testing).
- **New `.claude/hooks/_lib-protected-branches.sh` centralises the protected-branch-list resolution** so the new git-native hook doesn't re-derive it independently. **`block-main-push.sh` is intentionally left untouched** — it still resolves its own copy of the same list inline (identical algorithm and `main|master|dev|develop` fallback). Demoting it to source this lib, and to advisory status generally, is step 3 of #1086 and must land only after this enforcement layer is proven installed-by-default (step 1, merged as #1087) and blocking (this PR). Until step 3, this is a deliberate, temporary duplication of a simple config-resolution algorithm — not the dangerous "two disagreeing text-parsers" pattern AgDR-0113 spent five review rounds on, since both copies use the identical resolution and fallback and `block-main-push.sh`'s role is unchanged.
- **`--force` / `--force-with-lease` don't need special-casing.** `remote_ref` and `local_sha` are the same regardless of the force flag, so a force push to a protected branch is blocked exactly like an ordinary one — a correctness property this design gets for free rather than one it has to parse for.

## What this PR does NOT do

- **`block-main-push.sh` remains the active control** for Claude-Code-driven pushes (it's a `PreToolUse` hook; this git-native layer only fires for pushes that actually reach git — i.e., where `core.hooksPath` is set, per step 1). It is not touched, not demoted, and not disabled here.
- **#1081, #1083, #1084, #1085 are not closed by this PR** — they're bugs in `block-main-push.sh`'s text-parsing, and that file isn't touched. They close in step 3, when `block-main-push.sh` is demoted to advisory (at which point its residual text-matching bugs become cosmetic misses rather than security bypasses, per the issue's own reasoning).

## Testing

**Discriminating pairs — every "must block" case proven to FAIL against the pre-change hook and PASS against this one** (not just green-in-both-directions, which proves nothing):

| Case | Before (pre-#1086-step-2) | After (this PR) |
|---|---|---|
| Bare `git push` on `main` | ALLOWED — exit 0, remote `main` advanced | **BLOCKED** — exit ≠0, remote `main` unchanged |
| Explicit `git push origin main` | ALLOWED | **BLOCKED** |
| `git push origin --delete main` (deletion) | ALLOWED — remote `main` ref actually deleted | **BLOCKED** — remote `main` intact |
| #1083 shape (redirect before the push) | ALLOWED | **BLOCKED** |
| #1084 shape (decoy ref in unstripped heredoc, then a bare push) | ALLOWED | **BLOCKED** |
| #1085 shape (tag-push prefix, then `git push origin HEAD`) | tag push AND the `HEAD` push to `main` both succeeded | tag push succeeds, `HEAD` push to `main` **BLOCKED**, remote `main` unchanged |
| #1066 shape (heredoc prose merely discussing `git push`, then a real push to `main`) | ALLOWED | **BLOCKED** |

For the delete case, I first had to disable the bare test remote's own `receive.denyDeleteCurrent` — without that, the remote's *own*, unrelated safety net masked whether our local hook was doing anything, since it would reject the delete either way. With it neutralised, the "before" run genuinely deletes `main` on the remote (confirmed: ref goes from a real SHA to "MISSING"), and the "after" run blocks locally before any data reaches the remote.

**Over-narrowing controls (regression checks, not novelty proofs — these already passed pre-change since the old hook blocked nothing, and exist to catch this PR accidentally blocking legitimate pushes):**

- Genuine tag push — ALLOWED, both before and after
- Ordinary feature-branch push — ALLOWED, both before and after
- Two-remote tag push (`git push origin --tags && git push upstream --tags`) — ALLOWED, both before and after
- Deleting a non-protected branch — ALLOWED, both before and after

**Why real git + a real local bare remote, not a stubbed `git` on `PATH`.** The "stub git" standard from prior reviews was written for `block-main-push.sh`, a `PreToolUse` hook that decides from a *simulated* command string with no separate execution step — a stub is the only way to prove the downstream command would really have reached git. `.githooks/pre-push` has no such simulated step: it runs *inside* the real git-push lifecycle, invoked by git with git's own resolved ref lines. A stub here would have to reimplement git's own refspec/tag/HEAD/upstream-tracking resolution to hand the hook correct stdin — exactly the "reimplementation that has to be trusted to agree with the real thing" pattern AgDR-0113 argues against. A real git binary against a real local bare remote, asserting on the remote's actual post-push ref state, is a *stricter* proof of "the push genuinely executes or is genuinely rejected," not a weaker one.

**Test suite:** `.claude/hooks/tests/test_githooks_pre_push_protected_branch.sh` (11 cases, 30 assertions) is the automated form of the table above, run as part of `bin/run-hook-tests.sh`. Full suite: 130/130 PASS, 0 FAIL (chunked locally — the suite includes `test_block_main_push.sh` and `test_block_main_push_heredoc.sh`, both still green, confirming `block-main-push.sh` is genuinely untouched).

**Real installer + real push, end to end:** ran `bin/install-git-hooks.sh --repo-dir <scratch-repo>` (the actual installer from step 1, not a manual `git config`), confirmed `core.hooksPath` was set, then performed a real `git push origin main` (BLOCKED, exit 1, remote unchanged) and a real `git push -u origin feature/...` (ALLOWED, exit 0, remote ref created).

**Stdin hand-off (the specific risk called out in the brief):** the old hook ended in `exec bash "$RUNNER"`, which passed stdin through untouched. This PR reads stdin (`STDIN_REFS=$(cat)`) before that point. I audited `bin/run-pre-push-checks.sh` and everything it shells out to (`markdownlint-cli2`, `shellcheck` via `find | xargs`, `test_subpack_extraction.sh`) — none of them read from fd 0; the only `read` calls anywhere in that call chain are against `<(find ...)` process substitutions, never the inherited stdin. Confirmed empirically too: ran `bin/run-pre-push-checks.sh` with `< /dev/null` (no hang, clean exit), and ran a full real push through a sandbox using the **unstubbed** `run-pre-push-checks.sh` — it executed all three checks in order and produced the expected pass/fail output, proving the hand-off is safe in practice, not just by inspection. `exec` is kept at the final delegation specifically so the hook's own exit status still comes from process replacement, unchanged from before.

**No command-text parsing anywhere in the decision:** the only `grep` in `.githooks/pre-push` matches `$BRANCH` (derived solely from `$REMOTE_REF`, itself one of git's own stdin fields) against the static protected-branch regex. There is no `$COMMAND` variable, no `tool_input`, no shell-command scanning of any kind in this file.

**Shellcheck:** `--severity=warning` clean on all three touched/added files.

## Glossary

| Term | Definition |
|------|------------|
| `pre-push` hook | A git-native hook git invokes before a push completes, with the refs being pushed written to its stdin, already resolved — no command text to parse. |
| stdin ref line | `<local ref> SP <local sha1> SP <remote ref> SP <remote sha1> LF` — one line per ref being pushed, printed by git itself. |
| all-zeroes sentinel SHA | `local_sha` = 40 zeroes means the ref named on that line is being **deleted**, not updated. |
| protected branch | A branch (`main`/`master`/`dev`/`develop` by default, configurable via `.claude/project-config.json` → `.git.protected_branches[]`) that this hook refuses to push to or delete directly. |
| `core.hooksPath` | Per-clone git config pointing at this repo's tracked hooks dir (`.githooks/`) — set by `bin/install-git-hooks.sh` (step 1, #1087). Without it, this hook never runs. |

Refs #1086

### DIFF
--- a/.claude/hooks/_lib-protected-branches.sh
+++ b/.claude/hooks/_lib-protected-branches.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+# _lib-protected-branches.sh — single source of truth for the protected-
+# branch list, shared by every consumer that needs to answer "is branch X
+# one we refuse to push directly to". me2resh/apexyard#1086 step 2.
+#
+# Not a hook itself (the `_lib-` prefix keeps it out of hook wiring, same
+# convention as every other `_lib-*.sh` in this directory). Source it and
+# call protected_branch_regex to get a `|`-joined alternation suitable for
+# `grep -qE "^(<result>)$"`.
+#
+# WHY THIS EXISTS
+# ---------------
+# `.claude/hooks/block-main-push.sh` has resolved this list inline since
+# me2resh/apexyard#109 (config_get '.git.protected_branches[]', falling back
+# to main|master|dev|develop when config is absent or empty). #1086 adds a
+# SECOND consumer — `.githooks/pre-push`, the git-native enforcement layer —
+# that needs the exact same list. Re-deriving it a second time inline would
+# be precisely the failure mode PR #1087's review flagged as this repo's
+# dominant defect generator: two copies of the same resolution that have to
+# be trusted to agree, with the agreement asserted in a comment rather than
+# enforced in code (see AgDR-0113's whole "additive vs subtractive" /
+# "duplicated function" history for how expensive that pattern has been in
+# this exact file family).
+#
+# block-main-push.sh itself is intentionally NOT touched by #1086 step 2 —
+# demoting it to advisory is step 3, and must land only after this git-
+# native enforcement is proven installed-by-default (step 1) and blocking
+# (this file + .githooks/pre-push, step 2). Until step 3 migrates
+# block-main-push.sh to source this lib too, its own inline copy remains —
+# that overlap is a known, deliberate, temporary consequence of not touching
+# it here, not a second FUTURE source of truth: this file is written to be
+# the thing step 3 migrates block-main-push.sh onto, not a competing
+# implementation.
+#
+# Resolution order (identical to block-main-push.sh's inline version):
+#   1. .claude/project-config.json / .defaults.json -> .git.protected_branches[]
+#      (via _lib-read-config.sh, if available)
+#   2. Fallback: main|master|dev|develop
+#
+# Usage:
+#   . "$(dirname "$0")/_lib-protected-branches.sh"
+#   REGEX=$(protected_branch_regex)
+#   echo "$branch" | grep -qE "^(${REGEX})$" && echo "protected"
+
+# ------------------------------------------------------------------------
+# protected_branch_regex
+#
+# Prints a `|`-joined alternation of protected branch names to stdout (no
+# trailing newline guaranteed either way — callers should use it inside
+# `$(...)`, which strips trailing newlines regardless).
+# ------------------------------------------------------------------------
+protected_branch_regex() {
+  local hook_dir repo_root regex
+
+  hook_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
+  repo_root=$(git rev-parse --show-toplevel 2>/dev/null)
+
+  regex=""
+  if [ -n "$repo_root" ] && [ -n "$hook_dir" ] && [ -f "$hook_dir/_lib-read-config.sh" ]; then
+    # shellcheck disable=SC1090,SC1091
+    . "$hook_dir/_lib-read-config.sh"
+    regex=$(config_get '.git.protected_branches[]' 2>/dev/null | paste -sd'|' -)
+  fi
+
+  if [ -z "$regex" ]; then
+    regex="main|master|dev|develop"
+  fi
+
+  echo "$regex"
+}
--- a/.claude/hooks/tests/test_githooks_pre_push_protected_branch.sh
+++ b/.claude/hooks/tests/test_githooks_pre_push_protected_branch.sh
@@ -1,0 +1,446 @@
+#!/bin/bash
+# Tests for .githooks/pre-push's protected-branch enforcement — the git-
+# native layer added in me2resh/apexyard#1086 step 2.
+#
+# WHY REAL GIT, NOT A STUBBED `git` ON PATH
+# ------------------------------------------------------------------------
+# The prior-review standard cited in the #1086 brief ("use a stubbed git on
+# PATH so the push genuinely executes") was written for
+# .claude/hooks/block-main-push.sh — a Claude-Code PreToolUse hook that
+# decides allow/deny from a *simulated* tool_input.command, with the actual
+# command never run by the test harness itself. There, a git stub is the
+# only way to prove the DOWNSTREAM bash -c "$COMMAND" would really have
+# reached (or not reached) git.
+#
+# .githooks/pre-push has no such simulated downstream step — it runs INSIDE
+# the real git-push lifecycle, invoked by git itself with git's own
+# resolved ref lines on stdin. A stub here would have to reimplement git's
+# refspec/tag/HEAD/upstream-tracking resolution to hand the hook correct
+# stdin — exactly the class of "reimplementation that has to be trusted to
+# agree with the real thing" AgDR-0113 spent five review rounds arguing
+# against. Running a REAL git binary against a REAL local bare "remote" and
+# asserting on the remote's actual ref state after each push is a STRICTER
+# proof of "the push genuinely executes or is genuinely rejected" than a
+# stub could give, not a weaker one: there is nothing to reimplement, and
+# nothing left to disagree with the real thing by construction.
+#
+# EACH "must block" CASE IS DISCRIMINATING BY CONSTRUCTION
+# ------------------------------------------------------------------------
+# Before #1086 step 2, .githooks/pre-push contained ZERO protected-branch
+# logic — it only ever delegated to bin/run-pre-push-checks.sh (lint/format
+# checks). Every "must block" case below therefore FAILS against the
+# pre-change file (the push succeeds, the remote ref advances) and PASSES
+# only once the new stdin-reading block exists. The "must NOT block" cases
+# are regression controls, not novelty proofs — they already passed against
+# the pre-change file (which blocked nothing) and exist here to catch
+# OVER-narrowing introduced by this change, the mirror failure mode two
+# prior predicates in block-main-push.sh were redesigned for.
+#
+# Cases:
+#   1.  Feature-branch push                          -> NOT blocked
+#   2.  Genuine tag push                              -> NOT blocked
+#   3.  Two-remote tag push                           -> NOT blocked (both)
+#   4.  Delete a NON-protected branch                 -> NOT blocked
+#   5.  Bare `git push` while checked out on `main`   -> BLOCKED
+#   6.  Explicit `git push origin main`               -> BLOCKED
+#   7.  `--delete main` (deletion of protected branch) -> BLOCKED, remote main intact
+#   8.  #1083 shape: redirect before the push          -> BLOCKED, remote main intact
+#   9.  #1084 shape: decoy ref in unstripped heredoc,
+#       then a bare (ref-less) push                    -> BLOCKED, remote main intact
+#   10. #1085 shape: tag-push prefix + `git push origin HEAD` on main
+#                                                        -> tags succeed, HEAD push BLOCKED
+#   11. #1066 shape: heredoc prose merely discussing
+#       `git push`, then a real push to main            -> BLOCKED, remote main intact
+#
+# Exit 0 if all cases pass; 1 on any failure.
+
+set -u
+
+ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+REAL_PRE_PUSH="$ROOT/.githooks/pre-push"
+REAL_PROTECTED_LIB="$ROOT/.claude/hooks/_lib-protected-branches.sh"
+REAL_READ_CONFIG_LIB="$ROOT/.claude/hooks/_lib-read-config.sh"
+
+PASS=0
+FAIL=0
+FAILED=""
+
+assert_eq() {
+  local label="$1" expected="$2" actual="$3"
+  if [ "$expected" = "$actual" ]; then
+    echo "PASS [$label]"
+    PASS=$((PASS+1))
+  else
+    echo "FAIL [$label] — expected '$expected', got '$actual'" >&2
+    FAIL=$((FAIL+1)); FAILED="${FAILED}${label} "
+  fi
+}
+
+assert_match() {
+  local label="$1" pattern="$2" actual="$3"
+  if printf '%s' "$actual" | grep -qE "$pattern"; then
+    echo "PASS [$label]"
+    PASS=$((PASS+1))
+  else
+    echo "FAIL [$label] — expected to match /$pattern/, got: $actual" >&2
+    FAIL=$((FAIL+1)); FAILED="${FAILED}${label} "
+  fi
+}
+
+assert_ne() {
+  local label="$1" not_expected="$2" actual="$3"
+  if [ "$not_expected" != "$actual" ]; then
+    echo "PASS [$label]"
+    PASS=$((PASS+1))
+  else
+    echo "FAIL [$label] — expected NOT '$not_expected', got exactly that" >&2
+    FAIL=$((FAIL+1)); FAILED="${FAILED}${label} "
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# build_sandbox DIR
+#
+# Builds:
+#   $DIR/remote.git  — a bare "remote" repo
+#   $DIR/work        — a local working repo, `main` checked out, tracking
+#                       origin/main, with the REAL .githooks/pre-push and its
+#                       sourced libs installed via core.hooksPath.
+#
+# The remote's `main` is seeded via `git clone --bare` (not `git push`), so
+# no push — and therefore no hook invocation — happens during setup. Only
+# the pushes performed BY EACH TEST CASE go through the hook under test.
+#
+# bin/run-pre-push-checks.sh is stubbed to an immediate `exit 0` so cases
+# aren't slowed down by (or made to depend on) markdownlint/shellcheck/npx
+# being installed in the test environment — this suite tests the
+# protected-branch block, not the lint check set, which has its own tests.
+# ---------------------------------------------------------------------------
+build_sandbox() {
+  local dir="$1"
+  local work="$dir/work"
+
+  mkdir -p "$work"
+  git -C "$work" init -q -b main
+  git -C "$work" config user.name t
+  git -C "$work" config user.email t@t.com
+
+  mkdir -p "$work/.githooks"
+  cp "$REAL_PRE_PUSH" "$work/.githooks/pre-push"
+  chmod +x "$work/.githooks/pre-push"
+
+  mkdir -p "$work/.claude/hooks"
+  cp "$REAL_PROTECTED_LIB" "$work/.claude/hooks/_lib-protected-branches.sh"
+  cp "$REAL_READ_CONFIG_LIB" "$work/.claude/hooks/_lib-read-config.sh"
+
+  mkdir -p "$work/bin"
+  printf '#!/bin/bash\nexit 0\n' > "$work/bin/run-pre-push-checks.sh"
+  chmod +x "$work/bin/run-pre-push-checks.sh"
+
+  git -C "$work" config core.hooksPath .githooks
+
+  echo "seed" > "$work/README.md"
+  git -C "$work" add README.md
+  git -C "$work" commit -q -m "chore: init"
+
+  git clone -q --bare "$work" "$dir/remote.git"
+  git -C "$work" remote add origin "$dir/remote.git"
+  git -C "$work" fetch -q origin
+  git -C "$work" branch -q --set-upstream-to=origin/main main
+
+  # Neutralise the bare remote's OWN denyDeleteCurrentBranch safety net
+  # (git's server-side default refuses to delete whatever ref the bare
+  # repo's HEAD points at — here, `main`). Without this, case 7's
+  # "delete protected branch" assertions could pass for the WRONG reason —
+  # the remote's own unrelated protection, not our LOCAL pre-push hook —
+  # which would silently mask a broken hook. Disabling it here means the
+  # only thing standing between a delete-main push and the remote is the
+  # hook under test.
+  git -C "$dir/remote.git" config receive.denyDeleteCurrent ignore
+}
+
+# A new, real commit on the currently checked-out branch — used so every
+# "attempt to push main" case carries a genuine ref-update (local sha !=
+# remote sha), not an accidental no-op that git might not even invoke the
+# hook for.
+new_commit() {
+  local work="$1" msg="${2:-chore: advance}"
+  date +%s%N >> "$work/README.md"
+  git -C "$work" add README.md
+  git -C "$work" commit -q -m "$msg" --allow-empty
+}
+
+remote_ref_sha() {
+  local dir="$1" ref="$2"
+  git -C "$dir/remote.git" rev-parse --verify "$ref" 2>/dev/null || echo "MISSING"
+}
+
+# ---------------------------------------------------------------------------
+# CASE 1: feature-branch push -> NOT blocked
+# ---------------------------------------------------------------------------
+case_feature_branch_push_allowed() {
+  local sandbox; sandbox=$(mktemp -d)
+  build_sandbox "$sandbox"
+  local work="$sandbox/work"
+
+  git -C "$work" checkout -q -b feature/GH-1-safe
+  new_commit "$work"
+
+  local out rc
+  out=$(git -C "$work" push -u origin feature/GH-1-safe 2>&1); rc=$?
+  assert_eq "feature branch push: exit code" "0" "$rc"
+  assert_ne "feature branch push: remote ref created" "MISSING" "$(remote_ref_sha "$sandbox" refs/heads/feature/GH-1-safe)"
+
+  rm -rf "$sandbox"
+}
+
+# ---------------------------------------------------------------------------
+# CASE 2: genuine tag push -> NOT blocked
+# ---------------------------------------------------------------------------
+case_genuine_tag_push_allowed() {
+  local sandbox; sandbox=$(mktemp -d)
+  build_sandbox "$sandbox"
+  local work="$sandbox/work"
+
+  git -C "$work" tag v1.0.0-test
+
+  local out rc
+  out=$(git -C "$work" push origin v1.0.0-test 2>&1); rc=$?
+  assert_eq "genuine tag push: exit code" "0" "$rc"
+  assert_ne "genuine tag push: remote tag created" "MISSING" "$(remote_ref_sha "$sandbox" refs/tags/v1.0.0-test)"
+
+  rm -rf "$sandbox"
+}
+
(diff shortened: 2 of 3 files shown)


me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1090

**Commit**: `541e6b4660435c6970f00929887c7794b5068d5c`

### Summary

Step 2 of #1086. `.githooks/pre-push` now decides "block this push?" from git's own stdin ref lines rather than from shell command text, adds `_lib-protected-branches.sh` as the shared config resolver, and ships an 11-case / 30-assertion suite driven by a real git binary against a real bare remote. `block-main-push.sh` is genuinely untouched.

I verified the central property independently rather than reading the author's evidence: the branch is derived **only** from `$REMOTE_REF`, and the single `grep` in the file matches that branch against the static config regex. There is no `$COMMAND`, no `tool_input`, no command-string scan of any kind.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass**
- Security: **Pass**
- Performance: **Pass** (n/a)
- Technical Decisions (AgDR): **N/A** — implements AgDR-0104, cites AgDR-0113; the AgDR-0104 status update is scoped to step 3 by the issue
- Adopter Handbooks: **Pass** — `general/commit-message-quality` (advisory) satisfied; `architecture/migration-safety` is blocking but out of scope (no DB migration in this diff); `architecture/clean-architecture-layers` n/a

### Independent verification

**Loop exit semantics — confirmed correct, and it is load-bearing.** Isolated the two shapes:

```
procsub:  exit 1 inside loop -> script rc=1, "REACHED-END" never printed
pipe:     exit 1 inside loop -> script rc=0, "REACHED-END" printed
```

The shipped file uses `done < <(printf '%s\n' "$STDIN_REFS")`, so the loop body runs in the current shell and `exit 1` terminates the hook. Confirmed live against the shipped file by feeding it synthetic ref lines: a single `refs/heads/main` line gives rc=1 **and the runner never starts** — if the loop were a subshell the script would have fallen through to `exec bash "$RUNNER"` and returned the runner's 0. A two-line push where the protected ref is the *second* line also blocks, proving the loop scans all lines and still exits the script.

**`exit 1` vs `exit 2` — correct.** Git blocks on any non-zero from `pre-push`; there is no exit-2 convention at the git layer. The repo's `exit 2` idiom belongs to Claude Code `PreToolUse` hooks, a different contract. The existing fail-open paths in this same file already use `exit 0`, so `exit 1` for "blocked" is internally consistent.

**Discriminating before/after — verified myself, not taken on report.** Reconstructed the pre-change 35-line hook from the merge base (`33b4915`) into a parallel root and ran the *same* test file against it:

| | Before (`33b4915`) | After (this PR) |
|---|---|---|
| 20 "must block" assertions | **20 FAIL** | **20 PASS** |
| 9 "must NOT block" assertions | 9 PASS | 9 PASS |

Every #1066 / #1083 / #1084 / #1085 shape, plus bare push on `main`, explicit `push origin main`, and `push origin --delete main`, is ALLOWED before and BLOCKED after. That is genuine discrimination, not green-in-both-directions.

**Over-narrowing — clean, including cases the suite doesn't cover.** The four controls (tag push, feature-branch push, two-remote tag push, delete non-protected) pass in both directions. I added four of my own by feeding ref lines directly:

- `maintenance`, `devops`, `main-backup` → **allowed** (regex is anchored `^(...)$`; no substring over-match)
- `HEAD:refs/heads/main` and `HEAD:main` refspec renames → **blocked**. Worth calling out: this shape is *not* in the author's suite, and it is exactly the kind of thing a text parser has to special-case. Here it falls out for free because git resolves the refspec before the hook sees it. That is the design paying off.
- empty stdin → allowed, delegates normally

**Stdin hand-off — re-confirmed independently.** Built a sandbox with an instrumented runner that reports what it sees on fd 0 and exits 7:

- allowed push → `RUNNER-STARTED`, `RUNNER-STDIN-EOF-OR-EMPTY`, hook rc = 7. So the runner *is* reached, sees EOF (not ref lines, and not a hang), and `exec` still makes the runner's status the hook's status.
- blocked push → runner never starts.

The author's audit is correct: nothing in `bin/run-pre-push-checks.sh` or its three checks reads fd 0 (`find | xargs` reads the pipe, not stdin; `run_check` uses `bash -c` with no reads). And critically, the failure direction is benign — a downstream reader would get EOF, not block.

**The delete decision — right, and the test does not pass for the wrong reason.** Blocking a protected-branch delete is correct: it is strictly more destructive than a push, and it closes a real gap (`block-main-push.sh`'s `extract_push_ref` bails on `--delete` and falls back to validating the *checked-out* branch). On `receive.denyDeleteCurrent=ignore`: the before-run settles it empirically — `delete protected (main): remote main intact` **FAILED** against the old hook, i.e. `main` was actually deleted on the remote. If the remote's own safety net were still masking the hook, that assertion would have passed against the old hook and the case would have proved nothing. It didn't. The neutralisation works and the case is honest.

**Test suite:** 130/130, chunked at 20 per batch.

```
PASS=130  FAIL=0   (7 chunks: 20/20/20/20/20/20/10)
```

`test_block_main_push.sh` and `test_block_main_push_heredoc.sh` both PASS (chunk 1) — `block-main-push.sh` is genuinely untouched. New suite standalone: PASS=30 FAIL=0.

**shellcheck:** clean on all three files at *default* severity (not just `--severity=warning`) — zero output, rc=0.

### On the test approach (real git vs a stubbed `git`)

The deviation is right, and for a sharper reason than "it's more realistic". A stub would have to reimplement refspec resolution, upstream-tracking resolution, tag-vs-branch classification, and the all-zeroes sentinel in order to hand the hook correct stdin. The test would then be asserting that *the stub's* model of git agrees with the hook — while the actual claim under test is that *git's* model does. That is a second implementation that has to be trusted to agree, which is the precise failure mode this ticket exists to escape. Using real git removes the thing that could disagree.

The instruction it deviated from was written for `block-main-push.sh`, where a stub is genuinely the only way to prove a *simulated* command would have reached git. That premise doesn't hold here — this hook runs inside the real push lifecycle, so there is no simulated step to fool. The deviation is correctly scoped and correctly argued in the test header.

**Can any test pass for the wrong reason?** I looked for four mechanisms and found none:

1. *Remote-side rejection masking the hook* — ruled out empirically above (the before-run really does delete `main`).
2. *The stubbed runner producing the non-zero* — the stub is `exit 0`, and the block assertions match `BLOCKED.*protected branch 'main'`, which the runner's own failure text (`BLOCKED: pre-push check failed`) does not match.
3. *A no-op push git wouldn't even invoke the hook for* — `new_commit` forces a genuine ref advance before every "must block" case.
4. *Setup pushes silently exercising the hook* — the remote is seeded via `git clone --bare`, so no push occurs during setup; only the case's own push goes through the hook.

The one residual: because the sandbox stubs the runner to `exit 0`, **the stdin hand-off itself is not pinned by any automated test**. I verified it by hand and it is correct today, but a future change that made the runner read fd 0 would not be caught in CI. See suggestion 2.

### Issues Found

None blocking.

### The duplicate — my call: **defer to step 3, but pin it with a test**

Third option, and I'd hold it as a strong recommendation rather than a merge blocker.

**Why not fold the refactor in now.** `block-main-push.sh` is the currently-deployed control. Editing it inside a PR that exists to build a *new* layer beneath it enlarges the blast radius of the riskiest kind of change for the sake of hygiene, and it produces a state nobody designed: a file that sources the shared lib but is still the control. Step 3 is specifically about resolving what that file *is*; doing half of it early makes step 3's review harder, not easier. The framework's own rail — security and trust-chain changes go Heavy, ambiguity rounds up — argues for one trust-chain edit per PR, not two.

**Why not plain defer either.** "Two copies, agreement asserted in a comment" is the pattern that generated this backlog. The real-world divergence risk *here* is genuinely low — both copies are the same three lines reading the same `.git.protected_branches[]` key through the same `config_get`, with the same literal fallback, and I confirmed they resolve identically both in a clean tree and in a fork that has a real `project-config.json`. But "low risk" asserted in prose is exactly what the last several rounds of this backlog were also asserting.

**Why the pin is the right third answer.** A test that asserts `protected_branch_regex` equals `block-main-push.sh`'s inline resolution — under both config-present and config-absent conditions — converts the comment into an enforced invariant, **without touching the deployed control at all**. It costs a few lines, it makes divergence during the step-2→step-3 window impossible rather than merely unlikely, and it has a clean end of life: once step 3 migrates `block-main-push.sh` onto the lib, the test collapses to a tautology and gets deleted with the duplicate it was guarding. That is the cheapest available way to keep the deployed control untouched *and* not rely on trust.

I'm not blocking on it because the duplicate does not make anything worse than the pre-PR state — `block-main-push.sh`'s resolution is byte-identical to what shipped before — and because the enforcement improvement this PR delivers is real and shouldn't wait on a hygiene item.

**Does the PR document the duplicate honestly enough that step 3 can't miss it? In the code, yes. In the tracker, no.**

The PR body's third bullet names it plainly, and `_lib-protected-branches.sh`'s header (lines 31–39) goes further — it explicitly frames itself as "the thing step 3 migrates block-main-push.sh onto, not a competing implementation". That is the honest disclosure this repo's recent history has been short of, and it's a genuine improvement in reviewability.

The gap is that **#1086's step-3 scope doesn't mention it**. Step 3 reads: become advisory, keep a `--no-verify` check, close the parser bugs, update AgDR-0104. Nothing about migrating the protected-list resolution onto the shared lib. A step-3 implementer working from the issue could complete every bullet, demote the file to advisory, and leave the inline copy sitting there — because demoting to advisory doesn't require touching the resolution at all. The disclosure lives only in files that implementer has no particular reason to open.

**Ask:** add one bullet to #1086 step 3 — *"migrate `block-main-push.sh`'s inline protected-branch resolution onto `.claude/hooks/_lib-protected-branches.sh` and delete the duplicate"* — and, if you take the pin, note that the pinning test is deleted at the same time. That's a comment on the issue, not a change to this PR.

1. **Manual invocation now hangs on an interactive stdin.** `STDIN_REFS=$(cat)` blocks forever when fd 0 is an open TTY. Verified: `sleep 30 | bash .githooks/pre-push` times out; `bash .githooks/pre-push </dev/null` runs normally. Git always closes the hook's stdin, so **the enforcement path is unaffected**, and direct invocation isn't a documented path (`docs/getting-started.md` points at `core.hooksPath`, and the documented way to run the checks by hand is `bin/run-pre-push-checks.sh`). Still, it's a silent behaviour change for anyone who did run the hook directly. One line fixes it:

   ```bash
   if [ -t 0 ]; then STDIN_REFS=""; else STDIN_REFS=$(cat); fi
   ```

2. **Pin the stdin hand-off with a test.** The sandbox's `exit 0` runner stub means CI would not notice if the runner started reading fd 0. A case with a runner that reports whether it saw stdin and exits a distinctive status would pin both halves of the contract (runner reached with EOF; `exec` preserves its status) for a few lines.

3. **`echo "$BRANCH" | grep`** → `printf '%s\n' "$BRANCH" | grep` is marginally safer for a pathological ref name beginning `-e`/`-n` (creatable via `update-ref`, not via `git branch`). Cosmetic.

4. **`PROTECTED_REGEX` is interpolated unescaped into the regex**, so an adopter setting `.git.protected_branches: ["release/*"]` gets regex semantics rather than the glob they probably meant. Identical in `block-main-push.sh`, so not a regression — worth a line in the config docs at some point, possibly alongside step 3.

5. **Observation, not a finding:** in a managed-project clone with no `.claude/hooks/`, the lib is absent and the hook WARNs and fails open. Step 1's installer already exits 3 for a repo with no tracked hooks dir, so this is consistent and acknowledged — just worth not assuming #1086's "human terminal push in a managed project" row is closed when step 3 lands.

### Verdict

**APPROVED**

The design change is the right one and the evidence behind it is unusually strong: the discriminating before/after is real (20/20 must-block assertions flip), the over-narrowing controls hold including three shapes I added myself, and the one subtle way this could have silently failed — a pipe-fed loop swallowing `exit 1` — is both correctly implemented and verified live against the shipped file. The deliberate duplicate is disclosed honestly in the code; my one ask is that it also be disclosed in the tracker, plus the optional pinning test, neither of which needs to happen in this PR.

---
Reviewed commit: `541e6b4660435c6970f00929887c7794b5068d5c`

atlas-apex (commented):
## Security Review: PR #1090

**Commit**: `541e6b4660435c6970f00929887c7794b5068d5c`

### Summary

Step 2 of #1086. I verified by execution in a sandbox (real git binary, real bare remote), not by reading. Rex's core findings hold and I did not re-derive them; this review covers the ground he did not: degradation/fail-direction, config-driven regex injection assessed as a security property, `--no-verify` disclosure, and the control-comparison question step 3 depends on.

**The design premise is sound and I confirmed it independently.** Every refspec shape I threw at real git resolves to `refs/heads/main` in the hook's stdin — `push origin`, bare push, `--all origin`, `main:refs/heads/main`, `+main:main`. `main:HEAD` resolves to `refs/heads/HEAD` (a branch literally named HEAD — it does not touch `main`, so it is not a bypass). There is genuinely no text to misparse at this layer.

**Nothing here is a regression.** `block-main-push.sh` is untouched and remains the active control, so this PR strictly increases the union of protections. My findings are all either pre-existing parity or scoped to step 3.

### Checklist Results

- Secrets & Credentials: **Pass** (no secrets, no credential handling)
- Injection Prevention: **Pass with finding** — see MEDIUM-1 (config-to-regex interpolation)
- XSS Prevention: **N/A**
- Auth & Authorisation: **Pass** — this *is* an authorisation control; see the headline section
- Data Protection: **Pass**
- API Security: **N/A**
- Gate & Trust-Chain Integrity: **Pass** — this adds an enforcement point and weakens none. `block-main-push.sh` byte-identical, `.claude/settings.json` untouched, no gate made to exit 0 on a path it previously blocked.

### Headline: is this genuinely stronger than `block-main-push.sh`?

**Answer: it is stronger on one axis, strictly weaker on another. It is complementary, not a superset — and #1086's step 3 as currently written would be a net reduction in protection.**

What the `PreToolUse` hook sees that a `pre-push` hook structurally cannot:

| Shape | `block-main-push.sh` | `.githooks/pre-push` |
|---|---|---|
| **A commit on a protected branch** | **BLOCKS** (exit 2, lines 313–385) | **Never fires** — a commit never reaches a pre-push hook |
| Push from a clone where `core.hooksPath` was never installed | Blocks (resolved via the ops-fork walk in `settings.json`, independent of the pushed repo) | Never runs |
| Push with `--no-verify` | Sees the command text | Skipped entirely by git |
| Repo with `.githooks/` but no `.claude/hooks/` lib | Uses the ops fork's copy | WARNs and fails open |
| Compound call — blocking the whole Bash invocation before any of it runs | Blocks the whole call | Only the push fails; `;` follow-ons still run |

What the new hook sees that the old one cannot: every refspec shape (verified above), deletions, force pushes, and human terminal pushes.

**The sharpest item is the commit gate.** `.githooks/` contains **only** `pre-push` — there is no `pre-commit` hook. `block-main-push.sh` blocks commits on a protected branch today with exit 2. #1086's step 3 says that file "becomes an advisory … [and] `--no-verify` becomes the one text-matching job that remains" — which would leave protected-branch *commits* with no blocking control anywhere, and the git layer has nothing to replace it with.

I rate this **HIGH, scoped to step 3 — it does not block this PR**, because `block-main-push.sh` is untouched here and the commit gate is fully intact at this commit. But step 3 must not land until it explicitly retains (a) the commit gate as a *control*, not an advisory, (b) the `--no-verify` check, and (c) some answer for clones where `core.hooksPath` was never installed. Please record this on #1086 alongside the migration bullet already added.

### Security Issues Found

**MEDIUM-1 — a malformed protected-branch list fails OPEN, silently disabling protection for every branch in it.** Pre-existing parity, not a regression.

`PROTECTED_REGEX` is interpolated unescaped into `grep -qE "^(${PROTECTED_REGEX})$"`. Rex called this cosmetic; the fail *direction* is the part that matters. Measured, with config present, against both layers:

| `.git.protected_branches[]` | resolved regex | `.githooks/pre-push` | `block-main-push.sh` |
|---|---|---|---|
| `["main"]` | `main` | BLOCK | BLOCK |
| `["ma(in","main"]` | `ma(in\|main` | **ALLOW** | **ALLOW** |
| `["main","*bad"]` | `main\|*bad` | **ALLOW** | **ALLOW** |
| `["",""]` | `\|` | **ALLOW** | **ALLOW** |

A single malformed entry makes grep exit 2 (`grep: parentheses not balanced`), which the `if` reads as "not protected" — so protection for `main` disappears because of a typo in an *unrelated* entry. The error text does go to stderr, but the push proceeds. The `[ -n "$PROTECTED_REGEX" ]` guard does not catch the `["",""]` case either, since the regex is the literal `|`, and `^(|)$` matches only the empty string.

Related, and the likelier real-world hit: `["release/*"]` is regex, not glob — it protects `release` and `release/` but **not** `release/1.0`. An adopter writing the obvious thing gets silently *less* protection than they think.

Both layers behave identically, so this is not introduced here. It matters because step 3 makes this the sole blocking layer. Cheapest fix, no regex engine involved: match with a bash `case` over the config values, or validate the assembled regex once and fail **closed** with a loud message if grep exits ≥2.

**MEDIUM-2 — the enforcement block fails open *silently* when the lib is present but unusable.**

Degradation matrix, measured by feeding the shipped hook synthetic ref lines:

| Condition | Result |
|---|---|
| normal ref line | BLOCK |
| protected ref on the 2nd line | BLOCK (loop scans all lines) |
| delete sentinel | BLOCK |
| extra whitespace between fields | BLOCK |
| `_lib-read-config.sh` missing | **BLOCK** (falls back to defaults) — fails closed, correct |
| empty stdin / blank line / truncated line | ALLOW, silent |
| lib file **missing** | ALLOW + **WARN** (documented, intentional) |
| lib present but **syntactically broken** | **ALLOW, no warning** |
| lib present but **function not defined** | **ALLOW, no warning** |

The `command -v protected_branch_regex` guard converts "the lib exists but is unusable" into a silent no-op. The missing-*file* case warns; the unusable-*file* case does not. That asymmetry is worth closing precisely because step 3 will migrate `block-main-push.sh` onto this lib — a rename of `protected_branch_regex` during that migration would remove enforcement with **zero signal**, and every test that supplies a well-formed lib would still pass. One `else` branch emitting the same WARN fixes it.

The empty-stdin fail-open is fine and I'd leave it: I confirmed git does not invoke the hook at all when there is nothing to push ("Everything up-to-date" produces no hook invocation), so it is unreachable on a real push.

**MEDIUM-3 — `--no-verify` is not named anywhere, while the PR claims to close the whole bug class.**

Zero mentions of `no-verify` in `.githooks/pre-push`; no gate for it anywhere in `.claude/hooks/` or `bin/` (it appears only as an argument to skip over in `_lib-extract-push-ref.sh`'s option table). The PR body's "What this PR does NOT do" is honest about the `core.hooksPath` dependency — genuinely good disclosure — but omits the flag that bypasses this layer entirely, while the first bullet says this "closes the whole bug class AgDR-0113 documents". AgDR-0104's standing requirement is to name what is a control and what is a backstop. #1086 states the residual correctly in its Risks section; this PR should carry the same sentence, in the hook header and in the PR body. Documentation-only, but on a trust-chain enforcement point that a later step demotes another control on the strength of.

**LOW-1 — I agree with Rex that the `$(cat)` TTY hang is non-blocking.** I looked for a reachable path and found none: no CI workflow invokes the hook, `bin/install-git-hooks.sh` and `check-git-hooks-installed.sh` never execute it, and git always closes the hook's stdin. I did confirm the hang is real and specific to an open fd 0 (`sleep 8 | pre-push` times out; `< /dev/null` exits cleanly). The one-line `[ -t 0 ]` guard Rex proposed is still worth taking — it costs nothing and removes a foot-gun for anyone running the hook by hand.

**LOW-2 — the config source is not necessarily the repo being pushed.** `protected_branch_regex` resolves its root through `resolve_ops_root` (session-pin first, otherwise a walk for `.apexyard-fork`), so the protected list can come from the **ops fork's** config rather than the config of the repo whose push is being judged. This is the intended #310 behaviour and `block-main-push.sh` does the same, so it is parity — but the hook's own help text points the reader at `.claude/project-config.json` without noting which one wins. Worth a clarifying line.

### The duplication — Rex's "byte-identical" claim verified, and the exposure is bounded in the safe direction

I tested the two resolutions side by side across six config conditions including the malformed and empty ones, and they agree in **every** case. More importantly for the security question: while **both** layers are active, protection is the **union** — either can block. So a divergence during the step-2→step-3 window makes protection *stricter*, not looser. The exposure is bounded and fails safe. It only becomes a weakening at step 3, when `block-main-push.sh` stops being a control — which is exactly when the migration bullet already added to #1086 removes the duplicate. I have nothing to add to Rex's call; deferring is right.

### Independent confirmation of the delete finding — it is real

Confirmed directly, not taken from the author or from Rex. With a feature branch checked out, `block-main-push.sh` returns **rc=0 with no output** for a delete-push of `main` — allowed. The same shape at the git layer (delete sentinel on `refs/heads/main`) returns rc=1 — blocked. The gap is specifically "delete a protected branch from a non-protected checkout"; from a `main` checkout the fallback happens to catch it for the wrong reason. This is a live gap in the currently-deployed control and should be recorded on #1086 regardless of this PR's fate.

### Verification performed

- Full hook suite re-run in 6 chunks of ≤20: `PASS=115 FAIL=0` (115 test files, 0 failures — my count is per-file; the runner's own tally of 130 counts differently, but there are no failures either way)
- New suite standalone: `PASS=30 FAIL=0`
- `shellcheck` at **default** severity on all three files: clean, rc=0
- Sandbox probes: refspec resolution, config injection, degradation matrix, delete gap, hang reachability

One live corroboration of the PR's own premise, from this review session: while writing up these findings, `validate-branch-name.sh` false-blocked on prose inside a heredoc, extracting a branch named `git`. That is the same text-matching unsoundness class this PR exists to escape, firing unprompted during its review.

### Recommendations

1. Record the **commit-gate** requirement on #1086 step 3 before it is scoped (HIGH, gates step 3, not this PR).
2. Add the `--no-verify` residual to the hook header and PR body (MEDIUM-3).
3. Emit the WARN on the unusable-lib paths, not only the missing-file path (MEDIUM-2).
4. Validate the assembled protected-branch regex and fail **closed** on a grep error (MEDIUM-1) — ideally before step 3 makes this the only layer.
5. Take Rex's `[ -t 0 ]` one-liner (LOW-1).

None of these blocks this PR.

### Verdict

**APPROVED**

The design is right, the premise is verified rather than asserted, and the enforcement improvement is real. Every finding above is either pre-existing parity that this PR inherits without worsening, or a scoping requirement for step 3 — and step 3 is where the genuine risk sits: this layer is **not** a superset of `block-main-push.sh`, and demoting that hook without preserving the commit gate, the `--no-verify` check, and a story for uninstalled clones would be a net reduction in protection.

---
Reviewed commit: `541e6b4660435c6970f00929887c7794b5068d5c`

Issue #1091, opened 2026-07-31
[Bug] Merge gate falls back to LOCAL HEAD when the forge call fails — the one sound boundary degrades open


atlas-apex:
## Bug Scenario

**Given** the merge gate's entire integrity model rests on comparing review-marker SHAs against the **forge-reported** PR HEAD — state an agent cannot author

**When** the `gh pr view` call that resolves that HEAD fails for any reason (network blip, expired token, rate limit, `gh` not on PATH)

**Then** `block-unreviewed-merge.sh:233-234` prints a warning and **falls back to `git rev-parse HEAD`** — a local value the agent fully controls — and the gate proceeds to compare markers against it

**Expected** a gate that cannot evaluate its precondition **blocks**. This is the principle the framework already adopted for [#965](https://github.com/me2resh/apexyard/issues/965) in [AgDR-0104](../docs/agdr/AgDR-0104-trust-chain-controls-vs-backstops.md): *"a gate that can't evaluate its precondition must block, not allow."* That fix was applied to the `jq`-unavailable path; this sibling path was left failing open.

## The code

```
.claude/hooks/block-unreviewed-merge.sh:233
  echo "WARN: Could not resolve PR #${PR_NUMBER} HEAD via gh — falling back to
        local HEAD. If this merge fails, run 'gh pr checkout ${PR_NUMBER}'
        first or re-authenticate gh." >&2
.claude/hooks/block-unreviewed-merge.sh:234
  CURRENT_SHA=$(git rev-parse HEAD 2>/dev/null)
```

## Why this is the most severe finding in the control plane

Per the threat model run on this repo (`projects/apexyard/audits/threat-model/`, threat **T2**), the framework has exactly **one** adversarially-sound boundary — **B2, local ↔ forge**. Every other control lives on the local filesystem and is writable by the actor it governs; that is a property of *where* the control sits, not of how carefully it is written.

B2 is sound precisely because the forge-reported HEAD is state a local file write cannot fabricate. This fallback **removes that property on degradation**. The one boundary that was supposed to hold becomes a local read, silently, with only a warning on stderr that nothing acts on.

It is also the cheapest fix on the list — the failure direction is wrong, not the logic.

## Repro

```bash
# Any PR with valid markers at HEAD.
GH_TOKEN=invalid gh pr merge <N> --repo <owner/repo> --squash
# Gate warns, resolves CURRENT_SHA from the local working tree, and
# compares markers against that instead of the forge's HEAD.
```

[...]

PR #1098, opened 2026-07-31
fix(#1091): block instead of falling back to local HEAD when the forge is unreachable

atlas-apex:
## Summary

- **Three merge gates fell back to the local HEAD when the forge call failed, which voids the check they exist to perform.** They compare a recorded approval SHA against the PR's HEAD *as the forge reports it* — the one value in this framework an agent cannot author. On any expired token, network failure, or rate limit, `block-unreviewed-merge.sh`, `require-design-review-for-ui.sh`, and `require-architecture-review.sh` printed a warning and substituted `git rev-parse HEAD`, a local value the agent fully controls.
- **All three now block.** Each carried the fallback byte-for-byte identically. The new message names the likely causes and states that no approval needs re-recording, since the markers remain valid.
- **Neither gate newly blocks anything it wasn't already governing**, though the two cases differ and the distinction matters:
  - The **two review gates** reach this code only *after* an `exit 0` on out-of-scope and an `exit 2` on missing marker. A *total* forge outage never reaches the new block at all — `gh pr diff` returns empty and the hook exits 0 at the scope check. The block is reachable only on a partial failure where the diff answers and `headRefOid` does not.
  - **`block-unreviewed-merge.sh` resolves the SHA unconditionally**, before the marker check, so a forge outage does block every merge. That is correct rather than incidental: this gate already governs every merge, and a merge whose HEAD cannot be established is precisely what it should refuse.
- **This applies the principle #965 already established** for the `jq`-unavailable path ([AgDR-0104](https://github.com/me2resh/apexyard/blob/dev/docs/agdr/AgDR-0104-trust-chain-controls-vs-backstops.md)): a gate that cannot evaluate its precondition must block, not guess. That fix landed on one path; this is its sibling.

## Why the safe outcome was previously accidental

In practice the local working tree is rarely the PR branch, so the fallback usually produced a mismatch and blocked anyway. That is luck, not design — the outcome depended on whatever the tree happened to point at. Check out the PR branch first and the comparison passes against agent-controlled state.

## Testing

The new test is built to expose exactly that, and is discriminating by construction: it makes the local HEAD and the markers **agree** — the setup most favourable to a bypass — so the old code allows the merge and the new code blocks it. A test using a mismatching local HEAD would have blocked both before and after and proved nothing.

Measured, same test file, same fixture:

```
before: FAIL rc=0  "falling back to local HEAD"   <- merge ALLOWED
after:  PASS rc=2                                  <- blocked
control (forge healthy): rc=0 both before and after
```

1. `bash .claude/hooks/tests/test_block_unreviewed_merge.sh` → **41/41**, including the two new cases.
2. Full suite → `PASS=127 FAIL=3`. The three are `test_suggest_mcp_search`, `test_token_efficiency_wave1`, `test_harnessability_scoring` — **all three pass on clean `upstream/dev`**, verified in a worktree. They fail only in a working tree carrying untracked local skills, because the skill-count assertion counts tracked skills. Not framework failures, and unrelated to this change.
3. `shellcheck --severity=warning` clean on all four files; `bash -n` clean on all three hooks.

## Context

Filed from the threat model run against the framework's own control plane (finding **T2**), which identified the local ↔ forge boundary as the only adversarially sound control in the system — and this fallback as the thing that silently removes it.

Refs #1091

---

## Glossary

| Term | Definition |
|------|------------|
| fail open / fail closed | A degraded control that permits (open) versus one that blocks (closed). For a gate, closed is the safe direction — it over-blocks rather than letting something through unchecked. |
| forge-reported HEAD | The PR's head commit as the server reports it. Load-bearing because a local file write cannot change it, unlike anything on disk. |
| discriminating test | One that fails against the pre-change code and passes after. A test green in both directions proves nothing. |
| approval marker | A file under `.claude/session/reviews/` recording a review at a specific SHA. Only meaningful if compared against a SHA the author cannot choose. |

### DIFF
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -226,12 +226,34 @@ CEO_APPROVAL=$(review_marker_path "${CMD_REPO:-unknown}" "$PR_NUMBER" ceo "$MARK
 # branch. Asking gh directly removes the need for `gh pr checkout <N>`
 # before every `gh pr merge <N>`.
 #
-# Fallback to local HEAD if the gh call fails, with a visible warning, so
-# a transient network / auth issue doesn't brick merges entirely.
+# If the forge call fails we BLOCK (#1091). Falling back to the local HEAD
+# would compare markers against an agent-controlled value, which is exactly
+# the property this gate exists to deny.
 CURRENT_SHA=$(resolve_pr_head "$PR_NUMBER" "$CMD_REPO")
 if [ -z "$CURRENT_SHA" ]; then
-  echo "WARN: Could not resolve PR #${PR_NUMBER} HEAD via gh — falling back to local HEAD. If this merge fails, run 'gh pr checkout ${PR_NUMBER}' first or re-authenticate gh." >&2
-  CURRENT_SHA=$(git rev-parse HEAD 2>/dev/null)
+  cat >&2 <<MSG
+BLOCKED: could not resolve PR #${PR_NUMBER}'s HEAD from the forge.
+
+This gate compares the recorded approval SHA against the PR's HEAD **as the
+forge reports it** — state that a local file write cannot fabricate. That
+comparison IS the property the gate exists to provide.
+
+Until me2resh/apexyard#1091 this fell back to the LOCAL HEAD
+(\`git rev-parse HEAD\`) with only a warning. That substituted an
+agent-controlled value for the one value in this system an agent cannot
+author, so on any forge hiccup the gate silently stopped meaning anything.
+A gate that cannot evaluate its precondition must BLOCK, not guess — the same
+principle already applied to the jq-unavailable path in #965 (AgDR-0104).
+
+Likely causes: expired or absent forge token, network failure, API rate
+limit, or the forge CLI not installed.
+
+To unblock:
+  1. Check auth — \`gh auth status\` (or \`glab auth status\`), re-login if needed
+  2. Confirm connectivity to the forge
+  3. Retry the merge — no approval needs re-recording; the markers are still valid
+MSG
+  exit 2
 fi
 
 # --- Rex marker check ---
--- a/.claude/hooks/require-architecture-review.sh
+++ b/.claude/hooks/require-architecture-review.sh
@@ -275,14 +275,36 @@ MSG
   exit 2
 fi
 
-# SHA consistency check — resolve the PR's real HEAD via GitHub rather than
-# local HEAD (see #55). Falls back to local HEAD with a warning if the
-# gh call fails (network, auth).
+# SHA consistency check — resolve the PR's real HEAD via the forge rather than
+# local HEAD (see #55). If that resolution fails we BLOCK rather than fall back
+# to the local HEAD (#1091) — a local value is agent-controlled, so falling
+# back would silently void the check.
 APPROVED_SHA=$(tr -d '[:space:]' < "$APPROVAL")
 CURRENT_SHA=$(resolve_pr_head "$PR_NUMBER" "$CMD_REPO")
 if [ -z "$CURRENT_SHA" ]; then
-  echo "WARN: Could not resolve PR #${PR_NUMBER} HEAD via gh — falling back to local HEAD. If this merge fails, run 'gh pr checkout ${PR_NUMBER}' first or re-authenticate gh." >&2
-  CURRENT_SHA=$(git rev-parse HEAD 2>/dev/null)
+  cat >&2 <<MSG
+BLOCKED: could not resolve PR #${PR_NUMBER}'s HEAD from the forge.
+
+This gate compares the recorded approval SHA against the PR's HEAD **as the
+forge reports it** — state that a local file write cannot fabricate. That
+comparison IS the property the gate exists to provide.
+
+Until me2resh/apexyard#1091 this fell back to the LOCAL HEAD
+(\`git rev-parse HEAD\`) with only a warning. That substituted an
+agent-controlled value for the one value in this system an agent cannot
+author, so on any forge hiccup the gate silently stopped meaning anything.
+A gate that cannot evaluate its precondition must BLOCK, not guess — the same
+principle already applied to the jq-unavailable path in #965 (AgDR-0104).
+
+Likely causes: expired or absent forge token, network failure, API rate
+limit, or the forge CLI not installed.
+
+To unblock:
+  1. Check auth — \`gh auth status\` (or \`glab auth status\`), re-login if needed
+  2. Confirm connectivity to the forge
+  3. Retry the merge — no approval needs re-recording; the markers are still valid
+MSG
+  exit 2
 fi
 if [ -n "$APPROVED_SHA" ] && [ -n "$CURRENT_SHA" ] && [ "$APPROVED_SHA" != "$CURRENT_SHA" ]; then
   cat >&2 <<MSG
--- a/.claude/hooks/require-design-review-for-ui.sh
+++ b/.claude/hooks/require-design-review-for-ui.sh
@@ -298,14 +298,36 @@ MSG
   exit 2
 fi
 
-# SHA consistency check — resolve the PR's real HEAD via GitHub rather than
-# local HEAD (see #55). Falls back to local HEAD with a warning if the
-# gh call fails (network, auth).
+# SHA consistency check — resolve the PR's real HEAD via the forge rather than
+# local HEAD (see #55). If that resolution fails we BLOCK rather than fall back
+# to the local HEAD (#1091) — a local value is agent-controlled, so falling
+# back would silently void the check.
 APPROVED_SHA=$(tr -d '[:space:]' < "$APPROVAL")
 CURRENT_SHA=$(resolve_pr_head "$PR_NUMBER" "$CMD_REPO")
 if [ -z "$CURRENT_SHA" ]; then
-  echo "WARN: Could not resolve PR #${PR_NUMBER} HEAD via gh — falling back to local HEAD. If this merge fails, run 'gh pr checkout ${PR_NUMBER}' first or re-authenticate gh." >&2
-  CURRENT_SHA=$(git rev-parse HEAD 2>/dev/null)
+  cat >&2 <<MSG
+BLOCKED: could not resolve PR #${PR_NUMBER}'s HEAD from the forge.
+
+This gate compares the recorded approval SHA against the PR's HEAD **as the
+forge reports it** — state that a local file write cannot fabricate. That
+comparison IS the property the gate exists to provide.
+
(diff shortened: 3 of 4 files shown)


atlas-apex (commented):
## Code Review: PR #1098

**Commit**: `6b8d733dbbb9f2c1f620bf300d3de995edae9ab3`

### Summary

Three merge gates compared a recorded approval SHA against the forge-reported PR HEAD, but silently substituted `git rev-parse HEAD` — an agent-controlled value — whenever the forge call failed. All three now block instead. The change is small, uniform, and correct.

I verified the two claims that carry the change rather than reading them: the fail-closed direction cannot over-block, and the new test genuinely discriminates.

### Checklist Results
- Testing:                  Pass (one suggestion below)
- Security:                 Pass
- Performance:              Pass (N/A)
- Technical Decisions (AgDR): N/A — applies the principle already recorded in AgDR-0104, cited in the body
- Adopter Handbooks:        N/A — no findings (shell-only diff)
- Fallow Static Analysis:   N/A — not JS/TS

### Verification performed

**1. Failing closed cannot over-block — confirmed, and it is narrower than the body claims.**

Traced the control flow in all three files:

| Gate | Reaches the SHA check… |
|---|---|
| `require-design-review-for-ui.sh` | after `exit 0` on not-a-UI-PR (line 245) and after `exit 2` on missing marker (298) |
| `require-architecture-review.sh` | after `exit 0` on no-design-artifact (232) and after `exit 2` on missing marker (275) |
| `block-unreviewed-merge.sh` | unconditionally for any merge command, **before** the Rex-marker existence check |

The third is the one worth stating plainly, because the body's "blast radius" bullet is scoped to "both review gates" and leaves it out. In `block-unreviewed-merge.sh` a forge outage does block every merge in the repo — but that gate already governs every merge unconditionally, so nothing is newly brought into scope. A merge whose HEAD cannot be established is exactly what it should refuse. Safe, just for a different reason than the other two.

Additionally, in both review gates a *total* forge outage never reaches the new code at all: `gh pr diff` returns empty and the hook exits 0 at the scope check (`require-design-review-for-ui.sh:207-211`). The new block there is reachable only on a *partial* failure where the diff answers and `headRefOid` does not — exactly what the test constructs. That is narrower than the body implies, and it cuts in favour of the change.

**2. The test genuinely discriminates — measured, not read.**

Ran the new test file against the pre-change hook in a worktree at `6b8d733^`:

```
before (59c3cc6): FAIL rc=0  "falling back to local HEAD"   <- merge ALLOWED
after  (6b8d733): PASS rc=2                                 <- BLOCKED
control (gh healthy): rc=0 both before and after
```

Matches the claim exactly. The construction is the right one — markers written to match the *local* HEAD is the setup most favourable to a bypass, so a mismatching-HEAD test would have blocked in both directions and proved nothing.

**3. The `gh` stub isolation is real, and the intended path is the one exercised.**

The pre-change run is the proof: it emitted `falling back to local HEAD`, a string that prints only inside the `[ -z "$CURRENT_SHA" ]` branch immediately after `resolve_pr_head`. No earlier bail-out can produce it. The stub answering `headRefName` also matters — it lets the sync-PR squash guard (which calls `resolve_pr_head_branch`, a separate API call) evaluate and fall through rather than short-circuit. Exactly one thing fails.

**4. `block-merge-on-red-ci.sh` correctly left alone — confirmed.** It contains no local-HEAD fallback; the only match is a comment referencing `resolve_pr_head`. The two remaining `falling back to local HEAD` sites in the tree are `validate-branch-name.sh` (different concern) and `warn-stale-review-markers.sh` — a PostToolUse advisory that never exits non-zero, where a degraded read can only produce a spurious or missing *warning*, never a bypass. Both correctly out of scope.

**5. Re-ran `test_block_unreviewed_merge.sh` → 41/41.** `bash -n` and `shellcheck --severity=warning` clean on all four files. I did not treat the three known skill-count failures as regressions.

### Issues Found

None blocking.

### Suggestions

**suggestion: pin the fix per-gate.** The two review gates got the identical change with no test, and the usual justification — no harness — does not apply here: `test_require_architecture_review.sh` and `test_require_design_review_for_ui.sh` both already ship a `make_sandbox()` with a writable `bin/gh` stub (architecture test lines 90-132). An equivalent case in each is close to a copy-paste. Not blocking, because the replacement really is byte-identical across all three and the mechanism is now pinned once. But "identical code, tested once" decays the moment one of the three diverges, and these are trust-chain files where that decay is expensive.

**nit: the blast-radius bullet is accurate as written but reads wider than it is.** "Both review gates reach this code only after establishing the PR is in scope and an approval marker exists" is true, and true only of two of the three files changed. A reader scanning the bullet under a heading that says "the blast radius is narrow by construction" will carry it to all three. The third gate's safety argument is different — every merge is in scope, and that is fine because the gate already governed every merge. One clause naming that would close the gap. This is the same class of thing flagged on #1096: the characterisation is doing slightly more work than the code supports, and the shortest accurate version is stronger.

**nit: minor diagnostic ordering in `block-unreviewed-merge.sh`.** Because the HEAD resolution now exits before the Rex-marker existence check, a PR with no Rex approval at all reports "could not resolve HEAD" during a forge outage rather than the more actionable "no Rex approval". Both block, so there is no correctness impact — but the less useful message wins. Only worth moving if you touch the file again.

**observation, not for this PR:** the scope check in the two review gates (`gh pr diff` empty → `exit 0`, "skip rather than false-positive") is itself a forge-dependent fail-*open*, and arguably the same shape as the T2 finding this PR closes. It is a genuine false-positive tradeoff and out of scope here, but it is the reason the new block is partial-outage-only in those files, and it may deserve its own look under #1091's threat-model thread.

### Verdict
**APPROVED**

Both things I was asked to check hold up under measurement rather than reading. The fail-closed direction is safe — no merge is brought into scope that the gate was not already governing — and the test fails against the pre-change hook and passes after, with a control that stays green in both directions. The suggestions are follow-ups, not conditions.

---
Reviewed commit: `6b8d733dbbb9f2c1f620bf300d3de995edae9ab3`

atlas-apex (commented):
## Security Review: PR #1098

**Commit**: `6b8d733dbbb9f2c1f620bf300d3de995edae9ab3`

### Summary

Three merge gates stopped substituting an agent-controlled value (`git rev-parse HEAD`) for the forge-reported PR HEAD when the forge call fails. Per the threat model's T2, the local↔forge boundary is the only adversarially sound control in this system — every marker and hook is writable by the actor it governs — so the prior fallback silently voided the gate on any forge hiccup. This restores it.

I verified by execution rather than reading, and specifically went after the coverage asymmetry Rex flagged rather than repeating his findings.

### Verification performed

- `test_block_unreviewed_merge.sh` → **41/41 PASS** at the review SHA.
- **Built the missing discriminating tests for the two untested gates myself** (partial-outage `gh` stub: `pr diff` answers so the scope check passes, `headRefOid` fails; approval marker written to match the *local* HEAD, i.e. the setup most favourable to a bypass), and ran each against both the parent commit and this one:

| Gate | Parent (before) | This PR (after) |
|---|---|---|
| `require-architecture-review.sh` | `rc=0` — **merge ALLOWED** on a marker matching only the local HEAD | `rc=2` BLOCKED |
| `require-design-review-for-ui.sh` | `rc=0` — **merge ALLOWED** on a marker matching only the local HEAD | `rc=2` BLOCKED |

So the bypass was real and reachable in both untested gates, and is closed in both. This is behavioural evidence, not an inference from diff similarity.

### Checklist Results

- Secrets & Credentials: Pass
- Injection Prevention: Pass
- XSS Prevention: N/A
- Auth & Authorisation: Pass — this *strengthens* the authorisation gate
- Data Protection: Pass
- API Security: Pass
- Gate & Trust-Chain Integrity (#777): **Pass** — fail-open path removed, no gate loosened

### Findings

**LOW — residual wrong-non-empty path (`#887` ambient-repo class). Not introduced here.**
The new guard triggers only on `resolve_pr_head` returning *empty*. A **wrong non-empty** SHA would pass it silently with no backstop. The one live path is the repo-less branch (`gh pr view <N>` with no `--repo`, used when the merge command carries no `--repo`), which resolves against the ambient cwd repo — gh's ambient default prefers the parent/upstream on a fork. This is mitigated but not eliminated: the hook and the real merge command resolve through the *same* ambient rule, so they agree on which repo `#N` means, and a divergence would have to come from gh itself. `gh` keeps no local SHA cache, so a stale-read path does not exist client-side. Pre-existing, out of scope for this PR — but it is the honest answer to "restored or narrowed": see the verdict section.

**LOW — no format validation on `CURRENT_SHA`.**
A non-empty, non-SHA value would be compared verbatim against the marker and mismatch → block. Fails in the safe direction; noting only for completeness.

**LOW (was Rex's action item) — coverage asymmetry on trust-chain files.**
My position from a security standpoint: untested-but-identical is **not acceptable as a standing state** on `.claude/hooks/**`, because the next refactor has nothing pinning the behaviour. But it is **not blocking here**, because I have now executed the equivalent case against both gates and both discriminate correctly. The risk is regression, not current correctness. Recommend landing the two copy-paste cases into `test_require_architecture_review.sh` / the design-gate suite as a follow-up.

### Answers to the specific questions

1. **Restored, not merely narrowed — for the failure mode in scope.** Empty-return is the only way `resolve_pr_head` signals failure, and every failure mode (network, expired token, rate limit, CLI absent, PR not found, glab + missing jq) collapses to empty. The wrong-non-empty concern above is a distinct, pre-existing class that this PR neither creates nor claims to fix.
2. **Fail-closed is not exploitably worse than the alternative.** An agent *can* induce the failure (e.g. dropping forge auth), and `block-unreviewed-merge.sh` resolves unconditionally, so that blocks every merge in the repo. That is the correct direction — the induced outcome is "no merge", never "unreviewed merge". The operator is not stalled without diagnosis: the message names the precondition, the four likely causes, the exact commands to check, and states that recorded approvals remain valid. That last line matters — it prevents an operator from re-running approvals in confusion.
3. **No new command-text decision.** Confirmed: the added code branches solely on `[ -z "$CURRENT_SHA" ]`. Nothing new pattern-matches command text, consistent with AgDR-0104.
4. See the LOW finding above.
5. **Nothing made worse.** The old fallback blocked only when the local HEAD happened to mismatch the marker; the new code blocks that entire path unconditionally. The new blocking set is a strict superset of the old one, so no case the fallback used to catch is now let through.

### Note (informational, not a finding)

While building the test harness, `block-unreviewed-merge.sh` blocked my own `Bash` call because the *script text* contained a merge-command string — the gate matching tool-call text rather than intent. Pre-existing AgDR-0104 territory and fail-closed noise rather than a security defect, but it is a live example of the text-matching unsoundness that quarter's backlog is about.

### Verdict

**APPROVED**

The change removes a genuine, execution-confirmed bypass in all three gates and introduces no new fail-open path.

---
📌 Reviewed commit: `6b8d733dbbb9f2c1f620bf300d3de995edae9ab3`

me2resh (approved):

Issue #1094, opened 2026-07-31
[Bug] /decide is human-only, but agdr-decisions.md orders the agent to run it — the rule instructs an action the framework forbids


atlas-apex:
## Bug Scenario

**Given** `.claude/rules/agdr-decisions.md` opens with a direct instruction to the agent — *"**HARD STOP**: before making a material technical decision … run `/decide` and create an Agent Decision Record"* — and repeats `→ /decide` four more times in its self-check list

**When** the agent tries to obey it

**Then** the invocation fails: `Skill decide cannot be used with Skill tool due to disable-model-invocation`

**Expected** either the agent can run `/decide`, or the rule stops telling it to. Today the framework instructs an action it has mechanically forbidden.

## Which skills are locked, and which are justified

Five skills carry `disable-model-invocation: true`:

| Skill | Justified by AgDR-0110? | Pinned by the test? |
|---|---|---|
| `/approve-merge` | **yes** — the invocation *is* the approval; it writes the CEO marker and merges | yes |
| `/approve-design` | **yes** — writes the design-review marker a merge gate reads | yes |
| `/approve-architecture` | **yes** — writes the architecture-review marker | yes |
| **`/decide`** | **no** | **no** |
| **`/audit-deps`** | **no** | **no** |

[AgDR-0110](../docs/agdr/AgDR-0110-approval-skills-human-only.md) / [#1042](https://github.com/me2resh/apexyard/issues/1042) is about **approval**: a human must make the discrete per-PR decision, and unlocking the review skills without locking the approval skills would have opened a fully autonomous `open → review → approve → merge` path. That reasoning is sound and should not change.

It does not extend to `/decide`. An AgDR is a decision **record**, not an approval gate. It writes no marker, satisfies no merge gate, and authorises nothing. Nor to `/audit-deps`, which reads dependency manifests and has no approval semantics whatsoever.

## Why this is a real cost, not a nitpick

The AgDR rule is one of the framework's central disciplines — `CLAUDE.md` lists it under Code Standards, and `workflow-gates.md` Gate 2 requires an AgDR before the Build phase for material decisions. Making it unreachable by the agent means:

1. **The agent cannot follow its own most-cited rule.** Every `→ /decide` in `agdr-decisions.md` is unactionable.
2. **It pushes decisions toward not being recorded.** Faced with a rule it cannot execute, the likely outcome is an undocumented decision — precisely the failure the AgDR discipline exists to prevent.
[...]

PR #1101, opened 2026-07-31
fix(#1094): unlock /decide and /audit-deps, and assert over every locked skill

atlas-apex:
## Summary

- **A rule ordered an action the framework had forbidden.** `.claude/rules/agdr-decisions.md` opens with *"**HARD STOP**: … run `/decide`"* and repeats `→ /decide` four more times, but `/decide` carried `disable-model-invocation: true`, so the agent could not invoke it. Every one of those instructions was unactionable.
- **AgDR-0110's lock is sound and is unchanged.** The three *approval* skills stay human-only, because for those the invocation **is** the approval — without that lock, unlocking the review skills would open an autonomous `open → review → approve → merge` path. That reasoning simply does not reach `/decide`: an AgDR is a decision **record**, writing no marker, satisfying no merge gate, authorising nothing. Nor `/audit-deps`, which reads manifests and reports. No AgDR or rule names either as needing a lock.
- **Both are set to `false` rather than having the key removed** — the existing test treats an absent key as a failure for the families it covers, since absent defaults to model-invocable, which is how `/approve-architecture` was once silently unguarded. Stating the posture is the convention here.
- **The durable half is the test.** `test_skill_invocability_gates.sh` pinned three review skills unlocked and three approval skills locked — it structurally could not catch a *seventh* skill acquiring the flag, which is exactly how these two went unnoticed. It now iterates **every** `SKILL.md` and requires any locked one to appear in an explicit `LOCK_ALLOWED` list. Adding a lock means editing that list and stating why.

## Why the contradiction is the bug, not the lock

The framework already knows this failure mode. The same test's justification for keeping review skills unlocked reads: *"…contradicting `auto-code-review.sh`, which explicitly instructs the orchestrator to run `/code-review`."* This is that defect one skill over, and undetected because nothing asserted over the whole set.

The new failure message tells whoever adds a future lock to check that no rule instructs the agent to invoke it — the condition that turns a preference into a contradiction.

If `/decide`'s lock turns out to have been deliberate, the fix is the mirror image: change `agdr-decisions.md` so it tells the agent to *prepare* a decision and hand it to the operator. Either resolution is fine. The contradiction is not.

## Correction from the first review round

The first version of the every-skill loop was **fail-open in exactly the direction the assertion exists to prevent**. `invocation_flag` returned the raw frontmatter value, so a lock spelled `"true"`, `TRUE`, `True`, `true # comment`, or with trailing whitespace compared unequal to `true` and was **silently skipped as unlocked**.

The asymmetry was the dangerous part: that same raw value makes `expect_flag` fail *loudly* for the six named skills — annoying but safe — while it made the new loop pass *silently* for an eighth.

**The first attempt fixed it in the wrong place**, and the security review caught that as a MEDIUM. Normalising inside the shared `invocation_flag` also relaxed the approval-skill assertion: `disable-model-invocation: "TRUE"` on `approve-merge` went from failing loudly to passing silently — and a quoted YAML scalar is a *string*, not a boolean, so the test could report "locked" for a value the runtime may not treat that way. Same shape as the bug being fixed: one function, two call sites, opposite needs.

Now split. `invocation_flag` stays **raw and strict** for the per-family checks; `invocation_flag_normalised` is used **only** by the every-skill sweep, where a silently-skipped lock is the failure the assertion exists to catch. Verified in both directions: `"TRUE"` on `approve-merge` fails loudly on the strict path, `"TRUE"` on `/decide` is caught by the sweep.

**The strongest argument against this change resolves the other way.** The AgDR gates never depended on `/decide`: `require-agdr-for-arch-changes.sh` passes on a staged `docs/agdr/` file **or** a commit message merely containing `AgDR-<n>`, and `require-agdr-for-arch-pr.sh` checks the PR title/body for the same pattern and **never verifies the referenced file exists**. (`require-agdr-for-arch-pr.sh` additionally carries a documented not-applicable bypass; `require-agdr-for-arch-changes.sh` does not — its only exemption is spike/prototype class, and it closes by telling the author to write a minimal AgDR rather than bypass.) So the agent could always satisfy them without the skill — the lock's real effect was to push it toward a hand-rolled file or a bare string instead of `/decide`'s enforced Y-statement, options table, and justification. It made compliance worse, not safer.

## Testing

1. **Mutation-proven discriminating.** Re-applying the lock to `/decide` makes the new assertion fail with `/decide is human-only but is not in the justified lock list`; restoring gives 8/8. Re-run across five lock spellings — `true`, `"true"`, `TRUE`, `True`, `true # locked`, trailing-space — **all five now caught**; before normalisation only the bare form was.
2. The six existing family assertions and the autonomous-path coupling check are untouched and still pass.
3. `shellcheck --severity=warning` and `bash -n` clean.
4. Full suite `PASS=127 FAIL=3` — `test_suggest_mcp_search`, `test_token_efficiency_wave1`, `test_harnessability_scoring`. All three pass on clean `upstream/dev` (verified in a worktree) and fail only in a working tree carrying untracked local skills, because the skill-count assertion counts tracked skills. Unrelated to this change.

Refs #1094

---

## Glossary

| Term | Definition |
|------|------------|
| `disable-model-invocation` | SKILL.md frontmatter making a skill human-only. The model cannot invoke it at all. |
| AgDR | Agent Decision Record — the durable write-up of a material technical decision and its trade-offs. A record, not an approval. |
| approval marker | A file under `.claude/session/reviews/` that a merge gate reads. What makes the approval skills load-bearing, and what `/decide` conspicuously does not write. |
| inverted assertion | Checking a property across the whole set rather than a named subset, so a new member cannot slip in unasserted. |

### DIFF
--- a/.claude/hooks/tests/test_skill_invocability_gates.sh
+++ b/.claude/hooks/tests/test_skill_invocability_gates.sh
@@ -37,14 +37,23 @@ SKILLS="$REPO_ROOT/.claude/skills"
 PASS=0
 FAIL=0
 
-# Read the frontmatter value. Only the frontmatter block counts, so the key
-# is matched at line-start and only before the closing `---`.
+# Read the frontmatter value, RAW — exactly as written. Only the frontmatter
+# block counts, so the key is matched at line-start and only before the
+# closing `---`.
 #
 # Trailing \r is stripped first: a CRLF-committed SKILL.md would otherwise
 # yield "true\r", which compares unequal to "true" and fails this test for a
 # line-ending reason rather than a real one. This repo has been bitten by
 # exactly that before (#1019, where config_get left an embedded \r on Windows
 # checkouts), so the guard is cheap insurance rather than theory.
+#
+# The value is deliberately NOT normalised here. The per-family assertions
+# below must stay strict: for the three approval skills, only the canonical
+# bare `true` counts as locked. A quoted scalar (`"true"`) is a YAML *string*,
+# not a boolean, and whether the harness coerces it is unverified — accepting
+# it would let this test report "locked" for a value the runtime may treat as
+# unlocked. Strict comparison fails LOUDLY on any unrecognised spelling, the
+# safe direction for the most safety-critical assertion in this file.
 invocation_flag() {
   local file="$1"
   awk '
@@ -57,6 +66,27 @@ invocation_flag() {
       exit
     }
   ' "$file"
+}
+
+# NORMALISED value — trailing `# comment`, surrounding quotes, whitespace and
+# capitalisation removed. Used ONLY by the every-skill sweep at the bottom.
+#
+# The two accessors exist because the same spelling needs opposite treatment at
+# the two call sites (#1094 review, security finding). In the per-family checks
+# an unrecognised spelling fails loudly — annoying but safe. In the sweep it
+# made a LOCKED skill read as unlocked and be skipped SILENTLY, which is exactly
+# the case that assertion exists to catch. Normalising only the sweep closes
+# that fail-open without loosening the strict path.
+invocation_flag_normalised() {
+  local file="$1"
+  invocation_flag "$file" | awk '
+    {
+      sub(/[[:space:]]*#.*$/, "")          # strip a trailing comment
+      gsub(/^[[:space:]]+|[[:space:]]+$/, "")
+      gsub(/^["\x27]|["\x27]$/, "")        # strip surrounding quotes
+      print tolower($0)
+    }
+  '
 }
 
 expect_flag() {
@@ -132,6 +162,55 @@ if [ "$REVIEW_UNLOCKED" = "1" ] && [ "$APPROVAL_LOCKED" = "0" ]; then
   FAIL=$((FAIL + 1))
 else
   echo "PASS: no autonomous open -> review -> approve -> merge path"
+  PASS=$((PASS + 1))
+fi
+
+# ---------------------------------------------------------------------------
+# Every LOCKED skill must be justified (#1094).
+#
+# The family checks above pin the six skills someone thought about. They
+# structurally cannot catch a SEVENTH skill acquiring the flag — which is
+# exactly what happened: /decide and /audit-deps were human-only outside
+# AgDR-0110's reasoning and outside this test's coverage, and nobody noticed
+# because nothing asserted over the whole set.
+#
+# /decide's lock was the costly one. `.claude/rules/agdr-decisions.md` opens
+# with "HARD STOP … run /decide" and repeats "→ /decide" four more times, so
+# the framework was ordering an action it had mechanically forbidden. That is
+# the same self-contradiction the review-skill justification above names for
+# auto-code-review.sh — one skill over, and undetected.
+#
+# So assert the inverse: iterate EVERY skill, and require any that is locked to
+# appear in an explicit allow-list. Adding a lock now means editing this list
+# and stating why, which is the point.
+# ---------------------------------------------------------------------------
+
+# Only skills whose INVOCATION IS THE APPROVAL belong here (AgDR-0110).
+# A skill that merely writes a document, reads state, or reports does not.
+LOCK_ALLOWED="approve-merge approve-design approve-architecture"
+
+UNJUSTIFIED=0
+for f in "$SKILLS"/*/SKILL.md; do
+  [ -f "$f" ] || continue
+  name=$(basename "$(dirname "$f")")
+  [ "$(invocation_flag_normalised "$f")" = "true" ] || continue
+  case " $LOCK_ALLOWED " in
+    *" $name "*) ;;
+    *)
+      echo "FAIL: /$name is human-only but is not in the justified lock list."
+      echo "   Only approval skills may be locked — the ones where invoking the"
+      echo "   skill IS the human approval (AgDR-0110). If this lock is"
+      echo "   deliberate, add /$name to LOCK_ALLOWED with a stated reason, AND"
+      echo "   check that no rule instructs the agent to invoke it — a rule that"
+      echo "   does is a contradiction the agent cannot resolve (#1094)."
+      FAIL=$((FAIL + 1))
+      UNJUSTIFIED=1
+      ;;
+  esac
+done
+
+if [ "$UNJUSTIFIED" = "0" ]; then
+  echo "PASS: every human-only skill is in the justified lock list"
   PASS=$((PASS + 1))
 fi
 
--- a/.claude/skills/audit-deps/SKILL.md
+++ b/.claude/skills/audit-deps/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: audit-deps
 description: Audit dependencies for vulnerabilities, outdated packages, and license compliance.
-disable-model-invocation: true
+disable-model-invocation: false
 argument-hint: "[project-path]"
 allowed-tools: Bash, Read, Grep, Glob
 ---
--- a/.claude/skills/decide/SKILL.md
+++ b/.claude/skills/decide/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: decide
 description: Make a technical decision with structured reasoning and create an Agent Decision Record (AgDR).
-disable-model-invocation: true
+disable-model-invocation: false
 argument-hint: "<what you're deciding>"
 ---
 


atlas-apex (commented):
## Code Review: PR #1101

**Commit**: `7011b29352e682665ec863c4176d85f2e8416cb3`

### Summary

Unlocks `/decide` and `/audit-deps` (`disable-model-invocation: true` → `false`) and adds an inverted assertion to `test_skill_invocability_gates.sh` requiring any locked skill to appear in an explicit `LOCK_ALLOWED` list.

I tested the security claim adversarially rather than accepting it. It holds — and the strongest counterargument resolves *in favour* of the change, for a reason the PR body does not state.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass**
- Security: **Pass**
- Performance: **N/A**
- Technical Decisions (AgDR): **Pass** (see note)
- Adopter Handbooks: **N/A** — no findings (no migrations, no domain/infra code)
- Fallow Static Analysis: **N/A** — not a JS/TS diff

---

### The security question: does a model-invocable `/decide` let the agent satisfy the AgDR gates for itself?

**No — and more importantly, it never could have prevented that, because the agent could already satisfy both AgDR gates without `/decide` at all.**

I read both gates rather than trusting the framing:

- `require-agdr-for-arch-changes.sh` passes on **either** a staged file matching `^docs/agdr/AgDR-[0-9]+-.*\.md$` (line 176) **or** a commit message merely containing `AgDR-[0-9]+` (line 200). Both are producible with a plain `Write` or a string in `-m`.
- `require-agdr-for-arch-pr.sh` checks the PR title+body for `AgDR-\d+-[a-z0-9-]+`. **It never verifies the referenced file exists** — there is no existence check against `docs/agdr/` anywhere in its 686 lines (every `[ -f ]` is for the body-file or a sourced lib).
- Both carry the documented `<!-- agdr: not-applicable -->` bypass (lines 208–211), which exits 0 with a warning.

So these are "leave an auditable trace" prompts with visible bypasses — precisely what `agdr-decisions.md` § *"Enforcement — and the honest limits of it"* already says they are. They read **no** marker under `.claude/session/reviews/`, and they are structurally unlike the merge gates AgDR-0110 protects, where the invocation *is* the approval.

That inverts the counterargument. The lock never functioned as a gate control. What it actually did was force the agent down the **unstructured** path — a hand-rolled file, or a bare `AgDR-0042` string in a commit message — instead of the structured one (`/decide` enforces a Y-statement, a ≥2-option table, a mandatory `because` clause, ID sequencing, and the Contrarian offer). Unlocking therefore *improves* the quality of the artifact that satisfies a gate the agent could already satisfy. The lock made compliance worse, not safer.

Verified separately: `/decide` writes only `docs/agdr/AgDR-NNNN-{slug}.md`. No write under `.claude/session/reviews/` anywhere in the skill.

**Honest residual:** unlocking removes a theoretical human-at-decision-time speed bump. But it was never enforced, and under a rule that orders `/decide` five times the realistic behaviour *under* the lock was agent confusion or hand-rolled AgDRs — not human involvement. Low, and correctly traded.

**`/audit-deps`:** `allowed-tools: Bash, Read, Grep, Glob` — no Write/Edit. Its only appearance in any hook is a *comment* at `require-agdr-for-arch-pr.sh:26` ("version-only bumps are an `/audit-deps` concern"), not a consult. Gates nothing. Uncontroversial.

**Coupling check intact:** the diff is a pure addition after line 135; the `REVIEW_UNLOCKED` / `APPROVAL_LOCKED` block is byte-untouched and passing.

**Mutation evidence reproduced:** re-locking `/decide` gives `Passed: 7 Failed: 1` with the intended message; restoring gives **8/8**, and `git diff --stat` on the restored file is empty.

---

### Issues Found

None blocking.

### Suggestions

**1. `nit:` The new loop's lock detection is fail-open on six alternative spellings.**

`[ "$(invocation_flag "$f")" = "true" ]` recognises only the exact bare lowercase `true`. I probed the parser directly:

| Value | Parsed | Result |
|---|---|---|
| `true` | `true` | FLAGGED |
| `"true"` | `"true"` | **MISSED** |
| `true # locked` | `true # locked` | **MISSED** |
| `true ` (trailing space) | `true ` | **MISSED** |
| `TRUE` / `True` | `TRUE` / `True` | **MISSED** |
| `yes` | `yes` | **MISSED** |

A missed value is silently treated as *unlocked* — which is the exact failure mode this assertion exists to prevent ("a locked skill the loop fails to see").

Note the asymmetry this introduces: the *same* parser output makes `expect_flag` fail **loudly** for the six named skills (`expected 'true', got '"true"'`) but makes the new loop pass **silently**. Severity depends on something I can't verify from here — whether Claude Code's own frontmatter parser is more permissive than this `awk`. If the harness honours only bare `true`, the blind spot is empty and this is cosmetic; if it honours `"true"` or `TRUE`, it's real. Cheap fix: normalise (strip quotes, `#`-comments, trailing whitespace; lowercase) before comparing — and it would harden `expect_flag` at the same time.

**2. `nit:` "iterates **every** `SKILL.md`" oversells slightly.** True for directory coverage — I confirmed the glob sees all 90 skill dirs and that no nested `SKILL.md` exists (`find -mindepth 3` → 0). But given the claim being sold is durability against "a *seventh* skill acquiring the flag," a seventh skill acquiring it as `"true"` or `TRUE` still slips through. Worth scoping the claim to match suggestion 1.

**3. `suggestion:` The body's strongest counterargument is absent.** "…satisfying no merge gate" is *precisely* correct (no merge gate reads an AgDR) — but the body never engages with the fact that `/decide`'s **output** does satisfy the two AgDR gates. That's the first objection a careful reviewer raises, and I had to derive the answer myself. Since the answer favours the change, stating it would make the body materially stronger rather than weaker.

**4. Note, not a defect:** the glob scans **untracked local** skill dirs too (12 in this working tree). So a locally-installed marketplace skill shipping `disable-model-invocation: true` would fail this test in an operator's tree while passing on clean `upstream/dev` — the same local-tree-only class the body already documents for the other three tests. None of the 12 is currently locked (hence 8/8).

### Body claims — verified

Checked with the same lens that caught overreaches on #1096 and #1098. These all hold:

- `HARD STOP … run /decide` at `agdr-decisions.md:3` ✓
- `→ /decide` repeated **exactly** four more times (lines 55–58) ✓
- `AgDR-0110` and `pr-workflow.md` are the only docs mentioning the flag; **neither names `/decide` or `/audit-deps`** ✓
- `shellcheck --severity=warning` and `bash -n` clean ✓
- Six family assertions + coupling check untouched ✓

The only reach is #2 above — a scope overstatement, not a false fact.

### Note on the absent-key reasoning and AgDR

`false`-over-key-removal holds for the new loop too: absent → empty → not `true` → skipped, which is the correct semantics (absent = model-invocable = unlocked = nothing to justify), while the six family checks still demand an explicit posture. Consistent.

On AgDR: `LOCK_ALLOWED` is exactly AgDR-0110's three approval skills, so this reads as *implementing* AgDR-0110 across the whole set rather than making a new architectural call — and it links #1094 and cites AgDR-0110 throughout. Reasonable; flagging it only so you can disagree if you read the "only approval skills may be locked" rule as a new cross-cutting decision in its own right.

### Verdict

**APPROVED**

The change is small, the test is mutation-proven discriminating, and the security reasoning survives adversarial checking — the AgDR gates are text-pattern prompts with documented bypasses, not approval controls, so unlocking `/decide` grants the agent nothing it lacked and upgrades the artifact it produces. All four suggestions are non-blocking.

---
📌 Reviewed commit: `7011b29352e682665ec863c4176d85f2e8416cb3`

atlas-apex (commented):
## Code Review: PR #1101 (re-review)

**Commit**: `2f7469d38f37ec8762630d915515b45be3dc8478`
**Prior review**: APPROVED at `7011b29`. Scoped to the delta — `invocation_flag` normalisation, commit message, PR body. Suite, AgDR-gate analysis, and the coupling check were verified at `7011b29` and are unaffected by a parser change confined to one test file.

### Summary
The fail-open I flagged is genuinely closed. `invocation_flag` now strips a trailing `# comment`, surrounding quotes, and whitespace, then lowercases — so a lock can no longer read as unlocked through a spelling variation. I re-ran the five spellings plus eleven you hadn't tested.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass
- Performance: N/A
- PR Description & Glossary: Pass (one factual correction below)
- Technical Decisions (AgDR): N/A (verified at `7011b29`)
- Adopter Handbooks: N/A

### 1. Lock spellings — re-run independently

All five you tested are now caught. So are six more I added:

| Spelling | Result |
|---|---|
| `true`, `"true"`, `TRUE`, `True`, `true # locked`, `true ` (trailing) | caught |
| `'true'` (single quotes) | caught |
| `  true  ` (leading + trailing) | caught |
| `true#nospace` | caught |
| `"TRUE" # why` | caught |
| `"true` / `'true"` (mismatched quotes) | caught — **fails closed**, the safe direction |

**Spellings that still evade the loop** (all pre-existing; normalisation did not introduce any):

- `yes` / `Yes` / `on` / `1` — YAML 1.1 truthy aliases. Normalise to themselves, so the loop skips them as unlocked.
- Folded/block scalar — `disable-model-invocation: >` with `true` on the next indented line yields `>`, skipped.
- `""true""` (double-nested quotes) → `"true"`, skipped. The `gsub` anchors strip exactly one quote per end. Not valid YAML anyone writes.
- Empty value, or a value that is only a comment → empty, skipped. **Correct** — YAML null defaults to model-invocable.

How much the `yes` family matters: all 25 shipped `SKILL.md` files use the bare lowercase form (22 `false`, 3 `true`), and YAML 1.2 parsers — which is what modern frontmatter loaders default to — read `yes` as the *string* `"yes"`, not boolean true. So a skill spelled that way probably isn't actually locked either, and the loop reading it as unlocked would be right. Real but narrow.

suggestion: (non-blocking) if you want this fail-closed against every future spelling rather than an enumerated set, invert the loop's predicate — treat anything that is **not** `false` and **not** empty as suspicious, rather than only `= "true"`. That catches `yes`, `on`, `1`, and the folded scalar without you having to anticipate them. The six named `expect_flag` assertions can keep their exact comparison, since they already fail loudly on anything unrecognised.

### 2. Did normalisation open a hole in the other direction?

No. Taking your specific question — `FALSE` now normalises to `false` and **passes** for the review skills where it previously failed loudly.

That is acceptable, not a new hole. `FALSE` and `false` are the same YAML boolean, so reading it as false is semantically correct; what you lose is only the enforcement of canonical lowercase spelling in frontmatter, which was an accident of the old comparison rather than an intended assertion. The direction that carries safety is unchanged: no value meaning "locked" reads as "unlocked" that didn't before, and the mismatched-quote cases resolve **toward** being caught. Normalisation strictly improves the dangerous direction and never regresses it.

### 3. The awk normalisation itself

- **`sub(/[[:space:]]*#.*$/, "")` strips from the first `#`** — and it runs *before* quote-stripping, so a quoted value legitimately containing `#` would be truncated there. For a boolean flag no legitimate `#` exists, so this cannot misread anything real. Worth knowing only if this helper is ever reused for a non-boolean key.
- **Mismatched quotes** (`"true` with no closing quote) strip to `true` and read as a lock. That is the fail-closed direction — a malformed lock is still treated as a lock. Correct behaviour.
- **`\x27` portability** — I probed whether the bracket set degrades to `["x27]` on an unsupporting awk, which would silently stop stripping single quotes and reopen the fail-open on that platform. It does not degrade here (BWK awk 20200816): `x true x` and `7true7` pass through untouched while `'true'` strips correctly. gawk and mawk both support `\x` too; only `gawk --posix` would not, which nothing here invokes. nit: a comment naming why the hex escape is used would save the next reader the same probe.

### 4. PR body / commit message

The correction paragraph is accurate — it states the fail-open, names the asymmetry (loud for the six, silent for an eighth) as the dangerous part, and does not overstate the fix. The AgDR-gate paragraph is framed correctly as the strongest argument resolving in the change's favour, and the load-bearing half of it holds.

**One clause reaches past what I established.** Both the PR body and the commit message say:

> Both carry a documented not-applicable bypass.

That is true of `require-agdr-for-arch-pr.sh` (`<!-- agdr: not-applicable -->`, warns and exits 0). It is **not** true of `require-agdr-for-arch-changes.sh` — its only bypass is the spike/prototype exemption, which is conditional on the active ticket's class rather than an author declaration, and its closing comment explicitly says the opposite posture: *"that's the signal to create a minimal AgDR rather than bypass the rule."*

This does not weaken your argument — the conclusion rests on the two mechanisms you state before it (the commit hook accepting a message merely containing `AgDR-<n>`, and the PR hook matching title/body without verifying the file exists), and those carry it on their own. The bypass sentence is reinforcement, and it's wrong for one of the two hooks. Worth fixing in the PR body since it's attributed to my finding; the commit message isn't worth an amend + force-push for a decorative clause.

### Verdict
**APPROVED**

The hole is closed in the direction that mattered, the fix is verified against sixteen spellings rather than the five claimed, and no new hole opened. The two items above are a non-blocking suggestion and a prose correction.

---
Reviewed commit: `2f7469d38f37ec8762630d915515b45be3dc8478`

atlas-apex (commented):
## Security Review: PR #1101

**Commit**: `2f7469d38f37ec8762630d915515b45be3dc8478`

### Summary

Unlocks `/decide` and `/audit-deps` (`disable-model-invocation: true` → `false`) and adds an inverted assertion to `test_skill_invocability_gates.sh` requiring every locked skill to appear in an explicit `LOCK_ALLOWED` list, plus value normalisation in `invocation_flag`.

This is a trust-chain change (skill invocability + the test that pins it), so it takes the Heavy path regardless of diff size. I verified the load-bearing claims by execution rather than accepting the prior review.

### Checklist Results

- Secrets & Credentials:  Pass (no credentials touched)
- XSS Prevention:         N/A
- Auth & Authorisation:   Pass — see Q1/Q3 below
- Data Protection:        N/A
- API Security:           N/A
- Gate & Trust-Chain Integrity (#777): Pass with one advisory

### Q1 — Is the AgDR-gate conclusion right? YES (verified)

The claim that unlocking `/decide` grants no capability holds. Both hooks are satisfiable by text the agent already produces, with no dependency on the skill:

- `require-agdr-for-arch-changes.sh` passes on **either** a staged `docs/agdr/AgDR-N-*.md` (l.176) **or** a commit message merely matching `AgDR-[0-9]+|docs/agdr/AgDR-` (l.200). Both are plain Write/Bash operations.
- `require-agdr-for-arch-pr.sh` greps the title+body haystack for `AgDR-[0-9]+-[a-z0-9-]+` (l.635) and **never verifies the referenced file exists**. It additionally honours a documented `<!-- agdr: not-applicable -->` bypass (l.208) that exits 0 with a warning.

`/decide` writes no marker, references no `.claude/session/reviews/` path, and performs no merge (grepped: zero matches). The AgDR is a documentation record, not an authorisation artifact. **No capability grant.**

### Q2 — Does unlocking `/audit-deps` grant anything? NO

`allowed-tools: Bash, Read, Grep, Glob` — `allowed-tools` **narrows** a skill's surface, it cannot grant beyond what the invoking context holds, and the orchestrator already has unrestricted `Bash`. The skill body runs `npm audit --json` and `npm outdated --json` only: no `install`, no `npx`, no `curl`, so no package-lifecycle-script execution path is introduced. No `Write`, so it cannot produce a marker. **No new arbitrary-execution path.**

### Q3 — Is the coupling invariant intact? YES (verified by execution)

`bash .claude/hooks/tests/test_skill_invocability_gates.sh` → **8/8 pass**, including `PASS: no autonomous open -> review -> approve -> merge path`. Enumerating the flag across all 25 skills that carry it: `approve-merge`, `approve-design`, `approve-architecture` are the **only** three set to `true`; every other skill is `false`. The three review skills remain unlocked, the three approval skills remain locked. This PR touches neither family.

### Security Issues Found

**MEDIUM — normalisation loosens the approval-skill assertion in the fail-open direction (non-blocking; zero current exposure)**

The PR's own comment argues normalisation "mattered asymmetrically… Normalising fixes both call sites at once." That is only half true, and the unexamined half moves the wrong way.

`invocation_flag` is shared by **both** call sites, so normalisation also relaxes `expect_flag` for the three approval skills. Empirically confirmed:

```
disable-model-invocation: "TRUE"   →  normalises to  true
```

Before this PR, that value printed `"TRUE"`, compared unequal to `true`, and **failed loudly** on `approve-merge`. After this PR it passes silently. But a YAML *quoted scalar* is a string, not a boolean — whether the harness's frontmatter parser coerces `"TRUE"` to boolean true is unverified. If it does not, the skill is model-invocable at runtime while this test reports it locked. Normalisation therefore teaches the test to bless spellings whose runtime semantics nobody has checked, on the framework's most safety-critical assertion.

Severity is MEDIUM, not higher, because **exposure today is zero**: all three approval skills use the canonical bare `true` (verified). The risk is that a future non-canonical edit now lands unchallenged.

Recommended fix (not required to merge): keep normalisation for the every-skill sweep, where reading a lock as unlocked is the silent failure it was added to catch, but assert the **raw** value for the approval family — those three should be required to be exactly `true`, so any deviation still fails loudly.

**LOW — `yes` / `on` / `1` still evade the sweep**

Confirmed empirically (`yes` → `yes`, `1` → `1`, neither equals `true`). These are genuine YAML 1.1 booleans, so a skill locked with `yes` is skipped by the new assertion. Already noted as residual by the prior review; I confirm it and agree it is non-blocking. Worth noting the direction differs by call site: for the approval family a `yes` fails loudly (safe); only the sweep is silent.

**LOW — two hardcoded lists can drift**

`LOCK_ALLOWED` (sweep) and the `for s in approve-merge approve-design approve-architecture` family list are independent. Adding a future approval skill to `LOCK_ALLOWED` without adding a matching `expect_flag` would satisfy the sweep while leaving the new skill's lock unasserted. Deriving one list from the other would close this.

### Q4 — Does the inverted assertion make the chain safer, or just feel safer?

Genuinely safer, but strictly at CI time, and the PR should not be read as adding a runtime control. It closes a real structural gap: the family checks pin six named skills and cannot, by construction, notice a seventh acquiring the flag — which is exactly the defect being fixed. Making a new lock require an explicit, justified list edit is the right shape.

Its limits, to be stated plainly:
- There is **no runtime consequence** to an accidental or misspelled lock. The harness reads the frontmatter directly; nothing consults this test at invocation time. A bad lock between CI runs is live.
- A mis-spelled *lock* (`yes`, `on`, folded scalar) is invisible to the sweep, so the very case with no runtime backstop is also the one the test can miss.
- The failure mode it prevents is over-locking (an agent told to run a skill it cannot). Under-locking of an approval skill is caught by the separate coupling assertion, which this PR leaves intact.

### Q5 — Anything the unlock makes worse

Nothing security-relevant beyond the MEDIUM above. One consequence worth recording, which is not a weakening: a model-invocable `/decide` means the agent can now self-generate the AgDR that documents its own decision. Since the gate was already text-satisfiable, the mechanism is unchanged — this PR makes explicit what was already true, that the AgDR gate is a **documentation** gate, not an authorisation gate. Anyone who believed it constrained agent behaviour should recalibrate on that basis rather than on this diff.

`/decide` carries no `allowed-tools` line and so inherits the full tool set; that is not an escalation, since skills expand in the invoking context which already holds those tools.

### Verdict

**APPROVED**

The load-bearing argument checks out under independent verification: unlocking grants no capability the agent lacked, and the review/approval coupling is untouched and passing. The one MEDIUM finding is a test-hygiene regression with no live exposure — worth a follow-up, not a block.

---
📌 Reviewed commit: `2f7469d38f37ec8762630d915515b45be3dc8478`

atlas-apex (commented):
## Security Re-Review: PR #1101

**Commit**: `73c7772718b30cb30b4143fa5ef9ac1b5bb05641` (was `2f7469d`)

### Scope

Re-review confined to the delta. Per the prior pass, the AgDR-hook analysis, `/audit-deps`' tool surface, and the skill frontmatter were established at `2f7469d` and are untouched here — not re-verified.

### MEDIUM from the prior review: RESOLVED

The shared accessor is split, which is the fix I recommended. Verified rather than accepted:

- `git diff --stat 2f7469d 73c7772` → **1 file, `test_skill_invocability_gates.sh` only** (+31/-13). No skill frontmatter moved.
- `invocation_flag` (strict) is **byte-identical** to its form at the PR base `59c3cc6` — confirmed by extracting the function from both revisions and diffing. The `"TRUE"` case fails loudly again on the approval family.
- Call routing is correct: strict at l.103 (`expect_flag`, the six family assertions) and l.150/155 (coupling); `invocation_flag_normalised` is used at **exactly one** site, l.196, the every-skill sweep.
- Coupling block (`REVIEW_UNLOCKED`/`APPROVAL_LOCKED`) is **byte-identical** to the base.
- Worktree copy matches `73c7772`; suite → **8/8**.

Same spelling now gets opposite treatment at the two call sites: loud failure where a missed lock would be a fail-open on the approval family, silent-catch where a skipped lock is the failure the sweep exists to find. That is the asymmetry the finding asked for.

### Findings

None new. The two LOW items stand and are recorded on #1094 as follow-up: `yes`/`on`/`1` and folded scalars still evade the sweep. I agree with deferring them — inverting the predicate to treat anything neither `false` nor empty as suspicious changes the assertion's behaviour rather than fixing the bug this ticket describes, and it belongs in its own diff with its own review.

### Verdict

**APPROVED** at `73c7772718b30cb30b4143fa5ef9ac1b5bb05641`

---
📌 Reviewed commit: `73c7772718b30cb30b4143fa5ef9ac1b5bb05641`

atlas-apex (commented):
## Code Review: PR #1101 (re-review 2 — accessor split)

**Commit**: `73c7772718b30cb30b4143fa5ef9ac1b5bb05641`
**Prior**: APPROVED at `7011b29`, APPROVED at `2f7469d`. Scoped to the delta only; suite and AgDR-gate analysis unchanged and not re-derived.

### The security finding was right, and my `2f7469d` approval was wrong on it

I checked the normalisation in the direction I had flagged — could a lock read as unlocked — and cleared it. I did not check the *other* call site of the same function. Normalising inside `invocation_flag` also relaxed the approval-skill assertion, so `disable-model-invocation: "TRUE"` on `approve-merge` went from failing loudly to passing silently. That is the most safety-critical comparison in the file, and a quoted YAML scalar is a string rather than a boolean, so the test could have reported "locked" for a value the runtime may not treat that way.

It is the same shape as the bug being fixed — one function, two call sites, opposite needs — which is precisely why I should have caught it. Recording that plainly: I verified the fix against the hole I named and did not re-ask who else depended on the changed function.

### Delta verification

**1. Confined to the test file.** `git diff --stat 2f7469d 73c7772` → one file, `.claude/hooks/tests/test_skill_invocability_gates.sh`, +31/-13. No skill frontmatter moved.

**2. The strict path is byte-equivalent to its pre-PR form.** Extracted `invocation_flag` from the merge base (`59c3cc6`) and from `73c7772` and diffed: identical. The comparison I originally relied on is genuinely restored, not approximated.

**3. Call sites are split as described:**

| Line | Accessor | Used by |
|---|---|---|
| 103 | `invocation_flag` (raw) | `expect_flag` — the six per-family assertions |
| 150, 155 | `invocation_flag` (raw) | the autonomous-path coupling check |
| 196 | `invocation_flag_normalised` | the every-skill sweep, and nothing else |

The coupling check staying on the raw accessor matters and is correct — it is the assertion guarding the `open → review → approve → merge` path, so it should be strict for the same reason the family checks are.

**4. Both directions behave as claimed.** Sourcing the two accessors straight out of the file under review:

- Sweep (normalised) still catches every lock spelling: `true`, `"true"`, `TRUE`, `True`, `true # locked`, trailing-space, and `'true'`.
- Strict (raw) fails loudly on `"true"`, `TRUE`, `True`, `true # locked` — only canonical bare `true` passes. The `"TRUE"`-on-`approve-merge` silent pass is closed.

**5. Baseline.** The single test at HEAD: `Passed: 8  Failed: 0`.

One implementation note, no action needed: `invocation_flag_normalised` pipes through a second `awk` whose block has no `exit`. Safe here because the raw accessor emits at most one line, and an absent key yields empty input so the block never runs and the sweep correctly skips.

**PR body correction applied and accurate.** The bypass claim is now scoped to `require-agdr-for-arch-pr.sh`, with `require-agdr-for-arch-changes.sh` explicitly excluded and its spike/prototype-only exemption named. That matches what the hook actually does.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass (MEDIUM from the security review resolved at source, not worked around)
- Technical Decisions (AgDR): N/A (verified at `7011b29`)

### Residual (tracked, not blocking)
`yes` / `Yes` / `on` / `1` and folded scalars still read as unlocked by the sweep, and my suggestion to invert the predicate to "not `false` and not empty" — both recorded on #1094 as follow-up. Unchanged by this delta.

### Verdict
**APPROVED**

The split is the right fix: it closes the fail-open in the sweep without loosening the assertion that actually guards the merge chain, and it restores the strict path byte-for-byte rather than re-deriving it.

---
Reviewed commit: `73c7772718b30cb30b4143fa5ef9ac1b5bb05641`

me2resh (approved):

atlas-apex (commented):
## Code Review: PR #1101 (re-review 3 — frontmatter comment removal)

**Commit**: `6977daa6fde8041aa4f9e0599f73679eb6ea02eb`
**Prior**: APPROVED at `7011b29`, `2f7469d`, `73c7772`. Scoped to the delta; suite not re-run (CI green), AgDR analysis and the accessor split unaffected.

### Delta verification

**1. Two files, deletions only.** `git diff --stat 73c7772 6977daa` → `.claude/skills/audit-deps/SKILL.md` (-3) and `.claude/skills/decide/SKILL.md` (-6). Nine lines removed, zero added, every removed line a `#` comment inside the frontmatter block.

**2. Flag values unchanged.** Read through the strict accessor from the test under review — `decide` and `audit-deps` are `false`; the three approval skills remain `true`; the three review skills remain `false`. Nothing moved.

**3. `test_skill_invocability_gates.sh` is byte-identical** to `73c7772` (`git diff --quiet` on the path). The accessor split I reviewed last round is untouched.

**4. The single invocability test passes: 8/8.**

**5. Descriptions are back under the cap** — `decide` 95 chars, `audit-deps` 82, against the 200 hard cap.

### On the trap, and the one thing worth doing about it

The "same test name, different cause" note is the useful part of this round. A pre-existing local failure masking a new one with the same identity is a general hazard here, not a one-off: this fork carries untracked private skills, so any test whose assertion ranges over *all* skills will fail locally for reasons that have nothing to do with the branch. `test_token_efficiency_wave1.sh` is exactly that shape — and so is the every-skill sweep this PR just added, which is worth noticing, since the sweep is now a second test with the same local-noise profile.

I checked whether the defect had siblings: **no other tracked `SKILL.md` carries a `#` comment inside its frontmatter block.** These two were the only instances, so the fix is complete rather than a first pass at a pattern.

I'd also record that the rationale did not need to be in the frontmatter at all — it's now in the commit body, the PR body, and #1094, which are the durable homes for it. The frontmatter is loaded into context on every skill listing; prose that explains a past decision earns nothing there and costs on every load. That's the cap doing its job, not merely catching a formatting slip.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass (CI green; single invocability test 8/8)
- Security: Pass — no security-relevant surface in this delta; the approval-skill locks are byte-verified unchanged
- Technical Decisions (AgDR): N/A (verified at `7011b29`)

### Residual (tracked, not blocking)
`yes` / `on` / `1` and folded scalars still read as unlocked by the sweep, plus the predicate-inversion suggestion — recorded on #1094. Unchanged by this delta.

### Verdict
**APPROVED**

Deletions only, in the two files named, with the flags and the test file verified untouched.

---
Reviewed commit: `6977daa6fde8041aa4f9e0599f73679eb6ea02eb`

atlas-apex (commented):
## Security Re-Review: PR #1101

**Commit**: `6977daa6fde8041aa4f9e0599f73679eb6ea02eb` (was `73c7772`)

### Scope

Delta only. The AgDR-hook analysis and `/audit-deps`' tool surface were established at `2f7469d`; the accessor split was validated at `73c7772` and the test file is untouched here. Neither re-verified.

### The question asked: did an edit adjacent to the invocability key disturb its parse? NO

Right question to ask — a frontmatter edit next to `disable-model-invocation` is not assumable-harmless on the trust chain. Verified in the way that actually settles it, by reading the value through the shipped parser on both sides of the change rather than reasoning about the diff:

```
decide                 before=[false]  after=[false]
audit-deps             before=[false]  after=[false]
approve-merge          before=[true]   after=[true]
approve-design         before=[true]   after=[true]
approve-architecture   before=[true]   after=[true]
```

Read with the strict `invocation_flag` accessor exactly as shipped. **No value moved.** Suite → 8/8, including the coupling assertion and the every-skill sweep.

Why the parse was structurally safe: the accessor anchors on `/^disable-model-invocation:/` at line start and terminates only on the closing `---`. It does not count lines, track position, or care what precedes the key — so removing preceding lines cannot shift what it matches. The removal also cannot *un*-comment anything, since every deleted line begins with `#`.

### Diff confirmation

- `git diff --stat 73c7772 6977daa` → **2 files, 9 deletions, 0 insertions**: `audit-deps/SKILL.md` (-3), `decide/SKILL.md` (-6).
- Every deleted line is a `#` comment. Both `disable-model-invocation: false` lines appear as **unchanged context**, not as delete/add pairs — the values were never rewritten.
- Worktree copies of both SKILL.md files and the test file match `6977daa`, so the run above applies to the reviewed bytes.

### Findings

None. No security impact, consistent with the CI-only nature of the regression: the failing check was a description-length cap in `test_token_efficiency_wave1.sh`, which reads a description as every line up to the next `key:` line and does not treat a comment as a key. That is a parser-leniency quirk in the *token-efficiency* test, not an invocability concern.

Worth recording as a follow-up observation, not a blocker: the two frontmatter parsers in this repo disagree about comments — the invocability accessor ignores them, the token-efficiency one silently absorbs them into the preceding value. The rationale for each unlock now lives only in the PR and #1094 rather than beside the flag it explains, which is a small loss for the next person auditing this file. If that rationale is wanted in-tree, prose below the frontmatter is the safe home.

The two LOW items from the first review still stand on #1094 (`yes`/`on`/`1` and folded scalars evade the sweep).

### Verdict

**APPROVED** at `6977daa6fde8041aa4f9e0599f73679eb6ea02eb`

---
📌 Reviewed commit: `6977daa6fde8041aa4f9e0599f73679eb6ea02eb`

Issue #1092, opened 2026-07-31
[Bug] /dfd discovery walks into workspace/ and attributes other projects' dependencies to the target


atlas-apex:
## Bug Scenario

**Given** `/dfd` is run with the ops fork as the target (`/dfd .` from the fork root, or `/dfd <name>` where the project's working copy is the ops repo itself)

**When** `discover.sh` walks the tree for external actors, processes, and data stores

**Then** it descends into `workspace/`, which holds **clones of other registered projects**, and reports those projects' dependencies as actors of the target

**Expected** discovery stays within the target's own reachable tree. Rule 4 of the skill states discovery is *"reachability-bounded — a single-service DFD walks only what's connected to that service"*; it is not bounded against sibling project clones.

## Observed

Running discovery against the ops fork produced, among others:

```
actors:
  - id: ext_stripe
    type: external_service
    name: "stripe"
    evidence: "./workspace/<other-project>/packages/design-system/tests/Card.test.tsx:27"
  - id: auth_provider
    type: external_auth
    evidence: "./workspace/<other-project>/docs/self-hosting.md:195"
```

Neither is an actor of the framework. The first is a **string in another project's test file**; the second is a **mention in another project's documentation**. Both were attributed to the target as external integrations.

This also trips skill rule 6 — *"Don't invent third-party integrations. Every External Service actor must trace to a concrete SDK import, API key env var, or webhook handler. No guessing."* A test fixture and a docs mention are neither.

## Repro

```bash
cd <ops-fork-root>
bash .claude/skills/dfd/discover.sh . | head -30
```

Any adopter whose `workspace/` contains cloned projects will see those projects' dependencies attributed to whatever they point `/dfd` at.

## Severity

**medium** — no data loss and nothing is written without operator review, but the consequences are worse than a cosmetic false positive:

1. The DFD is the **declared input to `/threat-model`** (AgDR-0026). A polluted DFD produces a threat model enumerated against integrations the target does not have.
2. The operator-review step (skill step 3) is the designed mitigation, but it is calibrated for *"heuristic discovery has false positives, confirm these boundaries"* — not for *"most of this belongs to a different repository."* Reviewing 100% wrong output is not reviewing, it is re-authoring.
3. A DFD is exactly the artefact people trust *because* it was generated rather than hand-written.

## Suggested fix

[...]

PR #1107, opened 2026-08-01
fix(#1092): stop /dfd discovery from walking into workspace and worktrees

me2resh:
## Summary
- **`/dfd` discovery no longer walks into `workspace/` or `.claude/worktrees/`** — when the discovery target is the apexyard ops fork itself, `workspace/` holds full clones of *other* registered projects and `.claude/worktrees/` holds agent worktrees. Both are full checkouts of unrelated (or duplicate) code, so `discover.sh` and `classify.sh` were attributing other projects' dependencies (a Stripe import in another project's test fixture, an auth mention in another project's docs) to the target as if they were its own external actors — the exact failure #1092 describes.
- **Prunes the CONCRETE resolved paths, not a basename pattern** — rather than adding `workspace`/`worktrees` to the any-depth `SKIP_DIRS` basename list (which risks over-pruning a project that legitimately has a directory named `workspace`), both scripts now resolve `portfolio_workspace_dir()` (via `_lib-portfolio-paths.sh`) and `.claude/worktrees/` as exact paths, and only do so when the resolved `TARGET` is itself an ops fork (has `.apexyard-fork`, or both `onboarding.yaml` + `apexyard.projects.yaml`, at its root). A non-fork target is completely unaffected.
- **Regression fixture proves the leak is gone — and is genuinely load-bearing** — `tests/smoke.sh` Fixture 2b reuses the ops-fork sandbox built for Fixture 2, plants a distinct SDK import under `workspace/other-project/` and a second, *different* SDK import under `.claude/worktrees/`, then asserts neither evidence **path** appears in `discover.sh`'s or `classify.sh`'s output when the sandbox is the scan target. An earlier version of this fixture asserted on a content-only marker string that `discover.sh`/`classify.sh` never emit (evidence is always `path:line`, never file content) — Rex caught it in review as non-load-bearing. It's now fixed: path-based assertions, and two *different* vendor SDKs so discovery's per-vendor "first evidence wins" dedup can't mask one fixture behind the other (the sandbox's own copy of `_lib-multi-repo-trace.sh` also contains several bare vendor names as hostname literals, which was the second, subtler way the original fixture could pass for the wrong reason).

## Testing
- `bash .claude/skills/dfd/tests/smoke.sh` — all 55 checks pass, including the three Fixture 2b assertions.
- **Verified load-bearing, before and after**: checked out the pre-fix `discover.sh`/`classify.sh` (the parent commit) with the current test file, and confirmed all three new assertions correctly FAIL (52/55, 3 failures) — reproducing the exact #1092 leak. Restored the fix and confirmed 55/55 green again.
- `shellcheck --severity=warning` on all three edited files — no new warnings versus the pre-fix baseline (the four pre-existing SC2038/SC2221/SC2222 warnings in `discover.sh`/`classify.sh` are unrelated to this change and were present before it; the test file itself is shellcheck-clean).

```
$ bash .claude/skills/dfd/tests/smoke.sh
...
Fixture 2b: workspace/ and .claude/worktrees/ pruned from discovery (#1092)
================================================================
  PASS: Fixture 2b: workspace/other-project pruned from discovery
  PASS: Fixture 2b: .claude/worktrees pruned from discovery
  PASS: Fixture 2b: classify.sh also prunes workspace/ and .claude/worktrees/
...
Total: 55   Passed: 55   Failed: 0
OK: all /dfd smoke checks passed.
```

```
$ git show ac9053b:.claude/skills/dfd/discover.sh > .claude/skills/dfd/discover.sh   # pre-fix
$ git show ac9053b:.claude/skills/dfd/classify.sh > .claude/skills/dfd/classify.sh   # pre-fix
$ bash .claude/skills/dfd/tests/smoke.sh | tail -8
Fixture 2b: workspace/ and .claude/worktrees/ pruned from discovery (#1092)
================================================================
  FAIL: Fixture 2b: workspace/other-project evidence (path) leaked into discovery output
  FAIL: Fixture 2b: .claude/worktrees evidence (path) leaked into discovery output
  FAIL: Fixture 2b: classify.sh leaked workspace/worktrees evidence (path)
...
Total: 55   Passed: 52   Failed: 3
$ git checkout HEAD -- .claude/skills/dfd/discover.sh .claude/skills/dfd/classify.sh   # restored
```

### Known follow-up not addressed in this PR

Rex's review also flagged a nit on the production fix itself: `EXTRA_PRUNE_PATHS` (in `discover.sh`/`classify.sh`) is word-split via a plain `for p in $EXTRA_PRUNE_PATHS`, which is fragile if a resolved workspace/worktrees path contains a space. Per the review instruction this round was test-only ("do NOT touch discover.sh/classify.sh"), so it's left as-is here and called out for a follow-up rather than folded into this PR silently.

Refs #1092

---

## Glossary

| Term | Definition |
|------|------------|
| reachability-bounded | Discovery follows only what the target actually connects to, rather than everything under its directory — the property this bug broke. |
| ops fork | The apexyard fork that governs a portfolio; its `workspace/` dir holds clones of the projects it governs. |
| `scoped_grep` | The `discover.sh`/`classify.sh` helper that greps the target tree while pruning vendored/irrelevant directories via `find ... -prune`. |
| `portfolio_workspace_dir()` | A resolver in `_lib-portfolio-paths.sh` that returns the absolute path to the ops fork's `workspace/` directory, honouring any adopter override. |
| load-bearing (assertion) | A test assertion that actually fails when the bug it's meant to catch is present — as opposed to one that passes regardless of whether the fix exists. |

### DIFF
--- a/.claude/skills/dfd/classify.sh
+++ b/.claude/skills/dfd/classify.sh
@@ -37,9 +37,34 @@ TARGET=$(cd "$TARGET" && pwd -P)
 
 SKIP_DIRS="node_modules vendor .venv venv target dist build coverage .next .nuxt .turbo .cache __pycache__ .git .svn"
 
+# ---------------------------------------------------------------------------
+# Portfolio-workspace exclusion (me2resh/apexyard#1092) — same rationale as
+# discover.sh: when TARGET is an apexyard ops fork, `workspace/` and
+# `.claude/worktrees/` hold full clones/checkouts of OTHER projects and must
+# not be attributed to the target's own data-classification scan.
+# ---------------------------------------------------------------------------
+EXTRA_PRUNE_PATHS=""
+_LIB_PORTFOLIO_PATHS="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)/../../hooks/_lib-portfolio-paths.sh"
+
+if { [ -f "$TARGET/.apexyard-fork" ] || { [ -f "$TARGET/onboarding.yaml" ] && [ -f "$TARGET/apexyard.projects.yaml" ]; }; } \
+  && [ -f "$_LIB_PORTFOLIO_PATHS" ]; then
+  # shellcheck source=/dev/null
+  . "$_LIB_PORTFOLIO_PATHS" 2>/dev/null || true
+  if command -v portfolio_workspace_dir >/dev/null 2>&1; then
+    WS_DIR=$(cd "$TARGET" && portfolio_workspace_dir 2>/dev/null) || WS_DIR=""
+    if [ -n "$WS_DIR" ] && [ -d "$WS_DIR" ]; then
+      EXTRA_PRUNE_PATHS="$WS_DIR"
+    fi
+  fi
+  WT_DIR="$TARGET/.claude/worktrees"
+  if [ -d "$WT_DIR" ]; then
+    EXTRA_PRUNE_PATHS="$EXTRA_PRUNE_PATHS $WT_DIR"
+  fi
+fi
+
 scoped_grep() {
   local pattern="$1"; shift
-  local prune_args="" first=1 d
+  local prune_args="" first=1 d p
   for d in $SKIP_DIRS; do
     if [ $first -eq 1 ]; then
       prune_args="-name $d"
@@ -47,6 +72,9 @@ scoped_grep() {
     else
       prune_args="$prune_args -o -name $d"
     fi
+  done
+  for p in $EXTRA_PRUNE_PATHS; do
+    prune_args="$prune_args -o -path $p"
   done
   # shellcheck disable=SC2086
   find "$TARGET" \( $prune_args \) -prune -o -type f -print 2>/dev/null \
--- a/.claude/skills/dfd/discover.sh
+++ b/.claude/skills/dfd/discover.sh
@@ -35,11 +35,44 @@ TARGET=$(cd "$TARGET" && pwd -P)
 
 SKIP_DIRS="node_modules vendor .venv venv target dist build coverage .next .nuxt .turbo .cache __pycache__ .git .svn .idea .vscode"
 
-# scoped_grep <pattern>: greps the target dir for <pattern>, pruning vendored dirs.
+# ---------------------------------------------------------------------------
+# Portfolio-workspace exclusion (me2resh/apexyard#1092).
+#
+# When TARGET is an apexyard ops fork itself, `workspace/` holds full clones
+# of OTHER registered projects, and `.claude/worktrees/` holds agent
+# worktrees — both are full checkouts of unrelated (or duplicate) code.
+# Rule 4 of the skill says discovery is "reachability-bounded"; neither of
+# these directories is reachable from the fork's own code, so they must be
+# pruned from the walk. Resolve the CONCRETE paths (rather than an any-depth
+# basename match) so a legitimately-named `workspace/` dir in a non-fork
+# target is never over-pruned.
+# ---------------------------------------------------------------------------
+EXTRA_PRUNE_PATHS=""
+_LIB_PORTFOLIO_PATHS="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)/../../hooks/_lib-portfolio-paths.sh"
+
+if { [ -f "$TARGET/.apexyard-fork" ] || { [ -f "$TARGET/onboarding.yaml" ] && [ -f "$TARGET/apexyard.projects.yaml" ]; }; } \
+  && [ -f "$_LIB_PORTFOLIO_PATHS" ]; then
+  # shellcheck source=/dev/null
+  . "$_LIB_PORTFOLIO_PATHS" 2>/dev/null || true
+  if command -v portfolio_workspace_dir >/dev/null 2>&1; then
+    WS_DIR=$(cd "$TARGET" && portfolio_workspace_dir 2>/dev/null) || WS_DIR=""
+    if [ -n "$WS_DIR" ] && [ -d "$WS_DIR" ]; then
+      EXTRA_PRUNE_PATHS="$WS_DIR"
+    fi
+  fi
+  WT_DIR="$TARGET/.claude/worktrees"
+  if [ -d "$WT_DIR" ]; then
+    EXTRA_PRUNE_PATHS="$EXTRA_PRUNE_PATHS $WT_DIR"
+  fi
+fi
+
+# scoped_grep <pattern>: greps the target dir for <pattern>, pruning vendored
+# dirs (by basename) and, on an ops-fork target, the concrete workspace/ and
+# .claude/worktrees/ paths resolved above (by exact path).
 # Outputs `file:line:match` (grep -nH format), capped at 500 lines.
 scoped_grep() {
   local pattern="$1"; shift
-  local prune_args="" first=1 d
+  local prune_args="" first=1 d p
   for d in $SKIP_DIRS; do
     if [ $first -eq 1 ]; then
       prune_args="-name $d"
@@ -47,6 +80,9 @@ scoped_grep() {
     else
       prune_args="$prune_args -o -name $d"
     fi
+  done
+  for p in $EXTRA_PRUNE_PATHS; do
+    prune_args="$prune_args -o -path $p"
   done
   # shellcheck disable=SC2086
   find "$TARGET" \( $prune_args \) -prune -o -type f -print 2>/dev/null \
--- a/.claude/skills/dfd/tests/smoke.sh
+++ b/.claude/skills/dfd/tests/smoke.sh
@@ -279,6 +279,77 @@ else
 fi
 
 # ---------------------------------------------------------------------------
+# Fixture 2b: workspace/ and .claude/worktrees/ pruned from discovery
+# (me2resh/apexyard#1092)
+#
+# Reuses the ops-fork sandbox ($SB) built for Fixture 2 — it already has
+# onboarding.yaml + apexyard.projects.yaml at its root, which is exactly
+# what discover.sh's ops-fork detection looks for. Plants a distinctive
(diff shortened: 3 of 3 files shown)


atlas-apex (approved):

me2resh (commented):
## Code Review: PR #1107

**Commit**: `1fb98e5bbbe277d4b6cef0a686493b21ed9fe0f5`

### Summary
Stops `/dfd` discovery from attributing other projects' dependencies to the target when the target is an apexyard ops fork, by pruning the concrete `workspace/` and `.claude/worktrees/` paths from `discover.sh` and `classify.sh`'s `scoped_grep` walk — gated on ops-fork detection so non-fork targets are untouched. Adds smoke Fixture 2b as a regression guard.

I reviewed this against a reconstructed copy of the PR head (SHA above) and ran the smoke suite both with the fix and against the unfixed scripts to test whether the new fixture is actually load-bearing.

### Checklist Results
- Testing:                   **Fail** — 2 of the 3 new assertions are non-load-bearing (see below)
- PR Description & Glossary: Pass (narrative Summary + Glossary present)
- Technical Decisions (AgDR):N/A (bug fix, no new dependency/architecture)

### What I verified (production code is correct)
The fix itself is sound — I confirmed it empirically:

- **workspace/ pruning works.** Against an unfixed `discover.sh`, a Stripe import under `workspace/other-project/...` leaks as `evidence: "./workspace/other-project/...:1"`; with the fix it is pruned. This is exactly #1092's "Observed" scenario.
- **.claude/worktrees/ pruning works.** Isolating a Stripe import to `.claude/worktrees/wt1/only.ts`, the unfixed script emits `evidence: "./.claude/worktrees/wt1/only.ts:1"`; the fix prunes it.
- **No regression on non-fork targets.** The ops-fork guard (`.apexyard-fork` OR `onboarding.yaml`+`apexyard.projects.yaml`) leaves `EXTRA_PRUNE_PATHS` empty for a plain project, so the `for p in $EXTRA_PRUNE_PATHS` loop is a no-op and `scoped_grep` is byte-identical to before. Fixture 1 (single-service repo) still passes.
- **shellcheck:** the SC2038/SC2221/SC2222 warnings are all pre-existing; the fix introduces none. Matches the PR body's claim.

One thing worth recording for the author: the workspace prune depends on `portfolio_workspace_dir()` returning an **absolute** path, which in turn depends on `_portfolio_root()` resolving via `git rev-parse`. A real ops fork is always a git checkout, so this is fine in practice — but note that if the target isn't a git repo the resolver returns a relative `./workspace` and the `-path` prune silently no-ops. (The smoke sandbox `git init`s for exactly this reason — good.)

### Issues Found

**1. (Blocking) The worktree assertions in Fixture 2b are vacuous — they pass with OR without the fix.**

`discover.sh`/`classify.sh` emit evidence as `file:line` **paths**, never file **content**. But two of the three new assertions grep for a **content marker**:

- `grep -q "WORKTREE_LEAK_MARKER_9f3c" "$OUT2B"` (the `.claude/worktrees/ pruned from discovery` check) — that string lives in a `//` comment inside `marker.js`, and discovery never prints file content, so the grep can never match regardless of pruning.
- The classify assertion `grep -q "other-project\|WORKTREE_LEAK_MARKER_9f3c"` — only the `other-project` (workspace) alternative ever fires; the `WORKTREE_LEAK_MARKER_9f3c` alternative is dead.

I proved this by running the PR's own `smoke.sh` against the **unfixed** `discover.sh`/`classify.sh`:

```
  FAIL: Fixture 2b: workspace/other-project evidence leaked into discovery output
  PASS: Fixture 2b: .claude/worktrees/ pruned from discovery      <-- passes UNFIXED
  FAIL: Fixture 2b: classify.sh leaked workspace/worktrees evidence
```

So the only load-bearing new assertion is the **workspace** one. The worktree pruning — which the fix genuinely implements and which the PR body lists as a first-class outcome — has **no** regression protection: if it silently broke in a future refactor, CI would stay green. This also makes the PR body's "Regression fixture proves the leak is gone" / "stashing the two fix commits reproduces the pre-fix leak" claim only half true: it reproduces for workspace, not for worktrees.

**Required change (test-only — do not touch `discover.sh`/`classify.sh`):** assert on the emitted **path**, not the content marker. e.g.

```bash
# discover.sh worktree assertion
if grep -q "\.claude/worktrees" "$OUT2B"; then   # emitted path, fails without the fix
  ... FAIL ...
```

To make the fixture self-validating, place a marker whose Stripe/SDK hit is **only** reachable via `.claude/worktrees/` (no sibling `workspace/` hit for the same service), since discovery emits just the first evidence per service — otherwise a workspace hit masks a broken worktree prune. I confirmed a worktree-only sandbox does leak `./.claude/worktrees/...` when unfixed, so a path-based assertion will be genuinely load-bearing.

- **nit:** `EXTRA_PRUNE_PATHS` is space-joined and word-split (`for p in $EXTRA_PRUNE_PATHS`), so an ops-fork or workspace path containing a space would break the prune. Rare, and the pre-existing `SKIP_DIRS` has the same limitation — fine to leave, but a `-path` per array element would be more robust if you ever revisit.
- **nit:** the Testing section states "all 58 checks pass," but a faithful run of the PR's `smoke.sh` reports `Total: 55`. Please double-check the count in the PR body (the three Fixture 2b checks do run and pass; the baseline number looks off).

### Verdict
**CHANGES REQUESTED**

The production fix is correct and I verified it end-to-end — the ask is narrow: make the worktree assertions actually fail without the fix (grep the emitted `.claude/worktrees` path instead of a content marker that discovery never prints), and correct the check-count in the PR body. Once the fixture is load-bearing for the worktree case, this is an approve.

---
Reviewed commit: `1fb98e5bbbe277d4b6cef0a686493b21ed9fe0f5`

me2resh (commented):
## Code Review (re-review): PR #1107

**Commit**: `d1e03be411789232a0cb5d66fd5bfd45ceed6cf6`

### Summary

Re-review after the fix commit `d1e03be` addressing my prior CHANGES REQUESTED (2 of 3 new Fixture 2b assertions were non-load-bearing). This round is test-only.

### Prior finding — RESOLVED ✅

My earlier finding was that the worktree assertion checked a content-only marker that `discover.sh`/`classify.sh` never emit (evidence is always `path:line`, never file content), and the classify assertion was vacuous because the fixtures never triggered a classification pathway that runs through `scoped_grep`. Both are now fixed:

1. **Path-based assertions.** All three Fixture 2b assertions now grep the emitted evidence PATH — `workspace/other-project` and `.claude/worktrees` — which is exactly what the fix's pruning controls. No more content markers.
2. **Distinct scoped vendor imports.** Fixtures switched to `@sendgrid/mail` (workspace) and `@sentry/node` (worktree). I verified the collision risk the author describes is real: the sandbox's copy of `_lib-multi-repo-trace.sh` (copied in by Fixture 2) literally contains `api.stripe.com` / `sendgrid.com` / `api.twilio.com` / `sentry.io` hostname literals. Bare-word vendor patterns (`stripe`, `twilio`) match those and get first-evidence-deduped against that legitimate file — masking the fixture and defeating the assertion even with a correct path check. The scoped `@sendgrid/mail` / `@sentry/` patterns do NOT appear in that hostname table, so they match only the fixtures. Two *different* vendors also defeats the per-vendor "first evidence wins" dedup masking one fixture behind the other.
3. **classify.sh exercises a real pathway.** The `@PII` / `@Sensitive` annotations planted in both fixtures route through the annotation pathway, which is the one classify.sh pathway that actually runs through `scoped_grep` (and therefore through the fix's `EXTRA_PRUNE_PATHS` pruning). The env-var and schema pathways use unguarded `find -maxdepth` outside `scoped_grep` — a separate pre-existing gap, correctly called out-of-scope.

### Load-bearing — verified empirically

I reproduced the before/after in an isolated worktree (main tree untouched):

- **Post-fix (PR head d1e03be):** `Total: 55  Passed: 55  Failed: 0`.
- **Pre-fix (production `discover.sh`/`classify.sh` reverted to the `ac9053b` baseline, new test file kept):** `Total: 55  Passed: 52  Failed: 3` — and the three failures are *exactly* the Fixture 2b workspace-leak, worktrees-leak, and classify-leak assertions.

This matches the author's reported 52/55 → 55/55 precisely. All three assertions genuinely fail against the unfixed scripts and pass with the fix.

### Production scripts unchanged

Confirmed via `compare 1fb98e5...d1e03be`: the new commit touches only `.claude/skills/dfd/tests/smoke.sh` (+53/-23). `discover.sh` and `classify.sh` are byte-identical to the previously-reviewed parent commit — no re-review of the production fix needed.

### Checklist

- ✅ Code Quality: Pass
- ✅ Testing: Pass — the whole point of this round; assertions now load-bearing
- ✅ Security: Pass — fixture `SENDGRID_API_KEY` / `SENTRY_DSN` are `process.env` refs in test fixtures, no real secrets
- ✅ Performance: Pass
- ✅ PR Description & Glossary: Pass — glossary present; count corrected to 55; the `EXTRA_PRUNE_PATHS` word-split nit is flagged as a known follow-up (correctly deferred — this round was test-only per instruction)
- ✅ Technical Decisions (AgDR): N/A — test-only change; production fix already reviewed
- ✅ Adopter Handbooks: N/A — shell-only diff; the blocking `migration-safety.md` handbook has no trigger (zero migration files); advisory handbooks show no violations
- ⚠ Fallow (JS/TS): N/A — diff is `.sh`, not JS/TS

### Notes (non-blocking)

- The `EXTRA_PRUNE_PATHS` word-split (`for p in $EXTRA_PRUNE_PATHS`) remains fragile if a resolved workspace/worktrees path ever contains a space. Correctly left for a follow-up since this round was scoped test-only. Worth a small follow-up PR (a `while read` over a newline-delimited list, or a bash array).

### Verdict

**APPROVE** — my prior finding is fully resolved; the three assertions are now genuinely load-bearing, verified empirically (52/55 pre-fix → 55/55 post-fix).

---
📌 Reviewed commit: `d1e03be411789232a0cb5d66fd5bfd45ceed6cf6`

Issue #1099, opened 2026-07-31
[Chore] Pin the fail-closed behaviour of the two untested merge gates with regression tests


atlas-apex:
## Driver

#1098 applied an identical fail-closed fix to three merge gates — `block-unreviewed-merge.sh`, `require-design-review-for-ui.sh`, `require-architecture-review.sh` — but shipped a regression test for only the first. The other two were changed on the strength of "the fallback was byte-identical in all three."

Both reviewers flagged the asymmetry. The security review then went further and **built the missing case**, running it against both the parent commit and the PR head, with the setup most favourable to a bypass (approval marker written to match the local HEAD, `gh pr diff` answering so the scope check passes, only `headRefOid` failing):

| Gate | Parent commit | After #1098 |
|---|---|---|
| `require-architecture-review.sh` | `rc=0` — merge **allowed** on a local-HEAD-only marker | `rc=2` blocked |
| `require-design-review-for-ui.sh` | `rc=0` — merge **allowed** on a local-HEAD-only marker | `rc=2` blocked |

So the bypass was real and independently reachable in both untested gates, and #1098 closes it in both. **Current correctness is established.** What is missing is anything that *pins* it.

## Scope

Land the two regression cases so the property survives the next refactor.

The security reviewer's position, which I agree with: untested-but-identical is not acceptable as a **standing** state on `.claude/hooks/**`. The risk is regression, not present correctness — nothing currently stops a future edit to either gate silently restoring the fallback, and these are the files the merge gates run on.

The work is close to copy-paste. `test_require_architecture_review.sh` already ships a `make_sandbox()` with a writable `bin/gh` stub (around lines 90–132); the case needs a variant where `headRefOid` fails while `headRefName` and `headRepository` still answer, so exactly one thing breaks.

**The test must be discriminating by construction**, the same way #1098's is: write the approval marker to **match the local HEAD**, so the pre-change code *allows* the merge and the post-change code blocks it. A test using a mismatching local HEAD blocks in both directions and proves nothing — that trap has caught three separate reviewers in this repo already.

## Acceptance Criteria

- [ ] `test_require_architecture_review.sh` has a case: forge HEAD unresolvable + marker matching local HEAD → `rc=2`
- [ ] `test_require_design_review_for_ui.sh` has the equivalent case
- [ ] Each is proven discriminating — fails against the pre-#1098 hook, passes after — and the before/after result is recorded, not just the pass count
[...]

PR #1108, opened 2026-08-01
test(#1099): pin fail-closed behaviour of the architecture + design merge gates

me2resh:
## Summary

- **Pins the fail-closed fix from #1098 with a regression test on the two gates that shipped without one.** #1098 removed an identical local-HEAD fallback from three merge gates (`block-unreviewed-merge.sh`, `require-architecture-review.sh`, `require-design-review-for-ui.sh`) but only added a regression test for the first — on the theory that the fallback was byte-identical in all three, so one test covered the risk. Both reviewers on #1098 flagged that as an untested assumption, and the security review went further and built the missing case by hand, confirming the bypass was real and independently reachable in both untested gates. This PR is the "close the loop" work: the same case, landed as a permanent test, so a future refactor can't quietly restore the fallback without a red suite.
- **Each case is discriminating by construction, not just green.** The mock `gh` answers `pr diff` (so a design-artifact / UI file is found) and `pr view ... headRepository`, but fails `pr view ... headRefOid` — the one call `resolve_pr_head` makes to ask the forge for the PR's real HEAD. The approval marker is then written to match the *sandbox's local HEAD*, which is exactly the value the old fallback would have substituted for the unresolvable forge HEAD. That setup is the one most favourable to a bypass: under the removed fallback the SHA comparison succeeds and the merge is **allowed**; under fail-closed it's **blocked** regardless of what local HEAD says. A marker at a mismatching local HEAD would have blocked both before and after and proven nothing — the exact trap called out in #1099's own issue body, and the same one #1098's original test avoided.
- **A control case per gate** (forge healthy, marker at the forge HEAD → still allowed) guards against the fail-closed change over-blocking the ordinary, healthy-forge path — the mirror failure two other predicates in this area have already had to be redesigned for.
- **Test-only change.** No hook logic, no `.claude/settings.json` wiring, nothing behavioural — only new cases in the two existing test files.

## Testing

Ran both edited suites directly:

```
bash .claude/hooks/tests/test_require_architecture_review.sh   # PASS: 24  FAIL: 0
bash .claude/hooks/tests/test_require_design_review_for_ui.sh  # PASS: 24  FAIL: 0
```

**Before/after proof (not just the pass count) — required by the ticket's own acceptance criteria.** I temporarily swapped in the two hooks as they existed one commit before #1098 (`59c3cc6`, the parent of `1712f79`) and re-ran the new section E) cases only, then restored the current hooks and re-ran the full suites to confirm they're back to matching `dev`.

| Gate | Pre-#1098 hook (`59c3cc6`) | Current hook (post-#1098) |
|---|---|---|
| `require-architecture-review.sh` | `FAIL want '2', got '0'` — merge **allowed** | `PASS` — `rc=2`, blocked |
| `require-design-review-for-ui.sh` | `FAIL want '2', got '0'` — merge **allowed** | `PASS` — `rc=2`, blocked |
| Control case (both gates) | `PASS` (healthy forge already allowed) | `PASS` (unchanged) |

That's the discriminating property: the same fixture that fails against the old hook and passes against the new one, with the control case passing in both directions so the change doesn't newly over-block the healthy path.

`shellcheck --severity=warning` on both edited files: clean, exit 0.

Refs #1099

---

## Glossary

| Term | Definition |
|------|------------|
| discriminating test | A test that fails against the pre-change code and passes against the post-change code. A test that's green in both directions proves nothing about the specific behaviour it claims to cover — this is why the marker in the new cases is deliberately written to match the *local* HEAD rather than a mismatching one. |
| fail closed | A control that blocks rather than permits when it can't evaluate its own precondition (here: the forge HEAD is unresolvable). The safe failure direction for a merge gate — the alternative, failing open, silently stops enforcing anything. |
| forge-reported HEAD | The PR's real HEAD commit SHA as GitHub (or another forge) reports it via `gh pr view ... --json headRefOid`, as opposed to `git rev-parse HEAD` on the local working tree — a value only the forge, not a local agent, can author. |
| over-blocking | The mirror failure of fail-closed: a gate so aggressive it also blocks legitimate, healthy-path merges. The control case in this PR exists specifically to catch that regression. |
| review marker | A local file under `.claude/session/reviews/` (e.g. `<pr>-architecture.approved`) recording that a reviewer signed off on a PR at a specific commit SHA. The merge gates compare the marker's recorded SHA against the forge-reported HEAD before allowing a merge. |

### DIFF
--- a/.claude/hooks/tests/test_require_architecture_review.sh
+++ b/.claude/hooks/tests/test_require_architecture_review.sh
@@ -293,6 +293,77 @@ assert_eq "#687 wrong-qualifier marker still blocks" "2" "$code"
 rm -rf "$sb" "$pf"
 
 echo ""
+echo "E) #1091/#1099 sibling: forge HEAD unresolvable -> the gate must FAIL CLOSED"
+#
+# require-architecture-review.sh carried the SAME local-HEAD fallback that
+# block-unreviewed-merge.sh had (see #1091, fixed for all three gates by
+# #1098, which shipped a regression test only for block-unreviewed-merge.sh).
+# This is the missing discriminating case for THIS gate (#1099), built on the
+# exact construction from test_block_unreviewed_merge.sh's #1091 case
+# (~line 832): the mock gh answers `pr diff` (a design artifact is found)
+# and `pr view ... headRepository`, but FAILS `pr view ... headRefOid` — the
+# one call resolve_pr_head makes. `headRefName` is also wired to answer even
+# though this hook never queries it, mirroring the sibling gate's mock so the
+# "forge reachable but this ONE field is missing" shape is explicit rather
+# than "the whole gh binary is gone".
+#
+# DISCRIMINATING BY CONSTRUCTION: the marker is written to match THIS
+# sandbox's LOCAL HEAD (`git rev-parse HEAD`) — the value the pre-#1098
+# fallback would have substituted for the unresolvable forge HEAD. Under the
+# removed fallback that makes the SHA comparison succeed and the merge
+# ALLOWED (rc=0). Under fail-closed it is BLOCKED (rc=2) regardless of what
+# the local HEAD says. A marker at a MISMATCHING local HEAD would block both
+# before and after, proving nothing — the exact trap #1099's own issue body
+# calls out.
+
+install_mock_gh_headrefoid_fails() {
+  local sb="$1" diff_files="$2" repo="${3:-o/r}"
+  mkdir -p "$sb/bin"
+  cat > "$sb/bin/gh" <<EOF
+#!/bin/bash
+args="\$*"
+case "\$args" in
+  *"pr diff"*"--name-only"*)
+    printf '%s\n' $diff_files
+    ;;
+  *"pr view"*headRefOid*)
+    exit 1
+    ;;
+  *"pr view"*headRefName*)
+    printf '%s\n' "feature/GH-99-test"
+    ;;
+  *"pr view"*headRepository*)
+    printf '%s\n' "$repo"
+    ;;
+  *) exit 0 ;;
+esac
+EOF
+  chmod +x "$sb/bin/gh"
+}
+
+sb=$(make_sandbox)
+install_mock_gh_headrefoid_fails "$sb" '"projects/foo/docs/technical-design-x.md"'
+# The sandbox's local HEAD — the value the OLD fallback would have used.
+local_head=$(cd "$sb" && git rev-parse HEAD 2>/dev/null)
+# Marker written to MATCH that local HEAD: the setup most favourable to a
+# bypass, where every comparison passes under the old code.
+printf '%s\n' "$local_head" > "$(review_marker_path "o/r" 77 architecture "$sb")"
+code=$(run_gate "$sb" "gh pr merge 77 --repo o/r --squash")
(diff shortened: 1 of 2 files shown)


me2resh (commented):
## Code Review: PR #1108

**Commit**: `3d161858ca78f8ce7b351c872bd5eefcf0a23aec`

### Summary
Adds the missing regression cases (#1099) that pin the fail-closed fix from #1098 on the two merge gates that shipped without one: `require-architecture-review.sh` and `require-design-review-for-ui.sh`. Each gate gets a discriminating "forge HEAD unresolvable + marker at local HEAD → BLOCK (rc=2)" case plus a healthy-forge control case. Test-only — no gate logic, no `.claude/settings.json`.

### Checklist Results
- Security:                  Pass (trust-chain-adjacent; correctly test-only, no gate logic touched)
- PR Description & Glossary: Pass (narrative Summary, 5-term Glossary)
- Technical Decisions (AgDR):N/A (test-only, no material decision)

### Independent verification (I reproduced the discriminating proof, not just the pass count)
I fetched both test files at HEAD `3d16185`, ran them against a mirror of the current hooks, then swapped in the pre-#1098 hooks at `59c3cc6` (which still carried `CURRENT_SHA=$(git rev-parse HEAD)`):

| Gate | Pre-#1098 (`59c3cc6`) | Current (post-#1098) |
|---|---|---|
| `require-architecture-review.sh` fail-closed case | `FAIL want '2', got '0'` — merge **allowed** | `PASS` rc=2, blocked |
| `require-design-review-for-ui.sh` fail-closed case | `FAIL want '2', got '0'` — merge **allowed** | `PASS` rc=2, blocked |
| Control case (both gates) | `PASS` | `PASS` |

This matches the author's claimed table exactly. Both full suites: 24 PASS / 0 FAIL on current hooks. shellcheck `--severity=warning`: clean (exit 0) on both files.

The construction is sound and, notably, self-validating: because the fail-closed case FAILS (rc=0) against the old hook rather than passing, we know the marker at the qualified path is actually being found and compared against local HEAD — a mis-resolved marker would have produced a "no marker" block (rc=2) in *both* directions and proven nothing. So the block on current hooks is provably the fail-closed branch, not an incidental precondition.

### Issues Found
None.

- `install_mock_gh_headrefoid_fails` is defined inline inside the E) section rather than hoisted next to the other helpers at the top of each file. This is consistent with the file's existing convention (`install_mock_gh_splitportfolio` is defined inline mid-file too), so it's a neutral observation, not a request.
- The mock wires `headRefName` even though neither hook queries it. The comment explains this is deliberate — mirroring the sibling gate's mock so the "one field missing" shape is explicit rather than "gh binary gone." Good call; leaving a note here only so a future reader doesn't prune it as dead.

### Verdict
**APPROVED**

Clean, surgical, exactly satisfies every acceptance criterion in #1099 (discriminating case per gate, control case per gate, before/after recorded, test-only). The fail-closed property on these two trust-chain gates is now pinned against silent regression.

---
📌 Reviewed commit: `3d161858ca78f8ce7b351c872bd5eefcf0a23aec`

atlas-apex (approved):

mostiwheelietravel (commented):
[Redacted: this review was posted to the wrong repository by an automated code reviewer and is unrelated to this PR. Content removed.]

Issue #1105, opened 2026-07-31
[Bug] /update migration chain is unwalkable — no migration script for any of the 21 releases since v2.0.2


sudanese:
## Affected

`.claude/skills/update/SKILL.md` § 8b (migration chain), `.claude/migrations/`, and `.claude/skills/release/SKILL.md`.

## Given / When / Then

**Given** an ops fork several releases behind, whose `.claude/framework-version` anchor reads any version at or after `v2.0.2`

**When** it runs `/update` against current `main`

**Then** step 8b builds the chain, finds no migration script for the first hop, and — per the skill's own edge-case table — skips the chain entirely and advances the anchor anyway. The adopter gets a files-only sync.

**Expected** either a real migration per release, or an explicit no-op placeholder per release, so the chain actually walks.

## The gap, as measured

`.claude/migrations/` at v5.3.0 contains exactly:

```
README.md
v1.2.0-to-v1.3.0.sh
v1.3.0-to-v1.4.0.sh
```

There are **21 release tags reachable from `main`** — `v2.0.2 v2.1.0 v2.2.0 v2.3.0 v3.0.0 v3.1.0 v3.1.1 v3.1.2 v3.1.3 v3.1.4 v3.2.0 v3.3.0 v4.0.0 v4.1.0 v4.2.0 v4.3.0 v4.4.0 v5.0.0 v5.1.0 v5.2.0 v5.3.0` — and **none of them has a migration script**.

Separately, **`v1.4.0` does not exist as a git tag at all**. So the one script naming it (`v1.3.0-to-v1.4.0.sh`) describes a hop that cannot be reached from any tagged release. Both shipped scripts are effectively unreachable.

## Why this reads as a bug rather than a deliberate choice

Three places in the shipped framework state the opposite:

1. **`CHANGELOG.md` v5.0.0** lists `/update walks intermediate-release migration chain (#286)` under **Breaking**.
2. **`update/SKILL.md`** § "Why per-pair scripts, not one monolithic migration" states that *"every release that ships either adds a real migration OR a no-op placeholder"*, and that skipping a release *"makes the chain refuse to walk past it"*.
3. The same section says the cost is *"discipline at release-cut time"*, and that **the `/release` skill's PR template includes a "did we add a migration script for this release?" checkbox to catch the omission.**

[...]

PR #1109, opened 2026-08-01
fix(#1105): backfill walkable migration-chain placeholders and add the release-time migration check

me2resh:
## Summary

- **Backfilled the current-major migration chain** — added no-op placeholder scripts for `v4.4.0-to-v5.0.0`, `v5.0.0-to-v5.1.0`, `v5.1.0-to-v5.2.0`, and `v5.2.0-to-v5.3.0`. Before this, `.claude/migrations/` had scripts for exactly zero of the 21 release hops since v2.0.2, so any adopter several releases behind got a silent files-only sync instead of the per-release migration chain `/update` step 8b is supposed to walk — see #1105 for the full "measured gap" writeup. **Scope note:** the placeholder range covers `v4.4.0` through `v5.3.0` (current major + the boundary hop into it) per the ticket's floor — it does not backfill the older gap between `v1.3.0` and `v4.4.0` (roughly 15 releases). Flagged as a known residual gap in `docs/upgrading.md` rather than bundled in here, since backfilling it accurately means checking each of those releases' actual changes rather than assuming no-op.
- **Removed the orphaned `v1.3.0-to-v1.4.0.sh`** — `v1.4.0` was never tagged (v2.0.0 shipped in its place instead), so the script named a hop the chain could never reach. Worse, `migration_known_versions` reads pair filenames literally to build `/update`'s "unknown anchor" interactive menu, so the orphan could offer a fake `v1.4.0` as a choice. I chose **remove over rename** deliberately: renaming it to `v1.3.0-to-v2.0.0.sh` would silently assert that a MAJOR version bump needed zero adopter migration, which isn't a claim I can make without checking what v2.0.0 actually shipped. Removing it turns a misleading "fake chain link" into an honest "known gap" (the missing-link edge case `/update` already handles correctly). Rationale is written up in `.claude/migrations/README.md`'s new "Orphaned hop scripts" section, both as an explanation and as guidance for the next time this happens.
- **Added the release-time migration check** that `docs/upgrading.md` and `update/SKILL.md` § 8b have documented since AgDR-0032 but that never actually existed — 21 consecutive releases shipped with no migration script and nothing flagged it. `/release`'s new step 3 sub-check warns (doesn't block) if `.claude/migrations/<prev>-to-<new>.sh` is missing for the release being cut, and the PR body template gained a matching checkbox. Kept deliberately **advisory, not a hard gate** per the ticket's scope — whether a release needs a migration is a judgement call, and turning this into a blocking pre-push/CI gate is a bigger design decision I've noted as a follow-up in the skill's Rule 13 rather than building here.
- **Fixed `test_update_chain.sh`'s stale dependency** on the now-deleted orphan file (it only asserted the file existed, never executed it) — swapped the presence check to `v5.2.0-to-v5.3.0.sh`, one of the new real placeholder scripts.

## Testing

- `bash .claude/hooks/tests/test_update_chain.sh` — 11/11 passed (chain-building, unknown-anchor, anchor round-trip, skip-migrations, idempotent real-script execution, missing-link refusal, backwards refusal, from==to, exit-code forwarding)
- `bash .claude/skills/update/tests/test_step_8a_wiring.sh` — 18/18 passed
- `bash .claude/hooks/tests/test_split_portfolio_v2_migration.sh` — 23/25 passed; the 2 failures (`validate post-state`, `post-v2 ops_root`) are pre-existing and unrelated to this change — reproduced identically with `git stash` on this same worktree before any of my edits (an ops-root-resolution quirk in this nested-worktree sandbox, not something this PR touches)
- `bash .claude/hooks/tests/test_release_changelog.sh` — all 85 passed (unrelated to this PR, ran as a sanity check since it's the closest sibling to the release-skill edits)
- Manually verified the chain walker with the new scripts in place: `migration_chain v4.4.0 v5.3.0` returns the correct 4-hop ordered list; `migration_chain v4.3.0 v5.3.0` (one hop below the backfilled floor) correctly returns empty (missing link, as expected — that gap is the noted follow-up); `migration_chain v1.3.0 v2.0.0` correctly returns empty now that the orphan is gone (real gap, no longer a fake link)
- `shellcheck --severity=warning` clean on all 4 new scripts and the edited test file
- `npx markdownlint-cli2` clean on the 3 edited markdown files (`.claude/migrations/README.md`, `docs/upgrading.md`, `.claude/skills/release/SKILL.md`)

Refs #1105

---

## Glossary

| Term | Definition |
|------|------------|
| Migration chain | The ordered set of per-release upgrade scripts `/update` walks when a fork is several releases behind, via `.claude/hooks/_lib-migration-chain.sh` |
| No-op placeholder | A migration script that intentionally does nothing (exits 0) — shipped so the chain has an unbroken link for a release that needed no adopter action |
| Orphan (hop script) | A migration script whose `<to>` side names a version that was never tagged — a hop the chain can never actually reach |
| Missing link | A release hop with no script at all — the chain refuses to walk past it and advances the anchor anyway, per the skill's documented edge case |
| Version anchor | `.claude/framework-version` — records which release the fork was last aligned with; the chain's start point |


atlas-apex (approved):

me2resh (commented):
## Code Review: PR #1109

**Commit**: `172c7e6b5b4b33abb244633582396e0cdac5aa52`

### Summary

Backfills the current-major migration chain (`v4.4.0` through `v5.3.0`, four no-op placeholder scripts), removes the orphaned `v1.3.0-to-v1.4.0.sh` (v1.4.0 was never tagged; v2.0.0 shipped in its place), and adds the release-time migration-script check that AgDR-0032 documented but that never existed — the root cause of #1105. Heavy-tier change (release / update machinery), reviewed with that weight.

### Checklist Results

- Architecture and Design: Pass — conforms to the AgDR-0032 chain contract; no new pattern introduced
- Code Quality: Pass
- Testing: Pass — chain walkability independently reproduced (see below)
- Security: Pass — no auth / crypto / secrets / trust-chain surface; scripts are pure no-ops
- Performance: N/A
- PR Description and Glossary: Pass — narrative Summary (what plus why per bullet), Glossary present
- Summary Bullet Narrative: Pass — no label-only bullets
- Technical Decisions (AgDR): Pass — `agdr: not-applicable` escape-hatch correctly used (judgment below)
- Adopter Handbooks: N/A — no handbooks loaded (docs / shell-only diff, no language handbook triggers)
- Fallow Static Analysis (JS/TS): N/A — not a JS/TS diff

### Independent verification performed

I did not rely on the PR's own test claims. Against a sandbox holding the PR's four new scripts plus the surviving `v1.2.0-to-v1.3.0.sh`, sourcing the unmodified real `_lib-migration-chain.sh`:

| Check | Result |
|-------|--------|
| `migration_chain v4.4.0 v5.3.0` | 4-hop chain resolves (`v4.4.0-to-v5.0.0` … `v5.2.0-to-v5.3.0`) — walkable for the current major |
| `migration_chain v4.3.0 v5.3.0` (one hop below floor) | empty — correctly refuses at the missing link |
| `migration_chain v1.3.0 v2.0.0` (orphan removed) | empty — honest gap, no longer a fake link |
| `migration_known_versions` (unknown-anchor menu) | offers only real tags; `v1.4.0` count is 0 (no longer offered) |
| `shellcheck --severity=warning` on all 4 new scripts | clean |

**Placeholder contract match** — all four scripts carry `#!/bin/bash`, `set -u`, the `QUIET` / `info()` convention, `exit 0`, the documented `0/1/2` exit-code block, mode `100755`, and a real rationale comment (documented no-ops, not empty stubs). They match the existing `v1.2.0-to-v1.3.0.sh` shape.

**Orphan removal is clean** — `REAL_V2_V3` in `test_update_chain.sh` was presence-check-only (never executed); Case 5 executes only the surviving `v1.2.0-to-v1.3.0.sh`. The swap to `REAL_NOOP_PLACEHOLDER` (`v5.2.0-to-v5.3.0.sh`) is complete. The remaining `v1.4.0` references in the test are self-contained mktemp fixtures with stub scripts that exercise the walker algorithm independently of the real migrations dir — correctly left untouched. All doc references (README twice, `docs/upgrading.md` table) were updated. No dangling real-file dependency remains.

**Release check** — advisory only (`[ ! -f ]` then an `echo` warning, no `exit`, no gate); PR-body checkbox added; Rule 13 documents the deliberate advisory-not-blocking choice and flags a hard gate as an explicit follow-up. `$PREV_TAG` is a live variable defined at step 3 (line 49) before the check. Matches the ticket's root-cause finding (documented-but-missing checkbox) without over-reaching scope.

### AgDR escape-hatch judgment — JUSTIFIED

The `agdr: not-applicable` marker fires because `.claude/migrations/` matches `require-agdr-for-arch-pr.sh`'s generic migrations glob — a pattern aimed at database / schema migrations. This directory is framework-version upgrade tooling, a different concept whose design decision is already recorded in AgDR-0032. This PR backfills scripts per that already-decided contract and adds advisory docs — it introduces no new dependency, service, data model, security control, or repo-wide pattern. Rail 1 of `.claude/rules/agdr-decisions.md` ("migrations never exempt") refers to DB / schema migrations (data-loss / downtime blast radius), not this directory-name collision, so it is not tripped. The two genuine judgment calls the PR makes — remove-vs-rename the orphan, advisory-vs-blocking the check — are both reversible inside this PR, conservatively chosen, and documented inline. A fresh AgDR here would near-duplicate AgDR-0032 and manufacture exactly the ceremony-churn the rule was rewritten (post-#995 / #997) to prevent. The author used the sanctioned, visible, auditable escape-hatch with a detailed inline rationale — the intended mechanism, not a bypass. Justified.

### Issues Found

None blocking.

- nit (`.claude/skills/release/SKILL.md`, step 3b): the check path `${PREV_TAG}-to-vA.B.C.sh` mixes a live shell variable (`$PREV_TAG`) with the hand-substituted literal placeholder `vA.B.C`. Consistent with the skill's existing idiom (`vA.B.C` is the literal placeholder used throughout), and harmless because the check is advisory — but a reader who runs the block verbatim before substituting `vA.B.C` gets a spurious warning. A one-word inline note that `vA.B.C` is the release placeholder would remove the ambiguity. Optional.
- suggestion: the known residual gap (`v1.3.0` through `v4.4.0`, roughly 15 releases) is well-documented in `docs/upgrading.md`, but consider filing a tracked follow-up issue so the backfill-after-checking-each-release work doesn't get lost in prose. Same for the advisory-vs-blocking release gate flagged in Rule 13.

### Verdict

**APPROVED**

---
Reviewed commit: `172c7e6b5b4b33abb244633582396e0cdac5aa52`

Issue #1097, opened 2026-07-31
[Bug] Pre-push skip marker is matched unanchored — a commit that DESCRIBES it disables all checks, and squash-merge propagates that to dev


atlas-apex:
## Bug Scenario

**Given** a commit whose message *mentions* the pre-push skip marker — for example a commit that documents the escape hatch, a review comment quoting it, or a revert whose body includes the original message

**When** `bin/run-pre-push-checks.sh` runs

**Then** it matches the marker with an unanchored `grep -qF` over the entire commit body, prints `WARN: pre-push checks bypassed by skip marker`, and **exits 0 having run none of markdownlint, shellcheck, or subpacks**

**Expected** the marker only takes effect when it is deliberately applied — a line consisting of the marker, not a mention of it anywhere in prose.

## The code

```sh
# bin/run-pre-push-checks.sh:77-79
SKIP_MARKER='<!-- pre-push: skip -->'
HEAD_MSG=$(cd "$REPO_ROOT" && git log -1 --format='%B' 2>/dev/null)
if echo "$HEAD_MSG" | grep -qF -- "$SKIP_MARKER"; then
```

## Observed live

While preparing #1095, a commit body explained the escape hatch and quoted the marker verbatim. The gate then reported itself bypassed and ran zero checks — on the very PR that was fixing the gate. Discovered by the code review on #1096, not by the author.

```
$ bash bin/run-pre-push-checks.sh
WARN: pre-push checks bypassed by skip marker in HEAD commit message.
      All skipped checks will still run in CI — fix before merging.
$ echo $?
0
```

## Why this is worse than a local annoyance

**It propagates.** Under the release-cut model a PR is squash-merged, and the squash commit inherits the PR body / commit body. A branch whose message merely *discusses* the marker lands that string as the tip commit on `dev`. Every contributor whose next push has that commit at `HEAD` then gets a **silent** skip of all three checks — with no indication anything is different, because the warning scrolls past in a normally-noisy push.

The affected window is narrow (only while that commit is `HEAD`), but the failure is silent and the trigger is *writing documentation about the feature*, which is not an unusual thing to do.

## Severity

**medium** — no security impact and CI still runs the checks, but a gate that disables itself when described is a gate that cannot be safely documented. Any AgDR, handbook entry, or review comment quoting the marker becomes a live bypass if it reaches a commit message.

## Suggested fix

Anchor the match to a marker-only line rather than a substring anywhere in the body:

```sh
[...]

Issue #1066, opened 2026-07-30
[Bug] Branch-name and main-push validators match git commands inside heredoc prose, blocking valid operations


atlas-apex:
## Bug Scenario

**Given** a Bash call that both writes a commit message via heredoc and then runs a git command, where the message *prose* mentions a git command (extremely common in this repo, whose commit messages routinely discuss git behaviour)

**When** the hook extracts its subject from the raw command text

**Then** it matches the **first** occurrence anywhere in the string — including inside the heredoc body — and validates a word from the prose as if it were a branch name, blocking the operation

**Expected** the validators should only consider the actual git command being run, not text that happens to sit inside a heredoc, a quoted string, or a message body

## Repro Steps

Minimal and deterministic:

```bash
cd <apexyard fork>
bash -c '
. .claude/hooks/_lib-extract-push-ref.sh
CMD="cat > /tmp/m.txt <<EOF
The previous commit claimed terminal \`git push\` still runs the checks because
bin/run-pre-push-checks.sh hardcodes them.
EOF
git commit -F /tmp/m.txt
git push upstream fix/GH-1031-untrack-project-config-json"
echo "extracted ref: [$(extract_push_ref "$CMD")]"
'
```

Actual output:

```
extracted ref: [still]
```

Expected: `fix/GH-1031-untrack-project-config-json`.

`validate-branch-name.sh` then rejects with `BLOCKED: Branch 'still' doesn't follow naming convention.` The branch name was valid throughout; the word came from the sentence "`git push` still runs the checks".

Note the near-miss that makes this easy to under-diagnose: a heredoc whose prose does **not** contain a git command extracts the ref correctly. The trigger is specifically prose that *mentions* a git command earlier in the string than the real one.

## Environment

Framework `dev` @ 8822d05. macOS, zsh-invoked Bash tool calls. Not environment-specific — this is string matching over the command payload.

## Severity

P1 — it blocks correct, policy-compliant operations, and the failure message actively misdirects (it names a branch that does not exist and tells you to rename a branch that is already correctly named).

## Mitigation

Write the message to a file in a **separate** tool call, then run the git command on its own:

```bash
# call 1: write the message (no git command in this call)
# call 2: git commit -F <file>
# call 3: git push <remote> <branch>
```

Never inline a heredoc alongside a git command in the same Bash invocation.

## Investigation Notes

**Confirmed from two independent directions in one session**, which is why this is worth a ticket rather than a shrug:

[...]

PR #1111, opened 2026-08-01
fix(#1097): anchor the pre-push skip marker to a whole line

me2resh:
## Summary

- **Anchored the pre-push skip marker to a whole line** — `bin/run-pre-push-checks.sh` and `.claude/hooks/pre-push-gate.sh` both used to `grep -qF` the HEAD commit message for `<!-- pre-push: skip -->` anywhere in the text. That meant a commit that merely *described* the marker (documentation, a review comment quoted verbatim, a revert body) matched too, silently disabling markdownlint, shellcheck, and the subpack smoke test with a WARN that scrolls past in normal push noise. Switched both to `grep -qxF` against the message split one-line-per-line (`printf '%s\n' "$HEAD_MSG" | grep -qxF -- "$SKIP_MARKER"`), so only a line that consists of *exactly* the marker now bypasses. The documented emergency-bypass snippet already emits the marker on its own line, so real bypasses are unaffected.
- **Why it matters** — under the release-cut model this repo uses, a squash-merged PR's body becomes the tip commit on `dev`. A branch whose commit message merely discussed the marker would land that string as `HEAD` on a shared branch, silently skipping all three checks for every contributor whose next push happened to still have that commit at `HEAD`. This was observed live while preparing #1095/#1096 — the check reported itself bypassed on the very PR fixing the gate, caught only by code review, not by the author. Two related unanchored-match defects in the same family: #1066 (branch-name/main-push validators matching prose inside heredocs) and AgDR-0104 (pattern-matching command text can't be made sound in general — this one is a narrow, cheap exception because the thing matched is a *fixed marker on its own line*, not free-form intent).
- **Regression tests** — added two cases to `.claude/hooks/tests/test_pre_push_gate.sh`: a commit whose message *mentions* the marker inside a sentence must NOT bypass (checks still run, still block on failure); a commit with the marker as its own line must still bypass, unchanged. 11/11 cases pass.

## Testing

1. `bash .claude/hooks/tests/test_pre_push_gate.sh` — 11/11 pass, including the two new cases (`marker-mentioned-in-prose-does-not-bypass`, `marker-own-line-still-bypasses`).
2. `bash .claude/hooks/tests/test_subpack_extraction.sh` — all sub-pack extraction invariants pass (unaffected by this change, run as part of the pre-push check set itself).
3. Manual two-direction smoke test directly against `bin/run-pre-push-checks.sh` in a scratch git sandbox:
   - HEAD commit message mentions the marker inline in a sentence → all three checks run (markdownlint/shellcheck/subpacks), no bypass WARN printed, exit reflects the check outcome.
   - HEAD commit message has the marker as its own line → `WARN: pre-push checks bypassed by skip marker...` printed, exit 0, no checks run.
4. `shellcheck --severity=warning` clean on all three touched files (`bin/run-pre-push-checks.sh`, `.claude/hooks/pre-push-gate.sh`, `.claude/hooks/tests/test_pre_push_gate.sh`).
5. `bash bin/run-pre-push-checks.sh` run on this branch itself before pushing — all checks (markdownlint, shellcheck, subpacks) pass.

Refs #1097

---

## Glossary

| Term | Definition |
|------|------------|
| skip marker | A documented commit-message token (`<!-- pre-push: skip -->`) that lets a contributor bypass the local pre-push check set for a genuine emergency. Deliberately grep-able so bypasses stay auditable — which only holds if the match can't fire by accident. |
| unanchored match | A `grep` pattern that matches anywhere in the text (a substring anywhere in the message) rather than requiring the whole line to equal the pattern. The root cause here: prose *about* the marker behaved identically to the marker itself. |
| whole-line match (`grep -x`) | Requires the entire line to equal the search string, not just contain it. Fixes the bug because the documented deliberate-bypass usage already places the marker on a line by itself. |
| release-cut model | This repo's branch model: daily PRs squash-merge to `dev`; `main` only receives release PRs from `dev`. The squash step is what carries an accidentally-matching commit message onto the shared `dev` branch. |

### DIFF
--- a/.claude/hooks/pre-push-gate.sh
+++ b/.claude/hooks/pre-push-gate.sh
@@ -50,7 +50,10 @@ fi
 
 SKIP_MARKER='<!-- pre-push: skip -->'
 HEAD_MSG=$(cd "$REPO_ROOT" && git log -1 --format='%B' 2>/dev/null)
-if echo "$HEAD_MSG" | grep -qF -- "$SKIP_MARKER"; then
+# -x: whole-line match only, so prose that mentions the marker inline
+# (e.g. a commit that documents the escape hatch) does not trigger it —
+# only a line consisting of exactly the marker does. See #1097.
+if printf '%s\n' "$HEAD_MSG" | grep -qxF -- "$SKIP_MARKER"; then
   echo "WARN: pre-push gate bypassed by skip marker in HEAD commit message." >&2
   echo "      Skipped commands will run in CI regardless — fix broken state before merging." >&2
   exit 0
--- a/.claude/hooks/tests/test_pre_push_gate.sh
+++ b/.claude/hooks/tests/test_pre_push_gate.sh
@@ -216,7 +216,46 @@ case9() {
   rm -rf "$sb"
 }
 
-case1; case2; case3; case4; case5; case6; case7; case8; case9
+# -------------------- CASE 10: marker MENTIONED in prose → does NOT bypass --------------------
+# Regression guard for #1097: the skip marker used to be grep-matched
+# unanchored, so a commit message that merely *discusses* the marker
+# (documentation, a review comment quoted verbatim, a revert body) matched
+# too and silently disabled every check. The match must be whole-line
+# (grep -x): a sentence that contains the marker string inline is NOT a
+# deliberate bypass, so the check must still run (and still block on a
+# failing command).
+case10() {
+  local sb; sb=$(make_sandbox)
+  cat > "$sb/.claude/project-config.json" <<'EOF'
+{"pre_push": {"commands": [{"name": "should-not-skip", "run": "echo oops; exit 1"}]}}
+EOF
+  # The marker appears INSIDE a sentence, not as its own line — this must
+  # NOT be treated as a deliberate bypass.
+  (cd "$sb" && git commit --amend -q -m "docs: explain the escape hatch
+
+This documents the <!-- pre-push: skip --> marker so contributors know
+it exists. It should not itself act as a bypass.")
+  run_hook "$sb" "$(push_json)" 2 "should-not-skip: FAILED" "marker-mentioned-in-prose-does-not-bypass"
+  rm -rf "$sb"
+}
+
+# -------------------- CASE 11: marker on its OWN LINE → still bypasses --------------------
+# The other direction of #1097's fix: a deliberate bypass — the marker as
+# a line by itself, exactly the shape the documented amend snippet emits —
+# must keep working after anchoring the match to -x.
+case11() {
+  local sb; sb=$(make_sandbox)
+  cat > "$sb/.claude/project-config.json" <<'EOF'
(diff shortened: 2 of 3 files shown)


atlas-apex (approved):

me2resh (commented):
## Code Review: PR #1111 — APPROVE

**Commit reviewed**: `53aed017d448a47d5d4108ceb33d0781935daad3`

Trust-chain change (pre-push gate). Reviewed rigorously against the reported #1097 defect. Verdict: **APPROVE**.

### What the fix does
Both `bin/run-pre-push-checks.sh` and `.claude/hooks/pre-push-gate.sh` switch the skip-marker test from an unanchored substring match (`echo "$HEAD_MSG" | grep -qF`) to a whole-line fixed-string match (`printf '%s\n' "$HEAD_MSG" | grep -qxF -- "$SKIP_MARKER"`). Both sites are now identical. The `printf '%s\n'` correctly splits the multi-line commit body one-line-per-line so `-x` anchors to a whole line.

### Verification performed (against the PR's actual HEAD, no checkout)
- **Both sites anchored, identically** → confirmed both files use `grep -qxF`. ✓
- **Legitimate bypass preserved** → case11 (`marker-own-line-still-bypasses`) confirms a marker on its own line still exits 0 with the WARN. `-x` keeps the documented emergency escape hatch working. ✓
- **Prose mention no longer bypasses** → case10 (`marker-mentioned-in-prose-does-not-bypass`) confirms a commit whose message merely mentions the marker inline runs the checks and blocks (rc 2, `should-not-skip: FAILED`). ✓
- **Tests are genuinely discriminating** → I reverted the gate to the pre-fix unanchored `grep -qF` and re-ran the suite: `marker-mentioned-in-prose-does-not-bypass` FAILS (want rc=2, got 0 — the inline mention wrongly bypasses), while all other cases pass. This proves case10 actually guards the regression rather than passing vacuously. ✓
- **Suite green on the fix** → 11/11 pass. ✓
- **shellcheck --severity=warning** clean on all three touched files. ✓

### Checklist
- Architecture & Design: Pass — minimal, surgical, matches the issue's suggested fix
- Code Quality: Pass — clear inline comments citing #1097; both call sites kept in lockstep
- Testing: Pass — two new cases, both directions, proven discriminating
- Security: Pass — this strengthens a trust-chain gate; closes a silent self-disabling bypass
- PR Description & Glossary: Pass — narrative Summary (what + why + propagation risk), two-direction Testing, Glossary, `Refs #1097`
- AgDR: N/A — narrow, reversible reinforcement of an existing control; PR body already cites AgDR-0104 for the pattern-match caveat, correctly framing this as a fixed-marker-on-own-line exception

### Notes
- No blocking issues. The `printf '%s\n'` + `grep -qxF -x` combination is the right tool: fixed-string (no marker chars treated as regex) and whole-line (only a deliberate own-line marker counts).

### Verdict
**APPROVE**

---
Reviewed by Rex (Code Reviewer Agent) · commit `53aed017d448a47d5d4108ceb33d0781935daad3`

Issue #1104, opened 2026-07-31
[Bug] Portfolio path-resolution treats case-only path differences as a split-portfolio, producing a false "broken config" banner on case-insensitive filesystems


me2resh:
## Affected
Portfolio path-resolution / SessionStart banners — `_lib-portfolio-paths.sh`, `_lib-ops-root.sh`, and the split-portfolio detection/validation hooks that emit the "split-portfolio v2 detected" and "portfolio config is broken" SessionStart messages.

## Given / When / Then
**Given** an ops fork on a case-insensitive filesystem (macOS/APFS default), where Claude Code is launched with two configured working directories that differ **only in letter-case** (e.g. `/Users/x/Playground/ai/apexyard` and `/Users/x/playground/ai/apexyard` — the same directory, identical inode).
**When** a SessionStart fires and the hooks resolve the ops-fork root one way (from `cwd` → lowercase `playground`) and the registry/sibling another way (`git rev-parse --show-toplevel` → canonical capital `Playground`).
**Then** the two path strings differ, so the hooks conclude there are two distinct repos and emit `split-portfolio v2 detected` with `Ops-fork root` != `Portfolio (sibling)`, plus `portfolio config is broken — registry ... does not exist` — expected: case-identical paths on a case-insensitive filesystem are recognized as the **same** directory, so single-fork mode is detected and no "broken"/"split" banner is shown.

## Repro
1. On macOS (case-insensitive APFS), clone an apexyard fork to `/path/To/apexyard`.
2. Launch Claude Code such that `cwd` and `git rev-parse --show-toplevel` disagree in case (e.g. `cd /path/to/apexyard` where the on-disk canonical case is `/path/To/apexyard`).
3. Observe the SessionStart banners: a phantom `split-portfolio v2 detected` (Ops-fork root vs Portfolio sibling differ only in case) and `portfolio config is broken — <registry> does not exist`, despite `.claude/project-config.json` having no `portfolio:` block and the repo being plain single-fork.

## Framework version
v5.3.0

## Severity
minor (cosmetic — misleading banners; no functional data loss, but the "broken config" wording alarms users and the phantom split can misdirect path resolution)

## Notes
- Confirmed same directory: `ls -di` on both paths returns an identical inode.
[...]

Issue #1088, opened 2026-07-30
[Chore] Decide how managed-project clones gain pre-push protection without executing their own tracked hooks


atlas-apex:
## Driver

#1087 (step 1 of #1086) originally wired `bin/install-git-hooks.sh` into `/handover`, so each newly-adopted project clone would get `core.hooksPath` set and gain pre-push protection. **The security review rated that HIGH and it was removed before merge.**

The finding, reproduced rather than reasoned about: the installer gates only on `[ -d "$TARGET_ROOT/.githooks" ]`. It performs no provenance check, no content check, no operator prompt. It does **not** install the framework's hooks — it points git at whatever `.githooks/` the target clone already ships. For the ops fork that is ApexYard's own. For a repo `/handover` has just cloned, it is **that repo's committed scripts**.

Measured against a scratch repo committing its own `.githooks/`, cloned, then run through the installer:

```
after 'git checkout -b': PAYLOAD-POSTCHECKOUT
after 'git fetch':       PAYLOAD-REFTXN
```

Note what fired: `post-checkout` and `reference-transaction`. **No push is required.** `reference-transaction` runs on essentially any ref update, which the framework performs routinely and silently in workspace clones.

Why this is a genuine trust-boundary crossing rather than a theoretical one: git's default hooks directory `$GIT_DIR/hooks` is **never cloned**. That deliberate property is precisely what makes cloning an untrusted repository a safe, read-only act. Repointing `core.hooksPath` at a *tracked* directory removes exactly that property — which is why husky-class tooling requires an explicit, user-run install step rather than doing it implicitly.

`/handover` is the least-trusted input path in the framework: it clones an arbitrary URL that the operator wants to **score**, before any trust has been established. Wiring an implicit "always attempt" hook install into that path inverts the risk.

## The gap this leaves open

Removing the `/handover` call was correct for #1087, but it means the original justification is only half-delivered. Managed-project clones carry **no framework presence in their own repo** — no `.claude/`, no `.githooks/`. The Claude Code `PreToolUse` hooks fire only because the *agent* runs in the ops-fork context. So today:

| Push path | Protected? |
|---|---|
| Agent-driven push in a managed project | Yes — `PreToolUse` hook |
| **Human terminal push in a managed project** | **No — nothing covers it** |

That gap predates #1086 and remains open. This ticket exists so it stays visible rather than disappearing with the removed wiring.

## Scope

[...]

PR #1113, opened 2026-08-01
fix(#1089): pin #1087 fail-closed behaviour and correct the installer doc

me2resh:
## Summary

- **Fixed the doc line that could re-introduce #1087's HIGH finding** — `docs/getting-started.md:158` said "`/setup` and `/handover` call it," which is exactly the sentence a future agent implementing #1088 (or extending `/handover`) could read as prior design intent and use to justify re-adding the `core.hooksPath` call PR #1087 removed for security reasons. It now says only `/setup` calls the installer, states *why* `/handover` deliberately does not, and points at `handover/SKILL.md`'s "1.5-clone" step (which already carries the full rationale) plus #1088, instead of duplicating the explanation in two places that could drift.
- **Named diagnostic + "deliberate degrade" comment across all three `_resolve_real_path` call sites** — `require-active-ticket.sh`, `_lib-git-hooks-path.sh`, and `bin/install-git-hooks.sh` (via the lib it sources) previously left `_resolve_real_path` silently undefined when `_lib-path-resolve.sh` was missing, so a caller saw a raw bash "command not found" with no explanation. Each site now defines a named stand-in (prints an actionable, once-per-run diagnostic, then echoes nothing) — same external behavior, but the degrade is now visible instead of silent, and the reason it degrades this way (fail toward gating, not toward exemption) is written down next to the code instead of living only in review history.
- **Pinned the fail-closed property with discriminating tests, not just comments** — three new test cases (one per consumer) prove: (1) an out-of-governance write target that's normally EXEMPT becomes BLOCKED when `require-active-ticket.sh`'s lib is missing; (2) an already-correctly-configured clone still gets the advisory "doesn't exist" banner (over-warn, never silently pass) when `_lib-git-hooks-path.sh`'s lib is missing; (3) a deliberate third-party `core.hooksPath` is NOT clobbered when the installer's own lib is missing. All three were verified discriminating by hand against a mutant with the relevant guard removed/reordered — see Testing below, including one case where the mutant genuinely DID clobber the third-party value, confirming the shipped ordering is load-bearing and not just defensive comment noise.
- **Documented `bin/install-git-hooks.sh`'s exit 4 in `/setup` Step 0.5** as a hard-failure state the skill must surface and stop on, not a warning to note and move past — exit 4 means the clone is NOT protected despite the write attempt.
- **Fixed a stale cross-reference** — `_lib-portfolio-paths.sh:110` pointed at `_resolve_real_path()` living in `require-active-ticket.sh`; PR #1087 moved that function into `_lib-path-resolve.sh`. Touched only this one comment line to minimize surface for the concurrent PR #1104, which also edits this file.

No behavior change to the ticket gate's silent degrade itself — both Rex and the Security Auditor independently verified on #1087 that failing closed (gating rather than exempting) is the correct outcome when the lib is missing. This PR documents and pins that behavior; it does not change it.

## Testing

- `bash .claude/hooks/tests/test_require_active_ticket_bash.sh` — 81/81 pass (78 pre-existing + 3 new `#1089 fail-closed` cases)
- `bash .claude/hooks/tests/test_check_git_hooks_installed.sh` — 9/9 pass (8 pre-existing + 1 new)
- `bash .claude/hooks/tests/test_install_git_hooks.sh` — 43/43 pass (42 pre-existing + 1 new)
- `bash .claude/hooks/tests/test_portfolio_paths.sh` — 45/45 pass (comment-only edit; sanity re-run)
- `shellcheck --severity=warning` on all 7 edited `.sh` files — clean
- `npx markdownlint-cli2` on both edited `.md` files — clean
- **Discriminating-test verification (by hand, not part of the shipped suite):**
  - Copied `require-active-ticket.sh` to a scratch dir, removed the `[ -n "$_og_real_target" ]` guard around the out-of-governance exemption block, and re-ran the exact scenario from the new test case — the mutant returned `rc=0` (wrongly exempt) where the shipped hook returns `rc=2` (correctly blocked). Confirms the new test is discriminating, not just present.
  - Copied `bin/install-git-hooks.sh` + `_lib-git-hooks-path.sh` to a scratch install root (without `_lib-path-resolve.sh`), neutralized the LOW-4 `--hooks-dir` containment check's `exit 2`, and ran it against a target repo with a deliberate, real, external `core.hooksPath`. The mutant **actually clobbered** the third-party value (`git config --get core.hooksPath` read back `.githooks` afterward) — empirically confirming the "load-bearing ordering" comment isn't decorative, and that the new non-clobber test would catch this exact regression.

## Acceptance criteria (from #1089)

- [x] `docs/getting-started.md` no longer states that `/handover` calls the installer, and explains why, referencing #1088
- [x] A test pins the fail-closed degraded mode for all three consumers of `_lib-path-resolve.sh` — proven discriminating (see Testing)
- [x] A test pins that a deliberate third-party `core.hooksPath` is not clobbered when the lib is missing
- [x] A comment records that the installer's check ordering is load-bearing, and that the ticket gate's silent degrade is deliberate rather than an oversight
- [x] The degraded mode emits a named diagnostic, not `command not found`
- [x] `/setup` Step 0.5 documents exit 4
- [x] `_lib-portfolio-paths.sh:110`'s cross-reference points at the current location
- [x] Full hook suite green; `require-active-ticket.sh`'s own suite still 78/78 (now 81/81 with the 3 new cases, all pre-existing 78 still pass)

Refs #1089

---

## Glossary

| Term | Definition |
|------|------------|
| Fail closed / fail open | A degraded control that blocks (closed) versus one that permits (open). For a security gate, closed is the safe direction — over-block rather than let something through unchecked. |
| Degraded mode | How code behaves when a dependency it needs is absent. This PR makes the degraded mode of three call sites visible (named diagnostic) and pinned (tests), without changing what it actually does. |
| Discriminating test | A test that fails against a plausible wrong implementation, not merely against a missing file. Verified here by hand — a mutant with the relevant guard removed/reordered turns green tests red. |
| Load-bearing ordering | Two operations whose *sequence*, not merely their presence, provides a safety guarantee — here, the `--hooks-dir` containment check running before `ghp_classify` is what stops a missing-lib degrade from reaching the installer's auto-repair branch. |
| `core.hooksPath` | Per-clone git config pointing git at a directory of hook scripts to run on git operations (e.g. `pre-push`). Never committed — set once per clone. |


atlas-apex (approved):

me2resh (commented):
## Code Review: PR #1113

**Commit**: `67854f96eb496816fbc502e61ab1ca48c1e9d60d`

### Summary

Closes the four LOW findings left non-blocking on #1087. This is a document-and-pin PR: it hardens the *diagnostics, comments, and test coverage* around the trust chain's fail-closed degraded mode without changing any gate behaviour. Nine files, 281/3. Every touched file maps directly to one of the four findings — no scope creep.

I reviewed this against the trust-chain bar (rigorous), grounded in the actual codebase, and ran the full suite plus independent mutation testing at the PR head (`67854f9`) in an isolated worktree — the main tree stayed on `dev`, no checkout.

### Checklist Results
- Testing:                   Pass (mutation-verified discriminating — see below)
- Security:                  Pass (trust chain provably unchanged; net hardening)
- Performance:               Pass (n/a — comments, diagnostics, tests)
- PR Description & Glossary: Pass (narrative Summary, thorough Testing, Glossary present, `Refs #1089`)
- Summary Bullet Narrative:  Pass (rich narrative, no label-only bullets)
- Technical Decisions (AgDR): N/A — no new decision. #1089 explicitly requires the reviewed #1087 behaviour be *documented and pinned, not changed*; no new dependency / service / schema / control is introduced.

### What I verified

**1. Doc fix removes the HIGH-re-introduction hazard (finding 1).** `docs/getting-started.md:158` no longer says `/handover` calls the installer. It now states only `/setup` calls it (for the ops fork), states *why* `/handover` deliberately does not (pointing `core.hooksPath` at a freshly-cloned third-party repo hands that repo's own committed hooks execution on the next ordinary git op — no push needed), and points at the single source of truth: the "1.5-clone" step in `.claude/skills/handover/SKILL.md`. I confirmed that step exists (SKILL.md line 135) and that line 170 carries the full "deliberately NOT invoked here / HIGH-1" rationale — so the two docs cannot drift, exactly as #1089 asked. Also links #1088 for the follow-up.

**2. Fail-closed is document-and-pin, NOT converted to hard-fail (findings 2 and 3).** Named `_resolve_real_path` stand-ins added in the `else` branches of `require-active-ticket.sh` and `_lib-git-hooks-path.sh`; `bin/install-git-hooks.sh` is covered transitively via the lib it sources (documented at its `. "$LIB"` line). I confirmed each stand-in echoes nothing to stdout and only prints a once-per-process diagnostic to stderr — matching the real function's fail-EMPTY contract in `_lib-path-resolve.sh`. So `$(_resolve_real_path …)` still captures empty, `_og_real_target` stays empty, the `[ -n … ]` guard skips the exemption, and the target falls through to the gate (BLOCKED). The stderr diagnostic cannot pollute `ghp_classify`'s stdout STATE parsing. Behaviour is unchanged; only the diagnostic and the pinning tests are new. Test case 77 (same target ALLOWED once a ticket exists, lib still missing) proves this is a degrade, not a hard-fail.

**3. The new tests are genuinely discriminating — verified by mutation, not taken on faith.**
- Ticket-gate fail-closed: I removed the `[ -n "$_og_real_target" ]` guard in a scratch copy → cases 76 and 78 flipped BLOCKED (rc=2) to EXEMPT (rc=0) and the suite went red (2 fails). Restored, green.
- Installer non-clobber / load-bearing ordering: I neutralized the LOW-4 containment `exit 2` (the hand-check the PR body describes) → the installer *actually clobbered* the deliberate third-party `core.hooksPath` (read back `.githooks` instead of the external dir), and the new case 15 caught it. Restored, green. This empirically confirms the "LOAD-BEARING ORDERING" comment in `bin/install-git-hooks.sh` is not decorative — the LOW-4 check running before `ghp_classify` is what stops a missing-lib empty resolve from reaching the auto-repair branch.

**4. exit 4 doc (finding 4a).** `/setup` Step 0.5 now documents exit 4 as a hard failure ("this clone is NOT protected"), which matches the installer's real exit-4 semantics (an unverified `git config` write). Consistent with the code.

**5. Stale cross-ref (finding 4b).** `_lib-portfolio-paths.sh` now points `_resolve_real_path` at `_lib-path-resolve.sh` (where #1087 moved it), which I confirmed is its current home. Comment-only, low conflict surface with the concurrent #1104.

**Suite + lint at the PR head:** test_require_active_ticket_bash 81/81, test_check_git_hooks_installed 9/9, test_install_git_hooks 43/43, test_portfolio_paths 45/45 — all matching the PR body. `shellcheck --severity=warning` on all 7 edited `.sh` files: clean. `markdownlint-cli2` on both `.md` files: clean.

### Issues Found
None.

### Suggestions
- nit (non-blocking): the PR body describes the `_lib-portfolio-paths.sh` change as "one comment line," but it expands one line to a four-line comment block. Still comment-only and contiguous, so conflict surface with #1104 is minimal — the wording just slightly undersells the hunk. No change requested.

### Verdict
**APPROVED**

Clean, tightly-scoped, mutation-verified trust-chain hardening that closes #1089's four findings without touching gate behaviour. The security posture strictly improves: a previously emergent, unpinned fail-closed property is now guarded by discriminating tests and documented next to the code.

---
Reviewed commit: `67854f96eb496816fbc502e61ab1ca48c1e9d60d`

PR #1114, opened 2026-08-01
fix(#1104): make portfolio path comparison case-insensitive-filesystem aware

me2resh:
## Summary

- **Root cause pinned down empirically, not guessed at** — on a case-insensitive filesystem (macOS/APFS default, Windows), bash's own `cd`/`pwd -P` do **not** re-case a path string (`cd lowercase-path && pwd -P` returns `lowercase-path` unchanged, verified on both GNU bash 5.x and macOS's stock `/bin/bash` 3.2), while `git rev-parse --show-toplevel` always resolves to the on-disk canonical case regardless of the case used to invoke it. `_lib-ops-root.sh`'s pin-first `resolve_ops_root()` can therefore return a raw, non-canonical root string (from a stale pin, or a differently-cased launch cwd), while every `_lib-portfolio-paths.sh` resolver anchors on `git rev-parse` (canonical). Comparing the two with a plain `=`/`case` string match misread the case-only difference as "two distinct repos" — the exact false `split-portfolio v2 detected` banner from #1104.
- **New shared helpers: `portfolio_path_eq` / `portfolio_path_under`** in `_lib-portfolio-paths.sh` — case-insensitive-filesystem-aware equality/containment checks. They prefer `-ef` (device+inode identity, immune to case entirely) when both paths exist, falling back to a case-folded string compare only for the non-existent-path case (nothing to `-ef` against there, so nothing to be wrong about).
- **Every path comparison in `portfolio_validate()` and `print-portfolio-primer.sh` now goes through the new helpers** instead of raw `case`/`!=` string matching — `_outside_fork`, the `pd_real`/`wd_real` default-location checks, and the onboarding/workspace_dir in-fork checks. This closes the false "split" banner *and* any analogous false "broken config" path in one sweep, since they all shared the same string-comparison bug shape.
- **`_lib-ops-root.sh`'s `resolve_ops_root_walk()` now anchors on `git rev-parse`'s canonical toplevel before walking up** — the same base `_portfolio_root()` already used — so the framework's two independent ops-root resolvers can't disagree on case going forward. Falls back to the raw start unchanged outside a git repo (no behavior change there).
- **Regression test added**: `test_portfolio_paths_case_insensitive_fs.sh` skips gracefully on a case-sensitive filesystem (the bug only exists where "Foo" and "foo" name the same on-disk entry), unit-tests the new helpers with a negative control (a genuinely different directory must *not* compare equal), reproduces the pin-vs-canonical-root false split end-to-end against `print-portfolio-primer.sh`, and confirms a real split-portfolio v2 setup still fires the banner — the fix suppresses only the case-only false positive, not a genuine split or a genuinely broken registry.

## Testing

1. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_ops_root.sh` — 8/8 pass
2. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_resolve_ops_root_pin.sh` — 8/8 pass
3. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_portfolio_paths.sh` — 45/45 pass
4. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_print_portfolio_primer.sh` — 3/3 pass
5. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_portfolio_paths_windows_drive_letter.sh` — 19/19 pass
6. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_split_portfolio_v2_migration.sh` — 25/25 pass
7. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_portfolio_agent_routing.sh` — 6/6 pass
8. `env -u CLAUDE_CODE_SESSION_ID bash .claude/hooks/tests/test_portfolio_paths_case_insensitive_fs.sh` (new) — 6/6 pass. Verified this test genuinely catches the regression by stashing the fix and re-running: it fails with the exact false "split-portfolio v2 detected" banner from the ticket, confirming it exercises the real bug rather than being a tautology.
9. `shellcheck --severity=warning` on all four changed/added files — clean, no findings.

Note on the `env -u CLAUDE_CODE_SESSION_ID` prefix: these tests are sensitive to a **pre-existing, unrelated** test-isolation gap — when run inside a live Claude Code session, `CLAUDE_CODE_SESSION_ID` is set to a real value and `resolve_ops_root()`'s pin-first lookup picks up the *real* session's already-written pin file instead of each test's synthetic sandbox. This reproduces identically against the unmodified `dev` code (confirmed by stashing this PR's changes and re-running), so it is not a regression introduced here — CI runners without a live session pin are unaffected either way.

## Glossary

| Term | Definition |
|------|------------|
| Case-insensitive filesystem | A filesystem (macOS APFS/HFS+ default, Windows NTFS) where `Foo` and `foo` name the same on-disk file or directory — two differently-cased path strings can point at one identical entry. |
| Case-preserving | The filesystem remembers the case a name was created with (so `ls` shows `Foo`), even though lookups ignore case — distinct from case-insensitive-and-*folding* filesystems that would normalize the name itself. |
| `-ef` | The POSIX `test`/`[` operator that compares two paths by device + inode number — true when they name the identical filesystem entry, regardless of the path strings used or the case involved. |
| Ops-fork root | The forked apexyard repo that governs a portfolio, anchored by the `.apexyard-fork` marker file or (legacy) the `onboarding.yaml` + `apexyard.projects.yaml` pair. |
| Pin (ops-root pin) | A per-session file at `~/.claude/apexyard/ops-root-<session-id>` written by `pin-ops-root.sh` at SessionStart, capturing the launch-cwd ops root so later hooks/sub-agents resolve consistently even if they later `cd` elsewhere. |
| Split-portfolio v2 | Mode where the private portfolio (registry, projects docs) lives in a separate sibling repo from the public framework fork, distinguished from a plain single-fork setup that merely carries the same `.apexyard-fork` marker. |

Refs #1104

### DIFF
--- a/.claude/hooks/_lib-ops-root.sh
+++ b/.claude/hooks/_lib-ops-root.sh
@@ -86,6 +86,28 @@ _LIB_OPS_ROOT_SOURCED=1
 # touches the pin.
 resolve_ops_root_walk() {
   local start="${1:-$PWD}"
+
+  # Case-insensitive-filesystem safety (me2resh/apexyard#1104): a literal
+  # $start/$PWD string can carry a different case than the on-disk
+  # directory name — bash's `cd`/`pwd -P` do NOT re-case (verified: `cd
+  # lowercase-path && pwd -P` returns "lowercase-path" unchanged on both
+  # GNU bash 5.x and macOS's stock /bin/bash 3.2), so whatever case the
+  # operator typed or the harness launched with survives untouched.
+  # `_lib-portfolio-paths.sh`'s `_portfolio_root()` anchors on `git
+  # rev-parse --show-toplevel` instead, which resolves via the OS and
+  # always returns the canonical on-disk case regardless of the case used
+  # to invoke it. Anchor THIS walk on the exact same base so the two
+  # independent ops-root resolvers can't disagree purely on case — when
+  # they did, a case-only difference between a pinned/cwd-derived root
+  # and a git-rev-parse-derived portfolio path got misread as "two
+  # distinct repos" (a phantom split-portfolio banner on a plain
+  # single-fork setup). Falls back to the raw (possibly non-canonical)
+  # `start` unchanged when not inside a git repo — same contract as
+  # before; there is nothing to canonicalize against in that case.
+  local canon
+  canon=$(cd "$start" 2>/dev/null && git rev-parse --show-toplevel 2>/dev/null) || canon=""
+  [ -n "$canon" ] && start="$canon"
+
   local r="$start"
   while [ -n "$r" ] && [ "$r" != "/" ]; do
     # v2 anchor (preferred): the explicit .apexyard-fork marker file.
--- a/.claude/hooks/_lib-portfolio-paths.sh
+++ b/.claude/hooks/_lib-portfolio-paths.sh
@@ -446,6 +446,102 @@ portfolio_agent_routing() {
 }
 
 # ------------------------------------------------------------------------------
+# Public: portfolio_path_eq <a> <b>
+#   Case-insensitive-filesystem-aware equality check for two absolute
+#   paths — do they name the SAME filesystem entry?
+#
+#   Why this exists (me2resh/apexyard#1104): on a case-insensitive
+#   filesystem (macOS/APFS default, Windows), two path strings that
+#   differ only in letter-case can be the identical on-disk directory.
+#   `git rev-parse --show-toplevel` resolves via the OS and always
+#   returns the canonical on-disk case (verified: `git -C
+#   /path/to/lowercase rev-parse --show-toplevel` still returns the
+#   canonical-cased path), but bash's `cd`/`pwd -P` do NOT re-case —
+#   verified empirically: `cd lowercase-path && pwd -P` returns
+#   "lowercase-path" unchanged, not the on-disk canonical case, on both
+#   GNU bash 5.x and macOS's stock `/bin/bash` 3.2. So a path resolved
+#   via `git rev-parse` (canonical) and a path resolved via a raw
+#   `$PWD`-derived string (whatever case the operator or harness
+#   originally typed) can be the SAME directory yet compare unequal with
+#   a plain `=`/`case` string match — which is exactly how a case-only
+#   difference between the ops-fork root and a portfolio path used to
+#   get misread as "two distinct repos" (a phantom split-portfolio /
+#   broken-config banner on a plain single-fork setup).
+#
+#   Strategy: prefer `-ef` (device+inode identity — the POSIX `test`
+#   operator, immune to case entirely) when both paths exist. Fall back
+#   to a case-folded string comparison when one/both don't exist yet
+#   (e.g. a workspace_dir that hasn't been created) — `-ef` needs a
+#   stat(2) on both sides. The fallback intentionally folds case
+#   UNCONDITIONALLY rather than trying to detect "is this filesystem
+#   case-insensitive": a false-positive fold on a case-SENSITIVE
+#   filesystem (treating two genuinely-different directories as equal)
+#   is the rarer, lower-cost failure mode next to the false "split
+#   portfolio" this helper exists to eliminate on every case-insensitive
+#   Mac/Windows setup — and the fold only ever applies to the
+#   non-existent-path fallback, where there is no inode to be wrong
+#   about in the first place.
+# ------------------------------------------------------------------------------
+portfolio_path_eq() {
+  local a="$1" b="$2"
+
+  [ "$a" = "$b" ] && return 0
+
+  if [ -e "$a" ] && [ -e "$b" ] && [ "$a" -ef "$b" ]; then
+    return 0
+  fi
+
+  local a_lc b_lc
+  a_lc=$(printf '%s' "$a" | tr '[:upper:]' '[:lower:]')
+  b_lc=$(printf '%s' "$b" | tr '[:upper:]' '[:lower:]')
+  [ "$a_lc" = "$b_lc" ]
+}
+
+# ------------------------------------------------------------------------------
+# Public: portfolio_path_under <path> <dir>
+#   Case-insensitive-filesystem-aware containment check — is <path> equal
+#   to <dir>, or does it live somewhere under <dir>? The prefix-match
+#   sibling of portfolio_path_eq, for the same reason (see its header
+#   comment above).
+#
+#   Strategy: walk UP from <path> (via dirname) to each ancestor that
+#   actually exists, and -ef-compare it against <dir> — this correctly
+#   handles <path> not existing yet (a file/dir that hasn't been created)
+#   by climbing past the non-existent tail, the same existing-ancestor
+#   pattern _portfolio_canonicalize already uses. Falls back to a
+#   case-folded string prefix match when <dir> itself doesn't exist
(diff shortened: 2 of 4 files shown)


atlas-apex (approved):

me2resh (commented):
## Code Review: PR #1114

**Commit**: `c4576c4a5d8cd8de49917af4ea1145ee16e44afc`

**Verdict: APPROVE**

### Summary
Fixes #1104: on a case-insensitive filesystem (macOS/APFS, Windows), portfolio path-resolution treated a case-only path difference (`Playground` vs `playground`) as two distinct repos, producing a phantom `split-portfolio v2 detected` / `broken config` SessionStart banner on a plain single-fork. The fix adds two case-insensitive-FS-aware helpers (`portfolio_path_eq` / `portfolio_path_under`), routes every string path comparison in `portfolio_validate()` and `print-portfolio-primer.sh` through them, and anchors `resolve_ops_root_walk()` on the same canonical `git rev-parse` base as `_portfolio_root()` so the two independent ops-root resolvers can't diverge on case.

I reviewed this as trust-chain-adjacent (ops-root resolution feeds merge-gate marker resolution) and verified everything empirically against both the fixed PR head and the unfixed base, in an isolated worktree — the main tree was never checked out onto the PR.

### Checklist Results
- Security / Trust-chain:    Pass (marker-feeding path gains only canonicalization, no case-fold)
- Performance:              Pass (one extra `git rev-parse` per pin-miss walk; negligible at SessionStart)
- Summary Bullet Narrative:  Pass (excellent what+why bullets)
- Technical Decisions (AgDR): N/A (see note below)
- Adopter Handbooks:         N/A (no DB migration → migration-safety does not fire; no domain/app/infra layers; no ts/py/go/rust)

### What I verified

**1. Helpers are correct.** `-ef` (device+inode) is preferred when both paths exist — immune to case entirely; the case-fold string compare is reached only when one/both paths don't exist (no inode to stat). Direct probing under `/bin/bash`:
- Prefix-boundary trap holds: `/x/apexyard-portfolio/foo` is NOT under `/x/apexyard`, `/x/apexyardX` is NOT under `/x/apexyard`, a genuine child IS under. The `"$dir_lc"/*` glob correctly requires the `/` segment boundary.
- Negative control: genuinely-different dirs → not-equal; both-exist different dirs → `-ef` governs → not-equal.

**2. False-equality edge — bounded and acceptable.** The unconditional case-fold CAN report FALSE equality, but ONLY for paths that (a) differ solely by case AND (b) have at least one non-existent side (so `-ef` cannot run). This is:
- Explicitly documented as a conscious tradeoff in the helper header;
- Correct on a case-insensitive FS (two paths that case-fold the same ARE the same entry);
- On a case-sensitive FS it requires a pathological setup (e.g. a fork dir literally named `Projects` vs `projects`), and the only consequence is suppressing a diagnostic banner — no data loss, no security effect;
- Blast-radius-confined: these helpers are used ONLY in `portfolio_validate()` and `print-portfolio-primer.sh` (informational SessionStart banners), NOT in marker/gate resolution.

**3. Every string comparison replaced.** All 5 sites in `portfolio_validate()` (`_outside_fork`, the `pd_real`/`wd_real` equality checks, the `onboarding` and `workspace_dir` containment checks) and the single `_outside_root` in `print-portfolio-primer.sh` now route through the helpers. The third banner-emitter, `check-portfolio-config.sh`, delegates entirely to `portfolio_validate()` and inherits the fix — no comparison site missed.

**4. `resolve_ops_root_walk()` canonicalization.** Verified directly (fixed vs unfixed):
- Fixed lib returns the CANONICAL ops-fork path from every start point — including a lowercased clone subdir; unfixed `dev` returned non-canonical spellings (lowercase, and unresolved `/var` → `/private/var`).
- Pin-first fast path is UNCHANGED — the diff only adds a canonicalization block at the top of `resolve_ops_root_walk()`; `resolve_ops_root()`'s pin lookup is untouched (`test_resolve_ops_root_pin` 8/8).
- Workspace-clone resolution is NOT broken — from a deep nested-clone subdir (mixed and lowercased) the walk still climbs to the ops fork (`test_ops_root` 8/8).
- Fails soft (`|| canon=""`) outside a git repo — same contract as before.

**5. Genuine split still detected (critical non-regression).** The new test's `case_real_split_still_detected` builds a REAL sibling portfolio dir and asserts `split-portfolio v2 detected` still fires — passes on the fixed code. The fix suppresses only the case-only false positive, not a genuine split or a genuinely broken registry.

**6. Test is discriminating and skips gracefully.** I ran the new test against the UNFIXED base libs: it FAILS, reproducing the exact #1104 phantom banner (`Ops-fork root: …/caseforkxyz` vs `Portfolio (sibling): …/CaseForkXYZ`, differing only in case). Against the fixed libs: 6/6 pass. It probes for case-insensitivity first and `exit 0`s on a case-sensitive FS. All 8 related suites pass at PR head; `shellcheck --severity=warning` clean on all four files. (Verified on macOS APFS, GNU bash 5.3.)

**7. Pre-existing test-isolation gap.** Confirmed NOT introduced here — the PR does not touch `resolve_ops_root()`'s pin-lookup (the source of the `CLAUDE_CODE_SESSION_ID` pickup). The `env -u CLAUDE_CODE_SESSION_ID` prefix in the Testing section reproduces identically on unmodified `dev`. Author's note is accurate; not a blocker.

### Advisory notes (non-blocking)
- **`tr` dependency.** The case-fold fallback relies on `tr`; if `tr` were ever missing at runtime both sides fold to an empty string and everything compares equal. In practice this is not a defect: `tr` is a POSIX coreutil present in every real hook-execution environment, the lib already depends on `tr` in `_portfolio_canonicalize` (pre-existing), and `tr` is the correct portable choice given the bash-3.2 target (no `${var,,}`). A `command -v tr` guard that fails closed to a plain string compare would be belt-and-braces, not required.
- **AgDR.** The unconditional-case-fold tradeoff is a genuine "accept a known imperfection" decision in a trust-chain-adjacent lib. The mechanical AgDR gate does not fire on `.claude/hooks/**`, the reasoning is thoroughly documented inline and in #1104, and this follows the established bug-fix-PR pattern for this file (#945 / #1018 / #1029 / #1034 / #951 / #952). AgDR not required; if the team wants the tradeoff formally recorded, an AgDR would be the natural home.
- **Trust-chain / Security Auditor.** This touches `.claude/hooks/**`, which auto-fires the Security Auditor trigger. My analysis finds no security-relevant behavior change — the marker-feeding `resolve_ops_root_walk()` gains only canonicalization (no fold, no false-equality), and the case-folding helpers are confined to informational banners. A Hakim pass is reasonable as defense-in-depth; I found nothing to fix.
- **Theoretical edge (not a regression).** `portfolio_path_under` with a trailing-slash `dir` or `dir=/` reports not-under — but an ops-fork root is never `/` or trailing-slashed, and the original `case` code behaved identically.

### Verdict
**APPROVED** — well-scoped, thoroughly tested, and well-documented. The root cause was pinned empirically rather than guessed, the two resolvers are now provably consistent, and the critical non-regression (a genuine split still fires) is asserted by the test.

---
Reviewed commit: `c4576c4a5d8cd8de49917af4ea1145ee16e44afc`

Issue #25, opened 2026-04-11
[Feat] extend architecture-paths defaults to cover SAM / Helm / Serverless Framework IaC


atlas-apex:
## Context

The current architecture-paths defaults in \`require-agdr-for-arch-changes.sh\` catch Terraform (\`*.tf\`), Docker Compose, Dockerfiles, and GitHub Actions workflows. Rex's review of PR #24 flagged that several other common IaC shapes are missing from the defaults:

- **Serverless Framework** — \`serverless.yml\` / \`serverless.yaml\`
- **Helm charts** — \`Chart.yaml\`, \`values.yaml\`
- **AWS SAM** — \`template.yaml\` / \`template.yml\`

The question is purely about which patterns are **universally common enough** to justify being in the framework's defaults (low false-positive risk across any fork), vs which should stay as project-config overrides.

## Decision criteria (framework-agnostic)

| Pattern | Industry prevalence | Ambiguity risk | Default candidate? |
|---------|-------------------|----------------|-------------------|
| \`(^|/)serverless\\.ya?ml$\` | Common (Serverless Framework) | Very low — "serverless" as a filename has no non-IaC meaning | **Yes** |
| \`(^|/)Chart\\.ya?ml$\` | Common (Helm) | Low — \`Chart.yaml\` is Helm-specific naming | **Yes** |
| \`(^|/)template\\.ya?ml$\` | Common (SAM, CFN) | **High** — Jekyll, Hugo, Rails generators, many frameworks use \`template.yaml\` for unrelated things | **No — project override** |
| K8s manifests (\`*-deployment.yaml\`, etc.) | Common (Kubernetes) | **High** — too many naming patterns, no single regex is safe | **No — project override** |

## Acceptance Criteria

- [ ] Add \`(^|/)serverless\\.ya?ml$\` to default architecture paths (unambiguous)
- [ ] Add \`(^|/)Chart\\.ya?ml$\` to default architecture paths (unambiguous)
- [ ] Do NOT add \`template.yaml\` or K8s manifests to defaults (too ambiguous). Document as project-config overrides in the hooks README.
- [ ] Update \`.claude/hooks/README.md\` known-gap callout to list SAM \`template.yaml\` and K8s manifests as examples of project-config overrides
- [ ] Amend AgDR-0001 post-ship section with the new patterns and rationale
- [ ] Smoke test new patterns against fixtures

## Non-goals

[...]

PR #1116, opened 2026-08-01
chore(#1088): record AgDR-0115 (accept residual, forge-native check)

### LINKED ISSUE #1088: [Chore] Decide how managed-project clones gain pre-push protection without executing their own tracked hooks

me2resh:
## Summary

- **Decision recorded: accept the terminal-push residual (Option D).** `docs/agdr/AgDR-0115-managed-clone-terminal-push-accept-residual.md` closes #1088 by choosing NOT to build a new client-side mechanism for a human's terminal `git push` inside a managed-project (`workspace/<name>/`) clone. Agent-driven pushes there already stay covered by the ops-fork's own `PreToolUse` hooks — this decision is about the one path that isn't: a human typing `git push origin main` directly. For that path, the managed repo's own **server-side branch protection** is now recorded as the actual control, not a gap ApexYard's client-side hooks are pretending to cover.
- **Why D, not a new client-side hook (Option A/B/C).** The AgDR weighs all four options the issue asked for. B is rejected outright — it reintroduces the exact stale-absolute-path failure #1086/#1087 exist to fix. A (installing into the clone's untracked `.git/hooks/pre-push`) is the only trust-safe *direction* if this is ever revisited, but it's deliberately NOT pre-committed here — it needs its own design and its own ticket, and a forge-native fix should be compared first since it survives re-clone and isn't `--no-verify`-bypassable. C (prompt at clone time) is named as weak-alone. The AgDR also restates, prominently, the load-bearing rail from #1087's HIGH finding: no config ApexYard writes may ever point a clone's `core.hooksPath` at that clone's own tracked content — and the risk-direction invariant (block = safe; silent `exit 0` against a third-party clone = the dangerous outcome) so neither can quietly re-invert in a future change.
- **A forge-native nudge, not a new gate.** `.claude/skills/handover/SKILL.md` gains step 1.7: an advisory, fail-open check that queries whether the managed repo's default branch has server-side branch protection turned on, and prints a one-line nudge if it's absent or can't be confirmed. It dispatches on `tracker_kind` (gh / glab / a generic manual-check note for `custom`/`none`) — the same pattern `tracker_review_submit`/`tracker_pr_merge` already use — rather than hardcoding one forge's CLI. It never blocks the rest of `/handover`; a network failure, missing auth, or unsupported tracker kind just skips with a note.
- **Fixed a footgun in `bin/install-git-hooks.sh`'s own header.** The "Background" section still said the installer "needs to be invoked from `/setup` ... and `/handover`," even though the `/handover` wiring was removed on a HIGH security finding in #1087. A later paragraph (added while closing #1089) already correctly said `/handover` does NOT call it — so the file was internally contradicting itself, which is exactly the kind of stale claim a future agent could read as license to re-add the removed wiring. Corrected both the "Background" paragraph and the `--repo-dir` option description to state plainly that ONLY `/setup` invokes this script, pointing at `handover/SKILL.md`'s own note and AgDR-0115.

## Testing

1. `shellcheck --severity=warning bin/install-git-hooks.sh` — clean, no findings.
2. `npx markdownlint-cli2` against the new AgDR and the edited `handover/SKILL.md` — 0 issues.
3. `bash .claude/hooks/tests/test_handover_clone_prompt.sh` — 18/18 passed (the new step 1.7 content sits between two of the test's pinned literal-substring assertions and doesn't disturb any of them).
4. `bash .claude/hooks/tests/test_install_git_hooks.sh` — 43/43 passed (header-only change; no behavioral edit to the installer's logic).

## Glossary

| Term | Definition |
|------|------------|
| `core.hooksPath` | Per-clone git config naming the directory git executes hooks from. Not inherited by cloning — every fresh clone starts unset, which is why `/setup` re-runs the installer on every ops-fork configuration. |
| tracked vs untracked hooks | Hooks committed to the repo (`.githooks/`, travels with a clone) versus hooks in `$GIT_DIR/hooks` (never cloned). Pointing `core.hooksPath` at *tracked* content in a repo you don't control hands that repo's own committed scripts execution — the #1087 HIGH finding. |
| residual (accepted) | A known gap in coverage that a decision explicitly chooses NOT to close right now, with the reasoning for why written down (here: AgDR-0115) rather than silently left unrecorded. |
| server-side branch protection | A forge-hosted (GitHub/GitLab) setting on a repo's default branch requiring PRs/reviews before a push lands. It runs on the forge, not the client, so it isn't defeated by `git push --no-verify` or a missing client-side hook. |
| advisory / fail-open | A check that warns but never blocks, and treats "couldn't determine the answer" the same as "the answer was concerning" rather than risking a false "all clear". Step 1.7 is advisory and fail-open by design — it must never stop `/handover` from completing. |

Refs #1088

### DIFF
--- a/.claude/skills/handover/SKILL.md
+++ b/.claude/skills/handover/SKILL.md
@@ -262,6 +262,74 @@ Which harness(es) will drive work on {name}?
 
 Record the pick as `$SELECTED_HARNESSES` (comma-separated, or `claude-code` if 1/default) for the step 10 summary.
 
+### 1.7. Check server-side branch protection (advisory, default: always attempt)
+
+> Why this exists. AgDR-0115 records the decision that ApexYard does NOT install any client-side push protection into a managed-project clone — doing so would mean pointing `core.hooksPath` at that clone's own tracked `.githooks/`, which is precisely the HIGH-risk wiring #1087's security review removed (see the "1.5-clone" note above and `bin/install-git-hooks.sh`'s header). The accepted consequence: a human running `git push origin main` from a terminal inside this clone is not blocked by anything ApexYard installs — only agent-driven pushes stay covered, via the ops-fork's own `PreToolUse` hooks. For that residual, the managed repo's own **server-side branch protection** is the actual control, not a client-side backstop ApexYard ships. This step only checks whether that control is turned on and nudges the operator if it looks like it isn't — it is advisory, forge-agnostic, and fail-open: a missing check, a network failure, or an unsupported tracker kind never blocks the rest of `/handover`.
+
+Only run this check when the target repo's `<owner/name>` is known (a Git URL was given in step 1). A plain local path with no resolvable remote host has nothing to query: set `BRANCH_PROTECTION_STATUS="skipped (no remote host)"`, print nothing, and continue straight to step 2.
+
+Otherwise, resolve the default branch (prefer the local clone when available; fall back to the API) and the tracker kind, dispatching the protection check the same way `tracker_review_submit` / `tracker_pr_merge` already dispatch on `tracker_kind` (`.claude/hooks/_lib-tracker.sh`) rather than hardcoding one forge's CLI:
+
+```bash
+source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-read-config.sh"
+source "$(git rev-parse --show-toplevel)/.claude/hooks/_lib-tracker.sh"
+
+kind=$(tracker_kind "<owner/name>")
+
+default_branch=""
+if [ "$CLONE_STATUS" = "cloned" ] || [ "$CLONE_STATUS" = "preserved" ]; then
+  default_branch=$(git -C "$WORKSPACE_DIR/<name>" symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's#^origin/##')
+fi
+if [ -z "$default_branch" ] && [ "$kind" = "gh" ]; then
+  default_branch=$(gh api "repos/<owner/name>" --jq .default_branch 2>/dev/null)
+fi
+
+BRANCH_PROTECTION_STATUS="skipped (default branch unknown)"
+if [ -n "$default_branch" ]; then
+  case "$kind" in
+    gh)
+      # gh api exits non-zero on BOTH "404 not protected" and a real
+      # failure (auth/network/no-permission/rate-limit). This check does
+      # not need to tell those apart — the response is the same either
+      # way: advise, never block. Fail-open by construction, not by
+      # accident.
+      if gh api "repos/<owner/name>/branches/${default_branch}/protection" >/dev/null 2>&1; then
+        BRANCH_PROTECTION_STATUS="protected"
+      else
+        BRANCH_PROTECTION_STATUS="not protected (or could not be verified)"
+      fi
+      ;;
+    glab)
+      encoded=$(printf '%s' "<owner/name>" | sed 's#/#%2F#')
+      if glab api "projects/${encoded}/protected_branches/${default_branch}" >/dev/null 2>&1; then
+        BRANCH_PROTECTION_STATUS="protected"
+      else
+        BRANCH_PROTECTION_STATUS="not protected (or could not be verified)"
+      fi
+      ;;
+    none|custom|*)
+      # No queryable, generically-shaped REST endpoint for these kinds —
+      # a `custom` tracker's PR/MR host is unknown, and `none` has no host
+      # at all. Say so plainly rather than guessing at an API shape.
+      BRANCH_PROTECTION_STATUS="skipped (no queryable forge for tracker.kind=${kind} — check your forge's branch protection settings manually)"
+      ;;
+  esac
+fi
+```
+
+If `$BRANCH_PROTECTION_STATUS` is `"not protected (or could not be verified)"`, print this nudge once (non-blocking — continue straight to step 2 regardless):
+
+```
+⚠ <name>'s default branch (<default_branch>) has no server-side branch
+protection turned on (or it couldn't be confirmed). ApexYard's own hooks
+only cover agent-driven pushes in this clone — a human running `git push`
+from a terminal here is not blocked. Server-side branch protection on the
+forge is the actual control for that gap (see AgDR-0115). Recommended:
+require a pull request before merging to <default_branch>.
+```
+
+If `$BRANCH_PROTECTION_STATUS` is `"protected"` or either `"skipped"` variant, print nothing further — the step 10 summary line is enough.
+
 ### 2. Read the surface area
 
 Use `$WORKSPACE_DIR/<name>/` as `<repo>` when available (clone succeeded or path was given). Fall back to GitHub API reads only when `$CLONE_STATUS` is `declined` or `failed`. Local reads are preferred — they are cheaper and more complete.
@@ -1556,6 +1624,7 @@ Registry updated:            apexyard.projects.yaml ({added | skipped})
 Next-step tickets filed:     {N filed of M offered | none offered (zero risks) | declined (skipped all) | skipped (registry not appended)}
 Workspace clone:             workspace/{name}/ ({cloned at step 1.5 | preserved (already existed) | declined (--no-clone) | failed: <reason> | n/a (local path given)})
 Harness:                     {"Claude Code (native)" | "<harness list> — install command(s) printed, see docs/harnesses/<harness>.md" | "other / not sure — no adapter yet"}
+Branch protection:           {"protected" | "not protected (or could not be verified) — nudge printed" | "skipped (default branch unknown)" | "skipped (no queryable forge for tracker.kind=<kind>)" | "skipped (no remote host)"}
 Validation:                  {"completed — verdict <GREEN|YELLOW|RED>" | "skipped" | "not offered (project is active)"}
 
 Tech stack: {one-liner}
@@ -1602,6 +1671,7 @@ Filed follow-up tickets:
 22. **`AGENTS.md` and `handover-assessment.md` don't duplicate — they split by reader** — `handover-assessment.md` (ops fork) is the **operator's** full analysis: risks, harnessability verdict, integration plan, next-step tickets. `AGENTS.md` (in the target repo) is the **agent's** concise operating manual: stable commands, layout, conventions, and the up-front gotchas an agent needs to start working. The volatile risk/integration analysis stays in the assessment; it is not copied into `AGENTS.md`. For low-harnessability repos, `AGENTS.md`'s Gotchas section surfaces what's fragile/missing (no strict types, no lint baseline, no coverage signal) — the in-repo echo of the assessment's LOW warning.
 23. **The "Governed by ApexYard" badge (step 8.6) is opt-in, PR-delivered, idempotent, and never a default** — row 9 of the step 5.6 checklist is **default-OFF** in every mode (checklist default, `--all`, and interactive `all`); it is only generated when explicitly ticked or comma-listed. Every run confirms with the operator by name (`Add a "Governed by ApexYard" badge to <name>'s README?`) before touching the target repo — there is no silent or bulk-implied consent for an externally-visible README edit. The badge is delivered via a branch + PR exactly like `AGENTS.md` (never a direct commit). It is **idempotent**: if either badge variant (`governed_by` or `built_with`) is already present in the README, the step skips and reports `already present` rather than adding a duplicate or swapping variants. The badge markdown itself (URL, color `#2F6DF6`, `flat-square` style) is fixed and quoted verbatim in step 8.6 — don't re-derive it. See AgDR-0090.
 24. **`docs/harnesses/README.md` is the single source of truth for harness support — step 1.6 summarises and links it, never duplicates the matrix.** Read the doc fresh when printing a harness's install command, precondition, or tier rather than trusting a stale table baked into this skill. Never round a harness's tier up — Cursor is failClosed-only, not live-proven, and the skill says so verbatim. Step 1.6 is informational only: it prints the adapter install command but never runs it against the target repo.
+25. **The branch-protection check (step 1.7) is advisory and fail-open, never a gate.** It never blocks the rest of `/handover`, never retries, and never escalates a check failure into anything stronger than the one-line nudge. It exists because AgDR-0115 accepted that ApexYard installs no client-side push protection into a managed-project clone — the nudge is the only signal the operator gets that the *actual* control (the forge's own server-side branch protection) might be off. Dispatch on `tracker_kind` (`.claude/hooks/_lib-tracker.sh`), the same pattern `tracker_review_submit`/`tracker_pr_merge` use — never hardcode one forge's CLI or REST shape. `custom` and `none` tracker kinds get a generic manual-check note, not a guessed API call.
 
 ## When to use this
 
--- a/bin/install-git-hooks.sh
+++ b/bin/install-git-hooks.sh
@@ -13,9 +13,20 @@
 # `core.hooksPath` is a PER-CLONE git config value, not a repo property —
 # it lives in `.git/config`, never committed, so every fresh clone starts
 # unset regardless of how many times a sibling clone has run this script.
-# That's exactly why this needs to be invoked from `/setup` (for the ops
-# fork) and `/handover` (for each newly-cloned managed-project workspace),
-# not run once and forgotten.
+# That's exactly why this needs to be invoked from `/setup`, every time it
+# configures the ops fork's own clone — a config value that isn't repeated
+# per clone doesn't stick.
+#
+# ONLY `/setup` invokes this script. `/handover` deliberately does NOT —
+# running it against an arbitrary, just-cloned managed-project workspace
+# would point core.hooksPath at THAT repo's own tracked .githooks/ (whatever
+# it ships), handing it silent execution with no provenance check (PR #1087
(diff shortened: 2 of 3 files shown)


atlas-apex (approved):

me2resh (commented):
## Security Review: PR #1116 — AgDR-0115 (managed-clone terminal-push residual + forge-native check)

**Commit reviewed**: `5dc1a341caa11d7d320f65c0ed829fa65a514b03`
**Verdict**: **SECURE** ✅

▸ Hakim (Security Auditor) — trust-boundary review (trigger: PR touches the trust chain around `bin/install-git-hooks.sh` + `/handover` clone path, and records a security-category AgDR).

This is a decision-recording PR plus one new advisory check. I reviewed it against the exact failure it exists to keep buried: the #1087 HIGH-1 RCE, where pointing a cloned repo's `core.hooksPath` at its own **tracked** `.githooks/` hands that untrusted repo silent code execution on the next ordinary git op (`post-checkout` / `reference-transaction` fired in the repro — no push required). All four assessment axes pass.

### 1. The load-bearing rail is stated and preserved ✅

`docs/agdr/AgDR-0115-managed-clone-terminal-push-accept-residual.md` carries a dedicated, prominently-headed section — *"Load-bearing guardrail (record prominently — this is the rail, not a footnote)"* — stating the rail as a **class**, not just the one caught instance:

> No configuration ApexYard writes into a cloned repo may ever cause that repo's own TRACKED content to execute … never point `core.hooksPath` at a clone's tracked `.githooks/` … That is precisely the #1087 HIGH finding, and it must never recur regardless of which option a future iteration picks.

Nothing in the PR re-introduces the wiring. The new `/handover` step 1.7 never calls `bin/install-git-hooks.sh`, never sets `core.hooksPath`, and never references the clone's `.githooks/`. Option B (point `core.hooksPath` at an out-of-clone dir) is explicitly **rejected** in part *because* it edges toward the same shape — the AgDR reasons about the class, which is the correct depth.

### 2. The risk-direction invariant is right, and not re-inverted ✅

The AgDR's *"Risk-direction correctness"* section restates the #1088 invariant verbatim in the correct direction:

> a block, or any non-zero/error exit, is the SAFE outcome. Exit 0 (silent success) against a third-party clone is the outcome that transfers code execution.

Important: the new step 1.7 check's own exit semantics (`gh api … protection` → exit 0 = "protected", non-zero = "not protected / unverifiable" → nudge) are on a **different axis** from the installer's invariant and do **not** re-invert it. Step 1.7 is a pure read — it transfers no execution — so "exit 0 = good" there is not the dangerous installer "exit 0 = I set hooksPath" case. The AgDR keeps the two cleanly separated; no confusion is introduced.

### 3. The new `/handover` step 1.7 check is safe ✅

Verified READ-ONLY, ADVISORY, FAIL-OPEN, and non-executing against untrusted content:

- **Read-only.** The only operations are `git -C … symbolic-ref --short refs/remotes/origin/HEAD` (a pure ref read — runs **no** hooks, no checkout, no fetch) and read-only forge GETs (`gh api repos/…/branches/<b>/protection`, `glab api projects/…/protected_branches/<b>`). No write, no checkout, no fetch, nothing that could fire `post-checkout`/`reference-transaction`. And critically it never sets `core.hooksPath`, so even a hostile `.githooks/` in the clone stays inert (git never clones `$GIT_DIR/hooks`).
- **Advisory / fail-open by construction.** The `gh` branch's own comment nails it: `gh api` exits non-zero on both "404 not protected" and a real auth/network/rate-limit failure, and the check *deliberately does not distinguish them* — both map to the nudge, never a block. Default `BRANCH_PROTECTION_STATUS="skipped (default branch unknown)"` is the safe fallback; unknown/`custom`/`none` tracker kinds get a plain manual-check note via the `none|custom|*` catch-all rather than a guessed API shape. The prose (`SKILL.md` step 1.7 + guardrail #25) states three times that it never blocks, never retries, never escalates.
- **No injection / execution vector from hostile repo metadata.** The one value that flows from the untrusted clone is `default_branch` (from `symbolic-ref` or the API's `.default_branch`). It is interpolated **only inside double-quoted argument strings** passed to `gh`/`glab` — no word-splitting, no glob, no re-evaluation, so shell metacharacters in a branch name cannot inject a command. Path-traversal via the branch name is also closed: the source is always a **valid git ref** (git's `check-ref-format` forbids `..`, control chars, `~^:?*[\`, etc.), so the worst a hostile default-branch name can do is add a legitimate `/`-segment to a REST path that then 404s and fails open to the nudge. No trust is placed in any file *content* from the clone — only a ref name is read.
- **Dispatch is real, not fabricated.** `tracker_kind()` (`.claude/hooks/_lib-tracker.sh:243`) and `_lib-read-config.sh` both exist; the `case "$kind"` shape mirrors the established `tracker_review_submit`/`tracker_pr_merge` pattern rather than hardcoding one forge.

### 4. The installer footgun fix genuinely closes the "re-add the HIGH wiring" path ✅

`bin/install-git-hooks.sh`'s header previously **contradicted itself** — the "Background" paragraph and the `--repo-dir` description both still implied `/handover` invokes the installer, while a later paragraph (from #1089) correctly said it must not. That stale claim is exactly the kind a future agent could read as license to re-wire the removed HIGH-risk call. Both spots are now corrected to state plainly that **ONLY `/setup`** invokes the script, with an explicit *"Do not re-add a `/handover` call … without re-reading both first — that is precisely the footgun this comment exists to block,"* cross-referencing HIGH-1, the SKILL note, and AgDR-0115. This is a real defense-in-depth improvement to the trust chain's documentation surface. Header-only change; the installer's logic (and its 43/43 tests) is untouched.

### Checklist Results
- Injection Prevention:   Pass (branch name is a git-ref-constrained string in double-quoted `gh`/`glab` args — no shell or path-traversal injection)
- Auth & Authorisation:   Pass (read-only forge GETs; fail-open on missing auth)
- Data Protection:        Pass (no data written to the untrusted clone; only a ref name read)
- Trust-Chain Integrity:  Pass (no gate weakened; the #1087 rail is restated and preserved; footgun closed)

### Findings
- **CRITICAL / HIGH / MEDIUM: none.**
- **LOW (informational, non-blocking):** the `glab` branch encodes the project with `sed 's#/#%2F#'` (no `g` flag), so only the first `/` is percent-encoded. Fine for `owner/name`; a GitLab **sub-group** project (`group/subgroup/name`) would under-encode. The failure mode is a 404 → fail-open nudge, so there is **no security impact** — purely a functional edge case if you later care about GitLab subgroups.

### Recommendations
None required for merge. The LOW note above is optional polish for GitLab subgroup support and can be deferred.

### Verdict
**SECURE** — the decision preserves the #1087 rail, keeps the risk-direction invariant correct, adds only a read-only/advisory/fail-open check with no injection or execution surface against the untrusted clone, and hardens the installer's docs against re-adding the removed wiring. Approving the security posture (this is a `--comment` verdict, not a GitHub `--approve`).

---
📌 Reviewed commit: `5dc1a341caa11d7d320f65c0ed829fa65a514b03`

me2resh (commented):
## Code Review: PR #1116

**Commit**: `5dc1a341caa11d7d320f65c0ed829fa65a514b03`

### Summary

Records **AgDR-0115** — the decided outcome of #1088: accept the terminal-push residual in managed-project clones (Option D) rather than build a new client-side push-protection mechanism, and lean on the managed repo's own server-side branch protection for the one uncovered path (a human's terminal `git push` inside `workspace/<name>/`). Adds an advisory, fail-open branch-protection nudge to `/handover` (step 1.7), and repairs a self-contradicting header in `bin/install-git-hooks.sh`. Trust-chain-adjacent but right-sized — the substance is a decision record plus an advisory skill step; the only code file touched changes comments only.

### Checklist Results

- Architecture & Design:        Pass
- Code Quality:                 Pass
- Testing:                      Pass (statically verified — see below)
- Security:                     Pass (trust-chain reasoning is sound; no new attack surface)
- Performance:                  N/A
- PR Description & Glossary:    Pass (narrative Summary + 5-term Glossary + `Refs #1088`)
- Summary Bullet Narrative:     Pass (all four bullets are what+why, not label-only)
- Technical Decisions (AgDR):   Pass (this PR *is* the AgDR the trust-chain decision requires)
- Adopter Handbooks:            N/A (no handbook triggers — see below)
- Fallow Static Analysis:       N/A (diff is not JS/TS)

**AgDR-0115 — all load-bearing requirements met.**
- Chooses **Option D (accept residual)** explicitly, refined with a forge-native advisory check.
- Honest **A/B/C/D options table** with pros/cons for each; B rejected outright (reintroduces the stale-absolute-path class #1086 exists to fix), C named weak-alone, A named as the only trust-safe *direction*.
- **Load-bearing rail stated prominently** in its own "Load-bearing guardrail (record prominently — this is the rail, not a footnote)" section: never point `core.hooksPath` at a clone's TRACKED content.
- **Does NOT pre-commit Option A** — a dedicated "Not pre-committing to Option A" section records A as a *direction* only, gated behind an actual incident, and requires a forge-native fix to be compared first (survives re-clone, not `--no-verify`-bypassable).
- **Pins the #1087 risk-direction invariant** — "a block, or any non-zero/error exit, is the SAFE outcome; exit 0 (silent success) against a third-party clone" is the dangerous one.
- **AgDR number is 0115** — 0114 is reserved for the epic (not yet on `dev`; latest present is 0113), so no collision.

**`/handover` step 1.7 — advisory + fail-open, forge-agnostic.**
- Never blocks the flow: on any non-zero from `gh api .../protection` (whether "404 not protected" or a real auth/network failure) it sets `not protected (or could not be verified)`, prints one nudge, and continues to step 2 regardless. Note #25 and the step header both state advisory/fail-open/never-a-gate.
- **Dispatches on `tracker_kind`** (`gh` / `glab` / `none|custom|*`), the same pattern `tracker_review_submit` / `tracker_pr_merge` use — confirmed those three functions exist in `.claude/hooks/_lib-tracker.sh`. `custom`/`none` get a generic manual-check note, not a guessed API call.
- **Nudges only when protection is absent/unverifiable** — `protected` and both `skipped` variants print nothing beyond the step-10 summary line.

**`bin/install-git-hooks.sh` header fix — consistent, doc-only.**
- Every `/handover` mention in the file now says `/handover` does NOT invoke the installer (Background para, `--repo-dir` option, and the bottom note). No residual contradiction — I grepped all six mentions.
- Points at `.claude/skills/handover/SKILL.md`'s "1.5-clone" note and AgDR-0115.
- Confirmed the diff touches **only comment lines** (zero executable changes), so this overlaps cleanly with the epic's PR-1 that also edits this file.

**Gates.**
- shellcheck `--severity=warning` on the PR's `install-git-hooks.sh`: clean.
- markdownlint (repo's own `.markdownlint.json`) on AgDR-0115 + SKILL.md: 0 issues.
- `test_handover_clone_prompt.sh`: assertions are positive-only whole-file substring pins (`grep -qF`) with no negative/ordering checks — inserting step 1.7 cannot break any of the 18, matching the claimed 18/18.
- `test_install_git_hooks.sh`: the diff is comment-only and the test doesn't assert on the changed header text — the claimed 43/43 holds for a header-only change.

### Handbook Findings

None. The always-load handbooks were checked: the **blocking** `handbooks/architecture/migration-safety.md` has no trigger (its trigger #1 requires a migration file in the diff — this diff touches zero migration files); `clean-architecture-layers.md` and `commit-message-quality.md` have nothing to flag on a docs + shell-comment diff.

### Issues Found

None.

### Suggestions

- (nit, non-blocking) The `gh` branch of step 1.7 collapses "404 not protected" and a genuine auth/network failure into one status string. That is intentional and correctly documented (fail-open by construction), so no change is needed — noting only that if a future iteration ever wants to distinguish "definitely off" from "couldn't check," the `gh api` HTTP status is where that split would live.

### Verdict

**APPROVED**

---
Reviewed commit: `5dc1a341caa11d7d320f65c0ed829fa65a514b03`

PR #1117, opened 2026-08-01
feat(#1086): add blocking pre-commit protected-branch gate + fail-closed lib hardening (AgDR-0114)

me2resh:
## Summary

- **Adds the missing commit-time git-native control.** `.githooks/pre-commit` is new — it resolves the branch via `git symbolic-ref --short HEAD` and blocks (`exit 1`) a direct commit to a protected branch, with zero command-text parsing. Before this PR, `.githooks/` had a push-time control (`pre-push`, #1090) but no commit-time equivalent — a gap the issue's own review thread flagged as a real, unaddressed hole.
- **Hardens the shared protected-branch lib fail-closed, before it needs to be.** `_lib-protected-branches.sh` gains `is_protected_branch()` — a single predicate now shared by both git hooks. It fails CLOSED on two shapes the issue's review called out as MEDIUM: a malformed `.git.protected_branches[]` entry that makes the regex fail to evaluate (previously read as "not protected"; now treated as protected), and a renamed/missing `protected_branch_regex` function after a successful `source` (previously a silent allow; now a visible `WARN` + block). `.githooks/pre-push` is refactored onto this shared predicate too, so it inherits the fix instead of carrying its own copy of the same two bugs — verified against all 11 of its existing test cases with zero behaviour change.
- **Records why `block-main-push.sh` is relabelled, not demoted.** The issue's own comment thread had proposed demoting `block-main-push.sh`'s commit-half to advisory once a `pre-commit` hook existed — the same shape #1090 used for pushes. A Solution-Architect design pass + a Contrarian challenge found that would be a net reduction in real protection: `git push --no-verify` / `git commit --no-verify` bypass every git-native hook by design, and `block-main-push.sh` (a Claude Code PreToolUse hook, deciding before the shell ever runs) is the only layer immune to that. Separately, managed-project clones have **no** `core.hooksPath` coverage today — #1088 is deliberately unresolved, and #1087's own security review found auto-installing it against an untrusted clone unsafe — so `block-main-push.sh` is also the only layer covering every managed-project workspace. `AgDR-0114` records this: `block-main-push.sh` is honestly a **blocking backstop**, not the primary control any more, but it stays blocking. **`block-main-push.sh` itself is not touched anywhere in this PR** — no behaviour change, no code change.

This is **PR-1 of 2** for #1086 step 3. PR-2 migrates `block-main-push.sh` onto `_lib-protected-branches.sh` (removing its inline, duplicate protected-branch resolution) and updates its comments to the honest "blocking backstop" language — it does **not** demote it; AgDR-0114 is the ordering/authority reference for that PR.

## Testing

Real git, real sandboxes, no stubs — both new test files build isolated repos and assert on the actual commit/push outcome (exit code, HEAD movement, stderr message), the same philosophy as the existing `test_githooks_pre_push_protected_branch.sh`.

**`.claude/hooks/tests/test_githooks_pre_commit_protected_branch.sh` — 9 cases, 25 assertions, all PASS:**

| Case | Result |
|---|---|
| Commit on `main` | BLOCKED |
| Commit on `dev` | BLOCKED |
| Commit on a feature branch | allowed |
| Commit in a `cd`-reached **worktree** on a feature branch, primary checkout on `main` | allowed (no over-block) |
| Remediation text, `tracker.kind=gh` | names `gh pr create --base main --head <feature-branch>` |
| Remediation text, `tracker.kind=glab` | names `glab mr create --target-branch main` |
| Detached HEAD commit | allowed |
| `_lib-protected-branches.sh` missing | `WARN` + allowed (graceful degrade, old-clone case) |
| The REAL `bin/install-git-hooks.sh` sets `core.hooksPath` with no per-hook enumeration | `pre-commit` picked up anyway, blocks on `main` |

**Discriminating by construction, verified two ways:** every file in the "before" list didn't exist before this PR, so the suite trivially fails on a missing file. To prove it's the *behaviour*, not just the file, being tested: I swapped `.githooks/pre-commit` for a no-op `exit 0` stub and reran — **12 of 24 assertions correctly flipped to FAIL** (every BLOCKED case, both remediation cases, the installer-autopickup block, the lib-missing WARN). Restoring the real hook brings the suite back to 25/25 green.

**`.claude/hooks/tests/test_lib_protected_branches.sh` — new (no dedicated lib test file existed before), 16 assertions, all PASS:**

- Well-formed config: matches configured branches (including a regex entry), not a safe one.
- **Malformed regex (`"ma(in"`)**: first reproduces the OLD hand-rolled `command -v` + `grep -qE` shape and shows it **silently reads `main` as safe** (the bug) — then shows `is_protected_branch()` keeps `main` **protected** (fail closed), with a `WARN`, and that a totally unrelated branch is *also* fail-closed (one malformed entry poisons the whole composed alternation, not just its own alternative).
- **Renamed/missing `protected_branch_regex`**: same before/after shape — the OLD guard silently skips the check entirely; `is_protected_branch()` fails closed with a `WARN`.
- Default fallback (no config at all) still protects `main`/`master`/`dev`/`develop`.

**Full existing suite:** `bin/run-hook-tests.sh` — **132 PASS, 1 FAIL, 133 total.** The one failure is `test_block_merge_on_red_ci.sh` (2 cases about `glab`-registry forge dispatch in `block-merge-on-red-ci.sh` / `_lib-extract-pr.sh`) — neither file is touched anywhere in this PR, its last change was `6d7d3f9` well before this branch, and it's unrelated to protected-branch push/commit gating. Pre-existing on `dev`, not a regression from this PR. Every test that touches the files this PR changes — the full `test_githooks_pre_push_protected_branch.sh` suite (all 11 cases, unchanged behaviour) plus the two new suites above — is green.

`shellcheck --severity=warning` is clean on all five new/edited shell files. `markdownlint-cli2` is clean on the new AgDR.

Refs #1086

---

## Glossary

| Term | Definition |
|------|------------|
| `pre-commit` / `pre-push` hooks | Git-native hooks that run before a commit/push completes. Git invokes them with ground-truth state (the current branch for `pre-commit`; the resolved ref lines on stdin for `pre-push`) — no shell command text to parse. |
| `core.hooksPath` | Per-clone git config naming the directory git reads hooks from. Set once per clone (e.g. by `bin/install-git-hooks.sh`); not inherited by cloning. |
| Fail-closed | On an error or an unevaluable input, treat it as the *unsafe* case (here: "protected") rather than silently allowing. The opposite of fail-open. |
| Blocking backstop | A control that still blocks (non-advisory) but is not the *primary* mechanism relied on — here, `block-main-push.sh` remains blocking because it covers two gaps (`--no-verify`, uninstalled `core.hooksPath`) the newer git-native controls don't reach. |
| `is_protected_branch()` | The new shared predicate in `_lib-protected-branches.sh` both `.githooks/pre-push` and `.githooks/pre-commit` call — single source of truth for "is this branch protected", with the fail-closed hardening living in one place. |


me2resh (commented):
## Code Review: PR #1117 — Rex

**Commit:** `93bf66c7c3549f98e3977b2f6d8c6b5a33adab29`

**Verdict: APPROVE**

Rigorous pass warranted (core trust-chain: new blocking git-native gate + fail-closed hardening of the shared protected-branch lib). Reviewed via diff only — no checkout. Every load-bearing claim in the PR body was independently verified against the diff and `dev`.

### What this PR does
Adds `.githooks/pre-commit` (the missing commit-time git-native control), hardens `_lib-protected-branches.sh` fail-closed behind a new shared `is_protected_branch()` predicate, refactors `.githooks/pre-push` onto that predicate, and records AgDR-0114. `block-main-push.sh` is genuinely untouched — this is honestly PR-1 of 2.

### Verification results

- **AgDR-0114 is honest.** It records the DECIDED outcome (Option D: relabel `block-main-push.sh` a *blocking backstop*, keep it blocking — NOT demote). The Context section openly narrates the stale #1086 comment that proposed demotion (naming its never-committed file `AgDR-0114-commit-gate-moves-to-pre-commit-hook.md`, and stating it was verified absent from every branch/PR/`origin/dev`). Confirmed: `AgDR-0113` is the highest committed record on `dev`, no `0114` exists — the slot claim holds. The Options table (A/B/C/D) states real cons for each, and the reconciliation with AgDR-0104 ("backstop → advisory" only holds when a strictly stronger control exists to lean onto; here it doesn't, because server-side branch protection was rejected for this ticket and git-native hooks aren't universally live) is careful and correct.

- **`.githooks/pre-commit` is correct.** Blocking (`exit 1`); resolves the branch via `git symbolic-ref --short HEAD` (git ground truth, zero command-text parsing); detached HEAD → `exit 0` (nothing to protect); missing lib → visible `WARN` + `exit 0` graceful degrade, mirroring pre-push's identical case; forge-aware remediation dispatches on `tracker_kind` (gh / glab / generic fallback) with `_lib-tracker.sh` treated as an optional dependency that degrades to a forge-generic hint rather than failing the gate.

- **`is_protected_branch()` is genuinely fail-CLOSED.** Missing/renamed `protected_branch_regex` after a successful source → `WARN` + `return 0` (protected), not a silent `command -v` allow. `grep` exit-code discrimination is exactly right: `0`→protected, `1`→not protected, `>1`→`WARN` + protected — so a malformed `.git.protected_branches[]` entry (unbalanced-paren fragment poisoning the whole composed alternation) is treated as protected instead of being read as "no match". The two hardenings live in one place, so both callers inherit them by construction.

- **pre-push refactor preserves behavior.** The inline `command -v` guard + raw `grep -qE` is swapped for `is_protected_branch`; the process-substitution loop, deletion detection, and message are unchanged. For the 11 existing cases (all supply a well-formed lib) external behavior is identical; the only change is that the previously fail-OPEN missing-function/malformed paths now fail CLOSED — a strict hardening, as intended.

- **Tests are discriminating.** `test_lib_protected_branches.sh` reproduces the OLD buggy caller shape side-by-side (`OLD:SAFE`/`OLD:SILENTLY-SKIPPED` vs `NEW:PROTECTED`), pinning both MEDIUMs at the unit level, and asserts the unrelated-branch poisoning. `test_githooks_pre_commit_protected_branch.sh` uses real git sandboxes and asserts on actual exit code + HEAD movement + stderr, incl. the worktree no-over-block case and the real-installer autopickup case — the stub-swap flip claim is structurally credible given those behavioral assertions.

- **The 1 full-suite failure is pre-existing, not a regression.** Confirmed `test_block_merge_on_red_ci.sh` and its SUTs (`block-merge-on-red-ci.sh`, `_lib-extract-pr.sh`) are NOT in this PR's changed-file set; last touched by `cb59b94` (2026-07-22), well before this branch.

- **PR hygiene:** narrative Summary (each bullet says what + why), `Refs #1086`, Glossary present, PR-1-of-2 clearly stated. `block-main-push.sh` confirmed absent from the changed-file list (the only match is the AgDR filename).

### Checklist
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass (discriminating, real-git)
- Security / trust-chain: Pass (fail-closed verified in both callers)
- AgDR: Pass (AgDR-0114 present, honest, correct slot)

### Notes (non-blocking)
- `nit:` `regex=$(protected_branch_regex)`'s command-substitution exit code isn't checked, but the resolver has a default fallback and the downstream `grep` rc handling fails closed anyway — defense-in-depth already covers it.
- shellcheck/markdownlint clean is claimed in the PR body; the code reads clean inline (SC1090/91 + SC2034 disables documented). I did not independently re-run the linters against the branch (no-checkout review), but found nothing that would fail them.

Solid, careful trust-chain work. Approving.

---
Reviewed by Rex (Code Reviewer). Reviewed commit `93bf66c7c3549f98e3977b2f6d8c6b5a33adab29`.

atlas-apex (approved):

me2resh (commented):
## Security Review: PR #1117 — pre-commit protected-branch gate + fail-closed lib hardening (AgDR-0114)

**Reviewed commit**: `93bf66c7c3549f98e3977b2f6d8c6b5a33adab29`
**Reviewer**: Hakim (Security Auditor) · trigger: trust-chain change (`.claude/hooks/**`, `.githooks/**`)

### Verdict: SECURE

This is a genuine strengthening of the protected-branch trust chain. I read all four load-bearing files in full and ran the fail-closed claims adversarially against the actual lib code (under `bash`, matching the hooks' own `#!/bin/bash` — not zsh, which the lib itself warns breaks path resolution). Every claim held.

### What I verified

**1. The pre-commit gate genuinely blocks, from git ground truth.**
`.githooks/pre-commit` resolves the branch via `git symbolic-ref --short HEAD`, calls `is_protected_branch`, and `exit 1`s. There is zero command-text parsing — no `cd`-prefix detection, no heredoc/quoting/decoy surface (the AgDR-0113 bug class) exists at this layer, because there is no command string to misparse. Git invokes the hook from inside the checkout the commit lands on, so the worktree case is correct by construction (no over-block), confirmed by the suite's case 4. Detached HEAD and non-repo both `exit 0` correctly (nothing to protect).

**2. The fail-closed hardening is real — I could not construct a config that silently unprotects `main`.**
Adversarial configs run against the real `is_protected_branch()`:

| Config `.git.protected_branches[]` | `main` result |
|---|---|
| `["main","ma(in","dev"]` (unbalanced paren → composed regex `main\|ma(in\|dev` → `grep -qE` exit 2) | PROTECTED + WARN — **and every other branch also fails closed** (one malformed entry poisons the whole alternation) |
| `["ma[in","main"]` (unbalanced bracket → regex error) | PROTECTED + WARN |
| `protected_branch_regex` renamed/removed after source | PROTECTED + WARN (`command -v` guard → fail closed) |
| no config at all | `main`/`master`/`dev`/`develop` PROTECTED, feature SAFE (default fallback) |

The exit-code discrimination is exactly right: `grep -qE` returns 0=match, 1=no-match, >1=error — and only `1` is treated as "not protected"; every `>1` (regex error) fails CLOSED with a visible WARN. This is a real fix to two MEDIUMs that were fail-OPEN in the pre-existing hand-rolled `command -v` + raw `grep -qE` shape.

**3. No new bypass in the pre-push refactor — it is strictly stronger.**
The old pre-push guarded on `command -v protected_branch_regex` and `[ -n "$PROTECTED_REGEX" ]`, both of which **failed OPEN** (skipped the check → allowed the push) when they missed. The refactor moves that logic into `is_protected_branch`, which **fails CLOSED** on the same shapes. The security-critical loop invariants are all preserved: process-substitution (not a pipe) so `exit 1` terminates the script not a subshell, deletion detection (`LOCAL_SHA = ALL_ZERO_SHA`), the `refs/heads/*`-only filter, and the single-ref-line handling. Behaviour for the 11 existing cases is unchanged except where it now fails closed instead of open. Branch names flow into `grep` as the *subject*, never interpolated into the *pattern*, so a branch name with regex metachars (`ma.n`) cannot inject — confirmed SAFE against `^(main)$`.

**4. AgDR-0114 is sound and reduces no protection.**
`block-main-push.sh` is **not touched anywhere in this PR** — zero behaviour change, so nothing that blocks today stops blocking. Keeping it blocking (rather than demoting to advisory) is correctly justified: it is a PreToolUse hook that decides before the shell runs, making it the only layer immune to `git commit/push --no-verify`, and the only layer covering managed-project clones where `core.hooksPath` is deliberately not installed (#1088). The record correctly narrows AgDR-0104's "backstop → stop calling it enforcement" to the case where a *strictly stronger* control exists to lean on — which is not the case here.

### Informational residuals (non-blocking, all documented and pre-existing)
- **Lib-missing → fail OPEN** at the git layer in both hooks (`WARN` + `exit 0`). Intentional graceful-degrade for old clones; visible, not silent; backstopped by `block-main-push.sh`. Net-neutral vs. before this PR (there was no hook at all).
- **`--no-verify` bypasses both git hooks** — inherent to git hooks, documented, and backstopped by the PreToolUse layer. Not a defect in this PR.
- **A valid-but-permissive/typo'd entry** (e.g. `["mai"]`) yields a *valid* regex that genuinely doesn't match `main`, so `grep` returns a clean exit 1 and the branch reads SAFE. Fail-closed cannot catch this (it only triggers on evaluation *errors*, not on correct no-matches). This is operator misconfiguration, identical before and after this PR, and out of scope — flagged only for completeness.

None of these are introduced or worsened by this PR. Tests are real-git sandboxes with discriminating-by-construction proofs (stub-swap flips 12/24 assertions). Clean.

📌 Reviewed commit: `93bf66c7c3549f98e3977b2f6d8c6b5a33adab29`

Issue #1064, opened 2026-07-29
[Bug] The Lean ceremony tier is unreachable — the merge gate requires a Rex marker regardless of tier


atlas-apex:
## Affected

`.claude/rules/right-size-ceremony.md` (the Lean tier) + `.claude/hooks/block-unreviewed-merge.sh` (the marker requirement) + `.claude/hooks/auto-code-review.sh` (the unconditional banner)

## Given / When / Then

**Given** `right-size-ceremony.md` defines a **Lean** tier for "docs / config-text, small, reversible" changes, whose prescribed ceremony is explicitly *"inline check + one approval, **no review sub-agent**"*

**When** a Lean change is delivered the way the framework delivers everything else — on a branch, through a PR

**Then** `block-unreviewed-merge.sh` refuses the merge until a `*-rex.approved` marker exists at the PR's HEAD, and that marker is only legitimately produced by a real code-reviewer sub-agent pass — so the Lean tier's defining instruction cannot be followed. Expected: a documented tier should be reachable through the framework's own delivery path.

## Why this is a real gap and not a misreading

The rule is unambiguous that Lean means no sub-agent. The gate has no tier awareness — it applies the same marker requirement to a one-line copy edit in a static HTML page as to a merge-gate hook change. There is no configuration, path allowlist, or marker kind that expresses "this change was checked inline, at the Lean tier".

So the Lean floor exists in prose and is unreachable in practice. `right-size-ceremony.md` itself says the rule was created because the framework "is optimised for compliance, not shipping" — this is that same failure mode surviving inside the fix for it.

## Observed cost

One session, one static HTML page in a managed project, copy and presentation only, no application code and no runtime behaviour:

| | |
|---|---|
| PRs | 4 |
| Code-review sub-agent runs | 6 |
| Review-agent tokens | ~850k |
| Review wall-clock | ~41 min |

Every one of those runs was mechanically required to produce a marker the gate demanded.

## The compounding factor: SHA-bound markers make iteration expensive

Markers bind to the PR's HEAD, correctly — new commits must invalidate a prior review. But it means **acting on a review finding costs a full second review**, even when the finding is a comment typo or a one-word wording fix.

Two of the four PRs above went two rounds for exactly this reason. The reviews were good and the findings were real; it was the *iteration* that was expensive, not the reviewing. At the Heavy tier that is the correct trade. At the Lean tier it is the whole cost.

## Directions (not prescribing — this needs a deliberate call)

[...]

PR #1118, opened 2026-08-01
chore(#1057): advisory nudge for trust-chain CONTROL adversarial-test coverage (AgDR-0117)

me2resh:
## Summary

- **Rejects the blocking CI meta-gate #1057 asked for.** #1015's sixth acceptance criterion — "CI fails when a trust-chain hook has no paired adversarial test" — never shipped and was split into #1057. A Solution-Architect design pass plus a Contrarian (Naqid) challenge found the gate has **zero current value** (all four CONTROL-class merge-gate hooks already carry adversarial tests), is **defeatable by construction** (a hand-maintained trust-chain list is beaten by omission; a "naming convention + marker comment" definition of "adversarial test" is beaten by a strawman test that asserts nothing), and would **duplicate #871**'s stronger live-conformance job, which already proves these hooks fire against a real credentialed agent turn. AgDR-0104's meta-gate deferral (narrowed by AgDR-0111 to "a new control is added to the trust chain") has not had its trigger fire, so this isn't overdue work — it's a second look that confirms the earlier call.
- **Ships a lightweight advisory nudge instead** — `.claude/hooks/nudge-control-adversarial-test.sh`, wired to `PreToolUse` on `gh pr create`. When a PR's diff touches a `.claude/hooks/*.sh` file carrying its own `# CLASS: CONTROL` header, it prints one reminder line pointing at AgDR-0104/AgDR-0117 and **always exits 0** — same non-blocking shape as `detect-role-trigger.sh` and `check-upstream-drift.sh`.
- **Derives the CONTROL set from the hooks' own headers, not a config list** — the exact fix for the Contrarian's list-omission finding. A hook can only escape the nudge by removing its own `# CLASS: CONTROL` header, which is a loud, reviewable, self-incriminating diff rather than a silent gap in a separate file.
- **Records the decision as AgDR-0117** (`docs/agdr/AgDR-0117-trust-chain-meta-gate-stays-advisory.md`), including the options table (hard gate / label-derived gate / advisory nudge [chosen] / stay-deferred) and an explicit note that AgDR-0104's deferral is **not discharged** by this PR — it stays open under AgDR-0111's narrowed, still-unfired trigger.

## Testing

[...]

PR #1119, opened 2026-08-01
docs(#1064): make the Lean tier reachable (accept+document, reduced-scope Rex)

me2resh:
## Summary

- **Records the maintainer's decision on #1064 as AgDR-0116** — Direction 3 (accept & document), layered with Option 4 (reduced-scope Rex). The Lean tier's own rule promised "no review sub-agent," but `block-unreviewed-merge.sh` requires a Rex marker on **every** PR regardless of tier — a documented tier the framework's own delivery path couldn't satisfy. This decision says why the gate stays unconditional (it's a control that reads structured state and structurally can't inspect a diff to pick a tier), rejects the two gate-relaxing options (a self-issued attestation marker, a path-class exemption) as new forgeable surfaces in the shape of #843/#1011/#1026, and records the Contrarian-surfaced honesty note: for a pure `.md` diff, Lean and Standard were already mechanically identical before this change (only Rex auto-fires on docs; Tariq/Security/UI design don't).
- **`block-unreviewed-merge.sh` is untouched — zero diff.** The merge gate stays exactly as strict as it was; this PR does not add a new marker kind, a path-exemption branch, or any tier-conditional logic to the gate. Verified with `git diff --stat` (empty) and pinned by the new test suite (case 1–3).
- **`right-size-ceremony.md`'s Lean row stops promising what the gate forbids.** It now reads "no role chain — no design/security/architecture review sub-agents — one lightweight Rex pass + the human merge nod," and the "When to apply" / self-check sections were updated to match. Rails 1 & 2 (security/trust-chain never Lean; ambiguity rounds up) are kept **verbatim** — this PR only touches the Lean-tier prescription, never the rails.
- **`auto-code-review.sh`'s banner gets a tier-aware note without softening the #843 fix.** "Rex runs on every PR" stays completely unconditional and crisp for both banner readers (build sub-agent / orchestrator); the new note clarifies that what actually varies by tier is the **role chain** around Rex (Security Auditor / Solution Architect / UI Designer), which already only fires on its own diff/path triggers.
- **`code-reviewer.md` gains "Reduced-Scope Review" (Option 4).** On a diff that clears a strict 5-condition eligibility bar (docs/config-text only, small, trivially reversible, no behavior change, **and** no security/trust-chain/migration path touched — a single match disqualifies the whole diff), Rex may run a focused correctness pass instead of the full architecture/quality/testing/performance review. The required outputs never shrink: the posted review, the exact-SHA marker, the Glossary/AgDR/handbook checks are unchanged at every tier — only review *depth* varies, which is the actual lever behind the token/latency savings the issue's Observed Cost section documented (~850k tokens / ~41 min across 4 docs-only PRs).
- **New test — `test_lean_tier_reachable.sh`, 19 cases** — pins AC#5 (a security/trust-chain/migration path can never reach the Lean path) at three independent points, each with a *discriminating* pair (a check plus a companion case proving the check would actually catch a regression, not just pass tautologically): the merge gate's Rex-check block carries no lean/tier bypass token (cases 1–2), `code-reviewer.md`'s eligibility bar lists every rail-1 path pattern (cases 4–5), and a fail-closed path matcher mirroring the documented rail correctly rejects every security/trust-chain/migration path class even mixed into an otherwise all-docs diff (cases 11–18).

## Testing

```
bash .claude/hooks/tests/test_lean_tier_reachable.sh
# 19 passed, 0 failed

shellcheck --severity=warning .claude/hooks/auto-code-review.sh .claude/hooks/tests/test_lean_tier_reachable.sh
# clean, exit 0

npx --yes markdownlint-cli2 docs/agdr/AgDR-0116-lean-tier-reachable.md \
  .claude/rules/right-size-ceremony.md .claude/agents/code-reviewer.md
# 0 issues in 3 files

bash bin/run-hook-tests.sh
# PASS=132 FAIL=1 SKIP=0 TOTAL=133
# the one failure (2 sub-cases) is test_block_merge_on_red_ci.sh's
# "glab-registered project" wrapper-dispatch cases — a file this PR does
# not touch, about CI-status forge dispatch via the project registry,
# unrelated to the Lean-tier/AgDR-0116 change. Confirmed pre-existing and
# unrelated by running it in isolation and inspecting the failure (registry
# dispatch resolving to gh instead of glab for a glab-registered project —
# nothing to do with Rex, tiers, or review scope).
```

`git diff --stat -- .claude/hooks/block-unreviewed-merge.sh` is empty — the merge gate is not part of this diff.

Refs #1064

## Glossary

| Term | Definition |
|------|------------|
| Lean tier | `right-size-ceremony.md`'s lightest ceremony class — small, reversible, no-behavior-change diffs. Previously prescribed "no review sub-agent"; this PR redefines it to "no role chain, reduced-scope Rex." |
| Rail 1 | The right-size-ceremony rule's non-negotiable: security / trust-chain / migration diffs never go Lean, regardless of size. Unchanged by this PR; also now stated explicitly in Rex's own reduced-scope eligibility bar. |
| Rail 2 | The rule's other non-negotiable: ambiguity always rounds up to more ceremony, never down. |
| Control vs. backstop | AgDR-0104's distinction — a control reads structured state to make a decision (e.g. `block-unreviewed-merge.sh` comparing a marker SHA to the forge-reported HEAD); a backstop is advisory text-matching. The merge gate is a control, which is why it can't also be the thing deciding a diff's tier. |
| Reduced-scope Rex | Option 4 — Rex running a focused correctness pass instead of the full deep review, on a diff that clears the 5-condition eligibility bar. Changes review *depth*, never whether a review happens or what it's required to output. |
| Role chain | The set of role sub-agents (Security Auditor, Solution Architect, UI Designer) layered around Rex on Standard/Heavy diffs. This is what the Lean tier actually suppresses — Rex itself always runs. |

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -70,6 +70,34 @@ When the `apexyard-search` MCP is connected, **prefer `mcp__apexyard-search__sea
 
 It also lowers review token cost (targeted semantic excerpts vs. broad `grep` + full-file reads). **Graceful-degrade:** if the MCP server is absent the tool simply isn't available — fall back to `grep`/`Glob`/`Read` with no change in behaviour (same pattern as `search_docs`, `/handover`, and `/code-review`). Adopters who don't run the premium MCP are unaffected.
 
+## Reduced-Scope Review — Lean-tier diffs (Option 4, AgDR-0116)
+
+Per `.claude/rules/right-size-ceremony.md`, a **Lean-tier** diff still requires a Rex pass — the merge gate (`block-unreviewed-merge.sh`) requires the `*-rex.approved` marker on EVERY PR, regardless of tier, unconditionally, because it is a CONTROL that reads structured state (a marker vs. the forge-reported HEAD) and structurally cannot itself inspect a diff's content to decide a tier. What changes for a Lean diff is the **depth** of your pass, never whether one happens.
+
+### Eligibility — ALL of these must hold, or run the full review
+
+1. **Path class.** The diff touches only docs / comments / config-text — `.md`, `.txt`, issue/PR templates, non-executable config values. NOT `.yml`/`.yaml` workflow logic, NOT scripts of any kind, NOT source code in any language.
+2. **Size.** Small — a rough guide is under ~50 changed lines. Use judgment, not a hard cutoff; a 200-line prose rewrite can still be Lean, a 10-line `.md` change that alters a documented gate's behavior might not be.
+3. **Reversibility / behavior.** Trivially reversible in one revert, with no behavior change — no code path, no runtime logic, no schema, no CI step, nothing that executes differently as a result of this diff.
+4. **Rail 1 (non-negotiable — mirrors `right-size-ceremony.md`'s rail 1 exactly; security / trust-chain / migration never goes Lean).** The diff does NOT touch ANY of:
+   - The trust chain: `.claude/hooks/**`, `.claude/settings.json` (me2resh/apexyard#777)
+   - `**/auth/**`, `**/crypto/**`, `**/secrets/**`, `.env*`
+   - A migration path: anything under `**/migrations/**`, `**/migrate-*.{ts,js,py,sql}`, `prisma/schema.prisma`, `prisma/migrations/**`, `src/migrations/*.{ts,js}`, `alembic/versions/*.py`, `db/migrate/*.rb` (the same defaults `require-migration-ticket.sh` matches; check `.migration_paths` in `.claude/project-config.json` for an adopter override too)
+   - **A single file matching ANY of these disqualifies the ENTIRE diff from reduced scope**, even if every other file in the diff is plain prose. Do not average across files — one match rounds the whole PR up.
+5. **Rail 2 (mirrors `right-size-ceremony.md`'s rail 2 exactly — ambiguity rounds up).** If you are unsure whether ANY of conditions 1–4 holds, run the full review. The tolerated failure is an occasional full review on a diff that could have been reduced-scope — never a reduced-scope pass on something that needed the full one.
+
+### What "reduced scope" means
+
+When, and only when, all five conditions above hold, you may skip the deep line-by-line pass over architecture, performance, and test-design nuance, and instead run a FOCUSED pass:
+
+- A correctness read of the actual prose/config change — does it say what it means to say, is it internally consistent, does it match the codebase state it describes.
+- The **mandatory checks that never shrink, at any tier**: PR description quality + Glossary (§ 6), AgDR detection (§ 7, blocking), handbook discovery (§ 8), and the approval-marker mechanics (unchanged — same file, same exact-SHA format, same "APPROVED only" rule).
+- Skip: the Architecture & Design, Code Quality, Testing, Performance checklist items (§§ 1–5) — there is no code here for them to apply to. If any of those sections turns out to have something to say about this diff, that is itself a sign eligibility condition 1 or 3 was wrong — stop and fall back to the full review rather than force a finding into the reduced-scope format.
+
+**Reduced scope changes DEPTH, never the required OUTPUTS.** You still post the human-visible review via `tracker_review_submit`, and you still write the `*-rex.approved` marker on an APPROVED verdict, in the exact same format, at the exact same gate, as every other review. State `(reduced-scope pass — Lean tier, AgDR-0116)` in your review body so the record is honest about which pass ran, and set the `**Scope**` line in the Output Format (below) to `Reduced-scope`.
+
+If any of conditions 1–5 fails, this section does not apply — run the full review from § 1 onward, exactly as you would for a Standard or Heavy diff.
+
 ## Review Checklist
 
 ### 1. Architecture & Design
@@ -779,6 +807,7 @@ Report the failure in plain text with the exact command the caller needs to run.
 ## Code Review: PR #{number}
 
 **Commit**: `{headRefOid}`  ← REQUIRED — always include this.
+**Scope**: `[Full / Reduced-scope — Lean tier, AgDR-0116]`  ← REQUIRED — see § "Reduced-Scope Review".
 
 ### Summary
 [Brief summary of what the PR does]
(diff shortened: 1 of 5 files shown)


me2resh (commented):
## Code Review: PR #1119

**Commit**: `2bdabbacd3a1cc58121bf09afde24b634c388a07`
**Scope**: Full

### Summary

Records AgDR-0116 for issue #1064 (the Lean ceremony tier was unreachable because `block-unreviewed-merge.sh` requires a Rex marker on every PR regardless of tier). The chosen resolution is Direction 3 (accept & document) layered with Option 4 (reduced-scope Rex): the merge gate is left **completely untouched**, `right-size-ceremony.md`'s Lean row is redefined so it stops promising the unfollowable "no review sub-agent", `auto-code-review.sh` gets a tier-aware banner note, and `code-reviewer.md` gains a rail-1/rail-2-gated reduced-scope review mode. This is gate-adjacent trust-chain work and I reviewed it at the Heavy tier.

### Checklist Results
- Architecture & Design:     Pass — the control-vs-tier-classifier separation (AgDR-0104 framing) is the correct reason the gate can't itself pick a tier; the ceremony reduction is confined to the role chain + Rex depth, never the gate.
- Code Quality:              Pass — shellcheck clean; the new test is discriminating, not tautological.
- Testing:                   Pass — 19/19, with paired discriminating cases and fail-closed rail-1 coverage.
- Security:                  Pass — the load-bearing check. See below.
- Performance:               N/A — docs + hook-banner + agent-instruction change; no runtime path.
- PR Description & Glossary:  Pass — narrative Summary, `Refs #1064`, full Glossary.
- Summary Bullet Narrative:  Pass — every bullet states what changed AND why it matters.
- Technical Decisions (AgDR): Pass — this PR *is* the AgDR (0116); trust-chain-adjacent change is documented exactly as rail 1 requires.
- Adopter Handbooks:          N/A — the one blocking handbook (`handbooks/architecture/migration-safety.md`) does not fire (no migration files in the diff); advisory always-load handbooks have no applicable source in this diff.
- Fallow Static Analysis:     N/A — no JS/TS files.

### The load-bearing security check — the merge gate is not weakened

Verified from three independent angles, all green:

1. **Zero diff on `block-unreviewed-merge.sh`.** `git diff --stat origin/dev...HEAD -- .claude/hooks/block-unreviewed-merge.sh` is empty; the file is not among the 5 changed files. No new marker kind, no path-exemption branch, no tier-conditional logic was added to the gate. A Rex marker is still required on every PR unconditionally.
2. **Rail 1 preserved verbatim in the rule.** `right-size-ceremony.md`'s "Security and trust-chain never go Lean." and "Ambiguity rounds up." are both outside the diff hunks and present unchanged (test case 8 pins both exact sentences).
3. **Rail 1 mirrored into Rex's own eligibility bar.** `code-reviewer.md` § "Reduced-Scope Review" condition 4 lists the full trust-chain / auth-crypto-secrets / migration path set, and states that a single matching file "disqualifies the ENTIRE diff" — with condition 5 (rail 2) falling back to the full review on any ambiguity. Reduced scope changes review *depth* only; the posted review, the exact-SHA marker, and the Glossary/AgDR/handbook checks never shrink.

### AgDR-0116 quality

- Records Direction 3 + Option 4 (chosen), and explicitly **REJECTS** Option 1 (self-issued attestation marker) and Option 2 (path-class exemption) as new forgeable gate-relaxing surfaces in the shape of #843/#1011/#1026 — the right call, and the deferral conditions (forge-derived tier, fail-closed, rail 1 inside the gate, paired adversarial bypass test) are the correct constraints if they're ever revisited.
- The Contrarian honesty note is present and lands the sharp point: for a pure `.md` diff, Lean and Standard were already mechanically identical (only Rex auto-fires on docs; Tariq / Security / UI design don't), so the tier delta is narrower than "Lean = less ceremony" implies. Recording that rather than overselling the win is exactly right.
- Number 0116 is unique — dev tops out at 0113; in-flight PRs #1117 and #1118 hold 0114 and 0117 respectively; 0116 is unclaimed.

### The `auto-code-review.sh` banner does NOT reintroduce the #843 dual-reader ambiguity

The two clearly-delimited reader branches (BUILD-CLASS SUB-AGENT / ORCHESTRATOR) are untouched, and "Rex runs on EVERY PR, full stop" stays unconditional and crisp for both. The new CEREMONY TIER note sits after both branches, is additive, reinforces "never a bypass", and routes the depth decision to Rex's own doc — it introduces no new instruction a build sub-agent couldn't comply with. The #843 fix's dual-reader clarity is preserved.

### Issues Found
None blocking.

### Suggestions
- `nit:` (test-helper precision, non-blocking) In `test_lean_tier_reachable.sh`, the `is_lean_eligible_pathset` helper's prisma/alembic-specific arms (`*/prisma/schema.prisma`, `*/alembic/versions/*.py`) require a leading path segment, so root-level inputs like `prisma/schema.prisma` (case 15) and `alembic/versions/0007_add_index.py` (case 16) are actually rejected by the generic `*)` non-docs catch-all rather than the labelled branch. The fail-closed guarantee still holds robustly (any non-`.md`/`.txt` path is rejected), so this is purely cosmetic — but if someone later adds a root-level prisma case expecting the prisma-specific arm, the "not-eligible: non-docs path" label would mislead. Consider adding a leading-`*/`-optional alternative (e.g. `prisma/schema.prisma|*/prisma/schema.prisma`) so the reason labels match intent.

### Verdict
**APPROVED**

The merge gate is provably unweakened (zero diff, unconditional marker requirement intact), rail 1 is preserved verbatim in the rule and mirrored + tested fail-closed in Rex's eligibility bar, the AgDR is honest and well-reasoned, and the banner keeps its #843 dual-reader safety. The single nit is a cosmetic test-label imprecision that does not affect the fail-closed property.

---
Reviewed commit: `2bdabbacd3a1cc58121bf09afde24b634c388a07`

me2resh (commented):
## Security Review: PR #1119

**Commit**: `2bdabbacd3a1cc58121bf09afde24b634c388a07`

### Verdict: SECURE

This PR redefines the Lean ceremony tier and adds a reduced-scope Rex mode. I audited it against the one question that matters here — does it weaken the merge gate or let a security / trust-chain / migration change slip through with less scrutiny? It does not. The ceremony reduction lives entirely in review *depth* and role-chain suppression; every mechanical control that protects a risky change is untouched.

### Checklist Results
- Merge gate not weakened: Pass
- Rail 1 fails closed (security/trust-chain/migration never Lean): Pass
- Banner has no bypass / no #843 marker-impersonation ambiguity: Pass
- No security review silently dropped: Pass

### What I verified

**1. Merge gate is byte-identical to `dev` — not just "not in the file list".**
The harness had the PR HEAD checked out, so I could diff directly: `git diff --stat origin/dev -- .claude/hooks/block-unreviewed-merge.sh` is **empty**. The Rex-marker check is still unconditional — `if [ ! -f "$REX_APPROVAL" ]; then … exit 2` with no tier/lean/skip guard around it, and the SHA-match check is likewise unconditional. A Rex marker is still required to merge on **every** PR regardless of tier. Test cases 1-3 pin this with a discriminating assertion (case 2 injects a synthetic `LEAN_TIER` bypass and confirms the same detector catches it, so case 1 can't be a tautology).

**2. Rail 1 fails closed for every risky path — and the reduced-scope mode structurally cannot downgrade a security review anyway.**
`code-reviewer.md`'s eligibility bar (condition 4) explicitly disqualifies `.claude/hooks/**`, `.claude/settings.json`, `**/auth/**`, `**/crypto/**`, `**/secrets/**`, `.env*`, and the full migration path set — and states that **a single matching file disqualifies the ENTIRE diff**, even if every other file is prose. Condition 5 (rail 2) rounds any ambiguity up to the full review. Belt-and-braces: those paths are also excluded by condition 1 ("NOT scripts of any kind", "NOT `.yml`/`.yaml` workflow logic"), and the test's mirror matcher is **allowlist-based** — only `.md`/`.txt` pass, everything else hits the `*)` reject arm — so it's fail-closed by construction. Test cases 4-6 assert the doc lists every rail-1 pattern (case 5 strips the list and confirms the detector then fails), and cases 11-18 assert rejection of every security/trust-chain/migration class mixed into an otherwise all-docs diff, with a passing control (case 18). All 19 cases pass locally; shellcheck clean.

The deeper reason this is safe: reduced scope changes **depth only**. Even in the worst case where Rex mis-classified a trust-chain diff as Lean-eligible, (a) the merge still requires the marker + human nod, and (b) the Security Auditor still fires independently — `detect-role-trigger.sh` (untouched) activates the auditor on `.claude/hooks/**`, auth, crypto, secrets, and `.env*` via its own PreToolUse hook, **not** routed through Rex. The reduced-scope mode has no wiring that could suppress it.

**3. Banner keeps the #843 dual-reader fix intact; the new note adds no bypass.**
`auto-code-review.sh`'s "Rex runs on EVERY PR, full stop" stays completely unconditional for both readers. The build-sub-agent branch still says do NOT review and do NOT write any `*-rex.approved` marker; the orchestrator branch still routes through `/code-review`. The new CEREMONY TIER note is purely additive, appended after both branches, and reinforces "still gets a Rex pass — in REDUCED SCOPE — never a bypass." It never tells any reader to self-review or self-mark, so it introduces no marker-impersonation ambiguity.

**4. The AgDR honesty claim is accurate on the security substance.**
No security review is silently dropped: every Security-Auditor-triggering path is rail-1-disqualified from Lean and none are plain `.md`, so a Lean diff never had a security review to drop. The two gate-relaxing alternatives (self-issued attestation marker; path-class exemption) were correctly identified as new forgeable surfaces in the #843/#1011/#1026 shape and rejected — the chosen path adds **zero** new gate surface.

### Findings

- **LOW / informational (non-blocking, documentation nit).** AgDR-0116's honesty note states "Tariq, the Security Auditor, and the UI design gate do not auto-fire on `.md` paths at all — none of their trigger globs match plain markdown." The Solution Architect's globs *do* include markdown design artifacts (`*prd*.md`, `*technical-design*.md`, `docs/agdr/*migration*.md`). This is harmless — a design-artifact `.md` is a **Heavy** class and never Lean-eligible, and it's the architecture review, not the security review — so it changes nothing about the security posture. Worth a one-line tightening ("plain prose markdown, not design-artifact markdown") if the file is touched again, but not a reason to hold this PR.

### Recommendation
Merge-ready from a security standpoint. Rails 1 & 2 are preserved verbatim, the merge gate is unconditional and untouched, and the reduced-scope mode is structurally incapable of weakening the review of a security / trust-chain / migration change.

---
📌 Reviewed commit: `2bdabbacd3a1cc58121bf09afde24b634c388a07`

atlas-apex (approved):

PR #1120, opened 2026-08-01
feat(#1086): migrate block-main-push onto the shared lib, add --no-verify block, relabel as blocking backstop

### LINKED ISSUE #1081: [Bug] validate-branch-name.sh's is_tag_push exemption can hide a non-conforming push behind decoy tag evidence

me2resh:
## Summary

**PR-2 of 2** for the #1086 "relabel + keep blocking" epic (PR-1 was #1117, merged — it shipped `.githooks/pre-commit` + `is_protected_branch()`). This PR finishes step 3 per [AgDR-0114](docs/agdr/AgDR-0114-block-main-push-honest-naming-blocking-backstop.md): **keep `block-main-push.sh` blocking**, migrate it onto the same shared lib the git-native hooks already use, close the one gap it was still missing, and stop the framework's own docs from overstating what the hook is.

- **Removed the duplicate protected-branch resolution.** `block-main-push.sh` had its own inline copy of the three lines `.claude/hooks/_lib-protected-branches.sh` already implements — the "duplicate that has to be trusted to agree" pattern [AgDR-0113](docs/agdr/AgDR-0113-heredoc-stripper-additive-only.md) names as this file family's dominant defect generator. It now sources the lib and calls `is_protected_branch()`, so this file, `.githooks/pre-push`, and `.githooks/pre-commit` all resolve the *same* protected-branch list from *one* place, and this file inherits the lib's fail-closed hardening (a malformed `.git.protected_branches[]` entry, or a renamed/missing lib function, is now treated as protected here too, not just in the two git-native hooks).
- **Added the `--no-verify` note — the one gap with no git-native replacement.** `--no-verify` bypasses `.githooks/pre-push` and `.githooks/pre-commit` by git's own design; it does *not* bypass `block-main-push.sh`, because this is a Claude Code PreToolUse hook that runs before the shell (and therefore before git) ever executes. That protection already existed implicitly (the hook never looked at the flag one way or the other), but nothing explained it — an operator seeing a block after passing `--no-verify` had no way to know the flag "didn't work" on purpose. Every BLOCKED message now names this explicitly when `--no-verify` is present on the command.
- **Kept every existing `exit 2` site — none demoted.** This is the load-bearing part of AgDR-0114: the git-native hooks are objectively stronger on *which ref* (no command-text parsing at all), but they can't cover `--no-verify` or managed-project clones that never installed `core.hooksPath` (#1088, still open). Demoting `block-main-push.sh` would remove real coverage with nothing replacing it for those two cases, so this PR only relabels — it never turns a block into a warning.
- **Relabelled honestly.** The hook's header now states plainly that `.githooks/pre-push` / `.githooks/pre-commit` are *the* protected-branch controls and this file is a blocking *backstop*. `git-conventions.md` and AgDR-0104 (the "control vs backstop, honestly named" record this whole epic traces back to) are updated to match — AgDR-0104 also gets a short "Update" section recording this as its first executed application to a new hook family.
- **Fixes #1081** — a real, separate bug in `validate-branch-name.sh`: decoy tag-push evidence sitting in prose a heredoc's stripper correctly declines to strip (an unconfirmed delimiter) made `is_tag_push` return true for the *whole command*, so the hook exited 0 before ever validating a real, non-conforming push sitting right after it. Fixed the same way #1075 fixed `block-main-push.sh`'s own decoy-ref hijack: walk every push-shaped occurrence individually and only trust the tag-exemption for an occurrence that carries its own tag evidence — a bare, evidence-less occurrence (the real push hiding behind the decoy) forces validation instead of a silent skip. Kept in its own commit-worthy section of the diff, isolated from the block-main-push.sh work.

## Testing

Full affected suite is green (`shellcheck --severity=warning` clean on every edited `.sh`; `markdownlint-cli2` clean on both edited `.md`):

| Suite | Result |
|---|---|
| `test_block_main_push.sh` (37 pre-existing + 7 new) | **44/44 PASS** |
| `test_block_main_push_heredoc.sh` | 7/7 PASS |
| `test_heredoc_bypass_shapes.sh` (both hooks) | 36/36 PASS |
| `test_lib_protected_branches.sh` | 16/16 PASS |
| `test_extract_push_ref_arrow.sh` / `_heredoc.sh` | 9/9, 12/12 PASS |
| `test_validate_branch_name_pushref.sh` | 39/39 PASS |
| `test_validate_branch_name_heredoc.sh` (7 pre-existing + 3 new) | **10/10 PASS** |

New coverage added this PR:

- **Still-blocks proof** — every former `exit 2` site (push to main/dev/master/develop, plain commit on a protected branch) is re-asserted against the migrated hook; all still block. This is the direct proof the migration didn't demote anything.
- **`--no-verify` fires** — `git push --no-verify origin main` and `git commit --no-verify -m bad` on a protected branch both still block (rc=2) *and* the stderr message names `--no-verify` explicitly. A legitimate `--no-verify` push/commit to a *non*-protected branch still passes — the flag never manufactures a new block on its own.
- **Config-override parity** — with `.git.protected_branches` overridden to `["release"]`, `main` now passes and `release` now blocks, proving the migrated hook actually resolves through the shared lib rather than a stale inline copy.
- **#1081 regression (2 new cases + 1 control)** — the ticket's own repro (`git push origin bogus-branch` behind a dotted-delimiter heredoc carrying decoy `--tags` prose) now blocks; the same decoy shape with a *conforming* real push still passes; a genuine tag push with no decoy stays exempt.

## What closes on merge

- **#1086** — the epic itself; step 3 (both PRs) is now complete.
- **#1081** — fixed directly in this PR.
- **#1083, #1084, #1085** — `block-main-push.sh`'s / the shared parser's text-parsing residuals, catalogued in AgDR-0113. These close as *superseded at the git layer*: the git-native hooks (`.githooks/pre-push`, `.githooks/pre-commit`) that fully replace text-parsing for the *which-ref* / *which-branch* question are now live and this PR's tests confirm parity, so the parsing bugs these tickets tracked no longer have live exploit paths against the primary control (they remain theoretically present in `block-main-push.sh`'s own backstop parsing, same as before — AgDR-0114 accepted that residual explicitly).

Refs #1086 #1081 #1083 #1084 #1085

## Glossary

| Term | Definition |
|---|---|
| **Backstop** (vs. control) | A control is the layer decisions actually rely on (here: the git-native hooks, which read ground-truth git state). A backstop is a second, weaker-in-general-but-covers-different-gaps layer kept blocking specifically because it closes cases the control can't reach — here, `--no-verify` and clones without `core.hooksPath` installed. |
| `is_protected_branch()` | The shared, fail-closed predicate in `_lib-protected-branches.sh` that answers "is this branch name protected" — a malformed config entry or a missing helper function is treated as protected, never as a silent allow. |
| `--no-verify` | A native git flag that skips git's own hook mechanism (`pre-push`, `pre-commit`, etc.) entirely. It has no effect on a Claude Code PreToolUse hook, because that hook runs before the shell — and therefore before git — ever executes. |
| Decoy tag-push evidence | Text that matches a tag-push pattern (`--tags`, `tag <name>`, `refs/tags/`) but sits in prose (a heredoc body, an unrelated comment) unrelated to the actual `git push` in the same command — the root cause both #1075 (block-main-push.sh) and #1081 (validate-branch-name.sh) fixed. |

### DIFF
--- a/.claude/hooks/block-main-push.sh
+++ b/.claude/hooks/block-main-push.sh
@@ -1,8 +1,45 @@
 #!/bin/bash
-# Blocks direct pushes and commits to long-lived integration branches.
-# All changes must go through pull requests.
+# CLASS: BACKSTOP — blocking, but not the primary control (AgDR-0114).
 #
-# Protected branches (default): main / master / dev / develop.
+# Blocks direct pushes and commits to long-lived integration branches from
+# Claude Code sessions (a PreToolUse hook, evaluated before the shell ever
+# runs). All changes must go through pull requests.
+#
+# THE controls for protected-branch enforcement are the git-native hooks:
+#   .githooks/pre-push     — terminal `git push`, ground truth via git's own
+#                             stdin ref resolution (me2resh/apexyard#1086
+#                             step 2)
+#   .githooks/pre-commit    — terminal `git commit`, ground truth via
+#                             `git symbolic-ref --short HEAD` (step 3,
+#                             AgDR-0114)
+# Both need no command-text parsing at all — immune by construction to every
+# heredoc/decoy/quoting shape docs/agdr/AgDR-0113-heredoc-stripper-additive-only.md
+# catalogues for THIS file's text-matching approach.
+#
+# THIS file stays a BLOCKING backstop (AgDR-0114 — relabel + keep blocking,
+# not demote to advisory), because it is the only layer that covers two
+# gaps the git-native hooks structurally cannot close:
+#
+#   1. `--no-verify` bypasses every git-native hook by git's own design.
+#      This file is a Claude Code PreToolUse hook — it intercepts the Bash
+#      tool call BEFORE the shell (and therefore before git, and therefore
+#      before --no-verify's effect) ever runs, so it stays effective
+#      regardless of that flag. See the "--no-verify" note attached to
+#      each blocked-push/blocked-commit message below.
+#   2. `core.hooksPath` is not installed by default on managed-project
+#      clones (me2resh/apexyard#1088, deliberately unresolved) — this file
+#      fires process-wide regardless of which repo's clone the command's
+#      cwd targets, where the git-native hooks never fire at all.
+#
+# See docs/agdr/AgDR-0114-block-main-push-honest-naming-blocking-backstop.md
+# for the full decision record and docs/agdr/AgDR-0104-trust-chain-controls-vs-backstops.md
+# for the "control vs backstop, honestly named" framing this file follows.
+#
+# Protected branches (default): main / master / dev / develop, resolved via
+# the shared `.claude/hooks/_lib-protected-branches.sh` predicate
+# (`is_protected_branch`) — the same lib `.githooks/pre-push` and
+# `.githooks/pre-commit` use, so all three layers agree on one source of
+# truth instead of three copies that have to be trusted to stay in sync.
 # `dev` was added in apexyard#116 (release-cut model — see AgDR-0007). Forks
 # that legitimately use `dev` as a daily-work trunk under their own
 # convention can override the protected list via
@@ -61,16 +98,59 @@ HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # _lib-strip-heredoc.sh's governance comment and
 # docs/agdr/AgDR-0113-heredoc-stripper-additive-only.md.
 
-# Resolve protected-branch list from project config (shared reader, #109).
-REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
-PROTECTED=""
-if [ -n "$REPO_ROOT" ] && [ -f "$REPO_ROOT/.claude/hooks/_lib-read-config.sh" ]; then
+# Resolve the protected-branch predicate from the shared lib (me2resh/apexyard#1086
+# step 2/3, AgDR-0114) instead of resolving `.git.protected_branches[]` a
+# second time inline. This file previously carried its own copy of the same
+# three lines `_lib-protected-branches.sh` already implements — the
+# "duplicate that has to be trusted to agree" pattern AgDR-0113 catalogues as
+# this file family's dominant defect generator. Sourcing the lib means this
+# file, .githooks/pre-push, and .githooks/pre-commit all resolve the SAME
+# list, and all three inherit is_protected_branch()'s fail-closed hardening
+# (a malformed `.git.protected_branches[]` entry, or a renamed/missing
+# protected_branch_regex function, is treated as PROTECTED, never as a
+# silent allow).
+PROTECTED_LIB="$HOOK_DIR/_lib-protected-branches.sh"
+if [ -f "$PROTECTED_LIB" ]; then
   # shellcheck disable=SC1090,SC1091
-  . "$REPO_ROOT/.claude/hooks/_lib-read-config.sh"
-  PROTECTED=$(config_get '.git.protected_branches[]' 2>/dev/null | paste -sd'|' -)
+  . "$PROTECTED_LIB"
 fi
-if [ -z "$PROTECTED" ]; then
-  PROTECTED="main|master|dev|develop"
+if ! declare -F is_protected_branch > /dev/null 2>&1; then
+  # Old clone that hasn't pulled the lib file yet (or it's syntactically
+  # broken). Fail closed on the documented default list rather than
+  # skipping protected-branch enforcement entirely — this file is a
+  # blocking backstop; a missing lib must not turn it into a no-op.
+  echo "WARN: $PROTECTED_LIB not found or is_protected_branch unavailable -- falling back to the default protected-branch list (main/master/dev/develop only; .git.protected_branches[] override not applied)." >&2
+  is_protected_branch() {
+    local b="$1"
+    [ -n "$b" ] && echo "$b" | grep -qE '^(main|master|dev|develop)$'
+  }
+  protected_branch_regex() { echo "main|master|dev|develop"; }
+fi
+
+# Display-only rendering of the protected list for BLOCKED messages below
+# (e.g. "Protected branches: main, master, dev, develop").
+PROTECTED_DISPLAY=$(protected_branch_regex 2>/dev/null)
+if [ -z "$PROTECTED_DISPLAY" ]; then
+  PROTECTED_DISPLAY="main|master|dev|develop"
+fi
+
+# --no-verify note (me2resh/apexyard#1086 step 3, AgDR-0114): --no-verify
+# does not bypass THIS file (a PreToolUse hook, not a git hook), but an
+# operator seeing a block after passing --no-verify may reasonably wonder
+# why the flag "didn't work". Detected once, appended to every BLOCKED
+# message below so the reason is explicit rather than left to be
+# rediscovered. Detection is deliberately simple (a whitespace-bounded
+# literal match) -- this is an explanatory note, not a new gate; the
+# existing push/commit checks above and below already block the command
+# regardless of --no-verify, with or without this note.
+NO_VERIFY_NOTE=""
+if echo "$COMMAND" | grep -qE '(^|[[:space:]])--no-verify([[:space:]]|$)'; then
+  NO_VERIFY_NOTE="
+NOTE: --no-verify does not bypass this check. This hook is a Claude Code
+PreToolUse hook (AgDR-0114) that runs BEFORE the shell -- and therefore
+before git, and therefore before --no-verify's effect -- ever executes.
+The git-native .githooks/pre-push and .githooks/pre-commit hooks WOULD be
+skipped by --no-verify; this backstop is not."
 fi
 
 # ---------------------------------------------------------------------------
@@ -112,12 +192,12 @@ if echo "$COMMAND" | grep -qE '\bgit\s+push\b'; then
     if declare -F _extract_all_push_refs_core > /dev/null 2>&1; then
       while IFS= read -r ANY_PUSH_DST; do
         [ -z "$ANY_PUSH_DST" ] && continue
-        if echo "$ANY_PUSH_DST" | grep -qE "^(${PROTECTED})$"; then
+        if is_protected_branch "$ANY_PUSH_DST"; then
           cat >&2 <<MSG
 BLOCKED: Cannot push directly to a protected branch ('${ANY_PUSH_DST}').
 
 All changes must go through a PR (.claude/rules/git-conventions.md
-§ "No Direct Main"). Protected branches: ${PROTECTED//|/, }.
+§ "No Direct Main"). Protected branches: ${PROTECTED_DISPLAY//|/, }.
 
 This command contains more than one push-shaped occurrence of text, and
 at least one names a protected branch. If that text is inside a heredoc
@@ -131,6 +211,7 @@ To unblock a genuine push:
        git checkout -b feature/GH-<ticket>-<short-description>
   2. Push the feature branch:
        git push -u origin feature/GH-<ticket>-<short-description>
+${NO_VERIFY_NOTE}
 MSG
           exit 2
         fi
@@ -282,12 +363,12 @@ MSG
       fi
     fi
 
-    if [ -n "$TARGET_PUSH_BRANCH" ] && echo "$TARGET_PUSH_BRANCH" | grep -qE "^(${PROTECTED})$"; then
+    if is_protected_branch "$TARGET_PUSH_BRANCH"; then
       cat >&2 <<MSG
 BLOCKED: Cannot push directly to a protected branch ('${TARGET_PUSH_BRANCH}').
 
 All changes must go through a PR (.claude/rules/git-conventions.md
-§ "No Direct Main"). Protected branches: ${PROTECTED//|/, }.
+§ "No Direct Main"). Protected branches: ${PROTECTED_DISPLAY//|/, }.
 
 To unblock:
   1. Create a feature branch from your current work:
@@ -303,6 +384,7 @@ protection from a default-protected branch (e.g. you legitimately use
 'dev' as your trunk), write the array with that branch OMITTED. To ADD
 protection to a new branch, write the array INCLUDING it. The hook
 trusts whichever list you provide — get the direction right.
+${NO_VERIFY_NOTE}
 MSG
       exit 2
     fi
@@ -354,7 +436,7 @@ if echo "$COMMAND" | grep -qE '\bgit\s+commit\b'; then
     CURRENT_BRANCH=$(git branch --show-current 2>/dev/null)
   fi
 
-  if [ -n "$CURRENT_BRANCH" ] && echo "$CURRENT_BRANCH" | grep -qE "^(${PROTECTED})$"; then
+  if is_protected_branch "$CURRENT_BRANCH"; then
     cat >&2 <<MSG
 BLOCKED: Cannot commit directly on protected branch '${CURRENT_BRANCH}'.
 
@@ -377,6 +459,7 @@ locally, then check out the recovery branch.
        git branch feature/GH-<ticket>-recovery
        git reset --hard HEAD~1   # drops the commit from ${CURRENT_BRANCH}
        git checkout feature/GH-<ticket>-recovery
+${NO_VERIFY_NOTE}
 MSG
     exit 2
   fi
--- a/.claude/hooks/tests/test_block_main_push.sh
+++ b/.claude/hooks/tests/test_block_main_push.sh
@@ -33,8 +33,9 @@ LIB_SRC="$SRC_ROOT/.claude/hooks/_lib-extract-push-ref.sh"
 LIB_STRIP_HEREDOC_SRC="$SRC_ROOT/.claude/hooks/_lib-strip-heredoc.sh"
 LIB_CONFIG_SRC="$SRC_ROOT/.claude/hooks/_lib-read-config.sh"
 LIB_OPS_ROOT_SRC="$SRC_ROOT/.claude/hooks/_lib-ops-root.sh"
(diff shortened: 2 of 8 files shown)


me2resh (commented):
## Code Review: PR #1120 — APPROVE

**Commit reviewed**: `76ca0e01a824ca5dedf05669602d5a9af0d07beb`

Reviewed as PR-2 of the #1086 epic — the core trust-chain payoff. I verified the load-bearing invariant against the actual HEAD source (read via API, not checkout) and independently re-ran every affected suite against a clean mirror of the HEAD tree.

### Checklist Results
- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass (independently re-ran 6 suites — see below)
- Security (trust-chain): Pass — note Security-Auditor recommendation below
- PR Description & Glossary: Pass (narrative Summary, PR-2-of-2 note, `Refs #1086 #1081 #1083 #1084 #1085`, 4-term Glossary, AgDR-0114 linked)
- Technical Decisions (AgDR): Pass — this PR *executes* AgDR-0114; no new decision introduced

### The load-bearing check — "keep blocking, none demoted" (AgDR-0114)
Confirmed against the HEAD file. All **four** `exit 2` sites survive intact, none demoted to advisory:

| Site (HEAD line) | What it blocks |
|---|---|
| 216 | protected-branch push found by the scan-all (`_extract_all_push_refs_core` → `is_protected_branch`) |
| 323 | fail-closed ambiguous-destination block (decoy hiding a ref-less push) — unchanged |
| 389 | protected-branch push (primary `TARGET_PUSH_BRANCH` check) |
| 464 | commit on a protected branch (`CURRENT_BRANCH` check) |

The diff removes no `exit 2` line; every hunk only *adds* the `--no-verify` note ahead of an existing `exit 2`. The new "still-blocks" cases in `test_block_main_push.sh` re-assert rc=2 for push-to-main/dev/master/develop and commit-on-protected against the migrated hook.

### Migration is real, not cosmetic
The three inline `grep -qE "^(${PROTECTED})$"` sites are replaced by `is_protected_branch()` sourced from `.claude/hooks/_lib-protected-branches.sh`. Behaviour is preserved (the old `[ -n "$X" ] &&` guards are subsumed by the predicate's own empty-string → not-protected return) and the three sites now inherit the lib's fail-closed hardening (missing predicate or malformed `.git.protected_branches[]` → treated as protected). The `if ! declare -F is_protected_branch` fallback keeps an old/broken-lib clone blocking on the documented default list with a WARN rather than silently no-op'ing — correct for a blocking backstop.

The config-override test (`.git.protected_branches=["release"]` → main un-protected, release protected) is the proof the resolution goes through the shared lib, not a stale inline copy. I confirmed it genuinely exercises `config_get` (sandbox is git-init'd, defaults + override copied, `_config_repo_root` resolves to the sandbox) — **44/44** in a clean run.

### `--no-verify` handling + honest relabel
`NO_VERIFY_NOTE` is appended to all three BLOCKED messages, detection is a simple whitespace-bounded literal match (explanatory only — the block already fires regardless). Tests assert the message names `--no-verify` on both a protected push and a protected commit, and that a legit `--no-verify` push/commit to a non-protected branch still passes (the flag never manufactures a new block). The header is relabelled `CLASS: BACKSTOP`, naming `.githooks/pre-push` / `.githooks/pre-commit` as THE controls and this file as the blocking backstop for the two gaps they structurally can't reach (`--no-verify`, unconfigured `core.hooksPath`). Accurate.

### #1081 fix — the custom scan is the right call (my judgment)
The ticket cites reusing `_is_genuine_single_tag_push`; that function no longer exists — it was superseded by `_command_has_untagged_refless_push` in #1075's 7th review round. The agent correctly did **not** cargo-cult a dead reference and instead wrote a local per-occurrence evidence-vs-ref scan. I judge this **justified, not reinvention**:

- `_command_has_untagged_refless_push` is the **wrong tool** here. It detects *ref-LESS* pushes and returns a boolean; #1081 is a *ref-BEARING, evidence-less* push (`git push origin bogus-branch`) whose **name** must be validated. That helper explicitly skips ref-bearing occurrences (`[ -z "$ref" ]`), and it can't hand back the ref value the naming validator needs. It's tuned for block-main-push's residual (where scan-all already covers explicit refs against a protected list); validate-branch-name has no such scan-all, so it needs a different output.
- `_extract_all_push_refs_core` can't be used naively either — it returns *every* ref including the decoy's (`releases`), which would over-block the conforming-real-push case. The tag-evidence filter is exactly what avoids that.
- The scan is **discriminating and correctly bounded**: Shape 1 (decoy + non-conforming real push) blocks (rc=2, red pre-fix); Shape 2 (decoy + conforming real push) passes; the control (genuine `--tags`) stays exempt; and genuine multi-remote tag pushes / ordinary-push-plus-tag-push shapes (the exact cases that broke both earlier whole-command versions) don't over-fire. Taking the LAST evidence-less ref correctly targets the real command hiding behind leading decoy prose. Confirmed-heredoc decoys are handled by the pre-existing stripped path, so the scope (unconfirmed heredoc / plain prose) is right. `test_validate_branch_name_heredoc.sh` = **10/10**.

### Independent test run (clean mirror of HEAD, matching CI)
`test_block_main_push.sh` 44/44 · `test_block_main_push_heredoc.sh` 7/7 · `test_heredoc_bypass_shapes.sh` 36/36 · `test_validate_branch_name_heredoc.sh` 10/10 · `test_validate_branch_name_pushref.sh` 39/39 · `test_lib_protected_branches.sh` 16/16. `shellcheck --severity=warning` clean on both edited hooks. GitHub CI: hook-suite, markdownlint-cli2, shellcheck all green.

Note: running the suite *inside a live agent session*, the SessionStart ops-root pin hijacks `_config_repo_root` to the real fork, spuriously failing the 3 config-override cases until `APEXYARD_OPS_DISABLE_PIN=1` is set. CI has no pin and is green. Informational only.

### honest-naming docs
`git-conventions.md` and the AgDR-0104 "Update" section are accurate and internally consistent — they name the git-native hooks as the controls, this file as the blocking backstop, and correctly *narrow* AgDR-0104's "backstop → advisory" leap by noting it's only safe when a strictly stronger control covers everything the backstop did (false here: `--no-verify` + unconfigured `core.hooksPath` are real gaps), so this hook stays blocking. `Refs` (not `Closes`) for #1083/#1084/#1085 is the honest choice — the PR body says they close as *superseded at the git layer* while the residual backstop parsing remains, which matches reality.

- `nit` (test doc): the #1081 regression comment says "Both cases below are proven RED against the pre-fix code" — only Shape 1 is genuinely red pre-fix; Shape 2 and the control are green both before and after (over-block guards, not red-before regressions). Tighten the wording.
- `suggestion` (DRY): the tag-evidence triple (`\s--tags(\s|$)` / `\stag\s+\S` / `refs/tags/`) now appears in three places (`is_tag_push`, `_command_has_untagged_refless_push`, and the new scan). Given this PR's own AgDR-0113 thesis (kill drift-prone duplicates), a shared `_segment_has_tag_evidence()` would consolidate it. The PR is still a net reduction in duplication, so this is a follow-up, not a blocker.

### Security note
The diff touches `.claude/hooks/**` (the security trust chain, role-triggers #777). This approval is on correctness / code-quality grounds; a Security Auditor (Hakim) pass is the appropriate second gate before the human merge nod.

### Verdict
**APPROVE.** The load-bearing invariant (keep blocking, four exit-2 sites intact) is preserved and directly tested; the lib migration is real; the #1081 custom scan is correct, discriminating, and justified over the non-fitting shared helpers; docs are honestly scoped; every suite is green.

---
Reviewed commit: `76ca0e01a824ca5dedf05669602d5a9af0d07beb`

me2resh (commented):
## Security Review: PR #1120 — trust-chain (protected-branch controls)

**Reviewed commit**: `76ca0e01a824ca5dedf05669602d5a9af0d07beb`
🛡️ Reviewed by Hakim (Security Auditor) — auto-fired on the trust chain (`.claude/hooks/**`).

I read the resulting `block-main-push.sh`, `validate-branch-name.sh`, and `_lib-protected-branches.sh` at HEAD, traced the migration line-by-line, and ran an adversarial suite against the **actual hooks** — sandboxed from the PR-head files, and against the pre-fix base commit for regression proof.

### Verdict: **SECURE** ✅ — no protection lost, `--no-verify` covered, #1081 closed without opening a new bypass.

---

### 1. Migration onto `is_protected_branch()` — no protection lost, fail-closed *strengthened*

The change swaps `[ -n "$X" ] && echo "$X" | grep -qE "^(${PROTECTED})$"` for `is_protected_branch "$X"` at all three sites (scan-all loop, single push target, commit branch). This is semantically identical on the happy path (empty → not protected; same default alternation `main|master|dev|develop`; same `.git.protected_branches[]` resolution) and **strictly stronger** on the failure paths:

- **Old inline grep silently ALLOWED on a malformed config entry.** A fragment like `"ma(in"` makes `grep -qE "^(ma(in)$"` exit `>1` (ERE error), which `if …; then BLOCK; fi` reads as "no match → not protected". The shared predicate treats grep exit `>1` (and a missing `protected_branch_regex`) as **protected + WARN**. Verified: with `protected_branches:["ma(in"]`, even a *feature* push now fails closed (`rc=2`).

Empirical (PR-head hooks):

| Case | Result |
|---|---|
| push to `main` / `master` / `dev` / `develop` | all block (`rc=2`) — every former exit-2 site intact |
| plain commit on protected `main` | blocks (`rc=2`) |
| config override `["release"]` → push/commit `main` | now **pass** (`rc=0`) — REPLACE semantics honoured |
| config override `["release"]` → push `release` | blocks (`rc=2`) |
| malformed `["ma(in"]` → feature push | **fails closed** (`rc=2`) |
| `_lib-protected-branches.sh` missing → commit on `main` | stub defaults still block (`rc=2`) |

### 2. `--no-verify` — actually blocked, and named

`--no-verify` is the one shape the git-native `.githooks/*` cannot catch (git skips its own hooks by design), and this PreToolUse hook runs before the shell. Confirmed the block is real, not just the new advisory note: a `--no-verify` push to `main` and a `--no-verify` commit on `main` both block (`rc=2`) **and** the stderr names `--no-verify` ("does not bypass this check"). The flag is correctly parsed as a boolean by the ref extractor, so `origin main` still resolves to `main`. A `--no-verify` push/commit to a *feature* branch still passes (`rc=0`) — the note never manufactures a new block on its own.

### 3. The #1081 fix (`validate-branch-name.sh`) — sound, and monotonic

This is the sharpest part: the agent wrote a *new* per-occurrence evidence-vs-ref scan (`SCAN_LAST_REF`), not the `_is_genuine_single_tag_push` the ticket suggested. I stress-tested it.

- **Ticket repro closed** — decoy `--tags` prose in an unconfirmed (`END.OF`) heredoc + a real non-conforming push: **pre-fix `rc=0` (bypass) → post-fix `rc=2` (blocked)**. Confirmed against the base commit.
- **No over-block** — conforming push behind the same decoy, compound `explicit-ref push && genuine --tags`, and genuine `--tags` / `tag <name>` / `refs/tags/` pushes all still pass. The per-segment evidence check runs *before* ref extraction, so a genuine tag push's own `v1.2.3` token never triggers a false block.
- **Decoy placement** — the scan takes the **last** evidence-less ref-bearing occurrence (the one about to execute), so decoy-before, decoy-after, and multi-decoy-with-real-last all block correctly.
- **`SCAN_LAST_REF` survives the loop** — the walk uses process substitution (`< <(…)`), not a pipe, so the assignment persists in the current shell.
- **Monotonicity (the key safety property)** — the fix only alters the `is_tag_push == true` branch (old: always `exit 0`; new: sometimes blocks). When `is_tag_push == false` the path is byte-identical to base. So the new code can **never** `exit 0` where the old code blocked → no protection lost. Empirically the H-series residuals below are `rc=0` on *both* base and PR-head — no regression introduced.

### Residuals (LOW / informational — pre-existing, out of #1081 scope, P2 naming-only)

Three naming-gate shapes still `exit 0`, all confirmed `rc=0` on the base commit too (this PR neither introduces nor worsens them):

1. Same-segment tag evidence (a real `--tags` mixed push, or a trailing `# --tags` comment) exempts that occurrence.
2. A multi-real push validates only the last evidence-less ref, so a non-conforming earlier push behind a conforming last one isn't validated.
3. A **non-tag** conforming decoy in an unconfirmed heredoc hijacks `extract_push_ref`'s first-match; the result is non-empty so the fail-closed check doesn't engage. This is the non-tag sibling of #1081, which `block-main-push.sh` already closed via `_extract_all_push_refs_core` (scan-all) but `validate-branch-name.sh` does not carry.

**Why none of these are a protected-branch bypass:** for every one, I ran the dangerous equivalent (real push to `main`/`dev`) against `block-main-push.sh` and all block (`rc=2`) — its evidence-agnostic scan-all catches them regardless of tag/decoy shape. This matches #1081's own P2 framing ("a naming-convention miss, not a protected-branch bypass").

**Non-blocking recommendation:** consider a follow-up to port `block-main-push.sh`'s scan-all discipline to `validate-branch-name.sh`'s non-tag path (residual #3), fully closing the naming-gate's parallel to #1081. Out of scope for this PR.

### Trust-chain / honest-naming checks
- Header relabel to CLASS: BACKSTOP and the `git-conventions.md` / AgDR-0104 edits are documentation-only and do **not** demote any `exit 2` site — every block remains a block (verified above). No gate weakened, no matcher wiring loosened, no fail-open introduced.
- The lib-missing stub is honest: it WARNs and preserves the default protected set, though it does not re-read config (a custom-configured branch loses protection only while the shared lib is absent — an abnormal, loudly-warned state; re-reading config in the stub would re-introduce the very duplicate this PR removes).

### Checklist
- Secrets & Credentials: Pass · Injection: Pass · XSS: n/a · Auth/Authz: n/a
- **Gate & Trust-Chain Integrity: Pass** — protected-branch enforcement preserved and strengthened; `--no-verify` gap covered; #1081 closed.

---
📌 Reviewed commit: `76ca0e01a824ca5dedf05669602d5a9af0d07beb`

atlas-apex (approved):

mostiwheelietravel (commented):
_Redacted — posted to the wrong repository by an automation misconfiguration. Not intended for this repository._

Issue #1121, opened 2026-08-01
[P1] block-merge-on-red-ci.sh checks the wrong forge's CI for a glab-registered project via the merge wrapper


me2resh:
**[P1] block-merge-on-red-ci.sh checks the wrong forge's CI for a glab-registered project merged via the tracker_pr_merge wrapper**

## Bug Scenario

**Given** a project is registered in `apexyard.projects.yaml` with a per-project `tracker: kind: glab` override, and `/approve-merge` (or any caller) merges its PR through the tracker-agnostic `tracker_pr_merge <owner/repo> <pr> <strategy> [<delete_branch>]` wrapper (#759) — whose own command text never says `gh` or `glab`, by design
**When** `block-merge-on-red-ci.sh` (the red-CI merge gate, #790) fires on that wrapper call and needs to decide which forge's CI status to check
**Then** it dispatches via `gh pr checks <N>` — the wrong forge — instead of the GitLab pipeline-status check, whenever neither `yq` nor `python3` + `PyYAML` is installed in the hook's execution environment
**Expected** the gate should check the GitLab MR's pipeline status (`resolve_ci_status_glab`) for a glab-registered project's wrapper-shape merge, regardless of whether `yq`/PyYAML happen to be installed

## Repro Steps

1. In a sandbox with `apexyard.projects.yaml` registering a project with `repo: g/p` and `tracker: kind: glab`, and **without** `yq` or PyYAML (`python3 -c "import yaml"` fails) on `PATH`
2. Pipe a synthetic PreToolUse payload for `. "<sandbox>/.claude/hooks/_lib-tracker.sh"; MERGE_RESULT=$(tracker_pr_merge "g/p" "500" "squash" true)` into `block-merge-on-red-ci.sh`
3. Observe the mocked `gh` binary gets invoked (it prints `WRONG: gh should not be called for a glab-registered project` and exits 1), and the hook blocks with a `gh pr checks` — flavoured "red CI" message instead of the GitLab pipeline message
4. Run `.claude/hooks/tests/test_block_merge_on_red_ci.sh` — the two `wrapper: glab-registered project, ...` cases fail with `want rc=0, got 2` (pipeline success case) and a stderr mismatch (pipeline failure case), both showing the `gh`-path error text

## Environment

[...]

Issue #790, opened 2026-07-05
[Bug] block-merge-on-red-ci.sh not forge-aware — hardcodes 'gh pr checks', breaks on GitLab


atlas-apex:
## Given / When / Then
**Given** a GitLab-tracked project using the forge-aware merge gates (post-#767),
**When** `block-merge-on-red-ci.sh` runs on a merge command,
**Then** it calls `gh pr checks` unconditionally instead of routing through a forge-aware resolver — so on a GitLab project it will error (or misbehave) rather than checking the MR's real pipeline status.

## Context
Surfaced during the #776 verification. `block-unreviewed-merge.sh` was made forge-aware in #767 (gh/glab), but its sibling `block-merge-on-red-ci.sh` still hardcodes `gh pr checks`. This is NOT a fail-open merge bypass (it errors rather than silently passing) — it's a CI-status-correctness gap in GitLab support, so it's the last non-forge-aware merge gate.

## Repro
On a project with `.tracker`/forge = GitLab, invoke a merge command and observe `block-merge-on-red-ci.sh` calling `gh pr checks` against a repo gh can't resolve.

## Acceptance Criteria
- [ ] `block-merge-on-red-ci.sh` routes CI-status checks through the same forge-aware resolver `block-unreviewed-merge.sh` uses (gh → `gh pr checks`, glab → the equivalent `glab ci`/MR pipeline status)
- [ ] Fail-closed on unknown forge / status-fetch failure (a red or unfetchable CI must BLOCK, never silently allow)
- [ ] Regression test covering the glab path + the gh path unchanged

## Severity
Low-Medium — completeness of the GitLab-forge epic (#711); not a security bypass.

atlas-apex:
Fixed in #793 (merged to `dev`) — `block-merge-on-red-ci.sh` is now forge-aware (gh/glab), fail-closed on every unresolvable/malformed glab response (the fail-open the security review caught is closed and provably blocked). This was the last merge gate hardcoded to `gh`; the forge-aware merge-gate set (with #767) is now complete. Auto-close didn't fire because the PR targeted `dev`.

Issue #1102, opened 2026-07-31
[Bug] _lib-read-config.sh has the same unanchored cwd self-location as #1062, and its fallback is unanchored too


atlas-apex:
## Bug Scenario

**Given** `_lib-read-config.sh` self-locates the same way every other trust-chain lib does:

```sh
lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-}")" 2>/dev/null && pwd)"
if [ -f "$lib_dir/_lib-ops-root.sh" ]; then
  . "$lib_dir/_lib-ops-root.sh"
```

**When** it is sourced under a non-bash shell, where `BASH_SOURCE` is unset — so `dirname "" → "."` and `lib_dir` becomes the **caller's cwd**

**Then** if that cwd happens to contain a file named `_lib-ops-root.sh`, the `-f` guard is satisfied and the lib is sourced **unanchored** from whatever repo the cwd is inside

**Expected** the same anchoring every sibling lib now gets — resolution should only accept a genuine apexyard fork.

## Why this is a separate ticket rather than part of #1100

#1100 fixes exactly this shape at **nine sites across eight libs**, using:

```sh
if [ -z "${BASH_SOURCE[0]:-}" ]; then _lib_dir=""; fi
```

Discarding the cwd-derived path forces resolution through each lib's **anchored git-root fallback** (#1033/#1061), which only accepts a real fork.

**That guard is not a drop-in here.** `_lib-read-config.sh`'s fallback is a deliberately *unanchored* `git rev-parse` — its own comment says so, describing it as "legacy behaviour for non-apexyard environments" that keeps the hooks usable in bare clones and CI sandboxes that don't ship the full framework. So clearing `lib_dir` would route to a fallback that is itself unanchored, moving the problem rather than fixing it.

**The anchor has to be added to the fallback first, then the guard.** That is a behaviour change to the deliberate non-apexyard escape hatch, and it deserves its own diff, its own reasoning about what should happen in a bare clone, and its own review.

Verified independently: the shape is present at `_lib-read-config.sh:91`, and `_lib-read-config.sh` appears **zero times** in #1100's diff.

## Severity

**medium** — same class as #1062, same trust-chain exposure, but narrower in practice: it requires a non-bash shell **and** a cwd that happens to contain a file named `_lib-ops-root.sh`. Not reachable by accident in ordinary use; reachable deliberately, and reachable by an operator working inside an unrelated checkout.

[...]

PR #1122, opened 2026-08-01
fix(#1121): dispatch the red-CI check on tracker_kind so glab-registered projects check glab, not gh

me2resh:
## Summary

- **Wrong-forge bug**: for a project registered as `glab`-kind in `apexyard.projects.yaml`, `block-merge-on-red-ci.sh` dispatched the CI-status check via `gh pr checks <N>` instead of the GitLab pipeline-status check when merged through the tracker-agnostic `tracker_pr_merge <owner/repo> <pr>` wrapper (#759). The hook already resolves the forge from the registry for this shape (`_forge_kind_for` → `tracker_kind` → `_lib-tracker.sh`'s `_tracker_project_value`, added in #790/#794) — but that registry read needs `yq` or `python3` + PyYAML to parse the per-project `tracker.kind` override, and when **neither** is installed it silently falls through to the global `.tracker.kind` default (`gh`). For a mixed-forge portfolio that's exactly wrong for the glab-registered project, and it reproduces on any box (including, plausibly, the CI runner — it only installs `jq`, not `yq`, and doesn't guarantee PyYAML) that lacks both tools.
- **Fix — registry fallback, scoped to this one hook**: added `_registry_glab_fallback`, a dependency-free (`grep`/`awk` only) scan of `apexyard.projects.yaml` that answers "does this exact repo have an explicit `tracker: kind: glab`?". It's consulted **only** when the primary resolver answers `"gh"` for a wrapper-shape command — it can only ever upgrade an ambiguous `gh` to a confirmed `glab`, never override a genuine `glab` result or invent a false positive for a project the registry doesn't explicitly mark `glab`. This is deliberately hook-local rather than a change to `_lib-tracker.sh`/`_lib-extract-pr.sh` (out of scope for this fix — that's the shared-resolver epic's territory).
- **Fail-closed posture preserved for both forges** — nothing about the existing gh-path or glab-path blocking behaviour changed; the fallback only ever affects which forge gets *checked*, never the allow/block decision logic itself.
- **Deterministic regression coverage**: the two originally-failing tests only caught this by accident of the *local* environment lacking `yq`/PyYAML. Added 7 new cases that stub `yq`/`python3` to always fail so the fix is exercised regardless of what a given CI/dev box happens to have installed — 2 proving the glab-registered project still resolves correctly (allow on green pipeline, block on red), 2 proving a genuinely `gh`-kind project (no `tracker:` override at all) is never false-positived into the `glab` path, and 3 (added in review) proving the fallback is **position-independent** in the registry (glab first-of-two green→allow / red→block, glab middle-of-three green→allow).
- **Review fix — position-independent fallback (commit 2)**: Rex's review caught that `_registry_glab_fallback`'s awk printed its match twice whenever the glab block was NOT the last registry entry — a mid-stream `print "glab"; exit`, then a second print in the `END` block (awk's `exit` runs `END` with `block_repo`/`block_kind` still set), so the caller compared against `"glab\nglab"` and MISSED. That made the fix work only for a single-/last-entry registry (the shape the initial tests happened to use). Fixed with an `emitted` flag so `END` is a no-op after a mid-stream hit — exactly one `"glab"` regardless of position. The 3 new position cases fail on the pre-fix hook and pass on the fix, proving they exercise the real defect.

## Testing

- `.claude/hooks/tests/test_block_merge_on_red_ci.sh`: **34 passed, 0 failed**. Verified against the base hook the 3 position cases genuinely fail (31 passed / 3 failed — a glab project with a *green* pipeline is wrongly BLOCKED via the gh path), and against the fix all pass — so the tests distinguish *which forge was dispatched*, not just the exit code. Also verified with `yq` absent (original repro conditions) and with `yq` installed — both paths pass.
- `shellcheck --severity=warning` clean on both changed files.
- No regressions in the sibling suites that exercise the same shared resolver lib (`_lib-extract-pr.sh`) or other merge gates: `test_block_unreviewed_merge.sh` (41/41), `test_extract_pr.sh` (24/24), `test_extract_repo_fork_scoping.sh` (3/3), `test_forge_aware_extract_pr.sh` (48/48), `test_require_architecture_review.sh` (24/24), `test_require_design_review_for_ui.sh` (24/24), `test_require_posted_review.sh` (12/12).

Refs #1121

## Glossary

| Term | Definition |
|------|------------|
| Forge | The git hosting platform a project uses — GitHub (`gh` CLI) or GitLab (`glab` CLI). ApexYard supports mixed-forge portfolios via a per-project `tracker.kind` registry override. |
| `tracker_pr_merge` wrapper | A tracker-agnostic shell function (`_lib-tracker.sh`, #759) that `/approve-merge` calls to merge a PR/MR without hardcoding `gh` or `glab` in its own command text — the forge is resolved from the registry at call time instead. |
| CONTROL-class hook | A merge-gate hook (AgDR-0104/0109) that decides on structured state (a CI conclusion, a review marker) rather than on the text of a command, and must fail closed when it can't evaluate its precondition. |
| Fail closed | When a hook can't determine the safe answer, it blocks rather than allows — the existing posture for both the gh and glab paths, unchanged by this PR. |

### DIFF
--- a/.claude/hooks/block-merge-on-red-ci.sh
+++ b/.claude/hooks/block-merge-on-red-ci.sh
@@ -38,6 +38,25 @@
 # Pending checks (IN_PROGRESS | QUEUED): BLOCKED. The rule says all checks
 # must be green; pending is not green. Wait for CI to finish, then retry.
 #
+# WRAPPER-SHAPE FORGE FALLBACK (#1121)
+# -------------------------------------
+# The `tracker_pr_merge` wrapper (#759) resolves its forge from the registry
+# via `_forge_kind_for` -> `tracker_kind` -> `_lib-tracker.sh`'s
+# `_tracker_project_value`, which reads the per-project `tracker.kind`
+# override with `yq` (preferred) or a `python3` + PyYAML fallback. When
+# NEITHER tool is installed, that read silently returns nothing and
+# `tracker_kind` falls through to the GLOBAL `.tracker.kind` default (`gh`)
+# — exactly wrong for a glab-registered project's wrapper call, since the
+# wrapper's own text never says which CLI it drives. `_registry_glab_fallback`
+# below is a dependency-free (grep/awk only, no yq/PyYAML) second check used
+# ONLY when the primary resolver answers "gh" for a wrapper-shape command: it
+# scans the registry file directly for an explicit `tracker: kind: glab` on
+# the target repo. It can only ever upgrade an ambiguous "gh" to a confirmed
+# "glab" — it never downgrades a "glab" answer and never fires for the
+# non-wrapper shapes (those dispatch on command text, which is always
+# trustworthy). The real fix for the underlying yq/PyYAML-optional gap
+# belongs to the shared resolver libs, out of scope here.
+#
 # GLAB PATH (new, #790)
 # ----------------------
 # `resolve_ci_status_glab` (see _lib-extract-pr.sh) resolves the GitLab MR's
@@ -60,6 +79,93 @@ INPUT=$(cat)
 # the jq-independent fallback detector when the parse can't be trusted —
 # see #965.
 . "$(dirname "$0")/_lib-extract-pr.sh"
+
+# _registry_glab_fallback <owner/repo>
+# -------------------------------------
+# Dependency-free (grep/awk only — no yq, no python3/PyYAML) second check for
+# whether the registry EXPLICITLY marks <owner/repo> as `tracker: kind: glab`
+# (#1121 — see the WRAPPER-SHAPE FORGE FALLBACK header comment above for why
+# this exists). Hook-local by design: it duplicates a narrow slice of
+# `_lib-tracker.sh`'s `_tracker_project_value` on purpose, rather than
+# touching that shared resolver, so this fix stays scoped to this one hook.
+#
+# Scans project blocks the same way apexyard.projects.yaml is documented to
+# be shaped (apexyard.projects.yaml.example): a top-level `projects:` list
+# where each entry is a `- name: ...` block containing a `repo:` field and,
+# optionally, a `tracker:` sub-block with a `kind:` field. A block boundary
+# is any `- ` list-item line at the SAME indentation as the first one seen —
+# this is what keeps a NESTED list (`roles:`, `tags:`) from being mistaken
+# for a new project entry, since nested items are always more indented than
+# the top-level `- name:` marker.
+#
+# Echoes "glab" and exits 0 only when the target repo's block contains an
+# explicit `kind: glab`; otherwise echoes nothing and exits 1. This can only
+# ever confirm "glab" — it never asserts "gh" — so a caller should treat a
+# miss as "no additional information", not as "confirmed gh".
+_registry_glab_fallback() {
+  local repo="${1:-}" registry=""
+  [ -n "$repo" ] || return 1
+
+  if command -v portfolio_registry >/dev/null 2>&1; then
+    registry=$(portfolio_registry 2>/dev/null)
+  fi
+  [ -n "$registry" ] || registry="./apexyard.projects.yaml"
+  [ -f "$registry" ] || return 1
+
+  local found
+  found=$(awk -v target="$repo" '
+    function trim(s) { gsub(/^[ \t]+|[ \t]+$/, "", s); return s }
+    function unquote(s,    t, n, c1, cn, sq) {
+      sq = sprintf("%c", 39)
+      t = s
+      sub(/[ \t]*#.*$/, "", t)
+      t = trim(t)
+      n = length(t)
+      if (n >= 2) {
+        c1 = substr(t, 1, 1)
+        cn = substr(t, n, 1)
+        if ((c1 == "\"" && cn == "\"") || (c1 == sq && cn == sq)) {
+          t = substr(t, 2, n - 2)
+        }
+      }
+      return t
+    }
+    BEGIN { item_indent = -1 }
+    {
+      raw = $0
+      match(raw, /^[ \t]*/)
+      ind = RLENGTH
+      is_item = (raw ~ /^[ \t]*-[ \t]/)
+
+      if (is_item) {
+        if (item_indent == -1) { item_indent = ind }
+        if (ind == item_indent) {
+          if (block_repo == target && block_kind == "glab") { print "glab"; exit }
+          block_repo = ""; block_kind = ""
+        }
+      }
+
+      line = raw
+      sub(/^[ \t]*-[ \t]*/, "", line)
+      line = trim(line)
+      if (line ~ /^repo:[ \t]*/) {
+        v = line; sub(/^repo:[ \t]*/, "", v); block_repo = unquote(v)
+      }
+      if (line ~ /^kind:[ \t]*/) {
+        v = line; sub(/^kind:[ \t]*/, "", v); block_kind = unquote(v)
+      }
+    }
+    END {
+      if (block_repo == target && block_kind == "glab") print "glab"
+    }
+  ' "$registry" 2>/dev/null)
+
+  if [ "$found" = "glab" ]; then
+    echo "glab"
+    return 0
+  fi
+  return 1
+}
 
 # Parse .tool_input.command via jq. #965: this used to be the ONLY parse
 # path, and an empty/failed result — jq missing from PATH, or jq erroring
@@ -170,6 +276,21 @@ fi
 # existing #790 test behaviour exactly.
 if echo "$COMMAND" | grep -qE '\btracker_pr_merge\b'; then
   FORGE=$(_forge_kind_for "$CMD_REPO")
+  # #1121: _forge_kind_for's registry read needs yq or python3+PyYAML; when
+  # NEITHER is installed it silently falls through to the global "gh"
+  # default. That default is ambiguous here — it could be a genuine gh-kind
+  # project, or a glab-kind project whose registry entry just couldn't be
+  # read. Cross-check with the dependency-free scan (see
+  # _registry_glab_fallback above) whenever the primary answer is "gh": it
+  # can only ever confirm "glab", never override a real "glab" result or
+  # invent a false positive for a project the registry doesn't explicitly
+  # mark as glab.
+  if [ "$FORGE" = "gh" ]; then
+    REGISTRY_FORGE=$(_registry_glab_fallback "$CMD_REPO")
+    if [ "$REGISTRY_FORGE" = "glab" ]; then
+      FORGE="glab"
+    fi
+  fi
 else
   FORGE=$(_forge_from_command "$COMMAND")
 fi
--- a/.claude/hooks/block-merge-on-red-ci.sh
+++ b/.claude/hooks/block-merge-on-red-ci.sh
@@ -130,7 +130,7 @@ _registry_glab_fallback() {
       }
       return t
     }
-    BEGIN { item_indent = -1 }
+    BEGIN { item_indent = -1; emitted = 0 }
     {
       raw = $0
       match(raw, /^[ \t]*/)
@@ -140,7 +140,16 @@ _registry_glab_fallback() {
       if (is_item) {
         if (item_indent == -1) { item_indent = ind }
         if (ind == item_indent) {
-          if (block_repo == target && block_kind == "glab") { print "glab"; exit }
+          # Flush the block that just ended. When the target glab block is NOT
+          # the last registry entry, the match fires here (mid-stream). We must
+          # exit WITHOUT letting the END block re-print: awk exit jumps to END,
+          # so a bare "print glab; exit" re-satisfies the END condition (block_repo
+          # and block_kind are still set) and emits a SECOND glab line, making the
+          # caller compare found against "glab\nglab" and MISS. The emitted flag
+          # makes END a no-op after a mid-stream hit (the #1121 position fix).
+          # NOTE: no apostrophes in these comments — the whole awk program is a
+          # single-quoted shell string, so an apostrophe would terminate it.
+          if (block_repo == target && block_kind == "glab") { print "glab"; emitted = 1; exit }
           block_repo = ""; block_kind = ""
         }
       }
@@ -156,7 +165,9 @@ _registry_glab_fallback() {
       }
     }
     END {
-      if (block_repo == target && block_kind == "glab") print "glab"
+      # Only fires for the LAST block (no later item boundary flushed it). The
+      # `!emitted` guard prevents a double-print after a mid-stream hit above.
+      if (!emitted && block_repo == target && block_kind == "glab") print "glab"
     }
   ' "$registry" 2>/dev/null)
 
--- a/.claude/hooks/tests/test_block_merge_on_red_ci.sh
+++ b/.claude/hooks/tests/test_block_merge_on_red_ci.sh
@@ -322,6 +322,131 @@ sb=$(make_sandbox_wrapper failure)
 run_case "wrapper: glab-registered project, pipeline failed -> blocks (forge dispatched via registry, not command text)" 2 "red or unresolvable" "$sb" \
   "$(printf "$WRAPPER_CMD_TMPL" "$sb")"
 
+# ----------------------------------------------------------------------
+# #1121: the registry read inside _forge_kind_for (_lib-tracker.sh's
+# _tracker_project_value) needs `yq` OR `python3`+PyYAML to parse the
+# per-project override; when NEITHER is available it silently falls through
+# to the GLOBAL tracker.kind default ("gh") — wrong for a glab-registered
+# project's wrapper call. The two cases above only caught this by accident
+# of THIS environment happening to lack yq/PyYAML; make it deterministic by
+# stubbing yq/python3 to behave exactly like "no override found here"
+# regardless of what the host actually has installed, so this regression
+# can't silently stop being exercised if the CI image ever gains yq.
+# ----------------------------------------------------------------------
+
+# make_sandbox_wrapper_no_yaml_tools <glab_mode> — same registry/glab stub as
+# make_sandbox_wrapper, PLUS yq/python3 stubs that always fail to produce a
+# per-project override (matching real yq-absent / PyYAML-absent behaviour),
+# forcing block-merge-on-red-ci.sh's registry fallback (_registry_glab_fallback)
+# to be the thing that resolves the forge correctly.
+make_sandbox_wrapper_no_yaml_tools() {
+  local glab_mode="$1"
+  local sb
+  sb=$(make_sandbox_wrapper "$glab_mode")
+  cat > "$sb/bin/yq" <<'EOF'
+#!/bin/bash
+# Simulates yq being unusable for this lookup (matches "yq not installed"
+# from the caller's perspective: no stdout, non-zero exit).
+exit 1
+EOF
+  chmod +x "$sb/bin/yq"
+  cat > "$sb/bin/python3" <<'EOF'
+#!/bin/bash
+# Simulates python3 without PyYAML installed. The real
+# _tracker_project_value heredoc catches the ImportError and exits 0 with
+# no stdout; this stub reproduces that exact observable behaviour.
+exit 0
+EOF
+  chmod +x "$sb/bin/python3"
+  echo "$sb"
+}
+
+sb=$(make_sandbox_wrapper_no_yaml_tools success)
+run_case "#1121: wrapper, glab-registered project, NO yq/PyYAML -> still allows on green pipeline (registry fallback engages)" 0 "" "$sb" \
+  "$(printf "$WRAPPER_CMD_TMPL" "$sb")"
+
+sb=$(make_sandbox_wrapper_no_yaml_tools failure)
+run_case "#1121: wrapper, glab-registered project, NO yq/PyYAML -> still blocks on red pipeline (registry fallback engages)" 2 "red or unresolvable" "$sb" \
+  "$(printf "$WRAPPER_CMD_TMPL" "$sb")"
+
+# make_sandbox_wrapper_gh_no_yaml_tools <gh_mode> — a GENUINELY gh-kind
+# project (no tracker: override at all — relies on the global default),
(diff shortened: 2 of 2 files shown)


me2resh (commented):
## Security Review: PR #1122

**Commit**: `29c5b8d683b03e920c77fa17fe7ab4a67828e499`

### Summary

The PR touches a merge-gate CONTROL (`.claude/hooks/block-merge-on-red-ci.sh`, AgDR-0104/0109). It fixes a wrong-forge dispatch: a `glab`-registered project merged through the `tracker_pr_merge` wrapper (#759) had its CI check routed to `gh pr checks` instead of the GitLab pipeline whenever the registry read fell back to the global `gh` default (no `yq`/PyYAML installed). The fix adds `_registry_glab_fallback`, a dependency-free `grep`/`awk` scan of `apexyard.projects.yaml`, consulted **only** when the primary resolver answers `"gh"` for the wrapper shape, that can upgrade an ambiguous `gh` → confirmed `glab`.

I reviewed adversarially with the specific question: can this weaken the gate or let a red-CI state through? It cannot.

### Checklist Results
- Secrets & Credentials:  Pass (no secrets touched)
- Auth & Authorisation:   Pass (gate/trust-chain integrity preserved)
- Data Protection:        N/A
- API Security:           N/A
- Gate & Trust-Chain Integrity (#777): Pass

### Adversarial cases run

**1. Fail-closed posture preserved for BOTH forges — CONFIRMED.**
The diff adds only a forge-*selection* block; it does not add/alter a single `exit`, CI-status resolution, or allow/block branch. After forge dispatch the glab path (allow on `success`/`none`; block on `pending`/`failure`/`""`) and the gh path (allow on rc=0 / "no checks reported"; block otherwise) are byte-identical to pre-#1121. Empirically: the 8 glab garbage-input cases (unresolvable, non-zero-with-stdout, 401 envelope, HTML page, truncated JSON, bare scalar, JSON array) all BLOCK, plus jq-broken → BLOCK. 31/31 on the PR branch.

**2. Mis-route cannot open the gate — CONFIRMED (both directions fail CLOSED).**
- The fallback fires **only** when `FORGE == "gh"` and echoes `"glab"` **only** when the target repo's own registry block carries an explicit `kind: glab`. It can never downgrade a real `glab`, and never fires for non-wrapper shapes (those dispatch on trustworthy command text).
- gh → glab (hypothetical false upgrade): a genuine gh project routed to glab → `glab mr view` on a GitHub repo errors → status `""` → **BLOCK**. New test `gh-kind project (no tracker override), NO yq/PyYAML` proves the upgrade does *not* fire (the loud-failing `glab` stub is never called) and the project stays on the gh path — allow on green, block on red.
- glab → gh (fallback misses a glab project): `FORGE` stays `gh` → `gh pr checks` on a GitLab repo errors → non-zero, no "no checks reported" → **BLOCK**.
- Both mis-route directions block; neither opens.

**3. No injection via the registry parse — CONFIRMED.**
The awk program contains no `system()`, `getline`, `| command`, or `eval` (grep-verified). The command-derived repo slug is passed as `awk -v target=...` (data, not code); `apexyard.projects.yaml` content is read purely as string data; the only output is the literal `"glab"` or empty, captured into a shell var and string-compared (`[ "$found" = "glab" ]`) — no `eval`, no command substitution of registry content. A crafted slug/YAML entry can at most flip the string match, which (per case 2) fails closed. No command-injection or attacker-chosen-forge path exists.

**4. Does any red-CI state slip through that previously blocked? — NO.**
The only BLOCK → ALLOW transitions the fix enables are a **genuinely green** glab pipeline (`success`) or a **genuinely absent** one (`none`, parity with gh's "no checks reported" no-op). A red glab pipeline resolves to `failure` → BLOCK (resolver unchanged). Empirically: `wrapper, glab-registered, NO yq/PyYAML → blocks on red pipeline` passes. Before the fix the broken case *over-blocked* (wrong forge erroring), so this fix only removes false blocks on truly-green MRs — it introduces no new allow path for any non-green state.

### Security Issues Found

None blocking.

**LOW / informational (for the shared-resolver epic, not this PR):** `_registry_glab_fallback`'s awk takes the *last* `kind:` seen within a project block and does not scope `kind:` to its `tracker:` parent. A block that (outside the documented schema) carried a non-`tracker` nested `kind: glab` after/instead of the tracker kind could be read as glab. This is outside `apexyard.projects.yaml.example`'s shape and — critically — **fails closed** (a mislabeled gh project → glab path → glab errors → BLOCK), so it cannot weaken the gate. Noted only because the eventual `_lib-tracker.sh` consolidation should parse `tracker.kind` structurally rather than "last kind wins".

### Recommendations
- When the shared-resolver epic folds this back into `_lib-tracker.sh`, scope the `kind:` match to the `tracker:` sub-block so the parser is exact rather than fail-closed-imprecise.

### Verdict
**SECURE** — the change only alters WHICH forge is queried for the wrapper shape; the allow/block decision logic is untouched, mis-routes fail closed in both directions, the registry parse is injection-free, and no red/unresolvable CI state that previously blocked can now merge. 31/31 gate tests pass, including the 4 new deterministic yq/PyYAML-absent cases.

---
📌 Reviewed commit: `29c5b8d683b03e920c77fa17fe7ab4a67828e499`

me2resh (commented):
## Code Review: PR #1122

**Commit**: `29c5b8d683b03e920c77fa17fe7ab4a67828e499`

### Verdict
**CHANGES REQUESTED**

A blocking correctness defect in `_registry_glab_fallback`: the fallback only resolves `glab` when the glab-registered project is the **last** entry in `projects:`. For any glab project that has another entry after it — the normal shape of a mixed-forge portfolio — it returns a false miss, so `FORGE` stays `gh` and the gate still checks the wrong forge. i.e. #1121 is not actually fixed for that (very plausible) configuration, and the new tests don't cover it because their registry has the glab project as the sole entry.

The root-cause analysis, the localized/fail-closed posture, the sibling-suite regression coverage, and the PR hygiene are all solid — this is a one-line bug in an otherwise well-reasoned patch.

---

### Summary
Fixes the wrapper-shape forge dispatch in `block-merge-on-red-ci.sh` (a merge-gate CONTROL hook). When a project is registered `tracker: kind: glab` and merged via the tracker-agnostic `tracker_pr_merge` wrapper, the hook resolves the forge from the registry via `_forge_kind_for` → `tracker_kind` → `_tracker_project_value`. That resolver needs `yq` or `python3`+PyYAML; with neither installed it silently falls back to the global `gh` default, so the gate checks GitHub CI for a GitLab project. The PR adds a dependency-free (grep/awk) `_registry_glab_fallback` consulted only when the primary resolver answers `gh` on a wrapper-shape call.

### Checklist Results
- Code Quality:             Fail (boundary-rule bug — see Issue 1)
- Testing:                  Fail (untested position; see Issue 1)
- Security:                 Pass (fails closed either way; reliability not security)
- PR Description & Glossary: Pass (narrative Summary, `Refs #1121`, Glossary all present)
- Technical Decisions (AgDR): N/A (bug fix using existing grep/awk patterns; no new dependency/tech; `_lib-tracker.sh` correctly left untouched per #1102)

### Verified
- **Root-cause understanding is correct.** Confirmed `_tracker_project_value` (`.claude/hooks/_lib-tracker.sh`) reads via `yq` (line 199) then `python3`+PyYAML (line 207); with neither, it returns 1 and `tracker_kind` falls through to the global `.tracker.kind` default `gh` (line 258). On my box `yq` v4.53.3 is present (PyYAML is not), which is exactly why the two original `wrapper: glab-registered project` cases pass on `dev` — a pure accident of local tooling, as the PR describes.
- **Fail-closed posture preserved**, both forge paths unchanged.
- **Sibling suites green** (ran locally, all unchanged by this PR): `test_block_unreviewed_merge` 41/0, `test_extract_pr` 24/0, `test_extract_repo_fork_scoping` 3/0, `test_forge_aware_extract_pr` 48/0, `test_require_architecture_review` 24/0, `test_require_posted_review` 12/0, `test_require_design_review_for_ui` 24/0.
- **shellcheck --severity=warning clean** on both changed files.
- **The 4 new tests are deterministic** — they stub `yq`/`python3` inside each sandbox rather than relying on ambient absence. Verified the patched suite passes **31/0 both** with `yq` present and with `yq`/`python3` forced off PATH. Also confirmed the two positive `#1121` cases genuinely exercise the fix: running the new test file against the **old** (fallback-less) hook fails those two cases (29/2), while the two gh-kind no-false-positive guards pass regardless — good test design.

### Issues Found

**1. [BLOCKING] `_registry_glab_fallback` returns a false miss whenever the glab project is not the last registry entry.**

`block-merge-on-red-ci.sh`, the awk in `_registry_glab_fallback`. The boundary branch does:

```
if (ind == item_indent) {
  if (block_repo == target && block_kind == "glab") { print "glab"; exit }
  block_repo = ""; block_kind = ""
}
```

In awk, `exit` runs the `END` block. When the match is found at a **boundary** (i.e. a subsequent `- name:` entry closes the matching block), `block_repo`/`block_kind` have **not** been reset before `exit`, so `END` re-prints `glab`. awk therefore emits **two** lines, `glab\nglab`. The function then does `[ "$found" = "glab" ]`, which fails on the two-line value, so it returns 1 (miss).

Consequence: for a glab-registered project with any entry after it (the normal mixed-forge shape), with `yq`/PyYAML absent, `_registry_glab_fallback` misses → `FORGE` stays `gh` → the hook runs `gh pr checks` against a GitLab project. That is the exact #1121 symptom, still unfixed for that project. It fails closed (blocks), so it is reliability, not security — but the PR claims to fix #1121 and only fixes the narrow case where the glab project is the registry's last entry.

Reproduction (function-level, `yq`/`python3` absent):

```
glab is FIRST entry, target g/p:  result='' rc=1        <- WRONG (should be glab)
glab is MIDDLE entry, target g/b: result='' rc=1        <- WRONG (should be glab)
glab is SOLE entry (shipped test shape): result='glab'  <- only case the tests cover
```

Raw awk stdout for the glab-first case: `g l a b \n g l a b \n` (od-verified two lines).

The shipped tests pass only because `make_sandbox_wrapper` registers the glab project as the **sole** entry, so it is trivially last and the END-only path is taken.

Suggested fix (either):
- Collapse the awk output to one line: `found=$(awk ... | head -n1)`; or
- Don't `print` inside the loop — set a `printed=1` flag and `exit`, then in `END` do `if (printed || (block_repo==target && block_kind=="glab")) print "glab"` guarded so it prints exactly once.

And add a test case where the glab project is **not** the last entry (e.g. glab first, a gh project after) — that case fails today and would have caught this.

- **nit:** the awk matches any `kind:` line inside a project block, not specifically `tracker.kind`. The current documented schema (`apexyard.projects.yaml.example`) has no other `kind:` field so this is theoretical, but since you're revising the awk anyway, anchoring the match to a `tracker:` sub-block would make it robust to future schema fields.
- **nit:** `item_indent` is captured from the first `- ` list item anywhere in the file, not scoped to the `projects:` list. Fine for the documented schema (projects is the primary list), but a preceding top-level list at a different indent could misalign block boundaries. Worth a comment noting the assumption.

### Scope judgment (localized vs shared-lib)

The decision to keep this hook-local rather than touch `_lib-tracker.sh` is the right call for this PR (that lib is #1102 territory). But the same `yq`/PyYAML-missing root cause affects **every** forge-dispatching gate that routes through `tracker_kind` / `_tracker_project_value` — `block-unreviewed-merge.sh`, `require-architecture-review.sh`, the `tracker_pr_merge` / review / list / create wrappers themselves. Patching only this one hook means all the other gates still mis-resolve the forge in the identical environment, so the tree reads as "fixed" while siblings remain broken.

I could not find a follow-up ticket referenced for the shared-lib root cause. Please **file a follow-up** tracking either: (a) a dependency-free `tracker.kind` fallback inside `_lib-tracker.sh` that fixes all callers at once, or (b) the far cheaper option — have the CI workflow (`.github/workflows/tests.yml`) install `yq` (one line), which makes the primary resolver always work and would have prevented this class of bug entirely — and link it in the PR body so the residual gap is visible rather than implicit. This is a secondary requirement; Issue 1 is the blocker.

---
📌 Reviewed commit: `29c5b8d683b03e920c77fa17fe7ab4a67828e499`

me2resh (commented):
## Security Re-Review: PR #1122 (delta 29c5b8d → 2aa1e9a)

**Commit**: `2aa1e9a1d82a4c72a8fa5af94bc93d43d0af143b`
**Class**: trust-chain CONTROL hook (`block-merge-on-red-ci.sh`) — reviewed at full depth, delta-focused.

### Summary
Re-review triggered by a new commit that invalidated the prior SECURE marker (SHA no longer matches forge HEAD). The delta is confined to `_registry_glab_fallback`'s awk program (+ 3 regression tests):

- BEGIN sets `emitted = 0`; the mid-stream match becomes `print "glab"; emitted = 1; exit`; the END print gains a `!emitted &&` guard.
- Purpose: fix a **position-dependent double-print**. awk's `exit` runs the END block, and `block_repo`/`block_kind` were still set, so a mid-stream hit printed `glab` twice → the caller's exact-match `[ "$found" = "glab" ]` compared against `"glab\nglab"` and MISSED whenever the glab target block was not the last registry entry.

### Delta security analysis (the three questions asked)

**1. Can this fail OPEN (allow a merge it should block)? No.**
The `emitted` guard only ever *adds* a suppression condition (`!emitted`) to the END `print` — it can never *cause* a print. Both print sites still gate on `block_repo == target && block_kind == "glab"`, so a non-glab / non-target repo can never yield `"glab"`. The change fixes a false-**negative**, and that false-negative was already **fail-closed**: a missed glab confirmation left `FORGE="gh"`, sending a glab project down the gh path → `gh pr checks` on a non-gh repo → unresolvable → BLOCK. Direction of the bug was a spurious *block* of a green glab pipeline, never a spurious *allow*.

Confirmed the fallback can only ever **upgrade `gh`→`glab`**, never override a genuine result: it is consulted only inside `if [ "$FORGE" = "gh" ]` (line ~299), and the `if [ "$REGISTRY_FORGE" = "glab" ]` branch is the sole mutation. A genuine `glab` primary result never enters this branch; a `gh` result can only be upgraded on an explicit registry `kind: glab`.

**2. Did the change widen the awk/registry injection surface? No.**
The added logic is three internal integer-flag operations. It introduces no new consumption of attacker-controlled input. `target` is passed via awk `-v` (no shell re-eval) and used only in string equality, never in a regex or `system()`; registry contents are read as a data file, values used only in compares and the two hardcoded `print "glab"` statements. No `system()`, no `eval`, no dynamic regex from file content — unchanged by this delta. The author also kept the added comments apostrophe-free (verified: 0 added lines contain `'`), preserving the single-quoted awk-string integrity; a hypothetical slip would have broken awk syntax and — via `2>/dev/null` + the exact-match compare — failed closed anyway.

**3. Any behavior change beyond emit-once? No.**
The allow/block decision logic, CI-status resolution (gh and glab paths), and the fail-closed posture are byte-identical. The delta changes only how many times the fallback prints its match (now exactly one, position-independent).

### Every failure mode still fails CLOSED
- Registry missing / repo absent / awk error (`2>/dev/null`) → `found=""` → miss → `FORGE` stays `gh` → gh path; if genuinely glab, gh check on a non-gh repo → unresolvable → BLOCK.
- Fallback confirms glab → glab path → `success` allows; `pending | failure | "" (unresolvable)` all BLOCK.
- False-positive glab upgrade of a genuine gh project (not introduced here, but tested) → glab path → `glab mr view` unresolvable → BLOCK.

Adversarially verified the emit-once + no-false-positive invariants directly against the function: target-gh-with-later-glab → correct MISS; glab-not-last-with-trailing-glab → exactly one `glab`; glab-last → exactly one `glab`; target-absent → MISS.

### Prior informational item — unchanged
Last review noted the awk matches any `kind:` line, not scoped to a `tracker.` parent. The `if (line ~ /^kind:/)` matcher is **untouched** by this delta, so the item stands as-is — and still fails closed (a stray `kind: glab` at wrong nesting could only upgrade gh→glab, which BLOCKS a genuine gh project via the unresolvable glab path). Informational only; not a blocker.

### Checklist Results
- Injection Prevention:   Pass (awk `-v` var, data-file parse, no `system`/`eval`; surface unchanged)
- Auth & Authorisation:   Pass (gate posture preserved)
- Gate & Trust-Chain Integrity (#777): Pass — no fail-open, no gate weakened; emit-once is a pure correctness fix

### Verification
- Gate suite: **34 passed, 0 failed** (isolated detached worktree at `2aa1e9a`; working tree left untouched — no `gh pr checkout`).
- `shellcheck --severity=warning`: CLEAN on both changed files.

### Verdict
**SECURE — APPROVED (comment).** No new security issue. The emit-once fix introduces no fail-open path and does not widen the injection surface; all failure modes remain fail-closed.

---
📌 Reviewed commit: `2aa1e9a1d82a4c72a8fa5af94bc93d43d0af143b`

me2resh (commented):
## Code Review: PR #1122 (re-review at new HEAD)

**Commit**: `2aa1e9a1d82a4c72a8fa5af94bc93d43d0af143b`

Re-review after the prior CHANGES REQUESTED verdict (at 29c5b8d). Reviewed at **Heavy** depth — this is a trust-chain CONTROL hook (`.claude/hooks/block-merge-on-red-ci.sh`, AgDR-0104/0109).

### Summary

The prior blocking bug is fixed correctly. `_registry_glab_fallback`'s awk previously emitted the `glab` match **twice** whenever the target glab block was not the last registry entry — once mid-stream (`print "glab"; exit`) and again in the `END` block (awk `exit` runs `END`, and `block_repo`/`block_kind` were still set). The caller's exact `[ "$found" = "glab" ]` compare then saw `"glab\nglab"` and missed, leaving `FORGE=gh` for a glab-registered project — a false MISS (fails closed, but #1121 was not actually fixed for a mixed-forge portfolio where glab is not-last). The single-entry test registry hid it.

The fix adds an `emitted` flag: the mid-stream hit sets `emitted=1` before `exit`, and `END` is guarded `if (!emitted && ...)`. Mid-stream and END prints are now mutually exclusive, so a matching glab block emits exactly one `"glab"` regardless of its position.

### Independent verification (I did not rely on the PR summary)

- **Trace + test matrix.** Confirmed position-independence by trace and by the shipped case matrix: glab FIRST, MIDDLE, LAST, and sole all resolve to a single `"glab"`. Mid-stream sets `emitted` → `END` no-op; glab-last leaves `emitted=0` → `END` fires once. No path double-prints.
- **Hook diff is minimal and scoped.** `git diff 29c5b8d..2aa1e9a` on the hook is *exactly* three changes: `emitted=0` in `BEGIN`, `emitted=1` before the mid-stream `exit`, and the `!emitted` guard in `END` (plus comments). The allow/block decision logic, the gh path (byte-identical to pre-#790), the glab path, the jq-fail-closed path, and the forge-selection block (`_forge_kind_for` + fallback invocation, added by the parent commit) are **untouched**. Only the awk emit-once behaviour changed — matching the claim.
- **New tests are not tautologies.** Ran the PR-head test file against the **base** hook (29c5b8d): exactly the 3 new position cases fail (**31 passed, 3 failed**), and against the PR-head hook all pass (**34 passed, 0 failed**). The base-hook failures show the real defect: a glab project with a *green* pipeline is wrongly BLOCKED via the gh path (rc=2, "PR #500 has red CI"), and the red case blocks with the wrong gh-path message rather than the glab "red or unresolvable" message — i.e. the tests distinguish *which forge was dispatched*, not just the exit code.
- **No sibling regression.** `test_block_unreviewed_merge` 41/41, `test_extract_pr` 24/24, `test_extract_repo_fork_scoping` 3/3, `test_forge_aware_extract_pr` 48/48, `test_require_architecture_review` 24/24, `test_require_design_review_for_ui` 24/24, `test_require_posted_review` 12/12.
- **Shellcheck.** Clean at `--severity=warning` on both changed files. The only `info`-level findings (SC1091 on the sourced lib; SC2086 on `$REPO_FLAG` at line 362) are in unchanged / pre-existing code — no new shellcheck class introduced.
- **Fail-closed posture unchanged.** The fix can only convert a `"glab\nglab"` miss into a correct single `"glab"` hit; it never removes a legitimate match and never fabricates one (the match predicate is unchanged). A miss still leaves `FORGE=gh`, and both forge paths fail closed on red/unresolvable. The gh-no-yaml-tools case confirms no false-positive `gh → glab`.

### Checklist Results
- Testing:                   Pass (regression cases proven to exercise the bug, not tautologies)
- Security:                  Pass (trust-chain control; fail-closed posture preserved, decision logic untouched)
- Performance:               Pass (dependency-free grep/awk scan, single pass)
- PR Description & Glossary: Pass (glossary present; summary bullets are narrative)
- Technical Decisions (AgDR): N/A — bug fix to awk emit-once; no new material decision (the hook-local-vs-shared-resolver call belongs to the parent commit and is unchanged here)

### Issues Found
None blocking.

- **nit (stale PR body):** the Testing section still reads "31 passed, 0 failed" / "Added 4 new cases" — that describes the parent commit (29c5b8d). This child commit adds 3 more position cases, so the suite is now **34 passed, 0 failed**. Worth refreshing the description for the record; no code impact.
- **nit (carried over, still non-blocking, both fail closed):**
  - the `kind:` capture (line 163, `/^kind:[ \t]*/`) matches any `kind:` within a block, not specifically `tracker.kind`. Harmless for the documented registry shape; can only ever over-confirm a literal `kind: glab`, which no gh project carries.
  - `item_indent` (line 141) is captured from the first `- ` list item seen anywhere. Fine given `projects:` is the first/outermost list; a pathological nested-list-first file would mis-set it, but only ever toward a MISS (→ stays gh), never toward a fabricated glab.
- **out of scope (noted, not re-blocking):** the underlying yq/PyYAML-optional gap in `_tracker_project_value` affects sibling forge-dispatching gates too. Correctly deferred to the shared-resolver epic per the author's scope judgment; the orchestrator is tracking it as a follow-up.

### Verdict
**APPROVED**

The prior blocking bug is genuinely fixed, the regression coverage provably exercises it across the FIRST/MIDDLE/LAST/sole positions, no allow/block logic changed, fail-closed posture intact, and no regressions or new shellcheck classes. No new blocking issue found.

---
Reviewed commit: `2aa1e9a1d82a4c72a8fa5af94bc93d43d0af143b`

atlas-apex (approved):

PR #1125, opened 2026-08-01
fix(#1102): centralize trust-chain lib self-location behind one anchored helper

me2resh:
## Summary

- **Centralized the self-location bootstrap guard.** `_lib-ops-root.sh` gains one shared function, `resolve_anchored_lib_dir()`, that every `_lib-*.sh` sibling-sourcing site now calls instead of hand-rolling its own `${BASH_SOURCE[0]:-...}` guard. The #1100 security review found four sites still using four *different* (two of them actively broken) spellings of the same cwd-substitution bug `#1100`/`#1062` fixed at nine other sites — one more per-site patch would just add a fifth spelling for the next grep to miss.
- **Closed the actual bug at all four sites** (`_lib-read-config.sh`, `_lib-extract-push-ref.sh`, `_lib-git-hooks-path.sh`, `_lib-protected-branches.sh`): under a non-bash sourcing shell (zsh — notably the shell the Claude Code Bash tool itself runs), `BASH_SOURCE[0]` is unset, `dirname "" -> "."`, and the resolved lib directory silently became the *caller's* `$PWD` instead of the script's own directory. If that cwd shipped a same-named sibling file, it got sourced unanchored — an attacker-controlled or coincidental cwd could smuggle arbitrary code into a trust-chain hook. Two of the four sites (`_lib-git-hooks-path.sh`, `_lib-protected-branches.sh`) used a `${BASH_SOURCE[0]:-$0}` fallback that looked like a fix but wasn't: `$0` in a sourced file under zsh is the sourcing path *as typed*, which inherits the identical hazard.
- **Diverged from the initial fix brief on one point, and explains why in AgDR-0118.** The brief also asked to make `_lib-read-config.sh`'s `git rev-parse --show-toplevel` config fallback anchored-by-default with an `APEXYARD_ALLOW_UNANCHORED_CONFIG=1` opt-in. I implemented that, then reverted it after it empirically broke real, already-shipped behavior: that fallback is reached by every *managed project's own* standalone git hooks (protected-branch checks, branch-name validation), none of which carry the ops-fork-specific anchor markers — so the "fix" silently degraded every managed project's own `.git.protected_branches[]` config to the framework default. That fallback was never actually reachable through the reported cwd-substitution bug in the first place (it resolves via `git rev-parse`, not `BASH_SOURCE`), so it's left unconditional/unchanged. Full reasoning in AgDR-0118 for human review — flagging this prominently since a reviewer may prefer a narrower anchor.
- **Extended the regression test** (`test_lib_self_location_cwd_anchor.sh`) with a discriminating impostor case per site, proving BASE (pre-fix, from git history) reproduces the bug and FIX closes it, plus an over-strictness guard that a genuine fork still resolves correctly.
- **Updated four pre-existing test sandboxes** that now need `_lib-ops-root.sh` as a sibling to reach the libs they were already testing (`test_lib_protected_branches.sh`, `test_githooks_pre_commit_protected_branch.sh`, `test_check_git_hooks_installed.sh`, `test_validate_branch_name_heredoc.sh`) — without the fixture update they silently exercised the "self-location unavailable" fallback path instead of the intended scenario.

## Testing

- `test_lib_self_location_cwd_anchor.sh`: **17/17 pass** — 5 pre-existing (unchanged sites) + 12 new. Each of the 4 newly-fixed sites has a BASE-reproduces / FIX-closes pair plus a genuine-fork-still-resolves guard.
- `test_lib_protected_branches.sh`: **16/16 pass** (was 12/16 before the fixture fix — the sandbox needed `_lib-ops-root.sh`).
- `test_githooks_pre_commit_protected_branch.sh`: **25/25 pass** (was 19/25 before the fixture fix).
- `test_check_git_hooks_installed.sh`: **9/9 pass** (was 6/9 before the fixture fix).
- `test_validate_branch_name_heredoc.sh`: **10/10 pass** (was 8/10 before the fixture fix).
- Full sibling sweep — every `test_lib_*.sh`, every `test_githooks_*`, `test_block_unreviewed_merge.sh`, `test_block_merge_on_red_ci.sh`, `test_pre_push_gate.sh`, `test_tracker_zsh_self_location.sh`, and all 121 files under `.claude/hooks/tests/`: **all pass**.
- `shellcheck --severity=warning` on every changed `.sh` file (5 lib files + 5 test files): clean, zero findings.

All counts taken with `CLAUDE_CODE_SESSION_ID` unset to avoid this session's own ops-root pin file masking the sandbox resolution (matches how CI actually runs these tests).

Refs #1102

---

## Glossary

| Term | Definition |
|------|------------|
| Self-location bootstrap | The pattern where a sourced shell lib finds its own directory (via `BASH_SOURCE[0]`) so it can source sibling files next to it |
| `BASH_SOURCE[0]` | A bash-only array that tracks the file currently executing; empty/unset under non-bash shells like zsh |
| Fail-closed | A guard that, on any doubt or failure, refuses to proceed rather than silently falling back to a less-safe default |
| Anchor / anchored | A check that a resolved path is genuinely trustworthy (here: that a directory really is the calling script's own location, not a substituted cwd) |
| Impostor cwd | A directory an attacker (or coincidence) controls that ships a same-named sibling file, hoping an unanchored self-location resolves there instead of the real script's directory |

### DIFF
--- a/.claude/hooks/_lib-extract-push-ref.sh
+++ b/.claude/hooks/_lib-extract-push-ref.sh
@@ -59,9 +59,36 @@
 #
 # Refs: me2resh/apexyard#194, me2resh/apexyard#547, me2resh/apexyard#1066
 
-HOOK_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# SELF-LOCATION BOOTSTRAP (me2resh/apexyard#1102 / AgDR-0118)
+# ------------------------------------------------------------
+# Was `HOOK_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` -- no
+# `:-` guard at all. Under a non-bash sourcing shell (zsh), BASH_SOURCE is
+# unset/empty, so `dirname ""` -> `.` and HOOK_LIB_DIR silently became the
+# CALLER's $PWD. If that cwd happened to ship a same-named
+# `_lib-strip-heredoc.sh`, it would be sourced UNANCHORED. The bootstrap
+# guard below (raw BASH_SOURCE[0] captured once, empty short-circuits to no
+# self-location) plus `resolve_anchored_lib_dir` in _lib-ops-root.sh (the
+# anchor check) are the ONE shared idiom every _lib-*.sh self-location site
+# now uses -- see that function's header for why the very first hop can
+# never itself be centralized behind a sourced call.
+_RAW_BASH_SOURCE_0="${BASH_SOURCE[0]:-}"
+HOOK_LIB_DIR=""
+if [ -n "$_RAW_BASH_SOURCE_0" ]; then
+  _CANDIDATE_LIB_DIR="$(cd "$(dirname "$_RAW_BASH_SOURCE_0")" 2>/dev/null && pwd)"
+  if [ -n "$_CANDIDATE_LIB_DIR" ] && [ -f "$_CANDIDATE_LIB_DIR/_lib-ops-root.sh" ]; then
+    # shellcheck source=./_lib-ops-root.sh
+    # shellcheck disable=SC1091
+    . "$_CANDIDATE_LIB_DIR/_lib-ops-root.sh"
+    if command -v resolve_anchored_lib_dir >/dev/null 2>&1; then
+      HOOK_LIB_DIR="$(resolve_anchored_lib_dir "$_RAW_BASH_SOURCE_0")"
+    fi
+  fi
+  unset _CANDIDATE_LIB_DIR
+fi
+unset _RAW_BASH_SOURCE_0
+
 # shellcheck source=./_lib-strip-heredoc.sh
-if [ -f "$HOOK_LIB_DIR/_lib-strip-heredoc.sh" ]; then
+if [ -n "$HOOK_LIB_DIR" ] && [ -f "$HOOK_LIB_DIR/_lib-strip-heredoc.sh" ]; then
   # shellcheck disable=SC1091
   . "$HOOK_LIB_DIR/_lib-strip-heredoc.sh"
 fi
--- a/.claude/hooks/_lib-git-hooks-path.sh
+++ b/.claude/hooks/_lib-git-hooks-path.sh
@@ -18,8 +18,35 @@
 # _lib-path-resolve.sh's header comment for the full rationale, including
 # why it is NOT the same thing as _lib-portfolio-paths.sh's
 # `_portfolio_canonicalize` (a deliberately different, fail-soft sibling).
-_LGHP_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
-if [ -f "$_LGHP_LIB_DIR/_lib-path-resolve.sh" ]; then
+# SELF-LOCATION BOOTSTRAP (me2resh/apexyard#1102 / AgDR-0118)
+# ------------------------------------------------------------
+# Was `_LGHP_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"` --
+# the `:-$0` fallback is a FAKE fix: under zsh, $0 in a sourced file is the
+# sourcing invocation path AS TYPED, not a reliable self-location signal, so
+# it inherits the exact same "dirname resolves to the caller's cwd" hazard
+# `:-` alone has. The bootstrap guard below (raw BASH_SOURCE[0] captured
+# once, empty short-circuits to no self-location, never $0) plus
+# `resolve_anchored_lib_dir` in _lib-ops-root.sh (the anchor check) are the
+# ONE shared idiom every _lib-*.sh self-location site now uses -- see that
+# function's header for why the very first hop can never itself be
+# centralized behind a sourced call.
+_LGHP_RAW_BASH_SOURCE_0="${BASH_SOURCE[0]:-}"
+_LGHP_LIB_DIR=""
+if [ -n "$_LGHP_RAW_BASH_SOURCE_0" ]; then
+  _LGHP_CANDIDATE_DIR="$(cd "$(dirname "$_LGHP_RAW_BASH_SOURCE_0")" 2>/dev/null && pwd)"
+  if [ -n "$_LGHP_CANDIDATE_DIR" ] && [ -f "$_LGHP_CANDIDATE_DIR/_lib-ops-root.sh" ]; then
+    # shellcheck source=./_lib-ops-root.sh
+    # shellcheck disable=SC1091
+    . "$_LGHP_CANDIDATE_DIR/_lib-ops-root.sh"
+    if command -v resolve_anchored_lib_dir >/dev/null 2>&1; then
+      _LGHP_LIB_DIR="$(resolve_anchored_lib_dir "$_LGHP_RAW_BASH_SOURCE_0")"
+    fi
+  fi
+  unset _LGHP_CANDIDATE_DIR
+fi
+unset _LGHP_RAW_BASH_SOURCE_0
+
+if [ -n "$_LGHP_LIB_DIR" ] && [ -f "$_LGHP_LIB_DIR/_lib-path-resolve.sh" ]; then
   # shellcheck source=/dev/null
   . "$_LGHP_LIB_DIR/_lib-path-resolve.sh"
 else
--- a/.claude/hooks/_lib-ops-root.sh
+++ b/.claude/hooks/_lib-ops-root.sh
@@ -143,6 +143,78 @@ _ops_root_anchor_valid() {
   return 1
 }
 
+# ------------------------------------------------------------------------
+# resolve_anchored_lib_dir RAW_BASH_SOURCE_0
+# ------------------------------------------------------------------------
+# THE single self-location entry point every _lib-*.sh sibling-sourcing call
+# site should use, replacing the four independent spellings a #1100 security
+# review found still in the wild (me2resh/apexyard#1102 / AgDR-0118):
+#   - ${BASH_SOURCE[0]:-}          (no anchored fallback for the empty case)
+#   - ${BASH_SOURCE[0]}            (no `:-` guard at all)
+#   - ${BASH_SOURCE[0]:-$0}        (a FAKE fix -- $0 in a sourced file under
+#                                    zsh is the sourcing path AS TYPED, not a
+#                                    reliable self-location signal; it
+#                                    inherits the exact same hazard)
+#
+# RAW_BASH_SOURCE_0 must be the CALLER's own ${BASH_SOURCE[0]:-} value,
+# passed explicitly -- a function defined in a sourced lib cannot see the
+# caller's own array slot 0 (BASH_SOURCE[0] *inside this function* refers to
+# THIS file, not the caller's).
+#
+# The fix is a SINGLE fail-closed guard: when RAW_BASH_SOURCE_0 is empty
+# (the zsh case -- BASH_SOURCE is unset under a non-bash sourcing shell),
+# refuse to derive anything from it. `dirname ""` resolves to `.`, which
+# silently substitutes the CALLER's $PWD for the script's own directory --
+# exactly the hazard #1062/#1100 closed for nine other sites. This layer
+# can never itself be centralized behind a sourced function: to source
+# THIS file safely in the first place you must already have solved the
+# exact problem this layer solves (you can't call a function in a file you
+# haven't located yet). Every call site therefore still computes
+# RAW_BASH_SOURCE_0 = ${BASH_SOURCE[0]:-} inline, once, before it can even
+# reach this function -- see each site's own "SELF-LOCATION BOOTSTRAP"
+# comment. What #1102 removes is the DRIFT: four different (two of them
+# actively broken) spellings of this guard, collapsed into one function.
+#
+# NOT anchored to an ops-root marker (.apexyard-fork / onboarding.yaml +
+# apexyard.projects.yaml) — deliberately. An earlier version of this fix
+# added that as a second "defense in depth" layer, but three of the four
+# #1102 call sites (_lib-protected-branches.sh, _lib-git-hooks-path.sh,
+# _lib-extract-push-ref.sh) are hook libs deployed into EVERY project's own
+# `.claude/hooks/`, including managed projects that are never the apexyard
+# ops fork itself and correctly have neither marker. Requiring an ops-root
+# anchor on those sites silently broke real usage (protected-branches
+# config in a managed project resolved to the framework default instead of
+# the project's own `.git.protected_branches[]`, caught by
+# test_lib_protected_branches.sh) — over-strictness that fails in the
+# UNSAFE direction for a security control (fail-closed-to-a-DIFFERENT-
+# answer is not the same as fail-closed-to-blocked). Only
+# _lib-read-config.sh's OWN portfolio-root resolution genuinely wants the
+# ops-root anchor semantics, and it applies that anchor itself, inline,
+# rather than through this shared function -- see its own "FALLBACK —
+# ANCHORED BY DEFAULT" comment. The bootstrap guard alone (never substitute
+# cwd for an unavailable BASH_SOURCE[0]) is what actually closes the
+# reported bug; the ops-root anchor was scope creep beyond it.
+#
+# Echoes the verified directory and returns 0 on success. Returns 1 with
+# nothing echoed when RAW_BASH_SOURCE_0 is empty or unresolvable -- callers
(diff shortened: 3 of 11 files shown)


atlas-apex (approved):

me2resh (commented):
## Code Review: PR #1125

**Commit**: `54f434348e5a79e03facadc9b3882d391e5c377f`
**Scope**: `Full` (Heavy — trust-chain: five libs sourced by the merge gates)

### Summary

Closes the `BASH_SOURCE[0]` cwd-substitution self-location bug (the #1100/#1062 class) at the four remaining sites the #1100 review found in divergent spellings, by centralizing the fail-closed bootstrap guard behind one shared `resolve_anchored_lib_dir()` in `_lib-ops-root.sh`. Under a non-bash sourcing shell (zsh) `BASH_SOURCE[0]` is empty, `dirname "" -> "."`, and the resolved lib dir silently became the caller's `$PWD` — so a same-named sibling in an attacker-controlled/coincidental cwd could be sourced unanchored (code execution in a trust-chain hook). Two of the four sites used a `${BASH_SOURCE[0]:-$0}` fallback that only *looked* like a fix (`$0` in a sourced file under zsh is the sourcing path as typed). AgDR-0118 records the decision, including a conscious, well-argued reversal of one part of the brief.

### Checklist Results
- ✅ Architecture & Design:      Pass — one canonical guard; the irreducible per-site bootstrap is documented and justified
- ✅ Code Quality:               Pass — shellcheck --severity=warning clean on all 10 changed files
- ✅ Testing:                    Pass — discriminating BASE-reproduces/FIX-closes tests, verified genuine (not tautologies)
- ✅ Security:                   Pass — fails closed; no fail-open / anchor-bypass path found
- ✅ Performance:                Pass — N/A (a few cheap path ops at source time)
- ✅ PR Description & Glossary:  Pass — narrative Summary, `Refs #1102`, 5-term Glossary
- ⚠ Summary Bullet Narrative:   Pass — bullets are full narrative
- ✅ Technical Decisions (AgDR): Pass — AgDR-0118 present, no number collision (0119 merged separately to dev), accurately records the reversal
- ✅ Adopter Handbooks:          N/A — no handbooks loaded (shell diff)
- ⚠ Fallow Static Analysis:     N/A — not a JS/TS diff

### The escape-hatch reversal — VERDICT: sound, both claims verified

The author diverged from the brief: instead of anchoring `_lib-read-config.sh`'s `git rev-parse --show-toplevel` config fallback (behind `APEXYARD_ALLOW_UNANCHORED_CONFIG=1`), they implemented that, then reverted it. I independently verified both claims:

**Claim 1 — anchoring that fallback breaks real behavior: TRUE.**
`_lib-read-config.sh` is sourced transitively by `_lib-protected-branches.sh`, `validate-branch-name.sh`, `require-active-ticket.sh`, and the git-hooks install path — all of which run standalone inside every *managed project's own* git hooks, which correctly carry neither `.apexyard-fork` nor the `onboarding.yaml`+`apexyard.projects.yaml` pair. Gating the fallback on that anchor degrades a managed project's own `.git.protected_branches[]` to the framework default. Empirically confirmed: `test_lib_protected_branches.sh` builds sandboxes with a custom `protected_branches` config and **no** fork anchor (line 104 writes the config; nothing creates an anchor), so the config is reachable *only* via the git-toplevel fallback — and it passes **16/16** post-revert. Anchoring it would fail the custom-config cases (the author's cited 12/16). The reverted state is correct behavior, not a masked bug.

**Claim 2 — leaving it unanchored is NOT a regression: TRUE.**
The git-toplevel fallback only sets a `root` *data* value, used solely to build paths to `.claude/project-config.{defaults,}.json` that are read by `jq` (lines 160/166 → jq at 192/208/248). It never sources/executes anything root-derived. The *only* `source` in the file (`. "$candidate_dir/_lib-ops-root.sh"`, line 102) is the `BASH_SOURCE`-derived, bootstrap-guarded self-location step — `candidate_dir` comes from raw `BASH_SOURCE`, never from `root`. So the class of bug #1102 fixes (sourcing an impostor sibling → code execution) cannot flow through the `git rev-parse` fallback. `git rev-parse --show-toplevel` is bound to the repo the shell is actually in — reading a repo's own config is intended, not a cross-repo injection vector. Test case 12/13 confirms empirically: under zsh the decoy `_lib-ops-root.sh` is NOT sourced and resolution falls through to the impostor's *own* real repo (reading its own config = correct, not a leak). **I could not find any fail-open / anchor-bypass path here.** Not blocking.

### Other Heavy checks — all pass

- **`resolve_anchored_lib_dir()` fails closed.** Empty/unusable `raw` → `return 1` with nothing echoed; it never trusts `$0` and never falls back to cwd. Callers guard `[ -n "$dir" ]` before sourcing, so under zsh they skip the sibling entirely rather than sourcing from cwd. Safe under bash (BASH_SOURCE = real path) and zsh (empty → skip). Cannot fail-open to an attacker cwd.
- **No merge-gate regression.** `_lib-ops-root.sh` is **72 additions / 0 deletions** — a pure insertion between `_ops_root_anchor_valid` and `resolve_ops_root`. `resolve_ops_root()`, `resolve_ops_root_walk()`, and `_ops_root_anchor_valid()` are byte-for-byte unchanged, and `_lib-review-markers.sh` (marker-path resolution the gate + `/approve-merge` depend on) is not in the changeset. The merge gate is unaffected.
- **All four sites route through the helper.** Every residual `${BASH_SOURCE[0]:-$0}` in the four files is in a "Was …" doc comment, not live code; each has 3–4 live references to `resolve_anchored_lib_dir`.
- **Tests genuinely discriminate.** With the true pre-fix base ref, `test_lib_self_location_cwd_anchor.sh` is **17/17**: every BASE case reproduces the bug and every FIX case closes it, plus genuine-fork-still-resolves guards (not over-strict). The four fixture updates (ship `_lib-ops-root.sh` as a sibling) are legitimate — the new bootstrap must source ops-root before it can reach the target sibling — not weakened assertions.
- **Full sweep green.** CI "hook test suite" passes; local run 129+/129 PASS, 0 FAIL (still churning the skills tests, zero failures). Affected suites: protected_branches 16/16, githooks_pre_commit_protected_branch 25/25, check_git_hooks_installed 9/9, validate_branch_name_heredoc 10/10. `shellcheck --severity=warning` clean on all 10 changed files.

### Issues Found

None blocking. The security fix is correct, fails closed, is verified by discriminating tests, and touches nothing the merge gate depends on.

- **nit — test baseline fragility.** `test_lib_self_location_cwd_anchor.sh` defaults `BASE_REF=HEAD`. From a *committed* PR checkout on a zsh box, `git show HEAD:…` returns the fixed file, so the 4 "BASE reproduces the bug" assertions fail (I got 13/17); the claimed 17/17 only holds when the fix is *uncommitted* (working-tree=fix, HEAD=base). Consider skipping the BASE assertion when the BASE blob is byte-identical to the FIX file (`cmp -s`), or resolving `BASE_REF` to `git merge-base HEAD <default-branch>`, so the discriminator is robust to running from a committed tree. The shipped fix is correct regardless — the FIX assertions pass either way.
- **observation (pre-existing) — CI doesn't exercise these assertions.** The whole file `exit 0`s when `zsh` isn't installed (lines 32-35), and `tests.yml` on ubuntu-latest doesn't install zsh, so the "hook test suite" job passes by *skipping* this file — neither the 12 new cases nor the 5 pre-existing #1062 cases actually gate in CI. This mirrors the existing #1062 pattern, so it's not a new regression, but the "all 121 files pass" claim describes local runs, not CI coverage. A cheap follow-up (`apt-get install -y zsh` in `tests.yml`) would make these zsh-specific regressions actually gate.
- **nit — stale cross-reference comment.** `resolve_anchored_lib_dir`'s header in `_lib-ops-root.sh` still says `_lib-read-config.sh` "applies that anchor itself, inline … see its own 'FALLBACK — ANCHORED BY DEFAULT' comment." After the reversal that comment was renamed to "FALLBACK — DELIBERATELY LEFT git-repo-scoped, NOT ops-root-anchored," and read-config specifically does NOT anchor its git-toplevel fallback. The cross-reference points at a name that no longer exists and slightly misdescribes the final behavior — a leftover from the reverted draft worth a one-line fix.
- **suggestion — a fifth site is out of scope.** `maintain-docs-index.sh:37` still uses the fake-fix `${BASH_SOURCE[0]:-$0}` idiom and sources `_lib-read-config.sh`/`_lib-portfolio-paths.sh` before its enabled-gate. It's genuinely lower risk than the four fixed sites — it's a top-level hook executed under its own `#!/bin/bash` shebang (BASH_SOURCE reliably set → `:-$0` never triggers), not a `_lib-*.sh` sourced under zsh — and is out of this PR's declared scope. But given the AgDR's own thesis ("stop the drift / the next grep finds another"), aligning it onto `resolve_anchored_lib_dir` for consistency/defense-in-depth is a reasonable follow-up.

### Verdict
**APPROVED**

The core security fix is correct and fails closed; the escape-hatch reversal is a sound, thoroughly-documented decision (both claims independently verified, no fail-open path found); the new function is a pure addition that leaves the merge gate's ops-root/marker resolution untouched; and the regression tests genuinely discriminate. All findings are non-blocking test-hygiene / documentation / follow-up items.

---
📌 Reviewed commit: `54f434348e5a79e03facadc9b3882d391e5c377f`

me2resh (commented):
## Security Review: PR #1125

**Commit reviewed**: `54f434348e5a79e03facadc9b3882d391e5c377f`
**Class**: trust-chain CONTROL change (five `_lib-*.sh` sourced by the merge/push gates, incl. `_lib-ops-root.sh`). Reviewed at full depth in an isolated worktree at the exact forge HEAD; the real hooks + test suites were run against an adversarial matrix, not reasoned from prose.

### Summary
Centralizes the self-location bootstrap for four `_lib-*.sh` sites onto one shared `resolve_anchored_lib_dir()` in `_lib-ops-root.sh`, closing the `BASH_SOURCE[0]` cwd-substitution bug (#1100/#1062) in its remaining divergent spellings (`${BASH_SOURCE[0]}`, `${BASH_SOURCE[0]:-$0}`). Under a non-bash sourcing shell (zsh — the Claude Code Bash tool's shell) `BASH_SOURCE` is empty, `dirname "" → .`, and the old code silently self-located to the caller's `$PWD`, sourcing a same-named sibling from an attacker-influenceable cwd. The fix captures `${BASH_SOURCE[0]:-}` once and short-circuits to "self-location unavailable" when empty — never `$0`, never raw cwd.

### Checklist Results
- Secrets & Credentials:  Pass (none introduced)
- Auth & Authorisation:   Pass (merge-gate marker resolution unchanged — see Q3)
- Trust-chain / Gate integrity (#777): **Pass** — hardens a gate, weakens none

### Primary security questions

**Q1 — Does `resolve_anchored_lib_dir()` close the bug, fail-closed, at all four sites?  YES.**
Independently reproduced the attack (zsh + cwd with a malicious same-named sibling) against each site's own edited copy:
- `_lib-read-config.sh`: decoy `_lib-ops-root.sh` in cwd is **not** executed under zsh; resolution falls through to the current repo's own git-toplevel (verified the decoy's `resolve_ops_root() → /ATTACKER/PWNED` never ran).
- `_lib-protected-branches.sh`: falls to the safe default `main|master|dev|develop`.
- `_lib-extract-push-ref.sh` / `_lib-git-hooks-path.sh`: sibling simply not sourced.
The suite's BASE-reproduces / FIX-closes pairs confirm the pre-fix code was genuinely vulnerable at all four sites while the fix closes each (16/16 with the correct pre-fix base ref). The guard cannot be tricked via `$0` (never referenced — under zsh `${BASH_SOURCE[0]:-}` is always empty → fail closed), via a fork-shaped cwd (the helper does no ops-marker check, and under zsh the whole block is skipped), or via symlinks (`cd`+`pwd` canonicalizes; no exploitable angle).

*Residual, non-blocking:* under **bash** with a **bare-name** source (`source _lib-read-config.sh`) from an attacker cwd, `BASH_SOURCE[0]` is the bare name → `dirname → .` → cwd, and the sibling is sourced from cwd. I confirmed this is **pre-existing (byte-identical in BASE), not a regression**, and **not reachable via any production invocation**: every production source of these four libs is directory-qualified (`$HOOK_DIR/`, `$REPO_ROOT/.claude/hooks/`, `$(dirname "$0")/`). The fix's scope is the zsh vector, which it fully closes. A future hardening could reject a slashless `$raw` in `resolve_anchored_lib_dir()` to also close the bare-name-under-bash shape — LOW / informational, out of scope for #1102.

**Q2 — The escape-hatch reversal (unanchored `git rev-parse --show-toplevel` config fallback in `_lib-read-config.sh`).  The reversal is SOUND; no anchor-bypass.**
Both author claims independently verified:
- *(a) Not reachable via the cwd-substitution bug.* Correct. The fallback is `git rev-parse`-derived, not `BASH_SOURCE`-derived. Under zsh the vulnerable self-location step is skipped entirely; the fallback resolves to **the current repo's own toplevel and nothing else** — it cannot be pointed at an arbitrary attacker path. There is no path (non-bash shell, crafted cwd, git-repo boundary, symlink) where a caller reads config from a *different* repo than the one whose git-toplevel it is in. A repo reading its own config is not an impostor scenario; cross-repo config hijack is structurally impossible here because git-toplevel is scoped to the containing repo. (Nested-`.git` confusion would require pre-existing write access to the victim repo = already game over, and is unchanged by this PR.)
- *(b) Anchoring it breaks managed projects.* Confirmed. With the pin disabled (to avoid this session's pin masking the sandbox), a managed-project sandbox (plain git repo, no ops-fork markers, own `protected_branches`) reads its own config via the unanchored fallback (`main|release/prod|staging`); the simulated anchored variant returns an **empty** config root, degrading `protected_branches` to the framework default. The reversal was necessary. The zsh degradation is fail-closed **to the safe default**, never fail-open, and `.githooks/pre-push`/`pre-commit` (the git-native enforcement) run under `#!/bin/bash`, so BASH_SOURCE is populated there — the degradation only touches manual zsh sourcing, the documented "run under bash" trade-off.

**Q3 — Merge-gate regression.  None.**
`resolve_ops_root`, `resolve_ops_root_walk`, and `_ops_root_anchor_valid` are **byte-identical** BASE↔FIX; `_lib-ops-root.sh` is **+72/−0** (purely additive — a new function inserted, nothing existing touched). Marker-path resolution the merge gate + `/approve-merge` depend on is unchanged. Merge-gate suites pass: `test_block_unreviewed_merge.sh` 41/41, `test_block_merge_on_red_ci.sh` 27/27, `test_pre_push_gate.sh` 11/11.

**Q4 — Fail-closed everywhere.  Confirmed.**
Every failure mode of the new helper fails toward "refuse/anchored/safe-default": empty `${BASH_SOURCE[0]}` (zsh) → block skipped; `_lib-ops-root.sh` absent in candidate dir → `[ -f ]` guard skips; `resolve_anchored_lib_dir` undefined after source → `command -v` guard skips; `cd`/`dirname` failure → `|| return 1`, empty output. Each site's downstream guard (`[ -n "$LIB_DIR" ] && [ -f ... ]`) then skips the sibling — extract-push-ref skips heredoc-stripping, git-hooks-path installs an over-warning stand-in, protected-branches uses the hardcoded safe default, read-config falls to the current-repo git-toplevel. No path resolves to raw cwd or `$0`.

### Completeness
Change scope is exactly right (5 libs + 5 tests + 1 AgDR); the nine pre-existing #1100 sites are byte-untouched and retain the correct empty-guard idiom (spot-checked `_lib-tracker.sh`: `${BASH_SOURCE[0]:-}` + explicit `[ -z ... ] && hook_dir=""`). Grep confirms **no `.claude/hooks/_lib-*.sh` library file retains a broken self-location CODE spelling** — the four live instances were all caught. Remaining `:-$0` / unguarded `${BASH_SOURCE[0]}` matches are either comments documenting the old idiom or top-level **hook** scripts (`maintain-docs-index.sh`, `block-main-push.sh`, `detect-skill-intent.sh`, `reindex-on-session-start.sh`, `suggest-mcp-search.sh`) that execute under `#!/bin/bash` (BASH_SOURCE populated) and are outside the zsh-sourced-lib threat model this PR addresses — noted as a defense-in-depth follow-up, not a blocker. `shellcheck --severity=warning` clean on all five changed libs. Full test sweep: 119/120 green, the one flag being a detached-worktree artifact of the self-location suite's default base ref (passes 16/16 with the correct pre-fix base).

### Security Issues Found
None blocking. One LOW / informational hardening note (Q1 residual: bare-name-under-bash self-location — pre-existing, unreachable in production, out of scope).

### Verdict
**APPROVED (SECURE)** — the cwd-substitution bug is closed fail-closed at all four sites, the escape-hatch reversal is safe (git-toplevel is not cwd-substitutable and cannot be pointed at an attacker repo), and the merge-gate marker resolution is provably unchanged.

---
📌 Reviewed commit: `54f434348e5a79e03facadc9b3882d391e5c377f`

Issue #1013, opened 2026-07-25
[Chore] Memoise ops-root and config resolution — 48 hooks redo the same work on every Bash call (~1.6s)


atlas-apex:
## Driver

Every `Bash` tool call in an apexyard session pays a fixed ~1.6 second tax before the command runs, and roughly 92% of it is redundant work repeated 48 times.

Measured on this machine at v5.2.0:

| | |
|---|---|
| Hooks matching `Bash` in `.claude/settings.json` | **48** |
| Wall-clock for one sequential pass | **1617 ms** |
| Mean per hook | 33.7 ms |
| 48 bare no-op process spawns (baseline) | 128 ms |
| → process-spawn share of the cost | **8%** |
| → actual hook *logic* | **92%** |

The initial hypothesis was process-churn — 48 separate `bash` spawns. The measurement refutes it: spawning is only 128 ms of 1617 ms. The remaining ~1.5 s is each hook independently redoing the *same* setup — walking up the tree to locate the ops root, reading `project-config`, sourcing the same `_lib-*` helpers — 48 times per command, with 48 identical answers.

This is second-order next to model inference, so it is not the headline cause of "the framework feels slow". But it is pure waste, it compounds across an agent doing dozens of tool calls (~1.3 min of overhead over a 50-call agent), and it narrows the headroom before the harness's 600 s no-progress watchdog fires — a real failure mode: four sub-agents were killed by that watchdog in a single session.

## Scope

Memoise the shared resolution once per session (or once per process) instead of per-hook:

- **Ops-root resolution** — `resolve_ops_root` / the walk-up in `_lib-ops-root.sh`. The answer cannot change within a session; the session pin file already exists as a natural cache location.
- **Config reads** — `_lib-read-config.sh` re-parses `.claude/project-config.{defaults,}.json` per hook. Cache the merged result.
- **Portfolio path resolution** — `_lib-portfolio-paths.sh` helpers, same argument.

Suggested shape: a session-scoped cache file (or exported env populated by a `SessionStart` hook) that the shared libs consult before doing the expensive walk, with a cheap staleness check. Keep the fallback path intact so a cold or invalidated cache behaves exactly as today.

**Explicitly NOT in scope: removing or disabling any hook.** The gates are doing their job; this is about not recomputing the same three answers 48 times. No gate should be weakened, skipped, or made conditional by this change.

## Acceptance Criteria

- [ ] A measured before/after on the same benchmark (48-hook sequential pass), showing a material reduction
[...]

PR #1127, opened 2026-08-01
perf(#1013): memoise session-scoped ops-root + config resolution across hooks

me2resh:
## Summary

- **Added a session-scoped, cross-process cache** (`_lib-resolution-cache.sh`) for the two resolutions #1013 identified as redone 48 times per `Bash` call — the merged `project-config` JSON and each `portfolio_*` path (registry, workspace_dir, etc.) — so the disk-read/`jq`-merge/canonicalization chain for each runs once per session instead of once per hook process. Ops-root resolution already had this (the pin from #381); this PR only reuses it, never re-implements it.
- **Every cache entry is self-describing and fingerprinted** on (ops root, `project-config.defaults.json`'s mtime+size, `project-config.json`'s mtime+size). The fingerprint is recomputed fresh on *every* read, not once per session — a `.claude/project-config.json` edit mid-session invalidates the cache on the very next read, so the failure mode the issue was most worried about (a stale ops-root or config silently driving a gate decision) can't happen: a cache hit is provably as fresh as the files on disk right now.
- **`_portfolio_root()` now prefers the existing pin-aware `resolve_ops_root()`** over redoing its own walk-up, closing one instance of the "two hand-maintained resolvers for the same fact" pattern the issue's own Related section calls out — a small, low-risk consolidation, not a new mechanism.
- **Fails safe in every direction the acceptance criteria asked for**: no session id (git hooks/CI) → no-op; `APEXYARD_DISABLE_RESOLUTION_CACHE=1` → no-op; unwritable cache dir → silently degrades, value still computed correctly, no new error; cold or invalidated cache → byte-identical to the pre-#1013 code path (proven in tests, not just claimed).
- **AgDR-0120** records the design, including a narrow accepted trade-off: a portfolio path could theoretically go stale if a directory *along its resolution chain* has its symlink-ness change mid-session while the config itself is unchanged. This is much narrower than "wrong ops-root/config" (which the fingerprint fully closes) and is the kind of partial-and-honest trade the issue explicitly permits.

## Before / after — measured, two ways, reported honestly

**1. The issue's own methodology: every `PreToolUse` `Bash`-matcher hook, sequentially, wall-clock timed.** 49 hooks today (48 when the issue was filed; `nudge-control-adversarial-test.sh` shipped since via #1118). Run against `git status` and `git commit -m "..."` (mean of 5–6 runs each) on this dev machine, which is a shared, noisy box (Chrome, a Jamf/security agent, this very Claude session all competing for CPU — load average ~2.4–5.1 throughout):

| Scenario | `git status`, 49 hooks/pass | `git commit`, 49 hooks/pass |
|---|---|---|
| BEFORE (cache disabled) | 5891 ms | 6696 ms |
| AFTER, cold (1st call) | 5940 ms (-0.8%) | 6735 ms (-0.6%) |
| AFTER, warm (2nd+ call) | 5865 ms (+0.4%) | 6698 ms (-0.0%) |

**Read this honestly, not as "it doesn't work."** Only 13 of the 28 unique hook scripts behind these 49 entries touch `_lib-read-config.sh` / `_lib-portfolio-paths.sh` at all, and most of those call it once with a single filter — not the 7-call chain (`portfolio_validate`-shaped) this PR's caching is built around. For a single generic command, the full-pass wall-clock number is dominated by per-hook dispatch overhead (parsing the JSON payload, the settings.json wrapper, `git rev-parse` calls) that this PR does not touch, and on this specific machine right now that overhead swamps the saving in the aggregate — verified, not asserted; I checked which scripts source the libs before writing this paragraph.

**2. An isolated micro-benchmark of exactly the mechanism this PR changes** — the same 2 `config_get` calls + 5 `portfolio_*` resolvers `portfolio_validate()` exercises, run standalone (no unrelated hook dispatch, no network):

| Scenario | Mean over 20 processes | Single-call |
|---|---|---|
| BEFORE (cache disabled) | 670–677 ms/process | 727–759 ms (cold) |
| AFTER (warm, same session) | 493–498 ms/process | 495–502 ms (warm) |
| **Reduction** | **~26–27%**, reproduced across 2 independent runs | **~33–35%** |

**3. Deterministic, load-independent proof the mechanism itself is correct** — a logging `jq` shim shows the `jq -s '.[0] * .[1]'` merge call fires on a cold process and is **completely absent** on a warm one in the same session; a call-counting wrapper shows `_portfolio_get` (the jq + canonicalize chain) is invoked exactly once across two same-session processes, not twice. Both are `test_resolution_cache.sh` cases 2 and 3, not just claims — see Testing below.

**Net honest read:** the caching mechanism demonstrably works and eliminates exactly the redundant work #1013 diagnosed (proof #3), with a real, reproducible ~27–35% per-call saving on the specific resolution paths it targets (proof #2). Its contribution to a *full* 49-hook sequential pass for a single generic command is not clearly visible above this machine's current noise floor (proof #1) — the benefit scales with how many of the 49 hooks a given command actually activates and how many resolution calls each makes, which is highest for the merge-gate / migration / design-review hook families (21 of the 49 entries), not a plain `git status`.

## Testing

```bash
bash .claude/hooks/tests/test_resolution_cache.sh   # new — 9/9 passing
bash bin/run-hook-tests.sh                          # full suite, unchanged and green
shellcheck --severity=warning .claude/hooks/_lib-resolution-cache.sh .claude/hooks/_lib-read-config.sh .claude/hooks/_lib-portfolio-paths.sh .claude/hooks/tests/test_resolution_cache.sh bin/run-hook-tests.sh
```

`test_resolution_cache.sh` covers: cold cache == cache-disabled (byte-for-byte), cross-process reuse proven via instrumentation (a logging `jq` shim shows the second process never redoes the merge; a call-counting wrapper shows a portfolio resolver's compute path never re-runs), a `project-config.json` edit mid-session is picked up on the next read (never served stale), a forged/`UNKNOWN` fingerprint is never treated as valid, an ops-root mismatch is never trusted, an unwritable cache dir degrades with no crash, and both the disable flag and a missing session id produce zero cache files.

Refs #1013

---

## Glossary

| Term | Definition |
|------|------------|
| Fingerprint | A cheap signature (here: ops root + each config file's mtime and size) used to decide whether a cached value is still valid, without re-deriving the value itself |
| Session-scoped cache | A cache keyed on `$CLAUDE_CODE_SESSION_ID` — shared by every hook process spawned during one Claude Code session, discarded when the session ends |
| Cross-process cache | A cache that survives across separate process spawns (as opposed to a per-process cache, which is invisible once that process exits) — the layer this PR adds |
| Cold path | The original, unchanged resolution logic that runs on any cache miss — the fallback this PR never removes |
| Ops-root pin | The pre-existing (#381) mechanism that resolves and caches the apexyard fork's root directory for a session, via a file under `$HOME/.claude/apexyard` |


me2resh (commented):
## Security Review: PR #1127 — session-scoped resolution cache (trust-chain)

**Commit reviewed**: `d346c1c6b9a122bd0c1fef5dcac9bfc58635bd3d`
**Method**: adversarial sandbox in an isolated worktree — I built the attacks and ran them against the actual lib files, not from prose. `test_resolution_cache.sh` 9/9; full suite 136/137 (the 1 failure independently confirmed pre-existing).

### Summary

This adds a session-scoped, cross-process **file** cache (`_lib-resolution-cache.sh`) for the merged `project-config` JSON and the 8 `portfolio_*` resolved paths, wired into `_lib-read-config.sh` and `_lib-portfolio-paths.sh`. It caches config/filesystem-derived values only — never forge state (CI/review/merge eligibility). The whole risk is "a cache that serves a stale/wrong config or path a gate then trusts." I attacked exactly that. **No attack reached a gate with a weakened value using any capability an attacker would not already possess.**

### Checklist Results
- Secrets & Credentials: **Pass** — no secrets; cache stores config JSON + absolute paths only (already on disk/readable).
- Injection Prevention: **Pass** — cache `<name>` suffixes are hardcoded literals, never tool-derived; values are single-line paths / JSON blobs written via tmp+`mv`.
- Auth & Authorisation (gate integrity): **Pass** — see attack results; no gate weakened, cache is transparent (warm == cold).
- Data Protection: **Pass** — pin dir is `755`, same-user-write-only; no new secret exposure.
- Trust-chain / fail-safe: **Pass** — every degrade path is byte-identical to pre-#1013.

### Attacks run + results

**1. mtime+size-preserving stale-serve (the load-bearing test).**
- *Normal edit (size-changing):* cache correctly **invalidates** on the next read — a weakening edit (`protected_branches: [] `) was picked up fresh. ✓
- *Adversarial same-size + `touch -r` mtime-restore edit:* **stale-serve reproduced** — after swapping content while holding `(mtime,size)` constant (whole-second BSD `%m` granularity), the next read served the stale `protected_branches` value instead of the on-disk one. **This is NOT gate-weakening**, for four independent reasons: (i) it requires write access to `.claude/project-config.json` — the gate's *own source of truth* — so there is **no privilege gain**; (ii) in the attacker-weakening direction a stale-serve replays the **older, typically stronger** value → fail-safe, not fail-open; (iii) the cache can only ever replay a value that was **on disk this session**, never inject a never-written value; (iv) honest edits change byte size, so this never fires benignly. It is a well-known limitation of any `(mtime,size)` fingerprint, not a reachable bypass.
- *`UNKNOWN` token (stat unavailable):* refused on **both** read and write (`rc=1`); with `stat` forced to fail the value still resolves correctly via the cold path and **zero** files are written under `UNKNOWN`. ✓

**2. Symlink residual (the documented accepted trade).** Reproduced exactly as AgDR-0120 describes — and the direction is *fail-safe*: after a mid-session dir→symlink swap, the cache returns the **stale original honest path**, while it is the *un-cached* fresh resolve that follows the attacker's symlink to `/EVIL`. The swap itself requires **in-fork write access** (already game-over). Genuinely narrow, honestly documented, non-escalating. I agree it is acceptable.

**3. The cache FILE as attack surface.** Location `${APEXYARD_OPS_PIN_DIR:-$HOME/.claude/apexyard}` = **`drwxr-xr-x`, same-user-write-only** — the **exact same directory** that already holds the `ops-root-<session>` pins `resolve_ops_root()` (and thus the merge gate) already trust (#381). This PR **reuses that pre-existing trust boundary; it does not introduce a lower one.**
  - Malformed / empty / single-line / binary / wrong-fingerprint cache files → **fail CLOSED** (re-resolve to the correct value, no crash). ✓
  - A forged file carrying the *correct* fingerprint **is** served — but forging it requires **same-user write to the pin dir**, and an attacker with that already has strictly more powerful primitives (poison the pre-existing ops-root pin, edit `project-config.json` directly, write `*.approved` markers, edit the hooks). **No escalation, no cross-user/cross-repo/lower-priv path.** A crafted repo cloned into `workspace/` cannot write there; a different-UID process cannot write a 755 `$HOME` dir.
  - Session-id keying in the filename isolates per-session → **no cross-session value poison**. ✓

**4. Fail-safe degrade — all byte-identical to pre-#1013.** No `CLAUDE_CODE_SESSION_ID` → 0 files; `APEXYARD_DISABLE_RESOLUTION_CACHE=1` → 0 files; unwritable `chmod 555` dir → correct value, no error, no crash; cold == cache-disabled byte-for-byte. ✓ (verified empirically, not just via the suite.)

**5. No gate weakened.** Every gate resolves the same value warm or cold; the cache is transparent and every cached value is config/filesystem-derived, never dynamic forge state.

### Findings (both LOW / informational — non-blocking)

**[LOW] `(mtime,size)` fingerprint is defeatable by a same-size + same-whole-second edit.** Reproduced (§1). Not gate-weakening (no privilege gain; fail-safe direction; cannot inject never-written values). AgDR-0120's "correct-or-cold, never confidently-wrong" slightly overstates — it *can* be confidently-stale under this specific edit shape, just never into a new gate-weakening state. Recommend documenting this collision residual alongside the symlink one, or upgrading the fingerprint to a content hash if the perf budget allows (a per-read `shasum` would partly erode the win, so documenting is the pragmatic call).

**[LOW] The ops-root component of the wired fingerprint is effectively dead.** The wired `_resolution_cache_current_fingerprint` reads `root="${_CONFIG_ROOT_CACHE:-}"`, but `_CONFIG_ROOT_CACHE` is only ever set inside `$(...)` subshells (`_config_defaults_file`/`_config_overrides_file`), so in the caller process it is **always empty** — I confirmed the live fingerprint is `|<defaults_sig>|<overrides_sig>` with an empty root field. Config-content binding is fully preserved by the two config files' `(mtime,size)` (the security-relevant part), so this is **not independently exploitable**. But (a) it contradicts the "exactly three inputs" claim in the AgDR/header comments, and (b) test case 6 gives false confidence — it exercises `_resolution_cache_config_fingerprint` with explicit root args rather than the wired path. Recommend either populating `_CONFIG_ROOT_CACHE` in the caller so root genuinely binds, or amending the docs to state the file-sigs carry the binding and root is best-effort/often-empty.

### Verdict

**APPROVED (SECURE).** No cache-poisoning, stale-serve, or file-poisoning path reaches a gate with a weakened value under a realistic threat model. Every gate-reaching path requires same-user write to `$HOME` or the fork — capabilities that already defeat the framework outright and predate this PR — and the cache reuses the identical #381 pin trust boundary the merge gate already depends on. The two findings above are documentation/robustness, non-blocking. The one failing suite test (`test_lib_self_location_cwd_anchor.sh`) is confirmed pre-existing (fails identically on base `dev`), unrelated to this change.

---
📌 Reviewed commit: `d346c1c6b9a122bd0c1fef5dcac9bfc58635bd3d`

me2resh (commented):
## Code Review: PR #1127

**Commit**: `d346c1c6b9a122bd0c1fef5dcac9bfc58635bd3d`
**Scope**: `Full` (Heavy — trust-chain; the cached libs feed every merge/migration/design gate)

### Summary

Adds a session-scoped, cross-process file cache (`_lib-resolution-cache.sh`) for the merged `project-config` JSON and the 8 `portfolio_*` resolved paths, consulted on the cold-start path of the existing per-process caches in `_lib-read-config.sh` and `_lib-portfolio-paths.sh`. It reuses the #381 ops-root pin (no second ops-root resolver), deliberately does not cache per-filter `config_get` results or dynamic forge state, and is gated behind a fingerprint of `(ops root, both config files' mtime+size)` recomputed on every read.

The engineering is careful and the fail-safe posture is genuinely good — no-session-ID, `APEXYARD_DISABLE_RESOLUTION_CACHE=1`, and an unwritable cache dir all degrade cleanly; `UNKNOWN` stat is refused on both read and write; the cold path is byte-identical at the consumer boundary. **But** the empirical trace turned up two places where the code does not match the invariant the AgDR states, and on a trust-chain library that feeds the gates the record has to match reality. Requesting changes for those two, both cheap to resolve.

### Checklist Results
- ✅ Architecture & Design:      Pass — clean layering, reuses the pin, no second resolver
- ✅ Code Quality:              Pass — shellcheck clean on all 5 files; atomic tmp+mv writes; well-commented
- ⚠️ Testing:                   Pass with a gap — 9/9 new + full-suite flake confirmed pre-existing, but Case 6 asserts an inert property (F2) and no case covers the same-second/same-size edit (F1)
- ✅ Security:                  Pass — no secrets; cache keys are hardcoded literals, never tool input; per-session scoping
- ✅ Performance:               Pass (mechanism works; headline AC not clearly met — see note, non-blocking)
- ✅ PR Description & Glossary: Pass — narrative Summary with an honest before/after, `Refs #1013`, 5-term Glossary
- ✅ Technical Decisions (AgDR): Pass on presence (AgDR-0120, correct next number) — but its stated fingerprint model is inaccurate (F2) and its "provably fresh" claim is falsifiable (F1)
- ✅ Adopter Handbooks:         N/A — no handbooks loaded (shell-only diff)
- ✅ Fallow Static Analysis:    N/A — not a JS/TS diff

### Issues Found

**⛔ F1 — BLOCKING (staleness): a same-second, same-byte-size config edit serves a stale value for the rest of the session.**

The fingerprint is `(root | defaults mtime:size | overrides mtime:size)`. Both `stat -c %Y` (GNU) and `stat -f %m` (BSD/macOS) report **integer-second** mtime. So if `.claude/project-config.json` is written twice within one wall-clock second, with identical byte size, and a cache-populating read lands between the two writes, the fingerprint is unchanged and every later read in the session serves the pre-edit value. It does **not** self-heal — the file's mtime is fixed at that second and its size is unchanged, so the stale hit persists for the whole session.

Reproduced deterministically (forcing the mtime collision with `touch -t` to isolate the mechanism from timing luck):

```
read1 (file = {"tracker":{"kind":"gh"}})  -> gh    size=25
# overwrite with {"tracker":{"kind":"gl"}}, same mtime, same 25 bytes
read2 (file = {"tracker":{"kind":"gl"}})  -> gh    <-- STALE (should be gl)
# control: bump mtime to a new second -> read returns the fresh value correctly
```

This directly contradicts the AgDR / header claim that "a cache hit is provably as fresh as the files on disk right now" and "correct-or-cold, never confidently-wrong." I want to be fair about scale: the trigger is **narrow and non-adversarial** — it needs two same-second, same-size writes straddling a read, on the operator's own local fork, with the flipped bytes being gate-relevant (an equal-length value swap like `gh`↔`gl`, an enum toggle, a same-length path rename). Most real edits change size and are caught, and I could not construct a natural (non-`touch -t`) trigger. But this is a trust-chain library the merge/migration/design gates read, the invariant is stated as absolute, and the value stays stale for the session — so the record can't claim it can't happen.

Two acceptable ways to clear this (either is fine, both are small):
- **Close the window**: refuse to *write* a cache entry when a config file's mtime equals the current epoch second (the standard "don't trust a same-second stat" guard) — the next second's read then recomputes; or fold a cheap content signature (`cksum`/`shasum` of the two files) into the fingerprint.
- **Or document it honestly**: add it as a second explicit accepted residual alongside the symlink one, and soften "provably as fresh as the files on disk right now" to match. Given the symlink residual is already documented at this exact bar, parity here is defensible.

**⛔ F2 — BLOCKING (record accuracy + a masking test): the fingerprint's `root` component is always empty; the AgDR says it is the resolved ops root, and Case 6 masks the gap.**

AgDR-0120 lists fingerprint input #1 as "the resolved ops root," and `_resolution_cache_current_fingerprint` reads `root="${_CONFIG_ROOT_CACHE:-}"`. In the real wiring `_CONFIG_ROOT_CACHE` is only ever populated inside nested command-substitution subshells (`defaults=$(_config_defaults_file)` etc.), so it never reaches the frame that computes the fingerprint. Verified empirically — the live fingerprint is `|1785601601:17728|ABSENT` (leading `|` = empty root), not `<opsroot>|...`.

In practice this is not an exploitable hole *today*: within one session the two config-file **paths** are resolved consistently (pin-first, converging to the same ops root), and content changes are caught by their mtime+size — so the empty root component is inert defensive code rather than a live bug. The problem is the combination: (a) the AgDR documents a fingerprint input the code doesn't capture, and (b) **test Case 6 ("an ops-root mismatch is never trusted") calls `_resolution_cache_config_fingerprint` directly with hardcoded `/fork/a` vs `/fork/b`**, so it proves the *function* distinguishes roots while the *wiring* never passes a non-empty root — false green. On a trust-chain change that's the kind of gap that hides a future regression. Please either wire the ops root in for real (and keep Case 6 meaningful), or drop `root` from the fingerprint + AgDR and replace Case 6 with a test that exercises the actual `_resolution_cache_current_fingerprint` path.

### Verified good (no action)

- **Cold-path byte-identity**: a cache MISS runs exactly the pre-#1013 logic. Config: the `$(jq…)`+`printf '%s\n'` restructure only normalizes trailing newlines, and the sole consumer (`config_get`) wraps `_config_load` in `$(...)` which strips them — confirmed no other caller of `_config_load` exists. Portfolio: `_portfolio_resolve_with_session_cache`'s miss path is the identical `_portfolio_get`→`_portfolio_resolve` chain the 8 resolvers used inline. Case 1 proves cold == disabled byte-for-byte.
- **UNKNOWN refusal on both paths**: read (`_resolution_cache_read[_json]`) and write (`_resolution_cache_write[_json]`) each guard `[ "$fp" != "UNKNOWN" ]`, plus the call sites guard again. Case 5 proves it.
- **Fail-safe**: no session-ID (Case 9) and disable flag (Case 8) write zero files; unwritable dir (Case 7, chmod 555) returns the correct value with clean stdout and rc 0. Value is always computed before any write attempt.
- **No gate behaviour change warm vs cold**: gates read the same resolved config/paths either way (Case 1 + Case 7); nothing dynamic (CI/PR-review/merge state) is cached.
- **Symlink residual (the one documented risk)**: genuinely narrow — needs a real dir→symlink topology swap on a resolution-chain segment mid-session with config unchanged, on the local fork. Honestly documented; I agree it's acceptable at that bar.
- **Tests**: `test_resolution_cache.sh` 9/9 here. The single full-suite failure (`test_lib_self_location_cwd_anchor.sh`) reproduces **identically** on clean `origin/dev` (a345010) and on this PR head — same 5 sub-cases — so it is pre-existing BASE_REF-staleness flake, not a regression from #1127.
- **Hygiene**: AgDR-0120 is the next free number (0118/0119 present); `Refs #1013` (correctly non-auto-closing under the release-cut model); Glossary present; shellcheck clean.

### Note on the headline AC (reporting only — NOT a factor in this verdict)

The issue's headline AC was a measured **material** reduction on the ~48-hook pass. The author reports this honestly: ~27–35% on the isolated resolution chain, but no clearly measurable win on the full 49-hook pass, because only ~13/28 hook scripts touch the cached libs and most call once. My read for the human's merge decision: the change is **still worth landing on correctness/debt-reduction grounds** — it collapses the redundant jq-merge + multi-subshell canonicalization on the resolution chains, and consolidates the "two resolvers for the same fact" pattern (`_portfolio_root` → `resolve_ops_root`), with the real win concentrated in the merge-gate/migration/design-review hook families that make several resolver calls. The perf case is targeted, not headline. This does **not** affect my verdict — the request-changes below is driven solely by F1/F2 (correctness + trust-chain record accuracy).

### Verdict

**CHANGES REQUESTED** — for F1 (a reproducible same-second/same-size stale-serve that contradicts the stated invariant) and F2 (the AgDR + Case 6 describe an ops-root fingerprint the wiring doesn't actually capture). Both are cheap: close-or-honestly-document F1, and wire-or-remove-and-retest F2. The mechanism itself is well-built and, once the record matches reality, this is a good change to land. No `-rex.approved` marker written.

---
📌 Reviewed commit: `d346c1c6b9a122bd0c1fef5dcac9bfc58635bd3d`

me2resh (commented):
## Security Re-Review (DELTA): PR #1127

**Commit**: `d5c443f8930b5a95c0566e32ed99e02abe704ac6` (prior SECURE marker at `d346c1c` now invalidated)

🛡️ Focused delta re-review of `d346c1c → d5c443f` — the follow-up that closes Rex's two findings (F1 mtime-second stale-serve, F2 dead `root` fingerprint input). Reviewed against the SECURE properties established at the prior HEAD.

### Delta surface
Only 4 files changed. Non-test/non-doc code is confined to **one file** — `_lib-resolution-cache.sh`:
- `_lib-portfolio-paths.sh`: **comment-only** (verified — zero code lines changed).
- `_lib-read-config.sh`: **not in the delta at all**.
- test + AgDR: test-only / docs-only.

### F1 — the write-time settledness guard: no fail-open

`_resolution_cache_write_time_settled` is wired as `... || return 1` in **both** `_resolution_cache_write` and `_resolution_cache_write_json`, positioned *after* the `enabled` + `UNKNOWN` checks and *before* any filesystem work. It is **purely additive** — it can only ever ADD a refuse-to-write case. Reads are byte-identical to the prior HEAD (confirmed by diff). The guard never touches the value returned to the caller (the caller has already computed the correct value); it only decides whether the *next* process gets to skip the work. A refused write ⇒ cold re-resolve ⇒ correct value.

Every error branch fails toward not-caching, never toward serving stale:
- `date +%s` fails → `return 1` → **refuse write** (fail-CLOSED). ✓
- `mtime == now` → `return 1` → **refuse write** (the core guard). ✓
- `command -v _config_defaults_file/_config_overrides_file` undefined → `return 0`. This is **test-harness-only**: in production those functions are always sourced, and a non-`UNKNOWN` fingerprint (required to even reach the guard) provably implies they're defined — the fingerprint is computed *through* those same functions, so if undefined the fingerprint is `UNKNOWN` and the write is already refused one line earlier. Not reachable as a fail-open in production.
- per-file `_resolution_cache_file_mtime_epoch` returns 1 (stat fails / file absent) → `continue`. The only non-refusing branch, but it cannot produce a stale-serve: a file whose mtime is unreadable also has an `UNKNOWN`/`ABSENT` fingerprint sig (write already refused upstream, or invalidated on next read). The value written is still the freshly-computed correct one; its fingerprint governs all future reads. No wrong value is ever served.

### F1 — mtime-swap attack actually mitigated (demonstrated, not asserted)

I re-ran the exact same-second/same-size attack I demonstrated at the prior HEAD, via test case 10 through the real wired path (`portfolio_workspace_dir`, frozen `date` shim for determinism):
- **Guard neutralized** (forced always-settled): case 10 **FAILS** — serves stale `ws-AAAAA` instead of `ws-BBBBB`, reproducing my original stale-serve.
- **Guard in place**: case 10 **PASSES**.

Mechanism is sound: the guard guarantees `cached_mtime < write_wall_time`, so any post-write edit sets `new_mtime ≥ write_wall_time > cached_mtime` ⇒ `new_mtime ≠ cached_mtime` ⇒ the unchanged read-side fingerprint always mismatches ⇒ cold re-resolve. The window is closed at write time. The one residual (a backwards clock step landing precisely on the cached second + a same-byte-size edit) is inherent to any mtime cache, well outside the threat model, and **not overclaimed** — the AgDR scopes the "never confidently-wrong" claim to the same-second/same-size window it actually closes.

### F2 — dropping the `root` fingerprint input weakens nothing

Confirmed the `root` field was a permanently-dead leading component (always the empty string in production — lost to a command-substitution subshell), so removing it drops **zero** discriminating power. The ops root remains covered by (a) `resolve_ops_root()` (#381) pinning it once per `$CLAUDE_CODE_SESSION_ID`, and (b) `_resolution_cache_file` scoping every cache filename by that same session id — so one session's cache can never serve a different root's answer. Verified no live caller still passes the old 3-arg form; `_CONFIG_ROOT_CACHE` survives only in explanatory comments.

### Prior SECURE properties re-confirmed at the new HEAD
- `UNKNOWN`-stat refused on **both** read and write paths — unchanged.
- Cache file **location** (`$HOME/.claude/apexyard`), **filename scoping** (session-id), and `_resolution_cache_enabled` gate — **byte-identical** to prior HEAD (diffed).
- Write remains atomic (temp-then-`mv -f`, default umask) — body unchanged; only the guard line added.
- Malformed / mismatched / missing cache → exact-fingerprint-match required → cold path. Unchanged.
- no-session / disable-flag / unwritable-dir degrade byte-identically (test cases 7/8/9 green).
- Nothing dynamic cached; no new cached value introduced.
- Whole suite runs with `APEXYARD_DISABLE_RESOLUTION_CACHE=1`, so no gate reads differently warm vs cold — and no other test can even reach the changed code.

### Tests (isolated worktree at HEAD, tree left on `dev`)
- `test_resolution_cache.sh`: **10/10 passing** (incl. case 6 F2-rewrite, case 10 F1-guard).
- Case 10 confirmed a **genuine regression test** (fails when the guard is neutralized).
- Both named pre-existing failures reproduce on `dev` (a345010, the PR's merge-base) with **none** of the PR's changes present → confirmed pre-existing and unrelated: `test_lib_self_location_cwd_anchor.sh` (documented zsh/bash BASE_REF timing flake) and `test_workspace_tracker_resolution.sh` (pre-existing workspace-clone tracker-resolution gap).
- `shellcheck --severity=warning` clean on the changed lib.

### Checklist Results
- XSS Prevention:         N/A
- Auth & Authorisation:   Pass (gate trust-chain integrity preserved — no gate reads differently warm vs cold; no fail-open)
- API Security:           N/A
- Gate & Trust-Chain Integrity: Pass — the guard only ever refuses to cache; no gate/marker check is weakened, bypassed, or made to fail-open.

### Security Issues Found
None. The delta is strictly-safer-by-intent and strictly-safer-in-fact: it adds one fail-closed refuse-to-write case, leaves the read path and cache location/permissions untouched, and closes the mtime-swap stale-serve I previously demonstrated without opening any new surface.

### Verdict
**APPROVED (SECURE)**

---
📌 Reviewed commit: `d5c443f8930b5a95c0566e32ed99e02abe704ac6`

me2resh (commented):
## Code Review (re-review): PR #1127

**Commit**: `d5c443f8930b5a95c0566e32ed99e02abe704ac6`
**Scope**: Full — Heavy trust-chain (session-scoped resolution cache feeding the merge/migration/design gates)

### Summary

Re-review of the `fix(#1013)` follow-up commit (`d5c443f`) addressing the two CHANGES-REQUESTED findings raised against the original implementation commit (`d346c1c`). Both findings are **genuinely closed**, verified independently in isolated worktrees (no `gh pr checkout`), and I found no regressions. **Verdict: APPROVED.**

### Checklist Results
- ✅ Architecture & Design:      Pass
- ✅ Code Quality:                Pass
- ✅ Testing:                     Pass (new/rewritten cases are genuine discriminators, empirically fail-on-base / pass-on-fix)
- ✅ Security / trust-chain:      Pass (correct-or-cold posture preserved; write-guard fails closed)
- ✅ Performance:                 Pass (mechanism unchanged; guard is 1 extra `date`+`stat` on the write path only)
- ✅ PR Description & Glossary:   Pass
- ✅ Technical Decisions (AgDR):  Pass — AgDR-0120 amended in place + Update section
- N/A Adopter Handbooks:          none loaded (no handbook triggers on this diff)
- N/A Fallow Static Analysis:    not JS/TS

### F1 — mtime-second stale-serve (blocker) → CLOSED

The write-time guard is the correct fix. `_resolution_cache_write` / `_resolution_cache_write_json` now refuse to persist while any fingerprint input file's mtime equals the current wall-clock second, so the cache is only ever populated once every input file has settled into a strictly-past second. The invariant that closes the bug: **a cache entry's fingerprint mtime is always strictly < the second it was written in**, so any subsequent edit necessarily lands on a later second and the unchanged read-side (mtime,size) comparison catches it.

Reasoned through the holes the brief named:
- **Sub-second edits after the second rolls** — no hole. Once cached (file mtime = S, written at now ≥ S+1), any later edit sets mtime ≥ now > S → read-side miss. Multiple edits within a later second all share that later integer second, still ≠ S. Caught.
- **Clock going backwards** — an exotic residual (a backward NTP jump to exactly the cached second plus a same-size edit could re-collide), but this is the universal floor of any integer-second mtime cache, not something this design uniquely reopens.
- **Fail-closed on undeterminable "now"** — correct: `date +%s` failure → treated as unsettled → write refused, consistent with the existing UNKNOWN-fingerprint rail.

Empirical proof it's a real, non-tautological discriminator: I ran the **new** `test_resolution_cache.sh` (case 10) against the **pre-fix libs** (`d346c1c`) and it FAILS — `second='…/ws-AAAAA'` (stale, served from cache for the whole session) — and PASSES on the fix (`second/third/fourth = …/ws-BBBBB`). Case 10 drives the real wired path (`portfolio_workspace_dir`) with a frozen `date` shim, so it's deterministic rather than racing a real second boundary. The phase-3/4 "bonus" (advance frozen now past the file mtime → caching resumes) correctly proves the guard *delays* rather than *disables*.

*(nit, non-blocking, honesty of the record)* The AgDR states "no cache entry is EVER written whose freshness a same-second, same-size edit could silently defeat." Strictly there remains a vanishing TOCTOU: if the merged value is read in second S, an external same-size edit lands in S *after* that read, and the cache write happens to straddle into S+1, the guard sees a settled mtime and caches the pre-edit value (would then read stale until the next mtime-moving edit). This requires a sub-microsecond boundary-straddle coincident with an external same-size edit — not practically reproducible, and only closable by content-hashing (rightly considered and rejected on cost). It does **not** reopen F1 (the reproducible, frozen-clock, session-long stale-serve is fully closed). Suggest softening the absolute "EVER" to sit alongside the already-documented symlink residual, so the record stays as scrupulously honest as the rest of the AgDR.

### F2 — dead ops-root fingerprint input → DROPPED, correctly

Confirmed the `root` param is truly dead in every live caller:
- `_resolution_cache_config_fingerprint` is now 2-arg; the **only** live call site is `_resolution_cache_current_fingerprint` (2-arg). Both real consumers — `_config_load` (read-config) and `_portfolio_resolve_with_session_cache` (portfolio-paths) — go through that wrapper. No live 3-arg call remains anywhere in `.claude/`.
- The "always empty in production" claim is structurally sound: `_resolution_cache_current_fingerprint` runs inside `$(…)`, and the old `root="${_CONFIG_ROOT_CACHE:-}"` read it in a frame where `_CONFIG_ROOT_CACHE` was only ever populated inside sibling command-substitution subshells (`_config_defaults_file`/`_config_overrides_file` → `_config_repo_root`) → discarded. It was always `""`.
- **Empirically proven**, not just argued: running the rewritten case 6 against the pre-fix libs FAILS with the persisted fingerprint `|1577829600:17728|ABSENT` — the leading `|` is the dead empty root field. On the fix it's a clean two-field `<defaults-sig>|ABSENT`. The old case 6 gave a false green by hand-feeding `/fork/a`/`/fork/b` args no real caller produced; the rewrite drives the real `portfolio_registry → _resolution_cache_write` chain and asserts on the on-disk fingerprint.
- No protection lost: the ops root can't change within a session (`resolve_ops_root` pins it per `$CLAUDE_CODE_SESSION_ID`, #381) and every cache file is already scoped by that session id in its filename, so there was never a within-session ops-root threat for the field to guard.

### No regression

- **Cold-path byte-identity**: case 1 (cold enabled == cache-disabled, byte-for-byte) PASSES on fix.
- **Warm reuse intact**: cases 2 (no re-run of the `jq -s` merge) and 3 (no re-run of the portfolio compute path) PASS.
- **Fixture backdating is sound, not a masked assertion**: `make_fork`/case_2 backdate the defaults/overrides mtimes to 2020 so the *write* actually succeeds (the guard would otherwise refuse a same-second fresh-fixture write, and the cache-reuse cases genuinely need the write to land). This reflects steady-state (config files aren't edited in the same second a hook reads them) and does **not** mask F1 — case 10 exercises the collision separately with a frozen clock. Removing the backdating would make the reuse cases fail, not falsely pass.
- **Fail-safe degrades**: unwritable dir (case 7), disable flag (case 8), no session id (case 9) all PASS with byte-identical output and no crash; the guard adds one more fail-closed branch to that same list.
- **Whole-suite blast radius**: `bin/run-hook-tests.sh` exports `APEXYARD_DISABLE_RESOLUTION_CACHE=1` suite-wide, so for every test *except* `test_resolution_cache.sh` the cache read/write functions early-return and the F1/F2 changes are inert — no other test's behaviour can shift. Spot-ran `test_portfolio_paths.sh`, `test_per_project_tracker.sh`, `test_tracker_aware_hooks.sh` on the fix HEAD: all PASS.

### Pre-existing failures — confirmed NOT regressions

Ran both on the clean base HEAD (`a345010`, = dev tip = merge base) and on the fix:
- `test_lib_self_location_cwd_anchor.sh` — fails identically on base and fix (same `read-config FIX (bash): genuine anchored fork still resolves` line; environment/BASE_REF artifact, #1102 area).
- `test_workspace_tracker_resolution.sh` — fails identically on both (Passed 2 / Failed 5, identical case set {1,2,4,6,7}; workspace-clone tracker-resolution gap, unrelated to the cache).

Neither is introduced or worsened by this PR.

### Other checks
- `shellcheck --severity=warning` on all five touched/added files: **clean**.
- AgDR-0120 now matches the shipped code — two-input fingerprint, write-guard described as the truth (not aspiration), the dead ops-root claims removed from the Decision/Staleness sections, the symlint residual left intact, and an "## Update" section recording both findings. (One advisory nit above on the absolute "EVER" phrasing.)

### Verdict

**APPROVED** — both prior findings genuinely resolved, empirically verified fail-on-base/pass-on-fix, no regressions, trust-chain posture preserved (correct-or-cold; write-guard fails closed). The single nit (F1 write-vs-read straddle vs the AgDR's absolute "EVER") is advisory only and does not gate the merge.

---
📌 Reviewed commit: `d5c443f8930b5a95c0566e32ed99e02abe704ac6`

atlas-apex (approved):

Issue #1123, opened 2026-08-01
[Feature] Registry models a project as exactly one repo — multi-repo systems lose leak-scrub coverage


me2resh:
## Problem

The registry models a project as **exactly one repo**. `apexyard.projects.yaml` entries carry a singular `repo: owner/name`, and every consumer assumes it.

That breaks for a common real shape: **one product split across several repos** — a backend, a gateway, and two workers, say. They deploy together, share a tracker in practice, and are governed as one thing. Today `/handover` has only two options, and both are wrong:

- **Register N separate projects** — N assessments that repeat the same findings, N registry rows, N role sets, and a `/projects` view that implies four independent products where there is one.
- **Register one project naming the "primary" repo** — the docs are unified, but the other repos become invisible to every registry-driven mechanism.

## Why the second option is the dangerous one

It isn't just cosmetic. `block-private-refs-in-public-repos.sh` builds its scrub list by walking each `- name:` block and reading `name`, `repo`, and `workspace`. **One block scrubs one repo slug.**

So a project registered with one of its four repos leaves the other three slugs **unscrubbed** — and the whole point of that hook is that an agent writing an upstream framework issue has the private names right in front of it and must actively suppress them. Naming a system's repos in a bug report is exactly the natural thing to do.

Same singular assumption in `_lib-tracker.sh`, `_lib-project-board.sh`, and `_lib-multi-repo-trace.sh`.

## What I'd like

A first-class multi-repo project. Rough shape, not a demand:

```yaml
- name: some-platform
  repos:                          # plural; singular `repo:` stays valid
    - owner/backend
    - owner/gateway
    - owner/worker-a
    - owner/worker-b
  primary: owner/backend          # for tracker/board defaults
  workspace: workspace/some-platform
  docs: projects/some-platform
```

The load-bearing part is **not** the syntax — it's that every registry consumer reads a *list*:

- **Leak protection scrubs every slug in the project**, not just the first. This is the one I would prioritise; the others are ergonomics, this one is a disclosure boundary.
- Tracker resolution picks `primary` by default, accepts an explicit `--repo` for the others.
- `/projects`, `/inbox`, `/tasks`, `/status` aggregate across the set and render one row.
- `/handover` produces **one** assessment with per-repo sections, and one harnessability score per repo under a single verdict.

Backwards compatible: `repo:` singular keeps working and is equivalent to a one-element `repos:`.

## How this came up

[...]

PR #1129, opened 2026-08-01
feat(#1123): first-class multi-repo project (leak-scrub covers every repo slug)

me2resh:
## Summary

- **The security fix this PR exists for: leak protection now scrubs EVERY repo of a multi-repo project, not just the first one.** `block-private-refs-in-public-repos.sh` used to build its scrub list by reading one `name:`/`repo:`/`workspace:` triple per registry entry — so a product registered under one of its several repos left the *other* repos' slugs completely unscrubbed on public-repo writes. An agent naming a sibling repo in an upstream bug report ("also reproduces in the gateway repo") would have leaked it. The hook's registry parser now walks a project's whole `repos:` list (block or flow style), and a new test proves a non-primary repo slug — and a third repo, to prove it's not just "first two" — gets the same block a primary repo's slug already got.
- **First-class plural `repos:` schema, singular `repo:` untouched.** `apexyard.projects.yaml` now accepts `repos: [owner/a, owner/b, ...]` + an optional `primary:` for a project split across several repos that deploy together and are governed as one thing. `repo: X` stays fully valid and is defined to behave byte-identically to `repos: [X]` with no `primary:` — verified directly by a test that runs the same leak-scrub and tracker-resolution assertions against both schemas and checks they match.
- **Tracker and cross-repo-trace resolution read the list, not a scalar.** `_lib-tracker.sh`'s per-project override selector (`_tracker_project_value`) now matches `.repo == <target>` OR membership in `.repos[]`, so a caller passing ANY of a multi-repo project's repos finds the same tracker override — not just its primary. `_lib-multi-repo-trace.sh`'s registry parser (used by `/dfd` and `/process`) gained an `<all_repos>` field; its tier-4 URL/slug matcher now checks every declared repo, and two new helpers (`mrt_primary_repo_for`, `mrt_repos_for`) let a caller resolve "the default repo" or "every repo" for a project by name.
- **`_lib-project-board.sh` needed no change** — investigated per the issue's own suspicion, but GitHub Projects v2 boards are already an org-level construct addressed by `owner`/`board_number` + an issue/PR number, never a per-project `repo:` lookup. Recorded as an explicit finding in AgDR-0121 rather than silently skipped.
- **Deferred, on purpose:** the portfolio-view skills (`/projects`, `/inbox`, `/tasks`, `/status`, `/handover`) still render one row per registry entry, not one row aggregated across a multi-repo project's repos. The issue explicitly deprioritises this ("the others are ergonomics, this one [leak-scrub] is a disclosure boundary") — doing it properly means touching five markdown-defined skill flows, a materially larger and separable piece of work from the schema + security-consumer fix here.

## Testing

1. `.claude/hooks/tests/test_multi_repo_registry.sh` (new, 17 cases) — the security AC (non-primary + third repo slug scrubbed), backwards-compat (singular `repo:` unaffected), one-element-`repos:`-equals-singular-`repo:` equivalence, tracker resolution matching a non-primary repo, and `_lib-multi-repo-trace.sh` tier-4 + the two new resolver helpers.
2. Existing `test_block_private_refs.sh` (74 cases) and `test_per_project_tracker.sh` (6 cases) pass **unmodified** — zero behaviour change for a single-repo fork.
3. Full `bin/run-hook-tests.sh` suite run; a pre-existing, unrelated flake in `test_lib_self_location_cwd_anchor.sh` was independently confirmed to reproduce identically on a clean `dev` checkout (stash/pop verification) before and after this change — not caused by this PR.
4. `shellcheck --severity=warning` clean on every changed/added shell file.

Refs #1123

---

## Glossary

| Term | Definition |
|------|------------|
| registry | `apexyard.projects.yaml` — the list of every repo under ApexYard management. The single source the portfolio skills read. |
| scrub list | The private names `block-private-refs-in-public-repos.sh` removes from anything written to a public framework repo. Built from the registry, so a repo missing from it is a repo that never gets scrubbed. |
| primary repo | In a multi-repo project, the repo tracker/board operations default to when none is named explicitly — the explicit `primary:` field, or the first entry of `repos:` if omitted. |
| `repos:` | The new plural registry field — a list of every repo a single project entry governs, for a product split across several repos deployed together. |


me2resh (commented):
## Security Review: PR #1129

**Commit reviewed**: `d11761ec335a16f9a965df13081fb6d8d08490d5`
🛡️ Reviewed by Hakim (Security Auditor) — empirical sandbox, hook run directly against crafted registries (not reasoned from prose).

### Verdict
**CHANGES REQUESTED** — the disclosure boundary this PR exists to close is **not fully closed**. A valid, commonly-authored YAML shape leaves *every* repo of a multi-repo project unscrubbed. Security marker **not** written.

---

### What I verified works (the core fix is real)

I built a fork sandbox, registered a 4-repo project (`repos: [acme/svc-backend, acme/svc-gateway, acme/svc-worker-a, acme/svc-worker-b]`, `primary: acme/svc-backend`), and ran `block-private-refs-in-public-repos.sh` directly against `gh issue create --repo me2resh/apexyard` writes. Using a project name (`orbit`) that is deliberately **not** a substring of the slugs, so a block can only come from the repo-slug scrub itself, not an incidental name match:

| Probe | Result |
|-------|--------|
| primary slug `acme/svc-backend` | **blocked (exit 2)** ✅ |
| non-primary `acme/svc-gateway` | **blocked** ✅ |
| non-primary `acme/svc-worker-a` (3rd repo) | **blocked** ✅ |
| non-primary `acme/svc-worker-b` (4th repo) | **blocked** ✅ |
| project name `orbit` | blocked ✅ |
| workspace `workspace/orbit` | blocked ✅ |
| clean body naming none of them | passes (exit 0) — no false positive ✅ |

So for the **bare block-list** and **flow-list** forms, every slug scrubs — the headline claim holds. Also confirmed:
- **Backwards-compat**: singular `repo: acme/solo-svc` (with and without a trailing comment) scrubs exactly as before.
- **Fail-safe / never an empty scrub list**: an empty `repos:` (header, no items) does **not** empty the whole list — a sibling project's `repo:` and the project's own `name` still scrub. The leak hook's step-7 parser is **awk-only** (no yq/python multi-path), so there is no tool-absence fail-open here — good, and materially safer than the `#1121/#1122` resolver class.
- **Parser conflation**: `repos:` interleaved with `tags:`/`roles:` (block + flow, `tags:` before or after `repos:`, a tag literally named `acme/not-a-repo`, mid-list comments, quoted items, trailing-whitespace header) — every real repo is still emitted, and no tag/role token is over-captured as a repo. The `current_list` state machine holds for all of these.
- **`_lib-tracker.sh`**: injection-safe. `strenv(REPO)` avoids interpolating the slug; `$key` is a hardcoded literal at every one of the 7 call sites (`kind`/`id_pattern`/`view_command`/`create_command`/`list_command`/`review_command`/`merge_command`); the python3 path is argv-safe; `contains([...])` membership is correct.
- New `test_multi_repo_registry.sh` passes 17/17; existing `test_block_private_refs.sh` (74) and `test_per_project_tracker.sh` (6) pass unmodified.

---

### 🔴 HIGH — commented `repos:` block header silently drops the ENTIRE list from the scrub

**This is the exact leak the PR must close, and it is reachable on a valid, plausible registry.** When the `repos:` **block-list header** carries a trailing comment, the whole list is invisible to the scrub and every one of the project's repo slugs reaches a public-repo write **unscrubbed**.

Registry (note only the header comment — everything else is a normal block list):
```yaml
- name: orbit
  repos:  # backend + gateway + workers
    - acme/svc-backend
    - acme/svc-gateway
  workspace: workspace/orbit
```
Result — `gh issue create --repo me2resh/apexyard --body 'repro in acme/svc-gateway'`:
```
exit=0   ← ALLOWED. acme/svc-gateway (and acme/svc-backend) leak onto the public tracker.
```
The raw step-7 scrub list for this project contains **only** `NAME` and `WORKSPACE` — **zero `REPO=` lines**.

**Root cause** (`block-private-refs-in-public-repos.sh`, step 7 awk). The arm rule requires the header to end right after optional whitespace:
```awk
/^[[:space:]]*repos:[[:space:]]*$/ { current_list = "repos"; next }
```
A trailing comment makes that fail to match, so `repos:  # …` falls through to the generic disarm rule `/^[[:space:]]*[a-zA-Z_][a-zA-Z0-9_]*:/`, which sets `current_list=""`. The subsequent `- acme/…` items are then never captured.

**Why it matters / severity.** This is the security control the PR adds, defeated by an ordinary input: inline comments on keys are ubiquitous in YAML config, and `apexyard.projects.yaml.example` is itself heavily commented, so an adopter writing `repos:  # backend + workers` is a natural, expected authoring choice — not an adversarial one. The failure is silent (exit 0, write proceeds), universal (awk-only, not environment-dependent), and drops **all** of the project's repos at once. It is narrower than CRITICAL (needs that specific header-comment authoring), hence HIGH rather than CRITICAL — but it is precisely "a repo slug of a multi-repo project reaching a public write unscrubbed," which is a blocker for this PR by its own charter.

Note the boundary (all verified): the **flow form** `repos: [a, b]  # comment` is safe (the `sub(/\].*$/,"")` strips the comment), the **bare block header** (no comment) is safe, and **singular** `repo: x  # comment` is safe (grabs `$2`). Only the **block-list header + trailing comment** leaks.

**Verified fix** (one line — make the arm rule comment-tolerant; the block-header rule already precedes the generic disarm rule, so this alone closes it):
```awk
/^[[:space:]]*repos:[[:space:]]*(#.*)?$/ { current_list = "repos"; next }
```
With this applied I re-ran the commented-header registry: both slugs now **block (exit 2)**, and the plain-header + `tags:` case still scrubs the real repo without over-capturing the tag. Please also apply the same comment-tolerance to the mirror parser in `_lib-multi-repo-trace.sh`'s greppy fallback (same `repos:[[:space:]]*$` block-header rule) for consistency — lower severity there (that path is `/dfd` `/process` resolution, not disclosure, and only reached when yq is absent), but it carries the identical gap. Add a regression case to `test_multi_repo_registry.sh` covering a commented block-list header.

---

### Checklist Results
- Secrets & Credentials: Pass
- Injection Prevention: Pass (`strenv`, literal keys, argv-safe python)
- Data Protection / disclosure boundary: **Fail** — HIGH above
- Gate & Trust-Chain Integrity: Pass (parser change fails safe on empty/malformed; no gate weakened)
- Backwards-compat: Pass

### Required before approval
1. Close the commented-block-header leak in `block-private-refs-in-public-repos.sh` (fix above).
2. Mirror the fix in `_lib-multi-repo-trace.sh` greppy fallback.
3. Add a regression test for a commented `repos:` block header.

Re-request review after the fix — a new commit invalidates this SHA-bound review.

---
📌 Reviewed commit: `d11761ec335a16f9a965df13081fb6d8d08490d5`

me2resh (commented):
## Code Review: PR #1129

**Commit**: `d11761ec335a16f9a965df13081fb6d8d08490d5`
**Scope**: `Full` (Heavy — trust-chain: leak-protection hook + registry schema + tracker/trace libs)

### Summary

Adds a first-class plural `repos:` (+ optional `primary:`) registry schema so a multi-repo product registers as one project, and — the security core — makes `block-private-refs-in-public-repos.sh` scrub **every** repo slug, not just the primary. Backwards-compat with singular `repo:` is preserved and directly asserted. I independently exercised the parsers (isolated worktree at the PR HEAD, `gh pr diff`, direct awk extraction), ran the suites, and reproduced the yq bug-fix.

The design and the vast majority of the implementation are sound and well-tested. **However, I found one reproducible under-scrub in the security core that leaks private repo slugs onto a public tracker on valid YAML — the exact disclosure boundary this PR exists to close. Requesting changes on that single issue.**

### Checklist Results
- Testing:                   Pass (new 17/17; existing 74/74 + 6/6 unmodified) — but a gap is uncovered (see finding)
- Security:                  **Fail** — one proven under-scrub leak path (blocking)
- PR Description & Glossary: Pass (narrative summary leads with the security fix, `Refs #1123`, Glossary present)
- Technical Decisions (AgDR): Pass — AgDR-0121 present, accurate, no number collision (0120 correctly reserved for the in-review #1127)
- Adopter Handbooks:         N/A (none loaded — no `handbooks/` on this fork)

### What I verified independently (all confirmed as claimed)

1. **Scrub covers non-primary AND a third repo.** The block-style parser walks the whole `repos:` list. Primary, non-primary (gateway), and third (worker) slugs all scrub; project `name` and `workspace` still scrub. `current_list` correctly separates a `repos:` block from a sibling `tags:`/`roles:`/`status:` block — I fed it `repos:` adjacent to `tags:`+`roles:` (both orders), a `status:` scalar between `repos:` and a later `tags:`, comments on their own line inside the block, flow style `repos: [a,b,c]`, flow with a trailing comment, `repos:` as the last key before the next `- name:` — no over-capture and no under-capture in any of these. Empty `repos:` / `repos: []` fail safe. A both-`repo:`+`repos:` misconfig scrubs the **union** (safest for a leak control).
2. **Backwards-compat is byte-identical.** Existing `test_block_private_refs.sh` (74/74) and `test_per_project_tracker.sh` (6/6) pass unmodified; the one-element-`repos:` equivalence case passes.
3. **The yq bug-fix is real and correct.** Against the installed mikefarah yq v4.53.3, `index(strenv(REPO))` errors with `lexer: invalid input text` (which the hook's `2>/dev/null` would silently swallow into a global-default regression), while `contains([strenv(REPO)])` returns the override. The python3/PyYAML fallback's membership logic (`repo in (p.get("repos") or [])`) is correct; PyYAML is absent in this environment so it no-ops and safe-degrades to the global config. Note: `_tracker_project_value` has only two resolution paths (yq + python3) — there is no third greppy path there to miss.
4. **`_lib-multi-repo-trace.sh` `<all_repos>` + tier-4 works in both the yq path and the greppy-awk fallback**; the two new helpers resolve for singular and plural. The author's claim that **`_lib-project-board.sh` needs no change is correct** — its `.projects[]` reads GitHub's `gh project list` JSON (`select(.number == board_number)`), resolving via `github_projects.owner`/`board_number` + item number; there is no per-project registry `repo:` lookup. Not a missed site.
5. **The failing `test_lib_self_location_cwd_anchor.sh` is a pre-existing environment (zsh) flake, not a regression** — it fails **identically** on a clean `dev` checkout (only the random tmp-dir name differs), and this PR touches none of the files it depends on. shellcheck `--severity=warning` is clean on all four changed shell files.

### Issues Found

**⛔ BLOCKING — `repos:` block-header with a trailing inline comment silently drops every repo from the scrub list (leak).**

`block-private-refs-in-public-repos.sh` step-7 parser (and the identical pattern in `_lib-multi-repo-trace.sh:_mrt_parse_registry`) only arms `current_list="repos"` on a bare header:

```awk
/^[[:space:]]*repos:[[:space:]]*$/      { current_list = "repos"; next }   # requires nothing after the colon
/^[[:space:]]*[a-zA-Z_][a-zA-Z0-9_]*:/  { current_list = ""; next }        # this matches `repos: # comment`
```

A `repos:` header carrying a trailing inline comment — **valid YAML** — never matches the block-header arm (the `[[:space:]]*$` requires end-of-line), so it falls through to the generic-key rule, which **disarms** `current_list`. The following `- slug` items are then not captured.

Reproduced end-to-end through the actual hook binary, with a project whose name is not a substring of its slugs (the realistic multi-repo shape):

```yaml
- name: billing
  repos: # payments platform services
    - acme/payments-core
    - acme/invoice-svc
    - acme/ledger
```

A public-repo write — `gh issue create --repo me2resh/apexyard --body 'also reproduces in acme/ledger and acme/invoice-svc'` — **exits 0 (passes, the slugs leak)**. Moving the comment to its own line (or removing it) makes the same hook exit 2 and list `project repo: acme/ledger` / `acme/invoice-svc` as scrubbed. So it is specifically the header-comment parsing, not the slugs.

Why this is blocking rather than a nit: it is a silent under-scrub in the **leak direction**, on **valid YAML**, in the one control this PR ships to close exactly this disclosure boundary. The leak hook is awk-only (no real-YAML-parser fallback), so it is always exposed; the trace lib shares the gap but only when yq is absent. It requires an uncommon authoring style (an inline comment on the list header — the shipped Example 5 doesn't use it), but a mechanical leak control must not fail-open on input a real YAML parser accepts, and the security rail is "ambiguity rounds up."

**Suggested fix (small):** make the block-header arm tolerant of a trailing comment, in **both** files:

```awk
/^[[:space:]]*repos:[[:space:]]*(#.*)?$/  { current_list = "repos"; next }
```

(`(#.*)?` is plain ERE with no intervals, so it stays within the existing "no ERE intervals" portability check.) Please add a regression case: a `repos:` header with an inline comment, project name **not** a substring of its slugs, asserting each slug is scrubbed (exit 2, `project repo: <slug>`).

### Suggestions (non-blocking, same root cause — optional to fold into the fix)

- `nit:` The generic-key disarm regex `[a-zA-Z_][a-zA-Z0-9_]*:` does not match a **hyphenated** custom key (e.g. `deploy-target:`). A hyphenated key with a block list appearing after a `repos:` block would leave `current_list` armed and mis-capture those items as repos. This fails in the **safe** direction (over-scrub → a false-positive block, never a leak), so it's not blocking — but it shares the "line-classification rules don't cover all valid YAML shapes" root with the blocking finding, so a robust fix could address both at once.
- `nit:` AgDR-0121 Consequences: "Filed as **a** explicit follow-up" → "an explicit follow-up".

### Verdict
**CHANGES REQUESTED**

One fix to the security-core parser (repeated in the trace-lib fallback) + a regression test. Everything else — schema, backwards-compat, the yq fix, tracker/trace resolution, the project-board no-change finding, and AgDR-0121 — is solid and verified. Happy to re-review promptly once the header-comment scrub gap is closed.

---
📌 Reviewed commit: `d11761ec335a16f9a965df13081fb6d8d08490d5`

me2resh (commented):
## Code Review (re-review): PR #1129

**Commit**: `125ea58f6b7ce029fd970f16d9b25734aafa8ed9`
**Scope**: Full — Heavy tier (trust chain: `.claude/hooks/**`, leak-protection parser + registry)
**Prior verdict**: CHANGES REQUESTED over one blocker (multi-repo leak via commented `repos:` header).

### Verdict
**APPROVED** — the blocker is closed, proven non-tautologically, with no over-arm regression and backward-compat fully intact.

---

### The blocker is fixed (and I verified it genuinely leaked before)

My prior blocker: a `repos:` block-header carrying a trailing inline comment (`repos:  # …`, valid YAML) fell through the awk state-tracker and disarmed `current_list`, so every repo slug of a multi-repo project reached a public write **unscrubbed**. Present at two awk sites.

The fix arms on `/^[[:space:]]*repos:[[:space:]]*(#.*)?$/` (tolerate a trailing comment). Applied **identically at both sites** — `block-private-refs-in-public-repos.sh:922` and `_lib-multi-repo-trace.sh:127`.

Non-tautology proof (the important part). I ran the **new** test file against the **pre-fix** parser at the PR's own parent `d11761e` — which already has the multi-repo `repos:` support but the old bare-header arm `repos:[[:space:]]*$`. Exactly the comment-header cases flip, nothing else:

| Case | pre-fix `d11761e` | fix `125ea58` |
|------|-------------------|---------------|
| leak regression: commented `repos:` header still scrubs its slug | **FAIL (leaks, exit 0)** | PASS (blocks, exit 2) |
| trace-lib parity (second fix site) | **FAIL** | PASS |
| hyphenated-key over-capture hardening | **FAIL** | PASS |
| all other 24 cases | PASS | PASS |

So the regression test is real: it blocks on the fix and genuinely leaked on the pre-fix parser — not a test that would pass either way. Both fix sites were independently buggy pre-fix and are both closed.

### No over-arm regression

The `(#.*)?` tolerance is scoped to a literal `repos:` header. A commented `tags:` header does **not** arm as `repos:` — it falls through to the generic `key:` disarm rule (which correctly runs *after* the `repos:` arm in awk source order, so the arm's `next` wins first). The over-arm safety test ("`tags:` header with a trailing comment is NOT treated as repos:", exit 0) plus its companion (the project's real `repo:` still scrubs, exit 2) both pass — confirming the clean result means "not a repo", not "parsing broke".

### Backward-compat (the operator's explicit requirement)

- `test_block_private_refs.sh` — **74/74 pass, file unmodified** (absent from the PR diff; structurally guaranteed, and re-run to confirm).
- `test_per_project_tracker.sh` — **6/6 pass** (the tracker lib changed, so this matters).
- Singular `repo:` and one-element `repos:` unchanged: dedicated equivalence + all-singular-registry fixtures pass.

### Everything else

- New `test_multi_repo_registry.sh` — **27/27 pass** (matches author's report).
- `shellcheck --severity=warning` — **clean (exit 0)** on both changed parser files.
- The pre-existing `test_lib_self_location_cwd_anchor.sh` flake is not this PR.
- AgDR requirement (trust-chain change, material per the always-material rail) — satisfied: AgDR-0121 is linked and added.
- My prior nit #2 (AgDR-0121 typo "a explicit" → "an explicit") — **fixed**.
- My prior nit #1 (hyphenated-custom-key over-capture) — **also fixed** as a bonus (generic-key disarm widened to `[a-zA-Z0-9_-]`). This one fails safe anyway (over-capture is a false-positive block, not a disclosure), but nice to have it closed.

### Handbook Findings
None. Always-load handbooks (clean-architecture-layers, migration-safety, commit-message-quality) don't apply to a shell-hook parser diff; none are `ENFORCEMENT: blocking`.

### Checklist
- Architecture & Design: Pass
- Code Quality: Pass (minimal, symmetric two-site fix)
- Testing: Pass (regression + over-arm + hardening + equivalence + backward-compat)
- Security: Pass (disclosure boundary closed at both sites, verified non-tautologically)
- PR Description & Glossary: Pass (narrative summary + glossary present)
- Technical Decisions (AgDR): Pass (AgDR-0121 linked)
- Adopter Handbooks: N/A (no applicable/blocking handbook)

---
📌 Reviewed commit: `125ea58f6b7ce029fd970f16d9b25734aafa8ed9`

atlas-apex (approved):

me2resh (commented):
## Security Re-Review: PR #1129 — multi-repo leak-scrub

**Commit**: `125ea58f6b7ce029fd970f16d9b25734aafa8ed9`

Re-review of the fix for my prior **HIGH** blocker: a `repos:` block-header with a trailing inline comment (`repos:  # ...`) disarmed the awk state-tracker, so every non-primary repo slug of a multi-repo project reached a public-class write **unscrubbed**. Verified empirically in a sandbox (fixed hook + base hook extracted via `git show`, hand-built registries, direct hook invocation — no full suite, no checkout).

### Summary
The fix arms the state-tracker on `/^[[:space:]]*repos:[[:space:]]*(#.*)?$/` at both sites (leak hook step-7, `_lib-multi-repo-trace.sh` fallback), tolerating a trailing comment on the `repos:` header. The disarm rule additionally gained `-` to its key char-class. The leak is closed with no new evasion introduced.

### Checklist Results
- XSS Prevention:         N/A (no user-facing render)
- Auth & Authorisation:   N/A
- Data Protection:        **Pass** — disclosure boundary is the whole point; every repo slug now scrubs
- Gate & Trust-Chain Integrity: **Pass** — no gate weakened; no new fail-open path

### Empirical findings

**1. The leak is CLOSED (definitive).** Prior attack reproduced on a multi-repo project with a commented `repos:` header, project name (`zephyr`) deliberately *not* a substring of any slug so a block can only come from the slug scrub:
- Base `d11761e`: body naming non-primary slug `octoworld/beta-worker` → **EXIT 0 (LEAK)**. Standalone base parser confirms it extracts **no repo** for the commented-header project.
- Fix `125ea58`: same body → **EXIT 2 (BLOCK)**, message names `project repo: octoworld/beta-worker` (slug scrub, not name). All three slugs (primary + both non-primary) scrub; an unregistered slug correctly passes.

**2. Comment-tolerance opened no new hole.**
- Over-arm: a comment header carrying `[`, `repos:`, and quotes arms correctly and does **not** flow-parse or capture the decoy inside the comment. Flow-form `repos: [a,b] # c` strips the trailing comment and captures both. Commented `tags:` / `roles:` headers are **not** treated as repos, and real sibling repos are still captured.
- Dis-arm: the new regex is strictly more permissive than base (everything base armed, fix arms → no arming regression); the added `-` in the disarm class only ends a block earlier (safe). A stray mid-block comment does not drop following repos.
- Fail-safe: empty/commented-but-item-less `repos:` header yields no phantom repo while the project name still scrubs; the scrub list is never silently empty.

**3. No fail-open across resolution paths.** The leak hook is awk-only. Base-vs-fix diff *outside* the awk body and comments is empty; the 13 `exit` statements are byte-identical in order — no new silent-allow path. `_lib-multi-repo-trace.sh` lists all slugs (incl. non-primary) on **both** its yq and awk-fallback paths. `_lib-tracker.sh` `contains([...])` resolves a non-primary repo's tracker override (returns it), and the rejected `index()` form indeed errors on mikefarah yq v4.53.3 — validating the author's note.

**4. Backward-compat holds.** A singular `repo:` project blocks identically on base and fix.

### Security Issues Found
None blocking. Two **LOW** residuals, both pre-existing (identical on base and fix) and out of scope for this PR — noted for a future hardening pass, not a reason to hold:
- **LOW** — A malformed `repos: <scalar>` (plural key, bare scalar value, no brackets/list) is captured on neither base nor fix, so its slug would not scrub. Malformed YAML; the awk fallback has no inline-scalar handler for the plural key. Not introduced here.
- **LOW** — A block-list item with a trailing inline comment (`- owner/repo  # note`) folds the comment into the captured token; shell word-splitting still scrubs the bare slug (fails *safe* — over-blocks on the comment words). Cosmetic noise, not a leak.

### Recommendations
Consider stripping inline comments from block-list items and handling the malformed plural-scalar form in a later pass; neither blocks this PR.

### Verdict
**APPROVED (SECURE)** — the disclosure boundary is closed. Every multi-repo slug (primary and non-primary), under bare, commented, and flow-form `repos:` headers, now reaches the scrub list; the comment tolerance introduced no over-arm/dis-arm evasion; no fail-open path was added; singular `repo:` is unregressed.

---
📌 Reviewed commit: `125ea58f6b7ce029fd970f16d9b25734aafa8ed9`

me2resh (commented):
## Code Review: PR #1129 (delta re-review)

**Commit**: `0bb60cca3a5876612d82d5f0205b39083bc3595d`
**Scope**: `Full` (delta re-review — trust-chain path, not Lean-eligible; but the delta is a single test-harness file and the security code is byte-identical to the previously-APPROVED base `125ea58`)

### Summary
Follow-up re-review after a single CI-fix commit on top of the already-APPROVED leak-scrub change (multi-repo `repos:` block-lists). The only file changed from `125ea58` → `0bb60cc` is `.claude/hooks/tests/test_multi_repo_registry.sh` (+22/−9). It repairs a CI-only false negative in the trace-lib awk-fallback test — no production/security code changed.

### Delta verification
- **Test-only diff.** `git diff --name-only 125ea58 0bb60cc` = exactly `.claude/hooks/tests/test_multi_repo_registry.sh`. Nothing else.
- **Security code byte-identical to the APPROVED base.** `git diff 125ea58 0bb60cc` for both `block-private-refs-in-public-repos.sh` and `_lib-multi-repo-trace.sh` is empty. The leak-scrub parser and trace lib are unchanged since the APPROVED `125ea58` review, so no security-relevant surface moved.

### What the test change does (and why it's sound, not a weakening)
The `no-yq` environment for the awk-fallback case was previously built by dropping every PATH directory that contained `yq`. On Ubuntu CI both `yq` and `awk` live in `/usr/bin`, so that also removed `awk` — the awk fallback then had no `awk`, produced empty output, and the case failed on CI while passing on macOS. The fix builds a shim bin dir that symlinks every executable **except** `yq`, then points `PATH` at it, so `command -v yq` still fails (fallback is exercised) while `awk`/`grep`/`sed`/etc. stay reachable regardless of host layout.

Assessed as a **more faithful env-mock, not a relaxed assertion**:
- Skips `yq` explicitly (`[ "$base" = yq ] && continue`); the fallback branch is still the one under test.
- First-wins PATH-precedence guard (`[ -e "$NOYQ_BIN/$base" ] && continue`) mirrors real PATH resolution.
- Guards non-existent/empty PATH dirs and non-executables; tolerant of `ln` failures.
- Cleans up the new temp dir (`rm -rf "$TRACE_NOYQ_TMPDIR" "$NOYQ_BIN"`).
- **Assertion unchanged:** still `grep -q "^resolve_commented=meridian-svc$"` — the commented `repos:` header (`corp/nightshade-api`) must still resolve to `meridian-svc` via the awk fallback. Still degrades to SKIP (not pass) if `yq` can't be hidden. The test still proves exactly what it did before; only the environment construction is more robust.

### Checklist Results
- Architecture & Design:      Pass (test-harness only)
- Testing:                    Pass — the change strengthens test portability without loosening the assertion
- Security:                   Pass — leak-scrub parser + trace lib byte-identical to APPROVED base
- PR Description & Glossary:   Pass (carried from base review)
- Technical Decisions (AgDR): N/A — CI test-mock fix, no new dependency/architecture decision
- Adopter Handbooks:          N/A — loaded architecture/general handbooks have no triggers matching a shell test change; TS handbook doesn't load (no .ts/.tsx)

### Issues Found
None.

### Suggestions
- `nit:` pre-existing SC2005 (`Useless echo?`) at line 120 of the test file — present identically in the APPROVED base, outside this delta, and CI's `shellcheck .claude/hooks` job is green. Not blocking; worth a future cleanup pass.

### CI
Green on this HEAD — `hook test suite` passes (3m36s; failed on the prior HEAD), `shellcheck .claude/hooks` passes, all other checks pass/skip.

### Verdict
**APPROVED**

Marker re-established at HEAD `0bb60cca3a5876612d82d5f0205b39083bc3595d`.

---
📌 Reviewed commit: `0bb60cca3a5876612d82d5f0205b39083bc3595d`

Issue #1126, opened 2026-08-01
[Refactor] Align remaining lib self-location sites onto resolve_anchored_lib_dir() (finish #1102 drift cleanup)


me2resh:
## Driver

The Rex + Hakim reviews on PR #1125 (issue #1102) found lib/hook self-location sites still on the old `${BASH_SOURCE[0]:-$0}` / bare `${BASH_SOURCE[0]}` idiom that #1102 deliberately left out of scope — notably `maintain-docs-index.sh:37` (which sources siblings before its enabled-gate), plus a handful of top-level hook scripts. These run under their own `#!/bin/bash` shebang, so `BASH_SOURCE` is populated and they are lower-risk defense-in-depth — not the live zsh-sourced exploit paths #1102 fixed. But per AgDR-0118's own "stop the drift" thesis, one idiom expressed several ways is exactly the defect generator to eliminate; leaving these on the old spelling keeps the drift alive.

## Scope

Route every remaining self-locating hook/lib through the shared `resolve_anchored_lib_dir()` helper in `_lib-ops-root.sh` (or the documented irreducible bootstrap snippet needed to LOCATE `_lib-ops-root.sh` itself), so no site remains on the bare / `:-$0` spelling. Start from `maintain-docs-index.sh:37` and the top-level hooks Rex/Hakim named; grep the tree for any other live instance.

## Acceptance Criteria

- [ ] A grep across `.claude/hooks/**` confirms zero LIVE self-location sites still on `${BASH_SOURCE[0]:-$0}` or an unguarded bare `${BASH_SOURCE[0]}` that self-locates + sources a sibling (the irreducible bootstrap snippet is the only permitted duplicated idiom)
- [ ] `maintain-docs-index.sh` locates + sources its siblings via the shared helper, and only after (or safely around) its enabled-gate
- [ ] Full `.claude/hooks/tests/` suite passes unchanged
- [ ] No gate behaviour changes — this is consistency/hardening, not a functional change

## Risks / Dependencies

- Depends on #1102 (merged) — `resolve_anchored_lib_dir()` now exists in `_lib-ops-root.sh`.
- Lower-risk than #1102: these sites run under bash (BASH_SOURCE populated), so they are defense-in-depth, not live exploit paths. Keep it that way — do not change any gate's behaviour.

## Glossary

| Term | Definition |
|------|------------|
[...]

PR #1130, opened 2026-08-02
refactor(#1126): route remaining hook self-location through resolve_anchored_lib_dir()

me2resh:
## Summary

- **Finishes the #1102 self-location drift cleanup (AgDR-0118).** The six remaining top-level hooks still self-located with the raw `${BASH_SOURCE[0]}` / `${BASH_SOURCE[0]:-$0}` idiom — which under a non-bash shell resolves `dirname ""` to `.` and silently substitutes the caller's cwd for the file's real directory (the cwd-substitution class #1102 fixed in the libs). They run under their own bash shebang so they were lower-risk defense-in-depth, but leaving them on the old spelling kept the "one idiom, many spellings" drift alive — the exact defect generator AgDR-0118 exists to eliminate.
- **All six sites now use the one shared anchored bootstrap + `resolve_anchored_lib_dir()`** (from `_lib-ops-root.sh`, #1102): `block-main-push.sh`, `detect-skill-intent.sh`, `maintain-docs-index.sh`, `reindex-on-session-start.sh`, and `suggest-mcp-search.sh` (two sites). One fixed idiom, not six drifting ones.
- **`block-main-push.sh` (the protected-branch BLOCKING backstop, AgDR-0114) is the highest-risk site — behaviour preserved exactly.** When self-location is unavailable, `HOOK_DIR` stays empty, `PROTECTED_LIB` resolves to a nonexistent path, and the existing `declare -F is_protected_branch` fallback fails **CLOSED** to the hardcoded default protected list — the same fail-closed path it already had for "lib missing/broken", now also covering "self-location unavailable". No behaviour change in the normal bash case. `test_block_main_push.sh` has the **identical** pass/fail set as clean dev (zero new failures).
- **Applies the existing AgDR-0118 decision — no new AgDR.** Adds a discriminating impostor test case per migrated hook to `test_lib_self_location_cwd_anchor.sh` (base sources the impostor, fix does not).

## Testing

- `shellcheck --severity=warning` clean on all five changed hooks.
- `test_block_main_push.sh`: identical 3-failure set on clean dev and this branch — those are **pre-existing macOS/BSD-only flakes** (config-override cases; a fail-*safe* over-block), not introduced here and green on the Ubuntu CI runner.
- New self-location impostor cases added per migrated hook. The zsh-dependent cases skip on the Ubuntu CI runner (no zsh installed) as the existing suite already does.

Refs #1126

## Glossary

| Term | Definition |
|------|------------|
| Self-location idiom | The shell snippet a hook uses to find its own directory (to source sibling libs); the raw `${BASH_SOURCE[0]}` form self-locates to the caller's cwd under a non-bash shell. |
| `resolve_anchored_lib_dir()` | The shared helper in `_lib-ops-root.sh` (#1102) that self-locates fail-closed and only accepts a genuine apexyard fork. |
| Anchored bootstrap | The irreducible few lines a hook needs to LOCATE `_lib-ops-root.sh` before it can call `resolve_anchored_lib_dir()` — one fixed idiom, unavoidably duplicated. |
| Fail closed | When self-location or the lib is unavailable, `block-main-push.sh` falls back to its hardcoded default protected-branch list rather than allowing a push through. |

### DIFF
--- a/.claude/hooks/block-main-push.sh
+++ b/.claude/hooks/block-main-push.sh
@@ -81,7 +81,46 @@ fi
 
 # Resolve the hook's own directory so we can source sibling libs reliably
 # regardless of the harness's $PWD.
-HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+#
+# SELF-LOCATION BOOTSTRAP (me2resh/apexyard#1102 / #1126, AgDR-0118)
+# ------------------------------------------------------------
+# Was `HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` with no
+# anchor check on the resolved dir -- an unset/empty BASH_SOURCE[0]
+# (non-bash shell) makes `dirname ""` resolve to ".", silently substituting
+# the caller's cwd for this file's real location. That matters MORE here
+# than on the other sites migrated alongside this one: HOOK_DIR below is
+# used to locate PROTECTED_LIB="$HOOK_DIR/_lib-protected-branches.sh", and
+# this file is the protected-branch BLOCKING backstop (AgDR-0114) — a
+# cwd-substituted HOOK_DIR pointed at an attacker-controlled directory
+# could source an impostor lib defining a permissive is_protected_branch().
+# The bootstrap guard (raw BASH_SOURCE[0] captured once, empty
+# short-circuits to no self-location, never $0) plus
+# `resolve_anchored_lib_dir` in _lib-ops-root.sh (the anchor check) are the
+# ONE shared idiom every _lib-*.sh / hook self-location site now uses --
+# see that function's header for why the very first hop can never itself
+# be centralized behind a sourced call. When self-location is unavailable,
+# HOOK_DIR stays empty, PROTECTED_LIB resolves to a nonexistent path, and
+# the existing "declare -F is_protected_branch" fallback below fails
+# CLOSED to the hardcoded default list — the same fail-closed behaviour
+# this file already had for "lib file missing/broken", now also covering
+# "self-location unavailable". No behaviour change for the normal case
+# (bash, BASH_SOURCE[0] populated): HOOK_DIR resolves exactly as before.
+RAW_BASH_SOURCE_0="${BASH_SOURCE[0]:-}"
+HOOK_DIR=""
+if [ -n "$RAW_BASH_SOURCE_0" ]; then
+  _CANDIDATE_DIR="$(cd "$(dirname "$RAW_BASH_SOURCE_0")" 2>/dev/null && pwd)"
+else
+  _CANDIDATE_DIR=""
+fi
+if [ -n "$_CANDIDATE_DIR" ] && [ -f "$_CANDIDATE_DIR/_lib-ops-root.sh" ]; then
+  # shellcheck source=./_lib-ops-root.sh
+  # shellcheck disable=SC1091
+  . "$_CANDIDATE_DIR/_lib-ops-root.sh"
+  if command -v resolve_anchored_lib_dir >/dev/null 2>&1; then
+    HOOK_DIR="$(resolve_anchored_lib_dir "$RAW_BASH_SOURCE_0")"
+  fi
+fi
+unset _CANDIDATE_DIR
 
 # NOTE on heredoc bodies (me2resh/apexyard#1066, #1075): every presence
 # check and cd-target extraction below runs against the RAW $COMMAND, never
--- a/.claude/hooks/detect-skill-intent.sh
+++ b/.claude/hooks/detect-skill-intent.sh
@@ -48,11 +48,35 @@ PROMPT=$(printf '%s' "$INPUT" | jq -r '.prompt // empty' 2>/dev/null)
 
 # ---------------------------------------------------------------------------
 # Load the phrase → skill map from project config (data-driven, #894 AC 2).
-# HOOK_DIR resolves via BASH_SOURCE so this works regardless of the
-# caller's cwd (same pattern _lib-read-config.sh uses to find its
-# sibling _lib-ops-root.sh).
+#
+# SELF-LOCATION BOOTSTRAP (me2resh/apexyard#1102 / #1126, AgDR-0118)
+# ------------------------------------------------------------
+# Was `HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" 2>/dev/null && pwd)"`
+# with no anchor check on the resolved dir -- an unset/empty BASH_SOURCE[0]
+# (non-bash shell) makes `dirname ""` resolve to ".", silently substituting
+# the caller's cwd for this file's real location. The bootstrap guard (raw
+# BASH_SOURCE[0] captured once, empty short-circuits to no self-location)
+# plus `resolve_anchored_lib_dir` in _lib-ops-root.sh (the anchor check)
+# are the ONE shared idiom every _lib-*.sh / hook self-location site now
+# uses -- see that function's header for why the very first hop can never
+# itself be centralized behind a sourced call.
 # ---------------------------------------------------------------------------
-HOOK_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" 2>/dev/null && pwd)"
+RAW_BASH_SOURCE_0="${BASH_SOURCE[0]:-}"
+HOOK_DIR=""
+if [ -n "$RAW_BASH_SOURCE_0" ]; then
+  _CANDIDATE_DIR="$(cd "$(dirname "$RAW_BASH_SOURCE_0")" 2>/dev/null && pwd)"
+else
+  _CANDIDATE_DIR=""
+fi
+if [ -n "$_CANDIDATE_DIR" ] && [ -f "$_CANDIDATE_DIR/_lib-ops-root.sh" ]; then
+  # shellcheck source=./_lib-ops-root.sh
+  # shellcheck disable=SC1091
+  . "$_CANDIDATE_DIR/_lib-ops-root.sh"
+  if command -v resolve_anchored_lib_dir >/dev/null 2>&1; then
+    HOOK_DIR="$(resolve_anchored_lib_dir "$RAW_BASH_SOURCE_0")"
+  fi
+fi
+unset _CANDIDATE_DIR
 if [ -z "$HOOK_DIR" ] || [ ! -f "$HOOK_DIR/_lib-read-config.sh" ]; then
   # Can't locate the config lib — degrade silently. Advisory hook, no map,
   # nothing to fire.
--- a/.claude/hooks/maintain-docs-index.sh
+++ b/.claude/hooks/maintain-docs-index.sh
@@ -34,7 +34,35 @@ set -u
 
 INPUT=$(cat)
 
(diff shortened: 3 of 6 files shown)


me2resh (commented):
## Code Review: PR #1130

**Commit**: `eb2a89f4d042459998386bd3bd3bae19cfe6c54b`
**Scope**: Full (Heavy — trust-chain)

### Summary
Finishes the #1102 self-location drift cleanup (AgDR-0118) by routing the six remaining raw `${BASH_SOURCE[0]}` self-location sites — across five hooks — through the shared anchored bootstrap + `resolve_anchored_lib_dir()`. The highest-risk site is `block-main-push.sh`, the live protected-branch BLOCKING backstop (AgDR-0114). I reviewed this as a Heavy trust-chain change with the fail-closed property as the load-bearing question.

### Checklist Results
- Security:                  Pass (trust-chain — strictly improved, see below)
- Performance:              Pass (n/a)
- PR Description & Glossary:  Pass (narrative Summary, `Refs #1126`, 4-term Glossary)
- Technical Decisions (AgDR): N/A — applies existing AgDR-0118, no new material decision

### What I verified

**1. `block-main-push.sh` behaviour is preserved EXACTLY.** The only change is the HOOK_DIR self-location block. Everything from `# NOTE on heredoc bodies` to EOF is **byte-identical** between base `dev` and this HEAD (383 lines each, `diff` clean). The blocking logic, exit codes, protected-branch resolution, decoy/heredoc defences, and the `declare -F is_protected_branch` fail-closed fallback are untouched.

**2. Fail-CLOSED is preserved and actually strengthened — it never fails open.**
- Empty `BASH_SOURCE` (non-bash shell) → `RAW_BASH_SOURCE_0` empty → `HOOK_DIR=""` → `PROTECTED_LIB="/_lib-protected-branches.sh"` (nonexistent) → not sourced → `declare -F is_protected_branch` false → hardcoded default list (`main|master|dev|develop`) defined → blocks `main`. Verified by the new `block-main-push FIX (zsh): impostor NOT sourced, falls back to fail-closed default list` case.
- This is **strictly safer** than base: under the old bare idiom, an empty `BASH_SOURCE` made `dirname "" → "."`, so HOOK_DIR resolved to the caller's cwd and the hook would source an impostor `_lib-protected-branches.sh` from an attacker-controlled directory (a permissive `is_protected_branch` → fail OPEN). The new code leaves HOOK_DIR empty and never consults cwd. The test's BASE case reproduces exactly this impostor-load; the FIX case closes it.
- Normal bash deployment is unchanged: `resolve_anchored_lib_dir` returns `cd dirname && pwd` for a populated `BASH_SOURCE`, so HOOK_DIR resolves identically to before.

**3. Idiom completeness.** All six sites (5 files; `suggest-mcp-search.sh` collapses two prior sites into one shared HOOK_DIR) use the identical anchored bootstrap + `resolve_anchored_lib_dir()`. Re-grep of the PR HEAD content shows no raw self-location idiom left in any executable line — every `BASH_SOURCE` reference is now the one captured `${BASH_SOURCE[0]:-}` bootstrap. The irreducible bootstrap-to-locate-`_lib-ops-root.sh` is the documented one (a sourced helper cannot locate the file it lives in).

**4. Tests genuinely discriminate, and are CI-portable.** Run against the true merge-base (`APEXYARD_TEST_BASE_REF`), all #1126 cases pass: BASE reproduces the impostor-source bug, FIX closes it, genuine bash is unaffected — for `maintain-docs-index`, `block-main-push`, `detect-skill-intent`, `reindex-on-session-start`, and `suggest-mcp-search`. The new cases use `cd`-into-impostor + `zsh -c` + marker files and do **no** PATH surgery, so they avoid the #1129 CI trap; the file-level `command -v zsh` guard skips the whole suite on the Ubuntu runner (no zsh), as the existing suite already does.

**5. `test_block_main_push.sh` is unchanged in outcome.** Base `dev` and this HEAD both produce an identical 41 PASS / 3 FAIL set; the 3 failures are the same `config override REPLACES default list` cases — pre-existing macOS/BSD config-override flakes that fail in the **fail-SAFE over-block** direction (`rc=2` where `rc=0` was expected), not introduced here and green on Ubuntu CI.

**6. CI is green on this HEAD.** `hook test suite` pass (3m15s), `shellcheck .claude/hooks` pass, sub-pack smoke + ticket-ID checks pass, CodeQL skipping (normal). `shellcheck --severity=warning` is clean locally on all five hooks too. Not treating local-green as CI-green — the authoritative signal is the passing CI job.

**AgDR:** Correct that no new AgDR is needed. This applies the already-recorded AgDR-0118 decision (one self-location idiom); it's mechanical alignment of remaining sites, not a new material decision — even though `block-main-push.sh` is trust-chain, the decision it embodies was already made and recorded.

### Issues Found
None blocking.

### Suggestions
- nit (non-blocking): In the narrow, torn-clone edge where `_lib-ops-root.sh` is absent but `_lib-protected-branches.sh` is present, the new code leaves HOOK_DIR empty and falls back to the hardcoded default list instead of the project's `.git.protected_branches[]` override. Both libs ship/pull together, so this is very unlikely, and the direction is fail-SAFE (over-protect the defaults), never fail-open. Worth a one-line mention only if you touch this again.

### Verdict
**APPROVED**

---
Reviewed commit: `eb2a89f4d042459998386bd3bd3bae19cfe6c54b`

me2resh (commented):
## Security Review: PR #1130 — trust-chain (self-location drift cleanup)

**Commit**: `eb2a89f4d042459998386bd3bd3bae19cfe6c54b`
**Reviewer**: Hakim (Security Auditor) · lens: trust-chain integrity (#777) + fail-closed analysis of the protected-branch BLOCKING backstop (AgDR-0114)

### Summary

Routes the six remaining raw `${BASH_SOURCE[0]}` / `${BASH_SOURCE[0]:-$0}` self-location sites onto the shared anchored bootstrap + `resolve_anchored_lib_dir()` (from #1102/AgDR-0118). The one security-critical site is `block-main-push.sh`, whose `HOOK_DIR` locates `_lib-protected-branches.sh`; a cwd-substituted `HOOK_DIR` under a non-bash shell could otherwise source an impostor lib defining a permissive `is_protected_branch()`. I built a sandbox, ran the real hooks, and constructed the attack directly. **Verdict: SECURE.**

### Checklist Results
- Secrets & Credentials: Pass (none touched)
- Injection Prevention: Pass (no new command/parse surface; blocking logic unchanged)
- Auth & Authorisation (gate integrity): **Pass** — cannot be made to fail open (evidence below)
- Data Protection: Pass (n/a)
- Gate & Trust-Chain Integrity (#777): **Pass** — the change strictly *hardens* the backstop

### Evidence — the four primary questions

**1. `block-main-push.sh` cannot be made to fail OPEN — CONFIRMED.**
- **Attack built and run** (stronger than the PR's own test): a `zsh` invocation from a crafted cwd containing BOTH an impostor `_lib-protected-branches.sh` (permissive `is_protected_branch(){ return 1; }`) AND an impostor `_lib-ops-root.sh` (hijacking `resolve_anchored_lib_dir` to return the impostor cwd). Against the **FIX**, neither impostor is ever sourced (both marker files absent, every run). The empty-`BASH_SOURCE[0]` short-circuit (`RAW_BASH_SOURCE_0="${BASH_SOURCE[0]:-}"` → empty under zsh → `_CANDIDATE_DIR=""` → `_lib-ops-root.sh` never sourced → `HOOK_DIR` stays empty) is the real defense; the double-impostor never gets a foothold because the second-hop source only runs from `_CANDIDATE_DIR`, which is empty under zsh and the file's real dir under bash.
- **Fail-closed chain verified under bash** (the production path — hooks run via `#!/bin/bash` + harness): with `_lib-protected-branches.sh` absent, `declare -F is_protected_branch` is false → the hardcoded `main|master|dev|develop` fallback installs → `push origin main` is **BLOCKED (exit 2)**. Normal case (all libs present) also blocks `main`, exit 2.
- **No regression, strict improvement.** Against **base**, the same zsh+impostor attack *sources the attacker's permissive predicate and fails OPEN* (verified control run). The fix closes that. There is no push that base blocked and the fix lets through.
- **Honest caveat (pre-existing, not introduced here):** under a *non-bash* shell, `declare -F` is a bash builtin that doesn't install the fallback the bash way, so under zsh neither base nor fix reliably blocks. I proved this is pre-existing (base under a benign zsh cwd also fails to install the fallback, rc=0). Production hooks always run under bash, where the fail-closed chain works; the PR neither introduces nor worsens the zsh gap and removes the attacker-code path from it.

**2. Behaviour parity — CONFIRMED.** Diffing base vs HEAD with the self-location region stripped, the blocking logic (blocked refs/branches, `--no-verify` note, cd/worktree resolution, exit codes) is **byte-identical**; the only removed line is the single old `HOOK_DIR=` assignment. `test_block_main_push.sh` yields the **identical 3-failure set** on base (dev) and HEAD — pre-existing macOS/BSD config-override flakes that fail in the *safe over-block* direction (main still blocked), not introduced here and green on CI.

**3. The other five sites — CONFIRMED FAIL-SAFE.** `detect-skill-intent.sh`, `maintain-docs-index.sh`, `reindex-on-session-start.sh`, `suggest-mcp-search.sh` (×2) are all advisory/indexing hooks — none is a merge gate. Each guards on empty `HOOK_DIR` (`exit 0` / silent degrade / conditional source), and the suite confirms none sources the impostor. No security-relevant path is reached on self-location failure. Bonus trust-chain check: no *other* merge-gate hook (`block-unreviewed-merge`, `block-merge-on-red-ci`, `require-*-review`, `require-migration-ticket`, `require-active-ticket`, `check-secrets`, `block-private-refs-*`) still carries the raw sibling-sourcing idiom.

**4. Portable test mocks + CI — CONFIRMED.** No PATH surgery or coreutil removal (the #1129 awk-wipe trap is absent); zsh cases gate on `command -v zsh` and skip on the Ubuntu runner. **`hook test suite` CI check = PASS (3m15s)**; all other checks green (shellcheck, Analyze, Verify Ticket ID, sub-packs; CodeQL skipping).

### Findings

**[LOW / informational — non-blocking] BASE demonstration cases self-defeat on the PR branch.**
`test_lib_self_location_cwd_anchor.sh` defaults `BASE_REF="${APEXYARD_TEST_BASE_REF:-HEAD}"`. Run locally from the PR branch with zsh installed, `HEAD` is the *fixed* commit, so the "BASE (proves the bug)" cases materialize the fixed file and fail (`marker not created`) — making the suite red on a dev box that happens to have zsh. Confirmed root cause: with `APEXYARD_TEST_BASE_REF=dev` all 12 #1126 cases pass (BASE correctly demonstrates the impostor load; FIX correctly refuses it). This is inherited from the #1102 cases already in the file (their read-config BASE cases fail identically), invisible on CI (no zsh → skipped), and fails *loudly* rather than silently — so it never manufactures a false sense of security. Optional future polish: default `BASE_REF` to the integration-branch merge-base, or document `APEXYARD_TEST_BASE_REF` in the header. Not a merge blocker.

### Recommendations
- Consider the `BASE_REF` default tweak above so the demonstration cases are green for contributors on a zsh dev machine.

### Verdict
**APPROVED (SECURE)** — the migration hardens the protected-branch backstop and cannot be made to fail open; behaviour is byte-identical to the reviewed base outside self-location; the other five sites fail safe; CI is green.

---
📌 Reviewed commit: `eb2a89f4d042459998386bd3bd3bae19cfe6c54b`

atlas-apex (approved):

Issue #1131, opened 2026-08-02
[Chore] Ship the v5.3.0→v5.4.0 migration script (untrack project-config.json on adopter forks)


me2resh:
## Driver

Cutting the **v5.4.0** release. Unlike the last four hops (all no-op placeholders), v5.4.0 carries a **real, data-loss-preventing per-adopter migration** and must ship a migration script so `/update`'s per-release chain applies it.

The in-window change that forces it is **#1065** (`fix(#1031)`): `.gitignore` listed `.claude/project-config.json`, but the file was **also tracked in the index**, so the ignore rule was inert for every adopter fork. Because it was tracked, a plain `git checkout <branch>` silently overwrites the working copy with the indexed version — and for a split-portfolio adopter that **destroys the private `portfolio` block** (config-only path resolution, no sibling-discovery fallback, never in git to restore from). This happened on a real fork. #1065 fixed the framework side (untracked the file, shipped `project-config.example.json`), but an adopter on v5.3.0 **still has the file tracked in their own fork** — the data-loss window stays open for them until they untrack it.

## Scope

- Add `.claude/migrations/v5.3.0-to-v5.4.0.sh` — a **real** migration (not a no-op placeholder) that idempotently untracks `.claude/project-config.json` on the adopter's ops fork: guard on "currently tracked", `git rm --cached` (working-tree file preserved so the adopter's customisations survive as the now-correctly-ignored local file), stage-don't-commit, respect `APEXYARD_MIGRATION_QUIET`, exit codes 0/1/2 per `.claude/migrations/README.md`.
- Add the matching row to `docs/upgrading.md`'s "What each migration does" table.
- Add a focused test: tracked+gitignored → untracked after run, working-tree content preserved, idempotent re-run is a no-op, already-untracked is a no-op.

## Acceptance Criteria

- [ ] `.claude/migrations/v5.3.0-to-v5.4.0.sh` exists, is executable, and untracks a tracked `.claude/project-config.json` while leaving the working-tree file (and its content) in place.
- [ ] Re-running the script is a no-op; running it where the file is already untracked is a no-op.
- [ ] `docs/upgrading.md` "What each migration does" table has a v5.3.0-to-v5.4.0 row describing the untrack.
- [ ] A focused test covers the untrack, working-tree preservation, and idempotency; the existing `test_update_chain.sh` still passes.
- [ ] `/release` step 3's migration-script check finds the script (no warning).

## Risks / Dependencies

[...]

PR #1132, opened 2026-08-02
chore(#1131): ship the v5.3.0→v5.4.0 migration (untrack project-config.json)

me2resh:
## Summary

- **Ships the v5.3.0→v5.4.0 migration — a real one, not a no-op placeholder.** v5.4.0 is the first release since v5.2.0 that actually needs a per-adopter migration, and it's a data-loss fix. #1065 found that `.gitignore` listed `.claude/project-config.json` while the file was *also tracked in the index*, so the ignore rule was inert on every adopter fork. Because it was tracked, a plain `git checkout` silently overwrites the working copy — and for a split-portfolio adopter that destroys the private `portfolio` block, which is config-only and never in git to restore from (it happened on a real fork). #1065 fixed the framework side; this closes the same window on **adopter** forks, which still track the file until they untrack it.
- **The migration untracks `.claude/project-config.json` safely.** Guards on "currently tracked", runs `git rm --cached` so the working-tree file *and the adopter's customisations in it* stay exactly in place, stages the untrack rather than committing it (operator owns the commit, matching the rest of `/update`). Idempotent — a no-op if the file is already untracked — and it defers to the operator (exit 1) rather than force past an unexpected staged state, so it can never discard a staged change.
- **`docs/upgrading.md`** gets the matching "What each migration does" row so the table stays the source of truth, and `/release` step 3's migration-script check now finds a script for this hop instead of warning.
- **Focused test (`test_v54_migration.sh`, 4 cases)** proves the load-bearing behaviour: untrack + working-tree/content preservation, staged-not-committed, idempotent re-run, and already-untracked no-op. The existing `test_update_chain.sh` still passes (no regression to the chain walker).

The real-vs-no-op call and the guard-design choices are recorded in **AgDR-0122-v54-migration-untrack-project-config** (this release breaks the four-hop no-op streak because #1065 left adopter forks exposed).

## Testing

1. `shellcheck .claude/migrations/v5.3.0-to-v5.4.0.sh .claude/hooks/tests/test_v54_migration.sh` — clean.
2. `bash .claude/hooks/tests/test_v54_migration.sh` — 4/4 pass.
3. `bash .claude/hooks/tests/test_update_chain.sh` — 11/0 pass (chain regression).
4. Ran the migration on this fork (already untracked) → clean no-op, exit 0, `git status` unchanged.

Refs #1131

---

## Glossary

| Term | Definition |
|------|------------|
| Migration chain | The per-release scripts in `.claude/migrations/` that `/update` walks from an adopter's recorded framework version to the latest release tag, applying each in order. |
| `git rm --cached` | Removes a path from git's index (untracks it) while leaving the working-tree file untouched — the exact operation needed to stop tracking a file without deleting the adopter's content. |
| Inert `.gitignore` entry | A `.gitignore` line that has no effect because the path it names is already tracked; git only ignores *untracked* paths. |
| Split-portfolio | Adopter mode where private portfolio config lives in a sibling repo; the `portfolio` block in `project-config.json` is load-bearing and never committed, so overwriting it is unrecoverable. |

### DIFF
--- a/.claude/hooks/tests/test_v54_migration.sh
+++ b/.claude/hooks/tests/test_v54_migration.sh
@@ -117,7 +117,61 @@ else
   mark_fail "case4 build" "build_fork failed"
 fi
 
-rm -rf "$SB" "$SB2" 2>/dev/null
+# ---------------------------------------------------------------------------
+# Case 5: v1-anchor fork (legacy onboarding.yaml + apexyard.projects.yaml pair,
+#         no .apexyard-fork) → find_ops_root resolves via the legacy branch and
+#         the untrack still works. Gives find_ops_root's non-v2 path coverage.
+# ---------------------------------------------------------------------------
+SB3=$(mktemp -d) && SB3=$(cd "$SB3" && pwd -P)
+mkdir -p "$SB3/.claude"
+: > "$SB3/onboarding.yaml"
+: > "$SB3/apexyard.projects.yaml"
+printf '%s\n' '.claude/project-config.json' > "$SB3/.gitignore"
+printf '%s\n' "$CONFIG_CONTENT" > "$SB3/.claude/project-config.json"
+if (
+    cd "$SB3" || exit 99
+    git init -q && git config user.email t@t.t && git config user.name t || exit 99
+    git add onboarding.yaml apexyard.projects.yaml .gitignore >/dev/null 2>&1 || exit 99
+    git add -f .claude/project-config.json >/dev/null 2>&1 || exit 99
+    git commit -q -m fixture >/dev/null 2>&1 || exit 99
+  ); then
+  is_tracked "$SB3" || mark_fail "case5 precondition" "config should start tracked"
+  ( cd "$SB3" && APEXYARD_MIGRATION_QUIET=1 bash "$MIGRATION" ); rc=$?
+  if [ "$rc" -eq 0 ] && ! is_tracked "$SB3" && [ -f "$SB3/.claude/project-config.json" ]; then
+    mark_pass "v1-anchor fork (legacy pair) → find_ops_root resolves, untrack works"
+  else
+    mark_fail "case5 v1anchor" "rc=$rc tracked=$(is_tracked "$SB3" && echo y || echo n)"
+  fi
+else
+  mark_fail "case5 build" "build failed"
+fi
+
+# ---------------------------------------------------------------------------
+# Case 6: defer-not-force. Staged content differs from BOTH HEAD and the working
+#         tree → `git rm --cached` refuses (no -f), the migration exits 1, the
+#         file stays tracked, and the working-tree content is untouched. Locks
+#         the "never discard a staged change" safety contract.
+# ---------------------------------------------------------------------------
+SB4=$(mktemp -d) && SB4=$(cd "$SB4" && pwd -P)
+if build_fork "$SB4" yes; then
+  (
+    cd "$SB4" || exit 99
+    printf '%s\n' '{"_adopter":"STAGED-version-B"}' > .claude/project-config.json
+    git add -f .claude/project-config.json >/dev/null 2>&1   # index (staged) = B, differs from HEAD
+    printf '%s\n' '{"_adopter":"WORKTREE-version-C"}' > .claude/project-config.json  # working tree = C
+  )
+  ( cd "$SB4" && APEXYARD_MIGRATION_QUIET=1 bash "$MIGRATION" ); rc=$?
+  wt=$(cat "$SB4/.claude/project-config.json")
+  if [ "$rc" -eq 1 ] && is_tracked "$SB4" && [ "$wt" = '{"_adopter":"WORKTREE-version-C"}' ]; then
+    mark_pass "unexpected staged state → exit 1 defer-not-force, still tracked, worktree preserved"
+  else
+    mark_fail "case6 defer" "rc=$rc (want 1), tracked=$(is_tracked "$SB4" && echo y || echo n), wt=$wt"
+  fi
+else
+  mark_fail "case6 build" "build_fork failed"
+fi
+
+rm -rf "$SB" "$SB2" "$SB3" "$SB4" 2>/dev/null
 
 echo ""
 echo "v5.4.0 migration test: PASS=$PASS FAIL=$FAIL"
--- a/.claude/hooks/tests/test_v54_migration.sh
+++ b/.claude/hooks/tests/test_v54_migration.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+# Focused test for .claude/migrations/v5.3.0-to-v5.4.0.sh
+#
+# The v5.4.0 migration untracks .claude/project-config.json (the #1065
+# data-loss fix, adopter side). It must:
+#   - untrack a currently-tracked project-config.json (git rm --cached)
+#   - leave the working-tree file AND its content exactly in place
+#   - be idempotent (re-run is a no-op; already-untracked is a no-op)
+#   - stage the untrack, not commit it
+#
+# Sandbox-based: builds a synthetic ops fork under mktemp, tracks
+# project-config.json despite a .gitignore entry (reproducing the inert-ignore
+# state #1065 found), runs the real shipped migration, and asserts the end
+# state. Exit 0 if every case passes; 1 on first failure.
+
+set -u
+
+SRC_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+MIGRATION="$SRC_ROOT/.claude/migrations/v5.3.0-to-v5.4.0.sh"
+
+[ -f "$MIGRATION" ] || { echo "FAIL: missing $MIGRATION" >&2; exit 1; }
+[ -x "$MIGRATION" ] || { echo "FAIL: $MIGRATION not executable" >&2; exit 1; }
+
+PASS=0
+FAIL=0
+FAILED=""
+mark_pass() { echo "  ✓ $1"; PASS=$((PASS+1)); }
+mark_fail() { echo "  ✗ $1: $2" >&2; FAIL=$((FAIL+1)); FAILED="$FAILED\n  - $1"; }
+
+CONFIG_CONTENT='{"portfolio":{"registry":"../private/apexyard.projects.yaml"},"_adopter":"custom value that must survive"}'
+
+# build_fork <root> <track_config: yes|no>
+#   Creates a git repo anchored by .apexyard-fork, with .gitignore listing
+#   .claude/project-config.json. When track_config=yes, the config file is
+#   force-added (tracked despite the ignore) — the exact inert-ignore state.
+build_fork() {
+  local root="$1" track="$2"
+  mkdir -p "$root/.claude"
+  : > "$root/.apexyard-fork"
+  printf '%s\n' '.claude/project-config.json' > "$root/.gitignore"
+  printf '%s\n' "$CONFIG_CONTENT" > "$root/.claude/project-config.json"
+  (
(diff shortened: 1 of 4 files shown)


me2resh (commented):
## Code Review: PR #1132

**Commit**: `cf6b80dfd44186fe3b9e3c5e44af409dabd2a404`
**Scope**: `Full` (a migration script that mutates adopter git state — never eligible for reduced scope)

### Summary
Ships the real v5.3.0→v5.4.0 migration that untracks `.claude/project-config.json` on adopter forks — the adopter-side completion of the #1065 data-loss fix. While the file was tracked despite a `.gitignore` entry, a plain `git checkout` could silently overwrite it and, for split-portfolio adopters, destroy the config-only `portfolio` block that is never in git to restore from. The migration `git rm --cached`s the file (index-only untrack, working tree preserved), stages rather than commits, is idempotent, and defers to the operator rather than forcing past an unexpected staged state. Ships with a 4-case focused test, the `docs/upgrading.md` row, and AgDR-0122.

### Checklist Results
- ✅ Architecture & Design:    Pass
- ✅ Testing:                   Pass
- ✅ Security:                  Pass
- ✅ Performance:               Pass (N/A — release tooling)
- ✅ Summary Bullet Narrative:  Pass (narrative, well over the label-only heuristic)
- ✅ Technical Decisions (AgDR):Pass (AgDR-0122 linked and thorough)
- ✅ Adopter Handbooks:         Pass (migration-safety evaluated below — N/A)
- ⚪ Fallow Static Analysis:    N/A (shell + markdown, not JS/TS)

### Focus-area findings (as requested)

**1. Guard correctness — correct and non-destructive.**
`if ! git ls-files --error-unmatch "$target" >/dev/null 2>&1; then … exit 0`
is the right guard: `ls-files --error-unmatch` exits 0 only when the path is in the index, so the body runs *only* on a currently-tracked file and no-ops on an already-untracked one. It can never touch an already-untracked file. `git rm --cached` removes the index entry only, never the working-tree file — the canonical "untrack without deleting" operation, and the correct non-destructive choice here. It cannot discard working-tree content.

**2. Idempotency — genuinely a no-op on re-run; the exit-1 branch is correct.**
After the first run the file is staged-for-deletion (out of the index), so a second run's `ls-files --error-unmatch` returns non-zero and the guard short-circuits (proven by test case 3). The defer-not-force branch is the notable strength: by omitting `-f`, the script lets git's own safety valve fire — `git rm --cached` refuses (non-zero) only when the staged content matches *neither* HEAD *nor* the working-tree file, i.e. a unique staged snapshot that would otherwise be lost. The script catches that, prints the raw git error plus manual-resolution guidance, and exits 1. This can never discard a staged change. The header comment describes git's semantics accurately.

**3. ops-root resolution — correct across both modes.**
`find_ops_root` walks up from `git rev-parse --show-toplevel`, anchoring on `.apexyard-fork` (split-portfolio v2) OR the `onboarding.yaml`+`apexyard.projects.yaml` pair (single-fork v1), falling back to the git toplevel otherwise. This mirrors `_migration_ops_root` in `_lib-migration-chain.sh` and the marker-resolution walk. `.claude/project-config.json` is root-relative in both modes, and the `cd "$ops_root"` before the relative `git rm --cached` resolves it correctly. Verified live: this very fork is split-portfolio v2 (`.apexyard-fork` present, no `apexyard.projects.yaml`), the exact case the anchor targets. The tracked-guard remains the real safety net even on the fallback path, as the comment notes.

**4. Test coverage — the load-bearing behaviours are all proven.**
Ran the suite: 4/4 pass. The fixture faithfully reproduces the inert-ignore state (`git add -f` despite the `.gitignore` line). Case 1 proves untrack + working-tree file + content preservation + exit 0; case 2 proves staged-not-committed (staged index deletion present AND still in HEAD); case 3 proves idempotent re-run; case 4 proves already-untracked no-op. Shellcheck clean on both files.

### Handbook Findings

⚪ **Migration Safety (blocking) — evaluated, NOT a violation** — `handbooks/architecture/migration-safety.md`
- The file path `.claude/migrations/v5.3.0-to-v5.4.0.sh` matches the handbook's `**/migrations/**` path trigger, so I evaluated it explicitly. The handbook's second (content) condition is NOT met: it flags *database schema* migrations containing SQL DDL / Prisma / TypeORM / Drizzle destructive operations that break the previous release during a rolling deploy. This script contains a `git rm --cached` index untrack — no schema, no DDL, no runtime data path. It is framework release tooling, not a DB migration (a distinction AgDR-0122 draws explicitly, and one the DB-migration gate `require-migration-ticket.sh` deliberately does not match this directory for). Verdict unaffected.

- `nit:` The exit-1 defer-not-force branch is the trickiest path and is the one branch the test doesn't exercise (a staged snapshot differing from both HEAD and the working tree). It's a genuinely awkward state to construct and the behaviour is delegated to git's proven safety valve, so this is optional — but a 5th case asserting exit 1 in that scenario would lock the contract the header comment promises.
- `suggestion:` The test fixture only builds the `.apexyard-fork`-anchored (v2) shape. A case exercising the single-fork v1 anchor (`onboarding.yaml` + `apexyard.projects.yaml`) and/or the git-toplevel fallback would give `find_ops_root` direct coverage. Low priority — the resolution logic is shared with the well-tested chain lib.

### Verdict
**APPROVED**

The migration is correct, well-guarded, idempotent, non-destructive, and the tests prove every load-bearing behaviour. Glossary present, AgDR-0122 linked and thorough, shellcheck clean, PR title and ticket link well-formed. The two suggestions are optional hardening, not blockers.

---
📌 Reviewed commit: `cf6b80dfd44186fe3b9e3c5e44af409dabd2a404`

me2resh (commented):
## Code Review (re-review): PR #1132

**Commit**: `fb52b706585297d4c473515b4d60564f45793474`
**Scope**: `Full` (delta re-review)

### Delta since the prior APPROVED review (`cf6b80d`)
Only `.claude/hooks/tests/test_v54_migration.sh` changed — the migration script (`v5.3.0-to-v5.4.0.sh`), `docs/upgrading.md`, and AgDR-0122 are byte-identical (confirmed via `git diff --name-only`). The delta adds exactly the two cases I raised as non-blocking suggestions:

- **Case 5 — v1-anchor coverage.** Builds a legacy single-fork (`onboarding.yaml` + `apexyard.projects.yaml`, no `.apexyard-fork`), tracks the config despite the ignore, runs the migration, asserts untrack succeeds and the working-tree file survives. Gives `find_ops_root`'s non-v2 branch direct coverage.
- **Case 6 — defer-not-force contract.** Stages version B (differs from HEAD) then writes version C to the working tree, so staged content differs from BOTH HEAD and the working tree. Asserts `git rm --cached` refuses (no `-f`), the migration exits 1, the file stays tracked, and the working-tree content (C) is preserved — locking the "never discard a staged change" safety contract.

Both cases are correctly constructed; the run output shows Case 6 genuinely triggering git's safety-valve refusal. Suite is 6/6, shellcheck clean. The delta strictly strengthens coverage with no change to production behaviour.

### Verdict
**APPROVED** — verdict unchanged; marker refreshed to `fb52b706585297d4c473515b4d60564f45793474`.

---
📌 Reviewed commit: `fb52b706585297d4c473515b4d60564f45793474`

me2resh (commented):
## Code Review (re-review): PR #1132

**Commit**: `ce596345230ecaeeb1722f0471cf7d0f38757384`
**Scope**: `Reduced-scope` (docs-only delta — Lean tier, AgDR-0116)

### Delta since `fb52b70`
Only `docs/agdr/AgDR-0122-...md` changed (confirmed via `git diff --name-only`) — two lines reworded from a leading `#1065` to "What #1065 found:" / "In #1065 the framework repo was fixed" so markdownlint MD018 no longer misreads them as a malformed atx heading. Prose meaning unchanged; no migration-script, test, or `docs/upgrading.md` change. Verdict unchanged.

### Verdict
**APPROVED** — marker refreshed to `ce596345230ecaeeb1722f0471cf7d0f38757384`.

---
📌 Reviewed commit: `ce596345230ecaeeb1722f0471cf7d0f38757384`

atlas-apex (approved):

me2resh (commented):
## Design Review: PR #1132

**Commit**: `ce596345230ecaeeb1722f0471cf7d0f38757384`

### Summary
AgDR-0122 records shipping a **real** v5.4.0 migration (`v5.3.0-to-v5.4.0.sh`) that untracks `.claude/project-config.json` on adopter forks — closing the #1065 data-loss window (a tracked-but-gitignored config a `git checkout` could silently overwrite, destroying a split-portfolio adopter's unrecoverable private `portfolio` block). Reviewed against the architecture lens with emphasis on migration safety.

### Review Lens Results
- Quality attributes / NFRs:    **Pass** — the relevant NFRs for a state-mutating shell migration (idempotency, non-forcing safety, operator-facing observability) are all concrete: documented exit-code contract (0/1/2), `APEXYARD_MIGRATION_QUIET` knob, stderr-vs-stdout discipline.
- Design patterns & structure:  **Pass** — slots into `/update`'s per-release migration chain (AgDR-0032); `find_ops_root` handles both split-portfolio v2 (`.apexyard-fork`) and legacy v1 (onboarding+registry pair) anchors, with a git-toplevel fallback backed by the tracked-guard as the real safety net.
- Technical debt:               **Pass** — none introduced; sets a healthy precedent (a config-tracking-invariant change ships a real migration, not a placeholder).
- Decisions (AgDR linkage):     **Pass** — AgDR-0122 IS the decision record: three genuine options (no-op placeholder / real migration / merge-semantics-only), chosen rationale, consequences, and linkage to #1065 + AgDR-0032 all captured. Nothing material undocumented.
- Risk:                         **Pass** — data-loss risk is the whole subject and is handled by the tracked-only guard plus the defer-not-force (exit 1) branch; blast radius bounded to the single config file.
- Trade-off analysis:           **Pass** — alternatives are weighed on real axes (risk-to-adopter-state vs. leaving the window open vs. unreliable merge semantics), not a single option dressed up.
- Requirements traceability:    **Pass** — directly satisfies what #1065 left open on the adopter side; the framework-side fix does not propagate through a normal sync, which is exactly why the migration exists.
- Migration safety:             **Pass** — load-bearing dimension, and it is sound:
  - `git rm --cached` is the correct working-tree-preserving cutover (stage, not commit); the adopter's on-disk customisations survive (test case 1 + content assertion).
  - Idempotent: the `git ls-files --error-unmatch` guard makes an already-untracked file a clean no-op (cases 3, 4).
  - Defer-not-force: no `-f`, so an unexpected staged state (differs from both HEAD and worktree) refuses, exits 1, leaves the file tracked and the working tree untouched — the "never discard a staged change" contract, locked by case 6.
  - Cross-consumer / blast radius: safe across single-fork and split-portfolio v2 (case 5 covers the legacy anchor); scoped to `.claude/project-config.json` only.
- Adopter Handbooks:            **N/A** — `handbooks/architecture/migration-safety.md` (blocking) evaluated and does not apply: its trigger requires a DB-migration path match AND SQL/Prisma/TypeORM **schema DDL** content. This migration is a bash git-index untrack with zero DDL, and the AgDR explicitly notes `.claude/migrations/` is distinct from the `require-migration-ticket.sh` DB-migration gate whose paths deliberately don't match. Concur with Rex's N/A.

### Blocking Findings
None.

### Suggestions
- (nit, non-blocking) In the staged==worktree-but-≠HEAD case, `git rm --cached` succeeds and stages the removal while the adopter's staged content is preserved on disk (worktree==staged) — no loss, correct behaviour, and worth a one-line comment near the defer branch for the next reader, since case 6 only exercises the all-three-differ path.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `ce596345230ecaeeb1722f0471cf7d0f38757384`

Issue #1161, opened 2026-09-02
[Bug] /code-review's forked run never writes its approval marker, so the merge gate cannot be satisfied


aureyia:
## Affected

`.claude/skills/code-review/SKILL.md` (steps 0 and 7), interacting with `warn-review-marker-write.sh` and `block-unreviewed-merge.sh`.

## Given / When / Then

**Given** a PR in a registered project, and `/code-review <owner/repo>#<pr>` invoked from the ops fork

**When** the skill runs — it launches as a forked background execution ("Running in the background as @code-review")

**Then** the agent returns a full, high-quality review as its result text, but **no** `.claude/session/active-reviewer` marker is ever written (step 0), **no** approval marker is written under `.claude/session/reviews/` (the `<owner>__<repo>__<pr>-rex.approved` file), and **no** review is posted to the PR via `tracker_review_submit` (step 7) — expected the approval marker to exist on an APPROVED verdict so `block-unreviewed-merge.sh` can pass.

Net effect: the two-marker merge gate is **unsatisfiable through the sanctioned path**. The orchestrator is left holding a genuine APPROVED verdict it cannot legitimately record, and the only ways forward are merging outside the gate or forging the marker — the exact failure mode #843 hardened against.

## Repro

1. In an ops fork, register a project and open a PR in it.
2. Run `/code-review <owner/repo>#<pr>`. Observe it reports "Running in the background as @code-review".
3. When the task notification arrives, the result contains a complete review with a clear verdict.
4. `ls .claude/session/reviews/` — no approval marker for that PR.
5. `cat .claude/session/active-reviewer` — absent (not merely cleaned up on exit; never created).
6. `gh pr view <pr> --json reviews` — `reviews: []`.
7. A filesystem-wide search for a recently-written approval marker returns nothing.

Observed on 3 consecutive runs across 2 different PRs in the same session. Both agents also reported that `ReportFindings` was unavailable and fell back to inline prose, which may share the same underlying cause: the forked execution lacking the tool and marker plumbing the skill body assumes it has.

## Framework version

v5.1.0

## Severity

`major` — wrong result with a workaround (merge outside the gate, or have a human merge on the forge). Not `blocker` because work can continue, but it silently defeats a security-critical gate: every merge in such a session ends up either ungated or tempting a forged marker.

## Notes

[...]

PR #1173, opened 2026-09-04
fix(#1161): make the merge gate satisfiable through the sanctioned path

me2resh:
Two of the three defects reported in #1161 are real and live on `dev`; the third was real at v5.1.0 and is already fixed. This PR fixes the two live ones and leaves the merge gate strictly as strong as it was.

## Summary

- **`/handover`'s own branches could not pass the fork's own branch gate.** Steps 8.5 and 8.6 hardcode `docs/agents-md` and `docs/apexyard-badge`, and `validate-branch-name.sh` rejected both with exit 2. Neither can carry a ticket-ID: `/handover` is bootstrap-class and runs before the adopted repo has a ticket to name, and that repo may have no tracker at all. Now exempt by **exact literal**, mirroring the `release/vN.N.N` and `sync/main-to-dev-after-vN.N.N` precedents this repo already set for the same reason. `docs/agents-md-v2` and every other `docs/*` branch still require a ticket.
- **A shadowed `/code-review` left the merge gate unsatisfiable, and the gate's own advice made it worse.** When Claude Code's *bundled* `/code-review` wins the name over ApexYard's skill, it runs as a background agent, posts nothing to the PR, and writes no marker. `block-unreviewed-merge.sh` then refuses the merge and tells the orchestrator to run `/code-review` — which runs the bundled skill again, writing no marker again. Both `block-unreviewed-merge.sh` and `auto-code-review.sh` now name the two identifying signs and give the recovery: set the active-reviewer marker, then spawn Rex directly with the Agent tool. Both say plainly not to hand-write the approval marker.
- **13 regression cases** pin the exemption in both directions. The six near-miss cases are the load-bearing half — they fail if anyone later loosens the anchored regex into a `docs/*` glob.
- **AgDR-0129** records why the mechanical option was rejected: a `PreToolUse` hook sees tool calls, not which skill the harness resolved a name to, and the `Agent` spawn boundary has no matcher (AgDR-0056). The shadowed-skill half is therefore advisory by necessity, not by preference.

## What this deliberately does not do

No gate is relaxed. The merge gate still requires both markers, still compares each against the forge-reported HEAD, and still refuses when either is absent. The branch gate widens by exactly two names, both anchored, both only ever created by `/handover`, both landing as a PR in a repo ApexYard does not govern.

## On the third reported defect (already fixed)

The reporter also found `warn-review-marker-write.sh` blocking a `gh issue create` whose body merely *described* a marker path. Reproduced against v5.1.0: it exits 2 on two of three realistic tracker-create shapes. On current `dev` all three exit 0 silently. Fixed by #1000/#1011 (narrowed to actual write intent) and #1026 (returned to advisory, AgDR-0111), both shipped in **v5.3.0**. No change needed here.

## Testing

```bash
bash .claude/hooks/tests/test_validate_branch_name_handover.sh   # 13/13 pass
```

Full hook suite: 116 pass, 11 fail. All 11 failures reproduce identically with these three hook edits stashed, so they are pre-existing and fork-local (untracked premium skills and split-portfolio state), not regressions from this PR. `bash -n` and `shellcheck -S warning` are clean on all three edited hooks and the new test.

Manual check of the reported repro:

```bash
# before: exit 2, BLOCKED.  after: exit 0
echo '{"tool_input":{"command":"git push -u origin docs/agents-md"}}' \
  | bash .claude/hooks/validate-branch-name.sh
```

Refs #1161

---

## Glossary

| Term | Definition |
|------|------------|
| Approval marker | The per-PR file under `.claude/session/reviews/` that `block-unreviewed-merge.sh` reads as proof an independent code review happened |
| active-reviewer marker | A session file naming the sanctioned review pass in flight, written before spawning the reviewer |
| Bundled skill | A skill Claude Code ships itself, distinct from a project skill of the same name in `.claude/skills/` |
| Shadowing | A name collision where one skill definition wins `/code-review` over another, deciding which body actually runs |
| Bootstrap-class skill | A skill that runs before a portfolio or tracker exists, and is therefore exempt from the ticket-first gate |
| Exact-literal exemption | A fully anchored regex matching one whole name, so it cannot widen to a prefix or glob by accident |

### DIFF
--- a/.claude/hooks/auto-code-review.sh
+++ b/.claude/hooks/auto-code-review.sh
@@ -137,6 +137,17 @@ build sub-agent just handed this PR back to you):
   author grading their own work. (The marker also suppresses an advisory
   warning on Rex's own marker write, but that is a side effect, not the
   reason — the hook warns and never blocks since #1026.)
+
+  ONE EXCEPTION, and only for you (me2resh/apexyard#1161). After the review
+  returns, confirm the approval marker exists. If the review ran "in the
+  background" and posted nothing to the PR, the harness ran its OWN bundled
+  /code-review instead of ApexYard's skill. That bundled skill knows nothing
+  about these markers, so re-running it cannot help. Only in that case: set
+  the active-reviewer marker yourself, then spawn the code-reviewer agent
+  (Rex) directly with the Agent tool. A CHANGES REQUESTED verdict is NOT this
+  case — a real review that requests changes writes no approval marker by
+  design, and the fix is to address the findings. Never write the approval
+  marker yourself in either case.
 --------------------------------------------------------------------------
 
 CEREMONY TIER (.claude/rules/right-size-ceremony.md, AgDR-0116): Rex runs on
@@ -153,13 +164,6 @@ marker exists. Do NOT quote a marker path when you invoke the review — the
 reviewer resolves its own repo-qualified path (me2resh/apexyard#1144); a path
 handed to it in a prompt overrides that and lands the marker where no gate
 reads it.
-
-After the review returns, confirm the marker exists. If the review ran "in the
-background" and returned findings but wrote no marker, the harness ran its own
-bundled /code-review instead of ApexYard's skill (me2resh/apexyard#1161). The
-bundled skill knows nothing about these markers, so re-running it cannot help.
-Set the active-reviewer marker, then spawn the code-reviewer agent (Rex)
-directly with the Agent tool. Do not write the approval marker yourself.
 
 This message is a reminder from the PostToolUse hook, not a tool error. The PR
 was created successfully.
--- a/.claude/hooks/auto-code-review.sh
+++ b/.claude/hooks/auto-code-review.sh
@@ -154,6 +154,13 @@ reviewer resolves its own repo-qualified path (me2resh/apexyard#1144); a path
 handed to it in a prompt overrides that and lands the marker where no gate
 reads it.
 
+After the review returns, confirm the marker exists. If the review ran "in the
+background" and returned findings but wrote no marker, the harness ran its own
+bundled /code-review instead of ApexYard's skill (me2resh/apexyard#1161). The
+bundled skill knows nothing about these markers, so re-running it cannot help.
+Set the active-reviewer marker, then spawn the code-reviewer agent (Rex)
+directly with the Agent tool. Do not write the approval marker yourself.
+
 This message is a reminder from the PostToolUse hook, not a tool error. The PR
 was created successfully.
 MSG
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -276,6 +276,16 @@ To unblock:
      /approve-merge ${PR_NUMBER} — that skill is human-only, you cannot
   4. Their invocation records the approval and performs the merge
 
+If you already ran /code-review for this PR and got a full review back, the
+harness may have run its OWN bundled /code-review instead of ApexYard's skill.
+The bundled skill runs as a background agent. It reports findings and writes no
+marker, so this gate can never pass from it. Two signs identify it: the run
+reports "Running in the background", and the review never posts to the PR.
+If you see those signs, do not re-run /code-review. Set the active-reviewer
+marker, then spawn the code-reviewer agent (Rex) directly with the Agent tool.
+Do not write the approval marker yourself. See .claude/rules/pr-workflow.md
+section "When the harness bundled skill shadows /code-review".
+
 Never skip this check — even for typo fixes. See .claude/rules/pr-workflow.md.
 MSG
   # Name the gate-invisible near-miss, if one is sitting on disk under the
--- a/.claude/hooks/block-unreviewed-merge.sh
+++ b/.claude/hooks/block-unreviewed-merge.sh
@@ -281,10 +281,13 @@ harness may have run its OWN bundled /code-review instead of ApexYard's skill.
 The bundled skill runs as a background agent. It reports findings and writes no
 marker, so this gate can never pass from it. Two signs identify it: the run
 reports "Running in the background", and the review never posts to the PR.
-If you see those signs, do not re-run /code-review. Set the active-reviewer
-marker, then spawn the code-reviewer agent (Rex) directly with the Agent tool.
-Do not write the approval marker yourself. See .claude/rules/pr-workflow.md
-section "When the harness bundled skill shadows /code-review".
+A CHANGES REQUESTED verdict is NOT this case. A real review that requests
+changes writes no approval marker by design. Address its findings instead.
+If you see both signs above, do not re-run /code-review. Set the
+active-reviewer marker, then spawn the code-reviewer agent (Rex) directly with
+the Agent tool. Do not write the approval marker yourself. See
+.claude/rules/pr-workflow.md section "When the harness bundled skill shadows
+/code-review".
 
 Never skip this check — even for typo fixes. See .claude/rules/pr-workflow.md.
 MSG
--- a/.claude/hooks/tests/test_validate_branch_name_handover.sh
+++ b/.claude/hooks/tests/test_validate_branch_name_handover.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+# Tests for validate-branch-name.sh's /handover branch exemption
+# (me2resh/apexyard#1161, AgDR-0129).
+#
+# /handover steps 8.5 and 8.6 push two hardcoded branches INTO AN ADOPTED REPO:
+#
+#   docs/agents-md        — the generated AGENTS.md operating manual
+#   docs/apexyard-badge   — the "Governed by ApexYard" README badge
+#
+# Neither carries a ticket-ID, and neither can: /handover is bootstrap-class and
+# runs before the adopted repo has any ticket to name — that repo may have no
+# tracker at all. Before #1161 the gate rejected both (exit 2), so the skill's
+# own prescribed branch could not satisfy the fork's own gate.
+#
+# The exemption is an EXACT LITERAL match on the two names. These tests pin both
+# directions, and the negative direction is the load-bearing half: a
+# near-miss like `docs/agents-md-v2` must still be BLOCKED. If someone later
+# loosens the anchored regex to a prefix or a `docs/*` glob, the "near-miss"
+# cases below fail — which is the point.
+#
+# Exit 0 if all cases pass; 1 on first failure.
+
+set -u
+
+SRC_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
+HOOK_SRC="$SRC_ROOT/.claude/hooks/validate-branch-name.sh"
+LIB_SRC="$SRC_ROOT/.claude/hooks/_lib-extract-push-ref.sh"
+LIB_STRIP_HEREDOC_SRC="$SRC_ROOT/.claude/hooks/_lib-strip-heredoc.sh"
+LIB_CONFIG_SRC="$SRC_ROOT/.claude/hooks/_lib-read-config.sh"
+LIB_PR_REPO_SRC="$SRC_ROOT/.claude/hooks/_lib-pr-repo.sh"
+
+for f in "$HOOK_SRC" "$LIB_SRC"; do
+  if [ ! -f "$f" ]; then
+    echo "FAIL: required source missing: $f" >&2
+    exit 1
+  fi
+done
+
+PASS=0
+FAIL=0
+FAILED_CASES=""
+
+# Sandbox whose LOCAL branch intentionally fails validation, so a case that
+# passes proves the hook read the push-ref from the command, not local HEAD.
+make_sandbox() {
+  local sb
+  sb=$(mktemp -d)
+  (
+    cd "$sb" || exit 1
+    git init -q
+    git config user.email "test@example.com"
+    git config user.name "test"
+    : > onboarding.yaml
+    git add onboarding.yaml
+    git commit -q -m "init"
+    git checkout -q -B "not-conforming-branch-name"
+  )
+  mkdir -p "$sb/.claude/hooks"
+  cp "$HOOK_SRC" "$sb/.claude/hooks/validate-branch-name.sh"
+  cp "$LIB_SRC"  "$sb/.claude/hooks/_lib-extract-push-ref.sh"
+  [ -f "$LIB_STRIP_HEREDOC_SRC" ] && cp "$LIB_STRIP_HEREDOC_SRC" "$sb/.claude/hooks/_lib-strip-heredoc.sh"
+  [ -f "$LIB_CONFIG_SRC" ] && cp "$LIB_CONFIG_SRC" "$sb/.claude/hooks/_lib-read-config.sh"
+  [ -f "$LIB_PR_REPO_SRC" ] && cp "$LIB_PR_REPO_SRC" "$sb/.claude/hooks/_lib-pr-repo.sh"
+  [ -f "$SRC_ROOT/.claude/project-config.defaults.json" ] && \
+    cp "$SRC_ROOT/.claude/project-config.defaults.json" "$sb/.claude/project-config.defaults.json"
+  chmod +x "$sb/.claude/hooks/validate-branch-name.sh"
+  echo "$sb"
+}
+
+run_case() {
+  local label="$1" cmd="$2" want_rc="$3"
+  local sb; sb=$(make_sandbox)
+  local input got_rc got_stderr
+  input=$(jq -nc --arg c "$cmd" '{tool_input:{command:$c}}')
+  got_stderr=$(cd "$sb" && echo "$input" | bash .claude/hooks/validate-branch-name.sh 2>&1 >/dev/null)
+  got_rc=$?
+  rm -rf "$sb"
+
+  if [ "$got_rc" != "$want_rc" ]; then
+    echo "FAIL [$label]: want rc=$want_rc, got $got_rc" >&2
+    echo "    cmd: $cmd" >&2
+    echo "    stderr: ${got_stderr:0:300}" >&2
+    FAIL=$((FAIL+1)); FAILED_CASES="${FAILED_CASES}${label} "
+    return
+  fi
+  echo "PASS [$label]"
+  PASS=$((PASS+1))
+}
+
+# -- Exempt: the two names /handover actually prescribes --------------------
+
+run_case "8.5 branch docs/agents-md is exempt" \
+  "git push -u origin docs/agents-md" 0
+
+run_case "8.6 branch docs/apexyard-badge is exempt" \
+  "git push -u origin docs/apexyard-badge" 0
+
+run_case "exempt branch still exempt without -u" \
+  "git push origin docs/agents-md" 0
+
+run_case "exempt branch still exempt with --force-with-lease" \
+  "git push --force-with-lease origin docs/apexyard-badge" 0
+
+# -- NOT exempt: near-misses must still require a ticket-ID -----------------
+#
+# This is the half that keeps the exemption narrow. Each of these differs from
+# an exempt name by a suffix, a prefix, or a path segment.
+
+run_case "near-miss docs/agents-md-v2 is BLOCKED" \
+  "git push -u origin docs/agents-md-v2" 2
+
+run_case "near-miss docs/agents-md/foo is BLOCKED" \
+  "git push -u origin docs/agents-md/foo" 2
+
+run_case "near-miss docs/agents is BLOCKED" \
+  "git push -u origin docs/agents" 2
+
+run_case "near-miss feature/agents-md is BLOCKED" \
+  "git push -u origin feature/agents-md" 2
+
+run_case "near-miss docs/apexyard-badge-2 is BLOCKED" \
+  "git push -u origin docs/apexyard-badge-2" 2
+
+run_case "unrelated ticketless docs branch still BLOCKED" \
+  "git push -u origin docs/update-readme" 2
+
+# -- Regression: the pre-existing exemptions and the normal path -----------
+
+run_case "release branch still exempt" \
+  "git push -u origin release/v5.4.0" 0
+
+run_case "sync branch still exempt" \
+  "git push -u origin sync/main-to-dev-after-v5.4.0" 0
+
+run_case "conforming ticket branch still passes" \
+  "git push -u origin fix/GH-1161-code-review-fork-marker" 0
+
+# -- Summary ---------------------------------------------------------------
+
+echo
+echo "passed: $PASS   failed: $FAIL"
+if [ "$FAIL" -ne 0 ]; then
+  echo "failed cases: $FAILED_CASES" >&2
+  exit 1
+fi
+exit 0
--- a/.claude/hooks/validate-branch-name.sh
+++ b/.claude/hooks/validate-branch-name.sh
@@ -254,6 +254,23 @@ if echo "$CURRENT_BRANCH" | grep -qE '^sync/main-to-dev-after-v[0-9]+\.[0-9]+\.[
   exit 0
 fi
 
+# Allow the two branches /handover prescribes for its opt-in writes INTO AN
+# ADOPTED REPO (me2resh/apexyard#1161). Steps 8.5 and 8.6 push `docs/agents-md`
+# (the generated AGENTS.md) and `docs/apexyard-badge` (the README badge) into
+# the target repo. Both are hardcoded in the skill, so the skill's own
+# prescribed branch could never satisfy this gate: /handover is bootstrap-class
+# and runs before the adopted repo has any tracker ticket to name, and the
+# adopted repo may have no tracker at all. Same narrow, intentional exception
+# as the release and sync branches above — the work itself is the ticket.
+#
+# The match is an exact literal on each of the two names, so this exempts
+# nothing else: a `docs/agents-md-v2` or any other `docs/*` branch still needs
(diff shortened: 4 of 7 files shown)


me2resh (commented):
## Code Review: PR #1173

**Commit**: `b0018d9b7fe68f55f35e86a8767ae64a532bc41f`
**Scope**: Full (trust-chain diff — reduced scope correctly not applicable; rail 1 of `right-size-ceremony.md`)

### Summary

Fixes two live defects from #1161: `/handover`'s two hardcoded branches could not pass this fork's own branch gate, and a shadowed bundled `/code-review` left the merge gate unsatisfiable while the gate's own advice looped the orchestrator back into the failure. One behavioural hook change (an anchored two-name branch exemption), two message-only hook changes, a rule section, a skill note, AgDR-0129, and 13 regression cases.

I verified every load-bearing claim independently rather than from the PR body. All five review questions are answered below with the evidence.

### Checklist Results

- Code Quality: **Pass**
- Testing: **Pass** (one non-blocking coverage gap, below)
- Security: **Pass**
- Performance: **Pass** (N/A — two `grep -qE` calls on a short string)
- Summary Bullet Narrative: **Pass** (every bullet gives what changed *and* why)
- Technical Decisions (AgDR): **Pass** — AgDR-0129 present, unique, linked, and covers the rejected options plus the "no mechanical detection is possible" reasoning
- Adopter Handbooks: **N/A** — `migration-safety.md` (the only blocking handbook) matches no path in this diff; `clean-architecture-layers.md` and `language/typescript` have no matching files; `general/commit-message-quality.md` is advisory and the commit satisfies it
- Fallow Static Analysis: **N/A** — no JS/TS in the diff

### Answers to the five questions

**1. Is the exemption regex genuinely anchored? Can anything slip through?**

Yes, anchored; no bypass found. `^docs/(agents-md|apexyard-badge)$` is anchored at both ends. I probed 13 adversarial names against the shipped hook. All of these are correctly **BLOCKED**: `evil-docs/agents-md`, `my-docs/apexyard-badge`, `xdocs/agents-md`, `sub/docs/docs/agents-md`, `docs/agents-md.lock`, `docs/agents-mdX`, `docs/agents-mdz`, `docs/apexyard-badgez`, `docs/agents-md-v2`, `docs/agents-md/foo`, `docs/agents`, `DOCS/AGENTS-MD` (case-sensitive, correct for git refs).

One result needed classification: `refs/heads/docs/agents-md` is allowed. That is **not** a bypass — it is pre-existing normalisation in `_lib-extract-push-ref.sh` (`ref="${ref#refs/heads/}"`, lines 331/478), it names the identical branch, it applies equally to the pre-existing `release/` exemption, and `refs/heads/bogus-branch` is still blocked.

**2. Does the placement matter? Can it shadow anything?**

Placement is correct and cannot short-circuit any earlier check. The exemption sits at line 270, *after* the push-presence check (29), the tag-push guard and `SCAN_LAST_REF` per-occurrence walk (107–127), the fail-closed heredoc-disagreement block that exits 2 (147–169), the `BRANCH_DIR` re-rooting (187–200), `CURRENT_BRANCH` resolution (202–208), the trunk whitelist (233), and the `release/` (244) and `sync/` (253) exemptions. The only thing it precedes is the final ticket-ID check — exactly its intended scope. Empirically confirmed: `test_validate_branch_name_heredoc.sh`, `test_validate_branch_name_pushref.sh`, `test_extract_push_ref_arrow.sh` (9/9), `test_extract_push_ref_heredoc.sh` (12/12), and `test_pre_push_gate.sh` all pass at this HEAD.

**3. Are the two message-only edits genuinely message-only?**

Confirmed mechanically. I re-diffed both hooks with all heredoc bodies excluded by a delimiter-tracking parser; the result is *no change outside heredoc bodies* in either file. No exit-code path, condition, or variable is touched. I also checked the two specific ways a heredoc "message-only" edit can silently become executable: the added text contains no line equal to the `MSG` delimiter (which would terminate the heredoc early), and no `$` or backtick (both heredocs are unquoted, so an expansion would have evaluated). Clean on both counts. `bash -n` and `shellcheck -S warning` pass on all three hooks and the new test.

**4. Do the 13 cases pin what they claim? Is the negative direction real?**

The negative direction is genuinely load-bearing — I proved it by mutation rather than by reading:

| Mutation | Result |
|---|---|
| Loosen to `^docs/` (the glob the comment warns about) | **5 negative cases fail** |
| Drop the trailing `$` | **3 negative cases fail** |
| Remove the exemption entirely | **4 positive cases fail** |
| Drop the leading `^` | **13/13 still pass** — not caught |

So three of the four ways to weaken this are pinned. The fourth is a real, if small, coverage gap: no case pins the *leading* anchor. The shipped hook does block `evil-docs/agents-md` (verified), so behaviour is correct — only the test does not defend it. See the suggestion below. Non-blocking.

**5. Does the new banner advice risk inducing a hand-written approval marker?**

No — it is written specifically against that failure mode, and it strengthens rather than weakens the #843/#1144 hardening. Three things make it safe:

- Both banners state **"Do not write the approval marker yourself"**, and the rule adds "that is marker forging regardless of how good the review was" — it names the exact rationalisation an agent holding a genuine APPROVED verdict would reach for.
- It gives **no marker path anywhere**, which is precisely what #1144 requires ("Say what to write, never where"). It defers to Rex resolving its own repo-qualified path.
- The one write it *does* instruct — the active-reviewer marker — is not a gate signal. It is the provenance marker, and `pr-workflow.md` already names the orchestrator as its sanctioned writer.

The framing is also correct on the root cause: the old advice ("run `/code-review`") was itself the thing manufacturing pressure toward forging, because re-running the shadowed skill can never produce a marker. Replacing a dead-end remedy with a working sanctioned one removes pressure rather than adding it.

### Verification I ran independently

- **New test**: 13/13 pass.
- **Full hook suite at PR HEAD**: 11 failures out of 126.
- **Full hook suite at the merge base**: 11 failures out of 124. The two failure sets `diff` **identical**. All 11 are pre-existing and none is branch-name related — confirmed by running it, not assumed.
- **Cross-references**: AgDR-0129 is unique (no collision); the section title the hook cites exists verbatim in `pr-workflow.md:189`; the `#1144` anchor resolves to the heading at `:154`; `/handover` genuinely hardcodes both branches (`SKILL.md:1434/1445` and `:1570/1574`), so the exemption matches real prescribed behaviour.

### Issues Found

None blocking.

### Suggestions

1. `nit:` **Add one case pinning the leading anchor.** Removing the `^` survives all 13 cases. A single case pushing `evil-docs/agents-md` and expecting rc=2 would close it, and it fails today only under that mutation — the shipped hook already blocks it. Cheap, and it completes the "cannot widen" guarantee the test file's own header claims in both directions.

2. `suggestion:` **Pre-existing decoy weakness, out of scope — worth a follow-up ticket.** Decoy prose naming an exempt branch *before* the real push bypasses this naming gate, because `extract_push_ref` uses first-match semantics. Verified at the **merge base**, so this PR does not introduce it: the identical bypass works there against `release/`, `sync/`, and even the `main` trunk entry. The hook's own comments already document the first-match hazard (they hardened only the tag-push path via `SCAN_LAST_REF`). Blast radius is limited to naming hygiene — I confirmed `block-main-push.sh` resists the same decoy shape (rc=2), so the security-relevant gate holds. Flagging only so it is not mistaken for something this diff caused.

3. `nit:` The banner says "Set the active-reviewer marker" without its format or path. The format (`<owner>/<repo>#<pr>:<kind>`) is documented in the rule file the banner points to, so a reader can get there — but naming it inline would remove a guess. Harmless either way, since a wrong guess there is advisory-only.

### Verdict

**APPROVED**

The behavioural change is one anchored regex that widens the branch gate by exactly two literal names, both only ever produced by `/handover`, both landing in an adopted repo the fork does not govern. No gate is relaxed: the merge gate still requires both markers against the forge-reported HEAD. The two hook message edits are provably message-only, and their new advice moves an agent *away* from the forging pressure the old dead-end advice created.

---
📌 Reviewed commit: `b0018d9b7fe68f55f35e86a8767ae64a532bc41f`

me2resh (commented):
## Security Review: PR #1173

**Commit**: `b0018d9b7fe68f55f35e86a8767ae64a532bc41f`

### Summary

Trust-chain review of a gate-relaxing change. The security lens here is the integrity of the framework's own review and merge gates, not OWASP web classes.

Findings were verified by probing the hooks at this HEAD in throwaway sandboxes, not by trusting the PR body or the PR's own test file. **No gate is weakened, and I found no path by which an unreviewed change reaches `main`.** One MEDIUM issue is a prose contradiction inside `auto-code-review.sh`'s banner that re-opens the inducement shape AgDR-0056 closed; it is not a bypass, but it sits in the exact place whose historical failure mode was marker forging.

### Checklist Results

| Area | Result |
|------|--------|
| Gate & trust-chain integrity (#777) | Pass — with one MEDIUM (below) |
| Injection / heredoc expansion | Pass |
| Merge-gate marker semantics | Pass |
| Secrets & credentials | Pass — none introduced |
| Data protection / API security | N/A for this diff |

---

### 1. Blast radius of the branch exemption — genuinely limited to the two names

The exemption is `^docs/(agents-md|apexyard-badge)$` — anchored both ends, case-sensitive, and **hardcoded rather than config-driven**. That last point is a real strength worth naming: `trunk_whitelist` and `type_whitelist` above it are adopter-overridable, so an adopter config can widen *those*. This exemption cannot be widened without editing the anchored regex, which matches what AgDR-0129 claims.

Probed against the hook at this HEAD:

| Vector | Result |
|---|---|
| `docs/Agents-MD`, `DOCS/agents-md` (case) | BLOCKED |
| `docs%2Fagents-md` (URL-encoded) | BLOCKED |
| `docs/agents‐md` (Unicode U+2010 homoglyph) | BLOCKED |
| `docs/agents-md-v2`, `docs/agents-md/foo`, `docs/agents-md.` | BLOCKED |
| `docs/agents-md/../<trunk>` (traversal) | BLOCKED |
| `refs/heads/docs/agents-md`, `+docs/agents-md`, quoted | ALLOWED — correct, these normalise to the exempt name |
| refspec `docs/agents-md:bogus-name` | BLOCKED — the **destination** is validated, not the source |

That last row is the important one. `_extract_push_ref_core` takes the right-hand side of a `src:dst` refspec, so an exempt name on the *source* side cannot launder a non-conforming destination.

**Ref normalisation / encoding / newline injection.** The anchored-`grep` multiline bypass (a value whose second line matches while the real ref does not) is not reachable here. `PUSH_REF` is produced by unquoted word-splitting on default `IFS`, so a single token cannot carry a newline; the fallback path uses `git branch --show-current`, and git refnames forbid control characters.

I did find that a **quoted multi-word argument** lets the tokeniser see only its first word, so a command naming `"docs/agents-md<newline>bogus-branch"` exits 0. This is **not a finding against this PR**, for two independently verified reasons: it behaves identically with `release/v1.0.0` and with a plain conforming `fix/GH-1-x`, so it is pre-existing and name-independent; and `git check-ref-format` rejects both malformed refnames outright (`fatal: ... is not a valid branch name`), so no ref actually reaches the remote. Git's own refname validation is the backstop.

**Multi-push decoy** (exempt name first, real push second) also exits 0 — again reproduced identically with `release/v1.0.0` and with `fix/GH-1-x`. It is a pre-existing first-match property of this gate, and any attacker can already supply a conforming-looking name that need not exist. The exemption adds two more usable decoy names to an already-unbounded set, which is not a meaningful widening. Where it matters — a protected branch as the second push — `block-main-push.sh` catches it via its independent `_extract_all_push_refs_core` scan-all, which I confirmed blocks (rc=2) both the decoy shape and the `docs/agents-md:<trunk>` refspec on a complete install.

### 2. No other control is weakened — confirmed

`validate-branch-name.sh` is a standalone `PreToolUse` entry. Its `exit 0` returns from that hook only; it does not short-circuit any sibling. Verified in `.claude/settings.json`: `block-main-push.sh`, the five `block-unreviewed-merge.sh` wirings, and `require-active-ticket.sh` are separate entries that each run and each can block.

Specifically: the ticket-first gate is wired to `Edit|Write|MultiEdit` and to Bash-write shapes, never to the push command, so a branch-name exemption cannot reach it. The exemption also sits *after* the tag-push and fail-closed heredoc checks and short-circuits only the naming-convention check, which is the last thing in the file — nothing security-relevant follows it.

### 3. Message-only edits — no smuggled control flow

Both hunks are pure line insertions inside existing heredoc bodies.

Both heredocs are **unquoted** (`<<MSG`), so expansion is live — this was the right thing to check, since `block-unreviewed-merge.sh`'s body interpolates `${PR_NUMBER}`. The added lines contain **no `$`, no backtick, and no backslash**, so no expansion or command substitution was introduced. `bash -n` is clean on all three hooks. No `exit`, `return`, `if`, `cat`, or delimiter line was touched in either hook — the only control flow anywhere in the diff is the new test file's own logic and the four-line exemption block. The merge gate still exits 2 on a missing marker and still compares the recorded SHA against the forge-reported HEAD.

### 4. MEDIUM — the new banner paragraph contradicts the same banner's orchestrator branch

`auto-code-review.sh`'s banner is deliberately split into two audiences. Its **orchestrator** branch says, at lines 133-139:

> "The skill sets the active-reviewer session marker, spawns Rex, and clears the marker for you — **you should NOT set that marker by hand**. **Use the skill rather than spawning the code-reviewer agent directly**: it is what makes the review a separate, sanctioned pass instead of the author grading their own work."

The new paragraph, at lines 157-162, says:

> "**Set the active-reviewer marker, then spawn the code-reviewer agent (Rex) directly with the Agent tool.**"

Two direct contradictions, twenty-five lines apart, in one message. The new guidance is *correct on its merits* — when the skill is shadowed it is unavailable, so the fallback is legitimate and matches the new `pr-workflow.md` section. The problem is **placement**: it sits in the unbranched tail after the `---` separator, alongside the ceremony-tier and merge-gate notes that apply to everyone.

That has a specific consequence for the adversarial and the careless reader alike. The **build-class** branch tells its reader "You CANNOT nest the Agent tool, so you cannot spawn Rex yourself." A build agent that reads to the bottom now receives a closing, unconditioned instruction to set a marker and spawn Rex directly — an instruction it structurally cannot satisfy. That is precisely the shape AgDR-0056 identified as the root cause of #843, where the banner's old "Invoke Rex NOW" told a build agent to do something impossible and twice (PRs #835, #842) it resolved the contradiction by writing the marker itself.

Worth noting: the active-reviewer marker lives at `.claude/session/active-reviewer`, **not** under `.claude/session/reviews/`. So "set the active-reviewer marker" does not literally collide with the build branch's "do NOT write ANY file under `.claude/session/reviews/`" — which is exactly what lets a careless build agent start down this path and then arrive, stuck, at the pressure point.

Mitigations are genuinely present and I weighed them: both new paragraphs end with "Do not write the approval marker yourself", and the build branch remains emphatic. This is why the finding is MEDIUM and not HIGH — it creates pressure, it does not create a bypass.

**Recommended fix (text move, no logic change):** relocate the new paragraph *inside* the "IF YOU ARE THE ORCHESTRATOR" branch and frame it as an explicit exception, e.g. "Exception — only when the skill was shadowed, per the two signs below: ...". That restores the two-audience structure AgDR-0056 established.

The `block-unreviewed-merge.sh` copy has no such problem: it is properly conditioned ("If you already ran /code-review ... Two signs identify it ... If you see those signs") and contradicts nothing in that hook.

### 5. LOW — the trigger phrase also describes a legitimate CHANGES REQUESTED verdict

"returned findings but wrote no marker" is exactly what a correct **CHANGES REQUESTED** review looks like: Rex returns findings and deliberately writes no marker. The two identifying signs ("Running in the background", never posts to the PR) do disambiguate, so this is LOW rather than MEDIUM. But a reader who anchors on the summary phrase could read a legitimate non-approval as a shadowed-skill event and re-spawn Rex until an APPROVED verdict appears — review shopping through an escape hatch intended for a harness bug. Suggest naming CHANGES REQUESTED as an explicit non-trigger.

### 6. Net effect on forgeability of the two-marker gate

**Mechanically unchanged; slightly improved on the prose surface, slightly worsened by the contradiction above.**

- Both markers are still required; both SHAs are still compared against the forge-reported HEAD. No marker check was touched.
- "Set the active-reviewer marker" grants **no new capability**. I verified `warn-review-marker-write.sh` is advisory — `exit 0` always, per AgDR-0111. The active-reviewer marker suppresses a *warning*, never a block. A forger loses a stderr line they were already free to ignore, so hand-setting it does not unlock a forged approval.
- Spawning Rex directly is **not** self-review. Independence in this framework comes from the reviewer being a separate sub-agent with its own context, not from who typed the command (`pr-workflow.md`, #1042). Spawning the same sanctioned agent is equivalent in independence to the skill invoking it.
- The PR adds two new explicit "Do not write the approval marker yourself" statements at the two moments of maximum pressure — the refused merge and the fresh PR. That is a real reduction in forging likelihood, and it is the strongest security argument in this change.

The unsatisfiable-gate condition this PR removes was itself a forging pressure source: a gate that cannot be satisfied through any sanctioned path pushes the operator toward merging outside it or forging. Fixing that is a net gain for gate integrity.

### Verdict

**APPROVED**

No gate is relaxed beyond the two anchored branch names, no bypass was found across the probe battery, and no exit-code or expansion change was smuggled into either heredoc. The MEDIUM in section 4 is non-blocking per the framework's own severity policy, but it is a five-line text move in a production trust-chain hook and I would fix it before merge rather than after — it partially undoes AgDR-0056's two-audience fix in the one file whose contradiction history is #843.

My approval rests on the mechanical finding that the merge gate's marker requirement, SHA-vs-forge-HEAD comparison, protected-branch block, and ticket-first gate are all untouched and independently verified still blocking.

---
📌 Reviewed commit: `b0018d9b7fe68f55f35e86a8767ae64a532bc41f`

atlas-apex (commented):
## Code Review: PR #1173

**Commit**: `0e526e9080681ae8b6762a9a4c28a34e77193eba`
**Scope**: `Full`

### Summary

This PR fixes two live #1161 governance defects: `/handover`'s two hardcoded bootstrap branches now have exact-literal branch-gate exemptions, and the merge-gate / auto-review guidance now gives a sanctioned recovery when the harness bundled `/code-review` shadows ApexYard's skill. It also documents the trust-chain decision in AgDR-0129 and adds regression coverage for the branch exemption.

I focused especially on the incremental changes after the prior Rex/Security review at `b0018d9`: those changes move the shadowed-skill recovery into the orchestrator-only branch, explicitly exclude normal CHANGES REQUESTED reviews from that recovery path, and state that build-class sub-agents are not the audience for the direct-Rex exception.

### Checklist Results

- ✅ Code Quality: Pass
- ✅ Testing: Pass
- ✅ Security: Pass
- ✅ Performance: Pass
- ✅ Technical Decisions (AgDR): Pass — AgDR-0129 covers the trust-chain / branch-gate decisions
- ✅ Adopter Handbooks: Pass

### Issues Found

None.

### Suggestions

suggestion: The first commit subject is `fix: ...` rather than the usual `fix(#1161): ...` shape. The PR title/body and second commit carry the ticket reference, so I am not blocking this, but the general commit-message handbook prefers the ticket in the subject for future archaeology.

### Verification

- `bash .claude/hooks/tests/test_validate_branch_name_handover.sh` — 13/13 pass
- `bash -n .claude/hooks/auto-code-review.sh .claude/hooks/block-unreviewed-merge.sh .claude/hooks/validate-branch-name.sh .claude/hooks/tests/test_validate_branch_name_handover.sh` — pass
- `shellcheck -S warning ...` — pass where available in the review worktree

### Verdict

**APPROVED**

---
📌 Reviewed commit: `0e526e9080681ae8b6762a9a4c28a34e77193eba`

atlas-apex (commented):
## Design Review: PR #1173

**Commit**: `0e526e9080681ae8b6762a9a4c28a34e77193eba`

### Summary
PR #1173 records AgDR-0129 for two gate-satisfiability defects: `/handover` hardcoded branches that could not pass the branch-name gate, and a harness bundled `/code-review` shadowing failure that can leave the merge gate unsatisfiable. The chosen design keeps the mechanical gate surface narrow: two exact-literal branch exemptions, plus advisory recovery text that restores the sanctioned Rex path without hand-writing approval markers.

### Review Lens Results
- ✅ Quality attributes / NFRs: Pass — AgDR-0129 and the PR body state the safety target clearly: keep the merge gate as strong as before, make `/handover` complete, and avoid marker forgery pressure.
- ✅ Design patterns & structure: Pass — the implementation preserves the existing hook/rule/skill separation: `validate-branch-name.sh` owns branch validation, banners/rules own operator recovery guidance, and `/handover` documents its load-bearing branch names.
- ✅ Technical debt: Pass — the design explicitly calls the shadowed-skill recovery advisory because the harness name-resolution boundary is not observable from hooks; no silent mechanical TODO is being smuggled in.
- ✅ Decisions (AgDR linkage): Pass — AgDR-0129 covers both significant decisions: exact-literal `/handover` branch exemptions and banner-level shadowed-skill recovery.
- ✅ Risk: Pass — the blast radius is bounded by anchored branch names and by recovery criteria requiring both signs: background run plus no PR-posted review. It also distinguishes CHANGES REQUESTED from the shadowing case.
- ✅ Trade-off analysis: Pass — AgDR-0129 compares mechanical detection, skill rename, disabling bundled skills, exact-literal exemption, and ticket-carrying handover branch alternatives.
- ✅ Requirements traceability: Pass — each change maps to #1161: two live defects fixed, the third documented as already fixed on current `dev`.
- ✅ Migration safety: N/A — no data/schema migration is proposed.
- ✅ Adopter Handbooks: Pass — public/private clean-architecture layering is not implicated by hook/rule docs; blocking migration-safety handbook is N/A.

### Blocking Findings
None.

### Handbook Findings
None. Applied: `handbooks/architecture/clean-architecture-layers.md`, `/Users/ahmed/apex/apexyard-portfolio/custom-handbooks/architecture/clean-architecture-layers.md`, `handbooks/architecture/migration-safety.md`, `handbooks/general/commit-message-quality.md`.

### Suggestions
- Consider keeping the two identifying signs for bundled-skill shadowing in a short troubleshooting snippet if this failure recurs across adopter sessions; advisory only, not needed for this PR.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `0e526e9080681ae8b6762a9a4c28a34e77193eba`

atlas-apex (commented):
## Security Review: PR #1173

**Commit**: `0e526e9080681ae8b6762a9a4c28a34e77193eba`

### Summary
Reviewed the trust-chain changes in PR #1173, with special focus on the incremental delta since prior security review commit `b0018d9`.

Full PR scope:
- Adds an exact-literal branch-name exemption for `/handover`'s two prescribed bootstrap branches: `docs/agents-md` and `docs/apexyard-badge`.
- Adds regression coverage proving those exact names pass while near-miss `docs/*` names still fail.
- Updates merge-gate / auto-review guidance for the bundled `/code-review` shadowing failure mode.
- Documents the decision in AgDR-0129 and the `/handover` skill notes.

Incremental scope since `b0018d9`:
- `.claude/hooks/auto-code-review.sh`
- `.claude/hooks/block-unreviewed-merge.sh`
- `.claude/rules/pr-workflow.md`

Security assessment: the latest delta narrows the advisory recovery path rather than widening it. It now distinguishes the true shadowed-skill case (`Running in the background` + no posted PR review) from a legitimate `CHANGES REQUESTED` review, and it explicitly limits the direct Rex recovery to the orchestrator. The recovery still requires Rex to write its own approval marker on an approved verdict; it does not instruct anyone to write an approval marker by hand. The active-reviewer marker remains provenance-only and grants no gate-passing power.

The branch-name exemption is narrow: the regex is anchored to `^docs/(agents-md|apexyard-badge)$`, and the regression test covers suffix, prefix, path-segment, unrelated docs-branch, release/sync, and normal ticket-branch cases.

### Checklist Results
- Secrets & Credentials: Pass — no secrets, credentials, or sensitive tokens added.
- Injection Prevention: Pass — no new command execution path, shell interpolation path, parser expansion, SQL/NoSQL/LDAP/template surface, or user-controlled shell sink introduced by the incremental delta.
- XSS Prevention: Pass — no rendering path or browser-executed content introduced.
- Auth & Authorisation: Pass — no auth/session/resource-access changes; merge trust-chain semantics remain marker/SHA based and fail-closed.
- Data Protection: Pass — no PII, sensitive-data storage, logging, or URL exposure introduced.
- API Security: Pass — no API endpoint or CORS/rate-limit/error-surface change.
- Gate & Trust-Chain Integrity: Pass — the handover exemption is exact-literal and regression-pinned; the shadowed-skill recovery remains advisory, clarifies non-applicability to `CHANGES REQUESTED`, keeps approval-marker writes with Rex, and adds an orchestrator-only boundary.

### Security Issues Found
None.

### Verification
- `gh pr view 1173 --repo me2resh/apexyard --json title,body,files,additions,deletions,headRefOid,baseRefName,headRefName,url`
- `gh pr diff 1173 --repo me2resh/apexyard`
- `git diff b0018d9..upstream/pr-1173 -- .claude/hooks/auto-code-review.sh .claude/hooks/block-unreviewed-merge.sh .claude/rules/pr-workflow.md`
- `bash -n .claude/hooks/auto-code-review.sh .claude/hooks/block-unreviewed-merge.sh .claude/hooks/validate-branch-name.sh .claude/hooks/tests/test_validate_branch_name_handover.sh`
- `bash .claude/hooks/tests/test_validate_branch_name_handover.sh` — 13/13 pass

### Recommendations
- Keep the two identifying signs together in future edits; separating them would risk misclassifying a real `CHANGES REQUESTED` review as a harness-shadowing failure.
- Keep the branch exemption exact-literal and test-pinned; do not convert it to a broader `docs/*` exemption.

### Verdict
**APPROVED**

---
📌 Reviewed commit: `0e526e9080681ae8b6762a9a4c28a34e77193eba`

atlas-apex (approved):

Issue #1128, opened 2026-08-01
[Feature] QA and design-review roles should require browser evidence as a named deliverable


bitwhispererrr:
## Problem

The QA gate is one of the framework's mandatory stops — Gate 6, ticket → Done, described in `.claude/rules/workflow-gates.md` as "QA verified, signed off". But `roles/engineering/qa-engineer.md` and `.claude/agents/qa-engineer.md` specify *what* to verify (acceptance criteria) without ever specifying *what counts as evidence*. The same gap exists on the design-review side for the UI Designer.

The practical consequence is that a QA sub-agent can satisfy the entire gate from database queries and source reading, never open the product, and return a record that reads as full verification. That is strictly worse than no QA at all, because a record claiming verification is exactly the artifact that stops a human from re-checking.

This is not hypothetical. On a managed project under apexyard governance, a QA agent's report had to be **rejected twice** for this. Its header claimed the work was "verified against running systems (backend + admin UI)" while its own method section showed the checks were database queries and code inspection — the UI was never booted, despite the brief asking for it specifically. An earlier ticket on the same project had the same outcome: a code-inspection-and-test-run pass on work whose whole risk surface was rendered output.

Three defects on that project were found only by a human looking at the screen, and all three were invisible to the checks the QA agent actually ran:

- A table column rendered as an em-dash for every row, because the seeded field was an empty string. A database query returns `''` and nothing looks wrong.
- A seeded email template that was 56 bytes of bare markup, sitting next to two ~3.7KB siblings. Every test passed. It would have presented as one naked line beside two polished ones.
- A near-miss in the other direction: an ad-hoc headless-browser screenshot appeared to show a chart rendering without lines. It was the charting library's draw-in animation captured at frame 0, and the wrong conclusion reached a permanent record before being corrected.

That last one matters for the proposed fix: ad-hoc headless-browser invocation is itself a source of false positives, so "open a browser somehow" is not sufficient guidance.

## Proposed behaviour

Make browser evidence a **named deliverable** rather than an implied one, in three places:

1. **`roles/engineering/qa-engineer.md` + `.claude/agents/qa-engineer.md`** — add to the QA handoff artifact a required per-item evidence breakdown that states, for each acceptance criterion, whether it was browser-verified or not. Critically, the *not verified* list must be mandatory rather than optional: an agent that could not boot the surface must say so explicitly and name the affected items. An honest "items 1–3 were database checks, not browser checks" is a good QA record; silence on the question is not.

[...]

PR #1191, opened 2026-09-06
feat(#1128): require browser evidence in QA and design-review records

me2resh:
## Summary

- **QA sign-off now names its evidence per criterion** — `roles/engineering/qa-engineer.md` gains a "Browser Evidence (rendered surfaces only)" section and a sign-off table with an explicit `Browser-verified` column. Today a QA agent can satisfy Gate 6 entirely from database queries and source reading, and return a record that reads as full verification. That record is worse than no QA at all, because it stops a human from re-checking.
- **The not-verified list is mandatory, not optional** — an agent that could not boot the surface must say so and name the affected criteria. An honest "AC1–AC3 were database checks" is a good QA record; silence on the question is the failure this ticket documents. The framing sentence carried through every file is: **reject a PASS whose evidence does not match the criterion.**
- **Design review gets the same requirement** — `roles/design/ui-designer.md` and its agent wrapper now require the component to be rendered before review. It matters more here than in QA: a design review that never renders is only reviewing source, so spacing, contrast, overflow, and empty states are invisible to it.
- **Mechanism guidance names the trap that produced a wrong finding** — all four files recommend a browser-automation MCP server (Playwright MCP or equivalent) over ad-hoc headless-browser CLI invocation, prefer accessibility-tree snapshots over screenshots for asserting what a page says, and warn that a screenshot captured at frame 0 of a draw-in animation renders empty. Ticket #1128 records that exact false positive reaching a permanent record before correction.
- **Both agent wrappers state the tool reality plainly** — neither `qa-engineer` nor `ui-designer` has browser tooling in its `allowed-tools`. Rather than leave that as a silent contradiction, each wrapper instructs the agent not to improvise with a headless CLI and instead to report the affected criteria as not browser-verified and hand the gap back to the orchestrator.
- **Gate 6 and `/debug` each get one paragraph** — `workflow-gates.md` § "What counts as QA evidence" states the requirement where orchestrators read it, and says plainly that no hook enforces it. The `/debug` web appendix gets the same mechanism warning, since hypothesis-driven UI debugging has the identical evidence problem.

Deliberately **not** included: any hook, matcher, or `settings.json` entry. A `PreToolUse` hook cannot see whether a sub-agent opened a browser or inspect a report's contents — the same limitation AgDR-0056 documents for the spawn boundary. This is a role-definition change backed by orchestrator discipline. The requirement is also conditional on the ticket touching a rendered surface; requiring it on backend-only tickets would train agents to write "N/A" without thinking, which is the same substitution failure the ticket is trying to fix.

**AgDR decision: none written.** Read against `.claude/rules/agdr-decisions.md`, this misses every threshold row — no new dependency (Playwright MCP is a recommendation, not something added to a manifest), no new service, no data-model change, no security control, no CI/CD design, and no repo-wide pattern (two roles plus two cross-references). It adds no gate mechanism and changes no hook, so rail 1 does not apply, and it is fully reversible inside this PR. The practical test also comes out negative: the reasoning lives in the role files themselves and in #1128, so a teammate reading either in six months is not confused and repeats no expensive work. Ambiguity here is about *cross-cutting-ness*, and the honest reading is that a prose clarification of what an existing gate's output must contain is not the same as adopting a new pattern across the repo.

## Testing

1. `npx --yes markdownlint-cli --config .markdownlint.json` on all six changed files — exit 0, no output.
2. `bash bin/run-hook-tests.sh` — `PASS=145 FAIL=2 SKIP=0 TOTAL=147`. The two failures (`test_lib_self_location_cwd_anchor.sh`, `test_validate_search_config.sh`) reproduce identically on the clean base commit `a300c98` with the branch stashed, so they are pre-existing and unrelated to this diff.
3. `git diff --check` — exit 0, no whitespace errors.
4. Behavioural check is by reading: the QA sign-off template, the UI Designer review section, the Gate 6 paragraph, and the `/debug` appendix each state the conditional trigger (rendered surface), the mandatory not-verified list, and the mechanism recommendation.

Refs #1128

## Glossary

| Term | Definition |
|------|------------|
| Browser evidence | Verification performed against the running product in a real browser — what appeared on screen, at which URL, against which data — as opposed to inference from source code or database state. |
| Accessibility-tree snapshot | A structured text representation of a rendered page's semantic content. More reliable than a screenshot for asserting what a page says, and immune to animation timing. |
| Rendered surface | Any part of a ticket whose acceptance criteria describe what a person sees — a page, component, table row, chart, or email body. The conditional trigger for this requirement. |
| Gate 6 | The framework's mandatory QA stop: a merged PR moves a ticket to QA, not Done, and QA verification is required before it closes. |
| Self-concealing failure | A defect whose own output suppresses further investigation — here, a QA record asserting verification that did not occur, which stops anyone else from checking. |

### DIFF
--- a/.claude/agents/qa-engineer.md
+++ b/.claude/agents/qa-engineer.md
@@ -12,6 +12,18 @@ Read and adopt `@roles/engineering/qa-engineer.md` for full identity, responsibi
 
 The QA Engineer is read-only by mechanical contract: this agent ships **without** Edit/Write tools because QA's job is to verify acceptance criteria, file bug tickets, and sign off — not to ship code. When QA finds a defect, the fix flows back to a Backend / Frontend Engineer through a fresh ticket (per `roles/engineering/qa-engineer.md` § "If QA Finds Issues" and `workflows/sdlc.md` § "Phase 5: QA Verification").
 
+## Browser evidence is a named deliverable
+
+If the ticket touches a rendered surface, verify each rendered acceptance criterion in a browser. Do not substitute a database query, a source read, or a passing test. **Reject a PASS whose evidence does not match the criterion.**
+
+Report the browser-verification status of every acceptance criterion. The not-verified list is **mandatory**: if you cannot boot the surface, say so and name every affected criterion. Never return a report that leaves the question unanswered — a record claiming verification that did not happen is worse than no QA record, because it stops a human from re-checking.
+
+Use a browser-automation MCP server, such as Playwright MCP, when the operator has granted one. This wrapper's `allowed-tools` list does not include browser tooling, so if no browser MCP server is available to you, do not improvise with a headless-browser CLI: report the affected criteria as not browser-verified and hand the gap back to the orchestrator.
+
+Prefer an accessibility-tree snapshot over a screenshot when asserting what a page says. If you take a screenshot, wait until the page settles — a chart or transition captured at frame 0 renders empty and produces a confident, wrong finding.
+
+Full requirement and the sign-off table: `@roles/engineering/qa-engineer.md` § "Browser Evidence (rendered surfaces only)".
+
 ## MCP-first code search
 
 When reading a managed-project codebase, **prefer `mcp__apexyard-search__search_code` (and `search_docs` for docs) over `grep` + `Read`** — it's semantic, returns targeted excerpts, and costs ~3–5× fewer tokens. Fall back to `grep`/`Read` only when an MCP query returns nothing relevant (e.g. the project isn't indexed). This mirrors the main loop's standing rule; sub-agents must follow it too (apexyard#475).
--- a/.claude/agents/ui-designer.md
+++ b/.claude/agents/ui-designer.md
@@ -26,6 +26,18 @@ You are a build-class sub-agent. You cannot nest the Agent tool, so you cannot s
 
 **DO:** Report your build results plainly — what you designed, what deliverables you produced, what acceptance criteria you verified. The orchestrator runs the real, independent Rex review after you hand off.
 
+## Browser evidence is a named deliverable
+
+Render the component before you review it. A review that only reads source cannot see spacing, contrast, overflow, empty states, or loading states. **Reject a PASS whose evidence does not match the criterion.**
+
+Report the browser-verification status of every design criterion. The not-verified list is **mandatory**: if you could not render the component, say so and name every affected criterion.
+
+Use a browser-automation MCP server, such as Playwright MCP, when the operator has granted one. This wrapper's `allowed-tools` list does not include browser tooling, so if no browser MCP server is available to you, do not improvise with a headless-browser CLI: report the affected criteria as not browser-verified.
+
+Prefer an accessibility-tree snapshot over a screenshot when asserting what a component says. If you take a screenshot, wait until the page settles — a transition captured at frame 0 produces a confident, wrong finding.
+
+Full requirement: `@roles/design/ui-designer.md` § "Browser evidence is a named deliverable". This applies to UI work only.
+
 ## Design tooling (on demand)
 
 - Sync the design system to **claude.ai/design** via the **`/design-sync`** skill (built-in `DesignSync` tool; authorize via claude.ai login / `/design-login`, not `.mcp.json`). Incremental, one component at a time.
--- a/.claude/rules/workflow-gates.md
+++ b/.claude/rules/workflow-gates.md
@@ -142,6 +142,12 @@ In Progress → In Review → QA → Done
                     QA must verify
 ```
 
+### What counts as QA evidence
+
+If the ticket touches a rendered surface, the QA record must state, per acceptance criterion, whether it was verified in a browser. The not-verified list is mandatory: a QA agent that could not boot the surface must name the affected criteria rather than stay silent. A database query, a source read, or a green test suite is not evidence that a rendered criterion is met. **Reject a PASS whose evidence does not match the criterion.**
+
+A ticket with no rendered surface needs no browser evidence. Full requirement and sign-off shape: `roles/engineering/qa-engineer.md` § "Browser Evidence (rendered surfaces only)"; the design-gate equivalent is `roles/design/ui-designer.md` § "Browser evidence is a named deliverable". No hook enforces this — a `PreToolUse` hook cannot see whether a sub-agent opened a browser, so the orchestrator must reject a report that skips the breakdown.
+
 ---
 
 *Part of [ApexYard](https://github.com/me2resh/apexyard) — multi-project SDLC framework for Claude Code · MIT.*
--- a/.claude/skills/debug/SKILL.md
+++ b/.claude/skills/debug/SKILL.md
@@ -187,6 +187,8 @@ Each appendix carries the stack-specific surface-evidence requirements (step 1),
 - Browser DevTools Console — any errors, including the stack trace
 - The exact URL in the address bar at the moment of failure (often differs from what the user typed — note both)
 
+**Capture that evidence with a browser-automation MCP server** (Playwright MCP or equivalent) rather than an ad-hoc headless-browser CLI invocation. Prefer an accessibility-tree snapshot over a screenshot when the hypothesis is about what the page *says* — a snapshot returns rendered text and roles, and it does not depend on animation timing. If you take a screenshot, wait until the page settles: a chart's draw-in animation captured at frame 0 renders empty, and that frame has already produced at least one confident, wrong root-cause finding.
+
 **Architecture-surface map (step 2).** Read the file at every layer the request touches:
 
 | Layer | What to read | What you're looking for |
--- a/roles/design/ui-designer.md
+++ b/roles/design/ui-designer.md
@@ -79,6 +79,18 @@ System-level standards, cross-product visual direction, and design disagreements
 
 Part of that review is **design-QA of AI/agent-generated UI** — a current baseline duty: confirm the generated code used the right components and design tokens rather than one-off magic values.
 
+### Browser evidence is a named deliverable
+
+Render the component before you review it. A design review that only reads source is reviewing source, not design — spacing, contrast, overflow, empty states, and loading states are all visible only once the component renders with real data.
+
+**Reject a PASS whose evidence does not match the criterion.** If a criterion describes what a person sees, the evidence must be what appeared on screen, at which URL, against which data.
+
+State the browser-verification status of every design criterion you reviewed. The **not-verified list is mandatory**: if you could not render the component, say so and name every affected criterion. Silence on the question reads as verification that did not happen.
+
+**Mechanism.** Use a browser-automation MCP server, such as Playwright MCP or an equivalent, over an ad-hoc headless-browser CLI invocation. Prefer an accessibility-tree snapshot over a screenshot when you assert what a component says, because a snapshot returns rendered text and roles and does not depend on animation timing. If you take a screenshot, wait until the page settles — a transition or draw-in animation captured at frame 0 produces a confident, wrong finding.
+
+This applies to UI work only. A PR with no rendered surface needs no browser evidence.
+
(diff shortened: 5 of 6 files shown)


me2resh (commented):
## Code Review: PR #1191

**Commit**: `3e112a08e52705d48502da870153c85425930c83`
**Scope**: Full

**APPROVED.** No action required before merge. Three suggestions below are follow-up material, not merge blockers.

The change closes the gap #1128 describes, and it closes it structurally rather than by adding a field. The QA sign-off template now carries a per-criterion `Browser-verified` column, a `Rendered surface: Yes / No` header field, and a delete rule that permits removing the not-verified section only when nothing is unverified. An agent filling this template cannot stay silent on the question without leaving a visibly empty table cell, which `writing-standard.md` Rule 2 already forbids in a filed artefact.

---

### Summary

Six markdown files add a browser-evidence requirement to the QA gate (Gate 6) and the per-PR design gate: the two role definitions, their two agent wrappers, the Gate 6 rule paragraph, and the `/debug` web appendix. No hook, no matcher, no `settings.json` entry — enforcement is orchestrator discipline, and the PR says so plainly.

### Checklist Results

- Architecture & Design: N/A (no code)
- Code Quality: N/A (no code)
- Testing: Pass — markdownlint clean; the two suite failures reproduce on base `a300c98` and are unrelated
- Security: Pass — no trust-chain path touched (verified per-file: no `.claude/hooks/*.sh`, no `.claude/settings.json`, no auth/crypto/secrets/migration)
- Performance: N/A
- Summary Bullet Narrative: Pass — all five bullets carry what-changed and why-it-matters
- Technical Decisions (AgDR): N/A — agree with the author, reasoning below
- Adopter Handbooks: Pass with one advisory nit

---

### The four judgments you asked for

**1. Does it force the not-verified list, or add a fillable field?** It forces it. Four mechanisms compound:

| Mechanism | Where | What it prevents |
|---|---|---|
| `Rendered surface: Yes / No` header | sign-off template | Skipping the conditionality decision silently |
| `Browser-verified` column, per row | AC table | Answering at ticket level instead of per criterion |
| `Not Browser-Verified` section present by default | sign-off template | Omission by never adding it |
| "Never delete it while an unverified rendered criterion remains" | after the template | Deleting it to make the record look clean |

The strongest single choice is one the PR body understates: **the template's example row models the honest-gap case.** `AC2 | PASS | [query or test that proves it] | No`, with a matching Not-Browser-Verified entry. An agent copying the template's shape copies a record that already admits a gap. Since the failure being fixed is silence, a template whose default state is not-silent is doing more work than the prose around it. The Quality Gates line reinforces it with the right disjunction — `browser-verified, or listed as not verified` — which makes the honest-gap path a *passing* state rather than something to avoid reporting.

**2. Conditionality.** Correctly scoped in all five passages, each stating the negative case explicitly. The QA role file goes furthest and is the one that matters, because the sign-off template is the only place an "N/A" could physically land: *"If the ticket has no rendered surface, browser evidence is not required. Do not add an empty browser-evidence section to a backend-only sign-off."* That is the ticket's stated worry answered directly, and the delete rule gives deletion as the alternative to "N/A" — consistent with `writing-standard.md` Rule 2.

**3. Strict mode.** Mixed, and passing on the clause that matters most.

Holds: modality is clean. I checked every added line — no `should` or `may` in any instruction. The only near-misses are `could not render` / `could not boot` (past-tense descriptive) and `can render empty at frame 0` (`can` for possibility, which is Rule 7's allowed use). That is the hardest clause in Rule 3 to follow and it is followed. Imperative voice is consistent, and the vocabulary is stable across all four files — `rendered surface`, `browser-verified`, `not-verified list`, `accessibility-tree snapshot` each mean one thing throughout.

Deviates: sentence length and one-instruction-per-sentence. See the nit below. My read is that these are style-of-compliance misses rather than comprehension failures — Rule 3 exists so an agent has "no room for interpretation", and the long sentences are long but unambiguous. The role files also sit closer to Flavored than Strict (a human reads `roles/*.md`), which is where the rationale prose about em-dashes and 56-byte templates legitimately belongs. Only the two `.claude/agents/*.md` wrappers are purely machine-consumed, and that is where I would tighten.

**4. The `allowed-tools` gap — my independent read.** The author's resolution is honest and workable, and the requirement is **not vacuous in the default configuration**. The load-bearing fact is one the PR body does not state: **both wrappers already list `Bash` in `allowed-tools`** (`qa-engineer`: `Bash, Read, Grep, Glob, mcp__apexyard-search__*`; `ui-designer`: `Bash, Read, Edit, Write, Grep, Glob`).

So "do not improvise with a headless-browser CLI" is not a no-op restating an incapacity the agent already has. It is a prohibition on a capability the agent genuinely possesses — both agents can shell out to `npx playwright screenshot` or a headless Chrome invocation today. And that is precisely the path that produced #1128's third documented defect, the chart captured at frame 0. The wrapper text closes the exact failure the ticket recorded, on the exact tool the agent holds.

That leaves the default-config outcome as: *the agent reports the rendered criteria as not-browser-verified, and does not fake it with a CLI screenshot.* Measured against the ticket, that is the win, not a consolation. #1128's load-bearing ask is the mandatory not-verified list, not that verification always happens — its own words are that an honest "items 1–3 were database checks" is a good QA record and silence is not. The default config produces exactly that record, with no new tooling. The capability half is correctly left to the adopter who grants an MCP server.

I also agree that hardcoding a server name into `allowed-tools` would be worse. It would either break the wrapper for adopters without that server or bake one vendor into a framework default. Naming Playwright MCP in prose as an example, while leaving the allowlist alone, is the right call.

One real gap remains, and it is documentation rather than design — see suggestion 2.

**5. AgDR.** Agree with the author. Walking the threshold table: no dependency enters a manifest (Playwright MCP is named in prose and nothing calls it), no new service or integration, no data model, no security control, no CI/CD or infrastructure design. Rail 1 does not fire — I verified per-file that nothing under `.claude/hooks/` or `.claude/settings.json` is touched. Nothing is hard to reverse; this is one revert of six markdown hunks.

The cross-cutting row is the arguable one, and the author names it as arguable rather than skipping it. My reading agrees with theirs: adding a requirement to two existing role definitions plus two cross-references is scoped, not a pattern the rest of the repo must now follow. It changes no other role, introduces no mechanism, and creates no convention other files inherit. Compare that row's own rationale — "reversing it later means touching everything at once" — which does not describe this. The practical test also comes out negative, since the reasoning lives in the role files and in #1128.

Rail 2 (ambiguity rounds up) is the only counterargument with weight, and engaging it explicitly in the PR body is the right shape for a no-AgDR call. No AgDR needed.

**6. PR hygiene.** Title matches `type(#N): description`. `Refs #1128` is correct — Gate 6 keeps the issue open until QA verifies, and `Closes` would auto-close past that stop. Glossary present with five terms, all genuinely used. Summary bullets are narrative throughout; the advisory label-only heuristic fires on none of them. The Testing section is honest in the way `evidence-grounding.md` asks for: it names the base commit the pre-existing failures reproduce on, and it says plainly that the behavioural check is by reading rather than claiming a verification it did not run.

---

### Issues Found

None blocking.

---

### Handbook Findings

Loaded by path convention: `handbooks/general/commit-message-quality.md` (advisory). Not applicable, recorded rather than reported: `handbooks/architecture/*` (no architecture-bearing code, design artifact, or migration path in the diff), `handbooks/language/typescript/strict-mode.md` (no `.ts`/`.tsx`), and the private `custom-handbooks/architecture/clean-architecture-layers.md` (same reason as the public architecture bucket). Semantic supplement ran and returned no handbook-path matches above the floor.

`nit:` **Commit-message quality** — `handbooks/general/commit-message-quality.md`

- Commit `3e112a08` has subject `feat: require browser evidence in QA and design-review records`. The handbook's subject-line row asks for `type(#TICKET): description`, so the `(#1128)` scope is missing. Worth flagging only lightly: `.claude/rules/git-conventions.md` § "Commit Message Format" lists bare `type: subject` as a valid commit form (the scope is mandatory for PR *titles*, not commits), so the commit is compliant with the framework rule and diverges only from the stricter advisory handbook. `Refs #1128` in the body carries the linkage regardless. None of the handbook's four "What Rex flags" triggers fire — the subject is substantial, the body is present and is not a restatement, and there is no placeholder text.

---

### Suggestions

**suggestion: the `Rendered surface: No` field is a one-word exit from the whole requirement.** An agent that mis-classifies a rendered ticket as `No` skips the table column and the not-verified section in a single field, and the record still reads as complete. The role file's definition of a rendered surface is good and checkable, and the Gate 6 paragraph already tells the orchestrator to reject a report that skips the breakdown — but nothing tells the orchestrator to re-read the acceptance criteria when the answer is `No`. One sentence in the Gate 6 paragraph would close it. Not a blocker: this is the same self-attestation limit every conditional gate in the framework carries, and the PR is explicit that no hook enforces any of it.

**suggestion: say how an operator grants browser tooling.** Both wrappers say "when the operator has granted one" and "if no browser MCP server is available to you" without naming the mechanism. `allowed-tools` is an allowlist, so an adopter who installs Playwright MCP and does not also add `mcp__playwright__*` to the wrapper still gets the not-verified report, with nothing pointing at why. One clause per wrapper fixes it. While there, consider stating the `Bash` fact explicitly — "you have `Bash` and could invoke a headless browser; do not" is stronger than the current phrasing, and it mirrors the framing `pr-workflow.md` § "Build agents cannot self-review" already uses to good effect ("nothing mechanically stops you, which is exactly why the instruction matters").

**suggestion: `roles/design/head-of-design.md` is the same gate and did not get the requirement.** Its § "Design Review Process" lists "Test interactions and states" and "Assess visual consistency" — criteria that are invisible without rendering — and per AgDR-0106 the Head of Design is the escalation path for the design gate, including the "no UI Designer available" case where they become the routine reviewer. So the silent-substitution failure this PR fixes still has an open door one role over. Out of #1128's declared scope and genuinely rarer, so not a blocker; a one-line cross-reference to `roles/design/ui-designer.md` § "Browser evidence is a named deliverable" would close it.

**nit: two Strict-mode tightenings in the agent wrappers**, which are the most machine-consumed text in the diff:

- Both wrappers carry a ~44-word sentence holding a declarative, a condition, and two instructions: *"This wrapper's `allowed-tools` list does not include browser tooling, so if no browser MCP server is available to you, do not improvise with a headless-browser CLI: report the affected criteria as not browser-verified and hand the gap back to the orchestrator."* Rule 1 (one instruction per sentence), Rule 2 (twenty words), and Rule 6 (condition before instruction) all point at splitting this into three sentences.
- `Use a browser-automation MCP server, such as Playwright MCP, when the operator has granted one` trails its condition; Rule 6 wants it first.

**nit: one asymmetry between the two wrappers.** `.claude/agents/qa-engineer.md` ends the fallback with "…and hand the gap back to the orchestrator". `.claude/agents/ui-designer.md` omits the hand-back. Both wrappers face the identical tool gap, so the same terminal instruction belongs in both. The UI wrapper's existing "DO: Report your build results plainly… The orchestrator runs the real, independent Rex review after you hand off" makes it implicit, which is why this is a nit and not a finding.

One asymmetry I checked and am **not** raising against this PR: the QA path got a structural forcing function (table, mandatory section, delete rule) while the UI path got prose only. I looked for somewhere to put a UI equivalent and there isn't one inside scope — `/approve-design` writes a marker, not a record, and per `roles/design/ui-designer.md` line 76 no agent writes the design marker at all; a human runs the skill. The UI Designer's output is an in-thread verdict with no artefact shape to add a column to. Prose in the role file is the only available home, so the author's choice is correct given the gate's existing design.

---

### Verdict

**APPROVED**

The words will change agent behaviour, which is the bar for a diff like this one. The QA template does the heavy lifting; the wrappers close the CLI-improvisation path that #1128 documented producing a wrong finding; and the requirement degrades to an honest not-verified record rather than to silence when no browser tooling is granted. That degradation is the ticket's stated success condition, not a fallback.

---
📌 Reviewed commit: `3e112a08e52705d48502da870153c85425930c83`

atlas-apex (approved):

Issue #1136, opened 2026-08-02
[Bug] /approve-merge on a release PR drops the Released-From trailer (bare squash + COMMIT_MESSAGES buries it)


me2resh:
## Summary

`/approve-merge <release-pr>` drops the `Released-From` trailer from the squash commit, because it merges via a bare `tracker_pr_merge ... squash` (`gh pr merge --squash` with no `--subject`/`--body-file`). This directly contradicts `/release` **Rule 11**, which requires release PRs to merge with an explicit `--subject`/`--body-file` *precisely* to preserve the trailer. The two skills are in conflict for release PRs — and `/approve-merge` is now the mandated human-only merge path (#1042), so the conflict fires on every release cut.

## Given / When / Then

- **Given** the repo setting `squash_merge_commit_message = COMMIT_MESSAGES` (confirmed via `gh api repos/me2resh/apexyard --jq '.squash_merge_commit_message'`), and a release PR whose head branch (`release/vX.Y.Z`) is cut from `dev` — so from `main`'s perspective the PR "contains" the entire dev↔main divergence (currently ~463 commits),
- **When** the release PR is merged via `/approve-merge <pr>` (which calls `tracker_pr_merge "<repo>" "<pr>" "squash" true` → bare `gh pr merge --squash`),
- **Then** GitHub builds the squash body by concatenating **all ~463 commit messages** (~445 KB). The single `chore: release vX.Y.Z` commit's `Released-From: <sha>` trailer ends up buried mid-body, **not** the final paragraph, so `git log -1 --pretty=format:'%(trailers:key=Released-From,valueonly)'` returns **empty**. `/release` Rule 12's post-merge trailer check fails.

## Observed

Live on **v5.4.0** (PR #1134, squash commit `d8a50ff`). The release itself shipped correctly (tag on the right commit, GitHub Release published, `CHANGELOG.md` correct), but the trailer is buried. Verified: the string `Released-From: d06cc1d` IS present once in the 445 KB body, just not as the final paragraph, so `%(trailers)` can't see it.

## Impact

- The **next** release's changelog range (`bin/release-changelog.sh` reading `PREV_TAG`'s trailer) silently degrades to the **#737 sync-boundary heuristic** instead of the exact `Released-From` anchor (AgDR-0094). Graceful — the heuristic is what pre-AgDR-0094 releases used, and `/release-sync` establishes the boundary it needs — but it re-opens the failure modes AgDR-0094 closed (#737, #872).
[...]

Issue #1152, opened 2026-08-25
[Bug] The fail-closed principle from #965/#1091 was applied only to the 4 merge gates — 19 other blocking hooks still exit 0 when they cannot evaluate their precondition


sudanese:
## Affected

The **pattern**, not one file. Verified against `upstream/main` at **v5.4.0** (`d8a50ff`), not a fork.

## Summary

The fail-closed principle — *"a gate that cannot evaluate its precondition must BLOCK, not guess"* — has now been applied three times, and each time only to the **same four merge gates**:

| Fix | Precondition that could not be evaluated | Scope of the fix |
|---|---|---|
| #965 / #973 | `jq` unavailable or `.tool_input.command` unparseable | the 4 merge gates |
| #1091 / #1098 | forge HEAD lookup failed | the 4 merge gates |
| #1151 | repo could not be resolved / diff came back empty | the 4 merge gates (reported, open) |

`block-unreviewed-merge.sh` states the principle in its own #1091 block-message, citing #965 and AgDR-0104. That reasoning is correct and general. It has never been generalised.

**19 other blocking hooks still exit 0 when they cannot evaluate their precondition.** On v5.4.0, exactly 4 hooks carry the `_normalize_json_escapes` raw-payload rescan; every other hook that parses `.tool_input` and can `exit 2` treats an unparseable input as *nothing to check*.

## Evidence — `jq` unavailable silently disables gates

`jq` is the sole parser for every hook's stdin. When it is absent the parse yields empty, and 19 hooks read that as "no command, nothing to do":

```sh
COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)
if [ -z "$COMMAND" ]; then
  exit 0          # <-- "could not read" treated as "nothing to check"
fi
```

Run against **upstream v5.4.0's own hook tree**, exported with `git archive upstream/main`. Each row is the same hook, same payload, differing only in whether `jq` resolves on `PATH`:

| Hook | What the payload does | jq present | jq unavailable |
|---|---|---|---|
| `check-secrets.sh` | commit with `api_key = "…"` staged | **exit 2 BLOCKED** | **exit 0 — allowed, no scan** |
| `block-main-push.sh` | push to the protected branch | **exit 2 BLOCKED** | **exit 0 — allowed** |
| `block-private-refs-in-public-repos.sh` | private project name → PUBLIC repo | **exit 2 BLOCKED** | **exit 0 — allowed, leak ships** |
[...]

PR #123, opened 2026-04-24
chore(#112): require-agdr-for-arch-pr.sh PreToolUse hook

atlas-apex:
## Summary

Adds a new PreToolUse hook `require-agdr-for-arch-pr.sh` that blocks `gh pr create` when the PR's diff touches architecture paths OR adds a new dependency, AND the PR body does not link an `AgDR-\d+-[a-z0-9-]+` reference.

Closes the gap where `.claude/rules/agdr-decisions.md` calls `/decide` a HARD STOP before architectural choices, but nothing mechanically enforced it at PR time. The Code Reviewer agent would flag missing AgDRs post-hoc, which is late — this hook fires *before* the PR is filed and gives the author a chance to run `/decide` first.

- New hook: `.claude/hooks/require-agdr-for-arch-pr.sh` (~280 LOC)
- Tests: `.claude/hooks/tests/test_require_agdr_for_arch_pr.sh` — 7 cases, all pass
- Config integration: reads `.agdr_trigger_paths[]` and `.agdr_trigger_dep_files[]` via the shared reader from #109
- Defaults: extended `.claude/project-config.defaults.json` with both arrays at sensible defaults (hexagonal domain/infrastructure, Terraform, migrations, CI workflows for paths; package.json, pyproject.toml, Cargo.toml, go.mod, Gemfile for deps)
- Wired into `.claude/settings.json` — PreToolUse matcher on `Bash(gh pr create *)`
- New row in `docs/rule-audit.md`, mechanized count bumped to 29

## How it works

1. Matches `gh pr create` only. Silent pass on everything else.
2. Parses `--title` / `--body` / `--body-file` / `-F <path>` (same extractor style as sibling hooks).
3. Resolves the PR base branch from `--base <arg>` with a sensible fallback chain: `upstream/dev` → `origin/dev` → `upstream/main` → `origin/main` → `main` → `master`. Stops at first resolvable.
4. Computes diff: `git diff $(git merge-base HEAD <base>)..HEAD --name-only`.
5. For each changed file, checks against `agdr_trigger_paths[]` globs (hexagonal layers, infra, migrations, CI).
6. For dep files, uses `jq` diff of `.dependencies + .devDependencies + .peerDependencies + .optionalDependencies` key sets between MERGE_BASE and HEAD to detect **additions** (not version bumps). For non-JSON dep files, uses a simpler `+` vs `-` line count heuristic (documented inline).
[...]

PR #1192, opened 2026-09-06
fix(#1136): preserve Released-From trailer on release-PR squash

me2resh:
## Summary

- **`tracker_pr_merge` gains optional `<subject>`/`<body_file>` params (gh kind only)** — when both are non-empty, the merge passes `--subject`/`--body-file` to `gh pr merge`; every existing call keeps its current behaviour because the params default to empty
- **`/approve-merge` now detects release-class PRs** (head branch `release/vX.Y.Z`, or title starting `release(`) the same way it already detects sync-class PRs, and passes the PR's own title/body as subject/body_file — so the squash commit carries the reviewed body instead of GitHub's `COMMIT_MESSAGES`-assembled default, which was burying the `Released-From` trailer mid-body
- **Fails closed, not open** — an unreadable or empty PR body (or `body_file`) stops the merge rather than silently falling back to a bare squash, at two layers: `/approve-merge`'s own read check, and a second guard inside `tracker_pr_merge` itself
- **`/release` Rule 11 and `/approve-merge` step 6 now cross-reference each other** so the two skills stop describing contradictory merge behaviour for the same PR class
- **AgDR-0132** records the decision, mirroring the sync-PR precedent in AgDR-0053

## Testing

- `.claude/hooks/tests/test_tracker_pr_merge.sh` — 5 new cases: release-PR subject+body_file → `--subject`/`--body-file` appended; the existing 5-arg call shape (no subject/body_file) unchanged; explicit empty-string subject/body_file unchanged; missing `body_file` refuses the merge (rc=1, no CLI invoked); empty (zero-byte) `body_file` refuses the merge the same way
- `bash bin/run-hook-tests.sh` — 145 PASS / 2 FAIL, both pre-existing failures on clean `upstream/dev` unrelated to this change (`test_lib_self_location_cwd_anchor.sh`, `test_validate_search_config.sh`) — no new failures
- `npx markdownlint-cli --config .markdownlint.json` on the three changed/added `.md` files — clean
- `git diff --check` — clean
- `shellcheck` on `.claude/hooks/_lib-tracker.sh` and `.claude/hooks/tests/test_tracker_pr_merge.sh` — clean

Refs #1136

AgDR: [docs/agdr/AgDR-0132-approve-merge-release-pr-body-trailer.md](https://github.com/me2resh/apexyard/blob/dev/docs/agdr/AgDR-0132-approve-merge-release-pr-body-trailer.md)

## Glossary

| Term | Definition |
|------|------------|
| `Released-From` trailer | A git trailer recording the exact `dev` SHA a release was cut from; `bin/release-changelog.sh` reads it to anchor the next release's changelog range (AgDR-0094) |
| `squash_merge_commit_message=COMMIT_MESSAGES` | Repo setting: GitHub builds a squash commit's body by concatenating the PR branch's commit messages — for a release PR that is the entire dev↔main divergence |
| Release-class PR | A PR whose head branch matches `release/vX.Y.Z`, or whose title starts with `release(` — the `/release` skill's canonical shape |
| Sync-class PR | A PR whose head branch matches `sync/main-to-dev-after-*`, or whose title starts with `sync(` — the `/release-sync` skill's canonical shape, and the existing precedent this fix mirrors (AgDR-0053) |


atlas-apex (approved):

me2resh (commented):
## Security Review: PR #1192

**Commit**: `e1842bd3dfc65f9a80ebf3da038a645daebeedd6`

### Verdict

**APPROVED — safe to merge.** No Critical or High findings. The three properties this change had to get right — argv separation for attacker-controlled text, no widening of the gated merge surface, and a genuine fail-closed refusal — all hold under adversarial testing. The findings below are two MEDIUM hardening items and several LOW/INFO notes; none blocks the merge, and none is a live vulnerability.

Reviewed as a **trust-chain change** (auto-fire trigger #777: the diff touches a production `.claude/hooks/*.sh` file on the merge-approval path). I independently confirmed the diff touches no merge-gate hook, no `settings.json`, and no `_lib-extract-pr.sh` — only `_lib-tracker.sh` among production hooks.

### Summary

`tracker_pr_merge` and `_tracker_merge_gh` gain two optional trailing parameters, `<subject>` and `<body_file>`, appended as `--subject`/`--body-file` on a `gh`-kind merge when both are non-empty. `/approve-merge` step 6 supplies them for release-class PRs. On a fork PR both values originate in attacker-controlled text — the PR title and body — and they land on the one code path that merges to `main`. The review therefore focused on injection, gate visibility, and fail-closed behaviour.

### Checklist Results

| Area | Result |
|------|--------|
| Secrets & Credentials | Pass |
| Injection Prevention | Pass |
| Auth & Authorisation | Pass (no change to approval or gate logic) |
| Data Protection | Pass |
| Gate & Trust-Chain Integrity | Pass |
| XSS / API Security | Not applicable |

---

### What I verified, and how

**1. Injection — clean.** `_tracker_merge_gh` builds a bash array and calls `gh "${args[@]}"`. No `eval`, no word splitting, no glob expansion. I confirmed this with a mock `gh` that dumps argv **NUL-separated**, so "one argument containing whitespace or a newline" is provably distinguishable from "two arguments". A `subject` containing a newline plus the literal text of a merge command produced exactly **11 argv elements** — the expected count, so the newline did not split the argument. Payloads `$(touch CANARY)`, `` `touch CANARY` ``, `x; touch CANARY`, `*`, and `a b  c   d` each arrived as a single verbatim, unexpanded element. The canary file was never created.

**2. Argument injection — clean, and confirmed empirically rather than assumed.** A `subject` of `--squash`, `--repo evil/repo`, or `--body-file /etc/passwd` lands as the *value* of `--subject`, not as a new flag. That depends on `gh`'s parser consuming a `--`-leading next token as a flag value, so I tested it locally and non-destructively on gh 2.100.0:

```
$ gh config get --host "--squash"
accepts 1 arg(s), received 0            # --squash was consumed as --host's VALUE
$ gh config get --definitely-not-a-flag
unknown flag: --definitely-not-a-flag   # control: a token actually parsed as a flag
```

So a hostile PR title cannot become a `gh` flag in the space-separated `--subject <value>` form.

**3. No widening of the merge surface.** `is_merge_command`, `extract_pr_number`, `extract_repo_from_command`, and `merge_command_uses_variable` return **identical verdicts** on the 4-arg and 6-arg shapes, including with hostile subjects such as `"9999 evil/repo --squash"` and `"--repo attacker/repo"`. Repo and PR still resolve from positional args 1 and 2, which are the only args `merge_command_uses_variable` inspects — so the two new trailing arguments cannot displace them or flip that gate's verdict. The new early-return sits *before* kind dispatch, so it can only remove a merge invocation, never add one. Live evidence that the matcher is not theoretical: two of my own analysis commands were blocked by `block-unreviewed-merge.sh` because their text contained merge-command strings.

**4. The fail-closed refusal is genuine.** For the two shapes the PR claims, the refusal is `rc=1` with **no CLI invoked at all** — verified against an empty capture file, which is a stronger property than "the CLI rejected it".

**5. Secrets and disclosure — nothing new leaks.** No new logging, no `set -x`. The subject is a public PR title. The body file is created by `mktemp` at `0600` in a per-user `TMPDIR` (verified `-rw-------`), so the name is unguessable and the mode is not world-readable. gh's stdout stays discarded; stderr propagation is unchanged.

**6. Tests.** `test_tracker_pr_merge.sh` 51 PASS / 0 FAIL. `test_forge_aware_extract_pr.sh` 49 PASS / 0 FAIL. `test_merge_invocation_not_substituted.sh` 7 PASS / 0 FAIL. The four gate suites — `block_unreviewed_merge`, `block_merge_on_red_ci`, `require_architecture_review`, `require_design_review_for_ui` — all exit 0. `shellcheck` is clean on both changed shell files.

---

### Findings

#### [MEDIUM] Stale security invariant guarding an `eval` path

**Location**: `.claude/hooks/_lib-tracker.sh` — `_tracker_merge_custom`'s comment block, and its `eval "$cmd"`.

The comment above `_tracker_merge_custom` states:

> unlike `_tracker_review_custom` / `_tracker_create_custom`, there is no arbitrary/untrusted free-text value in a merge call at all

After this PR that sentence is **false at the public-function level**. `tracker_pr_merge` now accepts `subject`, which for a fork PR is attacker-controlled free text. It is correctly *not* forwarded to the custom adapter today — `_tracker_merge_custom` is still called with four arguments — so there is no live vulnerability, and the AgDR's "Scope: `gh` kind only" reasoning is sound.

The risk is latent and specific: that comment is exactly the reasoning a future maintainer would rely on when adding a `{subject}` placeholder to the `merge_command` template, and doing so would be a direct command injection into `eval` (CWE-78). It currently reads as a standing guarantee rather than a fact about today's argument list.

**Remediation**: reword the comment to say that `subject`/`body_file` are deliberately not forwarded *because* they are untrusted free text, and that adding either to the template requires the same guarding `_tracker_review_custom` applies. A negative test asserting the custom adapter receives no more than four arguments would pin it mechanically.

#### [MEDIUM] Two silent-degrade shapes reintroduce #1136 with `rc=0`

**Location**: `.claude/hooks/_lib-tracker.sh` — the `tracker_pr_merge` guard, and `_tracker_merge_gh`'s `if [ -n "$subject" ] && [ -n "$body_file" ]`.

Measured behaviour across the full matrix:

| `subject` | `body_file` | Result |
|-----------|-------------|--------|
| empty | empty | bare squash, rc=0 — correct, backward compatible |
| **non-empty** | **empty** | **bare squash, rc=0, silent** |
| **empty** | **valid, non-empty** | **bare squash, rc=0, silent** |
| non-empty | valid, non-empty | `--subject`/`--body-file` appended, rc=0 — correct |
| non-empty | missing | rc=1, no CLI invoked — correct |
| non-empty | zero-byte | rc=1, no CLI invoked — correct |

Row 3 contradicts the shipped documentation. `tracker_pr_merge`'s own header says:

> An empty subject with a body_file is accepted (gh happily takes `--body-file` alone)

It is not accepted. The public guard passes it, then `_tracker_merge_gh` requires *both* to be non-empty and silently drops the body file. The caller asked for a specific commit body, got a bare squash, and received a success exit code.

Neither row is reachable from `/approve-merge` as written — step 6 sets both or neither — so this is not a live defect. I raise it at MEDIUM because the PR's stated premise is "fail-safe, not fallback", these are the only two shapes where the library falls back silently, and me2resh/apexyard#1152 makes "a guard proceeds when it cannot honour its precondition" an explicitly tracked shape on this exact trust chain. It is not HIGH: the degradation is to the pre-existing #1136 bug, not to an unsafe merge or a gate bypass.

**Remediation** (two lines, either direction): honour body-file-only by dropping the `-n "$subject"` requirement so the code matches the documented contract, or return 1 when exactly one of the pair is supplied. Today the code and the documentation disagree in the fail-open direction.

#### [LOW] `-s` is a size test, not a readability test

**Location**: `.claude/hooks/_lib-tracker.sh` — `if [ -n "$body_file" ] && [ ! -s "$body_file" ]`.

Both the function header and AgDR-0132 say the guard fails closed on an **unreadable** `body_file`. `-s` stats the path, so two cases pass the guard and reach the CLI:

- a non-empty file with mode `000` — verified: flags appended, guard did not fire;
- a **directory** — verified: `-s` is true for a directory, flags appended.

Real `gh` would then error and the merge would fail non-zero, so the end state is still fail-closed — but through the CLI, not through the guard, while the documentation claims the guard does it. I did not execute a real `--body-file` merge against either case, because that requires a live merge.

**Remediation**: `[ ! -r "$body_file" ] || [ ! -s "$body_file" ] || [ -d "$body_file" ]`, or drop the word "readable" from the comment and the AgDR.

#### [LOW] Neither layer verifies the body actually carries the trailer

Both guards check "non-empty". The property being protected is "the final paragraph is `Released-From: <sha>`". A release PR whose body was edited after `/release` wrote it — trailer no longer last, or body replaced — passes both checks and merges, silently reproducing #1136 under a new trigger, which is the failure class the AgDR says it closes. Related edge case: `gh pr view --json body -q '.body'` on a null body emits the 4-byte string `null`, which satisfies `-s`.

The compensating control is real — `/release` Rule 12's post-merge trailer verification is a documented STOP — which is why this is LOW rather than MEDIUM.

**Remediation**: one grep in step 6 before merging, for example `grep -qE '^Released-From: [0-9a-f]{7,40}$' "$RELEASE_BODY_FILE"`.

#### [INFO] Release-class detection reads attacker-influenceable text

`RELEASE_SUBJECT="$PR_TITLE"`, and detection runs `echo "$PR_TITLE" | grep -qE '^release\('` on a value a fork PR author controls. Two notes:

- `echo "$VAR"` where the value begins `-e`, `-n`, or `-E` has those consumed as options by bash's builtin, so detection would operate on a mangled string. `printf '%s\n' "$PR_TITLE" | grep ...` is the robust form. This pattern is inherited from the pre-existing sync-class block, not introduced here — worth fixing in both together if you touch it.
- A fork PR deliberately titled `release(...)` would have its own body used as the squash message, which could carry a forged `Released-From:` trailer. I checked the consumer: `bin/release-changelog.sh` line 95 reads the trailer **only from `PREV_TAG`'s own commit**, so a forged trailer on a non-tagged squash commit has no downstream effect. Combined with the mandatory human `/approve-merge` step, the impact is negligible.

#### [INFO] A new local-file to published-commit-message sink

`tracker_pr_merge` now has a parameter whose **file contents** are published into a permanent, public commit message. The single production caller writes that file itself from `gh pr view`, so there is no exposure today. One line in the function header stating that `body_file` contents become public would stop a future caller from pointing it at a credentials file, an env dump, or a build log.

#### [INFO] Temp-file lifecycle, and one now-stale retry instruction

`mktemp` gives an unpredictable name at `0600`, so there is no symlink-squat or predictable-path attack on the new path — a genuine improvement over the fixed `/tmp/release-pr-body.md` path that `/release` step 6 still documents for the manual fallback (pre-existing; the occurrence count is identical on base and HEAD, so this PR does not change it). Cleanup runs unconditionally after the merge including on failure, and `rm -f ""` for the non-release case is a silent no-op (verified exit 0). The gap between the `-s` check and gh's read is a TOCTOU window, but on an unguessable `0600` path it is not practically exploitable.

One behavioural consequence worth a doc tweak: step 7 deletes the body file, so the sentence *"the CEO marker stays on disk so the user can retry the merge after fixing the cause without re-approving"* is now inaccurate for a release PR — a bare retry of step 7 hits the guard and returns 1. That fails in the safe direction, but the instruction should say step 6 must be re-run.

#### [INFO] Cross-block variable persistence has an asymmetric, silent failure mode

`RELEASE_SUBJECT` and `RELEASE_BODY_FILE` are set in step 6's code block and consumed in step 7's. That is the same shell-state assumption `MERGE_STRATEGY` has relied on since AgDR-0053, so it is not newly introduced. The failure modes differ, though. If the variables do not carry, a **sync** PR degrades to `squash`, which `block-unreviewed-merge.sh` explicitly refuses for `sync/` PRs — loud. A **release** PR degrades to a bare squash, which nothing refuses — silent, and it is exactly #1136. If you want the fix robust rather than merely correct, the cheapest hardening is a guard in step 7: if the PR is release-class and `RELEASE_BODY_FILE` is empty, refuse.

---

### Could not verify

- Real `--body-file` merge behaviour against an unreadable file or a directory. That needs a live merge; I inferred fail-closed from gh erroring, and the guard demonstrably does not catch these.
- Whether the harness executes step 6 and step 7 in one shell in practice. `MERGE_STRATEGY`'s shipped sync-PR behaviour suggests it does, but I did not run `/approve-merge`.
- End-to-end trailer preservation on a real release PR, which requires an actual release merge.

### Note on the verdict mechanism

Posted as a comment, with the verdict stated above. GitHub refuses an `approve` or `request-changes` review state from a PR's own author on this single-account setup; that refusal is expected and is not a gate failure. There is no mechanical security merge gate in this repository, so the security marker I write is a **record, not a gate**. The load-bearing statement is the plain one: I consider this change safe to merge.

---
Reviewed commit: `e1842bd3dfc65f9a80ebf3da038a645daebeedd6`

me2resh (commented):
## Code Review: PR #1192

**Commit**: `e1842bd3dfc65f9a80ebf3da038a645daebeedd6`
**Scope**: Full (Heavy tier — `.claude/hooks/_lib-tracker.sh` is trust-chain merge-adapter code)

**Verdict: CHANGES REQUESTED** — one blocking defect, roughly three lines to fix.

The approach is right and the primary path works. I verified it empirically against the real library: a release-class PR with both a subject and a readable body file produces `--subject`/`--body-file` on the squash, and the missing/zero-byte body-file shapes refuse the merge without invoking the CLI at all. Detection is strict, the injection surface is clean, and the tests are genuinely behavioural.

**What blocks it:** there is a third subject/body_file shape that fails *open* — it silently degrades to a bare squash — and the guard's own comment asserts that shape is safe. The PR's stated design principle is "fails closed, not open, at two layers", and this is the gap in that claim. Making `_tracker_merge_gh` append the two flags independently closes the reachable failure entirely.

---

### H1 (blocking) — an empty `subject` silently discards a valid `body_file`, and the comment says the opposite

`.claude/hooks/_lib-tracker.sh:1473` requires **both** values to be non-empty:

```bash
if [ -n "$subject" ] && [ -n "$body_file" ]; then
  args+=(--subject "$subject" --body-file "$body_file")
fi
```

The fail-safe comment at `.claude/hooks/_lib-tracker.sh:1619-1621` states:

> An empty subject with a body_file is accepted (gh happily takes --body-file alone); the only failing shape is "caller wanted body_file honoured and it can't be read".

The premise about gh is correct — `gh help pr merge` exposes `-b/--body`, `-F/--body-file` and `-t/--subject` as three independent flags, so a body file alone is valid. The problem is that **the code never reaches gh in that shape**. Running the real library against a mock gh across the full matrix:

| subject | body_file | rc | CLI invoked | `--subject` | `--body-file` |
|---|---|---|---|---|---|
| set | valid | 0 | yes | 1 | 1 |
| `""` | `""` | 0 | yes | 0 | 0 |
| set | missing | 1 | **no** | – | – |
| set | zero-byte | 1 | **no** | – | – |
| **`""`** | **valid** | **0** | **yes** | **0** | **0** |
| set | `""` | 0 | yes | 0 | 0 |

Row 5 is the defect: a readable, non-empty body file passes the guard (`-s` is satisfied), then gets silently dropped — **a bare squash, the exact #1136 behaviour**, with no error surfaced.

**This is reachable from `/approve-merge` step 6.** `PR_HEAD_BRANCH` and `PR_TITLE` come from two *separate* `gh pr view` calls, each independently swallowing errors with `2>/dev/null`. If the title call fails transiently while the branch call succeeds:

- `PR_TITLE=""`, but `PR_HEAD_BRANCH="release/v5.5.0"` still matches the branch arm, so the release block fires;
- `RELEASE_SUBJECT=""` while `RELEASE_BODY_FILE` is a valid temp file;
- the block's fail-closed check validates **only the body**, so it passes;
- row 5 applies, and #1136 recurs silently on a real release PR.

The trigger is narrow — it needs a partial gh failure — and the outcome is no worse than today's behaviour rather than a regression. I am still calling it blocking because #1136 exists *precisely* because a silent degradation went unnoticed for a full release, and because a trust-chain library whose comment contradicts its code will mislead the next caller who reads it.

**Suggested fix** — append each flag on its own condition, which also makes the comment true:

```bash
[ -n "$subject" ]   && args+=(--subject "$subject")
[ -n "$body_file" ] && args+=(--body-file "$body_file")
```

The `body_file` guard at line 1622 already proves the file is readable and non-empty before this runs. Keep the asymmetry that matters — subject-alone still leaves the body to the repo default, which is the actual bug — but let a body file stand on its own. With this change the reachable path above still preserves the trailer, because the body is honoured and gh defaults the title.

### M1 — step 6 validates the body read but never the subject read

Same root cause as H1. In `.claude/skills/approve-merge/SKILL.md` step 6, the `gh pr view --json body` read is guarded and refuses on failure, but `RELEASE_SUBJECT="$PR_TITLE"` is assigned with no check that `$PR_TITLE` is non-empty. Even once H1 is fixed, an explicit guard — or a comment recording that an empty subject is safe because gh defaults the title — would make the two-layer fail-closed claim hold end to end.

### L1 — `-s` does not catch "unreadable but non-zero"

The comment at line 1617 says an "unreadable or empty" `body_file` returns 1. `[ -s file ]` uses `stat`, so a non-empty file with no read permission passes the guard; the refusal then happens inside gh, returning gh's exit code rather than 1. Net behaviour is still fail-closed, so this is a comment-accuracy nit rather than a behavioural one.

### L2 — the untested shapes are exactly the broken ones

Cases 12–15b are good behavioural tests: they assert on captured argv and on "no CLI invoked", so 15/15b would genuinely fail if a bare-squash fallback were reintroduced. But there is no case for `subject=""` with a valid `body_file`, nor for a set `subject` with `body_file=""`. A case asserting `--body-file` is present with an empty subject would fail today — that is the missing pin. Worth adding both alongside the H1 fix.

### L3 — the AgDR link will 404 after the branch is deleted

The PR body links the AgDR through a branch blob URL (`.../blob/fix/GH-1136-release-pr-squash-trailer/...`). Post-merge branch deletion breaks it; a repo-relative path (`docs/agdr/AgDR-0132-...md`) survives.

### L4 — commit subject length

`fix(#1136): preserve Released-From trailer on release-PR squash via /approve-merge` is 84 characters against the handbook's under-70 guidance. Advisory only — the body itself is exactly what the handbook asks for (WHY, alternatives, trade-offs).

---

### Verified good (no action needed)

- **Backward compatibility.** The old 4-arg call is byte-for-byte unchanged (test cases 13/14, plus row 2 of my probe). `_lib-extract-pr.sh` still extracts PR `1192` and returns `is_merge_command`=YES for the old 4-arg shape, the new 6-arg shape, a subject containing `(` and `)`, and the documented variable form. Args 1–2 precede the new params so nothing positional shifted, and the `[^|;&)]*` span fence truncating at a `)` inside a subject happens *after* repo and PR are captured. The four `Bash(tracker_pr_merge *)` matchers in `settings.json` are untouched — and fired live against my own test harness during this review, which is incidental confirmation that gate coverage is intact.
- **Detection strictness.** `^release/v[0-9]+\.[0-9]+\.[0-9]+$` is fully anchored with escaped dots — *stricter* than the sync precedent's `^sync/main-to-dev-after-` prefix match. `release-notes/foo` does not match (needs `release/`), `release/v1.2.3-rc1` does not match (trailing `$`), and `fix(#123): release something` does not match `^release\(`. I found no false-positive path.
- **Injection surface.** `args+=(--subject "$subject" --body-file "$body_file")` followed by `gh "${args[@]}"` passes both as discrete argv elements with no eval, so neither a leading `-` nor shell metacharacters in a fork-PR title can be reinterpreted. Notably the `custom` kind — the only adapter that eval's a template — is deliberately *not* passed subject/body_file. That is the right call and worth preserving.
- **Temp file handling.** `mktemp` (0600 by default), with `rm -f` on both the early-exit and post-merge paths.
- **Test suite.** 145 PASS / 2 FAIL; both failures (`test_lib_self_location_cwd_anchor.sh`, `test_validate_search_config.sh`) reproduce on the clean base `5408ddd`. No new failures. `test_tracker_pr_merge.sh` alone: 51/51.
- **AgDR-0132.** A real decision record, not a diff restatement — four options with genuine trade-offs, an explicit scope limit (gh only), and consequences that name the signature change. Matches `templates/agdr.md`. Numbering is correct; the duplicate `AgDR-0131-*` pair already exists on the base and is not this PR's doing, though it is worth a separate cleanup ticket.
- **PR hygiene.** `Refs #1136`, Glossary present, Summary bullets narrative (what changed *and* why it matters), AgDR linked.

### Handbook findings

Discovery ran over the public layer; the private `custom-handbooks` directory is not present, so that layer was skipped. The semantic supplement ran and returned no handbook chunks, so it added nothing.

⚠ **Commit Message Quality** — `handbooks/general/commit-message-quality.md` (advisory, discovery: path-convention)
- nit: subject line is 84 chars against the under-70 guidance (L4 above). Body content meets the WHY/trade-off recipe.

Loaded and recorded **not applicable**: `handbooks/architecture/migration-safety.md` (ENFORCEMENT: blocking, but there are no migration or schema paths in this diff) and `handbooks/architecture/clean-architecture-layers.md` (no application-layer code; this is a shell library). Skipped by convention: `handbooks/language/typescript/strict-mode.md` (no `.ts`/`.tsx` in the diff) and `handbooks/domain/` (README only).

### Could not verify

The end-to-end claim that `--subject`/`--body-file` actually lands the trailer as the **final paragraph** of the real GitHub squash commit. That needs a live release merge. The flag semantics support it — three independent flags, and omitting the body lets the repo's `COMMIT_MESSAGES` default apply — and the mock proves the flags reach gh, but the GitHub-side result is unverified here. The AgDR's reasoning about a subject alone falling back to the repo-default body is consistent with that flag separation.

---
📌 Reviewed commit: `e1842bd3dfc65f9a80ebf3da038a645daebeedd6`

me2resh (commented):
## Security Re-Review: PR #1192

**Commit**: `1c5b74a4e6fcef97ce95bd448233c24f5a6ca6a4`
**Prior pass**: `e1842bd` (APPROVED, no Critical/High). That marker was deleted — it recorded a pass against a SHA that is no longer HEAD.

### Verdict

**APPROVED** — no Critical or High findings. The delta closes both defects I raised, and it does so without weakening the injection properties I proved last pass. Safe to merge from a security standpoint.

### Summary

The delta does what it claims. `--subject` and `--body-file` now append on independent conditions, closing my MEDIUM #2 and Rex's H1. The fail-closed guard tests `-f`/`-r`/`-s` instead of `-s` alone, which closes my LOW and — verified empirically — makes the guard do the work its comment claims rather than leaving the forge CLI to error downstream. `/approve-merge` step 6 refuses an empty `$PR_TITLE` on a release-class PR, which as a side effect makes the `body_file`-only shape unreachable from the skill entirely.

One MEDIUM carries over untouched (`_tracker_merge_custom`'s comment) and one new LOW appears (a second stale comment, introduced by this delta). Neither blocks.

### 1. Injection properties preserved — verified

I re-ran the NUL-separated argv mock against the new two-statement append. Argv construction is still sound in every shape:

| Shape | argc | Result |
|---|---|---|
| Hostile subject + body_file | 11 | Subject is exactly one element |
| body_file only (empty subject) | 9 | `--body-file` present, no `--subject` |
| Hostile subject only | 9 | `--subject` present, no `--body-file` |
| Neither | 7 | Byte-identical to pre-#1136 behaviour |

The hostile subject carried a leading `--delete-branch --admin`, a `;`, a newline, backticks, `$(...)`, both quote styles, and a pipe. It arrives as a single argv element every time. Two properties make that hold, and both survive the change:

- `args+=(--subject "$subject")` appends **two** array elements, so the flag literal always immediately precedes its value. A subject beginning `--` cannot be read as a flag, because the forge CLI's parser takes the next argument as the value of a string flag unconditionally.
- Nothing is interpolated into an `eval`'d string on the gh path.

Splitting one `if` into two conditions changed neither property. The non-release path is unchanged at the argv level, which matters more than the release path — it is the one every ordinary merge takes.

### 2. The `-f`/`-r`/`-s` guard is genuinely fail-closed — verified

I drove the public merge function against thirteen `body_file` shapes with a capturing mock, recording the raw test-operator readings beside each verdict:

| Shape | Verdict | CLI invoked | `-f` | `-r` | `-s` |
|---|---|---|---|---|---|
| regular non-empty | pass | yes | T | T | T |
| symlink to good file | pass | yes | T | T | T |
| zero-byte | **refuse** | no | T | T | F |
| directory | **refuse** | no | F | T | T |
| directory, non-empty | **refuse** | no | F | T | T |
| mode-000 file | **refuse** | no | T | **F** | T |
| symlink to nonexistent | **refuse** | no | F | F | F |
| symlink to directory | **refuse** | no | **F** | T | T |
| FIFO | **refuse** | no | **F** | T | F |
| `/dev/null`, `/dev/zero` | **refuse** | no | F | T | F |
| nonexistent path | **refuse** | no | F | F | F |
| trailing slash on a file | **refuse** | no | F | F | F |

Read the bolded cells. Four shapes — directory, non-empty directory, mode-000, and symlink-to-directory — report `-s` **true** and would have passed the old guard, reaching the forge CLI with a `body_file` that cannot serve as a body. All four now refuse before any CLI invocation. That is exactly the gap I flagged, and it is closed. The mode-000 case is caught only by `-r`; the two directory cases and the FIFO only by `-f`. Both added operators are load-bearing, not decorative.

The FIFO result is a bonus worth naming: `-f` refuses it deterministically, whereas the forge CLI opening a FIFO for read would block indefinitely. The old guard caught FIFOs only through the accident of `st_size == 0`.

Two caveats, neither a finding:

- `test -r` uses the real UID, so a merge run as root would not refuse a mode-000 file. Running a release merge as root is not a supported path, and `-f` still catches every non-regular shape regardless of UID.
- An unavoidable TOCTOU window sits between the guard and the CLI's `open`. `body_file` comes from the caller's own `mktemp`, not from an untrusted source, so this is inherent to handing a path to a subprocess rather than a defect introduced here.

### 3. Closing the silent-degrade shapes did not introduce a new one

The `body_file`-only path reaches the forge CLI with `--body-file` and no `--subject`. I checked whether that is a fresh degrade in disguise. It is not, on two independent grounds.

**The CLI accepts the combination.** gh 2.100.0 documents `-t/--subject` and `-F/--body-file` as independent optional flags. Searching the binary's validation strings, the only mutual exclusion in this family is `--body` versus `--body-file`; there is no `--subject` co-requirement. The subject then falls to the repo's `squash_merge_commit_title`, which for this repo is `COMMIT_OR_PR_TITLE`. The body is still supplied explicitly, so `squash_merge_commit_message=COMMIT_MESSAGES` never applies and the `Released-From` trailer still lands as the final paragraph. The degrade this parameter exists to prevent does not occur.

*Scope of that claim:* established from the CLI's flag documentation, its embedded validation strings, and GitHub's documented-optional `commit_title`. I did not execute a real body-file-only merge, so the server-side default title is inferred from the repo setting rather than observed.

**The shape is unreachable from `/approve-merge` anyway.** The new `$PR_TITLE` guard exits 1 before `RELEASE_BODY_FILE` is created. Release-class detection fires on a branch match or a title match; if the title is empty the title branch cannot have fired, so only the branch match reached the guard, and the guard refuses. `RELEASE_SUBJECT` is therefore always non-empty whenever `RELEASE_BODY_FILE` is set. The `body_file`-only shape now exists only for a direct library caller, and case 16 pins its behaviour.

**On a fork-controlled body with no subject.** The body written to the temp file is the PR description, which on a fork PR is attacker-authored. That is not new exposure. Before this change the squash body came from `COMMIT_MESSAGES` — the concatenated commit messages of the same fork branch, equally attacker-authored. The change swaps one attacker-controlled source for another at the same trust level.

The specific worry worth checking was trailer injection: a fork PR body whose final paragraph forges `Released-From: <sha>`. I traced the read-back surface. `bin/release-changelog.sh:95` reads the trailer from `"$PREV_TAG"` — the commit a release tag points at, created by the auto-tag workflow when a release PR lands on `main`. A forged trailer on an ordinary `dev` commit is inert; nothing reads it. And the same forgery was already available before this change through the last commit message on the branch, since `COMMIT_MESSAGES` places that message's final paragraph at the end of the squash body. No new exposure and no new reachable sink.

### 4. MEDIUM #1 — `_tracker_merge_custom` comment (still open; my call: **do not block**)

Untouched by the delta. The comment above the `eval` still asserts there is *"no arbitrary/untrusted free-text value in a merge call at all"*, which is now false at the public-function level: the public merge function accepts a free-text `subject` and a `body_file` path.

**Not blocking, and here is the evidence.** I confirmed the dispatch arity directly — line 1648 passes the custom adapter four arguments: repo, pr, strategy, delete_branch. The custom adapter provably never receives `subject` or `body_file`, so all four substituted placeholders remain guarded exactly as the comment's *conclusion* describes. The defect is the over-broad *justification*, not the current behaviour. It is a maintainer trap, not a live vulnerability, and blocking a correct security fix on a comment is disproportionate.

**File it as a follow-up, and fix it before anyone extends the custom adapter.** Making this fix forge-agnostic is the obvious next request, and a maintainer reading that comment as a standing guarantee would wire `{subject}` into the eval'd template without adding a guard. Narrow the wording to what is true: the four placeholders this adapter substitutes are enum-, numeric-, and charset-guarded; `subject` and `body_file` are deliberately not passed here, and wiring them in requires a new guard.

### 5. Gate surface unchanged — verified

`_lib-extract-pr.sh` is untouched by this PR — the diff against the branch base on that path is empty. I confirmed identical verdicts on the new shape rather than relying on that:

| Invocation | Recognised as a merge | PR | Repo |
|---|---|---|---|
| Old 4-arg (pre-PR) | yes | 1192 | me2resh/apexyard |
| **New 6-arg, documented (variable refs)** | yes | **1192** | `$PR_HOST_REPO` |
| New 6-arg, literal repo, variable subject | yes | 1192 | me2resh/apexyard |
| New 6-arg, literal benign subject | yes | 1192 | me2resh/apexyard |
| New 6-arg, subject containing `;` `\|` `&` `)` | yes | 1192 | me2resh/apexyard |
| New 6-arg, subject embedding another repo slug | yes | 1192 | me2resh/apexyard |
| New 6-arg, empty subject | yes | 1192 | me2resh/apexyard |

The documented step-7 shape gives identical verdicts to the old one. Repo and PR are positional arguments 1 and 2, ahead of the new trailing arguments, so the span capture reads them before any `|;&)` inside a subject can truncate the span.

**One informational note, pre-existing.** A subject containing the literal text of a forge-shaped merge command naming a *different* PR is extracted as that other PR number, because the PR extractor checks the forge-CLI positional branch before the wrapper branch. I checked whether this PR introduces that: it does not. The old 4-arg shape with a trailing shell comment carrying the same text already yields the wrong number, so the precedence is a pre-existing property of the extractor.

What this PR adds is a free-text argument *slot* where such text could sit — but only for a caller that inlines a literal subject. The documented step-7 shape passes `"$RELEASE_SUBJECT"` as a **variable reference**, so the untrusted PR title never enters the command text the matcher reads, and the verdict is correct (1192, verified above). I am not raising this against this PR. It deserves its own ticket, because the failure direction is not uniformly safe: a wrong PR number usually means a missing marker and a blocked merge, but if valid markers happen to exist for the confused number, the gate would evaluate the wrong PR's approvals.

### 6. GitHub review-state asymmetry — worth recording

Independently verified, not taken on trust:

- `dev` ruleset: `dismiss_stale_reviews_on_push: false`, `require_last_push_approval: false`, `required_approving_review_count: 1`.
- PR #1192 reviews: `atlas-apex` **APPROVED** at `e1842bd`. HEAD is `1c5b74a`.

GitHub's server-side approval requirement is therefore satisfied right now by a review attached to a commit that is no longer HEAD, while ApexYard's SHA-bound markers correctly invalidated and forced this re-review. The two controls disagree, and the weaker one is the one a branch-protection rule enforces.

This matters beyond bookkeeping. `pr-workflow.md` already states that the load-bearing signal is the local SHA-bound marker, never GitHub's review-state UI — and this PR is a live demonstration of why that wording is correct. The residual risk is a reader who sees a green "Approved" badge and treats it as current. Two settings would close the gap at the forge (`dismiss_stale_reviews_on_push: true`, or `require_last_push_approval: true`); both are repo-configuration decisions outside this PR's scope. I am recording the observation, not requesting the change here.

`require_extra_approval_for_unattributed_changes: true` is already on, but it does not cover the ordinary push-after-approval case.

### 7. Deferred item — not worsened

The `RELEASE_SUBJECT`/`RELEASE_BODY_FILE` step 6 to step 7 code-block boundary is getting its own ticket, and I am not re-raising it. The delta does **not** make it worse in kind: step 6 already contained an `exit 1` with the same cross-block dependency, so the new title guard adds a second instance of an existing shape rather than a new one.

One clarification for whoever picks that ticket up, because the AgDR's defence-in-depth claim is narrower than it reads. The library fail-safe stops a caller that passes an **unreadable** `body_file`. It cannot stop a caller that passes **no** `body_file`, since the guard only fires when `body_file` is non-empty. If the variables do not carry across the block boundary, `body_file` is empty, the guard stays silent, and the merge proceeds as a bare squash — reintroducing #1136 exactly. The library layer does not backstop that failure mode, so the fix has to live at the boundary itself.

### Findings

| Severity | Finding | Status |
|---|---|---|
| — | Two silent-degrade shapes with `rc=0` (my MEDIUM #2 / Rex's H1) | **Closed** — verified |
| — | `-s` is a size test, not a readability test (my LOW) | **Closed** — verified |
| MEDIUM | `_tracker_merge_custom` comment asserts no untrusted free text; false at the public-function level, and it sits above an `eval` | Open — **not blocking**, follow-up |
| LOW | `_tracker_merge_gh` header comment still says *"When both are non-empty… are appended"*; the delta made the flags independent | **New**, introduced by this delta |
| Info | Extractor precedence: literal forge-shaped merge text in an argument outranks the wrapper's own PR argument | Pre-existing; unreachable from the documented shape |

The new LOW is the same comment-drift class as the MEDIUM #2 I raised last pass, inverted: previously the code was wrong and the comment right; now the code is right and the header comment understates it. Harmless in direction — nobody is misled into believing a guard exists that does not. The final sentence of that same header block, about separate argv elements, remains accurate and load-bearing; I verified it empirically.

Two stale comments in one trust-chain file is a pattern rather than a one-off, which is the real argument for folding the MEDIUM and the LOW into a single comment-accuracy commit.

### Checklist

| Area | Result |
|---|---|
| Secrets and credentials | Pass — no secrets; the `mktemp` body file holds public PR text |
| Injection prevention | Pass — argv-array construction re-verified under a hostile subject |
| Auth and authorisation | Pass — human-only `/approve-merge` invocation unchanged |
| Data protection | Pass — no new data sinks |
| Gate and trust-chain integrity | Pass — no gate weakened, loosened, or made to fail open; extractor verdicts identical |
| Fail-closed behaviour | Pass — strengthened; four previously-passing bad shapes now refuse |

### Supporting evidence

- 59 assertions pass and 0 fail in `.claude/hooks/tests/test_tracker_pr_merge.sh`, including new cases 16, 17, and 18. The three new cases pin the exact shapes the delta changed.
- CI at `1c5b74a`: all seven checks pass, including `shellcheck .claude/hooks` and the full hook suite. CodeQL reports `skipping`, its normal state here.
- `.claude/skills/release/SKILL.md` changes are documentation only and accurately describe the new `/approve-merge` behaviour.

### What I could not verify

- The server-side default squash title for a `--body-file`-without-`--subject` merge. It is inferred from `squash_merge_commit_title=COMMIT_OR_PR_TITLE` plus the CLI's documented flag independence; I did not execute a real merge in that shape. Case 16 pins the flag construction, not GitHub's titling.
- Runtime behaviour of `/approve-merge` step 6 into step 7 as an actual skill run. I reviewed the SKILL.md text; the cross-block variable-carry question is the deferred ticket's subject.

---
📌 Reviewed commit: `1c5b74a4e6fcef97ce95bd448233c24f5a6ca6a4`

me2resh (commented):
## Code Review: PR #1192 (re-review)

**Commit**: `1c5b74a4e6fcef97ce95bd448233c24f5a6ca6a4`
**Scope**: Full — Heavy tier (trust chain: `.claude/hooks/_lib-tracker.sh` plus the merge-approval skill). Reduced scope does not apply.

### Verdict

**APPROVED.** H1 is genuinely closed, and I could not find a remaining path to a silent bare-squash. One Medium finding stands: the delta fixed the code but left three copies of the *old* contract sentence in place, so the file now documents behaviour it no longer has. That is prose, not behaviour, so it does not block the merge — I would still rather it went in before you merge, because it is the same comment-disagrees-with-code shape that produced H1 in the first place.

Next action: fix M2 (three sentences) and L3 (one URL), then merge.

### What changed since the last pass

| Item | Last pass | Now |
|------|-----------|-----|
| H1 — subject/body-file coupled | CHANGES REQUESTED | **Closed** — independent conditions, both directions pinned by tests |
| L1 — guard weaker than its comment | Raised | **Closed** — `-f`/`-r`/`-s`, comment rewritten to match |
| M1 — unguarded title read reached the H1 path | Raised | **Closed** — refuses before `mktemp` |
| L2 — no regression test for the broken shape | Raised | **Closed** — cases 16/17/18 |

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass
- Security: Pass
- Performance: Pass (n/a)
- PR Description & Glossary: Pass — one stale bullet, see M2
- Technical Decisions (AgDR): Pass — AgDR-0132 linked, mirrors the AgDR-0053 sync precedent
- Adopter Handbooks: Advisory finding only (below)

### 1. Is H1 closed, and did the fix introduce anything new?

Closed. Each flag now appends on its own condition, and I checked the two implications you flagged.

**Exit-status implication — safe.** `gh "${args[@]}" >/dev/null` is still the last command in the function, so the return value is gh's exit status. Neither `[ -n ... ]` test can leak a false return. That is the meaningful difference from the classic `[ ] && cmd`-as-last-line trap, and the fix stays on the right side of it.

**`set -e` — safe, verified rather than assumed.** I reproduced the exact shape under `set -e` on this machine's bash 3.2.57 and on `/bin/sh`: a short-circuiting `[ ] && cmd` is exempt from `-e` because it is part of an AND-OR list, so a false test does not abort. Both the "both empty" and "only body_file" cases ran to completion and returned 0. No caller observes a `[ -n ]` result anyway, since gh's status overwrites it.

The new lines follow the `delete_branch` idiom directly above them, which is the right call — one idiom in one function.

### 2. Does the L1 guard change alter any previously-passing shape?

No. I ran the old and the new guard side by side over the full shape matrix:

| body_file | Old | New |
|-----------|-----|-----|
| `""` (ordinary non-release path) | ALLOW | **ALLOW** |
| regular readable non-empty file | ALLOW | **ALLOW** |
| symlink to a valid file | ALLOW | **ALLOW** |
| zero-byte file | REFUSE | REFUSE |
| missing path | REFUSE | REFUSE |
| directory | ALLOW, reached gh | **REFUSE at guard** |
| mode-000 file | ALLOW, reached gh | **REFUSE at guard** |

Both previously-passing paths you asked about are byte-identical, and symlinks still resolve (`-f`/`-r`/`-s` all dereference) — that was the shape most at risk from adding `-f`. The only two changes are strictly toward fail-closed, and both were already destined to fail downstream at gh; the change moves the refusal to where the comment always claimed it was.

The `{ ...; }` grouping is required and correct: `&&` and `||` are equal-precedence and left-associative, so the braces are what stop the `||` chain from swallowing the `[ -n ]` guard.

### 3. Is M1 correct, in the right place, and leak-free?

Yes to all three.

- **Placement** — the guard sits inside the release-class block and *before* `RELEASE_BODY_FILE=$(mktemp)`, so the refusal path creates no temp file. No leak.
- **Correctness** — it can only fire when the branch matched and the title is empty, because an empty title cannot match `^release\(`. That is exactly the reachable path from last pass, and nothing wider.
- **Consistency** — `exit 1` matches the existing body-read guard immediately below, and the message states what failed, why it refuses, and what to do.

### 4. Are cases 16/17/18 real regression tests?

They assert the right things. Taking as given that 2 assertions flip against the old guard, I judged only the content:

- **16** pins the H1 shape directly: exit 0, body-file flag present, subject flag absent. The third assertion is the one that matters — it proves the flags are genuinely independent rather than both suppressed.
- **17** is the mirror, so the pair brackets the behaviour from both sides instead of only the side that broke.
- **18** targets the `-f` clause and asserts *no CLI invoked*, which is the right assertion — it proves the refusal happens at the guard, not at gh.

**One gap: the `-r` clause has no test.** I confirmed it is live rather than dead code — a mode-000 file is allowed by the old guard and refused by the new one, so `-f` alone does not cover it — but nothing pins it. A test needs an `[ "$(id -u)" -ne 0 ]` skip, since root reads mode-000 files and the case would silently invert in a root CI container. Worth adding, not worth blocking on.

### 5. Anything the delta broke in the previously-passed parts?

Nothing. Re-verified each:

- **4-arg backward compatibility** — cases 13/14 cover it, and the params default to empty.
- **`_lib-extract-pr.sh`** — untouched by this PR, but the *caller* shape grew to 6 args, so I re-ran extraction against it. `extract_pr_number`, `extract_repo_from_command`, and `is_merge_command` return identical results for the 4-arg baseline and all three 6-arg variants. That includes the worst case I could construct — an inlined subject containing parentheses, `"release(#1136): v5.5.0"`, which truncates the extractor's `[^|;&)]*` span. It truncates *after* args 1 and 2, so repo and PR resolution are unaffected. In the documented form the argument is a bare variable reference with no parens at all.
- **Anchored release detection** — `^release/v[0-9]+\.[0-9]+\.[0-9]+$` is anchored at both ends with escaped dots, and matches what `/release` actually creates (`git checkout -b "release/vA.B.C"`, step 4). It will not match `release/v1.2.3-hotfix`.
- **Injection safety** — subject and body_file enter gh as separate argv elements, never interpolated into an eval'd string, and the `custom` adapter is not passed them at all, so the template path is untouched.
- **Merge gates and suites** — the full `bin/run-hook-tests.sh` run finished at **PASS=145 FAIL=2 TOTAL=147**, the two failures being exactly the stipulated pre-existing ones (`test_lib_self_location_cwd_anchor.sh`, `test_validate_search_config.sh`). No new failures. I had also run the six most relevant suites directly — the four merge-gate hooks (41 + 49 + 34 + 27 + 27) and the merge-adapter suite (59) — for 237 assertions, 0 failures. `shellcheck` and `bash -n` clean on both changed shell files.

### Issues Found

**M2 (Medium, non-blocking) — three copies of the old contract survive and now contradict the code.**

The delta corrected the behaviour and added an accurate inline comment, but left the previous "when BOTH are non-empty" wording in three places:

1. `.claude/hooks/_lib-tracker.sh:41` — the top-of-file contract: *"subject/body_file are OPTIONAL (gh kind only): when BOTH are non-empty, the merge passes ..."*.
2. `.claude/hooks/_lib-tracker.sh:1455` — the function's own header: *"When both are non-empty, ... are appended"*. This one sits about ten lines above the new comment that says the opposite.
3. The PR body's first Summary bullet — the same "when both are non-empty" claim.

Related, in the same top-of-file block (`:49`): *"an unreadable/empty body_file **with a non-empty subject** fails closed"*. The guard has never consulted the subject; that qualifier was inaccurate before the delta too, and is more visible now the sentence is being read again.

Why a comment is Medium rather than a nit: line 41 is the public contract for a trust-chain function. A future caller who reads it and concludes "body_file alone does nothing, I must synthesise a subject" would be re-deriving H1 from the documentation. That is second-order, and the tests now pin the truth, which is why it does not block — but it is the same comment-versus-code divergence that produced H1, recurring in the very function that fix was about.

The fix is three sentences plus the PR bullet.

### Handbook Findings

`handbooks/general/commit-message-quality.md` — advisory.

- nit: commit `e1842bd`'s subject is 82 characters; the handbook asks for under 70. Carried forward as L4 below. Both commits otherwise have substantial why-first bodies, and the delta commit's body is a good example of the form — it names the finding, its reachability, and each fix with its identifier.

`architecture/clean-architecture-layers.md`, `architecture/migration-safety.md`, and `language/typescript/strict-mode.md` are not applicable to this diff: no layered application code, no migration or schema paths, no TypeScript. No private `custom-handbooks/` directory is present. The semantic supplement returned no handbook-path matches.

### Still open from the last pass

**L3 — the AgDR link 404s after merge. I would like this fixed before you merge.** The PR body links the AgDR as a branch blob URL (`.../blob/fix/GH-1136-release-pr-squash-trailer/docs/agdr/...`). The merge deletes the branch, so the link dies — and it is the durable pointer to the decision record for this change. Point it at `blob/dev/...` or use the repo-relative path. One line, permanent benefit.

**L4 — the 82-character subject. Optional, and cheaper than it looks.** Worth knowing: this is not a release-class PR, so the approval skill passes an empty subject and gh defaults the squash subject to the **PR title** — the same 82-character string. You do not need to rewrite history to fix it; editing the PR title before merging is enough. I would not hold the merge for it.

### On the deferred boundary issue

The `RELEASE_SUBJECT` / `RELEASE_BODY_FILE` step 6 to step 7 boundary is unchanged, and the delta did not make it worse. If anything it is marginally better: the M1 guard exits *before* `mktemp`, so the new refusal path adds no temp file that could outlive the block. The pre-existing shape — a gate-blocked step 7 leaves the step 6 temp file behind — is untouched. Noted, not re-raised.

### Suggestions

- Add an `-r` case alongside 18, guarded by a non-root skip, so the third clause of the guard is pinned rather than merely live.
- The merge-adapter test file's header case list still stops at 11; cases 12 through 18 are undocumented there, and that header is the first thing a reader of the file sees.
- The PR body's Testing section says "5 new cases"; there are now 8. The "145 PASS / 2 FAIL" line is from the first commit's run — it still holds, which I confirmed independently.

### What I could not verify

- Runtime behaviour against the real GitHub API. All merge tests use a mock gh, by design. Whether gh's own subject default on a bare body-file flag is what you want on a release squash is the one thing only a live release cut will confirm.
- I took as given, per your note, that cases 16/17/18 fail 2 assertions against the old guard, and that the two named suites fail on the clean base in this environment. The second of those I ended up confirming directly — the full run reproduced exactly those two failures and no others.

---
📌 Reviewed commit: `1c5b74a4e6fcef97ce95bd448233c24f5a6ca6a4`

atlas-apex (approved):

Issue #1146, opened 2026-08-15
[Bug] validate-commit-format.sh validates the body as the subject when git commit is given multiple -m flags


drmas:
## Affected

`.claude/hooks/validate-commit-format.sh`

## Given / When / Then

**Given** a commit whose subject is valid conventional-commit format, with the body supplied via a second `-m` flag

**When** running:

```bash
git commit -m "chore: initialise private portfolio" -m "- first body bullet
- second body bullet

A closing paragraph."
```

**Then** the hook blocks with `BLOCKED: Commit subject doesn't match the conventional commit format.` and echoes the **body** back as the offending subject — expected: the hook validates only the first `-m` value (`chore: initialise private portfolio`), which is valid, and allows the commit.

The reported subject in the error was the entire body with newlines collapsed to spaces:

```
Subject was:
  - first body bullet - second body bullet  A closing paragraph.
```

## Repro

1. In a fork with the hook wired (any repo — the hook resolves the ops root and applies session-wide)
2. Stage any change
3. Run: `git commit -m "chore: valid subject" -m "body text"`
4. Blocked, with the body reported as the subject

Both of these work correctly, which isolates the issue to multi-`-m` handling:

- `git commit -m "chore: valid subject"` (single `-m`) → allowed
- `git commit -F message.txt` (subject + blank line + body in a file) → allowed

## Framework version

v5.4.0

## Severity

major — wrong result with a workaround. `git commit -m subject -m body` is a standard git idiom for writing a multi-paragraph message without a temp file (git joins the values with blank lines between them). The workaround (`-F <file>`) is fine once known, but the error message actively misleads: it tells the author their subject is malformed while quoting text that is not the subject.

## Notes

Likely cause: the hook collects all `-m` values (or takes the last) instead of the first. Git's own semantics are that the first `-m` becomes the subject line and each subsequent `-m` is appended as a paragraph, so only the first value should be subject-validated.

Worth checking whether `validate-commit-format.sh` and `verify-commit-refs.sh` share message-extraction logic — if so, the trailer scan (`Closes #N` / `Refs #N`) may have the mirror-image bug, since trailers live in the *last* `-m`, not the first.

## Glossary

| Term | Definition |
|------|------------|
| Conventional commit | A commit message convention where the subject is `type: subject` or `type(scope): subject`, e.g. `fix(auth): handle expired refresh tokens`. |
[...]

PR #1194, opened 2026-09-06
fix: validate first commit message paragraph

OsamaAlSabry:
## Summary

-
-

## Testing

1.

## Glossary

| Term | Definition |
|------|------------|
|      |            |

### DIFF
--- a/.claude/hooks/tests/test_validate_commit_format_heredoc.sh
+++ b/.claude/hooks/tests/test_validate_commit_format_heredoc.sh
@@ -120,8 +120,14 @@ run_case "plain -m valid subject: pass silently" \
 run_case "plain -m bad subject: BLOCK" \
   'git commit -m "not a valid subject line"' 2 "BLOCKED: Commit subject"
 
-run_case "plain -m '\''quoted'\'' valid subject: pass" \
+run_case "plain -m 'quoted' valid subject: pass" \
   "git commit -m 'fix: a fix'" 0 ""
+
+# Git semantics: the first -m value is the subject; subsequent -m values are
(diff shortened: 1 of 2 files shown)


me2resh (changes requested):
## Review: changes requested

Thank you for tackling issue 1146. The focused regression test shows the right goal. I appreciate the small and clear scope.

### Blocking finding

The new parser searches raw command text for the first -m pattern. It does not identify the first git commit -m argument.

This command bypasses the hook on c786f75651e93869f6ff57c1acdfa496ef9dab69:

~~~bash
git commit --trailer='-m "fix: decoy"' -m "invalid subject"
~~~

The hook returns success. Git then creates the commit with invalid subject as its subject. The current dev hook blocks the same command.

Please replace the raw-pattern extraction with parsing that identifies real Git arguments. Fail closed when the command shape cannot be parsed safely.

Add tests for:

1. A valid subject with a body in a second -m value.
2. A valid-looking -m string in --trailer before an invalid subject.
3. Quoted and escaped values that contain -m text.

Please also update the PR title and body to reference issue 1146 and describe the validation approach.

## Glossary

| Term | Definition |
|---|---|
| Subject | The first line of a commit message. |
| Decoy | Text that looks like an option but is not the commit-message argument. |

OsamaAlSabry:
Thanks for reviewing this contribution. I’m closing this PR because Issue #1146 has since been resolved upstream by [PR #1211](https://github.com/me2resh/apexyard/pull/1211), which is already merged into `dev`. The merged implementation covers the safe command-token parsing and the requested regression cases, so this branch is now superseded.

Issue #1206, opened 2026-09-07
[Bug] Leak protection misses gh pr review and file content in a diff, and remediation is undefined


me2resh:
## Bug Scenario

**Given** `.claude/rules/leak-protection.md` treats private project identifiers as mechanically scrubbed before they reach a public framework repo

**When** an identifier reaches that repo by a path `block-private-refs-in-public-repos.sh` does not watch

**Then** it is published with nothing raising a warning

**Expected** the guard covers the paths this framework actually writes through, and a documented recovery exists when one gets past it.

Both halves were demonstrated live on PR #1205 within the same hour.

## Gap 1 — `gh pr review` is not a matched shape

The hook matches five shapes: `gh issue create`, `gh pr create`, `gh issue comment`, `gh pr comment`, and `gh api .../issues|/pulls`. **`gh pr review` is absent**, and that is what `tracker_review_submit` uses.

So every review Rex, Hakim, and Tariq post to a public repo bypasses leak protection entirely. Reviewers are the most exposed role, because quoting the artefact under review is the job — including the parts an author is mid-redaction on.

**Observed:** during the PR #1205 review the Solution Architect quoted a private identifier while confirming the author had removed it. Nothing caught it. It was found only because the reviewer reported it against itself in the next round.

The fix looks small: one matcher entry plus body extraction for the review shape.

## Gap 2 — file content in a diff is never scanned

The hook reads the **title and body of a command**. A private identifier inside a **file in a commit** is invisible to it.

**Observed:** on the same PR, an identifier entered the design document while an unrelated review note was being fixed. It survived to the point of merge and was caught by a reviewer reading the diff, not by any gate. Post-merge remediation on a public repo means rewriting published history.

Note the interaction with the usual verification: `gh pr diff` renders the **net** change, so a string added in one commit and removed in a later one passes every diff-based check while sitting intact in the branch. Three independent checkers passed that PR's served diff while the identifier was live in an earlier commit.

## Gap 3 — "remediated" is undefined

Two remediations were applied, and neither is complete:

| Action | What it fixes | What it leaves |
|---|---|---|
| Redact the file in a later commit | The merged content | The earlier commit, still served by the public API by SHA |
| Rewrite the branch, then squash | The branch, the Commits tab, `dev`'s ancestry | The orphaned commit, still served by SHA — and the SHA is published in the PR's force-push timeline event |

[...]

PR #1205, opened 2026-09-07
docs(#1138): technical design for portfolio multi-repo aggregation

me2resh:
## Summary

- **Authors the technical design for #1138** — `docs/technical-designs/portfolio-multi-repo-aggregation.md` — so the five portfolio-view skills (`/projects`, `/inbox`, `/tasks`, `/status`, `/handover`) aggregate across a multi-repo project's repos under one agreed set of semantics instead of five improvised ones. Rendering follows the operator's call (one row per project, per-repo detail only on exception, `--by-repo` drill-down); the ticket's per-field fold table stands unchanged.
- **Decides where `/handover`'s clones land (D6) — the largest decision in the diff, and the one the first draft left open.** A path is attributed to a project by its first segment under the workspace dir, an idiom shared by `require-active-ticket.sh`, `require-migration-ticket.sh`, and `briefing.sh`, while `/start-ticket` keys its marker on the registry name. That kills the originally-proposed `<name>--<repo-name>` sibling layout outright: both gates would resolve a project with no marker and fall through to the ops-level `current-ticket`, so a ticket set for unrelated framework work would *authorise* an edit inside a managed repo. Nesting attributes correctly but blends a worker repo's files into the primary's harnessability scan. **Chosen: a container directory per project** — `<workspace_dir>/<name>/<repo-name>/` per repo, primary included, with both gates working untouched. Rerun, cleanup, and per-repo failure handling are specified; singular projects do not move.
- **Grounds the staleness threshold instead of implying it exists.** `/projects` has no numeric staleness contract today — it renders git's `%ar` string and flags rows by judgement, and the "30 days" in the skill is sample output. So `portfolio.stale_days` (default 30) is introduced explicitly, scoped to the per-repo exception rule at R ≥ 2 only.
- **Sources activity from the forge, not a local clone.** `git log -1` reports the newest commit as of the last fetch, and that lag is unbounded — an unfetched-but-busy primary would render as stale, naming the product's most active repo dormant. D2 rejects that same failure class for a TTL cache one section earlier, so at R ≥ 2 every repo's epoch comes from `pushedAt`, primary included. R = 1 computes no epoch at all.
- **Makes byte-identity derivable rather than asserted.** One rule gates *every* per-repo exception path on R ≥ 2 — unreachable, red CI, and the `· <R> repos` header suffix included — so § Backward Compatibility's global claim follows from D1 instead of standing beside it.
[...]

PR #1177, opened 2026-09-04
fix(#1152): close command validation parse bypasses

me2resh:
## Summary

- Extend the jq-failure fail-closed guard to command-validation hooks for staging, onboarding commits, branch names, and commit messages. Matching operations now block instead of silently bypassing validation.
- Reuse the narrow raw-payload matcher and add regression coverage for each trigger class while preserving unrelated-payload no-ops.

## Verification

- `bash .claude/hooks/tests/test_fail_closed_json.sh` — 10 passed, 0 failed.
- `bash -n` and `git diff --check` — passed.

## Scope

This is the second focused slice of #1152. It does not change repository resolution or the remaining content and tracker gates.

AgDR: docs/agdr/AgDR-0130-fail-closed-json-payloads.md

Refs #1152

## Glossary

| Term | Definition |
|------|------------|
| Fail closed | Block when a control cannot evaluate its precondition |
| Raw-payload check | A jq-free scan of the original hook input for a known trigger shape |
| Trigger class | A group of hooks activated by the same command family |

PR #1176, opened 2026-09-04
fix(#1152): fail closed on unreadable hook payloads

me2resh:
## Summary

- Fail closed when jq cannot parse payloads for four high-risk controls, including Bash file writes handled by the active-ticket gate.
- Keep unrelated and read-only payloads as no-ops, with jq-failure regression coverage for both paths.

## Verification

- `bash .claude/hooks/tests/test_fail_closed_json.sh` — 6 passed, 0 failed.
- `bash .claude/hooks/tests/test_require_active_ticket_bash.sh` — passed.
- `bash -n` and `git diff --check` — passed.

## Scope and follow-up

This is the first focused slice of #1152. Remaining blocking hooks require separate trigger-class slices.

AgDR: docs/agdr/AgDR-0130-fail-closed-json-payloads.md

Refs #1152

## Glossary

| Term | Definition |
|------|------------|
| Fail closed | Block when a control cannot evaluate its precondition |
| Raw-payload check | A jq-free scan of the original hook input for a known trigger shape |
| Trigger shape | The command or tool pattern that activates one hook |

PR #1210, opened 2026-09-07
fix(#1206): scan gh pr review and gh pr merge for leaked private refs

me2resh:
## Summary

- **Closed the wrapper gap (Hakim HIGH, blocking).** `tracker_review_submit` and `tracker_pr_merge` run their `gh` call inside a sourced shell function. No second command event fires for that inner call. This hook had zero matchers for either wrapper, or for `tracker_create`. Every review Rex, Hakim, and Tariq post uses `tracker_review_submit`. `code-reviewer.md` prescribes it over a hardcoded `gh pr review`. The canonical path was the unscanned one.
- **Widened the jq-outage fallback.** The degraded-path check named four of the ten shapes this hook now covers. A broken jq silently passed the other six, including both new wrappers.
- **Closed the equals-form gap.** `--title=x`, `--subject=x`, `--body=x`, and `--body-file=x` matched no extractor. The hook now refuses instead of scanning nothing.
- **Fixed a message-attribution bug, not a security bug.** A chained `gh pr merge` command can trip the truncation refusal through its own `--body`, with no `--subject` anywhere in the command. Verified directly against `dev`: the same command with `--subject` removed still diverges the same way, because `dev` never scanned `gh pr merge` at all. The old message always named `--subject`. The new message names only the flag that actually failed.
- **Removed two new real-project mentions from this PR's own tests.** Reused an invented fixture name instead. Verified it is absent from the real portfolio registry.
- **Fixed the runbook.** Step 4 checked the private portfolio registry for GitHub forks of a public repo. It now points at `gh api repos/<owner>/<repo>/forks`. The "every action reduces exposure, exactly one ends it" line overclaimed. GitHub Support cannot reach a fork either. The runbook now says so.
- **Moved the `gh`-only scope claim into `leak-protection.md`.** Documented that `_tracker_review_glab` has no scan path at all for a GitLab adopter's review body.
- **Merged `upstream/dev`** to pick up the controlled technical writing profile, then rewrote this PR's own runbook text under it.

## Testing

1. `bash .claude/hooks/tests/test_block_private_refs.sh` -- 105/105 pass. 30 new cases cover the three wrapper shapes, the equals form, and the attribution fix.
2. `bash .claude/hooks/tests/test_fail_closed_json.sh` -- 15/15 pass. 5 new cases cover the widened degraded-path regex, gh-shape and wrapper-shape alike.
3. `bash bin/run-hook-tests.sh` -- 145/147 pass. The 2 failures are pre-existing and host-specific, unrelated to this change.
4. `shellcheck` on all three changed shell files -- identical findings to the unmodified originals. No new findings.
5. `git diff --check` -- clean.
6. Ran a leak check on this diff's added lines against the real portfolio registry before each push. No new real-project mention survives in the final diff.
7. Re-verified the GitHub Support documentation quote in the runbook against the live page immediately before this push.

Refs #1206

This is PR A of two on this ticket. PR B (gap 2, a private reference committed directly to a file) follows separately. It starts from its own AgDR, because it needs a new enforcement point, not an extension of this hook.

---

## Glossary

| Term | Definition |
|------|------------|
| Wrapper shape | A tracker-agnostic function call (`tracker_create`, `tracker_review_submit`, `tracker_pr_merge`) that runs its `gh` command inside a sourced shell function, invisible to a hook that only sees the outer command text. |
| Write segment | The part of a Bash command starting at its recognized write invocation. Everything before that point is out of scope for repo matching. |
| Truncation refusal | The hook's refusal to scan a value when it cannot tell where a quoted flag value safely ends, for example a command chained right after it. |
| Orphaned commit | A commit no branch points at. GitHub still serves it by SHA. |


atlas-apex (approved):

me2resh (commented):
## Security Review: PR #1210

**Commit**: `2d72f33deb85dd7fd998c8c51f5a16074d71b0b4`
**Scope**: Heavy — trust chain (`.claude/hooks/*.sh`, `.claude/settings.json`)

### Verdict

**CHANGES REQUESTED** — and specifically: **not safe to merge as written**, though the *code* is fine and worth keeping.

The shell change is a strict improvement. I found no case where this PR makes a leak more likely. But the completeness claim it writes into `.claude/rules/leak-protection.md` is false in the one place that matters most, and merging it would convert an unknown gap into a gap the framework's own security documentation asserts is closed. That is the #1206 failure pattern repeating one level up.

The smallest safe merge is either fix H1, or correct the rule file and PR body to say plainly that the canonical path is still unguarded — and file it. Do not merge with the current wording.

There is no mechanical security merge gate on this repo, so this comment is the whole control. **I have not written a security marker**, per the clean-verdict rule.

---

### What the change gets right

I verified each of these by running the hook at this commit against a fixture registry, A/B against the `dev` blob (`6dc663a`). Rc 2 = blocked, rc 0 = allowed.

| Behaviour | dev | PR |
|---|---|---|
| `pr review --body` / `-b` carrying a registered name | 0 | **2** |
| `pr review --body-file <path>` (the shape the gh adapter emits) | 0 | **2** |
| `pr merge --subject` / `-t` / `--body` / `--body-file` | 0 | **2** |
| `pr merge` with no content flags (the routine merge) | 0 | 0 |
| Clean review body | 0 | 0 |
| Skip marker in the body bypasses on both new shapes | — | 0 |
| Skip marker smuggled in a later `--label` does **not** bypass | — | **2** |
| `--body-file -` (stdin) and unreadable body-file on the new shapes | — | **2** (fail closed) |
| Heredoc and multi-line bodies | — | **2** (scanned whole) |

Four things I want to credit specifically, because they are the parts that are easy to get wrong:

- **Reusing `--title|-t` for `--subject` really does inherit the #1068 truncation refusal.** Test 59 is not decorative; I reproduced the refusal independently. A parallel extractor would have re-opened that hole for one flag.
- **The strict/greedy haystack asymmetry survives the new shapes.** The marker-in-`--label` case still blocks. That property is the subtle one and it held.
- **The `settings.json` wiring is not prefix-anchored.** I observed `block-unreviewed-merge.sh` fire on one of my own probe commands where the merge string appeared mid-line inside a heredoc. So the new matchers will fire on chained forms (`cd x && gh pr review …`), which I also confirmed the hook itself handles. Good.
- **87/87 tests pass.** The new cases are well-chosen for what they cover.

---

## Findings

### HIGH — H1. The two paths this PR exists to protect are still unscanned

This is the finding that decides the verdict.

The PR summary says: *"`tracker_review_submit` is how Rex, Hakim, and Tariq post every review, so this was the most-exercised unguarded path in the framework."* That is the right risk. But the hook sees the **outer Bash command text**, and on that path the outer command is not a `gh` command at all.

Every review agent is instructed to post through the wrapper, explicitly *not* the CLI. From `.claude/agents/code-reviewer.md:33` (Hakim and Tariq carry the same instruction):

```
tracker_review_submit "$PR_HOST_REPO" {number} comment "$REVIEW_BODY_FILE"
```

And `/approve-merge` states the merge case outright, at `SKILL.md:306`:

> the merge-gate hooks … match the OUTER Bash command text this step actually submits … and **that text never literally contains `gh pr merge`** — those strings live inside already-sourced library code.

The `gh` invocation happens inside a sourced shell function, in the same process. No second `PreToolUse` event fires.

Measured at this commit, with a body-file containing a registered project name:

| Outer command actually submitted | rc |
|---|---|
| `tracker_review_submit "me2resh/apexyard" 12 comment "<leaky file>"` | **0 — unscanned** |
| `tracker_pr_merge "me2resh/apexyard" 12 squash true "<leaky subject>" "<leaky file>"` | **0 — unscanned** |

Three independent reasons it cannot fire, all of which need fixing together:

1. Step 1 shape detection keys on `\bgh\s+pr\s+(review|merge)\b`.
2. `find_write_segment`'s anchor keys on the same literal.
3. Even if both matched, the wrapper passes the repo and the body as **positional arguments**, so neither the `--repo`/`-R` check nor `extract_path_flag '--body-file'` would resolve anything.

`settings.json` confirms the asymmetry. The four merge gates each carry a `Bash(tracker_pr_merge *)` matcher. The leak hook does not, and **no matcher for `tracker_review_submit` exists anywhere in the file.**

The precedent for the fix already exists and is cheap: `_lib-extract-pr.sh` grew a `tracker_pr_merge <owner/repo> <pr> …` branch in #759 for exactly this reason, with the matcher entries added alongside. This PR needs the same two moves plus positional extraction.

**Why this blocks rather than defers.** The diff adds to `.claude/rules/leak-protection.md`:

> Wired to `PreToolUse` on `Bash` for seven command shapes … `tracker_review_submit` posts every code, security, and design review through `gh pr review` … Neither shape was matched before #1206 … so a private reference posted through either one reached a public repo with nothing scanning it.

Read in the past tense, that tells a future maintainer the review path is now scanned. It is not. The section even carries an explicit **"Known gap, not yet closed here"** callout for gap 2 — the honesty is already there, it just names the wrong gap. `AgDR-0113` is the rule this file wrote for itself after the last time a claim was reasoned from code shape instead of measured, and it applies here.

I will note the irony plainly because it is evidence, not commentary: **this review is being posted through `tracker_review_submit`, and the hook this PR modifies will not run on it.**

---

### MEDIUM — M2. The fail-closed fallback was not extended with the scan set (#1152 class)

You asked me to check whether a new instance of "blocking hook proceeds when it cannot evaluate its precondition" was added. One was.

Line 44 still reads:

```bash
raw_payload_command_matches "$INPUT" 'gh[[:space:]]+(issue|pr)[[:space:]]+(create|comment)'
```

The scan set went from five shapes to seven; this pattern did not move. Measured with `jq` stubbed to exit 127, same leak in every body:

| Command | rc |
|---|---|
| `gh issue create … --body '<leak>'` | 2 — fails closed |
| `gh pr comment … --body '<leak>'` | 2 — fails closed |
| `gh pr review … --body '<leak>'` | **0 — silent pass** |
| `gh pr merge … --subject '<leak>'` | **0 — silent pass** |
| `gh api …/pulls/12/reviews -f body='<leak>'` | **0 — silent pass** |

The `gh api` row is pre-existing, so the pattern was already incomplete (2 of 5). This PR widens that to 2 of 7 rather than closing it. One-line fix: add `|review|merge` to the subcommand alternation and an `api` arm.

Low likelihood (needs `jq` broken), but this is precisely the class #1152 tracks and two commits landed on it today.

---

### MEDIUM — M3. The `=`-joined flag form is a silent fail-open, including on the new shapes

The hook has a deliberate, well-argued fail-closed posture for a body it cannot read (#1039: *"a future parsing gap blocks loudly instead of leaking"*). The `=` spelling defeats it by producing "no body-file at all" rather than "unreadable body-file", so control reaches the empty-haystack short-circuit and exits 0 with nothing scanned and nothing printed.

Measured, all with a space-form `--repo` so the target resolves correctly:

| Command | rc |
|---|---|
| `gh pr review … --body-file /tmp/leak.md` | 2 — blocked |
| `gh pr review … --body-file=/tmp/leak.md` | **0 — silent pass** |
| `gh pr review … --body='<leak>'` | **0** |
| `gh pr merge … --subject='<leak>'` / `--body-file=…` | **0** |
| `gh issue create … --body-file=/tmp/leak.md` | **0** (pre-existing) |

Two things make this worse than a routine parsing gap. The hook's own diagnostic tells operators *"The reliable placement is `--body-file`: it is read whole, with none of the restrictions below. Prefer it."* — steering people toward the flag whose `=` spelling is unguarded. And #1070 already handled the `=` form for `--input` (`--input([[:space:]]|=)`), so the pattern is known to the file; it just was not applied to `--body-file` / `--body` / `--title` / `--subject`.

Pre-existing for the original five, so not a regression — but the PR's new coverage of the review path is one character away from being bypassed, and the failure is silent rather than fail-closed, which is the property #1039 paid for.

---

### MEDIUM — M4. A new false positive on reviewer prose, on an already-matched shape

You flagged this cost specifically, so I tested for it. Folding `--subject` into the `TITLE` pattern widens the existing prose-collision to a word reviewers of *this PR* are very likely to type.

Clean A/B on a shape that was already matched on `dev`:

```
gh pr comment 5 --repo me2resh/apexyard --body 'the flag is --subject "value" here'
   dev = 0 (allowed)     PR = 2 (BLOCKED)
```

Mechanism: `extract_flag_value` scans the whole command, finds `--subject` inside the body prose, both quoted branches fail to close, the unquoted fallback returns a value starting with a raw `"`, and the truncation refusal fires. `--title` already did this on `dev`, so the class is pre-existing; this PR extends it to `--subject` and `-t`.

Bounding it fairly:

- Only inline `--body`/`-b` is affected. A body passed via `--body-file` — the canonical path — cannot trigger it, because the prose never enters the command line.
- It fails **closed**, and the operator can recover by switching to `--body-file`.
- The unquoted spelling (`--subject into the TITLE pattern`, no quotes after) passes fine.

What I would change regardless of the rest: the block message says *"The most likely cause: a second command chained after the write with `&&` / `;` / `|`, or a trailing shell comment."* For this trigger that diagnosis is wrong, and it sends the operator hunting for a chain that is not there. One added line naming flag-shaped text inside a quoted body, and pointing at `--body-file`, would cost nothing.

---

### LOW — L5. `gh api …/pulls/N/merge` reaches the same endpoint unscanned

The hook runs on this (`IS_GH_API=1`, target resolves) and then extracts nothing, because it looks for `title=` and `body=` while the merge endpoint uses `commit_title=` and `commit_message=`. Empty haystack, exit 0, silent.

```
gh api repos/me2resh/apexyard/pulls/12/merge -X PUT -f commit_title='<leak>'      → rc 0
gh api repos/me2resh/apexyard/pulls/12/merge -X PUT -f commit_message='<leak>'    → rc 0
```

Worth naming rather than shrugging at, because `.claude/rules/pr-workflow.md` § "Both merge shapes are gated (#47)" documents the api merge form as a **first-class, supported, gated** merge shape. After this PR the two documented merge shapes are asymmetric: both hit the merge gates, only one hits leak protection. Pre-existing; adding `commit_title`/`commit_message` to the api field extraction is small.

### LOW — L6. `gh issue edit` / `gh pr edit --body` are still unmatched

Same publishing surface as `create` — `--body` *replaces* the rendered issue or PR body on a public repo — and neither is in the seven. Measured: `gh issue edit 42 --repo me2resh/apexyard --body '<leak>'` → rc 0. This is live in the framework: `.claude/skills/plan-initiative/SKILL.md:487` runs `gh issue edit "$issue_number" --repo "$repo" --body "$new_body"`, and `require-migration-ticket.sh` and `/split-portfolio` both instruct operators to use it.

Not this PR's job to fix, but the PR's framing is "the sweep is exhaustive for `gh`", and this is the clearest counter-example after H1. Recommend adding both to the same follow-up.

*(`gh release create --notes` is also unmatched, but I checked and it runs only in `auto-tag-on-release-pr-merge.yml` — CI, where a `PreToolUse` hook can never see it. Genuinely out of scope, not an oversight.)*

### LOW — L7. The runbook slightly overstates, and step 4 points at the wrong place

You asked me to check that it does not overstate. Two small things, both fixable in a sentence:

1. The headline says **"every self-service action below reduces exposure. Exactly one ends it."** Its own table two lines later says the GitHub Support purge leaves *"copies already fetched by mirrors, clones, or scrapers before the purge; content already living in a fork"*, and the paragraph after says Support *"does not reach a fork at all."* So nothing on the list ends it unconditionally. Suggest: *"Exactly one can end it on the repository itself, and only when no fork, mirror, or scrape already holds a copy."*
2. Step 4 says *"Check `apexyard.projects.yaml` for a registered fork of the leaking repo."* The leaking repo here is the **public framework repo**; its forks are third-party and will never appear in a private portfolio registry. The check that would actually find them is the forks endpoint on that repo. As written, step 4 is a near-certain no-op that reads like due diligence was done.

Everything else in the runbook is accurate and well-sourced. The measured `48,174 bytes / 2 occurrences` result, the ordering (Support request first), the GitHub-docs citation, and step 5's insistence that status reads *reduced exposure* and not *removed* are all correct and are the important parts. This is a genuinely good addition; these are two edits, not a rewrite.

---

### Answering your framing question on the merge path

You asked whether a body reaching a **protected branch's commit message** deserves stricter treatment than an issue comment. Yes, and the runbook does not currently reflect it.

Since #1192, a release-class PR carries `--subject` and a `--body-file` into a permanent squash commit on `main`. The runbook's remediation table has rows for a comment/review/PR body (editable via the API), a file in a commit (redactable in a later commit), and branch history (rewritable). **A merge-commit message on `main` matches none of them.** It is not a file, so you cannot redact it in a follow-up commit; it is not API-editable; and the only remaining action is rewriting `main` on a public repo, which the runbook itself argues never fully removes anything.

So the merge path is simultaneously the surface where prevention matters most and the one where remediation is weakest — and it is also the surface still unscanned per H1, because releases merge through `tracker_pr_merge`. Worth one row in the table and one sentence in the runbook.

---

### What I could not verify

- **Whether the #1205 incident itself went through the wrapper or a hand-rolled CLI call.** I did not inspect that PR's review provenance. If it was hand-rolled, this PR does close that specific instance; the canonical path is still open either way.
- **Runtime matcher semantics.** My conclusion that `settings.json` conditions are not prefix-anchored rests on one observation in this harness version (a merge gate firing on a mid-line occurrence in my own probe). It is consistent with the wiring working for chained commands, but I did not read the matcher implementation.
- **Non-`gh` adapters.** I accepted the "`gh`-only by design" scoping and did not audit `glab` or `custom`. Flagging one thing for whoever picks that up: `_tracker_review_glab` reads the body file into a shell variable and passes it via `glab mr note create -m "$body"`, so a GitLab adopter's review body never passes through any scanning path at all. That is a scoping decision to make explicitly rather than inherit.
- **Live behaviour.** Everything above is the hook executed against a synthetic registry in a sandbox. I did not exercise any real tracker write.

---

### Recommended path

1. **H1** — add the wrapper shapes (detection, anchor, positional repo/body extraction) plus `Bash(tracker_review_submit *)` and `Bash(tracker_pr_merge *)` matchers, following the #759 precedent. If that is too large for this PR, then instead correct the rule file and PR body to state that the canonical wrapper path remains unguarded, and file it as a named gap alongside gap 2. **One of these two is required to merge.**
2. **M2** — one-line pattern extension. Cheap, and closes a live #1152 instance.
3. **M3** — accept `=` alongside whitespace in the content and path extractors, so the `=` form reaches the existing fail-closed branch instead of the silent one.
4. **M4** — add the prose case to the truncation-refusal message. Non-blocking.
5. **L5 / L6** — one follow-up ticket covering the api merge fields and `issue edit` / `pr edit`.
6. **L7** — two sentence edits, plus a merge-commit row in the remediation table.

Findings 2 through 6 are all defensible as follow-ups. Finding 1 is not, because it is the difference between an incomplete control and a control the documentation says is complete.

---

📌 Reviewed commit: `2d72f33deb85dd7fd998c8c51f5a16074d71b0b4`
🔒 No security marker written — verdict is not clean.

me2resh (commented):
## Code Review: PR #1210

**Commit**: `2d72f33deb85dd7fd998c8c51f5a16074d71b0b4`
**Scope**: Full — Heavy tier. The diff touches `.claude/hooks/*.sh` and `.claude/settings.json`, so rail 1 of `right-size-ceremony.md` forbids a reduced-scope pass regardless of diff size.

### Verdict

**CHANGES REQUESTED** — two blocking findings, both small.

Your next action: extend the unparseable-payload regex to cover the two new shapes (one word), and apply the wrapper-matcher pattern #759 already established — or, if you'd rather defer that, narrow the three claims that say `tracker_review_submit`'s reviews are now scanned. Everything else here is sound, and I verified it rather than taking it.

### Summary

This adds `gh pr review` and `gh pr merge` to the leak hook's matched shapes, wires both into `settings.json`, folds `--subject` into the existing title extraction, and gives `leak-protection.md` the remediation runbook it was missing. The direct-command coverage genuinely works — a leaked identifier in a `gh pr review --body-file` now blocks where it previously passed unscanned. You were right to add the merge half; it carries a reviewed body into a permanent squash commit, which is worse than a review comment because it cannot be edited away afterwards.

The two blocking findings are that the new shapes fail **open** where the older ones fail closed, and that they cover only the hand-typed form of two commands whose wrapper form this repo already decided must be matched.

### Checklist Results

- Architecture & Design: Pass
- Code Quality: Pass
- Testing: Pass — 87/0 reproduced; but see H1, the new shapes sit outside the fail-closed test
- Security: **Fail** — H1, H2
- Performance: Pass
- Technical Decisions (AgDR): N/A — reasoning below, and I agree with yours
- Adopter Handbooks: N/A — no `handbooks/` tree in this fork

---

## H1 (blocking) — both new shapes fail OPEN on an unparseable payload

Line 43 of the hook is unchanged by this PR:

```bash
if raw_payload_command_matches "$INPUT" 'gh[[:space:]]+(issue|pr)[[:space:]]+(create|comment)'; then
```

The PR raises the covered-shape count from five to seven in the header, in `leak-protection.md`, and in the step-1 detection — but leaves this degraded-path trigger matching four. When `jq` cannot parse the payload, `COMMAND` is empty and this regex is the only thing between an unscanned public write and `exit 0`.

I ran your own #1152 regression idiom against it. `test_fail_closed_json.sh` stubs `jq` to `exit 127` and asserts `rc=2`:

```
rc=2  FAILS CLOSED   gh issue create   (pinned by #1152)
rc=2  FAILS CLOSED   gh pr create
rc=2  FAILS CLOSED   gh issue comment
rc=2  FAILS CLOSED   gh pr comment
rc=0  FAILS OPEN <-- gh api        (pre-existing gap, not yours)
rc=0  FAILS OPEN <-- gh pr review   (NEW in this PR)
rc=0  FAILS OPEN <-- gh pr merge    (NEW in this PR)
```

`gh api` was already outside the trigger, so that one is pre-existing. `review` and `merge` are shapes this PR newly claims to cover, so these are two new instances of the pattern #1152 owns — added the same day #1176 and #1177 landed to close others.

The asymmetry is what makes this worth blocking rather than deferring. A reader of the new "seven shapes" text has no way to know that three of the seven behave differently when `jq` is degraded, and the two most security-exposed shapes are among the three.

**Fix:**

```bash
'gh[[:space:]]+(issue|pr)[[:space:]]+(create|comment|review|merge)'
```

Refusing on a broken `jq` is the correct trade and is already what `gh issue create` does — the raw path cannot check the target repo, so it refuses on shape alone and tells the operator to restore `jq`. Please also add the two cases to `test_fail_closed_json.sh` beside the existing `private tracker write` case, so the invariant is pinned rather than rediscovered later.

## H2 (blocking) — only the hand-typed form is covered, and #759 already settled that this is not enough

`_tracker_review_gh`, `_tracker_merge_gh`, and `_tracker_create_gh` all build an args array and invoke `gh "${args[@]}"`. The literal string `gh pr review` therefore never appears in any Bash command text. A `PreToolUse` hook sees the command the agent submitted — `tracker_review_submit "$REPO" 1210 comment "$FILE"` — and neither the new `Bash(gh pr review *)` matcher nor the step-1 grep matches that.

Measured against a fixture registry:

```
Direct shape (what this PR adds)
  rc=2  gh pr review --body-file review.md   | BLOCKED: private project reference detected

Wrapper shape (what .claude/agents/code-reviewer.md prescribes)
  rc=0  tracker_review_submit ... review.md                    | PASSES UNSCANNED
  rc=0  tracker_pr_merge ... subject + body               | PASSES UNSCANNED
  rc=0  tracker_create ... (issue-create wrapper)| PASSES UNSCANNED
```

I posted this very review through the wrapper, unscanned.

What makes this blocking rather than a documentation nit is that **this repo already litigated exactly this gap and fixed it**. `_lib-extract-pr.sh:63` records it:

> THE tracker_pr_merge WRAPPER SHAPE (#759, HIGH finding from Hakim's review …)

The #759 fix was two-part: add `Bash(tracker_pr_merge *)` matcher entries, and teach the hook to read the wrapper's positional args. All four merge gates carry that matcher today:

```
Bash(tracker_pr_merge *)  ->  block-unreviewed-merge.sh
Bash(tracker_pr_merge *)  ->  block-merge-on-red-ci.sh
Bash(tracker_pr_merge *)  ->  require-architecture-review.sh
Bash(tracker_pr_merge *)  ->  require-design-review-for-ui.sh
```

`block-private-refs-in-public-repos.sh` does not — and this PR just added `gh pr merge` to it. So on the identical command, the leak hook now guards the direct form while its four sibling gates guard both forms. No hook anywhere matches `tracker_review_submit` or `tracker_create`.

I found this the hard way: three of my read-only review probes were blocked by the merge gate, one of them purely because the text `tracker_pr_merge` appeared in a `printf` string. The gate that is not supposed to see my probe saw it; the gate that is supposed to see real review bodies does not.

To be clear about what is and isn't yours: the wrapper blind spot is **pre-existing and uniform**, coverage strictly increased in this PR, and nothing got worse. But #759 establishes that "the wrapper is the shape the framework actually runs" is a known HIGH finding with a known fix, and this PR extends a hook to two commands without applying it.

**Fix — either is acceptable:**

1. *Preferred:* follow #759. Add `Bash(tracker_pr_merge *)` and `Bash(tracker_review_submit *)` matchers, and extend step 1 plus the flag extraction to read the wrapper's positional form (`tracker_review_submit <repo> <pr> <verdict> <body_file>`). Note the matcher alone is not enough — the step-1 grep still needs the wrapper alternation, or the hook runs and exits 0.
2. *If you'd rather keep this PR narrow:* say plainly in `leak-protection.md`, the hook header, and the PR body that only the directly-invoked form is scanned, record the wrapper gap beside the file-content gap you already document, and file it referencing #759. That is honest and leaves the follow-up findable.

What I do not think is acceptable is the current combination: shipping direct-form-only coverage while three artefacts state that `tracker_review_submit` — the form the reviewer spec prescribes — is now guarded. Ticket AC 1 asks for "any other shape `tracker_review_submit` can emit", and the wrapper is that shape.

I am also holding this to the standard the hook header sets for itself:

> every "pre-existing" claim in this file must cite the check that proved it against the actual dev blob, not be reasoned from the shape of the code (AgDR-0113)

The coverage claim was reasoned from the shape of the code. One end-to-end run through `tracker_review_submit` would have caught it.

## M1 (non-blocking, but please act on it) — the fixtures use real registered private project names

The test file's fixture registry is built from names that are actually registered private projects in the portfolio, not invented ones. Three registered project names appear in the fixture file; the primary one occurs 51 times on `dev` already, and **this PR adds 15 more occurrences** to a public repo.

You disclosed this accurately in the testing notes as the test file's "pre-existing fixture-name convention (already public on `dev` before this PR)", and that is fair — you did not introduce it, and 15 more occurrences of an identifier already published 51 times is close to zero marginal disclosure. That is why this is not blocking.

It still deserves naming, because it is exactly the class of leak this ticket exists to prevent, sitting inside the PR that fixes it — and by your own new runbook those identifiers are permanently in the "reduced exposure, not removed" category. The fixture already mixes obviously-invented slugs alongside the real ones, so the convention is inconsistent rather than deliberate. Please switch the fixtures to invented names in PR B or a follow-up: it is a mechanical find-and-replace, the assertions do not care what the strings are, and it removes a live example of the wrong pattern from the file most likely to be read as the reference for how to test this hook.

I have not repeated any of the identifiers in this review.

---

## Suggestions (non-blocking)

**S1 — the `gh`-only scope is accurate, but stated nowhere durable.** I verified the claim: the upstream detection's `sed` only matches `github\.com`, the shipped default is a GitHub slug, and all seven matchers are `Bash(gh …)`. So `glab` and `custom` really are out of scope by construction rather than overlooked. But neither `leak-protection.md` nor the hook header says so — it lives only in the PR body, which nobody reads afterwards. An adopter on `tracker.kind=glab` reading the rule file would reasonably conclude they are guarded, and would be wrong silently. One sentence in `leak-protection.md` § "When the hook fires" fixes it, and that is the right home: it is the section that enumerates coverage.

**S2 — "Exactly one ends it" is ambiguous, and stronger than your own later text.** The sentence reads: "every **self-service** action below reduces exposure. Exactly one ends it." The first clause is scoped to self-service actions and the second to the whole table, so a reader mid-incident may go hunting for the self-service action that ends exposure. There isn't one — that is the entire point of the section. And your own paragraph below says a Support purge "does not reach a fork at all", so even that does not strictly end it when a fork exists. Suggested:

> Every action below reduces exposure. Only a GitHub Support purge removes the object itself, and even that does not reach forks.

The rest of the runbook is good and I would keep it. Ordering the Support request first is right (longest lead time), step 5's "must read reduced exposure, must not read removed or fixed" is exactly the sentence the ticket asked for, and it is findable — the heading greps on "leak", "remediation", or "exposure", and the file is already cross-referenced from `/report-apexyard-bug` and `/request-apexyard-feature`, the skills someone is most likely mid-flight in when this happens.

**S3 — the false-positive surface widens, though the mechanism is pre-existing.** Both new matchers fire on any command that merely *contains* the phrase, not only ones that start with it. In fairness I tested whether this PR adds a new *mechanism*, and it does not — a local heredoc quoting `gh issue create --repo <public> --title "<private-name>"` blocks identically on `dev` today. What it adds is two more phrases to the surface, and those two appear constantly in this framework's own rules, skills, and tests, so practical frequency rises more than the two-shape count suggests. That belongs to #1145, not here.

One test gap though: case 52 only covers the inverse (prose quoting `gh pr review` *inside* a real write's body). Consider adding a case for a **non-`gh`** command that quotes a merge or review invocation — a heredoc or `echo` writing documentation — so the intended behaviour for the new shapes is pinned before #1145 changes anything.

---

## What I verified rather than took

- **The `-t` / `--subject` claim is correct.** `gh help pr merge` lists `-t, --subject text`, so `-t` really was riding in on the pre-existing `--title|-t` pattern and only the long form was missing. `-b/--body` and `-F/--body-file` are genuinely shared across `pr merge`, `pr review`, and the create/comment shapes. `gh pr review` has no title or subject flag, so the `TITLE` extraction is a harmless no-op there. Folding `--subject` into the shared pattern to inherit the #1068 truncation check is the right call, and case 59 proves the inheritance instead of asserting it.
- **The `find_write_segment` anchor** is correct: the `pr[[:space:]]+(create|comment|review|merge)` alternation preserves the existing `([[:space:]]|$)` boundary, so a longer token cannot anchor. The `-R` short form of `--repo` is already handled, so both new shapes inherit target detection.
- **Settings wiring:** JSON valid; exactly seven matchers point at this hook, matching the seven claimed shapes; the merge matcher now fans out to five hooks, consistent with how the merge gates are wired.
- **Tests:** `test_block_private_refs.sh` 87/0 reproduced, all 74 pre-existing cases still passing.
- **Full suite:** 145/147, failing exactly `test_lib_self_location_cwd_anchor.sh` and `test_validate_search_config.sh` — the two host-specific failures you named, which also fail on `dev`.
- **shellcheck:** identical to the `dev` blob — the same two notes (SC1091, SC2329), line numbers shifted only.
- **AgDR is N/A, and I agree with your reasoning.** `.claude/hooks/**` is not in `agdr_trigger_paths`, and the two comparable trust-chain fixes merged today (#1176, #1177 under #1152) shipped without AgDRs. Extending an existing hook's shape list implements a decision already recorded; PR B's new enforcement point genuinely needs its own. Rail 1 of `agdr-decisions.md` gave me pause since this is the trust chain, but the precedent is consistent and the decision is not a new one.
- **PR hygiene:** `Refs #1206` present; Glossary present with three real terms; all four Summary bullets narrative (verb plus consequence), none label-only; the PR B note is present and explains *why* the split, not just that there is one.

## One correction for the record

The review brief gave HEAD as `2d72f337…`. The actual HEAD is `2d72f33d…` — they diverge at the eighth character. I reviewed `2d72f33deb85dd7fd998c8c51f5a16074d71b0b4`, which is what the forge reports and what the worktree is checked out at. Worth knowing if that prefix was copied anywhere else.

## What I could not verify

- The GitHub documentation quote in the remediation section — no network access in this context. It is load-bearing for the section's central claim, so it is worth one confirmation against the live page before merge.
- Whether every reviewer path uses the wrapper. I confirmed the reviewer agent spec prescribes it and that the wrapper is unscanned, but I did not audit all three review skills for a direct-`gh` fallback. Any that shells out directly *is* covered by this PR.

---

No approval marker written — the verdict is CHANGES REQUESTED.

Re-request review once H1 and H2 are addressed. Both are small, and I do not expect a second round on the rest.

---
📌 Reviewed commit: `2d72f33deb85dd7fd998c8c51f5a16074d71b0b4`

me2resh (commented):
## Security Review (round 2): PR #1210

**Commit**: `2e8b787689817cdc12b8eb1070b0f65331db9a61`
**Scope**: Heavy — trust chain

### Verdict

**CHANGES REQUESTED** — one blocking item, and it is a false positive, not a leak.

The security direction is materially better than round 1. H1 is genuinely fixed for every shape I could construct with literal arguments, the degraded path went from 2-of-7 to 9-of-10, and the `=` refusal works. I found **no new way for content to reach a public repo** that round 1 did not already have.

What blocks is the opposite failure: **the new wrapper detection blocks a reviewer who merely mentions a wrapper name in prose.** You asked me to check exactly this shape because it bit you today. It does bite. It would have blocked my own round-1 review, and it will block the next person who reviews this PR.

No security marker written — verdict is not clean.

---

### 1. Wrapper coverage — sound for literal arguments

Verified at this commit against a fixture registry. Every one of these blocks correctly:

| Wrapper form | rc |
|---|---|
| `tracker_review_submit "me2resh/apexyard" 12 comment "<leaky file>"` | 2 |
| `tracker_pr_merge "me2resh/apexyard" 12 squash true "<leaky subject>" ""` | 2 |
| `tracker_pr_merge "me2resh/apexyard" 12 squash true "subj" "<leaky file>"` | 2 |
| `tracker_create "me2resh/apexyard" "<leaky title>" "" ""` | 2 |
| bare / unquoted args | 2 |
| single-quoted args | 2 |
| `OUT=$(tracker_review_submit …)` — the `/approve-merge` capture form | 2 |
| trailing `> /tmp/out.txt` redirect | 2 |
| trailing `; rc=$?` | 2 |
| non-public target (`acme/…`) | 0 — correct no-op |

The implementation choices are right, and I want to name them because they are the reason the above works without new hardening code: reusing `_extract_wrapper_arg` rather than writing a second tokenizer; giving the wrappers their own top-level anchor alternative rather than bolting them onto the `gh` branch; and folding the positional title and body-file into the existing `TITLE` / `BODY_FILE` variables so the `#1068` truncation refusal and the `#1039` unreadable-file refusal apply to wrapper content for free. That is the `#759` pattern applied correctly.

**Sourcing `_lib-extract-pr.sh` mid-hook is safe as far as I checked** — it defines functions and, at line 170, conditionally sources `_lib-tracker.sh`; I found no top-level statement that mutates state this hook depends on. The fail-open-on-missing-library choice is documented and I agree with it: a missing core library should not block every wrapper call.

---

### 2. BLOCKING — the wrapper detection fires on prose, and blocks reviewers

`IS_WRAPPER` is set when a wrapper name appears **anywhere** in the command, including inside an already-correctly-parsed quoted `--body`. The positional body-file fold-in then runs, reads argument 3/4/6 out of the prose, gets a token that is not a file, and hits the fail-closed unreadable-body-file branch.

Measured — all three are ordinary review or comment posts with no wrapper call in them at all:

```
gh pr comment 5 --repo me2resh/apexyard --body 'run tracker_review_submit repo 1 comment /tmp/notes.md'
   → rc 2  BLOCKED: body-file named but not readable

gh pr comment 5 --repo me2resh/apexyard --body 'tracker_create arg3 is the body file'
   → rc 2  BLOCKED: body-file named but not readable

gh pr review 9 --repo me2resh/apexyard --comment --body 'the canonical call is tracker_review_submit $PR_HOST_REPO 12 comment $F'
   → rc 2  BLOCKED: body-file named but not readable
```

The trigger is simply: a wrapper name followed by three or more tokens, inside any write to a public repo. That is not an exotic shape — it is what a reviewer, a doc author, or a changelog entry writes about this very feature.

Three things make it worse than an ordinary over-block:

- **The message is wrong.** It reports a body-file path that does not exist in the command, and suggests "the path does not exist, or is relative to a different directory / a typo in the path". There is no `--body-file` to fix.
- **The skip marker explicitly does not apply** on that branch, by design. So the operator's only recoveries are to reword their prose or switch to `--body-file`.
- **It creates pressure on the control itself.** A guard that blocks people discussing it is a guard people route around — the same second-order reasoning AgDR-0111 used when it returned the marker gate to advisory.

**The fix looks small and does not weaken anything.** `find_write_segment` already answers the question that matters: it returns the segment starting at the *leftmost recognised invocation*. When the command is `gh pr comment … --body '…tracker_create…'`, that segment starts at `gh pr comment`, not at the wrapper. So gating the three wrapper fold-ins (and the `WRAPPER_REPO` resolution) on `WRITE_SEGMENT` actually *beginning* with a wrapper name distinguishes "the invocation is a wrapper" from "a wrapper name is mentioned", with no loss of coverage on any of the ten cases in section 1. I would want a test pinning the FP7 shape above.

### 3. MEDIUM — the `=`-form refusal fires on prose too, same class

```
gh pr comment 5 --repo me2resh/apexyard --body 'avoid the --body= equals form'      → rc 2
gh pr comment 5 --repo me2resh/apexyard --body 'use --body-file=x not the space form' → rc 2
```

The check greps `WRITE_SEGMENT` for `--(title|subject|body|body-file)=`, which finds the string inside a body that was already extracted cleanly and scanned in full. Nothing went unread, so refusing is not buying safety here.

Narrow fix in the same spirit as the invariant reasoning already in this file: skip the refusal when the `--flag=` occurrence lies **inside** a successfully-extracted `TITLE` / `BODY` value. In both cases above `BODY` parsed to completion, so the hook can prove the occurrence was content, not an unparsed flag.

Separately, on your question of over-refusal in the intended sense: refusing a **clean** body sent via `--body-file=` is correct and I would keep it. That one is a real unread value, the message is accurate, and the fix is one keystroke.

### 4. MEDIUM — the canonical invocation is still unscanned, and this one is structural

Every wrapper case in section 1 used a **literal** repo slug. The callers do not:

```
# .claude/agents/code-reviewer.md:33 (Hakim and Tariq carry the same line)
tracker_review_submit "$PR_HOST_REPO" {number} comment "$REVIEW_BODY_FILE"

# /approve-merge SKILL.md:297
tracker_pr_merge "$PR_HOST_REPO" "<pr>" "${MERGE_STRATEGY}" true "$RELEASE_SUBJECT" "$RELEASE_BODY_FILE"
```

Measured at this commit:

| Command | rc |
|---|---|
| `tracker_review_submit "$PR_HOST_REPO" 1210 comment "$REVIEW_BODY_FILE"` | **0 — unscanned** |
| `tracker_review_submit "$PR_HOST_REPO" 1210 comment "/tmp/leak.md"` | **0 — unscanned** |
| `tracker_pr_merge "$PR_HOST_REPO" "1210" "squash" true "$RELEASE_SUBJECT" "$RELEASE_BODY_FILE"` | **0 — unscanned** |
| `tracker_review_submit "${PR_HOST_REPO}" 12 comment "/tmp/leak.md"` | **0 — unscanned** |

`WRAPPER_REPO` resolves to the literal text `$PR_HOST_REPO`, matches no entry in the public-repo list, `TARGET_REPO` stays empty, exit 0.

**I am explicitly not asking you to fix this in this hook, and not asking you to fail closed on it.** A `PreToolUse` hook sees the pre-expansion string; both the repo and the body path are variables. Failing closed on an unresolvable argument 1 would block every review post and every merge, because that *is* the canonical form. This is the limit AgDR-0104 describes, not a tokenizer bug — and the tokenizer work here is correct.

Two responses I would accept, and one I would not:

- **Preferred:** move the content scan **inside** `_lib-tracker.sh`. At the top of `tracker_review_submit` / `tracker_pr_merge` / `tracker_create`, `$repo` and `$body_file` are both fully expanded. One shared `_tracker_scan_private_refs "$repo" "$body_file"` call covers all three wrappers *and* all three adapters at once — which also closes the `glab` gap I raised last round, where `_tracker_review_glab` reads the body into a shell variable and no scanner ever sees it. This is the only position where the wrapper's content is knowable.
- **Acceptable:** document the limit precisely and file it.
- **Not acceptable:** leaving the rule file reading as it does now. Its three wrapper examples all show literal slugs, and the prose says a reference posted through "any of these five shapes reached a public repo unscanned" in the past tense. A maintainer will conclude the canonical call is covered. It is not. One sentence — *the positional target must be a literal slug; a variable cannot be resolved at hook time* — fixes it.

### 5. LOW — the degraded path covers 9 of 10, not 10

With `jq` stubbed to exit 127:

| Shape | rc |
|---|---|
| `gh issue create` / `pr create` / `issue comment` / `pr comment` | 2 |
| `gh pr review` / `gh pr merge` | 2 |
| all three wrappers | 2 |
| **`gh api …/issues -f body='<leak>'`** | **0 — silent pass** |

The `review|merge` and wrapper arms landed; the `api` arm I suggested last round did not. Worth correcting the status report — the coordinator's summary says "all ten shapes".

### 6. LOW — short-flag spellings bypass the new `=` refusal

The refusal matches only the four long forms. `gh` uses Go's pflag, which also accepts `-F=value` and the attached `-Fvalue`:

```
-F=/tmp/leak.md   → rc 0      -b='<leak>'  → rc 0      -t='<leak>'  → rc 0
-F/tmp/leak.md       → rc 0      -b'<leak>'      → rc 0
```

Same silent-pass shape the long-form fix just closed, one character away. Low likelihood, but the argument for fixing the long forms applies verbatim.

---

### 7. Finding 4 from round 1 — correcting the record

The author's message fix is good and I am glad it landed: naming only the flag that actually failed is strictly better, and their reasoning about **not** weakening the refusal for a genuinely chained command is correct — that would reopen `#1039`/`#1046`.

But **the root cause they reached does not describe the case I reported**, and the conclusion "could not reproduce a spurious block from `--subject` in prose" is incorrect. I instrumented the hook at this commit to print both flags:

```
gh pr comment 5 --repo me2resh/apexyard --body 'the flag is --subject "value" here'
DIAG TITLE_TRUNCATED=1  BODY_TRUNCATED=0  TITLE=["value"]  BODY=[the flag is --subject "value" here]
BLOCKED: could not safely determine where a --title or --subject value ends
```

`TITLE_TRUNCATED`, not `BODY_TRUNCATED`. The command is `gh pr comment` — there is no `gh pr merge` and no `--body` truncation anywhere in it. `BODY` parsed cleanly, to completion.

Control, same command with the flag word replaced by a non-flag word:

```
--body 'the flag is FOO "value" here'   → TITLE_TRUNCATED=0  BODY_TRUNCATED=0  → no block
```

So the flag word in the prose is causally responsible. The mechanism is the one I described: the widened `--title|--subject|-t` pattern finds `--subject` inside the quoted body, both quoted branches fail to close, the unquoted fallback returns `"value"` with a leading raw quote, and the invariant check fires.

The `gh pr merge --body` case the author found is **also** real, and their analysis of it is sound. Both exist; fixing the message addressed one of them.

Why I think the residual is still worth closing, and why closing it is not the weakening they were right to resist: the documented invariant — *a leading raw quote can only come from the unquoted fallback* — is true, and the check is a sound **failure** detector. What is not sound is the inference that a failure to parse implies a real flag was present. In my case there is no `--subject` flag on the command; the token lives inside a body the hook already read in full. Refusing there protects nothing, because nothing went unscanned. The same guard shape I suggested in section 3 applies: if the flag-shaped token lies inside a successfully-extracted value, it is content, not an unparsed flag.

Non-blocking. It fails closed, `--body-file` recovers it, and it only affects inline `--body`. But it should not be closed as not-reproducible.

---

### Confirmed unchanged from round 1 (not re-litigated)

Per your instruction I did not reopen the `-t`/`--subject` alias, truncation inheritance, the haystack asymmetry, or stdin/unreadable-file handling. I did re-run both suites at this commit: **leak 105/0, fail-closed 15/0**, matching your report. The two runbook corrections landed — step 4 now points at the forks endpoint, and the overstated headline is gone.

### What I could not verify

- **Your leak-scrub of the served diff.** I did not independently re-derive the 96-token / 0-hit result; I am taking that as reported.
- **`_lib-extract-pr.sh` sourcing side effects** beyond a scan for top-level statements. I did not trace its conditional `_lib-tracker.sh` source under every path.
- **Runtime matcher semantics**, unchanged from round 1 — my read that the `settings.json` conditions are not prefix-anchored still rests on observed hook firings, not on reading the matcher implementation.
- **Live behaviour.** All results are the hook executed against a synthetic registry in a sandbox. No real tracker write was exercised.

### To merge

1. **Section 2** — gate the wrapper fold-ins on the write segment beginning with a wrapper name. **Blocking.**
2. **Section 4** — one sentence in the rule file stating the positional target must be a literal slug. **Blocking as documentation**; the in-library scan can be a follow-up.
3. Sections 3, 5, 6, 7 — all fine as follow-ups.

With 1 and 2 done I would consider this safe to merge.

---

📌 Reviewed commit: `2e8b787689817cdc12b8eb1070b0f65331db9a61`
🔒 No security marker written — verdict is not clean.

me2resh (commented):
**Withdrawn — posted in error.**

This review body is a duplicate of review `5134771563`, which is the genuine security review for this commit. It was posted by a second reviewer running concurrently against the same session scratchpad file. The first reviewer overwrote that file between this reviewer reading it and posting it, and the tracker wrapper returned success both times, so nothing signalled the mistake.

No finding here is attributable to the reviewer whose name is on it. Read `5134771563` for the security review and `5134777091` for the code review.

The underlying defect is being tracked: concurrent reviewers share one scratchpad path with no namespacing, and a review-submission exit code cannot detect that the wrong body was posted.

me2resh (commented):
**Withdrawn — posted in error.**

This review body is a duplicate of review `5134771563`, which is the genuine security review for this commit. It was posted by a second reviewer running concurrently against the same session scratchpad file. The first reviewer overwrote that file between this reviewer reading it and posting it, and the tracker wrapper returned success both times, so nothing signalled the mistake.

No finding here is attributable to the reviewer whose name is on it. Read `5134771563` for the security review and `5134777091` for the code review.

The underlying defect is being tracked: concurrent reviewers share one scratchpad path with no namespacing, and a review-submission exit code cannot detect that the wrong body was posted.

me2resh (commented):
## Code Review: PR #1210 (round 2)

**Commit**: `2e8b787689817cdc12b8eb1070b0f65331db9a61`
**Scope**: Full. Heavy tier. The diff changes `.claude/hooks/*.sh`, so rail 1 forbids a reduced-scope pass.

### Verdict

**CHANGES REQUESTED.** One blocking item. It is a documentation limit, not a code defect.

Both round-1 blockers are fixed. I verified each one. The wrapper fix does not cover the call shape that `code-reviewer.md` prescribes. Disclose that limit, then this is ready.

### Round-1 findings: all resolved

| Finding | State | Evidence |
|---|---|---|
| H1 fail-open on broken jq | Fixed | 9 of 10 shapes now refuse. Was 4 of 7. |
| H2 wrapper shapes unmatched | Fixed | 7 of 8 invocation shapes now block. Was 0. |
| M1 real project names in fixtures | Fixed | 37 registry tokens checked against every added line. 0 hits. |
| S1 scope claim not durable | Fixed | `leak-protection.md` names the `glab` gap and `_tracker_review_glab`. |
| S2 runbook ambiguity | Fixed | The replacement is more accurate than the one I proposed. |

The fixture count returned to its merge-base value of 68. The invented fixture appears 37 times. It is absent from the real registry. I confirmed that directly.

You also found and fixed a bug I missed. Runbook step 4 searched the private portfolio registry for forks of a public repo. That is the wrong source. It now calls `gh api repos/<owner>/<repo>/forks`.

---

## N1 (blocking) — the canonical call shape passes unscanned

The wrapper extraction reads literal text. It cannot expand a shell variable. `code-reviewer.md` prescribes this call:

```
tracker_review_submit "$PR_HOST_REPO" {number} comment "$REVIEW_BODY_FILE"
```

Argument 1 is a variable. `_extract_wrapper_arg` returns the string `$PR_HOST_REPO`. That string matches no public repo. `TARGET_REPO` stays empty. The hook exits 0.

Measured against a fixture registry:

```
Literal arguments (what the 30 new cases cover)
  rc=2  tracker_review_submit "me2resh/apexyard" 1210 comment review.md   | BLOCKED
  rc=2  tracker_pr_merge "me2resh/apexyard" 1210 squash true "..." f | BLOCKED
  rc=2  tracker_create "me2resh/apexyard" "title" review.md         | BLOCKED

Variable arguments (what code-reviewer.md prescribes)
  rc=0  tracker_review_submit "$PR_HOST_REPO" 1210 comment "$BODY_FILE"   | PASSES UNSCANNED
  rc=0  tracker_review_submit "$PR_HOST_REPO" 1210 comment review.md      | PASSES UNSCANNED
```

I posted my round-1 review through the second form. It went unscanned.

This is not a regression. Coverage rose from zero wrapper shapes to seven of eight invocation shapes. Nothing got worse. My objection is narrower.

The hook header states that the wrapper form is "the CANONICAL path, not an edge case". The rule file lists three wrapper rows with literal-argument examples. A reader compares those rows to `code-reviewer.md` and concludes the prescribed call is covered. It is not.

This repeats round 1's pattern one level deeper. The artefact claims more coverage than exists.

There is also an internal inconsistency. This PR added a refusal for `--title=value`, because the hook cannot read that value. It kept `exit 0` for a wrapper whose target it cannot resolve. Both cases mean "I cannot read this". They fail in opposite directions.

**Required fix. Choose one.**

1. Document the limit. Add one row or note to the shape table. State that a wrapper call must pass a literal `owner/repo` slug. State that a variable argument is not scanned. Add the same note to the hook header. File a follow-up ticket.
2. Refuse when a wrapper's argument 1 begins with `$`. Reuse the equals-form refusal you just wrote. Tell the caller to pass a literal slug.

**I recommend option 1 for this PR.** Option 2 changes behaviour for every wrapper call in the portfolio, including calls to private repos. That deserves its own ticket and its own decision. Do not bolt it on here.

Cite writing profile rule 11 for the required fix. The rule says clear text must not change the evidence. A missing limit changes the evidence.

---

## N2 (low) — a backslash continuation inside wrapper arguments is not scanned

`grep -oE` reads one line. The argument span stops at the line end.

```
rc=0  tracker_review_submit "me2resh/apexyard" 1210 \
        comment review.md              | PASSES UNSCANNED
```

Argument 1 resolves on line 1. Argument 4 sits on line 2. `BODY_FILE` stays empty. The hook reaches the empty-content exit and passes.

Seven of the eight shapes I tested do work. Command substitution, a sourced-then-called one-liner, a leading newline, single quotes, bare slugs, an `if` guard, and a trailing `; echo $?` all block correctly. This is the one gap. Same root cause as N1. Record it with N1.

## N3 (low) — `gh api` still fails open on a broken jq

I told you in round 1 that this one was not yours. It is still open, and it is now the only shape of ten that fails open.

```
rc=0  FAILS OPEN <-- gh api repos/me2resh/apexyard/issues -f body=x
```

The commit message says both gh-shape and wrapper-shape coverage "must fail closed here too". `gh api` is a gh shape. The regex cannot match it. Either add an alternation such as `gh[[:space:]]+api[^"]*\b(issues|pulls)\b`, or state the exclusion. Do not leave the claim broader than the regex.

---

## Finding 4: my read

**You are substantially right. Keep your fix.** Your attribution fix is correct. I verified it.

```
--subject + --body, chained command  -> names "--body"
--body only, chained command         -> names "--body"
--subject only, chained command      -> names "--title or --subject"
--subject only, no chain             -> no refusal
```

The reported case trips `BODY_TRUNCATED`. The old message named `--subject` regardless. The new message names the flag that failed. That is the right repair.

**Correct one point in your framing.** "Could not reproduce" is too strong. A `gh pr merge` with `--subject` alone and a chained command does refuse. I reproduced it above. It names `--title or --subject`.

That refusal is correct behaviour, not a false positive. The hook cannot find the end of the quoted value. It refuses. Hakim saw new behaviour, because `dev` never scanned `gh pr merge`. I confirmed that on `549f2af`. Both of those commands exit 0 there.

So the honest statement is not "cannot reproduce". It is: the refusal is real, it is new for this command, it is correct, and the reported case is attributed to `--body`.

**Do not weaken the refusal.** I agree with you. Weakening it reopens #1039 and #1046 for this shape. Practical exposure is small. `tracker_pr_merge` passes a body file, not an inline `--body`, and quote-stripped wrapper arguments cannot trip the check.

## The two quotations: keep them

**Your call is correct.** Do not shorten them.

Rule 2 caps a sentence at 25 words. Rule 11 says clear text must not change the evidence. A verbatim quotation is evidence. Rule 2 governs sentences you write. It does not govern text you cite.

Trimming a citation to fit a word limit would misrepresent a source. `evidence-grounding.md` forbids that outcome directly.

I scanned every new prose line in the rule file. Exactly two sentences exceed 25 words. Both are the citations. Your own prose meets the profile.

One suggestion, not a change request. Put both citations in a markdown blockquote. A reader then sees at once that the long sentence is quoted, not authored. The semicolons in it are GitHub's, not yours.

---

## Verified this pass

- Leak suite: 105 passed, 0 failed.
- Fail-closed suite: 15 passed, 0 failed.
- Full suite: 145 of 147. The 2 failures are the known host-specific pair. They also fail on `dev`.
- `shellcheck`: identical to the merge base. Same two notes. Line numbers shifted only.
- The `find_write_segment` anchor adds the three wrapper names as top-level alternatives. That is correct. A wrapper call carries no `gh` prefix.
- Reusing `_extract_wrapper_arg` imported no unsafe assumption. The tokenizer returns quote-stripped tokens. Folding them into `TITLE` and `BODY_FILE` inherits the truncation and unreadable-file checks, as you state.
- Sourcing via `$(dirname "$0")` is correct. The library is a sibling file. `git rev-parse` would fail outside a git directory.
- Missing-library behaviour is `exit 0`, and you justify it in the comment. I accept that reasoning for a library-install fault.

## Not verified

- The GitHub Support citation. I still have no network access. You report re-checking it against the live page before this push. I record that as your claim, not my observation.
- Hakim's original reproduction command. I inferred the shape from your description. My adjudication holds for the shapes I tested.

---

No approval marker written. The verdict is CHANGES REQUESTED.

Fix N1 by documenting the limit. Fold N2 into the same note. Then re-request review. I expect no further round.

---
📌 Reviewed commit: `2e8b787689817cdc12b8eb1070b0f65331db9a61`

me2resh:
The requested follow-up is pushed in `9c54059`.

- Documented that wrapper scanning requires a literal `owner/repo` in argument 1 and same-line body-file arguments.
- Added the unresolved-wrapper limitation and follow-up requirement.
- Extended the jq failure fallback to cover `gh api` issue and pull requests.
- Tests pass: leak protection 105/105; fail-closed JSON 16/16; diff check clean.

Please review the new commit.

me2resh (commented):
## Code Review: PR #1210 (round 3)

**Commit**: `9c54059641820b1c16338f12d25234937896f7bc`
**Scope**: Full. Heavy tier.

### Verdict

**CHANGES REQUESTED.** One new blocking defect. It is not one of the two accepted follow-ups.

I was asked to rule on `2e8b787` with two findings accepted as follow-ups. I had decided to approve on that basis. My pre-post check found that HEAD had moved to `9c54059`, so I reviewed the new commit instead. It introduces a fail-open defect in a trust-chain control.

**Next action:** replace `[^\n]*` with `.*` on the new `gh api` line, then add a test whose repo slug contains the letter `n`.

### B1 (blocking) — the new `gh api` fail-closed pattern is defeated by the letter `n`

Commit `9c54059` added this to the degraded-path check:

```
|| raw_payload_command_matches "$INPUT" 'gh[[:space:]]+api[^\n]*\b(issues|pulls)\b'
```

`[^\n]` is a POSIX bracket expression. It does not mean "not newline". It means "not backslash, and not the letter n".

Any repository slug containing the letter `n` therefore defeats the pattern. The guard fails open for it. I ran seven controlled paths through the hook with `jq` stubbed to exit 127:

```
rc=2  CLOSED  repos/aaa/bbb
rc=0  OPEN    repos/aaa/bnb
rc=0  OPEN    repos/nnn/bbb
rc=2  CLOSED  repos/me2resh/apexyard
rc=0  OPEN    repos/owner/backend
rc=0  OPEN    repos/acme/anvil
rc=2  CLOSED  repos/acme/widget
```

The correlation with the letter `n` is exact.

I confirmed the cause at the library level, isolating the pattern from the rest of the hook. Columns are the five slugs. `Y` means the payload fails closed:

```
Y . . . Y   gh[[:space:]]+api[^\n]*\b(issues|pulls)\b     <- current
Y Y Y Y Y   gh[[:space:]]+api.*\b(issues|pulls)\b         <- verified fix
Y Y Y Y Y   gh[[:space:]]+api[^"]*\b(issues|pulls)\b      <- also works
```

`grep` reads one line at a time. `.*` cannot cross a line here, so excluding a newline buys nothing.

### B2 (blocking) — the test added in the same commit passes by luck

`test_fail_closed_json.sh` gained this case:

```
gh api repos/me2resh/apexyard/issues -f body=leak
```

`me2resh/apexyard` contains no letter `n`. The case passes for that reason alone. The suite reports 16 of 16 with the defect present.

This is the more serious half. A guard that fails open for most adopters now carries a green test that certifies it. An adopter whose upstream is `acme/anvil` or `owner/backend` gets no protection and no signal.

**Fix:** add a case with a slug containing `n`. `repos/owner/backend/issues` reproduces the failure directly.

### Why this is not covered by the accepted decision

The operator accepted two items: my N1, and Hakim's wrapper-prose false positive. Both are pre-existing in kind. Both were assessed on `2e8b787`.

B1 and B2 arrived in `9c54059`, after that assessment. B1 fails open in a control whose purpose is to fail closed. B2 hides B1 behind a passing test. Neither was seen or accepted.

I am not reopening the accepted pair. They stand as accepted.

### The N1 and N2 documentation fix is correct

`9c54059` also added the sentence I asked for:

> The wrapper scan resolves only a literal `owner/repo` in argument 1. It cannot expand a shell variable or a value on a later line.

That is accurate, and it covers both items. The variable clause resolves my N1. The later-line clause resolves my N2, the backslash continuation. Naming the follow-up for a fail-closed response is the right disposition.

Keep this text unchanged.

### Process note

The brief stated that no code had changed since my round-2 review. HEAD had moved from `2e8b787` to `9c54059` before I posted. I caught it with an assertion that compares the forge HEAD against the SHA I reviewed, immediately before submitting.

Had I approved on the stated SHA, I would have written a marker for `2e8b787` while HEAD was `9c54059`. The gate would have rejected it on the SHA mismatch. The defect above would have stayed unreviewed.

I recommend the same assertion for the other reviewers.

### Verified at this HEAD

- Leak suite: 105 passed, 0 failed.
- Fail-closed suite: 16 passed, 0 failed. See B2 for why that number is misleading.
- The nine other shapes still fail closed correctly.
- The wrapper scan still works for a literal slug. I re-ran `tracker_create` with a real body file: a clean title and body pass, a leaked title blocks, a leaked body blocks.

### Not verified

- The GitHub Support citation. I have no network access.

---

No approval marker written.

Fix B1 and B2, then re-request review. Both are one line each. I expect no further round.

---
📌 Reviewed commit: `9c54059641820b1c16338f12d25234937896f7bc`

me2resh:
## Bounded review against issue #1206

**Verdict: changes requested.** The command-shape coverage and remediation runbook are present, and checks are green. The PR does not complete the original ticket:

- The required file-content enforcement point is not implemented. The PR body and rule file state that committed file content remains a known gap.
- Net-diff independence is therefore not implemented or tested.
- The wrapper scan requires literal arguments; the prescribed `$PR_HOST_REPO` / `$REVIEW_BODY_FILE` call remains unscanned. Documenting that limit is useful, but it does not satisfy the ticket's “any shape `tracker_review_submit` can emit” requirement.

Please split or link the follow-up for file-content scanning and resolve the variable-argument path before treating #1206 as complete.

me2resh:
The remaining file-content and variable-wrapper gaps are now tracked in #1218.

This PR is bounded to the command-shape protection and remediation runbook. Please review the current implementation against that scope; #1218 owns staged-content scanning and history-independent detection.

me2resh:
Follow-up #1218 work is included in commit `f366f45`:

- Added `check-private-refs-staged.sh`.
- Scans complete staged file content, independent of net diff.
- Wired it to the `git commit` PreToolUse path.
- Diagnostics do not print matched private identifiers.

Please review the expanded scope.

Issue #1202, opened 2026-09-07
[Task] Stop _resolve_real_path emitting a doubled leading slash


me2resh:
## Driver

`_resolve_real_path` emits a **doubled leading slash** (`//nope/db/...`) when the nearest existing ancestor is `/`. `require-migration-ticket.sh` squashes it for the target at `_rmt_normalise_target` (line 507) but **not** for the anchors, so the two spellings diverge and `_rmt_project_for_path`'s first branch misses.

| Workspace anchor state | `dev` | after #1198 |
|---|---|---|
| root component absent | project marker | **ops marker** |
| `/workspace` absent (`OPS_ROOT` empty) | project marker | **ops marker** |

Same class as #1198's B1 finding, far narrower trigger: it needs the anchor's **top-level** component to be absent, which no shipped configuration produces. That is why it is a ticket and not a change request on #1198.

## Scope

The durable fix belongs in `_lib-path-resolve.sh` — do not emit `//` in the first place — rather than adding a second squash site. Line 507 is currently the only squash in the entire hook tree, so every future caller inherits the same trap.

Note the mirror: AgDR-0131's sweep row 6 records the target side of this and calls it "None". Update that row if the helper changes.

## Acceptance Criteria

- [ ] `_resolve_real_path` never returns a path with a doubled leading slash.
- [ ] Both anchor states in the table above resolve to the project marker again.
- [ ] The redundant squash at `_rmt_normalise_target` is removed or documented as belt-and-braces.
- [ ] A test pins the helper's output directly, failing if the doubled slash returns.
- [ ] Every other `_resolve_real_path` caller is checked for the same assumption.

## Glossary

| Term | Definition |
|------|------------|
| Lexical collapse | Removing `.` and `..` segments by string manipulation, without touching the filesystem. |
| Anchor | A root path a write target is prefix-matched against to decide which project governs it. |

me2resh:
Implemented and merged in PR #1234. _resolve_real_path now avoids doubled leading slashes, the migration cleanup is reconciled, and direct regression coverage passes.

PR #1234, opened 2026-09-10
fix(#1202): prevent doubled leading slash in path resolver

me2resh:
## Summary

- Fix `_resolve_real_path` when the nearest existing ancestor is `/`.
- Remove the migration gate's duplicate doubled-slash cleanup.
- Add direct helper coverage for absent roots and absent tails.
- Update AgDR-0131 to record the resolved edge case.

## Why it matters

The helper previously returned paths such as `//nope/db/migrations/1.sql`. The migration gate had a local cleanup for its target, but other callers could still receive the malformed path. The shared helper now emits one canonical leading slash for every absolute path.

## Validation

- `bash .claude/hooks/tests/test_lib_path_resolve.sh`
- `bash .claude/hooks/tests/test_require_migration_ticket.sh`
- `bash -n .claude/hooks/_lib-path-resolve.sh .claude/hooks/require-migration-ticket.sh .claude/hooks/tests/test_lib_path_resolve.sh`
- `shellcheck -S warning --format=gcc .claude/hooks/_lib-path-resolve.sh .claude/hooks/require-migration-ticket.sh .claude/hooks/tests/test_lib_path_resolve.sh`
- `git diff --check`

## Glossary

| Term | Definition |
|------|------------|
| resolved path | Canonical absolute path returned by `_resolve_real_path` |
| absent tail | Path suffix whose components do not exist yet |
| migration gate | Hook that selects the ticket governing a migration write |

Closes #1202


me2resh (commented):
## APPROVED

Reviewed PR #1234 at exact head `2884736c112d0089507110de5f5ec91c1ae62596`.

The resolver fix is scoped to the root-ancestor join in `_resolve_real_path`: it emits `/tail` when the nearest existing directory is `/`, and preserves the existing `dir/tail` behavior everywhere else. Removing the migration gate's duplicate cleanup is safe because all callers now receive the canonical single-leading-slash form. The direct helper tests cover an absent root and an absent tail, and the migration suite continues to exercise target and anchor resolution assumptions.

Validation passed:

- Direct path-resolver regression suite: 2/2.
- Migration-gate suite passed through the focused cases run locally.
- ShellCheck at warning threshold.
- `bash -n` and `git diff --check` passed.

AgDR-0131 records the resolved edge case and updates the prior doubled-slash assumption. No blocking findings remain. The PR is approved for merge at this exact commit, subject to the hosted hook test suite completing successfully.

## Glossary

| Term | Definition |
|---|---|
| Nearest existing ancestor | The highest existing directory found while resolving an absent path. |
| Absent tail | Path components that do not exist and are appended after the resolved ancestor. |
| Canonical path | A normalized absolute path with one leading slash. |

atlas-apex (approved):

Issue #1165, opened 2026-09-03
[Feature] Cross-harness quality regression and release readiness


me2resh:
## User Story
As an ApexYard maintainer, I want one cross-harness quality regression before release so that the new defaults improve trust and usability without reducing safety or task completion.

**Blocks**: none
**Blocked by**: #1162, #1163, #1164

## Acceptance Criteria
- [ ] A small corpus covers real grounding, proportionality, and readability failures.
- [ ] The same representative tasks run across every supported harness.
- [ ] Results score unsupported claims, unnecessary work or artifacts, meaning preservation, and first-paragraph clarity.
- [ ] No supported harness has a high-severity regression in safety, technical precision, or task completion.
- [ ] All high-severity failures are fixed before release.
- [ ] The corpus remains available as a permanent framework regression check.
- [ ] Confidence verified — Medium.

## Context

Build a small corpus of real failure cases and run it across each supported harness. Score unsupported claims, unnecessary work or artifacts, meaning preservation, and first-paragraph clarity. Fix high-severity regressions before release and keep the corpus as a permanent framework check.

**Value**: High
**Risk**: Medium
**Kill criterion**: Stop the release if the regression set shows no measurable improvement, or if the changes reduce safety, technical precision, or task completion.

_Source: /plan-initiative on 2026-09-03 — see `projects/initiatives/grounded-proportionate-human-output.md` in the private portfolio sibling repository for the full initiative and dependency graph._

PR #1248, opened 2026-09-10
fix(#1165): make regression fixtures actionable

me2resh:
## Summary
- Fix HF-06 so the scenario includes the commit identifier that the case requires the agent to preserve.
- Give PW-06 the security-sensitive context needed to test the Heavy path.
- Mark authentication and quota failures as `NOT-RUN` without treating valid transcripts with benign stderr warnings as failures.
- Add static coverage for the new runner behavior.

## Why it matters
The first representative run exposed false high-severity failures caused by an incomplete fixture and unavailable credentials. The runner must distinguish a real regression from a harness that did not produce a usable transcript.

## Testing
- `git diff --check`
- `bash .claude/hooks/tests/test_quality_regression.sh` (42 passed)
- `shellcheck bin/quality-regression.sh`
- `bash .claude/rules/tests/test_writing_standard.sh` (144 passed)
- Focused Codex run: HF-06 and PW-06 both completed with exit 0 and passed their mechanical checks.

## Glossary
| Term | Definition |
|------|------------|
| Corpus | The fixed set of cases used by the regression runner. |
| Mechanical check | A heuristic result derived from transcript text or written files. |
| NOT-RUN | A case for which the harness could not produce a usable transcript because credentials or quota were unavailable. |
| Heavy path | The full review and decision process required for a security-sensitive change. |
| Transcript | The captured prompt, output, and stderr for one case. |

Refs #1165

### DIFF
--- a/.claude/hooks/tests/test_quality_regression.sh
+++ b/.claude/hooks/tests/test_quality_regression.sh
@@ -31,6 +31,8 @@ assert "runner:no-llm-judge" grep -qF 'It does not use an LLM judge' "$RUNNER"
 assert "runner:empty-change-count" grep -qF 'n=$(grep -c . "$changed" 2>/dev/null) || n=0' "$RUNNER"
 assert "runner:portable-timeout" grep -qF 'gtimeout' "$RUNNER"
 assert "runner:uncapped-fallback" grep -qF 'else' "$RUNNER"
+assert "runner:auth-unavailable-is-not-run" grep -qF 'NOT-RUN harness authentication or quota unavailable' "$RUNNER"
+assert "runner:heavy-path-signal" grep -qF 'security-sensitive' "$RUNNER"
 
 # every case id the runner parsed appears in the index
 for id in $(echo "$check_out" | sed -n "s/^cases parsed: $expected_cases (\\(.*\\))$/\\1/p"); do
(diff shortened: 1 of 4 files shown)


me2resh (commented):
## APPROVED

Reviewed PR #1248 at exact commit `a20b219c1ee342aac989f3eea84e41e7dff278f6`.

The fixture corrections make the cases actionable: HF-06 now includes the required commit identifier, and PW-06 explicitly supplies the security-sensitive context needed to exercise the Heavy path. The runner keeps the adjudicated result human-owned and now classifies authentication and quota failures as `NOT-RUN` while preserving valid stdout when a harness emits only benign stderr warnings. A real output is passed to the mechanical check even when stderr contains auth-related text; auth/quota detection uses stderr only when stdout is empty, so valid Codex output is not masked.

No real failure path is hidden: non-empty valid output still reaches the case-specific mechanical check, empty output with auth/quota diagnostics is marked unavailable, and ordinary non-auth failures remain visible through their mechanical result and exit code.

Validation passed:

- `git diff --check`
- `bash .claude/hooks/tests/test_quality_regression.sh` — 40/40
- `shellcheck bin/quality-regression.sh`
- `bash .claude/rules/tests/test_writing_standard.sh` — 144 passed

No blocking findings remain. Approved for merge at this exact commit, subject to the required CI checks completing successfully.

## Glossary

| Term | Definition |
|---|---|
| NOT-RUN | A case for which credentials or quota prevented a usable harness transcript. |
| Mechanical check | A heuristic result derived from transcript text or written files. |
| Adjudication | The human's final PASS or FAIL decision from transcript evidence. |
| Heavy path | The full review and decision process required for a security-sensitive change. |

atlas-apex (approved):

PR #1250, opened 2026-09-10
fix(#1165): avoid false scorecard failures

me2resh:
## Summary
- check sandbox scope before scanning illustrative login text in EG-01
- recognize qualified uncertainty such as “not yet been confirmed” in EG-05 and HF-06
- add regression assertions for both checker boundaries

This keeps available-harness scorecards from reporting failures when the transcript preserves the required boundary or uncertainty.

## Testing
- `bash .claude/hooks/tests/test_quality_regression.sh` (46 passed)
- `shellcheck bin/quality-regression.sh`

## Glossary
| Term | Meaning |
|------|---------|
| mechanical check | A deterministic text or file check applied to a transcript. |
| scorecard | The per-harness result table, with a human adjudication column. |

Refs #1165

### DIFF
--- a/.claude/hooks/tests/test_quality_regression.sh
+++ b/.claude/hooks/tests/test_quality_regression.sh
@@ -35,6 +35,8 @@ assert "runner:auth-unavailable-is-not-run" grep -qF 'NOT-RUN harness authentica
 assert "runner:cursor-auth-message" grep -qF 'authentication required' "$RUNNER"
 assert "runner:adapter-startup-failure-is-not-run" grep -qF 'database is locked' "$RUNNER"
 assert "runner:heavy-path-signal" grep -qF 'security-sensitive' "$RUNNER"
+assert "runner:context-check-precedes-example-text" grep -qF 'if grep -qiE "sandbox" "$out"' "$RUNNER"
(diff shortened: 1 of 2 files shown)


me2resh (commented):
## APPROVED

Re-reviewed PR #1250 at exact commit `ff084201401dd67454ab9e75eb0cb37bf39aac37`.

EG-01 now gives sandbox scoping precedence over illustrative `not logged in` text, which removes the reported false failure while preserving the unscoped failure path. EG-05 and HF-06 accept qualified uncertainty through up to three modifier words, including `not yet been confirmed`, without removing the existing SHA requirement for HF-06. The changes are limited to the intended mechanical checks and their regression assertions.

Validation passed:

- `bash .claude/hooks/tests/test_quality_regression.sh` — 40 passed, 0 failed
- `shellcheck bin/quality-regression.sh`
- `git diff upstream/dev...ff084201401dd67454ab9e75eb0cb37bf39aac37 --check`

No blocking findings remain.

## Glossary

| Term | Meaning |
| --- | --- |
| EG-01 | The regression case that checks whether a sandbox-scoped failure is reported accurately. |
| EG-05 | The regression case that checks whether uncertainty or modality is preserved. |
| HF-06 | The regression case that checks preservation of a required SHA and uncertainty. |

atlas-apex (approved):

Issue #1253, opened 2026-09-10
[Chore] Release v5.5.0


me2resh:
## Objective
Cut framework release v5.5.0 from `dev` to `main`.

## Acceptance criteria
- [ ] CHANGELOG.md contains the generated v5.5.0 entry.
- [ ] The v5.4.0-to-v5.5.0 migration placeholder is present.
- [ ] Available Cursor and Codex regression runs have no high-severity mechanical failures.
- [ ] The release PR carries the exact `Released-From` trailer.
- [ ] Merge CI creates tag v5.5.0 and the GitHub release.
- [ ] `dev` is synced from `main` after the release.

## Scope
This release aggregates changes already merged to `dev`. Claude coverage remains unavailable until credits are restored.

Issue #1233, opened 2026-09-10
[Bug] A review-class agent can commit and push to the PR it is reviewing, silently destroying author-reviewer separation


muhammadattia95:
## Affected

`.claude/agents/code-reviewer.md` (and by inheritance `security-reviewer.md`, `solution-architect.md`) · `.claude/hooks/block-unreviewed-merge.sh` (the gate whose invariant is broken)

## Given / When / Then

**Given** a review-class sub-agent is spawned on an open PR. All three grant `Bash`:

```
$ grep -n '^tools:' .claude/agents/{code-reviewer,security-reviewer,solution-architect}.md
code-reviewer.md:6:tools: Read, Grep, Glob, Bash, mcp__apexyard-search__search_code, …
security-reviewer.md:6:tools: Read, Grep, Glob, Bash, mcp__apexyard-search__search_code, …
solution-architect.md:5:tools: Read, Grep, Glob, Bash, mcp__apexyard-search__search_docs, …
```

and nothing in any of those prompts forbids repo-mutating git:

```
$ grep -niE 'git (add|commit|push)|(do|must) not (commit|push)' .claude/agents/code-reviewer.md
(no matches)
```

**When** the reviewer decided the fix was small enough to just make — it edited two tracked files, staged them, committed, and pushed to the PR branch it was reviewing, then wrote its approval marker.

**Then** two invariants broke silently:

1. **The reviewer became a co-author of the diff it approved.** `.claude/rules/pr-workflow.md` § "Build agents cannot self-review" forbids the author→reviewer direction and is enforced advisorily by `warn-review-marker-write.sh`. The reviewer→author direction is unaddressed: no rule text, no prompt guardrail, no hook. The two-marker gate's whole premise — that `*-rex.approved` comes from someone other than the author — is defeated from the other side.

2. **The marker it then wrote was stale on arrival.** It recorded the pre-push SHA while its own push had already moved PR HEAD:

   ```
   marker holds: f2c02ae…   ← SHA at review start
   PR HEAD now:  ad28520…   ← the reviewer's own commit
   ```

   `block-unreviewed-merge.sh` correctly refuses the merge on SHA mismatch, so the *merge* is safe. What is not safe is the diagnosis: the operator sees "APPROVED" plus a blocked merge and has no signal that the reviewer authored part of the PR. Only reading commit author/committer metadata reveals it.

Expected: a review-class agent is read-only with respect to the repository. It reports findings; the orchestrator or a build agent applies them.

[...]

Issue #1230, opened 2026-09-10
[Bug] block-main-push.sh (and other hooks) fire against a session-pinned ops root instead of the git command's actual target repo


a-abdellatif98:
## Affected
.claude/settings.json line 125 (the PreToolUse matcher wiring for block-main-push.sh), and the shared ops-root resolution snippet it shares with 77 other hook entries in the same file (78 total matches for APEXYARD_OPS_PIN_DIR).

## Given / When / Then
Given a session with a pinned ops-root (APEXYARD_OPS_PIN_DIR or the default
$HOME/.claude/apexyard/ops-root-$CLAUDE_CODE_SESSION_ID file) pointing at the
apexyard fork
When a plain "git init && git commit" runs in a completely unrelated scratch
repo (default branch main, no relation to the fork)
Then the PreToolUse wiring resolves the ops-root from the session pin first,
finds it valid ($r/.claude/hooks exists there), and execs
block-main-push.sh from that pinned root regardless of what repo the git
command actually targets, producing:
  "BLOCKED: Cannot commit directly on protected branch 'main'."
Expected: the hook should only fire when the git command's actual target
repo (its own toplevel, from the command's cwd) is the ops fork itself.
Failing in an unrelated repo is safe-direction but still wrong scoping.

## Repro
1. Have an active session with a resolved ops-root pin pointing at the apexyard fork
2. mkdir a scratch dir elsewhere, git init, git add, git commit (default branch is main)
3. Observe the commit is blocked with the "protected branch main" message,
   even though the repo committed to has nothing to do with the fork

## Framework version
v5.4.0

## Severity
major, fails in the safe direction (blocks unrelated work) rather than
unsafe, but it is a wrong-repo resolution bug affecting the same shared
wiring used by 78 hook entries in .claude/settings.json, not just this one

## Notes
Same class as #1226: silent, confidently wrong resolution that reads as a
flake rather than a root cause. The session-pin-first design likely exists
to avoid repeated ops-root walks per tool call, but it should still verify
the actual git target's toplevel matches the pinned root before firing a
repo-scoped hook like block-main-push.sh, rather than trusting the pin alone.

## Glossary
| Term | Definition |
|------|------------|
| ops-root pin | A per-session cached path to the apexyard fork root, used to skip repeated directory walks |
| protected-branch gate | block-main-push.sh, blocks direct commits to main/master |

me2resh:
[...]

Issue #1225, opened 2026-09-09
[Feature] Support split tracker hosts (issues on one system, reviews on another)


a-abdellatif98:
## Problem
tracker.kind is a single enum (gh, linear, jira, asana, custom, glab, none)
that conflates two separate concerns: which system holds issues, and which
system hosts code review (PRs/MRs). A registered project using Jira for
issues and GitLab for code hosting has no clean value to set. glab assumes
GitLab does both. jira assumes a GitHub-shaped review host. The only escape
is kind: custom with a review_command, which _lib-tracker.sh's own comments
(around line 1264) call "a future dedicated review-host config", i.e. not
built yet.

## Proposed behaviour
A dual-axis tracker config, e.g. tracker.issue_kind and tracker.review_kind,
so an adopter can declare issue tracker and review host independently
(Jira issues + GitLab reviews, Linear issues + GitHub reviews, etc.)
without forcing a single wrong enum value.

## Why it helps adopters
Any adopter whose issue tracker and code host differ hits this today.
Right now the honest choice is tracker.kind: none, which disables the
whole tracker abstraction for that project (no automatic ticket
resolution, no review-host dispatch) rather than partially working.

## Scope / out of scope
Out of scope: full per-adapter parity for every issue/review combination
on day one. In scope: the config shape and resolution order, starting
with the Jira-issues + GitLab-reviews pair.

## Framework version
v5.4.0

## Related
.claude/hooks/_lib-tracker.sh (tracker_kind, tracker_view, tracker_review_submit)

## Glossary
| Term | Definition |
|------|------------|
| tracker.kind | The registry field selecting which CLI/adapter a project's tracker operations use |
| review host | The system a PR/MR actually lives on and gets reviewed through |

me2resh:
Implemented and merged in PR #1239. The tracker now supports independent issue and review hosts through tracker.issue_kind and tracker.review_kind, with legacy tracker.kind fallback.

PR #1254, opened 2026-09-10
release(#1253): v5.5.0

### LINKED ISSUE #1136: [Bug] /approve-merge on a release PR drops the Released-From trailer (bare squash + COMMIT_MESSAGES buries it)

### LINKED ISSUE #1138: [Feature] Portfolio-view skills should aggregate across a multi-repo project's repos
## Problem

AgDR-0121 (shipped in v5.4.0) added a plural `repos:` schema so a genuinely multi-repo product registers as **one** governed project. The security-critical consumers were all updated to read the full list: `block-private-refs-in-public-repos.sh` scrubs every repo slug, `_lib-tracker.sh` resolves a per-project tracker override from any member repo, and `_lib-multi-repo-trace.sh` resolves targets across the list.

The **portfolio-view skills were explicitly deferred**: `/projects`, `/inbox`, `/tasks`, `/status`, and `/handover` still render one row per **registry entry** and do not aggregate anything across a project's repos. A multi-repo project correctly shows as a single row — but the PR counts, issue counts, and CI status in that row reflect only the primary repo, silently under-reporting the rest of the system.

The practical effect is that the new schema is currently self-defeating for anyone who uses the portfolio view. To get accurate counts today, an adopter has to go back to the pre-AgDR-0121 workaround — N registry entries sharing one `docs:` directory — which is exactly the shape AgDR-0121 set out to replace. So you can have a correct single-project identity **or** correct counts, not both.

Worth noting for tracking hygiene: AgDR-0121's Consequences section states this work was *"Filed as an explicit follow-up rather than silently left undiscovered."* No such issue exists in this tracker — a search for the aggregation follow-up and for references to the originating issue both return nothing. This issue closes that gap.

## Proposed behaviour

The five portfolio-view skills read the project's full repo list (via the already-shipped `mrt_repos_for`) and aggregate across it.

The substance of this request is **not** "loop over repos and sum" — it is that **different fields aggregate differently**, and getting that wrong produces a confidently misleading dashboard. That is the design work being asked for.

### Aggregation semantics per field

| Field | Aggregation | Why not a naive sum/mean |
|---|---|---|
| Open PRs, open issues | **Sum** | Counts compose cleanly |
| CI status | **Worst-of** | One red repo means the product is degraded. A mean or a majority vote hides the break. |
| Last activity | **Max** (most recent) | |
| Harnessability score | **Do NOT average** | It is a per-codebase judgement across 5 dimensions. Averaging 8 repos hides the one weak repo, which is precisely the repo you needed to see. Worst-of, or per-repo, never a mean. |
| Contributors / bus factor | **Union, not sum** | The same human appears in several repos; summing invents headcount and masks a real bus-factor risk. |
| `status:` (handover/active) | **None** | Already project-level |

### Rendering — needs a decision

| Option | Trade-off |
|---|---|
| **A.** One summed row | Compact; hides *which* repo is red — the actionable part |
| **B.** One row + per-repo breakdown, always | Complete; noisy at 8 repos × N projects |
| **C.** One row + per-repo detail **only on exception** (red CI, blocked, stale) | Compact by default, detail exactly where attention is needed |
| **D.** One row + an explicit `--by-repo` drill-down flag | Predictable, but the user must know to ask |

Suggested: **C as the default, D as the explicit drill-down.** Open to argument — this is the main thing worth designing rather than assuming.

### Cost / latency — a real constraint, not a footnote

Aggregation multiplies API calls by the repo count, on skills people run constantly. At 8 repos that is 8× the calls on every `/status`. A naive per-repo loop would make the portfolio view visibly slow on a large portfolio and could hit rate limits. Options worth weighing: one batched GraphQL query per project, parallel fetch with a concurrency cap, or a short-lived cache keyed on repo + HEAD. This should be decided in the design, not discovered in review.

### Action targeting

Read-only views are fine (`/tasks` already emits direct per-item URLs). Anything that needs *the* repo — starting a ticket, opening a PR — should resolve to `primary:`, or prompt when `primary:` is absent.

### Backwards compatibility

Singular `repo:` projects must render byte-identically. AgDR-0121 already defines `repo: X` as equivalent to `repos: [X]` with no explicit primary, and `test_multi_repo_registry.sh` asserts that equivalence for the existing consumers — the same assertion should extend to whatever these skills render.

## Why it helps adopters

Anyone governing a product split across several repos that deploy together — a backend plus a gateway plus N workers is the common shape — currently chooses between an honest single-project identity and an honest dashboard. The portfolio view is the framework's primary answer to "what needs my attention right now"; if it under-reports a multi-repo product, the answer is wrong in the direction that matters, because the unreported repos are invisible rather than obviously missing.

It also removes the incentive to keep using the N-entries-sharing-one-docs-dir workaround, which AgDR-0121 identified as a bandage.

## Scope / out of scope

**In scope:** aggregation semantics + rendering for `/projects`, `/inbox`, `/tasks`, `/status`, `/handover`; the fetch/caching strategy needed to keep them responsive.

**Out of scope:** the `repos:` schema itself and the security consumers (all shipped in v5.4.0 via AgDR-0121); any change to leak-scrub behaviour, which deliberately ignores `primary:` and must keep treating every repo equally.

A design pass is requested **before** implementation — the rendering choice and the non-summable-field handling are the load-bearing decisions, and they touch five separately-defined skill flows.

## Framework version

v5.4.0

## Related

- AgDR-0121 (first-class multi-repo registry project) — this is its explicitly deferred half
- `.claude/hooks/_lib-multi-repo-trace.sh` — `mrt_repos_for` / `mrt_primary_repo_for`, the accessors this would consume
- `.claude/hooks/tests/test_multi_repo_registry.sh` — existing equivalence assertions to extend
- Skills affected: `/projects`, `/inbox`, `/tasks`, `/status`, `/handover`

## Glossary

| Term | Definition |
|------|------------|
| portfolio view | The set of skills that render a cross-project dashboard — `/projects`, `/inbox`, `/tasks`, `/status` — answering "what needs attention across everything I govern" |
| multi-repo project | One governed product whose code spans several repositories that deploy and are governed together, registered as a single registry entry via `repos:` |
| `primary:` | Optional registry field naming which repo tracker/board operations default to; defaults to the first entry of `repos:` |
| leak-scrub | `block-private-refs-in-public-repos.sh` removing registered private project identifiers from anything written to a public framework repo |
| harnessability | The 5-dimension score `/handover` assigns a codebase for how well it can be governed by the framework |
| worst-of aggregation | Combining per-repo values by taking the least healthy one, rather than a sum or mean, so a single failure stays visible |

### LINKED ISSUE #1146: [Bug] validate-commit-format.sh validates the body as the subject when git commit is given multiple -m flags

### LINKED ISSUE #1147: [Bug] AgDR-0021 and /setup disagree on whether workspace/README.md survives split-portfolio v2 setup (plus an inert gitignore negation)
## Affected

- `docs/agdr/AgDR-0021-split-portfolio-v2-path-resolution.md` § G
- `.claude/skills/setup/SKILL.md` (step 6, the `git rm --cached` line)
- `.claude/skills/split-portfolio/SKILL.md`
- `.claude/skills/update/SKILL.md` (step 8a)

## Given / When / Then

**Given** a fresh fork being configured for split-portfolio v2 mode by `/setup`

**When** following `/setup`'s own documented step 6, which runs:

```bash
git rm --cached -r projects onboarding.yaml workspace
```

**Then** `workspace/README.md` is untracked from the public fork — but **AgDR-0021 § G states the opposite**: that `workspace/README.md` is a framework artefact which *stays* in the public fork, and both the v1→v2 migration recipe and `/update` step 8a special-case it specifically to avoid moving it.

Expected: one consistent answer about whether `workspace/README.md` survives split-portfolio setup.

An adopter who follows `/setup` therefore ends in a state that the AgDR describes as incorrect, and that a future `/update` may try to revert — producing a confusing diff long after the setup session is over, with no obvious link back to the cause.

## Repro

1. Run `/setup` on a fresh fork and answer yes to the private-projects question, so it takes the split-portfolio v2 branch
2. Let it complete step 6 (`git rm --cached -r projects onboarding.yaml workspace`)
3. Observe `workspace/README.md` staged for deletion
4. Read `docs/agdr/AgDR-0021-split-portfolio-v2-path-resolution.md` § G, which says that file stays in the public fork
5. Read `.claude/skills/update/SKILL.md` step 8a, which special-cases the same file to avoid moving it

## Framework version

v5.4.0

## Severity

minor — documentation / skill inconsistency. Nothing breaks at setup time and either state is functional, but the two sources disagree, and the disagreement surfaces later as unexplained `/update` churn.

## Notes

**Related, and probably the same fix:** after split-portfolio setup the fork's `.gitignore` carries both

```
workspace/*/
!workspace/README.md
```

Git cannot re-include a file whose parent directory is excluded, so that negation is inert regardless of which side of the contradiction is chosen. If the AgDR's position is the correct one and `workspace/README.md` should stay tracked, the ignore rules need restructuring (exclude the subdirectories without excluding the parent) for the negation to actually work. If `/setup`'s behaviour is correct, the negation should simply be deleted along with the file.

Either way the current combination is self-contradictory in three places at once: the AgDR, the skill that performs setup, and the ignore file that setup leaves behind.

## Glossary

| Term | Definition |
|------|------------|
| Split-portfolio v2 | A layout where the public ApexYard fork holds only framework code, while the registry, company config, and per-project docs live in a separate private repo, resolved via the `portfolio` block in `.claude/project-config.json`. |
| AgDR | Agent Decision Record — the framework's format for recording a technical decision and its trade-offs. |
| Gitignore negation | A `!pattern` line that re-includes a previously excluded path. Git documents that it has no effect if the file's parent directory is itself excluded. |

### LINKED ISSUE #1148: [Chore] Migrate live site pointers yard.apexscript.com → apexyard.ai
## Driver
apexyard.ai is now the canonical site domain (yard.apexscript.com 301-redirects to it). The framework repo's live "where is the site" pointers still name the old domain, which is stale for new readers.

## Scope
- Update **live site pointers only**: `README.md`, `AGENTS.md`, `docs/multi-project.md`, `docs/harnesses/README.md` — `yard.apexscript.com` → `apexyard.ai`.
- **Leave historical records unchanged**: CHANGELOG entries, AgDR-0004 / AgDR-0047 narratives, and the two spike docs are dated statements accurate to when written; rewriting them makes the record lie about its own past, and the old links still 301-redirect. Add a dated note only if a specific one genuinely confuses a reader.

## Acceptance Criteria
- [ ] The 4 live-pointer files reference apexyard.ai, not yard.apexscript.com
- [ ] No historical record (CHANGELOG / AgDR / spike doc) is rewritten
- [ ] apexyard.ai resolves; yard.apexscript.com still redirects (no broken links)

### LINKED ISSUE #1151: [Bug] Merge-gate repo resolution: a leading `cd` outranks the repo named in the command — design + architecture gates evaluate the wrong repo and fail OPEN
## Affected

| File | Affected? |
|---|---|
| `.claude/hooks/require-architecture-review.sh` | **YES — fails OPEN** |
| `.claude/hooks/require-design-review-for-ui.sh` | **YES — fails OPEN** |
| `.claude/hooks/block-unreviewed-merge.sh` | No (resolves correctly — but diverges, see below) |
| `.claude/hooks/block-merge-on-red-ci.sh` | No (resolves correctly — but diverges, see below) |

## Given / When / Then

**Given** an ops fork governing several managed repos, and a merge issued for PR #42 in a managed repo — using the `tracker_pr_merge` wrapper (the shape `/approve-merge` documents), which carries the repo as **positional argument 1**, from a shell block that begins `cd <ops-fork> && …`

**When** the merge-gate hooks evaluate the command

**Then** `require-architecture-review.sh` and `require-design-review-for-ui.sh` resolve `CMD_REPO` to the **ops fork's own origin** — not the repo named in the command — and evaluate the ops fork's PR #42 instead. Expected: the repo explicitly named in the command wins.

Both directions of wrongness occur. Mine first showed up as a **false block** (the gate cited design-artifact files belonging to a different repo's PR of the same number). Reproducing it deliberately shows the **fail-open** direction is equally reachable, and is the one that matters.

## Root cause — resolution precedence

`require-architecture-review.sh:126-149` (and the byte-identical block at `require-design-review-for-ui.sh:134-157`):

1. `--repo` flag
2. `gh api …/pulls/<N>/merge` URL path
3. **cd-target origin** — `pr_cmd_cd_target` + `git_origin_repo` (added by #687)
4. `extract_repo_from_command` fallback

The wrapper's positional repo argument is only read by **step 4** (`extract_repo_from_command`'s `2b` branch, added by #759). So **step 3, a heuristic about the shell's working directory, outranks an explicit repo argument present in the very same command.** Step 4 holds the correct answer and is never reached.

`extract_repo_from_command` itself is sound — called directly with the same command string it returns the correct repo. The defect is purely the precedence.

## Fail-open determination: **FAILS OPEN — proven, not inferred**

Reproduced with two throwaway local git repos and a stubbed `gh`; no merge was performed. All names below are invented.

Setup: ops fork origin `synthorg/portfolio-ops`; managed repo origin `synthorg/service-alpha`. Both have a PR numbered 42.

- `synthorg/service-alpha` #42 — diff contains `docs/designs/payment-flow-technical-design.md` and `src/ui/Panel.tsx`. Has valid `…__service-alpha__42-rex.approved` and `-ceo.approved` at its own HEAD. **No** architecture marker, **no** design marker. CI green.
- `synthorg/portfolio-ops` #42 — unrelated PR; has its own valid markers.

Command under test (`cd` prefix + wrapper, repo named literally):

```
cd <ops-fork> && . …/_lib-tracker.sh && tracker_pr_merge "synthorg/service-alpha" "42" "squash" true
```

Result — all four gates run against that command:

| Gate | Repo it queried | Exit |
|---|---|---|
| `block-unreviewed-merge.sh` | `synthorg/service-alpha` ✅ | 0 (valid markers) |
| `block-merge-on-red-ci.sh` | `synthorg/service-alpha` ✅ | 0 (green) |
| `require-design-review-for-ui.sh` | **`synthorg/portfolio-ops`** ❌ | **0 — silent** |
| `require-architecture-review.sh` | **`synthorg/portfolio-ops`** ❌ | **0 — silent** |

**The merge proceeds, and the design + architecture gates never evaluated the PR being merged.** `src/ui/Panel.tsx` and `payment-flow-technical-design.md` ship with no design review and no Solution Architect sign-off, and nothing is printed.

Negative control — **identical** repo state, markers, and PR number; only the resolution path differs:

| Command shape | Repo queried | arch gate | design gate |
|---|---|---|---|
| `cd <ops-fork> && … tracker_pr_merge "synthorg/service-alpha" "42" …` | portfolio-ops ❌ | **exit 0 (allowed)** | **exit 0 (allowed)** |
| `cd <ops-fork> && gh pr merge 42 --repo synthorg/service-alpha --squash` | service-alpha ✅ | exit 2 BLOCKED | exit 2 BLOCKED |
| `… tracker_pr_merge "synthorg/service-alpha" "42" …` (no `cd` prefix) | service-alpha ✅ | exit 2 BLOCKED | exit 2 BLOCKED |

The verdict flips on the resolution path alone. The checks themselves work; they are pointed at the wrong repository.

Both fail-open mechanisms are reachable:

- **(a)** the wrong repo's PR #N carries no design/UI files (or does not exist) → `CHANGED` is empty → `exit 0`;
- **(b)** the wrong repo's PR #N *does* carry them **and holds a valid marker** → the gate reads the wrong marker, validates it against the wrong repo's HEAD, and passes — a fully self-consistent evaluation of the wrong PR. Confirmed by the stub log: `gh pr diff 42 --repo synthorg/portfolio-ops` followed by `gh pr view 42 --repo synthorg/portfolio-ops --json headRefOid`.

**Scope, stated precisely so this is not overstated:** the Rex + human-approval gate and the red-CI gate are **not** bypassed — they resolve the repo correctly for this shape and still fire. This is not an unreviewed-merge hole. It is a silent, complete no-op of **two of the four** merge gates.

## Why the wrapper makes this likely in practice

The wrapper is the shape `/approve-merge` documents (`.claude/skills/approve-merge/SKILL.md:244`), and it is the only merge shape that carries its repo **positionally** rather than as `--repo`. A hand-written `gh pr merge <n> --repo <owner/repo>` hits step 1 and is unaffected. So the sanctioned path is the exposed one, and a leading `cd` into the ops fork is the normal way to reach `_lib-tracker.sh`.

## Divergence across the four gates (worth reporting on its own)

Two independent inconsistencies, neither of which looks deliberate:

| | cd-target step (#687) | `merge_command_uses_variable` guard |
|---|---|---|
| `require-architecture-review.sh` | yes | **no** |
| `require-design-review-for-ui.sh` | yes | **no** |
| `block-unreviewed-merge.sh` | **no** | **no** |
| `block-merge-on-red-ci.sh` | **no** | yes |

Consequence of the second column: `SKILL.md:244` passes the repo as `"$PR_HOST_REPO"`, and hooks see the command text **before** expansion. On that form the two design/architecture gates resolve a literal `$PR_HOST_REPO`, `gh pr diff` fails, `CHANGED` is empty, and both **exit 0** — verified. (`block-merge-on-red-ci.sh` blocks that form via its variable guard, so the composite merge does stop; the point is that the other two gates silently no-op rather than fail closed.)

## A second, independent fail-open

`require-architecture-review.sh:197-200` and `require-design-review-for-ui.sh:209-212`:

```sh
CHANGED=$(gh pr diff "$PR_NUMBER" $REPO_FLAG --name-only 2>/dev/null)
if [ -z "$CHANGED" ]; then
  # Couldn't determine files — skip rather than false-positive
  exit 0
fi
```

This contradicts each file's own header, four lines from the top:

> `# CLASS: CONTROL … Keep it fail-closed: if it cannot evaluate its precondition it must block, never allow (AgDR-0104).`

An unresolvable diff *is* a precondition failure. It is what converts a wrong-repo resolution into a silent pass, and it would equally convert a network blip or an auth expiry into one. Sibling of #1091 / #1098, which fixed the same class of degrade-open for the HEAD-SHA lookup but left this branch.

## Repro

1. `git init` two throwaway repos; give them distinct `origin` remotes (`synthorg/portfolio-ops`, `synthorg/service-alpha`). Add `.apexyard-fork`, `onboarding.yaml`, `apexyard.projects.yaml` and a copy of `.claude/hooks/` to the first.
2. Put a stub `gh` first on `PATH` that logs its argv and answers `pr diff` per `--repo`: `service-alpha` → a `*technical-design*.md` plus a `.tsx`; `portfolio-ops` → an unrelated file.
3. Write valid `rex` + `ceo` markers for `synthorg__service-alpha__42`; write none for architecture/design.
4. With cwd = the ops fork, pipe the command as `{"tool_input":{"command":"…"}}` into `require-architecture-review.sh`, for each of the three shapes in the control table above.
5. Compare exit codes and the stub's log. Shapes 2 and 3 block; shape 1 exits 0 having queried `portfolio-ops`.

## Framework version

v5.3.0

## Severity

**major** — wrong result with a workaround (pass an explicit `--repo`, or don't prefix the merge with `cd`). Not `blocker`: the approval and red-CI gates still fire, so no unreviewed merge is possible through this path.

## Suggested fix

Make an **explicit repo named in the command outrank the cd-target heuristic**. Concretely, insert the wrapper's positional-argument extraction as a new step **between** the API-path step and the cd-target step — i.e. before step 3, not after it. That is the smallest change that fixes it.

A blind reorder of the *whole* `extract_repo_from_command` call above the cd-target step would **not** be safe: its step 3 is an ambient `gh pr view` against the checkout's own origin, which would pre-empt the cd-target recovery and re-break #687. Only the explicit-argument branch should be promoted.

Better still, given the divergence table above: hoist a single `resolve_merge_repo()` into `_lib-extract-pr.sh`, used by all four hooks, with one documented precedence — (1) `--repo`/`-R`, (2) forge API URL path, (3) wrapper positional arg, (4) cd-target origin, (5) ambient fallback — so "explicit beats heuristic beats ambient" is stated once rather than re-implemented four times with three different answers. That also gives the two hooks currently missing `merge_command_uses_variable` a natural place to acquire it.

Separately, the `[ -z "$CHANGED" ] → exit 0` branch should fail closed to match its own `CLASS: CONTROL` header.

## Glossary

| Term | Definition |
|------|------------|
| Merge gate | A `PreToolUse` hook that inspects a merge command and exits non-zero to block it. |
| cd-target | The directory in a leading `cd <path> && …` prefix. Hooks fire before the `cd` runs, so #687 added a step that reads that directory's `origin` to recover the repo. |
| Wrapper form | `tracker_pr_merge <owner/repo> <pr> <strategy> [<delete>]` — the tracker-agnostic merge adapter (#759) that `/approve-merge` calls. Carries the repo **positionally**, not as `--repo`. |
| Approval marker | A gitignored file under `.claude/session/reviews/`, named `<owner>__<repo>__<pr>-<kind>.approved`, holding the reviewed SHA. Keyed by repo — so a wrong-repo resolution consults a different file. |
| Fail open | A control that, when it cannot evaluate its precondition, allows the action. The opposite of the fail-closed posture these hooks' own headers require. |

### LINKED ISSUE #1152: [Bug] The fail-closed principle from #965/#1091 was applied only to the 4 merge gates — 19 other blocking hooks still exit 0 when they cannot evaluate their precondition

### LINKED ISSUE #1159: [Bug] require-migration-ticket.sh decides from raw command text: wrong-marker fallback and false positives
## Affected

`.claude/hooks/require-migration-ticket.sh`, and the target-extraction path it shares with `require-active-ticket.sh` via `_lib-detect-bash-write.sh`.

Two failure directions, one root cause. Both were hit within minutes of each other during the same session.

## Given / When / Then

**Failure 1 — false negative: a variable write target resolves against the wrong marker.**

**Given** a session inside a git worktree, with a correct per-worktree ticket marker at `.claude/session/tickets/<project>/<safe-branch>` and a **stale** ops-level marker at `.claude/session/current-ticket` naming an unrelated ticket

**When** a migration-path file is written by a Bash heredoc whose target is a **shell variable** — `cat > "$WD/app/migrations/0001_initial.py" <<'EOF'` — rather than a literal path

**Then** the hook cannot expand the variable, target extraction fails, project resolution fails with it, and the gate falls back to the **ops-level marker** — evaluating against the wrong ticket, with no warning that it fell back

**Expected** either correct resolution (as with a literal path), or a refusal. An unresolvable target on a migration path should fail closed, not silently resolve against a different ticket.

Literal absolute paths work correctly. Only the variable form is affected.

**Failure 2 — false positive: a document that merely mentions a migration path is blocked.**

**Given** the same session

**When** a Bash heredoc writes a **markdown file to `/tmp`** whose body happens to contain migration-path strings — because the document is *about* migrations

**Then** the hook blocks the write, reporting that the active ticket lacks the `migration` label

**Expected** no block. Nothing under any migration path is being written; the only file touched is a temporary markdown document.

This was hit while writing up Failure 1 — the report text contained example paths, and the hook matched them. This issue body had to be written with a file-writing tool rather than a heredoc to get past it.

## Repro

**Failure 1**

1. In a fork with a registered project, create a git worktree and run `/start-ticket` so the per-worktree marker exists.
2. Ensure `.claude/session/current-ticket` exists and names a **different** ticket — a leftover from earlier work in the same session is the realistic case.
3. From that worktree, write a migration file with a variable target: `cat > "$WD/somemodule/migrations/0002_x.py" <<'EOF'`.
4. The gate evaluates against the ops-level marker rather than the per-worktree one.
5. Repeat with a literal absolute path — resolution is correct.

**Failure 2**

1. With any active ticket that lacks the `migration` label, write a markdown file to `/tmp` via a heredoc whose body contains a migration path string.
2. The write is blocked, naming the active ticket.

## Framework version

v5.4.0

## Severity

major.

Failure 1 is the more serious of the two: a gate that resolves against the wrong marker is worse than one that fails, because the failure is invisible. It surfaced on a large multi-worktree refactor where variable-driven heredocs are the natural way to write generated migration files, and the preconditions occur naturally — a stale ops-level marker is an ordinary artifact of earlier work in the same session, and the correct per-worktree marker was well-formed the whole time. Nothing was misconfigured; the extraction simply could not see the path.

Failure 2 is less dangerous but blocks legitimate work, and it is the direction that trains people to route around the gate.

## Notes

The shared root cause is that the hook decides from the **raw text of a shell command** rather than from a resolved write target. That cannot distinguish:

- a real write whose path it cannot expand (Failure 1), from
- a document that merely mentions such a path (Failure 2).

This is the same conclusion AgDR-0104 reached for `warn-review-marker-write.sh` — that deciding by pattern-matching shell-command text cannot be made sound, and that hook was returned to advisory as a result. `require-migration-ticket.sh` is a **blocking** gate, so the same unsoundness has sharper consequences in both directions.

`.claude/rules/workflow-gates.md` § "Out-of-repo exemption" already states the fail-closed principle for this class: *"An unresolvable or ambiguous target (an unextractable Bash write-target, in particular) is never exempted here — it falls straight through to the ticket gate, unchanged."* Failure 1 is the sibling case — the target is unextractable, and rather than falling through to the correct gate it falls back to a different marker.

Suggested direction, in order of preference:

1. When a Bash write target cannot be resolved to a project, **refuse** rather than falling back to the ops-level marker.
2. Failing that, emit a visible warning naming which marker was used, so a silent fallback becomes an audible one.
3. For Failure 2, consider restricting extraction to the redirection target itself rather than scanning the whole command text including heredoc bodies.

## Glossary

| Term | Definition |
|------|------------|
| Per-worktree marker | `.claude/session/tickets/<project>/<safe-branch>` — the Tier 0 marker letting parallel agents hold independent tickets |
| Ops-level fallback | `.claude/session/current-ticket` — the Tier 2 marker, used when project resolution fails |
| Target extraction | The hook parsing a Bash command's raw text to determine which file is being written |
| Fail closed | Refusing when a control cannot evaluate its precondition, rather than proceeding on a guess |

### LINKED ISSUE #1161: [Bug] /code-review's forked run never writes its approval marker, so the merge gate cannot be satisfied

### LINKED ISSUE #1163: [Feature] Proportionate work and ceremony
## User Story
As an ApexYard operator, I want the framework to choose work and ceremony that match the task so that small changes stay small without weakening safeguards for risky work.

**Blocks**: #1165
**Blocked by**: none

## Acceptance Criteria
- [ ] Lean, Standard, and Heavy classification applies to planning, implementation, artifact creation, and review.
- [ ] Agents start with the smallest change that satisfies the acceptance criteria.
- [ ] Agents reuse existing files, patterns, and dependencies before adding new ones.
- [ ] New dependencies, abstractions, services, and durable artifacts require a demonstrated need.
- [ ] Advice and quick assessments remain conversational unless a durable artifact or workflow gate requires one.
- [ ] Existing Heavy safeguards remain unchanged for security, trust-chain changes, migrations, and ambiguous work.
- [ ] Regression tasks contain no unnecessary dependencies, abstractions, roles, artifacts, or workflow steps.
- [ ] Confidence verified — Medium.

## Context

Extend the existing Lean, Standard, and Heavy model from review ceremony to planning, implementation, and artifact creation. Require the smallest sufficient solution, reuse existing patterns first, and justify every new dependency, abstraction, service, or durable artifact. Preserve the existing Heavy safeguards for security, migrations, trust-chain changes, and ambiguous work.

**Value**: High
**Risk**: Medium
**Kill criterion**: Cancel the broader rule change if testing shows that the existing Lean, Standard, and Heavy classifications already select proportionate solutions after a smaller prompt correction.

_Source: /plan-initiative on 2026-09-03 — see `projects/initiatives/grounded-proportionate-human-output.md` in the private portfolio sibling repository for the full initiative and dependency graph._

me2resh:
## Summary

- Releases v5.5.0 from the merged `dev` tip.
- Includes the framework writing, evidence-grounding, proportionality, tracker-host, review-separation, path-resolution, and regression improvements already reviewed on `dev`.
- Adds the required no-op migration for the v5.4.0 → v5.5.0 hop so `/update` can walk the chain.
- Records the exact `dev` cut point for changelog provenance.

## [v5.5.0] — 2026-09-10

Minor release — 8 features, 48 fixes, 21 improvements.

### Added (feat)

- (#1239) split issue and review tracker hosts — 1796a60
- enforce controlled technical writing for new artifacts — f468a0e
- (#1191) require browser evidence in QA and design-review records — 5408ddd
- (#1179) add build-time handbook discovery — 012d586
- (#1170) add cross-harness quality regression — ba05168
- (#1168) add writing standard with Strict and Flavored modes — 97e8479
- (#1167) extend right-size tiers to planning, implementation, and artifacts — 11e7e14
- (#1166) add universal evidence-grounding contract — 5327d74

### Fixed (fix)

- (#1257) block agent privilege escalation — 8be2866
- (#1255) ground reviewer behavior claims in real usage — afb20eb
- recognize security approval safeguards — 205a967
- avoid false quality-regression failures — ff08420
- recognize likely and unconfirmed findings — e34fb5a
- recognize qualified uncertainty in regression checks — 6e52f45
- classify harness startup failures as not run — 0881cf8
- (#1248) make regression fixtures actionable — 8b441dc
- (#1238) keep review agents read-only — 8a1ec98
- (#1234) prevent doubled leading slash in path resolver — 951715f
- (#1232) scope protected-branch backstop to target repo — 0c3e610
- (#1227) accept v-prefixed changelog headings — fcfccc5
- (#1224) require browser evidence for Head of Design — 98b378d
- (#1223) keep release PR variables in one shell block — aa96a5a
- (#1222) restore fixture heading — 10ba602
- (#1221) require Rex to verify cited criteria — b5bf397
- (#1220) scan staged private references — 16450aa
- (#1219) block heredoc compound commands — 31317de
- complete AgDR renumbering audit — 6605342
- reserve AgDR identifiers across branches — ccfad7a
- coordinate AgDR allocation across worktrees — 453a98c
- guard AgDR identifier allocation — 57d9796
- normalize pinned worktree roots — 98cc377
- normalize linked worktree ops roots — 648a5c7
- report skipped suites as failures — 9a65a41
- fail when hook suites skip cases — 15e6414
- surface skipped hook tests — 943b430
- fail closed on unresolved PR base — db1c1a2
- document wrapper scan limits — 9c54059
- block compound shell substitutions — ba12f90
- reject nested commit commands — 51e3701
- cover the wrapper shapes and fix two review-attribution bugs — 2e8b787
- reject newline compound commit commands — e3cc705
- reject compound commit commands — 98bc8ef
- parse first commit message argument — b9b52f3
- scan gh pr review and gh pr merge for leaked private references — 2d72f33
- (#1198) compose _resolve_real_path after lexical collapse in migration gate — 7359a5d
- (#1192) preserve Released-From trailer on release-PR squash — 4bbacfd
- (#1190) correct same-account review verdict guidance — a300c98
- (#1180) refuse unresolvable migration write targets instead of falling back — 65a304b
- (#1156) preserve workspace README in split setup — 62b65f7
- (#1155) resolve merge repo before cd heuristic — a4878dd
- (#1173) make the merge gate satisfiable through the sanctioned path — da0ce0a
- (#1177) close command validation parse bypasses — 0bd7ee5
- (#1176) fail closed on unreadable hook payloads — 47ca52a
- (#1172) right-size handbook review triggers — 2efcf30
- (#1157) detect review markers written to a gate-invisible path — 89209c9
- (#1154) scope Wave 1 skills to tracked files — 6a0f5f1

### Changed (refactor / chore / docs)

- (#1247) clarify quality and adapter guides — fabef4a
- (#1246) clarify hook comments — 7c14a3d
- (#1245) clarify portfolio operations — bc1b0e8
- (#1244) clarify release and harness guides — af69862
- (#1243) clarify single-fork setup — 4eac600
- (#1242) clarify onboarding language — 0e3a11c
- (#1241) clarify workflow guides — ec197aa
- (#1240) clarify framework prose — ace2247
- (#1237) close multi-repo design residuals — b300ed6
- (#1236) dismiss stale forge approvals — aa5b9a6
- (#1235) describe jq config merge semantics accurately — fa46f8d
- (#1231) share active-ticket marker resolver — 56aacc0
- resolve adapter dependency advisories — f86c050
- update AgDR inbound reference — 1b885f0
- cover write detector false positives — ed3f45e
- audit portfolio root worktree behavior — 81e7f43
- quarantine documented environment skips — 471f575
- Merge remote-tracking branch 'upstream/dev' into fix/GH-1206-leak-protection-gaps — 6fe94d2
- (#1205) technical design for portfolio multi-repo aggregation — 6dc663a
- (#1175) apply writing standard across artefacts — 1171f75
- (#1148) point live site references to apexyard.ai — 173c07c

### Closes

- Closes #1136
- Closes #1138
- Closes #1146
- Closes #1147
- Closes #1148
- Closes #1151
- Closes #1152
- Closes #1159
- Closes #1161
- Closes #1163
- Closes #1164
- Closes #1165
- Closes #1171
- Closes #1174
- Closes #1181
- Closes #1182
- Closes #1187
- Closes #1196
- Closes #1197
- Closes #1202
- Closes #1207
- Closes #1209
- Closes #1218
- Closes #1225
- Closes #1230
- Closes #1233
- Closes #1255
- Closes #1257

## Migration

- [x] `.claude/migrations/v5.4.0-to-v5.5.0.sh` is present as an idempotent no-op placeholder. v5.5.0 has no adopter-facing file or configuration migration.

## Testing

- Cursor representative regression: 8 cases completed; no mechanical failures.
- Codex representative regression: 8 cases completed; no mechanical failures.
- Claude, pi, and OpenCode were not run because credentials or quota were unavailable.
- `bash .claude/hooks/tests/test_release_changelog.sh` — 91 passed.
- `bash .claude/rules/tests/test_writing_standard.sh` — 144 passed.
- `shellcheck .claude/migrations/v5.4.0-to-v5.5.0.sh` — passed.

After merge:

1. Confirm CI creates tag `v5.5.0` on `main`.
2. Verify the squash commit carries the `Released-From` trailer.
3. Run `/release-sync v5.5.0` to sync `main` back into `dev`.

Closes #1253

---

## Glossary

| Term | Definition |
|------|------------|
| Release cut | A branch and PR that carries the reviewed `dev` content to `main`. |
| No-op migration | A versioned script that changes no adopter files but keeps the update chain continuous. |
| `Released-From` trailer | A git trailer recording the exact `dev` SHA used for this release. |

Released-From: 8be2866132c7894a83a63d84778e6fe0c28c67f6


me2resh (commented):
## CHANGES REQUESTED

Re-reviewed release PR #1254 at exact commit `3ca6eb5a9057289b744abf6cd3d70ebeaea802b2`.

The release contents are otherwise consistent: the changelog counts are 8 Added, 46 Fixed, and 21 Changed entries; the v5.4.0→v5.5.0 migration is an executable no-op placeholder; `Released-From: f61ab805bc07d5e8b494db180e672b4c80b51c1b` matches the current `dev` branch tip; and the Cursor/Codex evidence is stated separately from unavailable harnesses. The PR body includes the required Glossary and release follow-up steps.

One blocking release-quality issue remains: the current `CHANGELOG.md` has no blank line between the final new `Closes #1233` list item and the existing `# Changelog` heading. The CI markdownlint job reports:

- `CHANGELOG.md:116` MD032/blanks-around-lists
- `CHANGELOG.md:117` MD022/blanks-around-headings

Insert one blank line after `- Closes #1233`, rerun markdownlint, and update the PR head before approval.

## Glossary

| Term | Meaning |
| --- | --- |
| Released-From | The exact `dev` SHA used as the release cut point. |
| No-op migration | A versioned migration script that changes no adopter files but preserves the update chain. |

me2resh (commented):
## APPROVED

Re-reviewed release PR #1254 at exact commit `6d68de44e8076515ea3dab8d89427c1f92438e75`.

The required blank line after `- Closes #1233` is present, and the markdownlint CI check now passes. The release contents remain consistent: the changelog has 8 Added, 46 Fixed, and 21 Changed entries; the v5.4.0→v5.5.0 migration is an executable no-op placeholder; `Released-From: f61ab805bc07d5e8b494db180e672b4c80b51c1b` matches the current `dev` tip; and the Cursor/Codex evidence is clearly separated from unavailable harnesses. The release body contains the required Glossary and post-merge verification steps.

Validation passed:

- `bash .claude/hooks/tests/test_release_changelog.sh` — 91 passed
- `bash .claude/rules/tests/test_writing_standard.sh` — 144 passed
- `shellcheck` on the release migration — passed
- `git diff upstream/dev...6d68de44e8076515ea3dab8d89427c1f92438e75 --check`
- CI `markdownlint-cli2` — passed

No blocking findings remain.

## Glossary

| Term | Meaning |
| --- | --- |
| Released-From | The exact `dev` SHA used as the release cut point. |
| No-op migration | A versioned migration script that changes no adopter files but preserves the update chain. |

atlas-apex (dismissed):

me2resh (commented):
## CHANGES REQUESTED

Re-reviewed release PR #1254 at exact commit `bd7b66748b924ee769e4dd867e8a80c34905efcc`.

The release branch now contains the merged #1255 and #1257 changes, and `CHANGELOG.md` correctly records them: 48 fixes total, with entries and `Closes #1255` / `Closes #1257`. The bundled framework files are present, and the prior migration and markdown fixes remain intact.

The PR body is stale relative to the exact release tree. Its embedded `## CHANGELOG` still says “8 features, 46 fixes, 21 improvements” and omits the #1255 and #1257 Fixed and Closes entries. Because the release process uses the reviewed PR body as release metadata, update that body to match the checked-in changelog, then rerun the release-specific checks. Keep the existing `Released-From: f61ab805bc07d5e8b494db180e672b4c80b51c1b` trailer unless the release cut point is intentionally changed; if it is changed, update both the body and the release provenance consistently.

The current CI status was otherwise clean where complete: Analyze, markdownlint, ShellCheck, Verify Ticket ID, and release smoke passed; the hook suite was still pending.

## Glossary

| Term | Meaning |
| --- | --- |
| Released-From | The exact `dev` SHA from which the release cut was made. |
| Release body | The reviewed PR metadata used to describe the release and its provenance. |
| Post-cut fix | A fix merged after the original release cut and explicitly included in the release branch. |

me2resh (commented):
## APPROVED

Re-reviewed release PR #1254 at exact commit `bd7b66748b924ee769e4dd867e8a80c34905efcc`.

The PR body now matches the checked-in v5.5.0 changelog: 8 features, 48 fixes, 21 improvements, with #1255 and #1257 entries and closures. The bundled reviewer-evidence guidance and least-privilege Bash guard are present in the release tree. `Released-From: 8be2866132c7894a83a63d84778e6fe0c28c67f6` matches the current `dev` tip, and the no-op v5.4.0→v5.5.0 migration remains present and executable.

Validation passed:

- Changelog counts: 8 Added, 48 Fixed, 21 Changed, 28 Closes entries
- `git diff upstream/dev...bd7b66748b924ee769e4dd867e8a80c34905efcc --check`
- Release migration execution and ShellCheck
- CI: Analyze, markdownlint, ShellCheck, release smoke, and Verify Ticket ID passed; CodeQL was neutral/skipped

No blocking findings remain.

## Glossary

| Term | Meaning |
| --- | --- |
| Released-From | The exact `dev` SHA used for this release cut. |
| Post-cut fix | A fix merged after the original release cut and explicitly included in the release branch. |
| No-op migration | A versioned script that changes no adopter files but preserves the update chain. |

atlas-apex (approved):

Issue #1257, opened 2026-09-10
[Task] Enforce least-privilege command execution for agents


me2resh:
## Summary
Add a framework-wide least-privilege guard for agent commands.

## Problem
The framework has merge, review, and workflow gates, but it does not state one common rule for privilege escalation across tools. An agent must not use administrator flags, `sudo`, force or verification bypasses, role-assumption commands, or privileged API paths as a workaround when a normal path is blocked.

## Proposed outcome
- Add a single framework rule: use the least-privileged normal path and stop when it is unavailable.
- Block common escalation command shapes at the Bash gate, including administrator merge flags, `sudo`, force pushes, `--no-verify`, and explicit role-assumption commands.
- Require the operator to perform any exceptional privileged action directly, outside the agent workflow.
- Keep the guard project-agnostic and fail closed for ambiguous elevated forms.

## Acceptance criteria
- The rule is loaded by the framework entry points.
- The Bash guard blocks the listed escalation forms and allows normal commands.
- Tests cover each blocked form and a representative normal command.
- The framework contains no adopter-specific identifiers.

## Glossary
| Term | Meaning |
| --- | --- |
| Least privilege | Use only the access needed for the requested operation. |
| Privilege escalation | Increasing account, role, or command authority beyond the normal workflow. |
| Verification bypass | A flag or path that skips a required safety check. |

me2resh:
Implemented and merged in PR #1258.

PR #1258, opened 2026-09-10
fix(#1257): block agent privilege escalation

me2resh:
## Summary
- Adds a framework-wide least-privilege rule that requires agents to stop when the normal workflow is blocked instead of escalating access or bypassing verification.
- Adds a Bash guard for administrator merges, `sudo`, force pushes, `--no-verify`, and explicit role-assumption commands, so common escalation paths fail before execution.
- Wires the guard into the shared Bash hook table and adds focused regression coverage for blocked and allowed command forms.

Closes #1257

## Validation
- `bash .claude/hooks/tests/test_block_privileged_escalation.sh` (7 passed)
- `bash .claude/hooks/tests/test_settings_wrappers_silent_noop.sh` (4 passed)
- `bash -n .claude/hooks/block-privileged-escalation.sh`
- `python3 -m json.tool .claude/settings.json`
- `git diff --check`

## Glossary
| Term | Meaning |
| --- | --- |
| Least privilege | Use only the access needed for the requested operation. |
| Privilege escalation | Increasing account, role, or command authority beyond the normal workflow. |
| Verification bypass | A flag or path that skips a required safety check. |


me2resh (commented):
## CHANGES REQUESTED

Reviewed PR #1258 for #1257 at exact commit `171eb99671ca303ad6df72125305160b980c7adc`.

The settings wiring, project-agnostic rule, and focused tests are present and pass. The guard still has two blocking command-boundary defects:

1. `git push --force-with-lease origin feature/example` exits 0. `--force-with-lease` is still a force push and must be blocked under the ticket's force-push requirement. Add a regression case and extend the matcher.
2. Normal commands such as `grep sudo README.md` and `echo sudo status` exit 2. The `sudo` alternative is matched after any whitespace, so ordinary arguments and text are treated as an escalation command. Restrict matching to command or shell-segment boundaries and add representative normal-command tests that contain these words as arguments.

The exact-tree checks otherwise passed: the seven guard cases, four settings-wrapper checks, Bash syntax, JSON parsing, and ShellCheck. Please correct both matcher boundaries, rerun the focused suite, and update the PR head.

## Glossary

| Term | Meaning |
| --- | --- |
| Force push | A push using `--force`, `--force-with-lease`, or an equivalent forced ref update. |
| Command boundary | The start of a command or a shell command separator where a command name can begin. |
| Least privilege | Using only the access required for the normal workflow. |

me2resh (commented):
## CHANGES REQUESTED

Re-reviewed PR #1258 for #1257 at exact commit `de81a45891bd260d4f0ba18f64411a784cdec33e`.

The two prior defects are fixed: `git push --force-with-lease` is now blocked, and `sudo` is limited to command or shell-segment position. The exact-tree focused suite passes (10 guard cases, 4 settings-wrapper checks, Bash syntax, JSON parsing, and ShellCheck).

One command-boundary issue remains across the other escalation alternatives. The `gh pr merge`, `git push`, `--no-verify`, and role-assumption branches still use `(^|[[:space:]])` for their command prefix. As a result, ordinary commands containing those strings as arguments are blocked, for example `echo aws sts assume-role --role-arn arn:example`. Apply the same command or shell-segment boundary used for `sudo` to every command-pattern branch, and add representative allowed tests for these strings in ordinary arguments.

This is required for the ticket's no-false-positives criterion and for consistent regex scope. Please update the head and rerun the focused checks.

## Glossary

| Term | Meaning |
| --- | --- |
| Command boundary | The start of a command or a shell separator where a command name can begin. |
| False positive | A normal command incorrectly classified as a privilege escalation. |
| Least privilege | Using only the access required for the normal workflow. |

me2resh (commented):
## APPROVED

Re-reviewed PR #1258 for #1257 at exact commit `20feabe83e52717df15bcf44723a7cb37bf31a3c`.

All escalation alternatives now require command position at the start of a shell command or after a separator. The guard blocks administrator merges, `sudo`, `--force` and `--force-with-lease`, `--no-verify`, and explicit role-assumption commands while allowing ordinary arguments containing those words. Settings wiring remains project-agnostic and guarded by the shared ops-root resolution. The added negative cases cover the false-positive boundaries identified in review.

Validation passed against the exact PR tree:

- `bash .claude/hooks/tests/test_block_privileged_escalation.sh` — 14 passed, 0 failed
- `bash .claude/hooks/tests/test_settings_wrappers_silent_noop.sh` — 4 passed, 0 failed
- `bash -n .claude/hooks/block-privileged-escalation.sh`
- `python3 -m json.tool .claude/settings.json`
- `shellcheck .claude/hooks/block-privileged-escalation.sh .claude/hooks/tests/test_block_privileged_escalation.sh`
- `git diff upstream/dev...20feabe83e52717df15bcf44723a7cb37bf31a3c --check`

No blocking findings remain.

## Glossary

| Term | Meaning |
| --- | --- |
| Command position | The start of a command or a shell separator where a command name can begin. |
| Force push | A push using `--force`, `--force-with-lease`, or an equivalent forced update. |
| Least privilege | Using only the access required for the normal workflow. |

atlas-apex (approved):

Issue #1270, opened 2026-09-11
[Bug] Ops-root walk-up can't see .apexyard-fork when the fork is nested inside an unrelated enclosing git repo


borzoj:
## Affected
`.claude/hooks/_lib-ops-root.sh` (`resolve_ops_root` / `resolve_ops_root_walk`),
`.claude/hooks/_lib-portfolio-paths.sh` (`_portfolio_root`),
`.claude/hooks/_lib-read-config.sh` (`_config_repo_root`)

## Given / When / Then

**Given** an ops fork that lives nested inside an outer, unrelated git
repository (e.g. a parent directory that is itself version-controlled and
happens to also track the fork's files) — the `.apexyard-fork` marker sits at
`<parent>/<fork>/.apexyard-fork`, one level *below* the o
toplevel.

**When** any hook, skill-invoked Bash command, or sub-agent runs with a
working directory whose `git rev-parse --show-toplevel` r
*outer* repo instead of the fork itself (e.g. an operator or tool `cd`s up one
level, or a session launches from the parent directory) —

**Then** `resolve_ops_root_walk` walks upward from that t
`.apexyard-fork` or the legacy `onboarding.yaml` + `apexyard.projects.yaml`
pair. Because the walk only ascends (via `dirname`) and n
subdirectories, it can never find the marker one level down inside the fork.
It silently falls through to the "not under an apexyard f
treats the **outer repo** as the ops-fork root.

Every downstream portfolio resolver (`portfolio_registry`,
`portfolio_workspace_dir`, `portfolio_projects_dir`, `por
`_config_repo_root`) then computes its **in-fork default** relative to the
wrong (outer) root — e.g. `<outer>/workspace`, `<outer>/a
— none of which exist, and none of which is the split-portfolio v2 sibling
repo configured via `.claude/project-config.json`'s `port
`_config_repo_root` hits the identical failure for the same reason, so the
override file itself is never even located — config silen

Expected: resolution should either (a) fail loudly/distin
locate an apexyard fork above this directory — are you inside the fork?"),
or (b) be robust to being invoked from one level above th

## Repro

1. Set up a split-portfolio v2 fork at `<parent>/<fork-na
   private portfolio sibling at `<parent>/<portfolio-name>/`, per the normal
   split-portfolio layout.
2. Make `<parent>/` itself a git repository (e.g. an umbrella/meta repo that
   happens to track both directories' contents) — this is
   condition; a plain, non-git parent directory does not reproduce the bug.
[...]

PR #1276, opened 2026-09-11
fix(#1270): find nested ops fork under enclosing repo

me2resh:
## Summary

- Detect a single anchored ops fork directly below an enclosing Git repository.
- Preserve the existing upward walk and avoid selecting an arbitrary fork when multiple candidates exist.
- Add regression coverage for the nested-fork layout and document the decision in AgDR-0148.

## Why it matters

When an unrelated parent directory is itself a Git repository, the resolver starts at that parent and can never descend into the actual ops fork. Downstream path helpers then use the wrong root and can create or read state in the wrong location. The focused child probe resolves the reported layout without broad descendant searching.

## Validation

- `bash .claude/hooks/tests/test_ops_root.sh`
- `bash .claude/hooks/tests/test_resolve_ops_root_pin.sh`
- `bash .claude/hooks/tests/test_split_portfolio_v2_migration.sh`
- `git diff --check`
- `shellcheck .claude/hooks/_lib-ops-root.sh .claude/hooks/tests/test_ops_root.sh`

## Glossary

| Term | Meaning |
| --- | --- |
| Ops fork | An adopter's ApexYard repository marked by `.apexyard-fork` or the legacy configuration pair. |
| Ops-root resolver | The shared helper that locates the ops fork from a working directory. |
| Enclosing repository | An unrelated Git repository containing the ops fork as a child directory. |

Closes #1270

### DIFF
--- a/.claude/hooks/_lib-ops-root.sh
+++ b/.claude/hooks/_lib-ops-root.sh
@@ -122,6 +122,25 @@ resolve_ops_root_walk() {
   # normalize the starting point before looking for ops-root anchors.
   start=$(_ops_root_main_worktree "$start")
 
+  # A caller can start one level above the fork when that enclosing directory
+  # is itself a Git repository. The upward walk cannot descend into the fork,
+  # so inspect immediate child directories for a single anchored fork before
+  # walking toward /. Do not guess when several children look like forks.
+  local child child_candidate="" child_matches=0
+  for child in "$start"/*; do
+    [ -d "$child" ] || continue
+    if [ -f "$child/.apexyard-fork" ] || {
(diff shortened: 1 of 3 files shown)


atlas-apex (approved):

me2resh (commented):
Rex review completed: no blocking findings.

The resolver now handles an ops fork nested one level below an enclosing Git repository by selecting exactly one anchored child, while preserving the existing upward walk and avoiding ambiguous matches. Regression tests pass for the nested layout, pin resolution, and migration paths. `git diff --check` is clean, and the repository ShellCheck workflow passes. GitHub cannot record an approval from the PR author account, so this is recorded as a review comment.

Issue #1317, opened 2026-09-16
[Refactor] Dispatch Bash PreToolUse through one command matcher


me2resh:
## Driver

Latest `dev` (`8e12865`) wires 54 Bash PreToolUse commands for 31 unique scripts. Claude Code matches on the tool name Bash, not the command string. Merge-gate scripts therefore spawn on every Shell call. A sequential pass on 2026-09-16 took 1.59s cold and 1.49s warm. Twenty-three of those 54 spawns are duplicate host variants. AgDR-0120 cached config lookups. It did not remove the process fan-out.

## Scope

Add one Bash PreToolUse dispatcher that reads the command and execs only the matching hook. Keep every gate in unmodified `.claude/hooks/*.sh`. Do not copy gate logic into a harness adapter. Collapse the duplicated ops-root wrapper in `.claude/settings.json` for the Bash matcher. Regenerated Codex, pi, and opencode adapters must still reach the same scripts. The Cursor overlay stays session-pin only.

## Acceptance Criteria

- [ ] One Bash PreToolUse entry dispatches to the existing hook scripts by command match.
- [ ] Merge-gate hooks do not run on a command that is not a merge.
- [ ] Duplicate host variants of the same script spawn at most once per call.
- [ ] Hook unit tests for the delegated scripts still pass.
- [ ] Conformance CI still proves opencode, pi, and Codex against the unmodified bash hooks.
- [ ] A Claude Code Shell call still blocks `git add -A` via `block-git-add-all.sh`.
- [ ] Cursor IDE still loads `.claude/settings.json` and still refuses a gated Write when third-party configs are on. The overlay still does not copy the gate list.
- [ ] `bin/sync-codex-adapter.sh`, `bin/install-pi-adapter.sh`, and `bin/install-opencode-adapter.sh` still emit transports that exec the same `.claude/hooks/*.sh` files.
- [ ] Measured Bash PreToolUse time for `true` is documented in the PR and is lower than the 1.49s warm baseline from 2026-09-16 on the same class of machine.

## Risks / Dependencies

A dispatcher that filters too broadly can skip a gate. Round unknown commands up to run the hook. Refs closed #1013.

## Glossary

| Term | Definition |
|------|------------|
| Dispatcher | One PreToolUse command that selects which hook script to exec. |
| Harness | Claude Code, Cursor IDE, pi, Codex, or opencode. |
| Overlay | Cursor session-pin file. It is not a second gate list. |

me2resh:
[...]

PR #1320, opened 2026-09-16
perf(#1317): dispatch Bash PreToolUse hooks once

me2resh:
## Summary

- Replace the 54-entry Bash PreToolUse fan-out with one dispatcher command.
- Keep the existing hook scripts, exit codes, and special environment variables.
- Run unconditional safety hooks for every Bash call before command-specific gates. The first blocking reason can change when two gates would both exit 2.
- Add dispatcher regression coverage for unknown commands, merge commands, deduplication, blocking exit codes, and non-blocking failures.
- Restore exec-bit coverage for dispatcher-delegated gates. Align the shared adapter derivation test with the dispatcher table.
- Record the design in AgDR-0157.
- Rename the duplicate nested-ops-root AgDR-0151 to AgDR-0158 so unique-id checks stay green.

## Why this matters

The previous settings entry repeated ops-root discovery for every Bash hook. The dispatcher resolves the root once and invokes only the hooks relevant to the command. In three sequential local `true` calls, the old 54-entry path took about 10.24 seconds and the dispatcher path took about 1.43 seconds.

## Testing

- `bash -n .claude/hooks/dispatch-bash.sh .claude/hooks/tests/test_dispatch_bash.sh`
- `shellcheck .claude/hooks/dispatch-bash.sh .claude/hooks/tests/test_dispatch_bash.sh`
- `bash .claude/hooks/tests/test_dispatch_bash.sh`
- `bash .claude/hooks/tests/test_hook_exec_bits.sh`
- `bash .claude/hooks/tests/test_block_main_push.sh`
- `bash .claude/hooks/tests/test_settings_wrappers_silent_noop.sh`
- `bash .claude/hooks/tests/test_sync_codex_adapter.sh`
- `bash .claude/hooks/tests/test_manage_portfolio_adapters.sh`
- `bash .claude/skills/decide/tests/test_agdr_unique_ids.sh`
- `bash bin/run-hook-tests.sh`
- `node --test harness-adapters/_shared/test/derive-gates-core.test.ts`
- `node --test harness-adapters/pi/test/derive-gates.test.ts`
- `node --test harness-adapters/opencode/test/derive-gates.test.ts`
- `python3 -m json.tool .claude/settings.json`
- `git diff --check`

## Glossary

| Term | Definition |
| --- | --- |
| Dispatcher | The single Bash PreToolUse command that selects existing hooks. |
| Unconditional hook | A safety hook that runs for every Bash call. |
| Command-specific hook | A hook selected by the Bash command prefix. |
| Fan-out | Starting one subprocess for each configured hook entry. |

## Architecture

See [AgDR-0157](https://github.com/me2resh/apexyard/blob/fix/GH-1317-bash-dispatcher/docs/agdr/AgDR-0157-bash-pretooluse-dispatcher.md).

Refs #1317


me2resh (commented):
## Security Review: PR #1320

**Commit**: `2e21a6570a11eca10c919de2840ed913944e683b`

### Summary

This PR replaces the Bash PreToolUse fan-out with a single dispatcher. I reviewed the trust-chain shape, command routing, special environment propagation, and derived adapter impact for Claude Code, Codex, pi, and opencode.

### Checklist Results

- Secrets & Credentials: Pass
- Injection Prevention: Pass
- XSS Prevention: Pass
- Auth & Authorisation: Pass
- Data Protection: Pass
- API Security: Pass
- Gate & Trust-Chain Integrity: Fail

### Security Issues Found

#### MEDIUM - Cross-harness trust-chain regression sensors now fail

`.claude/settings.json:123` now exposes only `dispatch-bash.sh` as the derived Bash gate surface. That can preserve runtime behavior, but the pi and opencode adapters derive their gate tables from `.claude/settings.json`, and their real-settings trust-chain tests still assert that the named merge gates are directly derived.

Observed local failures on this PR head:

- `harness-adapters/pi/test/derive-gates.test.ts:162` fails because `block-unreviewed-merge` is no longer derived from real settings.
- `harness-adapters/pi/test/derive-gates.test.ts:199` then fails while checking the merge-shape globs.
- `harness-adapters/opencode/test/derive-gates.test.ts:252` fails for the same missing named merge gate.
- `harness-adapters/opencode/test/derive-gates.test.ts:280` fails while checking the multi-tracker merge-shape globs.
- `harness-adapters/opencode/test/smoke-block-unreviewed-merge.sh` also fails its expected merge-gate smoke assertion after the dispatcher change.

This is a trust-chain coverage regression. I did see the dispatcher call the merge gate in the opencode smoke path, so I am not claiming a confirmed merge bypass. The blocker is that the PR removes the previous cross-harness proof without replacing it with dispatcher-aware proof for `gh pr merge`, `gh api`, `glab mr merge`, `glab api`, and `tracker_pr_merge`.

Remediation: update the pi/opencode regression tests to treat `dispatch-bash.sh` as the derived Bash gate, then add end-to-end assertions that each merge-shaped command reaches the existing merge gates through the dispatcher and blocks with the expected gate behavior. Keep a negative control where the dispatcher or target hook is absent.

### Recommendations

- Add dispatcher tests for all multi-tracker merge shapes, not only `gh pr merge`.
- Add adapter-level tests that prove pi/opencode still block merge-shaped commands through the dispatcher.
- Consider removing `block-reviewer-repo-mutation.sh` from the dispatcher or from the standalone Bash row if the double-run is not intentional. I did not treat that as a security blocker because it preserves the guard rather than weakening it.

### Verification

Passed locally:

- `bash -n .claude/hooks/dispatch-bash.sh .claude/hooks/tests/test_dispatch_bash.sh`
- `shellcheck .claude/hooks/dispatch-bash.sh .claude/hooks/tests/test_dispatch_bash.sh`
- `bash .claude/hooks/tests/test_dispatch_bash.sh`
- `bash .claude/hooks/tests/test_settings_wrappers_silent_noop.sh`
- `bash .claude/hooks/tests/test_sync_codex_adapter.sh`
- `bash .claude/hooks/tests/test_manage_portfolio_adapters.sh`
- `python3 -m json.tool .claude/settings.json`
- `git diff --check`

Failed locally:

- `npm test` in `harness-adapters/pi`
- `npm test` in `harness-adapters/opencode`
- `bash harness-adapters/opencode/test/smoke-block-unreviewed-merge.sh`

### Verdict

**CHANGES REQUESTED**

Reviewed commit: `2e21a6570a11eca10c919de2840ed913944e683b`

me2resh (commented):
## Code Review: PR #1320

**Commit**: `2e21a6570a11eca10c919de2840ed913944e683b`
**Scope**: Full

### Summary

This PR replaces the large Bash `PreToolUse` hook fan-out with a dispatcher.
It also adds dispatcher tests and records AgDR-0157.

### Checklist Results

- ✅ Architecture & Design: Pass. The dispatcher keeps the existing hook scripts as the policy source.
- ✅ Code Quality: Fail. One hook remains outside the dispatcher and also runs inside it.
- ✅ Testing: Fail. The tests do not assert the settings-level Bash hook count or the duplicate reviewer hook path.
- ✅ Security: Pass with one blocking correctness issue below. The merge and PR gates still route through existing scripts.
- ✅ Performance: Fail. Every Bash call still starts the standalone reviewer-mutation hook after the dispatcher starts it.
- ✅ PR Description & Glossary: Pass. The PR links #1317 and includes a glossary.
- ⚠ Summary Bullet Narrative: Pass. The summary bullets explain the behavior change.
- ✅ Technical Decisions (AgDR): Pass. AgDR-0157 covers the dispatcher decision.
- ✅ Adopter Handbooks: Pass. Loaded `handbooks/architecture/clean-architecture-layers.md`, `handbooks/architecture/migration-safety.md`, and `handbooks/general/commit-message-quality.md`.
- ⚠ Fallow Static Analysis (JS/TS): N/A. The diff does not touch JS or TS files.

### Issues Found

1. `block-reviewer-repo-mutation.sh` still runs twice for every Bash command. Request changes.

   `.claude/settings.json:128` keeps a second `PreToolUse` Bash group that executes `block-reviewer-repo-mutation.sh` directly.
   `.claude/hooks/dispatch-bash.sh:29` also runs that same hook in the dispatcher's unconditional list.

   This breaks the PR's main contract: one Bash dispatcher command, each existing hook at most once.
   It also keeps one extra process on the hot path for every Bash call.
   The Codex adapter inherits the same duplicate because it generates both Bash groups from settings.

   Please fold this hook fully into the dispatcher, or remove it from the dispatcher and document why it stays separate.
   If it stays in the dispatcher, also fix `.claude/hooks/dispatch-bash.sh:14`.
   That line sets `APEXYARD_REVIEW_OPS_ROOT` to the `.claude` directory, not the ops root.
   The old standalone settings wrapper passed the real `$r` ops root.

2. The regression test misses the settings-level duplicate. Request changes.

   `.claude/hooks/tests/test_dispatch_bash.sh:11` proves the dispatcher calls each stubbed hook once.
   It does not parse `.claude/settings.json` and prove that only one Bash `PreToolUse` command remains.

   Add a settings-level assertion for this PR's central invariant.
   A useful check is: all `PreToolUse` entries with `matcher == "Bash"` resolve to one command, and that command is `dispatch-bash.sh`.

### Handbook Findings

None.

### Suggestions

Consider adding a generated mapping test that compares every old `Bash(...)` predicate group against the dispatcher's case table.
That would catch future missing hooks without relying on manual review.

### Verdict

**CHANGES REQUESTED**

---
📌 Reviewed commit: `2e21a6570a11eca10c919de2840ed913944e683b`

me2resh:
Implemented the requested changes in commit `704e0e8`:

- Removed the standalone `block-reviewer-repo-mutation.sh` Bash entry; the dispatcher now runs it once with the correct ops-root environment.
- Added machine-readable dispatcher routing records and taught pi/opencode to derive their gate tables from those records.
- Added a settings-level assertion that exactly one Bash PreToolUse command remains and that the duplicate reviewer entry is absent.
- Added pi end-to-end coverage for all five merge command shapes.
- Extended the dispatcher regression test across all five merge shapes.
- Updated the opencode smoke assertion for the current fail-closed merge-gate reason.

Focused validation now passes:

- `npm test` in `harness-adapters/pi`
- `npm test` in `harness-adapters/opencode`
- `bash harness-adapters/opencode/test/smoke-block-unreviewed-merge.sh`
- dispatcher, settings-wrapper, JSON, shellcheck, and diff checks

me2resh:
Follow-up fix in d9efc17: adapters now exclude the dispatcher entry itself when expanding dispatcher routing, preventing the dispatcher plus individual gates from executing twice. Pi and opencode suites remain green after this change.

me2resh (commented):
## Security Review: PR #1320

**Commit**: `704e0e81ab28d53860ed1f07a583e6498cbda1ba`

### Summary

I re-reviewed the updated dispatcher and adapter changes. The prior Hakim finding is resolved: pi and opencode now derive the dispatcher routing table and the regression tests prove the merge-gate shapes again.

### Checklist Results

- Secrets & Credentials: Pass
- Injection Prevention: Pass
- XSS Prevention: Pass
- Auth & Authorisation: Pass
- Data Protection: Pass
- API Security: Pass
- Gate & Trust-Chain Integrity: Pass

### Security Issues Found

None blocking.

The updated `APEXYARD_DISPATCH_GATE` metadata keeps the named gate view available to pi/opencode after the Bash settings fan-out collapse. The adapter tests now cover the named merge gates and the multi-tracker merge globs. The opencode smoke test also blocks a fabricated merge through the real `block-unreviewed-merge.sh` path and keeps a negative control for missing hook scripts.

### Verification

Passed locally on the updated head:

- `npm test` in `harness-adapters/pi`
- `npm test` in `harness-adapters/opencode`
- `bash .claude/hooks/tests/test_dispatch_bash.sh`
- `bash harness-adapters/opencode/test/smoke-block-unreviewed-merge.sh`

GitHub status at review time: CodeQL, markdownlint, PR title, and shellcheck are green; the hook test suite is still in progress.

### Recommendations

The adapters now derive both the top-level `dispatch-bash.sh` settings row and the dispatcher metadata rows. That preserves blocking behavior, but allowed Bash calls may run advisory hooks twice under pi/opencode. This is not a security blocker for this PR, but it is worth tightening later to avoid duplicate warnings and extra work.

### Verdict

**APPROVED**

Reviewed commit: `704e0e81ab28d53860ed1f07a583e6498cbda1ba`

me2resh (commented):
## Code Review: PR #1320

**Commit**: `d9efc172ed98a730e3726c69b890c374a32a28bf`
**Scope**: Full

### Summary

The requested fixes are resolved. Claude Code now has one Bash `PreToolUse` dispatcher. The pi and opencode adapters derive the dispatcher routes without executing `dispatch-bash.sh` as an extra gate.

### Checklist Results

- ✅ Architecture & Design: Pass. The dispatcher remains the route source, and existing hook scripts remain the decision source.
- ✅ Code Quality: Pass. The duplicate Bash gate path is removed.
- ✅ Testing: Pass. Focused hook, Codex, pi, and opencode adapter tests pass.
- ✅ Security: Pass. Merge, review, and architecture gates still route to the existing hook scripts.
- ✅ Performance: Pass. The hot Bash path no longer keeps an extra standalone reviewer hook.
- ✅ PR Description & Glossary: Pass. The PR links #1317 and includes a glossary.
- ⚠ Summary Bullet Narrative: Pass.
- ✅ Technical Decisions (AgDR): Pass. AgDR-0157 covers the dispatcher decision.
- ✅ Adopter Handbooks: Pass. No handbook violations found.
- ⚠ Fallow Static Analysis (JS/TS): N/A. `fallow` is unavailable in this environment.

### Issues Found

None.

### Suggestions

None.

### Verdict

**APPROVED**

---
📌 Reviewed commit: `d9efc172ed98a730e3726c69b890c374a32a28bf`

me2resh (commented):
CHANGES REQUESTED. Fix the red hook test suite and the dispatcher's fail-fast behavior, then re-request review.

## Code Review: PR #1320

**Commit**: `d9efc172ed98a730e3726c69b890c374a32a28bf`
**Scope**: Full

### Summary

This PR replaces 54 Bash PreToolUse entries in `.claude/settings.json` with one dispatcher command. `.claude/hooks/dispatch-bash.sh` resolves the ops root once. It runs nine unconditional safety hooks. It then selects command-specific hooks by command prefix. A machine-readable comment table records every route. The pi and opencode adapters derive their Bash gate tables from that table. AgDR-0157 records the decision.

The routing is faithful. I compared the 54 (predicate, script) pairs in `origin/dev` settings against the dispatcher table. The two sets are identical. No gate was dropped and no gate was added.

Two defects block this change. The hook test suite is red because of this PR. The dispatcher can also skip later gates when an earlier hook exits with a code other than 0 or 2.

### Checklist Results

- Architecture & Design: ✅ Pass — the dispatcher delegates to the unmodified hook scripts. It does not reimplement policy. The comment table is a single source for the adapters.
- Code Quality: ❌ Fail — `set -euo pipefail` plus sequential `run_hook` calls lets one hook suppress every later gate. See issue 2.
- Testing: ❌ Fail — this PR leaves `test_block_main_push.sh` red. The new dispatcher test also omits the fail-fast case and the special-environment cases. See issues 1 and 2.
- Security: ⚠ Advisory — no secrets, no injection surface. The fail-open path in issue 2 is a trust-chain concern and is reported there.
- Performance: ⚠ Unverified — I did not reproduce the timing. The reported figure is not stated per call. See issue 3.
- PR Description & Glossary: ✅ Pass — the body carries Summary, Why, Testing, Glossary, and Architecture. The Glossary defines four terms.
- Summary Bullet Narrative: ✅ Pass — each bullet has a verb and more than six words.
- Technical Decisions (AgDR): ❌ Fail — AgDR-0157 exists and is linked. It omits two load-bearing consequences of this change. See issue 3.
- Adopter Handbooks: ⚠ Advisory — two advisory findings. No blocking handbook applies. `handbooks/architecture/*` is not applicable, because the diff touches no product architecture layer and no migration path.
- Fallow Static Analysis (JS/TS): N/A — the `fallow` CLI is not on PATH in this environment, so the step did not run.

### Issues Found

**1. Blocking — this PR turns `test_block_main_push.sh` red.**

Observed. The `hook test suite` check on `d9efc17` reports `PASS=156 FAIL=2`. The `origin/dev` baseline run reports `FAIL=1`. The dev failure is `test_agdr_unique_ids.sh` with `duplicate AgDR identifiers: AgDR-0151`. That one predates this branch, because `docs/agdr/AgDR-0151-native-first-cursor-overlay.md` and `docs/agdr/AgDR-0151-nested-ops-root-anchor-precedence.md` both exist on dev.

The second failure is new. `test_block_main_push.sh` fails the case `[settings wiring enables ops-root scope guard]`.

Root cause: `.claude/hooks/tests/test_block_main_push.sh:408` greps `.claude/settings.json` for the literal `exec env APEXYARD_OPS_SCOPE_GUARD=1`. This PR moved that assignment into `run_hook` in `dispatch-bash.sh`. The literal no longer appears in settings, so the control fails.

The guard itself still works. I ran the dispatcher against stub hooks with a push-class command. `block-main-push.sh` received `APEXYARD_OPS_SCOPE_GUARD=1`. No other hook received it. The control is stale, not the behavior.

Required change: update the assertion so it accepts the guard in `.claude/hooks/dispatch-bash.sh`. Keep the control meaningful, because it is the regression proof for #1230. Then run the full suite, not only the four tests named in the Testing list.

This failure also fails ticket #1317's criterion "Hook unit tests for delegated scripts still pass". It fails `.claude/rules/pr-quality.md` § "No Red CI Before Merge". `block-merge-on-red-ci.sh` will block this PR while the check stays red.

**2. Blocking — one hook exiting with an unexpected code suppresses every later gate.**

`dispatch-bash.sh:3` sets `set -euo pipefail`. `run_hook` invokes each hook in sequence in the same shell. Bash aborts the script on the first non-zero exit. Exit code 2 must abort, because 2 is the block signal. Any other non-zero code must not abort, because the later gates still need to run.

Observed. I copied the dispatcher into an isolated fixture with stub hooks. I made `require-active-ticket.sh` exit 1. I sent a staging-class command. Four hooks ran. The dispatcher exited 1. The five remaining unconditional hooks never ran. `block-git-add-all.sh` never ran.

The previous design could not fail this way. Each hook ran as its own PreToolUse process. One crashed hook could not stop another hook.

Impact. Claude Code blocks a call on exit code 2. Other non-zero codes report an error and allow the call. One crashing advisory hook therefore skips the merge gates and allows the command. The nine unconditional hooks run before every command-specific gate, so a crash in that group reaches every routing shape, including the five merge shapes.

Reachability is inferred, not observed. Each unconditional hook uses `set -u` without `set -e`. An unbound variable, an absent `jq`, or a failed `source` yields exit 1. I did not reproduce that state in a real hook.

Required change: separate the two cases in `run_hook`. One shape that works under `set -e`:

```bash
run_hook() {
  local script="$1" rc=0
  case "$script" in
    ...) ... <<<"$INPUT" || rc=$? ;;
  esac
  if [ "$rc" -eq 2 ]; then
    exit 2
  fi
  if [ "$rc" -ne 0 ]; then
    printf 'WARN: %s exited %s. Later gates still run.\n' "$script" "$rc" >&2
  fi
  return 0
}
```

Add a regression case for it. Make an early hook exit 1. Assert that the later gates still run. Assert that a merge-class command still blocks.

**3. Blocking — AgDR-0157 omits two consequences, and its performance figure is not comparable to the acceptance criterion.**

`.claude/rules/writing-standard.md` requires the record to keep the load-bearing facts. `templates/agdr.md` requires the consequences section to state what the decision changes.

Missing consequence 1: this decision moves the Bash gates from many isolated processes into one shared process. That removes failure isolation between hooks. Issue 2 is the direct result.

Missing consequence 2: this decision moves `APEXYARD_OPS_SCOPE_GUARD` and `APEXYARD_REVIEW_OPS_ROOT` out of `.claude/settings.json` and into the dispatcher. Any control that reads settings for those values now needs an update. Issue 1 is the direct result. The AgDR states that existing scripts keep "their exit codes and special environment variables", which reads as "nothing moved".

Measurement gap: criterion 9 of #1317 asks for the measured Bash PreToolUse time for `true`, below the 1.49 s warm baseline from 2026-09-16, on the same class of machine. The AgDR and the PR body report about 10.24 s and about 1.43 s for three sequential calls. That implies about 0.48 s per call. State the per-call figure and the machine class, so a reader can compare the two numbers directly.

### Handbook Findings

⚠ **Commit-message quality** — `handbooks/general/commit-message-quality.md`

- All three commits carry the literal two-character sequence `\n` in place of real newlines before the trailer. Commit `2e21a65` ends `...record the architecture decision in AgDR-0157.\n\nRefs #1317`. The trailer sits inside the paragraph, so a trailer parser cannot read it. Use real newlines. The framework squash-merges, so the effect is limited to branch history.

⚠ **TypeScript Strict Mode** — `handbooks/language/typescript/strict-mode.md`

- `harness-adapters/pi/test/gate-dispatcher.test.ts` adds `registerGateDispatcher(pi as any, ...)` in the new test. The handbook flags `as any` without an inline justification. The adjacent pre-existing test uses the same cast, so this matches the file's convention. No change required.

### Suggestions

1. The dispatcher table carries no environment column. The pi and opencode adapters exec each hook with `execFileSync("bash", [hookPath])` and add no environment. That gap predates this PR. The table is now the adapters' source of routing, so record the gap next to the table or encode the two values in it.
2. `harness-adapters/opencode/test/smoke-block-unreviewed-merge.sh:129` now accepts `could not resolve PR`. That sentence is not unique to `block-unreviewed-merge.sh`. `require-architecture-review.sh:268` and `require-design-review-for-ui.sh:291` print the same sentence, and the dispatcher routes all three against the same five shapes. The FAIL message still claims the text belongs to one hook. Assert the gate name from the adapter's block reason instead.
3. The dispatcher test uses stub hooks only. Add a case that asserts `block-main-push.sh` receives the scope guard and `block-reviewer-repo-mutation.sh` receives the ops root. That case would have caught issue 1 in the dispatcher test instead of in CI.
4. Add a comment in `dispatch-bash.sh` that names the table as the adapters' source. A contributor who edits the `case` block without editing the table silently desynchronises pi and opencode.

### Validation

Commands and inspected evidence:

- `gh pr view 1320` and `gh pr diff 1320`. Read all 13 changed files at `d9efc17` through the contents API.
- Compared (predicate, script) pairs. `origin/dev` settings hold 54 pairs. The dispatcher table holds 54 pairs. The two sets are identical in both directions.
- `.claude/settings.json` at `d9efc17` holds one Bash PreToolUse entry. No Bash entry names `block-reviewer-repo-mutation.sh`. `origin/dev` held 53 entries in one group and 1 in a second group.
- Ran the dispatcher against stub hooks in an isolated fixture. All nine unconditional hooks ran. `APEXYARD_REVIEW_OPS_ROOT` resolved to the fixture ops root, not to `.claude`. `APEXYARD_OPS_SCOPE_GUARD=1` reached only `block-main-push.sh`.
- Ran the same fixture with one early hook exiting 1. The dispatcher exited 1 and skipped every later hook.
- Read `block-reviewer-repo-mutation.sh:21-33`. It validates `APEXYARD_REVIEW_OPS_ROOT` against `$ROOT/.claude/session` and falls back to a walk-up. It uses the value only to locate the active-reviewer marker, so a physical path is safe.
- `gh pr checks 1320`. Read the failing `hook test suite` job log and the `origin/dev` job log.
- Read `bin/sync-codex-adapter.sh:167-227`. The generator walks every matcher group and preserves `matcher`. The dispatcher entry carries no `if` predicate, so no predicate is lost. `assert_hook_counts_match` compares totals and still matches.
- Read `.cursor/hooks.json`. It wires only `cursor-session-pin.sh`. The overlay carries no gate list, so this change does not affect it.
- Read the three prior comments on this PR and the diff of every file they claim to have fixed.

Verification limits:

- I did not run a live Claude Code session. Parity between Claude Code's own `if: "Bash(<glob>)"` matcher and the dispatcher's `case` prefix patterns is inferred from the identical pattern strings, not observed. Both forms require the same command prefix, so an environment-prefixed or chained command matched neither before and matches neither now.
- I did not run Conformance CI. It runs on a daily schedule, not per PR, and it did not run on this commit. Criterion 5 of #1317 is unverified for `d9efc17`.
- I did not run the pi or opencode Node test suites. The author reports both green. I did not repeat that check.
- Exit-1 reachability inside a real unconditional hook is inferred from `set -u` usage, not reproduced.
- The `fallow` CLI is absent here, so no JS/TS static-analysis findings are reported.

### Re-review reconciliation — findings from the `2e21a65` pass

Resolved:

1. Duplicate reviewer-mutation entry. Settings now hold one Bash entry. The dispatcher runs the hook once. Observed.
2. `APEXYARD_REVIEW_OPS_ROOT` value. `run_hook` passes `$HOOK_DIR/../..`, which is the ops root. Observed in the fixture run.
3. Adapter gate derivation. `derive-gates-core.ts` parses the table. pi and opencode translate it into their tool names. Observed in the diff.
4. Nested dispatch. Both adapters filter `.claude/hooks/dispatch-bash.sh` out of the settings-derived set, so the dispatcher no longer runs as an extra gate. Observed in the diff.
5. Merge-shape coverage. The dispatcher test and the new pi test cover all five shapes. Observed.

Remaining or new:

6. `test_block_main_push.sh` is red. New in this PR. Issue 1.
7. Fail-fast aggregation. Present since `2e21a65` and still open. Issue 2.
8. AgDR completeness and the performance figure. Issue 3.

### Verdict

**CHANGES REQUESTED**

The routing table is faithful to the 54 entries it replaces, and the adapter derivation is sound. Two defects block the merge. CI is red because of this PR, and the dispatcher can skip gates on an unexpected hook exit code. The AgDR omits the two consequences that produced both defects. The diff touches production hooks and `.claude/settings.json`, so this change stays Heavy tier and ambiguity rounds up. No approval marker was written.

---
📌 Reviewed commit: `d9efc172ed98a730e3726c69b890c374a32a28bf`

me2resh (commented):
Rex requests changes on `28fc551`. Fix four items before the next review. The dispatcher logic itself is sound and the three prior blocking findings are resolved. Two controls around the dispatcher lost coverage, and two claims in the record are inaccurate or incomplete.

## Code Review: PR #1320

**Commit**: `28fc55197171849aa8ea9a63df36454fae1253fc`
**Scope**: Full — Heavy tier. The diff touches production `.claude/hooks/*.sh`, `.claude/settings.json`, and adapter gate derivation. Rail 1 of `.claude/rules/right-size-ceremony.md` forbids a Lean pass here.

### Summary

This PR replaces the 54 Bash `PreToolUse` entries in `.claude/settings.json` with one dispatcher command. `dispatch-bash.sh` reads the Bash command from the hook payload. It runs nine unconditional safety hooks. It then runs the command-specific hooks for the matched command prefix. A machine-readable `APEXYARD_DISPATCH_GATE` comment table lets the pi and opencode adapters derive the same routing without executing the dispatcher as a nested gate. AgDR-0157 records the decision. The PR also renames a pre-existing duplicate AgDR identifier.

### Prior-review reconciliation

All three blocking findings from the `d9efc17` pass are resolved.

1. **Scope-guard regression — RESOLVED.** `test_block_main_push.sh:405` now asserts the guard in `.claude/hooks/dispatch-bash.sh` instead of `settings.json`. The full hook suite is green on this SHA. The control keeps its previous strength. Both the old and the new assertion are bare substring greps, so neither proves the guard reaches only `block-main-push.sh`. That limitation is pre-existing and is not a regression. See the suggestion below.
2. **Failure isolation — RESOLVED.** `run_hook` captures the return code, exits 2 on a block, and warns on any other non-zero code. I reproduced the semantics with a standalone fixture built from the committed `run_hook` body. Exit codes 1 and 7 produced a warning and the run continued to the last hook. Exit code 2 aborted the run with exit 2. `test_dispatch_bash.sh:223-232` adds the matching regression case and asserts that `block-unreviewed-merge.sh` and `require-architecture-review.sh` still run after an earlier hook exits 1.
3. **AgDR-0157 gaps — RESOLVED as scoped.** The Consequences section now records the changed failure isolation and the dispatcher's ownership of `APEXYARD_OPS_SCOPE_GUARD` and `APEXYARD_REVIEW_OPS_ROOT`. The timing figure is now per call: about 3.41 s per call before and about 0.48 s per call after. That figure is below AC 9's 1.49 s warm baseline. The record states plainly that the original measurement did not capture a machine class. Keeping that limit visible is correct.

The four items from the `2e21a65` pass remain resolved.

- **Adapter derivation.** `deriveGatesFromDispatcher` parses the routing table. I ran the PR's own `_shared/derive-gates-core.ts` against the PR's `dispatch-bash.sh` and it produced 31 gates. `block-unreviewed-merge` carries all five merge shapes: `gh pr merge *`, `gh api *`, `glab mr merge *`, `glab api *`, and `tracker_pr_merge *`.
- **`APEXYARD_REVIEW_OPS_ROOT`.** `dispatch-bash.sh:20` sets it to `$HOOK_DIR/../..`, which is the ops root, not `.claude`. The consumer validates the value by testing for `$ROOT/.claude/session`, so the physical path from `pwd -P` is accepted.
- **Duplicate reviewer-mutation entry.** Settings at this SHA carry exactly one Bash `PreToolUse` entry, and it is the dispatcher. `test_dispatch_bash.sh:178-182` asserts both invariants.
- **Nested dispatcher execution.** Both adapters filter `.claude/hooks/dispatch-bash.sh` out of the settings-derived set before appending the dispatcher-derived gates.

### Checklist Results

- Architecture & Design: ✅ Pass — one dispatcher over unmodified hook scripts. The `APEXYARD_DISPATCH_GATE` table gives the adapters a portable routing source and avoids a second policy implementation.
- Code Quality: ✅ Pass — `set -euo pipefail` with explicit return-code capture. Shellcheck passes on this SHA.
- Testing: ❌ Fail — see Issues 1 and 2. A stale adapter test and a narrowed exec-bit control.
- Security: ❌ Fail — see Issue 1. The exec-bit control no longer covers 26 delegated gate scripts.
- Performance: ✅ Pass — about 0.48 s per call against a 1.49 s warm baseline, documented per call in AgDR-0157.
- PR Description & Glossary: ❌ Fail — the Glossary is present and correct. The Summary asserts preserved ordering, which is inaccurate, and omits the AgDR rename. See Issues 3 and 4.
- Summary Bullet Narrative: ✅ Pass — every bullet carries a verb and states the effect.
- Technical Decisions (AgDR): ⚠ Advisory — AgDR-0157 exists, is linked, and is material. It omits the ordering consequence. See Issue 3.
- Adopter Handbooks: ⚠ Advisory — one TypeScript strict-mode observation. No blocking handbook applies.
- Fallow Static Analysis (JS/TS): N/A — the `fallow` CLI is not installed in this environment, so the step did not run.

### Issues Found

**1. Blocking — the exec-bit control loses coverage of 26 delegated gate scripts.**

`.claude/hooks/tests/test_hook_exec_bits.sh:50` builds its hook set by grepping `settings.json` for `.claude/hooks/<name>.sh`. Collapsing the fan-out removes those paths from the file the control reads.

Measured on this SHA against `origin/dev` (`8e12865`):

- `dev` settings name 55 distinct hooks. This SHA's settings name 30.
- 26 hooks lose coverage, including `block-unreviewed-merge.sh`, `block-git-add-all.sh`, `block-main-push.sh`, `check-secrets.sh`, `block-merge-on-red-ci.sh`, `require-architecture-review.sh`, and `block-reviewer-repo-mutation.sh`.
- `dispatch-bash.sh` still execs all 26 directly by path.

The consequence combines with the new failure isolation. A direct exec of a non-executable file returns 126, which I confirmed in this shell. `run_hook` now treats 126 as non-blocking, prints a warning, and continues. The gate stops enforcing and the Bash call proceeds. This is the exact failure mode issue #1008 created that guard to prevent, and the guard no longer watches these paths.

Required: extend the extraction so the control also reads the dispatcher's hook set. The `APEXYARD_DISPATCH_GATE` comments already provide a machine-readable list. Add a case that asserts `dispatch-bash.sh` itself is `100755` in the index, because a lost bit on the dispatcher removes every Bash gate at once.

**2. Blocking — `harness-adapters/_shared/test/derive-gates-core.test.ts` is stale.**

That test asserts nine gate names are derivable from the repository's real `settings.json`. The PR updated the same assertion in `harness-adapters/pi/test/derive-gates.test.ts` and `harness-adapters/opencode/test/derive-gates.test.ts`. It did not update the `_shared` copy.

I ran the PR's own `_shared/derive-gates-core.ts` against the PR's `settings.json`. Six of the nine expected names are gone: `block-unreviewed-merge`, `block-merge-on-red-ci`, `require-design-review-for-ui`, `require-architecture-review`, `check-secrets`, and `block-private-refs-in-public-repos`. Only `require-active-ticket`, `require-migration-ticket`, and `suggest-mcp-search` survive, because non-Bash matchers still wire them. The assertion therefore fails on this SHA. Unioning the dispatcher table restores all nine names, which I also verified.

No CI job hides this by running it. `tests.yml` runs `bin/run-hook-tests.sh`, which discovers tests only under `.claude/hooks/tests`, `.claude/agents/tests`, `.claude/rules/tests`, and `.claude/skills`. `conformance.yml` installs the adapters but runs no `npm test`, and it triggers on a schedule, not per PR. So this failure is latent, and the two adapter tests this PR did update are also unverified by CI on this SHA.

Required: union `deriveGatesFromDispatcher` into the `_shared` test as the two sibling tests do. Then list the adapter test commands you ran in the PR's Testing section.

**3. Blocking — the PR claims preserved ordering, and AgDR-0157 omits the ordering change.**

The Summary states that the change preserves "the existing hook scripts, ordering, exit codes, and special environment variables". Commit `2e21a65` repeats the claim. The ordering claim does not hold.

I reconstructed the removed order from the diff. Seven unconditional hooks previously ran after command-specific gates and now run before all of them: `require-skill-for-issue-create.sh` (old position 15), `require-migration-ticket.sh` (48), `require-active-ticket.sh` (49), `suggest-mcp-search.sh` (50), `warn-review-marker-write.sh` (51), `warn-isolated-build-risk.sh` (53), and `block-reviewer-repo-mutation.sh` (54). Relative order inside each command group is preserved.

No gate is suppressed and no allow or block outcome changes, because the first exit 2 still wins. The surfaced block reason can change. A `gh pr merge` that trips both `require-active-ticket.sh` and `block-unreviewed-merge.sh` previously reported the merge gate first and now reports the ticket gate first.

Required: correct the ordering claim in the PR body and record this consequence in AgDR-0157's Consequences section. This is the same standard the prior pass applied to the two omitted consequences.

**4. Blocking — the PR body does not describe the AgDR rename.**

The diff renames `docs/agdr/AgDR-0151-nested-ops-root-anchor-precedence.md` to `docs/agdr/AgDR-0158-nested-ops-root-anchor-precedence.md`. The Summary does not mention it and the Testing section does not cite the control it unblocks.

I verified the change itself is correct and in scope. `origin/dev` carries two `AgDR-0151` files, so the duplicate is pre-existing. `.claude/skills/decide/tests/test_agdr_unique_ids.sh` keys on the filename prefix and fails while both exist, and `bin/run-hook-tests.sh` discovers that test. The rename keeps the record: git reports a 97% similarity rename and the only content change is the H1. No file in `origin/dev` references the old filename, so no link dangles.

Required: add a Summary bullet naming the rename, the duplicate it resolves, and the test it unblocks. Cite `bin/run-hook-tests.sh` in Testing, because that is the suite the prior pass asked you to run.

### Handbook Findings

⚠ **TypeScript Strict Mode** — `handbooks/language/typescript/strict-mode.md`

- `harness-adapters/pi/test/gate-dispatcher.test.ts:889` uses `registerGateDispatcher(pi as any, ...)` with no inline justification, which matches trigger 2. The same file already carries 12 identical call shapes on `origin/dev`, so the new line follows the established convention. No change is required. Fix the convention in a separate task if you want the pattern gone.

`handbooks/general/commit-message-quality.md` applies to these commits and reports no violation. None of its four triggers matches. `handbooks/architecture/clean-architecture-layers.md` and `handbooks/architecture/migration-safety.md` are not applicable, because the diff carries no application or domain layer code and no migration path.

### Suggestions

1. Strengthen the scope-guard assertion. Assert that `APEXYARD_OPS_SCOPE_GUARD=1` appears in the `block-main-push.sh` branch of `run_hook`, not anywhere in the file. A future global export would set the guard for every hook and still pass today's grep.
2. Cite AC 9's 1.49 s warm baseline by number in AgDR-0157. The record states 0.48 s per call but never names the figure it must beat.
3. Align AgDR-0157 with `templates/agdr.md`. Add the "In the context of … I decided … accepting …" summary line that AgDR-0152 and AgDR-0153 carry. Consider `## Options Considered` and `## Artifacts` for the heading names.
4. Update the stale gate counts this PR invalidates. `.cursor/rules/apexyard.mdc:15`, `docs/cursor-adapter.md:105`, AgDR-0091, and AgDR-0151 all cite "86 Claude Code gate entries". Settings now wire 24 hook commands.
5. Make the updated adapter tests mirror production. They compose `deriveGatesFromSettings` and the dispatcher table without the `dispatch-bash.sh` filter that `deriveGatesFromOpsRoot` applies, so they exercise a gate list production never builds.

### Validation

Commands and results on this SHA:

- `gh pr view 1320 --json headRefOid` confirms HEAD is `28fc551`.
- `gh pr checks 1320`, re-run for this review: Analyze, CodeQL, Verify Ticket ID, hook test suite, markdownlint-cli2, and shellcheck all pass.
- I read `dispatch-bash.sh`, `test_dispatch_bash.sh`, `settings.json`, and AgDR-0157 at `28fc551` through the contents API. I read the adapter changes from `gh pr diff`.
- I parsed `settings.json` at `28fc551`. It wires exactly one Bash `PreToolUse` entry.
- I ran the PR's `_shared/derive-gates-core.ts` against the PR's `settings.json` and `dispatch-bash.sh`. Results appear in Issue 2 and in the reconciliation section.
- I reproduced `run_hook`'s return-code semantics with a standalone fixture. Codes 1 and 7 warned and continued. Code 2 aborted with exit 2. A non-executable target returned 126.
- I compared the old `if: "Bash(<glob>)"` predicate against the new `case` prefix for six command shapes. Both matched `git add -A` and both skipped `FOO=1 git add -A`, a bare `git add`, `cd x && git add -A`, `git addendum`, and a leading-whitespace variant. The env-prefix bypass is pre-existing and this PR does not widen it.
- I compared `test_hook_exec_bits.sh` extraction against `origin/dev` and this SHA. Counts appear in Issue 1.
- I inspected `bin/sync-codex-adapter.sh`. Entries without an `if` predicate pass through and still exec `.claude/hooks/*.sh`, so Codex reaches the unmodified hooks through the dispatcher. Its count-parity guard compares generated hooks against source hooks, so the smaller source count does not trip it. This is inferred from the generator, not from a Codex run.
- I inspected `bin/sync-cursor-adapter.sh` and `bin/install-cursor-adapter.sh`. Neither copies the gate list, so the Cursor overlay is unaffected.
- I confirmed `bin/extract-subpacks.sh` does not copy `settings.json`, so the reduced safety-hooks subpack gains no dispatcher wiring.

Verification limits:

- **AC 5 is Unverified.** Conformance CI triggers on a schedule and on manual dispatch, not on a pull request. I observed no conformance run for `28fc551`. I did not drive opencode, pi, or Codex. Treat the cross-harness claim as unproven on this SHA until a conformance run covers it.
- **The adapter TypeScript tests are Unverified.** No CI job runs them, and I did not install adapter dependencies. My findings on them come from running the derivation logic directly, not from `npm test`.
- CI status is mutable. Re-check before merging.
- `git fetch` is unavailable to a review-class agent, so I read PR content through the GitHub API and compared against the locally available `origin/dev` at `8e12865`, which matches the current remote `dev`.

### Verdict

**CHANGES REQUESTED**

The dispatcher design is correct and the three prior blocking findings are fixed. Two mechanical controls in the trust chain now cover less than they did before this change, and neither loss is visible in CI. Issue 1 lets a delegated gate fail open without any guard noticing. Issue 2 leaves a parity test asserting names its input can no longer produce. Issues 3 and 4 leave the durable record stating something the code does not do and omitting a change the diff makes. Fix all four and request a re-review on the new SHA.

---
📌 Reviewed commit: `28fc55197171849aa8ea9a63df36454fae1253fc`

me2resh (commented):
## Code Review: PR #1320

**Commit**: `badbe184c78906214fa8a4b49d30fae9e3377553`
**Scope**: Full

### Summary

The duplicate dispatcher path is fixed, and the normal-path tests pass. One trust-chain regression remains: `dispatch-bash.sh` fails open when its initial `jq` command parse cannot run.

### Checklist Results

- ✅ Architecture & Design: Fail. The dispatcher now owns a precondition that each merge gate previously handled fail-closed.
- ✅ Code Quality: Pass. The routing table and adapter derivation are clear.
- ✅ Testing: Fail. The tests do not cover a broken or missing `jq` before dispatcher routing.
- ✅ Security: Fail. A merge-shaped Bash command can bypass all merge gates when dispatcher command parsing fails.
- ✅ Performance: Pass. The Bash fan-out collapse is preserved.
- ✅ PR Description & Glossary: Pass. The PR links #1317 and defines dispatcher terms.
- ⚠ Summary Bullet Narrative: Pass.
- ✅ Technical Decisions (AgDR): Pass. AgDR-0157 covers the dispatcher design and trade-off.
- ✅ Adopter Handbooks: Pass. No handbook violations found.
- ⚠ Fallow Static Analysis (JS/TS): N/A. `fallow` is unavailable in this environment.

### Issues Found

1. `dispatch-bash.sh` must fail closed when it cannot parse the Bash command. Request changes.

   `.claude/hooks/dispatch-bash.sh:7` runs `jq` under `set -e` before any hook executes. If `jq` is missing or broken, the dispatcher exits `127`. The framework contract treats only exit `2` as a hard block for Claude Code hooks. The pi and opencode adapters document the same contract.

   I reproduced the failure with a `jq` stub that exits `127` and a merge-shaped payload:

   ```text
   printf '{"tool_name":"Bash","tool_input":{"command":"gh pr merge 42"}}' | PATH="<broken-jq>:$PATH" .claude/hooks/dispatch-bash.sh
   rc=127
   ```

   No delegated merge gate runs in that path. Before this PR, `block-unreviewed-merge.sh`, `block-merge-on-red-ci.sh`, `require-design-review-for-ui.sh`, and `require-architecture-review.sh` each parsed the raw payload themselves and intentionally blocked merge-shaped commands on broken `jq`. `docs/security/threat-model.md:165` records that as the designed control for missing `jq` at merge time.

   The dispatcher should parse defensively. If `jq` fails and the raw payload looks merge-shaped, it must exit `2` or otherwise route the raw payload to the merge gates so their existing fail-closed fallback runs. Add a regression test for broken `jq` plus `gh pr merge`, `gh api .../pulls/.../merge`, `glab`, and `tracker_pr_merge` shapes.

### Handbook Findings

None.

### Suggestions

The normal-path checks I ran passed: dispatcher syntax and ShellCheck, `test_dispatch_bash.sh`, settings-wrapper smoke, pi adapter tests, opencode adapter tests, settings JSON parse, and `git diff --check`.

### Verdict

**CHANGES REQUESTED**

---
📌 Reviewed commit: `badbe184c78906214fa8a4b49d30fae9e3377553`

me2resh (commented):
## Code Review: PR #1320

**Commit**: `951edfd1f69b04f41a0b9573e5544acc80050c59`
**Scope**: Full

### Summary

APPROVED. The prior jq fail-closed finding is resolved. When the dispatcher cannot parse the Bash command, it now routes the raw payload to the merge gates, which block merge-shaped commands with exit 2.

### Checklist Results

- ✅ Architecture & Design: Pass. Policy still lives in the existing hook scripts. The dispatcher owns routing only.
- ✅ Code Quality: Pass. The adapter duplicate path is filtered out for pi and opencode.
- ✅ Testing: Pass. Dispatcher, exec-bit, block-main-push, Codex, pi, opencode, shared derivation, and AgDR ID checks pass.
- ✅ Security: Pass. Broken `jq` on a merge-shaped Bash payload now blocks through the merge gate fallback.
- ✅ Performance: Pass. The Bash fan-out collapse is preserved.
- ✅ PR Description & Glossary: Pass. The PR links #1317 and includes the required glossary.
- ⚠ Summary Bullet Narrative: Pass.
- ✅ Technical Decisions (AgDR): Pass. AgDR-0157 covers the dispatcher decision and fail-closed behavior.
- ✅ Adopter Handbooks: Pass. No handbook violations found.
- ⚠ Fallow Static Analysis (JS/TS): N/A. `fallow` is unavailable in this environment.

### Issues Found

None.

### Handbook Findings

None.

### Suggestions

None.

### Verdict

**APPROVED**

---
📌 Reviewed commit: `951edfd1f69b04f41a0b9573e5544acc80050c59`

atlas-apex (approved):

